PR #321, opened 2014-11-20
STORM-563. Kafka Spout doesn't pick up from the beginning of the queue unless forceFromStart specified.

harshach:


kishorvpatil:
@harshach would you like to provide update on this?

harshach:
@kishorvpatil  this JIRA will be addressed as part of STORM-650 refactoring.

PR #328, opened 2014-11-25
[STORM-570] replace table sorter with data table.  Also upgraded several ...

revans2:
...javascript libraries to latest versions to make integration simpler.

I know this is kind of a lot. I upgraded to the latest version of several libraries including jquery and bootstrap.  Then I replaced table sorter with Data Table and a prebuild bootstrap integration.  I configured all of the tables to use DataTable.  I enabled pagination and filtering for the tables that make since, and added state storage for all tables.  The main index's state is stored in HTML5 local storage, but the other pages have is stored in the HTML 5 SessionStorage so that they are cleaned up eventually, but can remain specific to the topology/component.

The UI looks a bit different, mostly because the default theme for bootstrap changed.


kishorvpatil:
+1

kishorvpatil:
@revans2 could you please upmerge the change again?

kishorvpatil:
@revans2 Thank you. I have merged in the changes. Sorry for delay.

PR #336, opened 2014-12-03
STORM-583: Initiali check-in for storm-eventhubs.

shanyu:
Signed-off-by: Shanyu Zhao <shzhao@microsoft.com>

### DIFF
--- a/external/storm-eventhubs/README.md
+++ b/external/storm-eventhubs/README.md
@@ -1,0 +1,37 @@
+storm-eventhubs
+=====================
+
+Storm spout and bolt implementation for Microsoft Azure Eventhubs
+
+### build ###
+	mvn clean package
+
+### run sample topology ###
+To run the sample topology, you need to modify the config.properties file with
+the eventhubs configurations. Here is an example:
+
+	eventhubspout.username = [username: policy name in EventHubs Portal]
+	eventhubspout.password = [password: shared access key in EventHubs Portal]
+	eventhubspout.namespace = [namespace]
+	eventhubspout.entitypath = [entitypath]
+	eventhubspout.partitions.count = [partitioncount]
+
+	# if not provided, will use storm's zookeeper settings
+	# zookeeper.connectionstring=zookeeper0:2181,zookeeper1:2181,zookeeper2:2181
+
+	eventhubspout.checkpoint.interval = 10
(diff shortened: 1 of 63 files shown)


sasincj:
@ptgoetz Is storm-eventhubs available to use?

harshach:
@sasincj its in the process of getting into storm/external. But you can use by going into the shanyu's branch directly

harshach:
@ptgoetz  @revans2  I am +1 on merging this in

revans2:
@sasincj 
I don't understand the EventHub system well enough to really be able to judge if the code is functioning properly in corner cases, etc.  I get at a high level what it does (http://azure.microsoft.com/en-us/services/event-hubs/) but I have never used it myself.  The code itself looks OK, but I am not comfortable voting on it either way with my current knowledge.  

I am a bit nervous about pulling this in as we don't have a very good way to test it beyond the unit tests. EventHubs is a Microsoft Azure specific system.  I cannot spin up my own instance of it to validate that the code is doing what I expect prior to a release.  Although it may be a bit painful, I can do that for Kafka, Redis, HDFS, and HBase integration that we have accepted in so far.  That doesn't mean we cannot/should not pull it in.  To me it mostly means that we need more assurances and support prior to doing so.

sasincj:
@revans2 Without storm-eventhubs, Developers who are using storm with eventhub has to write their own Spout to receive the events. So I thought it would save developer's time when we have it as storm's add-on. For now I'll use it from @shanyu 's branch and let you know my experience...

ptgoetz:
+1 for accepting this as a code donation from Microsoft and moving forward with the IP clearance process.

COMMTTERS please note that this must not be merged until the IP clearance process is completed successfully. I am working with @shanyu to complete those requirements.

ptgoetz:
@shanyu Also, once the source code tarball for IP clearance has been created, please don't add any commits to this pull request. Any necessary modifications can take place later after IP clearance is complete.

shanyu:
@ptgoetz Got it. I will keep the pull request intact during the IP clearance phase. Thanks a lot!

shanyu:
@ptgoetz It has been a while, is the IP clearance done? Thx

ptgoetz:
Thanks @shanyu. I merged this into master.

PR #349, opened 2014-12-15
STORM-593: remove endpoint-socket-lock for worker-data

tedxia:
PR for [STORM-593](https://issues.apache.org/jira/browse/STORM-593)

cached-node+port->socket in worker-data is atom, there is no need of rwlock endpoint-socket-lock to protect cached-node+port->socket. And after use rwlock, there will be competition between refresh-connections and message send.

### DIFF
--- a/storm-core/src/clj/backtype/storm/daemon/worker.clj
+++ b/storm-core/src/clj/backtype/storm/daemon/worker.clj
@@ -232,7 +232,6 @@
       :task->component (HashMap. (storm-task-info topology storm-conf)) ; for optimized access when used in tasks later on
       :component->stream->fields (component->stream->fields (:system-topology <>))
       :component->sorted-tasks (->> (:task->component <>) reverse-map (map-val sort))
-      :endpoint-socket-lock (mk-rw-lock)
       :cached-node+port->socket (atom {})
       :cached-task->node+port (atom {})
       :transfer-queue transfer-queue
@@ -302,9 +301,8 @@
                            port)
                           ]
                          )))
-              (write-locked (:endpoint-socket-lock worker)
-                (reset! (:cached-task->node+port worker)
-                        (HashMap. my-assignment)))
+              (reset! (:cached-task->node+port worker)
+                (HashMap. my-assignment))
               (doseq [endpoint remove-connections]
                 (.close (get @(:cached-node+port->socket worker) endpoint)))
               (apply swap!
@@ -335,16 +333,14 @@
         drainer (TransferDrainer.)
         node+port->socket (:cached-node+port->socket worker)
         task->node+port (:cached-task->node+port worker)
-        endpoint-socket-lock (:endpoint-socket-lock worker)
         ]
     (disruptor/clojure-handler
       (fn [packets _ batch-end?]
         (.add drainer packets)
         
         (when batch-end?
-          (read-locked endpoint-socket-lock
-            (let [node+port->socket @node+port->socket]
-              (.send drainer node+port->socket)))
+          (let [node+port->socket @node+port->socket]
+            (.send drainer node+port->socket))
           (.clear drainer))))))
 
 (defn launch-receive-thread [worker]


storm-core/src/clj/backtype/storm/daemon/worker.clj
```diff
@@ -302,9 +301,8 @@
                            port)
                           ]
                          )))
-              (write-locked (:endpoint-socket-lock worker)
-                (reset! (:cached-task->node+port worker)
-                        (HashMap. my-assignment)))
```
d2r: The only other place `:cached-task->node+port` is used, it is only a read in `mk-transfer-fn`.  Since it is already an atom, we probably do not need this lock.

storm-core/src/clj/backtype/storm/daemon/worker.clj
```diff
@@ -335,16 +333,14 @@
         drainer (TransferDrainer.)
         node+port->socket (:cached-node+port->socket worker)
```
d2r: There is a write-lock around `:cached-node+port->socket`, but we do not use it here in this function after binding it.

storm-core/src/clj/backtype/storm/daemon/worker.clj
```diff
@@ -335,16 +333,14 @@
...
     (disruptor/clojure-handler
       (fn [packets _ batch-end?]
         (.add drainer packets)
         
         (when batch-end?
-          (read-locked endpoint-socket-lock
-            (let [node+port->socket @node+port->socket]
-              (.send drainer node+port->socket)))
+          (let [node+port->socket @node+port->socket]
+            (.send drainer node+port->socket))
```
d2r: @nathanmarz, I could be missing something, but I do not think we need a read lock here to protect against the send:

There is no corresponding write-lock around `:cached-node+port->socket`.  (I do not think we need one anyway, since it is updated via a `swap!` in `mk-refresh-connections`, and that should be atomic.)

nathanmarz:
Without the r/w lock the following can happen:

1. Sender S grabs a socket to worker A for task B
2. Refresh connections learns task B is now on worker C, changes the connection, and closes the old socket
3. Sender S tries to send data on a closed socket

An atom does not protect against this.

storm-core/src/clj/backtype/storm/daemon/worker.clj
d2r: OK, I am convinced we do need the lock here.  The lock is not so much protecting `:cached-node+port->socket` as it is guaranteeing (along with the read-lock below) that anyone accessing the atom after we enter this lock will get the data in new HashMap that has been swapped into it.

In fact, it seems there could still be a race we send to an endpoint after it is closed, but before it is removed.  We might want to extend the write lock to protect readers getting a mapping with non-null, closed connections:

```Diff
               (write-locked (:endpoint-socket-lock worker)
                 (reset! (:cached-task->node+port worker)
-                        (HashMap. my-assignment)))
-              (doseq [endpoint remove-connections]
-                (.close (get @(:cached-node+port->socket worker) endpoint)))
-              (apply swap!
-                     (:cached-node+port->socket worker)
-                     #(HashMap. (apply dissoc (into {} %1) %&))
-                     remove-connections)
+                        (HashMap. my-assignment))
+                (doseq [endpoint remove-connections]
+                  (.close (get @(:cached-node+port->socket worker) endpoint)))
+                (apply swap!
+                       (:cached-node+port->socket worker)
+                       #(HashMap. (apply dissoc (into {} %1) %&))
+                       remove-connections))
```
It is OK if the packets are sent to a null IConnection, since TransferDrainer does a [null check](https://github.com/apache/storm/blob/36e99fa2dfdd13cd43d8fa8c558c670cd7750ed0/storm-core/src/jvm/backtype/storm/utils/TransferDrainer.java#L50).  But it is not OK to try and send with a closed, but non-null connection.

Maybe it would be good to add documentation near `write-locked` explaining that it is a barrier to anyone else reading from the node+port->socket mapping.  Otherwise, it is not clear how it is helping.

storm-core/src/clj/backtype/storm/daemon/worker.clj
nathanmarz: Taking a look at it again, it looks like the introduction of TransferDrainer is a regression. The way that class works doesn't make a whole lot of sense... it has separate add and send methods but those are only ever used in the same method. Looking up the node+port for the task should happen in the read lock as well.

nathanmarz:
Taking a look at it again, it looks like the introduction of TransferDrainer is a regression. The way that class works doesn't make a whole lot of sense... it has separate add and send methods but those are only ever used in the same method. Looking up the node+port for the task should happen in the read lock as well.

The send code should be updated to fix this.

storm-core/src/clj/backtype/storm/daemon/worker.clj
d2r: We do need this. See [comment](https://github.com/apache/storm/pull/349/files#r27413373)

d2r:
> The send code should be updated to fix this.

@nathanmarz, should we also extend the write lock? [comment](https://github.com/apache/storm/pull/349#discussion_r27413373)

nathanmarz:
Looking inside task->node+port needs to be done inside the read lock (as it was done before this TaskDrainer class got inserted here). If that's done, there's no possible way for the sender to get its hands on a connection that's closed or going to be closed while it's using it.

To elaborate on that last point: once the write lock code is finished, all entries in task->node+port are up to date and accurate, and no connections for those node+ports will be closed. This means it's impossible for the code in read-locked to get a to-be-closed connection since it gets everything by looking at task->node+port first.

The write-locked code was carefully written this way to minimize the amount of work that actually happens in the write lock. This is important because when that code is active, tasks are prevented from sending messages. The way it is now, it's just setting the value of a single in-memory atom – so basically no work at all. If it were making/closing connections, it would be a completely different story.

In short, the code in the write-lock is fine – it's the code in the read lock that needs to be fixed. As part of that, the code looking up the node+port for a task needs to be moved back to this function and not happen before the tuple goes on the transfer queue.

storm-core/src/clj/backtype/storm/daemon/worker.clj
d2r: We should not extend the write lock, we should check tasks->node+port in the read lock when we are sending. See [comment](https://github.com/apache/storm/pull/349#issuecomment-87778672)

d2r:
Yes, I see before 861a92eab8740cfc0f83ac4d7ade9a2ab04a8b9b we checked task->node+port.

I'll file an issue to track, since it is a different problem/fix than covered by this JIRA issue and pull request.

d2r:
[STORM-737](https://issues.apache.org/jira/browse/STORM-737)

PR #352, opened 2014-12-19
[STORM-591] exclude logback.xml from jar

Lewuathe:
`logback.xml` should not be included jar file because it will override other configuration who use your jar. 
To avoid this case exclude `logback.xml` file from storm-core jar file.


Lewuathe:
I found logback searches logback-test.xml first and at eventually I renamed xml file and excluded it when it is packaged. I think it is a better way because it can show this xml file is only used for test explicitly.

d2r:
I like the change, but I think now we need to update for log4j2. @Lewuathe would you be willing to convert the syntax for log4j2?

d2r:
@Lewuathe ,
The file has been converted to log4j2.xml, so all we need to do get your branch upmerged.  Would you take a look?

knusbaum:

PR #354, opened 2014-12-19
Storm-166: Nimbus HA design doc and implementation.

Parth-Brahmbhatt:
I have deleted the bit torrent implementation from this pull request as the only available bit torrent library does not support tracker less torrents. In absence of tracker less torrents a single tracker becomes a single point of failure and a multi tracker implementation requires that if a tracker host fails the replacement host has same dns/network configuration. 

Some manual tests I executed:
* start 3 nimbuses, test simple word count topology works. try storm list/activate/deactivate/rebalance/kill from ui and CLI.
* set the replication factor to 2 run the first test again.
* bring up a new nimbus, ensure it catches up and competes for leader lock.
* with 3 nimbuses and 2 topologies, delete one topology code from each non leader nimbus. After killing master nimbus, ensure one of them eventually becomes leader.


Parth-Brahmbhatt:
@revans2 @ptgoetz @harshach @d2r @nathanmarz I will appreciate if you guys can review this.

revans2:
OK I want to start off by saying that I am getting some compilation errors.  First the code now requires the use of java 7 to compile.  I don't think it is that big of a deal, but the pom.xml still lists the java source and target as 1.6.

Also I am getting an error with ```net.sf:jargs:jar:1.0``` not being found when I try to run.

```mvn clean install -DskipTests```

Which appears to be a dependency of ttorrent.  Is there something special you had to do to make this compile?  Perhaps there is just something odd going on with my maven, as dependency:tree seems to be able to find it OK.

**Comment by revans2** [comment]
Oh and it looks like the bittorrent code is not even used right now, so I don't wee a reason to have it as a dependency.

**Comment by Parth-Brahmbhatt** [comment]
@revans2 I did not do anything special it works on my machine but I just checked and the default JVM on my machine is 1.7. I removed the bit torrent code as I mentioned in the first comment on this PR but I forgot to remove it from pom.xml. Pushing a change to remove it. Thanks for taking the time to review.

**Comment by Parth-Brahmbhatt** [comment]
@revans2  I also removed one unused import that was causing the 1.7 requirement, I tried locally with 1.6 and it worked fine. Let me know if you run into some other issues.

**Comment by revans2** [comment]
From reading through the design document, my initial impressions are that we are coupling the Nimbus fail over and leader election too closely to having a persistent store for the data.  

It feels to me like we want two different things.  One is a highly available store for blobs (we are working on an API for something similar to this for STORM-411) that we can write into it and query it to know that the blob has been persisted.  This could mean adequate replication, or whatever that blob store feels it needs.

The second thing we want is leader election/fail over for nimbus.

By separating the two, we can easily have something where we are running on YARN, and we only have one nimbus instance, but the data is stored in HDFS.  Nimbus crashes a new one comes up else where and everything should work just fine.  Or we are running on EC2 and we want nimbus to be hot/warm, but I don't want 3 instances of nimbus because it costs too much, just 2 so I store the data in S3 instead.  Exposing the replication count feels like it is an internal detail of storage system, that we really don't care that much about.

It really feels like it would give us a lot more flexibility to have the storage API completely separate from the fail over and leader election code.

File: docs/documentation/nimbus-ha-design.md [hunk 1]
```diff
@@ -0,0 +1,223 @@
...
+code locally before assuming the leadership role. If any active topology code is missing, the node will not accept the 
+leadership role instead it will release the lock and wait till it has all the code before requeueing for leader lock. 
+
+The first implementation will be Zookeeper based. If the zookeeper connection is lost/resetted resulting in loss of lock
+or the spot in queue the implementation will take care of updating the state such that isLeader() will reflect the 
+current status.The leader like actions must finish in less than minimumOf(connectionTimeout, SessionTimeout) to ensure
+the lock was held by nimbus for the entire duration of the action (Not sure if we want to just state this expectation 
+and ensure that zk configurations are set high enough which will result in higher failover time or we actually want to 
+create some sort of rollback mechanism for all actions, the second option needs a lot of code). If a nimbus that is not 
+leader receives a request that only a leader can perform it will throw a RunTimeException.
```
revans2: For fail-over, what we really want is some fencing around the state changes.  Since most, but not all, state is stored in zookeeper, it feels like writes to zookeeper are the proper place to put the fencing.  There are a few pieces of state that nimbus holds in memory for a short period of time, for example topology X is about to be killed.  In the first go around I think that is fine, but it would probably be good to look at persisting them in the future.

The fencing would only really need to be right before a write to ZK.  It would check that we are the leader right before we persist the data.  if we are not throw an exception and don't do the write.  That should be simple enough.

revans2:
I am also rather nervous about clients talking directly to zookeeper to know where nimbus is.  Zookeeper is really the scaling bottle neck in storm storm right now, and each time a client connects to it, a write occurs.  This is going to significantly increase the write load on the ZK cluster used with storm.

I think this is fine for a small cluster and I am OK with it going in without this, but if we want a decently large cluster we need an option where the client will try to connect to a hard coded list of hosts and ask them which one is the leader.

storm-core/src/clj/backtype/storm/cluster.clj
```diff
@@ -309,6 +324,16 @@
           (swap! assignment-version-callback assoc storm-id callback))
         (get-version cluster-state (assignment-path storm-id) (not-nil? callback)))
 
+      (code-distributor
+        [this callback]
+        (when callback
+          (reset! code-distributor-callback callback))
+        (get-children cluster-state CODE-DISTRIBUTOR-SUBTREE (not-nil? callback)))
+
+      (code-distributor-info
```
revans2: Why are we calling this code-distributor when it is leader election?

storm-core/src/clj/backtype/storm/daemon/supervisor.clj
```diff
@@ -31,6 +32,7 @@
 
 (defmulti download-storm-code cluster-mode)
 (defmulti launch-worker (fn [supervisor & _] (cluster-mode (:conf supervisor))))
+(defmulti mk-bt-tracker cluster-mode)
```
revans2: Can we rename this? it is not bit torrent so bt-tracker is really a confusing name.

revans2: Should this just be removed? we are still downloading from nimbus, so this seems like it is not needed here.

storm-core/test/clj/backtype/storm/security/auth/nimbus_auth_test.clj
```diff
@@ -48,6 +48,7 @@
         nimbus-server (ThriftServer. (:daemon-conf cluster-map)
                                      (Nimbus$Processor. (:nimbus cluster-map)) 
                                      ThriftConnectionType/NIMBUS)]
+    (Thread/sleep 2000)
```
revans2: Sleeps in tests are problematic.  Is there a better way that we can check if we are ready instead of just waiting?

storm-core/src/clj/backtype/storm/cluster.clj
Parth-Brahmbhatt: @revans2 its not leader-election stuff. This is actually the code distributor interface as discussed in the design doc which is separate from leader election. Essentially we have a zookeeper node /storm/code-distributor which will have one entry for each topology under it and every nimbus that has code metadata file will be listed under /storm-codedistributor/topologyid/ . 

I think the design doc discusses the metafile and code-distributor in more details but in a nut shell our code distributor protocol relies on a meta file that each host must download from some other host, most likely the leader nimbus but not necessary. The meta file has the details about the actual topology code/conf download location. In case of hdfs it is a directory location in hdfs , in case of local file system it is a local file system directory location. This is important because when we start supporting bit torrent the .torrent file will still need to be distributed. out side of bit torrent protocol.

storm-core/src/clj/backtype/storm/daemon/supervisor.clj
Parth-Brahmbhatt: yes, will have to rename all things bit torrent. I will push the change soon.

revans2:
I have done a quick read through the code, but nothing too deep.  I think a lot of the work seems fine, but I really would like to see a much better split between the leader election and the persistent storage.  Most of the code is already there, but it still feels tightly coupled.

I'll try to find some more time tomorrow to dig deeper into the changes.

storm-core/test/clj/backtype/storm/security/auth/nimbus_auth_test.clj
Parth-Brahmbhatt: I don't think this was intended change. Removed and ran the test 3 times and it never failed.

Parth-Brahmbhatt:
@jmVeiga Thanks for catching that and sorry about the bad push. I just tested a wordcount topology and it seems to be working now. I also found out that one of the test is broken and I will push the fix shortly but you should be able to build with mvn clean install -DskipTests=true for now for testing purpose. Appreciate the time you are taking for testing the feature.

harshach:
@Parth-Brahmbhatt I apologize for not getting to this earlier. I took a first-pass at the code and tested it in cluster. Here are my thoughts and nitpicks :)
1. nimbus.min.replication.count default should be 1 . Lets consider that leader nimbus is also a replica.
Also can we rename this as "topology.replication.count" . This config affects topology replicas and specifying "min" in the config indicates that minimum has to be reached for the success but in this even if the min.replication is not reached topology can be deployed.
2. nimbus.max.replication.wait.time.sec this should be higher than 0 as default. Can be renamed to "topology.max.replication.wait.time.secs"
3. needs doc on how to use nimbus.min.replication.count . Either user can specify as part of toplogy config? another way is to nimbus.min.replication.count via storm jar
example: ./bin/storm jar examples/storm-starter/storm-starter-topologies-0.10.0-SNAPSHOT.jar storm.starter.WordCountTopology wordcount3 -c "nimbus.min.replication.count=0"
I am not sure if this value being validated anywhere i.e user can pass in -1 and the topology still gets uploaded.
4. If a user kills a topology , topology jars are not being deleted from the non-leader replica.
5. **Improvement** we can do this in a follow-up JIRA. If a user already uploads a topology and wants to increase its replication factor. Not sure if this is required, thought it would be helpful.
6. **Improvement** follow-up JIRA .
TopologySummary should also contain what its replication. Helpful for admin/users to check if a topology has enough replicas.
7. renew-credentials should happen even if nimbus is leader or not . This make sures that a non-leader nimbus still have valid ticket. In a scenario where we don't renew ticket for non-leader nimbus and ticket expires. If the leader nimbus dies , this non-leader nimbus becomes leader which is running with a expired ticket.
8. The above should also apply to cleanup-corrupt-topologies . This might be what causing the issue mentioned in 4.
9. Mostly a question. From the code and doc it looks to me all the clients need to call zk-leader-elector to find the leader nimbus. This is ok for clojure or java code. There can be a usecase where users want to submit a topology without using StormSubmitter, is there a way to provide a thrift api or
some other mechanism to discover the leader nimbus by passing in zookeeper connect string.

Overall the code looks great and tested the failure scenarios during topology deployment and topology opertaions. It looks good to me.  Great work @Parth-Brahmbhatt

harshach:
If anyone interested testing nimbus ha in secure setup . Here is a vagrant setup to make things bit easier https://github.com/harshach/storm-vagrant/tree/nimbus-ha-security .  Instructions on how to set up are under here https://github.com/harshach/storm-vagrant/blob/nimbus-ha-security/README.md .

Parth-Brahmbhatt:
@harshach Thanks for taking the time to review. 
1.  nimbus.min.replication.count=1, makese sense.Currently in wait_for_desired_code_replication method I am passing "conf" and not "storm-conf" or "total-storm-conf" which has the user overrides so this configuration can not be modified per topology. I didn't think about the scenario where users have topologies with different replication needs but I think its a good idea to give this option so modified both the name of the config and the code.
2. nimbus.max.replication.wait.time.sec=0, again will change the name and code so this becomes a topology level config. Also for starters updating the default value to 60 seconds.
3. I have documented the configs and meaning as part of nimbus-ha-design in the configuration section. I will update the section to indicate the changes made in point 1 and 2.
4. That is a new bug and I will try to reproduce it locally. The local files from nimbuses are deleted via "clean-inbox" thread and that thread is scheduled for non leader nimbuses as well so not really sure why this would happen.
5. By default all nimbuses will sync-code so even if user indicates he only wants replication of n, it is  just the lower bound and if cluster has m nimbuses where m > n , all m nimbuses will replicate topology code. If by dynamically modifying the replication count you mean some API like "Deactivate my topology and reactivate only when this replication count is achieved" then yes we can discuss it in separate jira but my initial thought is this won't be very useful as ideally the admin should investigate why all nimbuses don't have the topology replicated.
6. Agreed, I can do this as part of this PR only.
7. Renew credentials thread only renews creds per topology and stores the creds for that topology in zookeeper under /storm/credentials/storm-id. I don't think it is responsible for renewing nimbus's own credentials. Given a non leader nimbus process will never use a topology credentials for any purpose and should have its own TGT for nimbus operation I am not sure if we want to enable this thread.  You already know I don't understand the storm security part well :-).  can you explain why nimbus would need topology creds?
8. cleanup-corrupt-topology thread lists local directory and diffs it with active storm-ids. It does not cleanup local state but instead deletes the zookeeper state for missing storm-ids. This means an out of sync nimbus host can delete a perfectly working topology. I think you mean to activate the clean-inbox thread which clears the local state and that thread is already activated for non leader nimbus.
9. As mentioned in the design doc I think a thrift API is a great idea for nimbus discovery and I plan to add them. I felt the changes I have as part of this single PR are already too big for review and it might be better to do it as part of a follow up task. Given both you and @revans2 have asked for this I am going to create a child task and submit a PR against this branch.

Thanks for the vagrant setup, I tried and failed to have a working multi nimbus secure setup. With your secure vagrant multi nimbus setup I can now test nimbus HA in secure environment.

harshach:
@Parth-Brahmbhatt 
1. I think I misunderstood the config here if your initial intent is to keep cluster wide and keep a storm.yaml config for cluster wide replication and have min replication I think what you have is works fine.   I thought this config can be overridden by the topology users and hence my comment. Please ignore as it doesn't look like your intent.
2. same as above
3. sounds good
5. Agree with you
7. This is my bad and you are right , the current code you have works fine.
8. Thanks for the details , ok now I understood it bit better :)
9. Sounds good.

Parth-Brahmbhatt:
@harshach My original intention was to have a single replication count for a cluster but after your comments I realized the users may want to change this setting per topology based on their availability requirements which may vary for each topology so I think its a great suggestion. I already made changes so the config is topology based and the replication count will be reflected in ui/rest/thrift. Unfortunately I don't think posting a PR against my own branch will show up in the jira but you can view the changes here https://github.com/Parth-Brahmbhatt/incubator-storm/pull/2/files

harshach:
Thanks @Parth-Brahmbhatt . Tested the above changes looks good to me.

Parth-Brahmbhatt:
upmerged to master and added the discovery using a thrift API.

Parth-Brahmbhatt:
@revans2 I have added the nimbus discovery using thrift API to address your concerns around clients connecting directly to zookeeper. 

For fencing around zk, non leader nimbuses still need to write some information (e.g. they write /storm/code-distributor/ what topologies they have locally replicated,and now they also write their nimbus summary under /storm/nimbuses) which means I can not put a single fencing around all zk write. One way or another I will have to differentiate between leader operations vs non leader operation. I prefer to put the fencing upfront so any in memory states are also not altered.

The coupling between storage layer and leader election is a side effect of having to support a protocol like bittorrent in absence of any highly available storage from which other hosts could download .torrent files. I can make the assumption that ICodeDistributor implementation are always backed by highly available storage systems and that will allow for removal of coupling with replication count but then unless STORM-411 is done we will have to force users to use HDFS to achieve nimbus HA.

Parth-Brahmbhatt:
@ptgoetz When you have time can you please review this?

ptgoetz:
@Parth-Brahmbhatt I'm in the process, but it might take a while. I want to run some tests with different failure scenarios, but so far I like what I see.

tdunning:
This is a bit late, but a ZK capability that might help with fencing is the multi operation.  What you can do is to a check on a leader election znode in the same multi that you are doing a state update.  If the leader election znode does not have the expected value, the state update will not be done.

You may also be able to use version numbers to achieve the desired fencing.  You could do a read of the state, verify that the correct master is listed in the state and then do a modify that requires a known version.  This will fail if somebody else has updated the state, but will succeed if the updating node is still the master.

These sorts of guaranteed updates are more secure than most other forms of fencing because they are guaranteed against race conditions, partitions and time faults.

ptgoetz:
I'm +1 for merging this, however this feature is targeted for a post-0.10.0 (AKA security release).

I'd like to propose that this be merged to a feature branch (e.g. "0.11.x-branch") to allow for further testing and development. That should enable others to more easily build and test with these changes. It would also relieve @Parth-Brahmbhatt from constantly having to upmerge this pull request.

Granted that would likely shift some of that burden to committers such that patches to master will also have to be committed/merged to that branch.

I'd be interested to hear others' thoughts.

harshach:
@ptgoetz @revans2 @Parth-Brahmbhatt  I am +1 on the patch as well. I like the idea of 0.11-branch makes it easier for anyone testing it too.

Parth-Brahmbhatt:
@revans2 @harshach @ptgoetz upmerged to master.

Parth-Brahmbhatt:
@revans2 @harshach @ptgoetz upmerged to master one more time and this time it took me a lot longer due to my own changes (rolling upgrade support and run as user support). I would like to mention that since last 3-4 commits I have only tested locally by running mvn clean install and running wordcount topology. Given the number of changes, I will try to re-test the HA test cases and rolling upgrade test cases one more time in my vagrant setup and update the PR.

vesense:
+1 looks good to me.

ptgoetz:
+1

I'll proceed with crating the necessary branches.

ptgoetz:
Thanks @Parth-Brahmbhatt. I've merged this to 0.11.x-branch.

ptgoetz:
**NOTE:**  0.11.x-branch has been renamed to 'nimbus-ha-branach'.

longdafeng:
@ptgoetz @revans2 @Parth-Brahmbhatt, 

Sorry for late discuss the HA desgin. 
Strongly recommend using JStorm's Nimbus HA Design. It is pretty stable.  JStorm's Nimbus HA has been released for one year and has been approved stable.

The logic is very simple. The code less 1000 lines.

1. Every nimbus will try to own the znode /nimbus_master, the winner will be the nimbus's master. Slaves will watch and timely check the znode, once it disappear, slaves will try to own it. During slaves check the znode, it will sync binary from master.
2. All client API firstly connection ZK and get to know who are the nimbus Master.

(1) The core code in nimbus:
https://github.com/alibaba/jstorm/blob/master/jstorm-server/src/main/java/com/alibaba/jstorm/schedule/FollowerRunnable.java

(2) How to find the master of nimbus:
https://github.com/alibaba/jstorm/blob/master/jstorm-client/src/main/java/backtype/storm/security/auth/ThriftClient.java

vesense:
@longdafeng GOOD JOB! it's simple and efficient.

Parth-Brahmbhatt:
Disclaimer I did not thoroughly look at the code but I am commenting based on your design description of Jstorm.

@longdafeng Did you have a chance to take a look at the current design? We are using curator for leader election which seems to be a very well tested library and is not really far from what you have proposed for leader election.

As for the length of the code, I don't completely agree with that being a good metric for most things. Due to the usage of an existing library the actual code for leader election in current PR is much smaller, 53 lines. https://github.com/Parth-Brahmbhatt/incubator-storm/blob/STORM-166/storm-core/src/clj/backtype/storm/zookeeper.clj#L250.

On top of that as part of this PR several of us had concerns around all clients connecting to zk to identify leader nimbus , as each new zk connection is a write to zk. We have partially fixed the issue by introducing thrift APIs for nimbus discovery which should be more efficient then the original approach and I plan to add caching at nimbus layer which should further improve the performance.

As @ptgoetz mentioned in the jira, we do not want  user's topologies getting lost once nimbus accepts it and we also do not want to force all users to have a dependency on a fully replicated storage layer like HDFS. In current design by adding a code replication interface we are guaranteeing that once a topology is in active state it will be fully replicated, which seems to be another missing feature in your proposal. Its still a choice between availability and initial topology submission time which the users can chose based on their topology.replication.count config setting.

We also added few more features like UI improvements, nimbus summary being stored in zk, thrift API modification so users can figure out replication factor of their topologies,  compatibility with rolling upgrade feature. All of which in my opinion are good admin tools and this feature will be incomplete without it.

I appreciate any feedback you can provide based on your experience of running Nimbus HA in production for a year. Please take some time to review the current design and let us know if you have any concerns.

longdafeng:
@Parth-Brahmbhatt , I don't suggest using "org.apache.curator.framework.recipes.leader"

In fact, We had implemented HA with "org.apache.curator.framework.recipes.leader" in the first version of JStorm Nimbus HA,  It occur a lot of problem when Zookeeper load is big. Later We use low level of curator api to implement HA. it's much more stable even when zookeeper load is big.


Right now, JStorm Nimbus HA is the third version.

revans2:
@longdafeng I am glad to see the JStorm community starting to help out here.  I respect your experience with this, especially since this is your third iteration on the code.  ZK load is a big issue and concern for me so I am all for trying to adopt what JStorm has done in the area of HA.  Because we are working on a separate feature branch for this, where we can break APIs if need be, perhaps we can check in the code as it is now, and file a follow on JIRA to adopt the model, configs, and ideally some of the code that JStorm is using.  This would also make combining the two simpler in the future.

@Parth-Brahmbhatt @ptgoetz do either of you have an opinion on this?

ptgoetz:
@revans2 @longdafeng @Parth-Brahmbhatt I haven't had a chance to fully review the JStorm Nimbus HA implementation yet, but I'm open to incorporating any of its concepts/implementation.

But to be clear, we have to complete IP clearance before any decisions are made regarding the JStorm code.

Parth-Brahmbhatt:
@revans2 @ptgoetz @harshach I have merged with master one more time. We are still using curator's LeaderLatch recipe. Nimbus discovery is done via an API so all clients don't have to connect to zookeeper.

harshach:
I am +1 on merging into master.

Parth-Brahmbhatt:
@revans2 Can you please review this PR when you have time? Given you have been involved in the original PR I don't want to commit this until I get a confirmation from you.

revans2:
Sorry I took so long on this.  I am +1 on merging this in. I see a +1 from @harshach before the upmerge and a +1 from @ptgoetz from a long time ago. The upmerge looks like it was mostly trivial changes so I am just going to merge this into master.

PR #357, opened 2014-12-22
STORM-600: Update version of jacoco plugin.

revans2:


d2r:
+1  I compiled and tested on OSX with jdk 1.8.  The build was successful and the tests passed.

PR #360, opened 2014-12-23
Storm-456: changing encoding to ring/url-encode.

Parth-Brahmbhatt:


revans2:
For the most part this looks good and I am +1 on merging this in.  However, I was wondering why we don't just replace the the backtype.storm.util url-encode with the ring version?  or delete it and replace all of the calls to it everywhere.  If it doesn't do the right thing we should stop using it everywhere.

Parth-Brahmbhatt:
@revans2 I am not sure if it does not do the right thing, but our decoding in ui is always done by ring decoder and it definitely does not match with the encoding done by the java encoder. I like the idea of using the ring encoder everywhere so I made the changes.

harshach:
@revans2 @Parth-Brahmbhatt  I've the same question on url-encode. I think we should replace the backtype.storm.util url-encode with ring . The only place url-encode used other than ui is cluster.clj. 
It's good to have one url-encode everywhere.

Parth-Brahmbhatt:
@harshach @revans2  made the suggested changes.

harshach:
+1

harshach:
@Parth-Brahmbhatt  Thanks for the patch , pushed into the master.

PR #361, opened 2014-12-23
STORM-603 Log errors when required kafka params are missing

curtisallen:
https://issues.apache.org/jira/browse/STORM-603

I was upgrading our topologies to storm-0.9.3 this [commit](https://github.com/apache/storm/commit/2596e335f27a57784a93a57823bd93dde587909f) introduced a change that threw me for a loop. When submitting my topology I got the following error.
```
[main] ERROR storm.kafka.DynamicBrokersReader - Couldn't connect to zookeeper
 java.lang.IllegalArgumentException: Don't know how to convert null to int
 	at backtype.storm.utils.Utils.getInt(Utils.java:301) ~[storm-core-0.9.3.jar:0.9.3]
 	at storm.kafka.DynamicBrokersReader.<init>(DynamicBrokersReader.java:47) ~[gambit-storm-threads-0.0.1-SNAPSHOT-jar-with-dependencies.jar:na]
 	at com.pearson.gambit.threads.storm.ThreadsTopology.main(ThreadsTopology.java:45) [gambit-storm-threads-0.0.1-SNAPSHOT-jar-with-dependencies.jar:na]
 Exception in thread "main" java.lang.RuntimeException: java.lang.RuntimeException: java.lang.NullPointerException
 	at storm.kafka.DynamicBrokersReader.getBrokerInfo(DynamicBrokersReader.java:81)
 	at com.pearson.gambit.threads.storm.ThreadsTopology.main(ThreadsTopology.java:48)
 Caused by: java.lang.RuntimeException: java.lang.NullPointerException
 	at storm.kafka.DynamicBrokersReader.getNumPartitions(DynamicBrokersReader.java:94)
 	at storm.kafka.DynamicBrokersReader.getBrokerInfo(DynamicBrokersReader.java:65)
 	... 1 more
 	at storm.kafka.DynamicBrokersReader.getNumPartitions(DynamicBrokersReader.java:91)
 	... 2 more 
```

It took me a while to figure out that this error stems from missing the `backtype.storm.Config.STORM_ZOOKEEPER_CONNECTION_TIMEOUT` property in the `conf` map provided to the `storm.kafka.DynamicBrokersReader` constructer. It would be nice to check the required configuration parameters and throw a RuntimeException if any are missing. 
This pull requests adds logging when required errors are missing.

### DIFF
--- a/external/storm-kafka/src/jvm/storm/kafka/DynamicBrokersReader.java
+++ b/external/storm-kafka/src/jvm/storm/kafka/DynamicBrokersReader.java
@@ -41,6 +41,14 @@ public class DynamicBrokersReader {
     private String _topic;
 
     public DynamicBrokersReader(Map conf, String zkStr, String zkPath, String topic) {
+        // Check required parameters
+        if(conf == null) {LOG.error("conf cannot be null");}
+        validateConfig(conf);
+
+        if(zkStr == null) {LOG.error("zkString cannot be null");}
+        if(zkPath == null) {LOG.error("zkPath cannot be null");}
+        if(topic == null) {LOG.error("topic cannot be null");}
+
         _zkPath = zkPath;
         _topic = topic;
         try {
@@ -53,6 +61,7 @@ public class DynamicBrokersReader {
             _curator.start();
         } catch (Exception ex) {
             LOG.error("Couldn't connect to zookeeper", ex);
+            throw new RuntimeException(ex);
         }
     }
 
@@ -146,6 +155,21 @@ public class DynamicBrokersReader {
             return new Broker(host, port);
         } catch (UnsupportedEncodingException e) {
             throw new RuntimeException(e);
+        }
+    }
+
+    private void validateConfig(final Map conf) {
+        if(conf.get(Config.STORM_ZOOKEEPER_SESSION_TIMEOUT) == null) {
+            LOG.error("{} cannot be null", Config.STORM_ZOOKEEPER_SESSION_TIMEOUT);
+        }
+        if(conf.get(Config.STORM_ZOOKEEPER_CONNECTION_TIMEOUT) == null) {
+            LOG.error("{} cannot be null", Config.STORM_ZOOKEEPER_CONNECTION_TIMEOUT);
+        }
+        if(conf.get(Config.STORM_ZOOKEEPER_RETRY_TIMES) == null) {
+            LOG.error("{} cannot be null", Config.STORM_ZOOKEEPER_RETRY_TIMES);
+        }
+        if(conf.get(Config.STORM_ZOOKEEPER_RETRY_INTERVAL) == null) {
+            LOG.error("{} cannot be null", Config.STORM_ZOOKEEPER_RETRY_INTERVAL);
         }
     }
 
--- a/external/storm-kafka/src/jvm/storm/kafka/DynamicBrokersReader.java
+++ b/external/storm-kafka/src/jvm/storm/kafka/DynamicBrokersReader.java
@@ -19,6 +19,7 @@ package storm.kafka;
 
 import backtype.storm.Config;
 import backtype.storm.utils.Utils;
+import com.google.common.base.Preconditions;
 import org.apache.curator.framework.CuratorFramework;
 import org.apache.curator.framework.CuratorFrameworkFactory;
 import org.apache.curator.retry.RetryNTimes;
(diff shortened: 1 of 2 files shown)


external/storm-kafka/src/jvm/storm/kafka/DynamicBrokersReader.java
```diff
@@ -149,4 +158,19 @@ private Broker getBrokerHost(byte[] contents) {
         }
     }
 
+    private void validateConfig(final Map conf) {
+        if(conf.get(Config.STORM_ZOOKEEPER_SESSION_TIMEOUT) == null) {
```
harshach: These are all storm configs , even if user didn't pass these or set in storm.yaml there are defaults . So not sure checking for this necessary.

external/storm-kafka/src/jvm/storm/kafka/DynamicBrokersReader.java
```diff
@@ -41,6 +41,14 @@
     private String _topic;
 
     public DynamicBrokersReader(Map conf, String zkStr, String zkPath, String topic) {
+        // Check required parameters
+        if(conf == null) {LOG.error("conf cannot be null");}
```
harshach: Instead putting it in DynamicBrokersReader it will be better if we have this in KafkaConfig and SpoutConfig and have a validate method there.

external/storm-kafka/src/jvm/storm/kafka/DynamicBrokersReader.java
curtisallen: @harshach there are no defaults for these values, hence the reason for this pull request. The constructor for the DynamicBrokerReader accepts a `java.util.Map` Which must contain the values that are verified in this `validateConfig` method.

external/storm-kafka/src/jvm/storm/kafka/DynamicBrokersReader.java
curtisallen: @harshach Agreed. I didn't take that approach because it would require significantly more changes to the spout since `DynamicBrokerReader` doesn't reference `KafkaConfig` or `SpoutConfig` currently. Also it would require a modification to `DynamicBrokerReader`'s constructor to accept a `KafkaConfig` instance breaking legacy code. I thought that would be a big change for a minor storm release.

revans2:
@curtisallen Sorry it has taken so long to review this.  I think the changes look OK.  I am curious why we only logging an error instead of throwing an exception.  Wouldn't we just end up throwing an NPE later on in the code?  Why not fail fast?

@harshach did @curtisallen answer all of your questions?  Do you have any other follow up?

curtisallen:
@revans2 Thanks for looking at this. Before we didn't throw any exceptions when the config was incorrect. I added this [line](https://github.com/curtisallen/storm/blob/STORM-603/external/storm-kafka/src/jvm/storm/kafka/DynamicBrokersReader.java#L64) that throws an exception if the config is incorrect. Right now it's a RuntimeException however, would you prefer I throw a null pointer exception earlier?

revans2:
@curtisallen I am not really concerned about it being an NPE, or a RuntimeException, my concern is two fold.  

First If we mis configure something like Config we log [something](https://github.com/curtisallen/storm/blob/STORM-603/external/storm-kafka/src/jvm/storm/kafka/DynamicBrokersReader.java#L45) it and then later end up throwing a RuntimeException.  In this case the RuntimeException is wrapping an NPE that will only have the line number of where the NPE happened.  If all we had were the logs to go off of then it would not be a big deal, but in storm the exception and the stack trace often are returned and displayed on the UI.  I would much rather see  "Config cannot be null" on the UI instead of NPE on line 162 of DynamicBrokersReader.

My second concern is around failing fast.  In the case of _zkPath or _topic being null.  There is nothing that explicitly uses them in the constructor.  The failure will happen much later if ever as is the case for _zkPath and _topic, as they appear to only be used as a part of string concatenation with a +.  Which when done with a null writes "null" to the string instead of throwing an NPE, so we could end up trying to read from "/kafka/topics/null/partitions".

If something is wrong in a way that we cannot recover from I would much rather throw an exception the moment we know it is wrong rather than log an error and hope that the exception will be thrown later on, or that somehow we will manage to not actually need this config to be set.

curtisallen:
@revans2 agreed, the concerns you brought make a lot of sense to me. I should have some time later this week to fail fast ;)

revans2:
@curtisallen Thanks.  If you could post a comment here when it is updated I'll take a look and we should hopefully be able to get it in soon.

curtisallen:
@revans2 now we throw NPEs if the values are missing. Leveraged guava's `Preconditions` . Let me know if you need anything else. Thanks!

revans2:
+1 looks great.  Thanks for doing this

PR #363, opened 2014-12-24
STORM-596: remove config topology.receiver.buffer.size

caofangkun:
topology.receiver.buffer.size has no effect and not necessary for now.
Should we just remove it ?


caofangkun:
@harshach 
Could you please have a review on this PR? Thanks a lot.

revans2:
This is technically a non-backwards compatible change, but I think it is OK.  I am +1.

storm-core/src/clj/backtype/storm/daemon/worker.clj
```diff
@@ -355,8 +355,7 @@
     (:storm-id worker)
     (:receiver-thread-count worker)
     (:port worker)
-    (:transfer-local-fn worker)
-    (-> worker :storm-conf (get TOPOLOGY-RECEIVER-BUFFER-SIZE))
+    (:transfer-local-fn worker))
```
revans2: There is an extra ')' here

revans2:
No one else has decided to review this, and it still compiles and I am still +1 so I will merge it in.

PR #364, opened 2014-12-26
[STORM-607] storm-hbase HBaseMapState should support user to customize the hbase-key & hbase-qualifier

dashengju:
In HBaseMapState, user can specific hbase-qualifier by Options, and the hbase-key is composed by all the keys by multiPut's  keys.
for example, If I have stream with (deal_id, date, pv ), grouped by (deal_id, date), then, the hbase-key is composed by(deal_id, date), the hbase-qualifier is pv.
But when I want hbase-key is deal_id, and hbase-qualifier is pv+date, HBaseMapState can not support this.

==================================================================================

With this patch, user can customize the hbase-key & hbase-qualifier by interface TridentHBaseMapMapper. 

As a old user, can use the old style like this:
HBaseMapState.Options hbaseOptions = new HBaseMapState.Options();
hbaseOptions.mapMapper = new SimpleTridentHBaseMapMapper("pv");

Also, you can customize the hbase-key & hbase-qualifier like this:

    public class WithDateHBaseMapMapper implements TridentHBaseMapMapper {
        private String qualifierPrefix;

        public WithDateHBaseMapMapper(String qualifierPrefix) {
            this.qualifierPrefix = qualifierPrefix;
        }

        @Override
        public byte[] rowKey(List<Object> keys) {
            String dealId = String.valueOf(keys.get(0));
            return dealId.getBytes();
        }

        @Override
        public String qualifier(List<Object> keys) {
            String dateStr = String.valueOf(keys.get(1));
            return qualifierPrefix + "_" + dateStr;
        }
    }

HBaseMapState.Options hbaseOptions = new HBaseMapState.Options();
hbaseOptions.mapMapper = new WithDateHBaseMapMapper("pv");

### DIFF
--- a/external/storm-hbase/src/main/java/org/apache/storm/hbase/trident/mapper/SimpleTridentHBaseMapMapper.java
+++ b/external/storm-hbase/src/main/java/org/apache/storm/hbase/trident/mapper/SimpleTridentHBaseMapMapper.java
@@ -1,0 +1,50 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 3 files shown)


dashengju:
@ptgoetz , can you help to review this PR?

knusbaum:
This looks ok to me, but it's not my area, so I'll wait for someone else to look it over as well before I +1.

revans2:
@dashengju Sorry the review took so long.  The changes look good to me +1

PR #366, opened 2014-12-29
[STORM-610] Check the return value of fts_close().

mnogu:


revans2:
+1 another really good catch.

PR #369, opened 2014-12-29
STORM-613: Fix wrong getOffset return value

tedxia:
PR for [STORM-613](https://issues.apache.org/jira/browse/STORM-613)

When judge getOffset return value, use 0 not NO_OFFSET as invalid offset return value.


revans2:
+1 looks fine to me.

PR #373, opened 2015-01-06
STORM-248 STORM-322 are fixed and upmerged

fmazoyer:
Details are in JIRA https://issues.apache.org/jira/browse/STORM-248.
As a collateral benefit, this also addresses JIRA https://issues.apache.org/jira/browse/STORM-322.
Could you please review?


bin/storm
```diff
@@ -74,7 +74,8 @@ if (not os.path.isfile(os.path.join(USER_CONF_DIR, "storm.yaml"))):
 STORM_RELEASE_DIR = os.path.join(STORM_DIR, "RELEASE")
 STORM_LIB_DIR = os.path.join(STORM_DIR, "lib")
 STORM_BIN_DIR = os.path.join(STORM_DIR, "bin")
-STORM_LOGBACK_FILE_PATH = os.path.join(STORM_DIR, "logback", "cluster.xml")
+STORM_LOGBACK_DIR = os.path.join(STORM_DIR, "logback")
+STORM_LOGBACK_FILE_PATH = os.path.join(STORM_LOGBACK_DIR, "cluster.xml")
```
revans2: I don't think this is used any more.

revans2:
One minor comment, but other then that it looks good.

fmazoyer:
Thank you for your comment, this is done :-)
Could you please review ?

revans2:
Looks good to me +1.  I'll give it a few days to let others have a chance to review, and then check it in if there are no other comments.

bin/storm
```diff
@@ -312,6 +312,13 @@ def repl():
     cppaths = [CLUSTER_CONF_DIR]
     exec_storm_class("clojure.main", jvmtype="-client", extrajars=cppaths)
 
+def get_logback_conf_dir():
+    cppaths = [CLUSTER_CONF_DIR]
+    storm_logback_conf_dir = confvalue("storm.logback.conf.dir", cppaths)
+    if(storm_logback_conf_dir == None or storm_logback_conf_dir == "nil"):
```
harshach: Here storm_logback_conf_dir == "nil" will match if the storm_logback_conf_dir is set to  "nil" . I think you meant to check if the string is empty, a better way probably is "if not storm_logback_conf"

revans2: No it is actually "nil".  You can try it.

```
$ storm localconfvalue something.bogus
something.bogus: nil
```

revans2: If we want it to behave differently we need to fix confvalue.

harshach: Sorry, you are right it is "nil".

harshach:
+1

fmazoyer:
Hi Robert,
do you think there are still some things to do on my side ?
Thanks for everything :-)

harshach:
@fmazoyer  Thanks for the patch. I pushed it into master.

d2r:
@harshach, would you resolve STORM-248?  I Resolved STORM-322 already.

PR #374, opened 2015-01-06
Storm-616 : Storm-jdbc connector.

Parth-Brahmbhatt:


external/storm-jdbc/README.md
```diff
@@ -0,0 +1,208 @@
+#Storm JDBC
+Storm/Trident integration for JDBC. This package includes the core bolts and trident states that allows a storm topology
+to either insert storm tuples in a database table or to execute select queries against a database and enrich tuples 
+in a storm topology. This code uses HikariCP for connection pooling. See http://brettwooldridge.github.io/HikariCP.
+
+## Inserting into a database.
+The bolt and trindet state included in this package for inserting data into a database tables are tied to a single table.
```
harshach: typo

external/storm-jdbc/src/main/java/org/apache/storm/jdbc/bolt/JdbcBolt.java
```diff
@@ -0,0 +1,91 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.storm.jdbc.bolt;
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
```
harshach: license included twice

external/storm-jdbc/README.md
```diff
@@ -0,0 +1,208 @@
...
+select dept_name from department, user_department where department.dept_id = user_department.dept_id and user_department.user_id = ?;
+```
+### Execution
+Run the `org.apache.storm.jdbc.topology.UserPersistanceTopology` class using storm jar command. The class expects 5 args
+storm jar org.apache.storm.jdbc.topology.UserPersistanceTopology <dataSourceClassName> <dataSource.url> <user> <password> [topology name]
+
+Mysql Example:
+```
+storm jar ~/repo/incubator-storm/external/storm-jdbc/target/storm-jdbc-0.10.0-SNAPSHOT-jar-with-dependencies.jar 
+org.apache.storm.jdbc.topology.UserPersistanceTridentTopology  com.mysql.jdbc.jdbc2.optional.MysqlDataSource
```
harshach: this should be UserPersistentTopology and "password" argument should be removed.

external/storm-jdbc/src/test/java/org/apache/storm/jdbc/topology/AbstractUserTopology.java
```diff
@@ -0,0 +1,102 @@
...
+    protected JdbcLookupMapper jdbcLookupMapper;
+
+    protected static final String TABLE_NAME = "user";
+    protected static final String JDBC_CONF = "jdbc.conf";
+    protected static final String SELECT_QUERY = "select dept_name from department, user_department where department.dept_id = user_department.dept_id" +
+            " and user_department.user_id = ?";
+
+    public void execute(String[] args) throws Exception {
+        if (args.length != 4 && args.length != 5) {
+            System.out.println("Usage: " + this.getClass().getSimpleName() + " <dataSourceClassName> <dataSource.url> "
```
harshach: This is condition will never work, it supposed to be "or"?

external/storm-jdbc/README.md
Parth-Brahmbhatt: fixed.

external/storm-jdbc/src/main/java/org/apache/storm/jdbc/bolt/JdbcBolt.java
Parth-Brahmbhatt: fixed.

external/storm-jdbc/README.md
Parth-Brahmbhatt: password argument is still needed why do you think it needs to be removed?

external/storm-jdbc/src/test/java/org/apache/storm/jdbc/topology/AbstractUserTopology.java
```diff
@@ -0,0 +1,102 @@
...
+
+        if (args.length == 4) {
+            LocalCluster cluster = new LocalCluster();
+            cluster.submitTopology("test", config, getTopology());
+            Thread.sleep(30000);
+            cluster.killTopology("test");
+            cluster.shutdown();
+            System.exit(0);
+        } else {
+            StormSubmitter.submitTopology(args[5], config, getTopology());
```
harshach: this should be args[4]?

external/storm-jdbc/src/main/java/org/apache/storm/jdbc/bolt/JdbcBolt.java
```diff
@@ -0,0 +1,91 @@
...
+ * <p/>
+ * Note: Each JdbcBolt defined in a topology is tied to a specific table.
+ */
+public class JdbcBolt extends AbstractJdbcBolt {
+    private static final Logger LOG = LoggerFactory.getLogger(JdbcBolt.class);
+
+    private String tableName;
+    private JdbcMapper jdbcMapper;
+
+    public JdbcBolt(String configKey) {
```
harshach: I am ok with current implementation but to be consistent with other connectors (hdfs, hive) can we make this as withConfigKey instead of a constructor param?

external/storm-jdbc/src/test/java/org/apache/storm/jdbc/topology/AbstractUserTopology.java
Parth-Brahmbhatt: yes, fixed. Sorry about this, originally I required that the user passed the table name as argument but I figured it is better to just execute all the DDLs as part of test so I removed it.

harshach:
overall I am +1 on merging this. There are minor nit-picks in the above comments. I would like to see the unit tests for the bolt and state. Probably we can use in-process jdbc db to test it out. We can do this as part of another JIRA.
@ptgoetz @revans2  please take a look at the PR. If we need a sponsor for jdbc I can volunteer. Thanks.

external/storm-jdbc/src/main/java/org/apache/storm/jdbc/bolt/AbstractJdbcBolt.java
```diff
@@ -0,0 +1,53 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.storm.jdbc.bolt;
+
+import backtype.storm.task.OutputCollector;
+import backtype.storm.task.TopologyContext;
+import backtype.storm.topology.base.BaseRichBolt;
+import org.apache.commons.lang.Validate;
+import org.apache.storm.jdbc.common.JDBCClient;
```
itaifrenkel: Notice that JDBCClient and JdbcMapper has different casing notations.

external/storm-jdbc/src/main/java/org/apache/storm/jdbc/common/JDBCClient.java
```diff
@@ -0,0 +1,209 @@
...
+public class JDBCClient {
+    private static final Logger LOG = LoggerFactory.getLogger(JDBCClient.class);
+
+    private HikariDataSource dataSource;
+
+    public JDBCClient(Map<String, Object> map) {
+        Properties properties = new Properties();
+        properties.putAll(map);
+        HikariConfig config = new HikariConfig(properties);
+        this.dataSource = new HikariDataSource(config);
```
itaifrenkel: Each bolt has its own JDBCClient. Each JDBCClient has its own data source. Does it mean that each bolt has its own dedicated connection pool ?

external/storm-jdbc/src/main/java/org/apache/storm/jdbc/bolt/JdbcBolt.java
```diff
@@ -0,0 +1,91 @@
...
+    @Override
+    public void execute(Tuple tuple) {
+        try {
+            List<Column> columns = jdbcMapper.getColumns(tuple);
+            List<List<Column>> columnLists = new ArrayList<List<Column>>();
+            columnLists.add(columns);
+            this.jdbcClient.insert(this.tableName, columnLists);
+        } catch (Exception e) {
+            LOG.warn("Failing tuple.", e);
+            this.collector.fail(tuple);
```
itaifrenkel: this.collector.reportError(e) ?

external/storm-jdbc/src/main/java/org/apache/storm/jdbc/bolt/JdbcLookupBolt.java
```diff
@@ -0,0 +1,80 @@
...
+                    List<Values> values = jdbcLookupMapper.toTuple(tuple, row);
+                    for (Values value : values) {
+                        collector.emit(value);
+                    }
+                }
+            }
+            this.collector.ack(tuple);
+        } catch (Exception e) {
+            LOG.info("Failed to execute a select query {} on tuple {} ", this.selectQuery, tuple);
+            this.collector.fail(tuple);
```
itaifrenkel: this.collector.reportError(e) ?

external/storm-jdbc/src/main/java/org/apache/storm/jdbc/bolt/JdbcBolt.java
```diff
@@ -0,0 +1,91 @@
...
+
+    @Override
+    public void execute(Tuple tuple) {
+        try {
+            List<Column> columns = jdbcMapper.getColumns(tuple);
+            List<List<Column>> columnLists = new ArrayList<List<Column>>();
+            columnLists.add(columns);
+            this.jdbcClient.insert(this.tableName, columnLists);
+        } catch (Exception e) {
+            LOG.warn("Failing tuple.", e);
```
itaifrenkel: We usually do not log each exception. we let the reportError decide what to do with it.  So to use this module we would have to turn off this logging completely.

external/storm-jdbc/src/main/java/org/apache/storm/jdbc/bolt/JdbcLookupBolt.java
```diff
@@ -0,0 +1,80 @@
...
+                for (List<Column> row : result) {
+                    List<Values> values = jdbcLookupMapper.toTuple(tuple, row);
+                    for (Values value : values) {
+                        collector.emit(value);
+                    }
+                }
+            }
+            this.collector.ack(tuple);
+        } catch (Exception e) {
+            LOG.info("Failed to execute a select query {} on tuple {} ", this.selectQuery, tuple);
```
itaifrenkel: We usually do not log each exception. we let the reportError decide what to do with it.  So to use this module we would have to turn off this logging completely.

external/storm-jdbc/src/main/java/org/apache/storm/jdbc/bolt/JdbcLookupBolt.java
itaifrenkel: Another approach is to derive from IBasicBolt and let the default error handling kick in.

external/storm-jdbc/src/main/java/org/apache/storm/jdbc/bolt/AbstractJdbcBolt.java
Parth-Brahmbhatt: fixed.

external/storm-jdbc/src/main/java/org/apache/storm/jdbc/common/JDBCClient.java
Parth-Brahmbhatt: yes.

itaifrenkel: So for each bolt, how many connections do you advise to be in the pool? Given each bolt has one thread then I would say 1. But then it would loose time reconnecting each tuple that may have caused some kind of error. So then I would be inclined to say 2 connections, but then for 10 bolts, I would need 20 connections. What is the recommended value you use in production?

external/storm-jdbc/README.md
harshach: my bad I meant to say "user". The example given above for submitting the UserPersistanceTridentTopology contains more args than required
"storm jar ~/repo/incubator-storm/external/storm-jdbc/target/storm-jdbc-0.10.0-SNAPSHOT-jar-with-dependencies.jar 
org.apache.storm.jdbc.topology.UserPersistanceTridentTopology  com.mysql.jdbc.jdbc2.optional.MysqlDataSource 
jdbc:mysql://localhost/test root password user UserPersistenceTopology"
Here "root password user" I think "user" needs to be removed.

harshach:
@Parth-Brahmbhatt  this is also need to be included under storm-dist/binary/src/main/assembly/binary.xml  like here https://github.com/apache/storm/blob/master/storm-dist/binary/src/main/assembly/binary.xml#L93

external/storm-jdbc/src/main/java/org/apache/storm/jdbc/bolt/JdbcBolt.java
Parth-Brahmbhatt: I am not seeing this being used in any other connector instead all the other connectors, hdfs/hbase/kafka just log the error. I am fine with adding it as it just seems to report the error so the ui can display it but disabling the log may result in loss of information which can make debugging harder. I propose to report the error but also log it.

external/storm-jdbc/src/main/java/org/apache/storm/jdbc/bolt/JdbcLookupBolt.java
Parth-Brahmbhatt: same as above.

external/storm-jdbc/src/main/java/org/apache/storm/jdbc/common/JDBCClient.java
```diff
@@ -0,0 +1,209 @@
...
+            if(LOG.isDebugEnabled()) {
+                LOG.debug("Executing query " + query);
+            }
+
+            PreparedStatement preparedStatement = connection.prepareStatement(query);
+            for(List<Column> columnList : columnLists) {
+                setPreparedStatementParams(preparedStatement, columnList);
+            }
+
+            return preparedStatement.executeUpdate();
```
itaifrenkel: executeUpdate could block for a very long time since preparedState.setQueryTimeout() was not called. This should be preferably configurable.

external/storm-jdbc/src/main/java/org/apache/storm/jdbc/bolt/JdbcBolt.java
itaifrenkel: I am not qualified for advising best practices in Storm codebase. Our experience has shown that in our environment a database that is not responding for any reason would flood the logs with long stack traces causing far worse side effects. It could happen to others. To reuse this code I would probably need to copy-paste the entire bolt, which is ok, since most of the logic resides elsewhere.

external/storm-jdbc/src/main/java/org/apache/storm/jdbc/common/JDBCClient.java
Parth-Brahmbhatt: added a new config queryTimeOutSecs which is defaulted to 30 seconds which is equal to topology.message.timeout.secs. user can override during bolt construction.

Parth-Brahmbhatt:
@harshach  modified storm-dist/pom.xml and also removed the config key from the constructor and added withConfigKey  method to be consistent with other connectors.

external/storm-jdbc/src/main/java/org/apache/storm/jdbc/bolt/JdbcBolt.java
Parth-Brahmbhatt: fixed.

external/storm-jdbc/README.md
Parth-Brahmbhatt: removed.

external/storm-jdbc/src/main/java/org/apache/storm/jdbc/common/JDBCClient.java
```diff
@@ -0,0 +1,213 @@
...
+import com.zaxxer.hikari.HikariDataSource;
+import org.apache.commons.lang.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.sql.*;
+import java.sql.Date;
+import java.util.*;
+
+public class JdbcClient {
```
itaifrenkel: The name of the file has different casing than the name of the class.

Parth-Brahmbhatt: @itaifrenkel fixed.

harshach:
@revans2 @ptgoetz can you take a look at this PR. Thanks.

external/storm-jdbc/src/main/java/org/apache/storm/jdbc/bolt/JdbcBolt.java
ptgoetz: +1 for reportError() before fail. Perhaps we should document this as a best practice and revisit other modules to implement it there as well.

external/storm-jdbc/src/main/java/org/apache/storm/jdbc/bolt/JdbcBolt.java
ptgoetz: I agree. `reportError()` has built-in protections against being flooded (and overloading ZK as a result).

external/storm-jdbc/src/main/java/org/apache/storm/jdbc/bolt/JdbcBolt.java
```diff
@@ -0,0 +1,81 @@
...
+ * <p/>
+ * Note: Each JdbcBolt defined in a topology is tied to a specific table.
+ */
+public class JdbcBolt extends AbstractJdbcBolt {
+    private static final Logger LOG = LoggerFactory.getLogger(JdbcBolt.class);
+
+    private String tableName;
+    private JdbcMapper jdbcMapper;
+
+    public JdbcBolt withConfigKey(String configKey) {
```
ptgoetz: I think the idea behind config key here, like in other modules, is to allow multiple instances of the bolt use different configurations (e.g. pointing to different databases, schemas, etc.).

That being said, I'm not really partial to using either a builder method or constructor arg, as long as it's exposed as an option. Either way, I'd like to try to keep it consistent across modules so the APIs have a similar "feel". One option for the builder method would be to define a default config key value that gets used unless overridden with `withConfigKey()`. We should also clearly document this behavior.

external/storm-jdbc/src/main/java/org/apache/storm/jdbc/bolt/JdbcBolt.java
```diff
@@ -0,0 +1,81 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Basic bolt for writing to any Database table.
+ * <p/>
+ * Note: Each JdbcBolt defined in a topology is tied to a specific table.
+ */
+public class JdbcBolt extends AbstractJdbcBolt {
```
ptgoetz: I agree with @revans2. While this deviates from some of the other older modules, I think this is a good convention to follow moving forward.

external/storm-jdbc/src/main/java/org/apache/storm/jdbc/common/JdbcClient.java
```diff
@@ -0,0 +1,213 @@
...
+                @Override
+                public String apply(Column input) {
+                    return input.getColumnName();
+                }
+            });
+            String columns = Joiner.on(",").join(columnNames);
+            sb.append(columns).append(") values ( ");
+
+            String placeHolders = StringUtils.chop(StringUtils.repeat("?,", columnNames.size()));
+            sb.append(placeHolders).append(")");
```
ptgoetz: To build on @revans2's comment: We're building preparedStatements each time, but using them only once and discarding them. I'd like to see them cached for re-use (probably with some sort of eviction policy to protect against blowing out memory).

I don't see this as a blocker, but something that should be considered here or as an early improvement.

external/storm-jdbc/src/main/java/org/apache/storm/jdbc/bolt/JdbcBolt.java
Parth-Brahmbhatt: Moved back to constructor. I am not sure what you mean when you say we could just pass the config into the Bolt. Hikari has many configuration https://github.com/brettwooldridge/HikariCP. The configKey is so user can provide any subset of those configs. so the configKey is just name of the config key in your submitted storm-conf, the value of this should be a map of hikari config keys to their corresponding values.

I can extract a connectionProvider interface and create a Hikari implementation that users have to construct to initialize the Bolts. This will mean the bolts will not have a config Key, users will be able to use whatever connection pool they feel most comfortable with and the configuration details will live in the connection provider. Do you think that will be better than what we currently have.

external/storm-jdbc/src/main/java/org/apache/storm/jdbc/bolt/JdbcBolt.java
Parth-Brahmbhatt: Done.

external/storm-jdbc/src/main/java/org/apache/storm/jdbc/bolt/JdbcBolt.java
Parth-Brahmbhatt: Done.

external/storm-jdbc/pom.xml
```diff
@@ -0,0 +1,125 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-lang3</artifactId>
+            <version>3.3</version>
+        </dependency>
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+            <version>17.0</version>
```
ptgoetz: Is there a way we can use the guava dependency already included with Storm (I would understand if not, just double-checking).

external/storm-jdbc/src/main/java/org/apache/storm/jdbc/bolt/JdbcLookupBolt.java
```diff
@@ -0,0 +1,86 @@
...
+    public void execute(Tuple tuple) {
+        try {
+            List<Column> columns = jdbcLookupMapper.getColumns(tuple);
+            List<List<Column>> result = jdbcClient.select(this.selectQuery, columns);
+
+            if (result != null && result.size() != 0) {
+                for (List<Column> row : result) {
+                    List<Values> values = jdbcLookupMapper.toTuple(tuple, row);
+                    for (Values value : values) {
+                        collector.emit(value);
```
Parth-Brahmbhatt: Done.

external/storm-jdbc/src/main/java/org/apache/storm/jdbc/common/Column.java
```diff
@@ -0,0 +1,87 @@
...
+ */
+package org.apache.storm.jdbc.common;
+
+
+import java.io.Serializable;
+import java.lang.reflect.Field;
+import java.sql.Date;
+import java.sql.Time;
+import java.sql.Timestamp;
+import java.sql.Types;
```
Parth-Brahmbhatt: I removed some methods and forgot to remove unused imports. Done.

external/storm-jdbc/src/main/java/org/apache/storm/jdbc/common/Column.java
```diff
@@ -0,0 +1,87 @@
...
+
+public class Column<T> implements Serializable {
+
+    private String columnName;
+    private T val;
+    private int sqlType;
+
+    public Column(String columnName, T val, int sqlType) {
```
Parth-Brahmbhatt: I have added a javadoc that should help to clarify things. If we consider a database table as a list of rows and each row as a list of columns, this class represent a single instance of a column, so for a row [userID =1 , userName = "Foo"] we will have two instances of Column class
new Column("userId", 1, Types.INTEGER) and new Column("userName", "Foo", Types.Varchar). 

I am not sure why java's jdbc API decided to go with integer to represent sql Types instead of using more readable Enums but the sqlType has value mapped to java.sql.Types where each constant represents a database type.

ptgoetz:
Overall I think this is a good start and I am supportive. @revans2 brought up some good points and I tried to expand/clarify in several places. I feel more efficient use of prepared statements is an important. The ability to invoke stored procedures is something we might want to consider, but not a requirement, especially since calling stored procedures can be a VERY DB-dependent thing.

I'm also willing to act as a sponsor.

external/storm-jdbc/src/main/java/org/apache/storm/jdbc/common/JdbcClient.java
```diff
@@ -0,0 +1,213 @@
...
+
+            String query = sb.toString();
+            if(LOG.isDebugEnabled()) {
+                LOG.debug("Executing query " + query);
```
Parth-Brahmbhatt: Done.

external/storm-jdbc/src/main/java/org/apache/storm/jdbc/common/JdbcClient.java
Parth-Brahmbhatt: We are using prepared statements with place holders so we are leveraging the statement cache., not sure if you mean something else by "JDBC stored queries". The statement cache is generally on server side and as long as we are using preparedstatements with place holders "?" we should get the benefit of not having to compile/parse the sql and the cached execution plan.  

The only reason we construct the sql query each time is because this is more flexible then a single static sql. For inserts people can have columns with default values and their topology can chose to emit values for those columns sometimes and sometimes they can just ignore those columns and both scenarios will be supported.

I can provide an override sql that user can provide , personally I would rather wait till we actually come across a DB or user that runs into some issue before adding that.

external/storm-jdbc/src/main/java/org/apache/storm/jdbc/bolt/JdbcBolt.java
revans2: OK So I see two ways to do the Hikari configuration.
The current way.
```
Map<String, String> jdbcConf = new HashMap<String, String>();
jdbcConf.put("a","b");
stormConf.put("something", jdbcConf);
bolt.withConfigKey("something");
```

or we could just pass the jdbc config to the Bolt, and not send it through the topology config first.

```
Map<String, String> jdbcConf = new HashMap<String, String>();
jdbcConf.put("a","b");
bolt.withConfig(jdbcConf);
``` 
I don't understand why we need the extra step of storing a config inside another config, and providing a layer of indirection.

As for constructor values vs builder values, it is a question of required vs optional parameters.

```
Map<String, String> jdbcConf = new HashMap<String, String>();
jdbcConf.put("a","b");
stormConf.put("something", jdbcConf);
//oops forgot this bolt.withConfigKey("something");
```
results in an NPE after the topology is launched.  If it were a constructor parameter.
```
Map<String, String> jdbcConf = new HashMap<String, String>();
jdbcConf.put("a","b");
stormConf.put("something", jdbcConf);
new Bolt(/*forgot "something"*/);
```
is a compile time error.  I am fine with having a builder pattern for anything that is optional, or even a way to override/extend require values.  I see it as defensive programming.

external/storm-jdbc/src/main/java/org/apache/storm/jdbc/common/Column.java
revans2: The javadocs is what I really was interested in.  The markdown doc was really great, but an int for sqlType is really confusing without some documentation right there.

external/storm-jdbc/pom.xml
Parth-Brahmbhatt: I tried doing that but without a direct dependency the compile phase fails as storm-core is provided which is non transitive. I could change guava to provided scope, will that work?

external/storm-jdbc/src/main/java/org/apache/storm/jdbc/bolt/JdbcInsertBolt.java
```diff
@@ -0,0 +1,79 @@
...
+        return this;
+    }
+
+    @Override
+    public void execute(Tuple tuple) {
+        try {
+            List<Column> columns = jdbcMapper.getColumns(tuple);
+            List<List<Column>> columnLists = new ArrayList<List<Column>>();
+            columnLists.add(columns);
+            this.jdbcClient.insert(this.tableName, columnLists);
```
revans2: For me the issue is with required configuration vs optional configuration.  configKey is not the only required configuration.  table name and jdbcMapper both appear to be required in the current code.  I would much rather have them be parameters to the constructor instead of builder methods, or have a reasonable default set in the constructor.

external/storm-jdbc/src/main/java/org/apache/storm/jdbc/bolt/JdbcLookupBolt.java
```diff
@@ -0,0 +1,84 @@
...
+    public JdbcLookupBolt withQueryTimeoutSecs(int queryTimeoutSecs) {
+        this.queryTimeoutSecs = queryTimeoutSecs;
+        return this;
+    }
+
+    @Override
+    public void execute(Tuple tuple) {
+        try {
+            List<Column> columns = jdbcLookupMapper.getColumns(tuple);
+            List<List<Column>> result = jdbcClient.select(this.selectQuery, columns);
```
revans2: Similar comment to JdbcInsertBolt.  jdbcLookupMapper and selectQuery both look to be required, as such I would really rather have them be constructor parameters.

external/storm-jdbc/src/main/java/org/apache/storm/jdbc/common/JdbcClient.java
revans2: I am not a DB expert I read up on JDBC some more and talked to @kishorvpatil I think the current code is going to be fine.  I am a little concerned about the overhead of doing string manipulations to recreate each prepared statement, but I am OK with waiting for it to become a problem before we implement some sort of caching.

revans2:
So I am OK with the code as it is now.  I would prefer to see required configuration values become constructor arguments, but I am not going to block this going in without it.

external/storm-jdbc/src/main/java/org/apache/storm/jdbc/bolt/JdbcLookupBolt.java
Parth-Brahmbhatt: This is done, added all required param as part of constructors.

revans2:
I am +1 for merging this in now.  But I am not a DB expert so I would feel more comfortable with others giving it a +1 too before going forward.

ptgoetz:
+1 looks good to me. I don't think we need the LICENSE file here, but that's a trivial change that can be done at merge time.

harshach:
+1 Tested the code with mysql.

Crystark:
This would indeed be nice but after trying it out it seems that if a batch has more than one tuple to insert, then only the last tuple is inserted.

In [JdbcClient#insert](https://github.com/Parth-Brahmbhatt/incubator-storm/blob/STORM-616/external/storm-jdbc/src/main/java/org/apache/storm/jdbc/common/JdbcClient.java#L48) we can see that there's only one set of `VALUES` (line 61-64) but then it loops on all the columns and each one overrides the previous values that were set as SQL parameters leading to only the last one beeing used.

external/storm-jdbc/src/main/java/org/apache/storm/jdbc/common/JdbcClient.java
```diff
@@ -0,0 +1,213 @@
...
+            String query = sb.toString();
+
+            LOG.debug("Executing query {}", query);
+
+
+            PreparedStatement preparedStatement = connection.prepareStatement(query);
+            preparedStatement.setQueryTimeout(queryTimeoutSecs);
+            for(List<Column> columnList : columnLists) {
+                setPreparedStatementParams(preparedStatement, columnList);
+            }
```
Crystark: Here each `columnList ` of `columnLists` overrides the previous value.

external/storm-jdbc/src/main/java/org/apache/storm/jdbc/common/JdbcClient.java
Crystark: The problem at this part is that there's only one set of values though this method works with multiple set of values.

Parth-Brahmbhatt:
@Crystark  super sorry about this I am not sure how the heck did I miss this. The code you pointed at is fine but I should be using executeBatch instead of executeUpdate. I have modified the code to use executeBatch and also modified the test to test the bug you pointed out.

Crystark:
Works like a charm :)

harshach:
+1.  @Parth-Brahmbhatt  Thanks for the quick fix . @Crystark  Thanks for review and nice catch.
@revans2  @ptgoetz  I think this is ready to merge it in will give it a day or two before pushing it to master.

harshach:
Thanks @Parth-Brahmbhatt  merged into master.

revans2:
I am now getting a compilation error.
``` [ERROR] external/storm-jdbc/src/main/java/org/apache/storm/jdbc/common/Util.java:[30,29] incomparable types: int and java.lang.Object```  Not sure if this is JDK8 specific or not.

revans2:
Yes it looks like it is a JDK8 issue.  JDK6 compiles the code just fine. I'll file a new JIRA for it.

Crystark:
On more thing i noticed: [Column.hashCode](https://github.com/apache/storm/blob/master/external/storm-jdbc/src/main/java/org/apache/storm/jdbc/common/Column.java#L99) doesn't support the value beeing null. I've tried using those columns as keys of a HashMap and got a NPE.

Parth-Brahmbhatt:
@Crystark  You are right I will fix it.

PR #377, opened 2015-01-09
[STORM-620] Duplicate maven plugin declaration

Lewuathe:
Fix duplicate maven-javadoc-plugin declaration in pom.xml reporting section

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -723,10 +723,6 @@
                 </configuration>
             </plugin>
             <plugin>
-            <groupId>org.apache.maven.plugins</groupId>
-            <artifactId>maven-javadoc-plugin</artifactId>
-            </plugin>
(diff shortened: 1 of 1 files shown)


revans2:
The javadoc plugin you removed is in the reporting section of the pom.  That is used when running `mvn site`  The other one is part of the dist profile, The dist profile is more or less dead code right now, but it's intention is to be run when we are building the binary and source distribution tar balls.   I think both are needed if we ever get the dist profile working.

-1

If you disagree and can let me know why this is valid, I am happy to change my opinion.  Otherwise please close this pull request.

revans2:
@Lewuathe very sorry I missed where it is in two places in the reporting section.  You are 100% correct.  +1 for mering this in.

PR #381, opened 2015-01-14
[Storm 599] Use use nimbus's cached heartbeats rather than fetching again from ZK, take two

d2r:
This is a retry of the reverted pull request #356.

In #356, a merge error (we think?) on the branch used for the pull request caused part of the change on one line to go missing.

```Diff
- beats (.executor-beats storm-cluster-state storm-id (:executor->node+port assignment))
+ beats (map-val :heartbeat @(:heartbeats-cache nimbus))
```

The new code should have been:
```Diff
+ beats (map-val :heartbeat (get @(:heartbeats-cache nimbus) storm-id))
```

This caused incorrect retrieval of heartbeats and therefore metrics were not populated as noted by @harshach [here](https://issues.apache.org/jira/browse/STORM-599?focusedCommentId=14275512&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-14275512).

Testing again with this branch in a clean environment shows that metrics are updated as expected once per `nimbus.monitor.freq.secs`, which defaults to 10s.


revans2:
+1 I manually brought up the topology and it is behaving as expected.  The metrics are updated every 10 seconds like clockwork.

harshach:
+1. There might be minor issue that metrics gets updated in bulk i.e if  I refresh it doesn't reflect the new values immediately it will update after few retries. Before if I want to get latest metrics I can keep refreshing and gets the latest metrics. This is not a big issue I am +1 on merging this in. Thanks.

PR #385, opened 2015-01-19
STORM-469

rekhajoshm:


Parth-Brahmbhatt:
-1. I don't think the change is necessary. Storm ships with a logviewer daemon. If users want more details on some exception in ui, they have the ability to do so by ensuring that they have the logviewer daemons started which should allow them to navigate and view the entire log file through ui.

rekhajoshm:
Parth-Brahmbhatt: As you like it, Parth :-)
IMO, if a truncated errormsg is shown , a link on it to the logs is a good ui design., and more user friendly.
Agree with Jason Kania on it.It might be even cooler to show last few bytes of log link, or complete log link…as in hadoop jobtracker design.

rekhajoshm:
jmlogan: you are right about the extra files.few more changes.will update commit.thanks!

harshach:
@rekhajoshm  Thanks for the patch. Few observations here 
1) you are changing interfaces to pass a codeDir param. I can't find it where codeDir used to fixed this issue. Whenever you are changing apis like this you should introduce a new method definition not change the existing ones as these are storm-core apis.  Please also explain whats the necessity of codeDir.
2) you added dependency-reduced-pom.xml and any reason to remove org.clojars.runa from pom.xml. Also maven-eclipse.xml shouldn't be part of this commit.
I am -1 on the current approach, changing apis to solve this is not a good idea. This pretty much breaks current topologies.
I suggest you to look into ui/core.clj , one approach is to have a search functionality in logviewer  and provide a link next to error message with param to logviewer as the errormessage. When user clicks on that it goes to logviewer and highlight the error in the log file.

rekhajoshm:
Thanks Harshch for having a look.Somehow you seem to have reviewed both storm-469 and storm-528 commits together?
1. commit 2931487 is for storm-469, tested and works on my storm cluster and gives me link to truncated text the log link, which works.
2. commit ef19e71 is for storm-528 and as informed to jmlogan, i will update the files in a while.Thanks!

harshach:
@rekhajoshm  Ok. Open up a different pull request for each of those JIRAs. Once the PR approved by the committers we will merge pr directly into master branch. combining two JIRAs into one PR is not recommended.

rekhajoshm:
created separate pulls.

PR #388, opened 2015-01-20
STORM-469

rekhajoshm:
@harshach @Parth-Brahmbhatt For https://issues.apache.org/jira/browse/STORM-469
IMO, if a truncated errormsg is shown , a link on it to the logs is a good ui design., and more user friendly.
Agree with Jason Kania on it.So just that.It might be even cooler to show last few bytes of log link, or complete log link…as in hadoop jobtracker design.

### DIFF
--- a/storm-core/src/ui/public/templates/component-page-template.html
+++ b/storm-core/src/ui/public/templates/component-page-template.html
@@ -479,7 +479,7 @@
         <td>{{errorHost}}</td>
         <td><a href="{{errorWorkerLogLink}}">{{errorPort}}</a></td>
         <td>
-          <span id="{{errorLapsedSecs}}" class="errorSpan">{{error}}</span>
+          <span id="{{errorLapsedSecs}}" class="errorSpan"><a href="{{errorWorkerLogLink}}">{{error}}</a></span>
         </td>
       </tr>
       {{/componentErrors}}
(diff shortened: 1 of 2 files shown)


d2r:
I agree it is definitely a great idea to put the link on the UI, and this is what #150 did when it added `errorWorkerLogLink` to storm.

But I am not sure it adds anything to create a second link nearby to the same place.  We can already click on the port number (similar to the executor status links).  It was also pointed out to me that some users expect to be able to access URLs that are in the error snippet text itself, but making the whole snippet a link makes that use case difficult.

Is it OK just to use the link on the port number as it is currently?

rekhajoshm:
@d2r Thanks, and that is another thing - why do we have log link on port number?
That does not make any sense at all.The truncated text is where the user expects a error log link on.

revans2:
+1 I think the change is a good start to this problem.

d2r:
> @d2r Thanks, and that is another thing - why do we have log link on port number?
> That does not make any sense at all.The truncated text is where the user expects a error log link on.

@rekhajoshm , sorry, somehow I did not get notified of this post.

The worker log link on the port number corresponds to the way the worker log link is designed into the Executors table on the Component page.  The two are thus consistent.  But that does not mean it makes more sense than having the link to the worker log somewhere else. You have raised a good point.

It turns out that this change breaks the red-coloring of errors reported less than 5 minutes in age (since it makes the entire text a blue link), so we need to revert this change for now.

I am -1, since it breaks functionality.

PR #389, opened 2015-01-20
Storm 528

rekhajoshm:
@harshach - For https://issues.apache.org/jira/browse/STORM-528
The overriding of codeDir is intended to remove duplicates python, ruby, js external resources in codebase. ATM, I have not intentionally updated all clients to use this signature to override codeDir as that would change API drastically, and need consensus.


revans2:
There is a lot of code in here that I don't understand how it relates to STORM-528.  Setting the code dir, is not really something that I think we want to expose, except a controlled way for testing.  I also don't understand all of the code for max pending.  It looks like you have code from multiple JIRA here.

rekhajoshm:
closing to correct branch issue.thanks

PR #390, opened 2015-01-21
[STORM-630] Support for Clojure 1.6.0

Lewuathe:
Through upgrading clojure 1.6.0, there are name space conflicts in `zip-contains-dir?`. Renamed it.
It looks there is no other dependencies of `some?` function and a topology, `ExclamationTopology` seems fine on my test cluster. If there is any other concern about upgrading clojure, please let me know.


revans2:
+1 the code looks fine to me.

PR #391, opened 2015-01-22
STORM-635. logviewer returns 404 if storm_home/logs is a symlinked dir.

harshach:


revans2:
The change to check the parent directory is the root dir was put in on purpose for security reasons.  if someone puts in a file called "../../../etc/passwd" there could be some serious security issues involved.

-1

I am fine with supporting symlinks but we need to have a way to restrict what can be accessed through the logviewer.

harshach:
@revans2 my bad I missed security fix that went in before. Please check the latest patch . Thanks.

Parth-Brahmbhatt:
+1.

revans2:
+1 looks good to me.

PR #393, opened 2015-01-22
STORM-633. Nimbus - HTTP Error 413 full HEAD if using kerberos authentication.

harshach:


ptgoetz:
+1 (since this is just a doc update, we probably don't need to wait for additional review)

PR #296, opened 2014-10-16
STORM-532:Supervisor should restart worker immediately, if the worker pr...

caofangkun:
1: kill -9 7469  (worker process id)
2: check $STORM_HOME/logs/supervisor.log 
2015-01-27T10:45:40.439+0800 b.s.d.supervisor [INFO] Worker Process 10c49685-7f61-42d7-8146-4ac48f807bc0 exited with code: 137
2015-01-27T10:45:42.272+0800 b.s.util [INFO] Error when trying to check 7469.
2015-01-27T10:45:42.284+0800 b.s.d.supervisor [INFO] Shutting down and clearing state for id 10c49685-7f61-42d7-8146-4ac48f807bc0. Current supervisor time: 1422326742. State: :process-not-exists, Heartbeat: #backtype.storm.daemon.common.WorkerHeartbeat{:time-secs 1422326739, :storm-id "ExclamationTopology-1-1422266117", :executors #{[4 4] [7 7] [10 10] [13 13] [16 16] [-1 -1] [1 1]}, :port 6703, :process-id "7469"}
2015-01-27T10:45:42.284+0800 b.s.d.supervisor [INFO] Shutting down b298ca62-feba-4c1f-a41e-55c3e5b7ae99:10c49685-7f61-42d7-8146-4ac48f807bc0
2015-01-27T10:45:42.285+0800 b.s.config [INFO] GET worker-user 10c49685-7f61-42d7-8146-4ac48f807bc0
2015-01-27T10:45:42.312+0800 b.s.util [INFO] Error when trying to kill 7469. Process is probably already dead.
2015-01-27T10:45:43.319+0800 b.s.util [INFO] Error when trying to kill 7469. Process is probably already dead.
2015-01-27T10:45:43.441+0800 b.s.config [INFO] REMOVE worker-user 10c49685-7f61-42d7-8146-4ac48f807bc0
2015-01-27T10:45:43.442+0800 b.s.d.supervisor [INFO] Shut down b298ca62-feba-4c1f-a41e-55c3e5b7ae99:10c49685-7f61-42d7-8146-4ac48f807bc0
[...]

PR #396, opened 2015-01-27
STORM-638:UI should show up process-id of the Worker to which an Executor is assigned

caofangkun:
with process id


caofangkun:
Pay Attention: conflicts  with https://github.com/apache/storm/pull/296

storm-core/src/storm.thrift
```diff
@@ -205,7 +205,8 @@ struct ExecutorSummary {
   2: required string component_id;
   3: required string host;
   4: required i32 port;
-  5: required i32 uptime_secs;
+  5: required i32 process_id;
```
revans2: You cannot change a thrift ID like this.  It breaks backwards compatibility too much.  Please make sure that you add new tags instead of changing existing ones.  It is also usually preferable to make the new tag optional, so you can have forward and backwards compatibility.

storm-core/src/ui/public/templates/component-page-template.html
```diff
@@ -211,6 +216,7 @@
         <td>{{uptime}}</td>
         <td>{{host}}</td>
         <td><a href="{{workerLogLink}}">{{port}}</a></td>
+        <td>{{processId}}
```
revans2: Can we close the td tag?  I know it is not required but all the other ones are closed.

revans2:
I'm not convinced having the PID on the UI is really that helpful, but it does provide more information for possible automation or monitoring, so I am OK with it.

harshach:
@revans2 not quite comfortable with exposing pid on UI. UI is supposed to be available for all the users. In non-secure env it can be run without any filter so I am against exposing pid on the UI.

storm-core/src/storm.thrift
caofangkun: Fixed. Thank you. Please review this again.

revans2:
@harshach, If you are running the UI without a filter I would not consider that secure.  There is other more damaging information exposed on the UI already.  Also what is having the pid of the process going to let me do as an attacher?  If I cannot log onto the box the pid is really rather useless.  If I can log onto the box I can already get the pid by running ```ps -aef | grep '<topology-id>-<port-number>'```.  Granted I am not a malicious individual so perhaps I am missing something or perhaps an attach will show up in the future that is somehow related to the pid.  It is totally possible, but I don't see it as being that likely.

@caofangkun, The generated thrift code no longer compiles.  It looks like you regenerated with thrift7, and we have moved on to thrift 0.9.2.  Please regenerate the code again.

revans2:
Sorry it is not the code generation after all.  Although the  comment in the generated code with a 0.7.0 is off putting, and I would prefer it to be generated with 0.9.2.  by making the field optional it changes the constructor, so now backtype/storm/daemon/nimbus.clj:1325:47 is failing to compile, because the constructor is different.

vesense:
unnecessary

knusbaum:
-0

I don't really think this is a useful thing to add to the UI. If we want to add it to the JSON API, I'm okay with that, but let's not display it on the UI.

HeartSaVioR:
Agreed on @knusbaum.
It can reduce one ps command for actual users to operate (ps, kill, etc.) specific worker by hand, but I'm curious that there're other benefits.
I also think that since we're already defining "port", we would be better to encapsulate worker process into "port" and provide a way to operate such things via "port".

Or we can think about providing workers' information (usage of CPU, MEM, etc. but except secure things) into UI. Maybe we could provide a way to kill specific worker from UI, but I wonder there's such use cases.

knusbaum:
Closing this.

PR #401, opened 2015-01-29
STORM-130: Supervisor getting killed due to java.io.FileNotFoundException: File '../stormconf.ser' does not exist.

harshach:


harshach:
@ptgoetz  @d2r @revans2 @Parth-Brahmbhatt  Please take a look at the new PR. I apologize for the whitespace issue. There are still few left in this patch. As Taylor suggested please ?w=1 to the URL.

storm-core/src/clj/backtype/storm/daemon/supervisor.clj
```diff
@@ -349,6 +350,20 @@
                         (keys keepers))
            (zipmap (vals new-worker-ids) (keys new-worker-ids))
            ))
+
+    ;; check storm topology code dir exists before launching workers
+    (doseq [[port assignment] reassign-executors]
+      (let [storm-cluster-state (:storm-cluster-state supervisor)
+            downloaded-storm-ids (set (read-downloaded-storm-ids conf))
+            storm-id (:storm-id assignment)
+            assignment-info (.assignment-info-with-version storm-cluster-state storm-id nil)
```
revans2: This is going to cause a large load on zookeeper.  There is a reason we have the cache for the assignment info, and we only download it when it has changed.  We really should use the cache here if we can.

harshach: @revans2 are you talking about .assignment-info-with-version?

revans2:
For the most part it looks good, I would like to see us use the cached assignment instead of going to zookeeper again to get it.  I would also like to understand better how we avoid issues with downloading the same thing in two different threads.  I'm not sure I understand the exact cause of what puts us in this situation to begin with, so I don't feel comfortable that we will not try to download the same thing twice and cause some issues when we do.

harshach:
@revans2 thanks for the review. I am looking for cached assignment code will send updated patch. 
To address downloading the same code twice. It can happen in download-storm-code method I am checking if the topology code dir exists in the supervisor it won't overwrite and deletes the tmp dir.  More details on this comment on how this can happen https://issues.apache.org/jira/browse/STORM-130?focusedCommentId=14296241&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-14296241

harshach:
@revans2 I am not sure if we are caching assignments on supervisor. Can you point me to that part of the code. Also this part of the code gets called if there are any new assignments and I am not sure even in a busy cluster unless workers are getting killed often they shouldn't be too many new assignments

storm-core/src/clj/backtype/storm/daemon/supervisor.clj
revans2: Yes.  We could replace it with ```@(:assignment-versions supervisor)``` and presumably get the same result without hitting zookeeper.

revans2:
Ok So this is for an existing topology that moved and then moved back.  I understand now about the race.  I am +1 for the change, once we avoid hitting ZK quite so frequently.

harshach:
@revans2 thanks for pointer on using cached assignment. I also spoke to @Parth-Brahmbhatt  about download-storm-code, added the lock to make sure there won't be multiple threads downloading the same code . Please take a look.

revans2:
+1 looks good to me.

ptgoetz:
+1. I'd also like to see this back-ported to the 0.9.3 branch, but that shouldn't block this from getting merged to master.

harshach:
@ptgoetz  will send a new PR against 0.9.3 branch. Thanks.

harshach:
@revans2 @ptgoetz  opened a PR against 0.9.3-branch https://github.com/apache/storm/pull/418 . Also updated the current PR to handle another case.
This issue can happen if supervisors are down and user kills a topology , restart supervisors. These supervisors can crash with same issue as sync-process can start after cleaning up local deleted topology files . In general having try catch around launch-worker is good idea as there is no need to kill the supervisor itself if the launch-worker fails.

PR #402, opened 2015-01-29
STORM-162: Load Balancing Shuffle Grouping

revans2:
This pull request adds a notion of load that is measured and sent to upstream workers through the messaging system.  Shuffle grouping was then updated so that if it is configured to do so it will take the load into account.

### DIFF
--- a/storm-core/src/jvm/backtype/storm/Config.java
+++ b/storm-core/src/jvm/backtype/storm/Config.java
@@ -199,6 +199,12 @@ public class Config extends HashMap<String, Object> {
     public static final Object TOPOLOGY_TUPLE_SERIALIZER_SCHEMA = String.class;
 
     /**
+     * Disable load aware grouping support.
+     */
+    public static final String TOPOLOGY_DISABLE_LOADAWARE = "topology.disable.loadaware";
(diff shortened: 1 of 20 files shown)


storm-core/src/jvm/backtype/storm/Config.java
```diff
@@ -199,6 +199,12 @@
     public static final Object TOPOLOGY_TUPLE_SERIALIZER_SCHEMA = String.class;
 
     /**
+     * Disable load aware grouping support.
+     */
+    public static final String TOPOLOGY_DISABLE_LOADAWARE = "topology.disable.loadaware";
```
d2r: Could we call this something like `TOPOLOGY_DISABLE_LOADAWARE_MESSAGING` so it is clear what ceases to be aware of load?

revans2: Great idea, I couldn't really come up with a decent description.

d2r:
@revans2 can you upmerge?

revans2:
I have some more things to do.  There was a suggestion on the JIRA about trying to incorporate process latency along with the queue length for better results, and I would also like to look at adding in support for network distance to the routing too.  So I am closing this for now, and will either put up a new pull request or reopen this one when I have the code ready.

Luckyfang:
test

PR #403, opened 2015-01-30
[STORM-639] storm-maven-plugin not found

Lewuathe:
maven compile is fail due to missing storm-maven-plugin. 
This plugin should be installed before compiling storm-core project.

### DIFF
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -228,6 +228,8 @@ To pull in a merge request you should generally follow the command line instruct
 The following commands must be run from the top-level directory.
 
     # Build the code and run the tests (requires nodejs, python and ruby installed) 
+    # `mvn clean package` will fail because storm-core requires storm-maven-plugin.
+    # This plugin should be installed before compiling storm-core.
     $ mvn clean install
 
     # Build the code and run the tests, with specifying default test timeout (in millisecond)
--- a/storm-buildtools/storm-maven-plugins/pom.xml
+++ b/storm-buildtools/storm-maven-plugins/pom.xml
@@ -76,6 +76,16 @@
           </execution>
         </executions>
       </plugin>
+        <plugin>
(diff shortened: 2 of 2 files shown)


d2r:
I am seeing this happen too.  Even with this change, if I try to compile with `mvn compile` from the top level, I see the following error:
```
Error resolving version for plugin 'org.apache.storm:storm-maven-plugins'
```
After `mvn install`, it does compile OK.

Lewuathe:
@d2r Thank you for reporting. I'll fix to succeed compile with `mvn compile`.

Lewuathe:
It seems that there is no way to install `storm-maven-plugin` in local repository before compiling `storm-core` because with `compile` phase there is no jar files enough to install. So I think there are two options. 
1. Update documents to install `storm-maven-plugin` before compiling `storm-core` project.
2. Deploy `storm-maven-plugin` on remote repository.

What do you think? If someone has any other ideas, please let me know.

d2r:
> Update documents to install storm-maven-plugin before compiling storm-core project.

Yeah... I was afraid of that.  It is a binary dependency, and maven does not support making this kind of thing work.

So I think your suggestion of changing the documents is better.

Lewuathe:
I changed documentation. Can someone review it?

d2r:
> Update documents to install storm-maven-plugin before compiling storm-core project.

It looks as though the documents already say how to build it so that it works.  It does not claim that doing the build directly from storm-core without the plugin jar on the classpath will succeed.  So now I am not sure we want to add to the document one way not to build storm.

Really, it would be good if the plugin were pushed to maven repositories separately from storm, like @revans2 [suggested](https://issues.apache.org/jira/browse/STORM-639?focusedCommentId=14324305&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-14324305).

revans2:
I forgot to add +1

PR #408, opened 2015-02-02
STORM-648. storm examples BasicDRPCTopology fails without running.

harshach:


ptgoetz:
+1 I think this will just prevent stacktraces due to the immediate shutdown, but I'm fine with the change.

PR #409, opened 2015-02-03
STORM-652. Use latest junit 4.11 .

harshach:


ptgoetz:
+1 The upgrade does not seem to cause any issues with tests.

PR #413, opened 2015-02-03
[STORM-653] missing DRPC HTTP port in SECURITY.md

Lewuathe:
There is no description about DRPC HTTP port in SECURITY.md.
This port is used by DRPC Client as default 3774


kishorvpatil:
+1

PR #427, opened 2015-02-10
STORM-651: Rename "ui" service to "storm ui" and temp.txt should be in T...

Parth-Brahmbhatt:
...EMP folder.Add jdk and jres bin and libs dirs to lib path.


harshach:
+1

ptgoetz:
+1

PR #428, opened 2015-02-11
STORM-329: fix cascading Storm failure by improving reconnection strategy and buffering messages

miguno:
This is an improved version of the original pull request discussed at https://github.com/apache/storm/pull/268.  Please refer to the discussion in the link above.

**Note**:  Please give attribution to @tedxia when merging the pull request as he did a lot (most?) of the work in this pull request.

The changes of this pull request include:

- Most importantly, we fix a bug in Storm that may cause a cascading failure in a Storm cluster, to the point where the whole cluster becomes unusable.  This is achieved by the work described in the next bullet points.
- We refactor and improve the Netty messaging backend, notably the client.
- During the initial startup of a topology, Storm will now wait until worker (Netty) connections are ready for operation.  See the [original discussion thread](https://github.com/apache/storm/pull/268) for the detailed explanation and justification of this change.

@clockfly, @tedxia: Please add any further comments to STORM-329 to this pull request, if possible.

PR #268, opened 2014-09-24
STORM-329 : buffer message in client and reconnect remote server async

tedxia:
Description can be found at https://issues.apache.org/jira/browse/STORM-329
The main change are:
1. set a client buffer limit, if the buffer message exceed the limit, it will wait util there are enough space for the new arrival message.
2. if we lose the connection with remote worker, the reconnect logic will be asynchronously and will not block other worker.
3. there are timer for cache messages, if the messages are cache for more than "topology.message.timeout.secs", we will drop the cache message.

PR #429, opened 2015-02-11
STORM-329: fix cascading Storm failure by improving reconnection strategy and buffering messages

miguno:
**This PR contains the same code as https://github.com/apache/storm/pull/428 but as a single commit for a cleaner commit history of our Storm repo.**

--

This is an improved version of the original pull request discussed at https://github.com/apache/storm/pull/268.  Please refer to the discussion in the link above.

**Note**:  Please give attribution to @tedxia when merging the pull request as he did a lot (most?) of the work in this pull request.

The changes of this pull request include:

- Most importantly, we fix a bug in Storm that may cause a cascading failure in a Storm cluster, to the point where the whole cluster becomes unusable.  This is achieved by the work described in the next bullet points.
- We refactor and improve the Netty messaging backend, notably the client.
- During the initial startup of a topology, Storm will now wait until worker (Netty) connections are ready for operation.  See the [original discussion thread](https://github.com/apache/storm/pull/268) for the detailed explanation and justification of this change.

@clockfly, @tedxia: Please add any further comments to STORM-329 to this pull request, if possible.


danielschonfeld:
I admit to not understanding all the intricacies of the code as i'm still coming to terms with the different parts of storm.  However, does this PR handle the case of the worker that remained alive trying to refresh a connection to a dead worker or a dead worker that is in the process of starting up?

I seem to see the reverse being handled, but not in the direction mentioned above.  Of course that could also be cause i don't understand the code fully.

miguno:
This patch allows a worker to properly detect that the connection to a peer becomes unavailable -- for whatever reason (the remote worker is dead or restarting, there was a network glitch, etc). Also, any reconnection attempts are now async so that reconnecting will not block other activities of the worker (such as sending messages to other workers it is still connected to).

So to your question: this patch includes the case of the worker that remained alive trying to (re)connect to a dead peer.

Does that help?

nathanmarz:
I retract my earlier -1. It was mentioned that this enables backpressure for unacked topologies. Is this the case? If so, this is a great new feature of Storm and there should be tests added testing this behavior. Namely, it should test that:

- Spouts stop emitting when all buffers fill up
- Topology recovers on worker death and only a subset of messages are dropped in that case

miguno:
Thanks for your feedback, Nathan.

As far as I understand this patch does not enable backpressure. But: because there is no backpressure (yet) that we can rely on, this patch will improve at least the situation during the startup phase of a topology to prevent that a) an unacked topo will not lose messages during the startup, and b) we do not need to unnecessarily replay messages in case of acked topos during their startup.  This is achieved by checking that all worker connections are ready before the topology starts processing data.

So backpressure is still an open feature. Backpressure was IIRC mentioned in the initial PR because there was a deficiency (dating back to a ZMQ related TODO) that caused problems related to this PR/Storm tickets (327, 404, and one more).  However, this patch does make the best of the current situation even in the absence of backpressure.  But first and foremost this patch fixes a (critical) cascading failure that can bring Storm clusters to a halt.

Please correct me if I'm mistaken in my summary.

clockfly:
+1

danielschonfeld:
Doesn't dropping the messages coming from a non ack/fail caring spout negate the 'at least once' attempt of storm? I mean doesn't that kinda force you to make all your spouts ack/fail aware where before you could have gotten away without it?

In other words.  There is a chance that if the worker that died is the one containing the spout and if the first bolt is located on another worker, that technically at-least once wasn't tried but rather fell to the floor right away.

miguno:
If you need at-least-once processing you must use an acking topology, which will allow Storm to replay lost messages.  If instead you go with an unacking topology (= no guaranteed message processing) then you may run into data loss.  There re pros and cons for each variant, and e.g. in our case we use both depending on the use case.

Also: The semantics described above have been in Storm right from the beginning.  None of these have been changed by this pull request.

storm-core/src/jvm/backtype/storm/messaging/netty/Client.java
```diff
@@ -42,344 +42,577 @@
...
+     * succeed  or not, and how long the recovery will take.
+     */
+    private void handleMessagesWhenConnectionIsUnavailable(Iterator<TaskMessage> msgs) {
+        LOG.error("connection to {} is unavailable", dstAddressPrefixedName);
+        dropPendingMessages(msgs);
+    }
+
+    private void dropPendingMessages(Iterator<TaskMessage> msgs) {
+        // We consume the iterator by traversing and thus "emptying" it.
+        int msgCount = iteratorSize(msgs);
```
revans2: We are throwing away messages, it would be nice to include that in the state so we can know how many messages we threw away waiting for a worker to come up.

storm-core/src/jvm/backtype/storm/messaging/netty/Client.java
```diff
@@ -42,344 +42,577 @@
...
+    /**
+     * The channel used for all write operations from this client to the remote destination.
+     */
+    private final AtomicReference<Channel> channelRef = new AtomicReference<Channel>(null);
+
+
+    /**
+     * Maximum number of reconnection attempts we will perform after a disconnect before giving up.
+     */
+    private final int maxReconnectionAttempts;
```
revans2: I'm not sure we want a maximum number of reconnection attempts, or if we have it we need to think about how we intend to recover from it.  This could be a separate JIRA, but we ran into a situation where we hit the maximum number of reconnection attempts, and the exception was eaten because it was thrown from a background thread and it just killed the background thread.  This code appears to do the same thing.

storm-core/src/jvm/backtype/storm/messaging/netty/Client.java
miguno: Fixed.  Good catch, Bobby!

storm-core/src/jvm/backtype/storm/messaging/netty/Client.java
miguno: I see your point, Bobby, and I also can confirm the behavior you described.

Right now the behavior is as follows.  If the maximum number of reconnection attempts is reached, we enable the flag `closing`.  One effect of this is that any subsequent `send()`ing of messages will result in those messages being discarded.

Example log message:

```
2015-02-16 10:06:29 b.s.m.n.Client [ERROR] discarding 1 messages because the Netty client to Netty-Client-supervisor4/10.0.0.104:6701 is being closed
```

A second effect is that the Netty client will never attempt to reconnect again (by design).  But this has a negative effect, too, and I think this is what Bobby is alluding to.  If a Netty client's `closing` is true (= because max reconnects was reached), then a Storm task using this Netty client will never be able again to send a message to the target remote destination of the Netty client.  Only if the Storm task itself were to die will such a reconnect be possible because we would then also start a new Netty client.  The current code will not, as Bobby is pointing out, cause the parent Storm task (or even the Netty client) to die -- instead the client will stay forever in the `closing` state, and the parent Storm task will continue to call the client's `send()` method for new messages, which in turn will forever drop any such messages.

Off the hip I'd say there are at least three approaches for addressing this:

1. Let the Netty client die if max retries is reached, so that the Storm task has the chance to re-create a client and thus break out of the client's discard-messages-forever state.
2. Let the Storm task that uses the Netty client die if (one of its possibly many) Netty clients dies, so that by restarting the task we'll also get a new Netty client.
3. Remove the max retries semantics as well as the corresponding setting from Storm's configuration.  Here, a Netty client will continue to reconnect to a remote destination forever.  The possible negative impact of these reconnects (e.g. number of TCP connection attempts in a cluster) are kept in check by our exponential backoff policy for such connection retries.

My personal comments to these three approaches:

- I do not like (1) because I feel it introduces potentially confusing semantics: We keep having a max retries setting, but it is not really a hard limit anymore.  It rather becomes a "max retries until we recreate a Netty client", and would also reset any exponential backoff strategy of the "previous" Netty client instance (cf. `StormBoundedExponentialBackoffRetry`).  If we do want such resets (but I don't think we do at this point), then a cleaner approach would be to implement such resetting inside the retry policy (again, cf. `StormBoundedExponentialBackoffRetry`).
- I do not like (2) because a single "bad" Netty client would be able to take down a Storm task, which among other things would also impact any other, working Netty clients of the Storm task.
- Option (3) seems a reasonable approach, although it breaks backwards compatibility with regard to Storm's configuration (because we'd now ignore `storm.messaging.netty.max_retries`).

Personally, I am fine with either a separate JIRA or, if consensus is reached quickly here, to address this directly in this pull request.

If we want to address this directly in the pull request, we only need to change a single line in `Client.java` (apart from follow-up changes such as updating `conf/defaults.yaml` to remove `storm.messaging.netty.max_retries`):

```java
    private boolean reconnectingAllowed() {
        // BEFORE:
        // return !closing && connectionAttempts.get() <= (maxReconnectionAttempts + 1);
        return !closing;
    }
```

revans2: I personally prefer option 3, no maximum number of reconnection attempts.  Having the client decide that it is done, before nimbus does feels like it is asking for trouble.

revans2:
I am +1 for the code as is.  Perhaps a separate JIRA for the reconnection attempts would be best.  The issue has been in for quite a while now, and this code was not intended to fix it to begin with.

revans2:
I am seeing a lot of tests timing out with this change.  Has anyone else seen this?

miguno:
> I am seeing a lot of tests timing out with this change. Has anyone else seen this?

Hmm.  All the tests are passing for me (and they have been since a while).

Do you have any pointers?

miguno:
And FWIW, with the code in this PR the total test suite takes about 5mins to complete.

```
$ mvn clean install
...

[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary:
[INFO]
[INFO] Storm .............................................. SUCCESS [  1.696 s]
[INFO] maven-shade-clojure-transformer .................... SUCCESS [  1.930 s]
[INFO] storm-maven-plugins ................................ SUCCESS [  2.033 s]
[INFO] Storm Core ......................................... SUCCESS [03:56 min]
[INFO] storm-starter ...................................... SUCCESS [  6.616 s]
[INFO] storm-kafka ........................................ SUCCESS [ 48.469 s]
[INFO] storm-hdfs ......................................... SUCCESS [  1.875 s]
[INFO] storm-hbase ........................................ SUCCESS [  2.271 s]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 05:02 min
[INFO] Finished at: 2015-02-17T17:49:19+01:00
[INFO] Final Memory: 67M/282M
[INFO] ------------------------------------------------------------------------
```

revans2:
I don't it could just be my mac acting funny.  I'll dig into it, but don't block the pull request on me.  I saw similar things on a different pull request that corrected themselves spontaneously.  My guess right now is that it could have something to do with being on battery power vs being plugged into the wall.

revans2:
OK It is something odd with my mac. It looks like the difference is wired vs wireless networking :), or possibly even switching between the two.

miguno:
Glad to hear it was only a temporary issue. :-)

To summarize the current status of this pull request:

- We have two +1 already.
- The max-reconnection-attempts issue that Bobby brought up would be addressed in a separate JIRA ticket (which I can create if no one else wants to).  And we already have an initial idea how to solve the issue.

Let me know if I forgot something.

@nathanmarz Are you ok, too?

storm-core/src/jvm/backtype/storm/messaging/netty/Client.java
ptgoetz: I'm in favor of option 3 as well. I'm not that concerned about `storm.messaging.netty.max_retries` being ignored. We could probably just log a warning that that configuration option is deprecated and will be ignored if the value is set.

ptgoetz:
+1 I was able to verify the fix, and am in favor of merging. I'd also like to apply it to the 0.9.x branch as I feel it's an important fix.

miguno:
FYI: I created [STORM-677: Maximum retries strategy may cause data loss](https://issues.apache.org/jira/browse/STORM-677) to address the issue that Bobby brought up in this discussion.

miguno:
> I'd also like to apply it to the 0.9.x branch as I feel it's an important fix.

That's a good idea, Taylor.  Would you also like volunteer to merge this PR in general? (otherwise let me know and I'll take care of it)

miguno:
PS: We may also want to update the original [STORM-329](https://issues.apache.org/jira/browse/STORM-329) ticket description to reflect the changes in this PR.

ptgoetz:
@miguno No problem, I can do the merge to master. I've done a test merge to the 0.9.3 branch, and it didn't seem too bad, but I'm still debugging some test failures.

miguno:
Thanks, Taylor!  Let me know if I can help with sorting out the test failures.

Also regarding JIRA: I forgot to mention that it looks like we need to update STORM-404 and STORM-510 as well as STORM-329 has been said to cover those issues, too.  It looks to me like this is actually the case, but I'd prefer another pair of eyes to be sure we're not mistakenly closing two related tickets.

nathanmarz:
Nimbus only knows a worker is having trouble when it stops sending heartbeats. If a worker gets into a bad state, the worst thing to do is have it continue trying to limp along in that bad state. It should instead suicide as quickly as possible. It seems counterintuitive, but this aggressive suiciding behavior actually makes things more robust as it prevents processes from getting into weird, potentially undefined states. This has been a crucial design principle in Storm from the beginning. One consequence of it is that any crucial system thread that receives an unrecoverable exception must suicide the process rather than die quietly. 

For the connection retry problem, it's a tricky situation since it may not be able to connect because the other worker is still getting set up. So the retry policy should be somehow related to the launch timeouts for worker processes specified in the configuration. Not being able to connect after the launch timeout + a certain number of attempts + a buffer period would certainly qualify as a weird state, so the process should suicide in that case. *Suiciding and restarting gets the worker back to a known state*. 

So in this case, I am heavily in favor of Option 2. I don't care about killing the other tasks in the worker because this is a rare situation. It is infinitely more important to get the worker back to a known, robust state than risk leaving it in a weird state permanently.

I would like to see these issues addressed as part of this patch.

@miguno Thanks for the explanation on this patch's relation to backpressure – we'll handle that in a future patch.

revans2:
@nathanmarz I'd like to move the conversation over to https://issues.apache.org/jira/browse/STORM-677.

miguno:
@nathanmarz Thanks for the detailed feedback on the max-retries issue.  As Bobby suggested, would you mind if we decouple the work on max-retries (tracked at STORM-677) from this pull request / STORM-329?  The max-retries issue has been a problem in Storm for a while, since before this pull request.  Decoupling would also help us with backtracking any future issues to either ticket.  Right now, I feel there's almost too much meat in this pull request already (as we kinda covered STORM-329, STORM-404, and STORM-510 in one big swing).

Would that be ok for you?

ptgoetz:
@miguno Regarding the merge to 0.9.x, I didn't have that much time to debug the test failures. There are 2 errors in `auth_test.clj` when run as part of the maven build. When run in a repl, I don't get any failures.

Anyway, the results of the merge are here:

https://github.com/ptgoetz/storm/tree/storm-329-0.9

I'll try to dig in a little deeper when I have more time.

miguno:
What were the exact REPL and mvn commands?

ptgoetz:
For the build, I just ran `mvn clean install` at the top level of the source tree.

For the repl, I ran (in the Intellij repl):

```
(load-file "/path/to/checkout/storm/security/auth/auth_test.clj")
(ns backtype.storm.security.auth.auth-test)
(run-tests)
```

That yielded no errors.

If I run the same in the maven-clojue-plugin repl:

```
cd storm-core
mvn clojure:repl
```

```
(load-file "/path/to/checkout/storm/security/auth/auth_test.clj")
(ns backtype.storm.security.auth.auth-test)
(run-tests)
```

I get one error.

miguno:
FYI: I took the opportunity to update [DEVELOPER.md](https://github.com/apache/storm/blob/master/DEVELOPER.md#building) with documentation how to run tests selectively in the REPL.

ptgoetz:
I merged this locally, and it tested out fine, but there are now a few merge conflicts. Resolving them seemed straightforward, but now the behavior I see in the unit tests is a never-ending loop of zookeeper reconnects.

@revans2 @d2r @harshach Do you mind doing a sanity check against what was recently merged?

The results of the merge can be found here:https://github.com/ptgoetz/storm/tree/storm-329-merge

ptgoetz:
Disregard last message. It was a merge mistake (picked right when I should have picked left).

All tests are passing now.

miguno:
Phew. :-)

Thanks for merging, Taylor!

miguno:
Oh, Taylor. Could you also update STORM-404 and STORM-510 as appropriate?

ptgoetz:
I finally got this successfully back ported to the 0.9.x branch, with all tests passing. I will merge that soon after more testing and update all associated JIRAs.

Thanks for all the effort that went into this!

danielschonfeld:
@ptgoetz sorry for being the newbie here, but does this mean a new jar will be built of 0.9.3 that will include this? or will we have to wait for 0.10.0? 

we could really use this fix

ptgoetz:
@miguno yep, that's what I meant.

ptgoetz:
@danielschonfeld No worries. There will be a 0.9.4 release that includes this fix. No need to wait for 0.10.0.

danielschonfeld:
@ptgoetz what is the time frame for 0.9.4? (roughly)

ptgoetz:
@danielschonfeld Soon. ;)

We are probably a handful of weeks out, hopefully less.

If you have the resources, testing an RC and reporting back is the best way to accelerate a release.

ptgoetz:
Patch ported and merged to 0.9.x branch.

Benchmarked several topologies (both core and Trident) before (0.9.3) and after (0.9.4-SNAPSHOT) this patch and found no performance regressions.

3in:
@ptgoetz 
If you have the resources, testing an RC and reporting back is the best way to accelerate a release.
---------
where do i get the rc, i will have a try,
https://github.com/ptgoetz/storm/tree/storm-329-merge    not existed
thanks

PR #430, opened 2015-02-12
SimpleTransportPlugin Using TThreadPoolServer

darionyaphet:
I'm reading storm's source code and I found it's very difference from version 0.9.3 . Storm Nimbus's ThriftServer is create by SimpleTransportPlugin. In getServer() function build a THsHaServer server to process supervisor's request . 
I know TThreadPoolServer maybe support higher throughput and lesser lantency. So I make this pull request. 
Also some config could be remove such as queueSize .


d2r:
-0 unless we have evidence of a real performance issue that needs fixing here.

@darionyaphet, If this pull request is related to an existing JIRA issue, it would help us out if you would update the title of this pull request with [STORM-X].  

Here is a [link](https://issues.apache.org/jira/browse/STORM-510?jql=project%20%3D%20STORM%20AND%20resolution%20%3D%20Unresolved%20ORDER%20BY%20priority%20DESC) to some currently unresolved issues in the JIRA system, for reference.

d2r:
@darionyaphet, I will be closing this pull request soon since it is inactive.  Please re-open if it is a mistake.  If you do, it would be good to have a Jira issue and some performance numbers that demonstrate the issue.

PR #432, opened 2015-02-15
STORM-702: Exhibitor support

atdixon:
This introduces optional exhibitor configuration properties to storm config--the options parallel the zookeeper options where relevant:

   storm.exhibitor.servers:
      - "localhost"
   storm.exhibitor.port: 8080
   storm.exhibitor.poll.millis: 1000
   storm.exhibitor.retry.times: 5
   storm.exhibitor.retry.interval: 1000
   storm.exhibitor.retry.intervalceiling.millis: 30000

If exhibitor.servers are not provided, the behavior of storm is identical to prior. If exhibitor.servers are provided then the curator connection uses exhibitor, using any configured zookeeper.servers for the curator-exhibitor backup connection string.

I've tested this locally, will be deploying this patch in coming weeks in my own production environment/s.

### DIFF
--- a/storm-core/src/jvm/backtype/storm/utils/Utils.java
+++ b/storm-core/src/jvm/backtype/storm/utils/Utils.java
@@ -35,21 +35,34 @@ import org.apache.commons.compress.archivers.tar.TarArchiveEntry;
 import org.apache.commons.compress.archivers.tar.TarArchiveInputStream;
 import org.apache.commons.io.input.ClassLoaderObjectInputStream;
 import org.apache.commons.lang.StringUtils;
+import org.apache.curator.ensemble.exhibitor.DefaultExhibitorRestClient;
+import org.apache.curator.ensemble.exhibitor.ExhibitorEnsembleProvider;
+import org.apache.curator.ensemble.exhibitor.Exhibitors;
+import org.apache.curator.framework.CuratorFramework;
+import org.apache.curator.framework.CuratorFrameworkFactory;
+import org.apache.thrift.TException;
+import org.apache.zookeeper.ZooDefs;
+import org.apache.zookeeper.data.ACL;
+import org.apache.zookeeper.data.Id;
+import org.json.simple.JSONValue;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.yaml.snakeyaml.Yaml;
+import org.yaml.snakeyaml.constructor.SafeConstructor;
+
 import java.io.BufferedReader;
-import java.io.ByteArrayInputStream;
-import java.io.ByteArrayOutputStream;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
-import java.io.ObjectInputStream;
-import java.io.ObjectOutputStream;
 import java.net.URL;
 import java.net.URLDecoder;
 import java.nio.ByteBuffer;
 import java.nio.channels.Channels;
 import java.nio.channels.WritableByteChannel;
 import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
 import java.util.Enumeration;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -58,6 +71,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.TreeMap;
 import java.util.UUID;
+
 import java.util.Collection;
 import java.util.Collections;
 
@@ -666,6 +680,13 @@ public class Utils {
         }
     }
 
+    public static String getString(Object o) {
+        if (null == o) {
+            throw new IllegalArgumentException("Don't know how to convert null to String");
+        }
+        return o.toString();
+    }
+
     public static Integer getInt(Object o) {
         Integer result = getInt(o, null);
         if (null == result) {
@@ -1016,7 +1037,7 @@ public class Utils {
 
     public static CuratorFramework newCurator(Map conf, List<String> servers, Object port, String root, ZookeeperAuthInfo auth) {
         List<String> serverPorts = new ArrayList<String>();
-        for (String zkServer : (List<String>) servers) {
+        for (String zkServer: servers) {
             serverPorts.add(zkServer + ":" + Utils.getInt(port));
         }
         String zkStr = StringUtils.join(serverPorts, ",") + root;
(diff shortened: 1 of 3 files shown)


lazyval:
For those who wanders what exhibitor is -- [it's a zookeeper supervisor used for monitoring, back up and visualization](https://github.com/Netflix/exhibitor). 

@atdixon are you sure this properties should be kept in `storm.yaml`? Why should storm be aware of them? Is there a ticket backing this change (as it's not clear what are benefits of such support)?

atdixon:
Zookeeper and its clients require static list of hosts in its configuration [1]. Exhibitor (in addition to supervision, backup, etc.) can provide the Zookeeper server list to Curator clients (via Curator's Exhibitor support [2]), so that the client can dynamically update its Zookeeper servers list.

This provides a more robust connection to our Zookeeper ensemble from Storm (supporting zookeeper cluster scale-up/down, whole node replacements, rolling upgrades, outages, etc. all w/o restarting Storm.) It also allows Storm to discover our Zookeeper cluster through curator-exhibitor instead of requiring the full Zookeeper ensemble to be hard-listed into the Storm configuration.

With this patch, you can (optionally) provide Exhibitor servers *in place of* Zookeeper servers in the storm config and Storm will discover your Zookeeper ensemble. This is why it should go into storm.yaml.

[1] http://zookeeper.apache.org/doc/r3.5.0-alpha/zookeeperProgrammers.html#ch_gotchas
[2] http://curator.apache.org/exhibitor.html

revans2:
I like the idea of this, and I am +1 on the concept, but we really need a JIRA for tracking this.  Could you please file a JIRA and include the JIRA number in the title of this pull request.

atdixon:
Done, I've added the ticket number to the title. Here is the ticket: https://issues.apache.org/jira/browse/STORM-702

atdixon:
Note: I've added a short bit on design options and why I recommend this implementation in a comment attached to the JIRA ticket.

harshach:
@atdixon @revans2 with this patch can a user who doesn't want use exhibitor avoid it by not having the config or the user must have the config in their storm.yaml.

atdixon:
The config is optional. If not present, everything works as usual.

revans2:
The code looks good to me and there is a JIRA for it.  I am +1, but @harshach I'll wait to hear from you before I check anything in.

storm-core/src/jvm/backtype/storm/utils/Utils.java
```diff
@@ -363,9 +384,30 @@ public static CuratorFramework newCurator(Map conf, List<String> servers, Object
...
 
-    protected static void setupBuilder(CuratorFrameworkFactory.Builder builder, String zkStr, Map conf, ZookeeperAuthInfo auth)
+    protected static void setupBuilder(CuratorFrameworkFactory.Builder builder, final String zkStr, Map conf, ZookeeperAuthInfo auth)
     {
-        builder.connectString(zkStr)
+        List<String> exhibitorServers = getStrings(conf.get(Config.STORM_EXHIBITOR_SERVERS));
+        if (!exhibitorServers.isEmpty()) {
+            // use exhibitor servers
+            builder.ensembleProvider(new ExhibitorEnsembleProvider(
+                new Exhibitors(exhibitorServers, Utils.getInt(conf.get(Config.STORM_EXHIBITOR_PORT), 8080),
```
harshach: Instead of having 8080 as port here in the code can you add it to defaults.yaml

harshach:
@atdixon  minor nit pick above. I am trying to test this by running exhibitor, can you provide a sample config for running exhibitor with single host zookeeper. 
I am also concerned about how it will run in secure mode. There is an option to set ACLs https://github.com/Netflix/exhibitor/wiki/Running-Exhibitor but we need to test this as well.
cc: @revans2

storm-core/src/jvm/backtype/storm/utils/Utils.java
```diff
@@ -363,9 +384,30 @@ public static CuratorFramework newCurator(Map conf, List<String> servers, Object
...
+                    new Exhibitors.BackupConnectionStringProvider() {
+                        @Override
+                        public String getBackupConnectionString() throws Exception {
+                            // use zk servers as backup if they exist
+                            return zkStr;
+                        }}),
+                new DefaultExhibitorRestClient(),
+                (String) Utils.get(conf, Config.STORM_EXHIBITOR_URIPATH, "/exhibitor/v1/cluster/list"),
```
harshach: this should also be in defaults.yaml

atdixon:
@harshach implemented your suggestions, thank you.

The *storm* configuration I used to test (w/ single zk node, too) is in the first comment of this thread.

To configure and run exhibitor locally, I used:

java -cp './lib/*' com.netflix.exhibitor.application.ExhibitorMain -c file --defaultconfig ./exhibitor-default.properties --hostname localhost

And my exhibitor-default.properties looked like so:

<pre>
java-environment=
zookeeper-data-directory=/Users/atdixon/work/srvr/exhibitor/1.5.4/zookeeper-3.4.6/snapshots
servers-spec=S\:1\:localhost
cleanup-period-ms=300000
auto-manage-instances-fixed-ensemble-size=0
zookeeper-install-directory=/Users/atdixon/work/srvr/exhibitor/1.5.4/zookeeper-3.4.6
check-ms=30000
zookeeper-log-directory=/Users/atdixon/work/srvr/exhibitor/1.5.4/zookeeper-3.4.6/transactions
log-index-directory=/Users/atdixon/work/srvr/exhibitor/1.5.4/zookeeper-3.4.6/transactions
backup-period-ms=60000
client-port=2181
cleanup-max-files=20
backup-max-store-ms=86400000
connect-port=2888
backup-extra=
observer-threshold=0
log4j-properties=
auto-manage-instances-apply-all-at-once=1
election-port=3888
zoo-cfg-extra=tickTime\=2000&initLimit\=10&syncLimit\=5&&datadir.autocreate\=false
auto-manage-instances-settling-period-ms=0
auto-manage-instances=1
</pre>

storm-core/src/jvm/backtype/storm/utils/Utils.java
```diff
@@ -306,7 +327,30 @@ public static ComponentCommon getComponentCommon(StormTopology topology, String
         }
         throw new IllegalArgumentException("Could not find component with id " + id);
     }
-    
+
+    public static List<String> getStrings(final Object o) {
+        if (o == null) {
+            return Collections.emptyList();
```
knusbaum: All of the other lists here are mutable, but this isn't. 
I'd rather see a new ArrayList, since it's a public method.

knusbaum:
@revans2, @harshach  Any updates here?

revans2:
@atdixon sorry this dropped off of my radar.  The code looks like it needs to be upmerged and retested, but it looks like it should not be too difficult.  Mostly white-space changes.

Do you still want this to go in?

atdixon:
@revans2 i've rebased and made updates per cr. mvn tests are passing. will try to test full integration scenario tomorrow.

atdixon:
I've tested this fix against the latest code (see thread above for how I am testing).

Anything else to consider?

I've noticed there is "storm.yaml.example" -- perhaps it would be a good idea to add into this file how to configure exhibitor. Thoughts?

revans2:
+1 again.  It would be nice to add some documentation, but I would like to get this in for a 1.0 branch, so we can probably do it in a follow on JIRA.

PR #312, opened 2014-11-13
Update "get-task-object" function, change the type of first param "topology" from ^TopologyContext to ^StormTopology

BuDongDong:
update "get-task-object" function, change the type of first param "topology" from ^TopologyContext to ^StormTopology

PR #433, opened 2015-02-16
Update "get-task-object" function, change the type of first param "topology" from ^TopologyContext to ^StormTopology.(JIRA [STORM-554])

BuDongDong:
Update "get-task-object" function, change the type of first param "topology" from ^TopologyContext to ^StormTopology. the "get-task-object" function in task.clj,the type of first param "topology" should be ^StormTopology not ^TopologyContext.

the "get-task-object" is called by "mk-task-data" function in task.clj, "mk-task-data" is defined as following:

(defn mk-task-data [executor-data task-id]
  (recursive-map
    :executor-data executor-data
    :task-id task-id
    :system-context (system-topology-context (:worker executor-data) executor-data task-id)
    :user-context (user-topology-context (:worker executor-data) executor-data task-id)
    :builtin-metrics (builtin-metrics/make-data (:type executor-data))
    :tasks-fn (mk-tasks-fn <>)
    :object (get-task-object (.getRawTopology ^TopologyContext (:system-context <>)) (:component-id executor-data))))

(:system-context <>) return TopologyContext instance, TopologyContext extends GeneralTopologyContext, the TopologyContext instance has StormTopology _topology. “getRawTopology” method of TopologyContext must return StormTopology _topology. (.getRawTopology ^TopologyContext (:system-context <>) return StormTopology instance not TopologyContext instance. so the type of param topology is StormTopology not TopologyContext.

at the same time, "get-task-object" call "get_spouts" and "get_bolts" of the param topology. you can find the "get_spouts" and "get_bolts" function are only defined in StormTopology, and StormTopology is not a subclass of TopologyContext. so i think the type of param topology is StormTopology not TopologyContext.

### DIFF
--- a/storm-core/src/clj/backtype/storm/daemon/task.clj
+++ b/storm-core/src/clj/backtype/storm/daemon/task.clj
@@ -18,7 +18,7 @@
   (:use [backtype.storm bootstrap])
   (:import [backtype.storm.hooks ITaskHook])
   (:import [backtype.storm.tuple Tuple])
-  (:import [backtype.storm.generated SpoutSpec Bolt StateSpoutSpec])
+  (:import [backtype.storm.generated SpoutSpec Bolt StateSpoutSpec StormTopology])
   (:import [backtype.storm.hooks.info SpoutAckInfo SpoutFailInfo
             EmitInfo BoltFailInfo BoltAckInfo])
   (:require [backtype.storm [tuple :as tuple]])
--- a/storm-core/src/clj/backtype/storm/daemon/task.clj
+++ b/storm-core/src/clj/backtype/storm/daemon/task.clj
@@ -61,7 +61,7 @@
     (:topology worker))
    tid))
 
-(defn- get-task-object [^TopologyContext topology component-id]
+(defn- get-task-object [^StormTopology topology component-id]
   (let [spouts (.get_spouts topology)
         bolts (.get_bolts topology)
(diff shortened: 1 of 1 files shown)


d2r:
+1 Yes, it appears the type hint is wrong.  This works anyway because the method calls in `get-task-object` are correct for the actual class (`StormTopology`) of the instance.  But we should correct the type hint.

d2r:
@BuDongDong Would you add [STORM-554] to the title of this pull request, and close #312 ?

BuDongDong:
@d2r I have added [STORM-554] to the title of this pull request and closed  #312.

d2r:
@BuDongDong, OK, got a compile error because STORM-441 went in since you made your change.  So we need to up-merge this branch and add an import for StormTopology, since it is no longer pulled in via bootstrap.

```Diff
-  (:import [backtype.storm.generated SpoutSpec Bolt StateSpoutSpec])
+  (:import [backtype.storm.generated SpoutSpec Bolt StateSpoutSpec
+                                     StormTopology])
```

Git will merge the file successfully despite this conflict, so if you could make this change to tasks.clj and push to your branch, hopefully it will work.

BuDongDong:
@d2r I haved imported backtype.storm.generated.StormTopology in task.clj.

d2r:
OK, thanks.  I will take care of the merge conflict.  I want to run the tests once.

d2r:
+1

PR #434, opened 2015-02-18
STORM-675. Allow users to have storm-env.sh under config dir to set custom JAVA_HOME and other env variables.

harshach:


HeartSaVioR:
AFAIK, STORM-487 was intended to lay the foundation of unifying storm launch script.
Do we have plan to change storm.cmd to use storm.py?

harshach:
@HeartSaVioR yes but wanted to keep the .cmd files for at least a release or two as there might be users who are depend on these. Since storm.py is now windows compatabile we can phase out the current storm.cmd

HeartSaVioR:
Great! Looks good to me.
Btw, it's better to point out storm.py's actual changed lines since it's hard to find difference.

harshach:
@HeartSaVioR bin/storm which used to be a python script now moved to bin/storm.py and in place of bin/storm we added storm shell script to give users a way to configure storm env variables. Because of this renaming files you see the diff like that.

HeartSaVioR:
@harshach I know, I just want to point out that bin/storm is not only renamed but also modified a bit.

Parth-Brahmbhatt:
+1.

harshach:
@ptgoetz can you please take a look at this.

bin/storm
```diff
@@ -1,5 +1,7 @@
-#!/usr/bin/python
-
+#!/bin/bash
+#
+# Copyright 2014 The Apache Software Foundation
```
ptgoetz: We shouldn't need the copyright year in the license header.

conf/storm-env.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+#
+# Copyright 2014 The Apache Software Foundation
```
ptgoetz: Same as previous comment.

harshach:
@ptgoetz Thanks for the review. Removed the year from LICENSE.

ptgoetz:
+1. The diffs of `storm` vs. `storm.py` are minimal and related to windows compatibility.

For easy reference, here they are:

```
$ diff -w storm storm.py
37a38,40
> def is_windows():
>     return sys.platform.startswith('win')
>
174c177
<         JAVA_CMD, jvmtype, get_config_opts(),
---
>         "java", jvmtype, get_config_opts(),
184c187
<     else:
---
>     elif is_windows():
186a190,191
>     else:
>         os.execvp(JAVA_CMD, all_args)
```

PR #431, opened 2015-02-12
STORM-670: restore java 1.6 compatibility (storm-kafka)

ptgoetz:
Not sure how long we want to cling to 1.6 compatibility, but this is a very small change.

PR #435, opened 2015-02-19
STORM-680: storm-kafka fails to compile with JDK 1.6

jayapriya90:
Replaced Long.compare() method of JDK 7 with compareTo() method of JDK 6.


jayapriya90:
Closing pull request as it is duplicate of STORM-670. #431

PR #436, opened 2015-02-19
[STORM-681] Auto insert license header with genthrift.sh

Lewuathe:
Current genthrift.sh does not insert license headers into generated source codes. These java codes and python codes should have license headers. 
And documentation about this command.


Parth-Brahmbhatt:
+1. I am not sure if the date change is something we want to include for all the existing files that has no other change.

Lewuathe:
@Parth-Brahmbhatt Yes, I'm not sure too. But I think it is better to include the files generated by new `genthrift.sh` script in order to keep and confirm whole source code consistency.

storm-core/src/genthrift.sh
```diff
@@ -17,6 +17,16 @@
...
 thrift7 --gen java:beans,hashcode,nocamel --gen py:utf8strings storm.thrift
+for file in gen-javabean/backtype/storm/generated/* ; do
+  cat java_license_header.txt ${file} > ${file}.tmp
+  mv ${file}.tmp ${file}
+done
+cat py_license_header.txt gen-py/__init__.py > gen-py/__init__.py.tmp
+mv gen-py/__init__.py.tmp gen-py/__init__.py
+for file in gen-py/storm/* ; do
+  cat py_license_header.txt ${file} > ${file}.tmp
+  mv ${file}.tmp ${file}
```
revans2: Could we include a `-f` on the mv command?  I know a lot of us have mv configured as an alias to `mv -i` to avoid unexpected overwrites and this would cause problems.

revans2:
I just had the one comment.  As for the date in the generated files I see two options.

1) We remove it with a sed script.  It is optional, and thrift is not following the formatting convention for it anyways
2) We eventually move to generating the code for the build each time.

I don't think either of these should be something we worry about now.  The thrift code changes rarely enough that I am fine with ignoring it until it is a problem.

knusbaum:
We should also change `thrift7` to `thrift` seeing as we've moved onto Thrift 0.9.2

Lewuathe:
@revans2 Thank you for reviewing. I exclude generated files in this patch. And I'll update.

@knusbaum Yes. We should change `thrift7` to `thrift`. Thank you.

revans2:
+1 looks good to me.

knusbaum:
+1

PR #440, opened 2015-02-24
STORM-687. Storm UI does not display up to date information despite refreshes in IE.

harshach:


revans2:
Why are we turning off caching on the client side, instead of having proper cache headers on the server response?  The code looks fine to me, just a bit confused if this is IE specific and it is ignoring the caching headers, or if we just are not including the proper ones.

harshach:
@revans2 I tried with setting Cache-Control to no-cache but didnt' see any affect IE but adding no-cache, no-store works with all the browsers. Removed client side settings from JS.

revans2:
Thanks, that looks better to me.  +1

Parth-Brahmbhatt:
+1.

PR #445, opened 2015-02-26
STORM-689. SimpleACLAuthorizer should provide a way to restrict who can submit topologies.

harshach:


harshach:
@revans2 can you please take a look at this.

harshach:
@revans2  pinging for a review. Thanks.

storm-core/src/jvm/backtype/storm/Config.java
```diff
@@ -339,6 +339,13 @@
     public static final Object NIMBUS_ADMINS_SCHEMA = ConfigValidation.StringsValidator;
 
     /**
+     * A list of users that are the only ones allowed to run user operation on storm cluster.
+     * To use this set nimbus.authorizer to backtype.storm.security.auth.authorizer.SimpleACLAuthorizer
+     */
+    public static final String NIMBUS_USERS = "nimbus.users";
```
Parth-Brahmbhatt: We should modify security.MD to document this config.

Parth-Brahmbhatt:
+1. I think most users would much rather define a nimbus.groups m e can probably file a separate jira for that.

storm-core/src/jvm/backtype/storm/security/auth/authorizer/SimpleACLAuthorizer.java
```diff
@@ -96,7 +101,10 @@ public boolean permit(ReqContext context, String operation, Map topology_conf) {
         }
 
         if (_userCommands.contains(operation)) {
-            return true;
+            if (_nimbusUsers.size() > 0 && _nimbusUsers.contains(user))
+                return true;
+            else if (_nimbusUsers.size() == 0)
+                return true;
```
revans2: I personally think 
```
return _nimbusUsers.size() == 0 || _nimbusUsers.contains(user);
```
is more readable than what is here, but this is really minor.

revans2:
Sorry it took so long for the review.  It looks good, and I agree having groups of allowed users would be a good step.

harshach:
@Parth-Brahmbhatt @revans2 Thanks for the review. I updated the PR to include nimbus.groups as well and added doc under SECURITY.md. Please take a look.

Parth-Brahmbhatt:
+1.

vesense:
+1

revans2:
+1

PR #446, opened 2015-02-26
STORM-496. task.clj missing debug for logging spout and bolt emit values.

harshach:


harshach:
@revans2  @ptgoetz  can you please take a look at this.

ptgoetz:
+1

PR #448, opened 2015-02-27
STORM-446: Allow superusers to impersonate other users in secure mode.

Parth-Brahmbhatt:


storm-core/src/jvm/backtype/storm/security/auth/authorizer/SimpleACLAuthorizer.java
```diff
@@ -50,6 +50,7 @@
     protected Set<String> _supervisors;
     protected IPrincipalToLocal _ptol;
     protected IGroupMappingServiceProvider _groupMappingProvider;
+    protected ImpersonationAuthorizer _impersonationAuthorizer;
```
revans2: I don't really like the idea of having all authorizers have to be updated to support the impersonation code.  I would much rather have the code spliced in at a much higher level that cannot be turned off.  Otherwise a custom IAuthorizer instance that does not know to update after this change is now wide open for anyone to impersonate anyone else.  The logging changes should also probably happen at the same level.

revans2:
I haven't gone deeply into the code, but for the most part it looks good.  Just one request that should hopefully not be too difficult to change.

storm-core/src/jvm/backtype/storm/security/auth/authorizer/SimpleACLAuthorizer.java
Parth-Brahmbhatt: Ok, I am going to add this authorization as part of nimbus check-authorization! method. I will also add "nimbus.impersonation.authorizer" config to allow for overriding behavior. I will also add the logging changes there.

Parth-Brahmbhatt:
@revans2 @harshach  For supporting the actual jira issue "ui server being able to impersonate users" we probably need to support wildcards ("allow userX to impersonate all users(*) from localhost") or allow the "nimbus.admin" users to impersonate all users from all hosts.  Any preferences?

revans2:
Yes we would need something like that for this to work.  I'm not sure I want all administrators to have that power though.

Parth-Brahmbhatt:
@revans2 I agree. I can allow users to specify regex for matching but that seems like an overkill. For now   I am just going to make * as special character for both groups and hosts.

Parth-Brahmbhatt:
@revans2 I added the fixes. I also added the doAs support to REST API.

storm-core/src/clj/backtype/storm/daemon/nimbus.clj
```diff
@@ -778,9 +779,22 @@
...
+
+       (if (.isImpersonating ctx)
+         (do
+          (log-warn "principal: " (.realPrincipal ctx) " is trying to impersonate principal: " (.principal ctx))
+          (if impersonation-authorizer
+           (if-not (.permit impersonation-authorizer ctx operation check-conf)
+             (throw (AuthorizationException. (str "principal " (.realPrincipal ctx) " is not authorized to impersonate
+                        principal " (.principal ctx) " from host " (.remoteAddress ctx) " Please see SECURITY.MD to learn
+                        how to configure impersonation acls."))))
+           (log-warn "impersonation attempt but " NIMBUS-IMPERSONATION-AUTHORIZER " has no authorizer configured. potential
```
revans2: Do we want to fail open or fail closed in this case.  For me if someone tries to do impersonation and we have not configured anything for this I would rather fail closed.

storm-core/src/clj/backtype/storm/ui/core.clj
```diff
@@ -45,28 +45,46 @@
...
+            (if-not (.permit *UI-IMPERSONATION-HANDLER* context op topology-conf)
+              (let [principal (.principal context)
+                    real-principal (.realPrincipal context)
+                    user (if principal (.getName principal) "unknown")
+                    real-user (if real-principal (.getName real-principal) "unknown")
+                    remote-address (.remoteAddress context)]
+                (throw (AuthorizationException.
+                         (str "user '" real-user "' is not authorized to impersonate user '" user "' from host '" remote-address "'. Please
+                         see SECURITY.MD to learn how to configure impersonation ACL.")))))
+          (log-warn " principal " (.realPrincipal context) " is trying to impersonate " (.principal context) " but "
```
revans2: Again do we want to fail open or fail closed?

revans2:
For the most part it looks great.  Just a little curious if we really want to fail open when no impersonation authorizer is configured?  I am +1 either way.

Parth-Brahmbhatt:
We should fail close, however currently if the acl authorizer is no set we fail open. Not sure why did we make that choice but I did not want alternating behavior. My personal preference is to fail close on any potential security configuration mistake.

revans2:
The fail open on the general authorizer was to preserve backwards compatibility for existing connections.  But I can see the point there too, we probably should file a JIRA to change it to fail closed there too.  Perhaps we can do it for both at the same time.  Like I said I am still +1 with the code as it is.

STORM-UI-REST-API.md
```diff
@@ -32,6 +32,11 @@ You can use a tool such as `curl` to talk to the REST API:
     # Note: We assume ui.port is configured to the default value of 8080.
     $ curl http://<ui-host>:8080/api/v1/cluster/configuration
 
+##Impersonating a user in secure environment
+In a secure environment an authenticated user can impersonate another user. To impersonate a user the caller must pass
+`doAsUser` param or header with value set to the user that the request needs to be performed as. Please see SECURITY.MD
+to learn more about how to setup impersonation ACLs and authorization. The rest API uses the same configs and acls that
+are used by nimbus.
```
harshach: can you add a curl example for this.

Parth-Brahmbhatt: Added and upmerged.

harshach:
@Parth-Brahmbhatt  Overall code looks good to me. I am trying to run few tests in secure cluster  but during mvn build the following unit test is failing  <testcase name="test-populate-req-context-on-null-user" classname="backtype.storm.security.auth.DefaultHttpCredentialsPlugin-test">
            <failure>expected: (= 0 (-&gt; handler (.populateContext context req) (.subject) (.getPrincipals) (.size)))
  actual: (not (= 0 1))
      at: run-test2660255144893935715.clj:85</failure>
        </testcase>
Can you please take a look.

Parth-Brahmbhatt:
@harshach sorry about that I introduced that failure when I added the rest support and forgot to run the unit tests after that. I have fixed the failing tests.

harshach:
@Parth-Brahmbhatt  Thanks for the quick fix. Ran some tests on the cluster this looks great. +1

PR #451, opened 2015-02-28
STORM-691 Add basic lookup / persist bolts

HeartSaVioR:
* Add Basic lookup / persist Bolts
  * support data types : string, list, hash, set, sorted set, hyperloglog
* rename util package to common

 Type | Read | Write
-------- | ------- | ------
STRING | GET (key) | SET (key, value)
HASH | HGET (key, field) | HSET (key, field, value)
LIST | LPOP (key) | RPUSH (key, value)
SET | SCARD (key) | SADD (key, member)
SORTED SET | ZSCORE (key, member) | ZADD (key, score, member)
HLL (HyperLogLog) | PFCOUNT (key) | PFADD (key, element)

### DIFF
--- a/external/storm-redis/src/main/java/org/apache/storm/redis/bolt/AbstractRedisBolt.java
+++ b/external/storm-redis/src/main/java/org/apache/storm/redis/bolt/AbstractRedisBolt.java
@@ -20,10 +20,10 @@ package org.apache.storm.redis.bolt;
 import backtype.storm.task.OutputCollector;
 import backtype.storm.task.TopologyContext;
 import backtype.storm.topology.base.BaseRichBolt;
-import org.apache.storm.redis.util.config.JedisClusterConfig;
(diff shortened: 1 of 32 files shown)


HeartSaVioR:
When this PR will be merged to master, we can apply RedisDataTypeDescription to Trident in order to support more data types.

HeartSaVioR:
@revans2 Could you take a look?

revans2:
The code looks fine to me.  My only comment is that it would be nice to have a default in most of the case statements, mostly as defensive programming so if something gets off we wold get an error instead of falling through and ignoring the tuples.

HeartSaVioR:
@revans2 Thanks for reviewing! It's updated. Could you please check again?

ptgoetz:
+1

HeartSaVioR:
@revans2 Could you take a look again?

revans2:
+1, again sorry about this taking so long, I'll merge it in.

PR #453, opened 2015-03-03
STORM-693: when bolt fails to write tuple, it should report error instea...

Parth-Brahmbhatt:
...d of silently acking.


ptgoetz:
+1

ptgoetz:
Merged to 0.9.x branch.

PR #455, opened 2015-03-03
STORM-625: don't leak netty clients when workers move

revans2:


revans2:
Turns out that @kishorvpatil found a bug in this during some testing and should be putting up a fixed pull request shortly.

PR #459, opened 2015-03-09
[STORM-2999] Add generics to static type of method parameters and variables where applicable

krichter722:
https://issues.apache.org/jira/browse/STORM-2999

### DIFF
--- a/examples/storm-hbase-examples/src/main/java/org/apache/storm/hbase/topology/TotalWordCounter.java
+++ b/examples/storm-hbase-examples/src/main/java/org/apache/storm/hbase/topology/TotalWordCounter.java
@@ -37,7 +37,6 @@ public class TotalWordCounter implements IBasicBolt {
     private BigInteger total = BigInteger.ZERO;
     private static final Logger LOG = LoggerFactory.getLogger(TotalWordCounter.class);
     private static final Random RANDOM = new Random();
-    @SuppressWarnings("rawtypes")
     public void prepare(Map<String, Object> topoConf, TopologyContext context) {
     }
 
--- a/examples/storm-hbase-examples/src/main/java/org/apache/storm/hbase/topology/WordCounter.java
+++ b/examples/storm-hbase-examples/src/main/java/org/apache/storm/hbase/topology/WordCounter.java
@@ -31,7 +31,6 @@ import static org.apache.storm.utils.Utils.tuple;
 public class WordCounter implements IBasicBolt {
 
 
-    @SuppressWarnings("rawtypes")
     public void prepare(Map<String, Object> topoConf, TopologyContext context) {
     }
 
--- a/examples/storm-hbase-examples/src/main/java/org/apache/storm/hbase/topology/WordSpout.java
+++ b/examples/storm-hbase-examples/src/main/java/org/apache/storm/hbase/topology/WordSpout.java
@@ -45,7 +45,6 @@ public class WordSpout implements IRichSpout {
         return this.isDistributed;
     }
 
-    @SuppressWarnings("rawtypes")
     public void open(Map<String, Object> conf, TopologyContext context, SpoutOutputCollector collector) {
         this.collector = collector;
     }
--- a/examples/storm-hdfs-examples/src/main/java/org/apache/storm/hdfs/spout/HdfsSpoutTopology.java
+++ b/examples/storm-hdfs-examples/src/main/java/org/apache/storm/hdfs/spout/HdfsSpoutTopology.java
@@ -132,7 +132,7 @@ public class HdfsSpoutTopology {
     builder.setBolt(BOLT_ID, bolt, 1).shuffleGrouping(SPOUT_ID);
 
     // 4 - submit topology, wait for a few min and terminate it
-    Map clusterConf = Utils.readStormConfig();
+    Map<String, Object> clusterConf = Utils.readStormConfig();
     StormSubmitter.submitTopologyWithProgressBar(topologyName, conf, builder.createTopology());
     Nimbus.Iface client = NimbusClient.getConfiguredClient(clusterConf).getClient();
 
--- a/examples/storm-hive-examples/src/main/java/org/apache/storm/hive/trident/TridentHiveTopology.java
+++ b/examples/storm-hive-examples/src/main/java/org/apache/storm/hive/trident/TridentHiveTopology.java
@@ -177,7 +177,7 @@ public class TridentHiveTopology {
         }
 
         @Override
-        public Map getComponentConfiguration() {
+        public Map<String, Object> getComponentConfiguration() {
             Config conf = new Config();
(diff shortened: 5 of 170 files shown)


knusbaum:
-1 
I don't really see the benefit to this.

d2r:
@knusbaum depending on what @krichter722 is seeking to fix, I could see benefit to cleaning up things like excessive compiler warnings.

@krichter722, what in particular does this fix?

krichter722:
> @krichter722, what in particular does this fix?

Compiler warnings.

revans2:
@knusbaum are you still a -1 on this?

HeartSaVioR:
@knusbaum 
Kindly reminder: are you still a -1 on this?
It may be better to write explicitly that any types can be here so that we can distinguish it is by intention and not missed to write type information.

d2r:
@krichter722, I am willing to disregard @knusbaum 's -1, but this PR needs to be updated as most of these files have been moved within the project.

I checked a few of these, and it seems they have already been modified to avoid the compile errors.

Would you take a look?

knusbaum:
Yes, I'll retracting my -1. Sorry for letting this one languish.

krichter722:
> I checked a few of these, and it seems they have already been modified to avoid the compile errors.

I seems like `IBolt.prepare`'s `topoConf` has been changed to `Map<String, Object>` everywhere which is a good idea (and a better idea than `<?,?>`). Now that the interface is improved in puncto type safety, the generic types should be used everywhere they're encountered/used anyway, but just not passed down or up the hierarchy. This applies to all topology configuration, metadata and authentication maps since they're using `<String, Object>` at some point in the hierarchy or in a relationship.

I pushed those changes to the generics branch. Since this issue is about `IBolt` I can open a new PR, but it's very similar.

I don't seem to be able to figure out the necessary type for `IScheduler.config` which is sometimes `Map<String, Object>` and `Map<String, Map<String, [Number or Double]>` which is possible in a non-type safe environment, but unnecessarily hard to understand. So, I let it untouched.

There're a lot of usages of the idiom

    Map<String, Object> someMap;
    Map<String, Object> nestedMap = (Map<String, Object>) someMap.get("someLiteral");

which should imo be encapsulated in a configuration class since the map should contain nothing more complex than wrappers, maybe a single valued collection (list or set), but a map in a map value might cause problems at some point (and probably only does not now because of the excellent test coverage). Also, `"someLiteral"` is the [magic-number anti-pattern](https://en.wikipedia.org/wiki/Magic_number_(programming)#Unnamed_numerical_constants).

> Yes, I'll retracting my -1. Sorry for letting this one languish.

No problem. I'm aware that some of the changes do not introduce improvement immediately because Storm serializes and deserializes a lot, but the changes allow to discover design flaws and to fail earlier in case of a type mismatch.

knusbaum:
@krichter722 Yes, these maps are vestiges of the old clojure code. It would probably be preferable to do as you suggest, but I'm not sure how much work there is to do there. IIRC, the config objects are pretty well spread and used throughout the code, and there's some other support code, tests, validation, etc. That may have to be significantly changed. I think it's definitely worth consideration, though. Maybe a dev list email would be a good place to get opinions.

krichter722:
> It would probably be preferable to do as you suggest, regarding a configuration class, but I'm not sure how much work there is to do there.

I wouldn't do it for all Maps, just the ones carrying complex types (I'm sticking to my description above). On the other hand, having type safe configuration classes for all Maps increases type safety to a maximum, but that'd be a few hundred person hours (I'm saying that after going through a large portion of the code making extremely trivial changes).

> I think it's definitely worth consideration, though. Maybe a dev list email would be a good place to get opinions.

When I find more time, sure. Are you OK with merging this PR? It's a contained unit of work and essentially a pareto optimization that cannot cause any runtime exception that would not already occur and there're no compile time issues.

external/storm-kafka/pom.xml
```diff
@@ -57,7 +57,7 @@
                 <artifactId>maven-checkstyle-plugin</artifactId>
                 <!--Note - the version would be inherited-->
                 <configuration>
-                    <maxAllowedViolations>557</maxAllowedViolations>
+                    <maxAllowedViolations>561</maxAllowedViolations>
```
d2r: @krichter722 curious: what caused us to increase this?

krichter722: Right, I should have mentioned this on my own. The usage of checkstyle is not clear to me since it enforces extremely strict rules like max. 80 characters line length whereas the code still contains easy to fix issues like variable names. Therefore I assumed that checkstyle has been added without fixing the large amount of complaints and the value 557 has simply been set to the number of complaints. I figured out that it's very hard to diff the violations with and without this PR, however I'm pretty sure that I only added 4 lines longer than 80 characters. I can check again if the 4 more issues besides the existing 557 are an absolute showstopper to merge this PR.

krichter722:
> Is there a Jira issue associated with this pull request? If not, we can make one.

I found https://issues.apache.org/jira/browse/STORM-1048 in issues reported by me which I forgot until now. It either has been fixed before this PR or is definitely fixed with this PR.

external/storm-kafka/pom.xml
knusbaum: What we did with checkstyle, for better or worse, was agree on a set of rules and then apply those rules. We didn't go through and fix every checkstyle violation when we did that. However, we agreed that any pull requests should not increase the number of alerts, and preferably decrease them. It would be better if you could reformat any new 80+character lines to please the checkstyle monster.

knusbaum:
@krichter722 I would either rename that Jira and update the description a bit, or open a new one. Once you've done that, add the jira number into the PR title - look at other Storm PRs for examples.

After that and the checkstyle stuff, I don't see anything else that should prevent this from going in.
:+1:

krichter722:
> I would either rename that Jira and update the description a bit, or open a new one. Once you've done that, add the jira number into the PR title - look at other Storm PRs for examples.
>
> After that and the checkstyle stuff, I don't see anything else that should prevent this from going in.

@knusbaum Both done.

krichter722:
I just realized that this PR makes a lot of `@SuppressWarning("rawtypes")` unnecessary. What do you think about adding that as a second commit to this PR?

d2r:
> I just realized that this PR makes a lot of @SuppressWarning("rawtypes") unnecessary. What do you think about adding that as a second commit to this PR?

@krichter722 Sorry for the delay; yes we can hold off if you want to clean up these unnecessary annotations.

krichter722:
> we can hold off if you want to clean up these unnecessary annotations.

Done.

knusbaum:
+1

roshannaik:
@krichter722  thanks for this!

d2r:
New changes are to remove warning suppression attributes and look good to me. +1 again.

Thanks @krichter722 for the changes and for your patience. I have merged this to master.

PR #461, opened 2015-03-10
[STORM-625] Don't leak netty clients when worker moves or reuse netty client.

kishorvpatil:
- Adding tracking of netty connection by key and reuse whenever client is available.
- Fix included original changes from #455.


revans2:
+1, but since some of the code came from me another committer is going to need to give a +1 too.

knusbaum:
+1

knusbaum:
+1

PR #462, opened 2015-03-10
[STORM-703] With hash key option for RedisMapState, only get values for keys in batch

dkatten:
This commit fixes a bug whereby if the state updater is constructed with a hash key (ie, the state will be stored as a key in a redis hash, versus as a key in the top-level redis space), each call to multiGet would request the entire hash and iterate to extract only the values in the hash relevant to the batch.

This can cause an inordinate amount of network traffic (and actually caused our interfaces to fall over) for states with either a moderately high cardinality or large values. Instead, the call to Redis should be an hmget (hash multiget) that takes the hash key as its first argument and an array of strings as the keys to fetch from that key, thereby retrieving only the requested values.

The change also deprecates and removes buildValuesFromMap.

### DIFF
--- a/external/storm-redis/src/main/java/org/apache/storm/redis/trident/state/RedisMapState.java
+++ b/external/storm-redis/src/main/java/org/apache/storm/redis/trident/state/RedisMapState.java
@@ -223,8 +223,10 @@ public class RedisMapState<T> implements IBackingMap<T> {
         if (keys.size() == 0) {
             return Collections.emptyList();
         }
+
+        String[] stringKeys = buildKeys(keys);
+
         if (Strings.isNullOrEmpty(this.options.hkey)) {
-            String[] stringKeys = buildKeys(keys);
             Jedis jedis = null;
             try {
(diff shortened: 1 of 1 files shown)


HeartSaVioR:
@dkatten I've just missed it. Thanks! +1
@revans2 Could you take a look, please?

revans2:
The change looks good to me. I am +1 for the change.  @dkatten if you want to file a JIRA under STORM at issues.apache.org that would be great.  If you would rather I can file it for you.

dkatten:
I'll file it. Thanks!

dkatten:
STORM-703
https://issues.apache.org/jira/browse/STORM-703

HeartSaVioR:
@revans2 It's filed to JIRA so seems like we can merge it.

revans2:
Sorry I really don't follow all the JIRa updates as closely as I should.  @HeartSaVioR thanks for kicking me :) I'll merge it in.

PR #467, opened 2015-03-18
STORM-711: modifying all connectors to use collector.reportError instead...

Parth-Brahmbhatt:
... of logging and use tuple anchoring when emitting a tuple.


harshach:
+1

PR #468, opened 2015-03-18
[STORM-712] Storm daemons shutdown if OutOfMemoryError occurs in any thread 

kishorvpatil:
- Creating `ExtendedThreadPoolExecutor` to pass uncaughtExceptions as part of `afterExecute`
- All daemon main threads set `uncaughtExceptionHandler`.

### DIFF
--- a/storm-core/src/jvm/backtype/storm/drpc/DRPCSpout.java
+++ b/storm-core/src/jvm/backtype/storm/drpc/DRPCSpout.java
@@ -28,6 +28,7 @@ import backtype.storm.topology.OutputFieldsDeclarer;
 import backtype.storm.topology.base.BaseRichSpout;
 import backtype.storm.tuple.Fields;
 import backtype.storm.tuple.Values;
+import backtype.storm.utils.ExtendedThreadPoolExecutor;
 import backtype.storm.utils.ServiceRegistry;
 import backtype.storm.utils.Utils;
 import java.util.ArrayList;
@@ -40,6 +41,10 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
 import java.util.concurrent.Callable;
+import java.util.concurrent.SynchronousQueue;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.apache.thrift.TException;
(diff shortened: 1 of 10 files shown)


lazyval:
And you sure it's not already happening?

kishorvpatil:
@lazyval, sorry for delay in responding. We have seen this happen in case some of our clusters. When the `ExecutorService`, executes the `FutureTask`, it  captures even `Error` and not processed by any thread's `DefaultUnCaughtExceptionHandler`. This can lead to server not shutdown even after OOM.

revans2:
+1 the code looks good to me.   I had to do something similar for hadoop.

@lazyval do you have any more questions or concerns about this?

storm-core/src/jvm/backtype/storm/utils/Utils.java
```diff
@@ -533,4 +533,16 @@ private static SerializationDelegate getSerializationDelegate(Map stormConf) {
...
+    public static void handleUncaughtException(Throwable t) {
+      if(t!= null && t instanceof OutOfMemoryError) {
+         try {
+             System.err.println("Halting due to Out Of Memory Error..." + Thread.currentThread().getName());
+         } catch (Throwable err) {
+             //Again we done want to exit because of logging issues.
+         }
+         Runtime.getRuntime().halt(-1);
+      }
+    }
```
lazyval: Why log only OOMs? Am I right that all other exceptions/errors will be silently swallowed?

storm-core/src/jvm/backtype/storm/drpc/DRPCSpout.java
```diff
@@ -129,7 +133,9 @@ private void checkFutures() {
     public void open(Map conf, TopologyContext context, SpoutOutputCollector collector) {
         _collector = collector;
         if(_local_drpc_id==null) {
-            _backround = Executors.newCachedThreadPool();
+            _backround = new ExtendedThreadPoolExecutor(0, Integer.MAX_VALUE,
+                60L, TimeUnit.SECONDS,
+                new SynchronousQueue<Runnable>());
```
lazyval: You swapped out thread pool that caches threads with implementation that does proper exception handling, are you sure there won't be any performance degradations?

storm-core/src/jvm/backtype/storm/utils/Utils.java
longdafeng: Please don't use "System.err.println", it is likely to lose all System.out/err message when start storm like "nohup storm nimbus >/dev/null 2>&1 &"
it is better use log interface.

storm-core/src/jvm/backtype/storm/drpc/DRPCSpout.java
kishorvpatil: @lazyval the `newCachedThreadPool` uses these very arguments to generate `ThreadPoolExcecutor` for cached threads. So it won't have any additional performance overhead.

storm-core/src/jvm/backtype/storm/utils/Utils.java
kishorvpatil: @lazyval, using log interface could rethrow OOM which won't be handled properly. Hence using `System.err.println`

kishorvpatil:
@lazyval Thanks for pointing out. I have updated the PR to rethrow non OOM Errors. Also, main server thread should exit on Error.

lazyval:
:+1:

revans2:
+1

PR #470, opened 2015-03-19
STORM-713: Include topic information with Kafka metrics

chawco:
Currently, Kafka metrics are emitted with partition information, but no information about which topic that partition belongs to. This patch includes the topic information in the Storm metrics emitted by the Kafka spout so that multiple input topics may be tracked from a single topology.


harshach:
@chawco patch looks good to me but can you open a apache jira for it. Thanks.

chawco:
Sure thing, I've filed [STORM-713 ](https://issues.apache.org/jira/browse/STORM-713) for this issue.

Parth-Brahmbhatt:
You should update the commit message so it begins with JIRA number. +1.

chawco:
I've amended the offending commit. Thanks!

harshach:
+1

PR #471, opened 2015-03-20
UserGroupInformation should use the same config with UserProvider

onlychoice:
As the title,  if UserGroupInformation is not set with the hbaseConfig, UserGroupInformation with initialize conf with new Configuration(), and this is not the conf used for UserProvider, and this will wrong behavior: I have set hadoop.security.authentication to kerberos in hbase-site.xml, but without this fix, UserGroupInformation.isSecurityEnabled() will always return false.

With UserGroupInformation.setConfiguration you need krb5 configuration, either default value or through System.setProperty("java.security.krb5.conf", krbPath);

### DIFF
--- a/external/storm-hbase/src/main/java/org/apache/storm/hbase/security/HBaseSecurityUtil.java
+++ b/external/storm-hbase/src/main/java/org/apache/storm/hbase/security/HBaseSecurityUtil.java
@@ -46,6 +46,7 @@ public class HBaseSecurityUtil {
         if (conf.get(TOPOLOGY_AUTO_CREDENTIALS) == null ||
                 !(((List) conf.get(TOPOLOGY_AUTO_CREDENTIALS)).contains(AutoHBase.class.getName()))) {
             LOG.info("Logging in using keytab as AutoHBase is not specified for " + TOPOLOGY_AUTO_CREDENTIALS);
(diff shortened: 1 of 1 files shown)


harshach:
@onlychoice you need to package hbase-site.xml along with the topology jar. Did you try doing that.

d2r:
@onlychoice , did you try what @harshach  suggests?

If we still need a change here, would you create a Jira issue for this work?

onlychoice:
@d2r I have packaged hbase-site.xml in the topology jar in the first place, this is not the reason.

I'd like to do that, but where can I create the jira issue?

revans2:
http://issues.apache.org/jira file it under STORM.

revans2:
Oh and it is not hbase-site.xml that you need to configure.  It is hadoop-site.xml.  UserGroupInformation is a Hadoop class that hbase reuses, but it has no knowledge of hbase-* configs.  It only looks for hadoop configs.

revans2:
@onlychoice I just looked at the documentation and I cannot find it anywhere for setting hadoop-site.xml vs hbase-site.xml etc.  If the work around is good for you could you at least file a JIRA for us to update the documentation to explain how to access secure hbase correctly.

knusbaum:
Closing this. Reopen if it's still an issue.

PR #475, opened 2015-03-23
STORM-587. trident transactional state in zk should be namespaced with topology id.

harshach:


Parth-Brahmbhatt:
+1.

harshach:
@ptgoetz @revans2 can you please take a look at this PR.

storm-core/src/jvm/backtype/storm/transactional/state/TransactionalState.java
```diff
@@ -59,7 +59,8 @@ protected TransactionalState(Map conf, String id, Map componentConf, String subr
                               .get(Config.TOPOLOGY_KRYO_REGISTER));
             }
             String transactionalRoot = (String)conf.get(Config.TRANSACTIONAL_ZOOKEEPER_ROOT);
-            String rootDir = transactionalRoot + "/" + id + "/" + subroot;
+            String topologyName = (String)conf.get(Config.TOPOLOGY_NAME);
```
caofangkun: What if topologyName is null?
Should we give these configs default values？ 
Something like:
```
String topologyName = Utils.parseString(conf.get(Config.TOPOLOGY_NAME), "default-topology-name");
```

revans2:
-1, I know it is convoluted the path through the code for this, but ID is already the topology name.  I don't see a reason to break backwards compatibility to have the topology name in there twice.

Please don't do topology id either, because then state would never be saved from one run of a trident topology to another.  The best we could do is to isolate it by user name.

revans2:
Sorry I am wrong, I need to look at it more. It is convoluted the path through the code, which is why I need to look into this more.  Right now it appears to be the stream id, which seems incorrect to me, but I'll read through it.

revans2:
This is the path I saw through the code.  Please let me know if I missed anything.
https://github.com/apache/storm/blob/master/storm-core/src/jvm/storm/trident/topology/state/TransactionalState.java#L53
https://github.com/apache/storm/blob/master/storm-core/src/jvm/storm/trident/spout/TridentSpoutCoordinator.java#L56
https://github.com/apache/storm/blob/master/storm-core/src/jvm/storm/trident/spout/TridentSpoutCoordinator.java#L48-51
https://github.com/apache/storm/blob/master/storm-core/src/jvm/storm/trident/topology/TridentTopologyBuilder.java#L160
https://github.com/apache/storm/blob/master/storm-core/src/jvm/storm/trident/topology/TridentTopologyBuilder.java#L278-282
https://github.com/apache/storm/blob/master/storm-core/src/jvm/storm/trident/topology/TridentTopologyBuilder.java#L71-76
https://github.com/apache/storm/blob/master/storm-core/src/jvm/storm/trident/TridentTopology.java#L403
https://github.com/apache/storm/blob/master/storm-core/src/jvm/storm/trident/planner/SpoutNode.java#L33-35
https://github.com/apache/storm/blob/master/storm-core/src/jvm/storm/trident/TridentTopology.java#L119-122

I am a bit nervous about breaking backwards compatibility without a migration path.  I would be OK with the change if we had a tool that could upgrade/downgrade a trident topology state.  I suppose it would be OK to also have an option to not include the topology name, but I would prefer to not have to deal with that from a state perspective.

revans2:
@harshach is this still something that you want?  If not please close the pull request if so I am happy to talk about a migration plan from one state layout to another.

ptgoetz:
Unless others feel strongly about this change, I think it's okay to close this.

PR #476, opened 2015-03-23
STORM-720: Storm.cmd should return ERRORLEVEL before exiting main block

rtandonmsft:
PR for https://issues.apache.org/jira/browse/STORM-720

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -296,6 +296,7 @@
  * STORM-2379: ElasticSearch connector changes to use new APIs.
 
 ## 1.2.0
+ * STORM-2640: Deprecate KafkaConsumer.subscribe API option, make KafkaConsumer.assign the default
  * STORM-2650: Expand property substitution test for Flux to cover non-string properties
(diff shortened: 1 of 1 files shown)


harshach:
+1

bin/storm.cmd
```diff
@@ -90,6 +90,7 @@
   )
   set path=%PATH%;%STORM_BIN_DIR%;%STORM_SBIN_DIR%
   call start /b "%storm-command%" "%JAVA%" %JAVA_HEAP_MAX% %STORM_OPTS% %STORM_LOG_FILE% %CLASS% %storm-command-arguments%
+  exit /b %ERRORLEVEL%
   goto :eof
```
knusbaum: We can get rid of this `goto :eof` now, right?

rtandonmsft: @knusbaum Yes, I do not see any control flow issue of removing or keeping it with this change. I did not do it to keep the sub routine blocks consistent so that all end with goto :eof, plus this is an already deployed and working change in Apache Storm on Azure HDInsight for more than 6 months.

harshach:
@knusbaum are we good to merge this?

PR #477, opened 2015-03-24
[STORM-468] java.io.NotSerializableException should be explained

jangie:
Am pretty sure this is not the best way. Am open to suggestions.


docs/documentation/FAQ.md
```diff
@@ -62,6 +62,10 @@ You can join streams with join, merge or multiReduce.
 
 At time of writing, you can't emit to multiple output streams from Trident -- see [STORM-68](https://issues.apache.org/jira/browse/STORM-68)
 
+### Why am I getting a NotSerializableException/IllegalStateException when my topology is being started up?
+
+Within the Storm lifecycle, the topology is instantiated and then serialized to byte format to be stored in ZooKeeper, prior to the topology being executed. Within this step, if a spout or bolt within the topology has an initialized unserializable property, serialization will fail. If there is a need for a field that is unserializable, initialize it within the bolt or spout's prepare method, which is run after the topology is delivered to the worker.
```
HeartSaVioR: I'm not sure but it may be better to document "serialized fields' values are reused when workers are relaunched whether live worker modifies it or not.". 
I had this issue earlier, and sure it was by mistake, but it would be hard to find bug.

Wish to hear everyone's thought.

ptgoetz:
Ping (mostly to other committers to solicit review/discussion).

This is not a bad idea.

revans2:
+1 for the concept.  The formatting of all of the code is off.  Looks like they are using tabs, not spaces.  I would also like to see @HeartSaVioR's comment addressed.

There also appear to be merge conflicts now.  @jangie sorry this has lingered so very long.  This is a great concept.  Is there any hope that you can address these comments?

jangie:
@revans2 Sure, I'll try and take a look later tonight for a fixup for the conflicts and spacing preference, thanks!

revans2:
@jangie great. I want to apologize too 10 months is way too long for a pull request to linger.  We really need to do better with some of these.

dossett:
Would it be cleaner to just catch NotSerializableException directly and rethrow an IllegalStateException?

jangie:
@dossett At initial time of the PR, I believe that the NotSerializableException was wrapped inside of a RuntimeException (and may very well still be).

For what it's worth, I believe I've addressed the other comments, I guess we'll see what Travis says.

dossett:
@jangie Yes, my comment was based a sloppy reading of your PR.  Nice work!

PR #478, opened 2015-03-24
STORM-553:add eclipse java formatter file for Apache Storm Java Code

caofangkun:


longdafeng:
why "indentation.size" and "tabulation.size" are 2, normally they are 4.

caofangkun:
@longdafeng 
Different tools may define different tab size,the tab size may equivalent to 2 or 4 or 8 spaces/characters.
Defined tab size to 2 spaces, will be much easier to convert Tab(4) to 4 spaces and Tab(8) to 8 spaces.

ptgoetz:
I'm all for adopting a code style guide for all languages (Java, Clojure, Python, etc.), but I'd rather see a style specification documented first, then follow up with templates that implement it.

I do feel that we have a lot of code style technical debt in both Java/Clojure. If we're moving away from clojure, we don't need to worry about style too much there, but we should establish a style guide for Java source, and start cleaning up compiler warnings, PMD/Checkstyle/Findbugs issues, etc.

I would suggest a discussion on the dev@ list, and hope it doesn't become too religious. ;)

revans2:
+1 for what @ptgoetz said.

d2r:
+1 @ptgoetz , and I thought there was already a style for Clojure that we were supposed to use. I think even after moving mostly to Java, we will still have Clojure left.  May as well include it.

ptgoetz:
@d2r Good point. I know there was some work done around moving to idiomatic clojure and making style consistent, but I don't think we ever put together a style guide. So our clojure code is relatively consistent in terms of style, but our java code is kind of all over the map.

I'd be in favor for establishing style guides for any languages we use.

knusbaum:
I'm going to close this. Please feel free to reopen or create a new PR once we have a style guide in place.

PR #482, opened 2015-03-25
STORM-723 Remove RedisStateSetUpdater / RedisStateSetCountQuerier which didn't tested and have a bug

HeartSaVioR:
@dashengju and me decided to remove RedisStateSetUpdater / RedisStateSetCountQuerier because it didn't documented, didn't have any test cases, and have bugs.

We can introduce counter / various data types later.


dashengju:
+1

revans2:
I am +1 on the change, but my only question is do we want to do this only for all branches? or only for 0.10.X?

HeartSaVioR:
I think we should remove these 2 files from all branches, since we'll introduce new classes if we need.

revans2:
OK, I'll merge this in.

HeartSaVioR:
@revans2 Thanks for merging! Especially I know that cherry-picking commits is not a easy way to do. ;)

revans2:
@HeartSaVioR yes, but this is a very easy cherry-pick, when we are depleting code. I am running the tests on 0.10.x-branch now to be sure everything looks good, and then I will merge it in and resolve the JIRA.

PR #483, opened 2015-03-26
STORM-727: nimbus_auth_test should not use default nimbus port.

Parth-Brahmbhatt:


d2r:
It would be even better to use an ephemeral port, as with util/available-port.

Parth-Brahmbhatt:
@d2r done.

harshach:
+1

d2r:
+1 Ran all tests on OSX: JDK 8

PR #484, opened 2015-03-26
[STORM-728] Put emitted and transferred stats under correct columns

d2r:
The column headers were consistent with other UI tables, so I swapped the data.


harshach:
+1

revans2:
+1

PR #485, opened 2015-03-26
[STORM-729] Include Executors (Window Hint) if the component is of Bolt type

d2r:
This was missing only from the Bolt case.


harshach:
+1

revans2:
+1

PR #486, opened 2015-03-26
STORM-704 Apply Travis CI to Apache Storm Project

HeartSaVioR:
* support build matrix (openjdk6, openjdk7, oraclejdk7, oraclejdk8)
* change log level to WARN while running unit tests
* print out clojure test reports (only errors)

Sure we should ask Apache Infra to set up Travis CI (with asfgit account), which is not an issue cause many projects on Apache already uses Travis CI.
ex. https://issues.apache.org/jira/browse/INFRA-6161


HeartSaVioR:
Applied first on my forked repo: https://travis-ci.org/HeartSaVioR/storm/jobs/56021892

Could any committers help requesting set up Travis CI to Apache Infra?

HeartSaVioR:
Failing build (from storm-core) output for example : https://s3.amazonaws.com/archive.travis-ci.org/jobs/56016756/log.txt

HeartSaVioR:
@revans2 
Could you take a look? 
We were talking about adopting Travis CI and I feel you would like this, and there're no hurdle left (except Apache Infra)

revans2:
@HeartSaVioR Sorry it has taken me so long to take a look.  It appears that for https://travis-ci.org/HeartSaVioR/storm/jobs/56016756 you are missing one of the shell bolt/spout dependencies.  We currently need ruby, python, and nodejs installed.  The errors about 
```39877 [Thread-128] ERROR backtype.storm.task.ShellBolt - Halting process: ShellBolt died.
java.lang.RuntimeException: backtype.storm.multilang.NoOutputException: Pipe to subprocess seems to be broken! No output read.```

indicate that one of these is not installed/on the path.

File: dev-tools/travis/travis-build.sh [hunk 1]
```diff
@@ -0,0 +1,35 @@
...
+TRAVIS_SCRIPT_DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
+
+cd ${STORM_SRC_ROOT_DIR}
+
+# Travis CI doesn't allow stdout bigger than 4M, so we have to reduce log while running tests
+export LOG_LEVEL=WARN
+# We should concern that Travis CI could be very slow cause it uses VM
+export export STORM_TEST_TIMEOUT_MS=100000
+
+mvn clean test
```
revans2: After some new changes went in around packaging shell dependencies this need to be changed to 
```mvn clean package```
or maven gets angry about unpacking something that was not already packaged

File: dev-tools/travis/travis-build.sh [hunk 2]
```diff
@@ -0,0 +1,35 @@
...
+STORM_SRC_ROOT_DIR=$1
+
+TRAVIS_SCRIPT_DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
+
+cd ${STORM_SRC_ROOT_DIR}
+
+# Travis CI doesn't allow stdout bigger than 4M, so we have to reduce log while running tests
+export LOG_LEVEL=WARN
+# We should concern that Travis CI could be very slow cause it uses VM
+export export STORM_TEST_TIMEOUT_MS=100000
```
knusbaum: Double export?

dev-tools/travis/travis-build.sh
revans2: Actually I was wrong.  travis-ci runs mvn clean install -DskipTests before building, so we probably just want to run ```mvn test``` and skip the clean.

revans2:
I cloned your changes and have been playing around with them.  For me the test all fail before reaching the clojure test.

https://travis-ci.org/revans2/incubator-storm/jobs/59781670

I think @kishorvpatil is looking into the test failures.  I don't see them on my box, so I think it has something to do with how fast the box is.

dev-tools/travis/travis-build.sh
HeartSaVioR: Yes, it should be changed for now.
I don't want to lean on Travis CI's behavior, but we would love to let test run as fast as possible, so just run ```mvn test``` with describing would be great.

dev-tools/travis/travis-build.sh
HeartSaVioR: Thanks for finding! I'll fix it.

.travis.yml
```diff
@@ -0,0 +1,7 @@
+language: java
+jdk:
+  - openjdk6
+  - openjdk7
```
revans2: Both openjdk6 and openjdk7 seem to be failing all the time.

https://travis-ci.org/revans2/incubator-storm/jobs/59783142

This looks like a bug in openjdk and glibc combining together.  We may not want to test with them if it is not something that we cannot fix.

HeartSaVioR:
@revans2 
About multilang, I missed due to not failing tests. We may like to make multilang tests more reliable.
I'll find a reason about what dependencies are missing, and find a way to install such things.
About test failures, we can control STORM_TEST_TIMEOUT_MS to respect slow box, but maybe some tests are still needs fast (maybe not very slow) machine.

.travis.yml
HeartSaVioR: We can start with removing two things now and make new branch which targets two JDKs sometimes.
Travis CI will test all PRs and branches, so it would be fine.

HeartSaVioR:
Mailed to Travis CI support to how can I get python, ruby, nodejs installed.
If we really need to install these manually, I'll lend a debug VM from Travis CI.

HeartSaVioR:
Last build is successful except multilang tests.
https://travis-ci.org/HeartSaVioR/storm/builds/59788403

revans2:
Yes, so the issue that we have run into before, and may be the same now, is that ruby does not come with JSON support by default on ubuntu.  But that was a much older version or ruby and ubuntu.  So who knows.

HeartSaVioR:
https://travis-ci.org/HeartSaVioR/storm/builds/59794019 succeed and https://travis-ci.org/HeartSaVioR/storm/builds/59795636 failed, which means failed tests could be a kind of random failures or relied on machine power.

First one is DisruptorQueueTest, as you already know.

Failed tests:   testMessageDisorder(backtype.storm.utils.DisruptorQueueTest): We expect to receive first published message first, but received null expected:<1> but was:<null>

Second one is Kafka Tests, which I met intermittently from my dev. machine.
 
  testKeyValue(storm.kafka.TridentKafkaTest): Could not send message with key = key-123 to topic = test
  generateTuplesWithKeyAndKeyValueScheme(storm.kafka.KafkaUtilsTest): Error fetching data from [Partition{host=localhost:52047, partition=0}] for topic [testTopic]: [OFFSET_OUT_OF_RANGE]

HeartSaVioR:
Infra setup seems to be completed. I can see current PR's build status.

revans2:
@HeartSaVioR great to hear that the setup is complete.  I'm not sure if we want to merge this in as is, knowing that the majority of test cases will fail, or if we want to wait for the tests to pass more regularly.

HeartSaVioR:
@revans2 
I'm trying to resolve multilang tests, so I'll let you know.
Regarding random failure, IMO we already addressed these to JIRA issue, and it's better to expose instead of hiding so we all know tests issue and will try to resolve.

revans2:
Can we try and do something like 

```
mvn test -fae
BUILD_RET_VAL=$?
```

or even better not fail ever, but check the output files for failures. I tend to use the following but you could update your script to do something similar.

```
mvn test -Pnative -fn
echo "Errors or Failures:"
egrep -il '<fail|<error' ../*/target/test-reports/*.xml | xargs -I '{}' bash -c "echo -n '{}':' '; egrep -ic '<fail|<error' '{}'"
egrep -iq '<fail|<error' ../*/target/test-reports/*.xml 
BUILD_RET_VAL=$? 
```

It doesn't print out the full error stack traces, but it does capture any failures

HeartSaVioR:
@revans2 
Thanks! I'll try it.

HeartSaVioR:
mvn with -fae makes sense since retrying tests may require at least collaborator to github repository so even committers are not authorized to retry tests.

HeartSaVioR:
Seems like multilang tests cannot verify it runs properly after initialized. 
NoOutputException has raised at Travis tests all the time but test-reports doesn't have fail or error.

https://travis-ci.org/HeartSaVioR/storm/jobs/59953593

HeartSaVioR:
If we have file "a" containing "abc"...

> egrep -il 'abc' a | xargs -I '{}' bash -c "echo -n '{}':' '; egrep -ic 'abc' '{}'"
a: 1
> echo $?
0
> egrep -il 'fg' a | xargs -I '{}' bash -c "echo -n '{}':' '; egrep -ic 'fg' '{}'"
> echo $?
0

It could capture count of string matching patterns in files, but it shouldn't be used to retval.

Note: I've tested it with OSX, with zsh.

HeartSaVioR:
Btw, I modified .travis.yml to install recent version of ruby and node, and still have no luck with multilang tests. It doesn't print out any helpful outputs, but I'll keep giving a try.

HeartSaVioR:
I'm about to be sure that we're better to introduce Travis CI now though some digging points still exist.
Cause it just catches storm-eventhubs build issue (https://issues.apache.org/jira/browse/STORM-800) automatically.
https://travis-ci.org/apache/storm/builds/59957899

HeartSaVioR:
Please note that ```mvn with -fae``` still ban external modules to test when storm-core failed.

revans2:
OK I am +1 for merging this in.  We can adjust things as with error reporting and logging as time goes on.

But I would like others opinions on this too.  @d2r @knusbaum @ptgoetz or any others do you have an opinion on merging this in?

knusbaum:
I'm +1 for merging. I'd rather have it in so it can yell at us about fixing the broken tests.

PR #490, opened 2015-03-29
STORM-731. Support compiling with HBase-1.0.0.

harshach:


ptgoetz:
+1, though it's not necessary because you are a committer. If you can apply the patch, even if you have to do so manually, without violating the spirit  of it (yes, I know, that's subjective...), then you can do so.

It would be good form to notify dev@ of any actions you had to take, etc. to make the patch apply.

If there are any problems, it can always be reverted.

PR #494, opened 2015-03-31
stormjar.jar on classpath first in supervisor workers

emidln:
If stormjar.jar is first on the classpath for the supervisor workers,
libraries AOT-compiled and provided by Storm for Storm UI will take
a backseat to libraries the user provides.

### DIFF
--- a/storm-core/src/clj/backtype/storm/daemon/supervisor.clj
+++ b/storm-core/src/clj/backtype/storm/daemon/supervisor.clj
@@ -669,9 +669,9 @@
           topo-classpath (if-let [cp (storm-conf TOPOLOGY-CLASSPATH)]
                            [cp]
                            [])
-          classpath (-> (current-classpath)
-                        (add-to-classpath [stormjar])
-                        (add-to-classpath topo-classpath))
+          classpath (-> stormjar 
+                        (add-to-classpath [(current-classpath)])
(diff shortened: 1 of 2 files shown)


revans2:
The problem here is one of classpath hell. The reason someone wants their jar first in the classpath is because there is a conflict between something that storm provides and something that the worker wants.  In some cases the developers of the dependency in question have maintained binary backwards compatibility and this will work fine, but in too many cases, i.e. guava, compatibility is not maintained. If the incompatibility results in a method not found error or something like that it is simple to debug, but I have seen issues, like with disruptor, where the errors are a lot more subtle, and everything mostly works.  Debugging those over a mailing list when I don't really know what dependencies storm is running against is a real pain. 

I am fine with having the option to make the user jar first on the classpath, but I would want it to be off by default.  @emidln if you want to do that please file a JIRA for it at https://issues.apache.org/jira under STORM and then update the title of the pull request to include the JIRA number, i.e. `STORM-XXXX: optionally let stormjar.jar fist on classpath`

d2r:
@emidln , it has been a while since the last comment.  Would you create a Jira issue for this work, address the comment by @revans2 , and resolve conflicts with the master branch?

knusbaum:
Closing this.

PR #120, opened 2014-05-21
STORM-188: Allow user to specifiy full configuration path when running storm command

clockfly:
We can use system config "storm.conf.file" to specify a custom config file.

Before this patch, we will only try to lookup this file on jvm classpath.

This patch will give us the ability to specify full/relative filesystem path as configuration file path, thus will give us more flexibility, while we still are able to look it up on jvm classpath.

Now we are able to submit storm topology like this: 

```bash
storm jar job.jar --config /tmp/dynamic-configuration-path.yaml
```

We don't need to add /tmp into our classpath(also inappropriate).

PR #495, opened 2015-03-31
STORM-188. Allow user to specifiy full configuration path when running storm command

harshach:


harshach:
This PR is taken from https://github.com/apache/storm/pull/120#issuecomment-69435220

@clockfly @revans2 can you please take a look.

clockfly:
+1

storm-core/src/jvm/backtype/storm/utils/Utils.java
```diff
@@ -135,35 +137,68 @@ public static void sleep(long millis) {
...
+            throw new IOException(
+                    "Could not find config file, name not specified");
+        }
+
+        HashSet<URL> resources = new HashSet<URL>(findResources(configFilePath));
+        if (resources.isEmpty()) {
+            File configFile = new File(configFilePath);
+            if (configFile.exists()) {
+                return new FileInputStream(configFile);
+            }
```
lazyval: Why not throw IOException, when path contains no actual file?

clockfly: @lazyval 

mustExist flag will control whether we should fail the method.
···public static Map findAndReadConfigFile(String name, boolean mustExist) ···

When we cannot find a resource, we will return null; then we check whether mustExist flag is enabled to determine whether we should throw RuntimeException.

storm-core/src/jvm/backtype/storm/utils/Utils.java
```diff
@@ -135,35 +137,68 @@ public static void sleep(long millis) {
     }
 
     public static Map findAndReadConfigFile(String name, boolean mustExist) {
+        InputStream in = null;
+        Boolean confFileEmpty = false;
```
lazyval: no need in boxed value

storm-core/src/jvm/backtype/storm/utils/Utils.java
lazyval: Yeah, I just a bit concerned that flow is driven by both exceptions and flag/null.

storm-core/src/jvm/backtype/storm/utils/Utils.java
harshach: @lazyval  what you mean by boxed value?

lazyval: There are two variables types in Java: primitive (like boolean, int, long, short and so on), and their boxed counterparts (Boolean, Integer, Long, ...), which are seen as reference types. It's minor, but unless you're working with collections, there is no need to use later one.

harshach: @lazyval  I know about primitives and objects in java . Yes makes sense to use a boolean.

harshach:
@clockfly @lazyval updated the patch for fixing the last comment.

storm-core/src/jvm/backtype/storm/utils/Utils.java
harshach: @lazyval there are cases where storm wants to use defaults.yaml if there is not storm.yaml. If one wants to do a quick test they can just run the storm cluster by not having any storm.yaml . Hence the reason we have mustexist as a flag to indicate if we want to throw an error or not.

lazyval:
:+1:

revans2:
I think overall it is good and I am +1.

PR #497, opened 2015-03-31
STORM-742 Let ShellBolt treat all messages to update heartbeat

HeartSaVioR:
It's for avoiding design constraint being introduced from STORM-513.
ShellSpout is already applied.

### DIFF
--- a/storm-core/src/jvm/backtype/storm/task/ShellBolt.java
+++ b/storm-core/src/jvm/backtype/storm/task/ShellBolt.java
@@ -321,9 +321,11 @@ public class ShellBolt implements IBolt {
                     if (command == null) {
                         throw new IllegalArgumentException("Command not found in bolt message: " + shellMsg);
                     }
-                    if (command.equals("sync")) {
-                        setHeartbeat();
-                    } else if(command.equals("ack")) {
+
+                    setHeartbeat();
+
+                    // We don't need to take care of sync, cause we're always updating heartbeat
+                    if(command.equals("ack")) {
                         handleAck(shellMsg.getId());
                     } else if (command.equals("fail")) {
                         handleFail(shellMsg.getId());


HeartSaVioR:
@dashengju 
It could be hard to make unit test, so could you apply this patch to your situation and report it works?
Thanks in advance!

dashengju:
thanks for submit this patch quickly.
we will test in our situation today.

But i have another consideration: does this increase too many heartbeat msg to subprocess?  every emit from subprocess will put a heartbeat tuple.

dashengju:
I checked the code again, my consideration will not happen, it will send heartbeat tuple every second. please ignore.

HeartSaVioR:
Yes, I've not modified about sending heartbeat tuple, so it doesn't make any issues. :)

HeartSaVioR:
Thinking it once more, it cannot resolve @dashengju 's scenario from STORM-738 since users can skip acking with NO-ACK mode.
I'll find alternative way to resolve it.

But PR can resolve ACK mode with too busy bolt, so this PR is still valid.

HeartSaVioR:
@dashengju 
I filed new issue regarding very busy bolt with Non-ACK mode. https://issues.apache.org/jira/browse/STORM-758
Maybe you would be interested that instead of this, so please have a look at my idea in new issue.
Thanks!

dashengju:
@HeartSaVioR  

We have tested the patch. It solved most of the problem, but it still exist in some cases.

Currently, even with ACK,  the subprocess handle one tuple, emit the result, and read all the tuples from stdin to get the id, read all the tuples from stdin needs long time, which cause exceed timeout. 

So we limit the queue-size in ShellBolt to 100, just make the read all the tuples from stdin operation a short time can avoid the problem.

HeartSaVioR:
@ptgoetz @revans2 
I'd like you to take a look. Thanks!

HeartSaVioR:
@dashengju Could you share your patch (limiting queue size) if you don't mind?

HeartSaVioR:
Now I changed my mind to stick with current design constraint, this PR should be merged to solve heartbeat issue in specific scenarios.

Please take a look and comment. Thanks!

HeartSaVioR:
@ptgoetz @revans2 @harshach 
Multi-lang users need this PR to overcome heartbeat design constraint. 
Could you take a look? Thanks in advance!

revans2:
@HeartSaVioR Sorry I have been way behind on looking at pull requests, among other things.  It looks good to me +1.  Do you want to pull this back into 0.10 too?

HeartSaVioR:
@revans2 
Thanks for the review. :)
Yes, it would be better to include it to next release, since several users already reported such situation.

revans2:
OK I'll check it into both branches.

HeartSaVioR:
@revans2 Thanks for merging!

itaifrenkel:
@HeartSaVioR I also recommend this fix forter@70f5689
Otherwise the bolt can fail when preparing in a busy build server (race condition between reader and heartbeat threads)
Also, I am not sure that anything other than fail/ack/sync really means the bolt is not in an endless loop.

PR #499, opened 2015-04-02
[STORM-746] Skip ack init when there are no output tasks

d2r:
See [STORM-746](https://issues.apache.org/jira/browse/STORM-746) for the description.

### DIFF
--- a/storm-core/src/clj/backtype/storm/daemon/executor.clj
+++ b/storm-core/src/clj/backtype/storm/daemon/executor.clj
@@ -526,7 +526,8 @@
                                                                         out-tuple
                                                                         overflow-buffer)
                                                            ))
-                                         (if rooted?
+                                         (if (and rooted?
+                                                  (seq out-ids)) ;; not empty
                                            (do
                                              (.put pending root-id [task-id
                                                                     message-id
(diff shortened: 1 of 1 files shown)


revans2:
My only concern with something like this would be the possible overhead on normal operation when we are sending a tuple that is going somewhere else.  But the overhead does not look that high so I am +1 on the change. (all the tests passed for me), but I would like to hear other peoples opinions on this.

storm-core/src/clj/backtype/storm/daemon/executor.clj
```diff
@@ -526,7 +526,8 @@
                                                                         out-tuple
                                                                         overflow-buffer)
                                                            ))
-                                         (if rooted?
+                                         (if (and rooted?
+                                                  (seq out-ids)) ;; not empty
```
knusbaum: `(not (.isEmpty out-ids))` might be better. 
`seq` is going to create a new ISeq and do copying every execution of this loop. 
Probably not a big deal, but it's easy to make better.

knusbaum:
Except for my comment above, I'm +1.

storm-core/src/clj/backtype/storm/daemon/executor.clj
d2r: Yes, this is more work than we need. It creates an [IteratorSeq](https://github.com/clojure/clojure/blob/d6baf6e1fb6e2b9ab3e3a459de5ab1d4381938aa/src/jvm/clojure/lang/IteratorSeq.java#L32-L37) each time, since the input is an ArrayList. I'll fix it.

revans2:
+1

@knusbaum are you still OK with this.

knusbaum:
Yes, I'm +1

PR #500, opened 2015-04-02
Issue assignment-version/info-with-version-callback when any assignment changes

lightyang:
JIRA: https://issues.apache.org/jira/browse/STORM-747


revans2:
+1

PR #501, opened 2015-04-02
An option for new/updated Redis keys to expire in RedisMapState

emulvaney:
This is much the same as the mechanism already available via `RedisStateUpdater` for `RedisState` instances.  Users can now set `expireIntervalSec` in their `RedisMapState.Options` instance, to set an expiry timeout on each Redis key updated by the state.

### DIFF
--- a/external/storm-redis/src/main/java/org/apache/storm/redis/trident/state/RedisMapState.java
+++ b/external/storm-redis/src/main/java/org/apache/storm/redis/trident/state/RedisMapState.java
@@ -72,7 +72,7 @@ public class RedisMapState<T> implements IBackingMap<T> {
     public static class Options<T> implements Serializable {
         public int localCacheSize = 1000;
         public String globalKey = "$REDIS-MAP-STATE-GLOBAL";
-        KeyFactory keyFactory = null;
+        public KeyFactory keyFactory = null;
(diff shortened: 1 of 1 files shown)


HeartSaVioR:
Looks good to me overall.

Btw, you're encouraged to request for pull to master branch.
If committer thinks that should be included for 0.10.0, it would be cherry-picked to 0.10.x branch.
You're encouraged to file a JIRA, too.

HeartSaVioR:
Please see https://github.com/apache/storm/blob/master/DEVELOPER.md for how to contribute.

HeartSaVioR:
You can apply your changeset to RedisClusterMapState though it can't use Pipeline for performance.

emulvaney:
Thanks for the feedback.  `RedisClusterMapState`, yes that makes sense.  I haven't set up a Redis cluster yet, but I can see that being useful in the future.

I guess I just skimmed the DEVELOPER file; I thought JIRA could be skipped for small changes.  No worries, I'll write one up.  I'll also switch the target to `master`, but I might have to make another pull request.  I can't see where to change the target branch.

Anyway, it's a holiday weekend, so I'll probably not get to any of this until Monday or Tuesday.  I'm busy being social in meat-space.  :-)

PR #502, opened 2015-04-03
STORM-749: Remove CSRF check.

Parth-Brahmbhatt:


harshach:
+1

HeartSaVioR:
LGTM.
Maybe it would be better to remove "antiForgeryToken" from /api/v1/topology/:id's sample json.

storm-core/src/clj/backtype/storm/ui/core.clj
```diff
@@ -975,7 +971,7 @@
             name (.get_name tplg)]
         (.activate nimbus name)
         (log-message "Activating topology '" name "'")))
-    (json-response (topology-op-response id "deactivate") (m "callback")))
+    (json-response (topology-op-response id "activate") (m "callback")))
```
d2r: Good catch.  This needs to be fixed in 0.10.x-branch also.

d2r:
> Maybe it would be better to remove "antiForgeryToken" from /api/v1/topology/:id's sample json.

Yes, STORM-UI-REST-API.md line 379.

d2r:
https://github.com/Parth-Brahmbhatt/incubator-storm/blob/b7ed77727a1d1e76fc775ac6776b6580e25a1afe/storm-core/src/ui/public/topology.html#L125-L128

Should remove this too, since the template is gone.

Parth-Brahmbhatt:
@d2r @HeartSaVioR fixed.

d2r:
Thanks.  I am +1 for this when we have a similar pull request for the 0.10.x-branch.

HeartSaVioR:
I have a question. When we want to reflect PR into master and existing branch, is it encouraged to make two PRs pointing for each branches, or just one PR with cherry-pick?

d2r:
> I have a question. When we want to reflect PR into master and existing branch, is it encouraged to make two PRs pointing for each branches, or just one PR with cherry-pick?

I think we could cherry-pick, but either way you would have a new branch before the code is merged in.

1) check out 0.10.x-branch
2) create new branch
3) cherry-pick change (test, etc...)
4) make pull request to apache 0.10.x-branch

I do not think we should cherry-pick commits directly to a branch without a pull request, because it is a code change that should have a review.

@ptgoetz, any comment?

Parth-Brahmbhatt:
Here is Pr against 0.10-x branch https://github.com/apache/storm/pull/512

d2r:
+1

PR #504, opened 2015-04-05
[STORM-753] Improve RedisStateQuerier to convert List<Values> from Redis value

HeartSaVioR:
Please see https://issues.apache.org/jira/browse/STORM-753 for more details.

Applied some refactoring to reduce code duplication, and some optimization regarding round-trip / Pipeline.


HeartSaVioR:
Also applied https://github.com/apache/storm/pull/462 for RedisClusterMapState.

HeartSaVioR:
Upmerged.
@revans2 Could you take a look? Thanks!

HeartSaVioR:
For convenience, I posted new PR (#567) which applies #504 into 0.10.x-branch.

external/storm-redis/README.md
```diff
@@ -184,7 +184,8 @@ RedisState
         JedisPoolConfig poolConfig = new JedisPoolConfig.Builder()
                                         .setHost(redisHost).setPort(redisPort)
                                         .build();
-        TridentTupleMapper tupleMapper = new WordCountTupleMapper();
+        RedisStoreMapper storeMapper = new WordCountStoreMapper();
```
revans2: storeMapper does not appear to be used later on, and should probably be removed.

external/storm-redis/README.md
```diff
@@ -205,7 +211,8 @@ RedisClusterState
         }
         JedisClusterConfig clusterConfig = new JedisClusterConfig.Builder().setNodes(nodes)
                                         .build();
-        TridentTupleMapper tupleMapper = new WordCountTupleMapper();
+        RedisStoreMapper storeMapper = new WordCountStoreMapper();
```
revans2: storeMapper does not appear to be used later on, and should probably be removed.

revans2: Or the tupleMapper needs to be updated.

external/storm-redis/README.md
```diff
@@ -194,6 +195,11 @@ RedisState
                                 fields,
                                 new RedisStateUpdater("test_", tupleMapper).withExpire(86400000),
```
revans2: Actually it looks like this needs to be updated instead.

external/storm-redis/src/main/java/org/apache/storm/redis/trident/state/AbstractRedisMapState.java
```diff
@@ -0,0 +1,87 @@
...
+ * limitations under the License.
+ */
+package org.apache.storm.redis.trident.state;
+
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Maps;
+import storm.trident.state.*;
+import storm.trident.state.map.IBackingMap;
+
+import java.util.*;
```
revans2: I don't really like wild care imports.  They can have unintended consequences when moving to a new version of a library.

external/storm-redis/src/main/java/org/apache/storm/redis/trident/state/Options.java
```diff
@@ -0,0 +1,33 @@
...
+import storm.trident.state.Serializer;
+
+import java.io.Serializable;
+
+public class Options<T> implements Serializable {
+	private static final RedisDataTypeDescription DEFAULT_REDIS_DATATYPE = new RedisDataTypeDescription(RedisDataTypeDescription.RedisDataType.STRING);
+
+	public int localCacheSize = 1000;
+	public String globalKey = "$REDIS-MAP-STATE-GLOBAL";
+	KeyFactory keyFactory = null;
```
revans2: Why is the key factory not public?  Do we not want people to override it?

Do we want setters available for these instead of just public data values?

revans2:
In general it looks good.  I would like to see the docs updated to match the test code, and I would like to understand why options.keyFactory is not public.  

I would also love to see some javadocs for the classes being added, although that is not strictly necessary.  Great job by the way.

external/storm-redis/README.md
HeartSaVioR: @revans2 Yes, right. Seems like I forgot to replace tupleMapper. ;( I'll take care of it.

external/storm-redis/src/main/java/org/apache/storm/redis/trident/state/AbstractRedisMapState.java
HeartSaVioR: @revans2 
I agree.
IntelliJ default setting converts some imports to wildcard automatically. Turned off now.

external/storm-redis/src/main/java/org/apache/storm/redis/trident/state/Options.java
HeartSaVioR: @revans2 
Maybe I missed it. keyFactory should be defined public or all fields should have setter.

HeartSaVioR:
@revans2 
Thanks for your careful review! I addressed all of your comments.

I have been using private / protected field and public getter/setter unless it has any special reasons.
Storm-redis has several 'default' or 'public' fields in some classes including Options which I'd like to restrict.
I'll file on new issue addressing it.

Since most of classes in storm-redis doesn't have javadoc, it would be better to treat it to new issue.
I'll file on new issue for javadoc, too.

HeartSaVioR:
Restricting fields: https://issues.apache.org/jira/browse/STORM-844
Javadoc: https://issues.apache.org/jira/browse/STORM-843

ptgoetz:
I'm +1, but with the understanding that I have not done any testing beyond unit tests. I would like to see additional Committer reviews before merging.

revans2:
+1 the changes look good to me.

HeartSaVioR:
@revans2 @ptgoetz Thanks for reviewing. :)

PR #507, opened 2015-04-06
STORM-750: Set Config serialVersionUID

revans2:


knusbaum:
+1

kishorvpatil:
+1

Parth-Brahmbhatt:
+1.

ptgoetz:
Did this just go from pull request to merged in less than 15 minutes?

I know it's a very minor change, and has all the requisite +1s, but that seems a little hasty to me.

revans2:
@ptgoetz,

I agree that it was rather fast (7 mins), but I thought with the bylaws allowing for a retroactive -1 to revert the change, that it was not that big of a deal.

d2r:
@revans2 Yeah, the bylaws do say a minimum of 1 day for code changes.

revans2:
@d2r @ptgoetz 

Crud you are right, I thought that was dropped with the move to retroactive -1.  Sorry about that.  If anyone wants me to revert the change and wait until tomorrow I will be happy to.

ptgoetz:
@revans2,

True, we can retroactively revert if a -1 comes in, but the bylaws state that the minimum approval time for a code change is 1 day.

That's there to give everyone a chance to review (we're all in different time zones). In this case it's a very minor change and I don't think anyone will -1 it, but we should at least give everyone the chance to review a code change.

ptgoetz:
@revans2 I thought about that (-1 to force a revert), but I decided against it. It was an honest mistake made in the spirit of adhering to the bylaws. As I mentioned earlier it is such an innocuous change that I doubt anyone would be against it.

Nonetheless, in the future I think we should stick with the 24-hour guideline.

I'm +1 for leaving it as is, if anyone disagrees, we can always revert.

knusbaum:
We should update the wording of the bylaws, then. I read through them when this went in and (apparently) misinterpreted them.

The rules for code changes state both: "The code can be committed after the first +1" and "Minimum Length: 1 day from initial patch"

I thought, like @revans2, that we'd dropped the 1-day waiting for the retroactive -1. I'm fine with either, but whichever we choose, we should make it more clear in the wording of the bylaws.

ptgoetz:
@knusbaum I agree.

Seems like an AND vs. OR issue between those two sentences. IMO it should be AND.

To me, it seems only fair to allow the sun to pass through all time zones before committing a patch given the geographical distribution of out community.

PR #508, opened 2015-04-06
STORM-757: Simulated time can leak out on errors

revans2:


ptgoetz:
Nice catch.

revans2:
I caught it, because I had other bugs and was beating my head wondering why were the tests getting stuck instead of failing.

revans2:
Wait there is a bug in my test changes :( I'll through up a fix shortly.

revans2:
OK I fixed the bug I found, it should be good for a review now.

knusbaum:
+1

HeartSaVioR:
LGTM. Great!

Btw, I feel it would be another chance of adopting CI (Jenkins, Travis, or anything) to catch some bugs from each PR before merge.

nathanmarz:
-1

Why get rid of the wait-for-condition function?

HeartSaVioR:
@revans2 Regarding timeout, it would be better to use TEST-TIMEOUT-MS on timeout value.

Just a two cents, wait-for-condition doesn't treat timeout so it could be stuck.
So tests should use while-timeout unless wait-for-condition does take care of timeout.
While while-timeout provides more informations on timeout, it would be better for wait-for-condition use while-timeout, too.

ptgoetz:
@nathanmarz To be valid, a veto (-1) must be accompanied by a justification. Can you elaborate on your objections?

If you're just looking for more information, then a -0 might be more appropriate. I'm sure any Committer would seek to address your concerns before merging. Vetoes are rather heavy-handed and can be disruptive. If you feel strongly that this patch should be blocked, then you need to explain why, and be ready and willing to support or retract the veto.

revans2:
@nathanmarz I removed wait-for-condition, because after updating the tests to have a timeout, like @HeartSaVioR mentioned, it was not longer used anywhere in the codebase.  If you would like me to add it back in, or add it back in with a timeout I would be happy to.  I was just trying to clean up dead code.

nathanmarz:
I'm still -1. You're duplicating the same code everywhere in this patch. Just keep the wait-for-condition function and add a timeout parameter. Then have a 1-arity version that defaults the timeout to 5000ms.

revans2:
@nathanmarz Thanks for the review I did like you asked and the code does look a lot cleaner.  I moved wait-for-condition into testing.clj because the while-timeout macro is rather testing specific, it throws an AssertionError and does a thread dump on timeout to help with debugging.

nathanmarz:
+1

PR #509, opened 2015-04-06
STORM-748: Package Multi-Lang scripts so they don't have to be duplicated

ptgoetz:
This moves storm's multi-lang components to a single location so they can be referenced from, rather than copied to other projects.

Each language implementation then becomes an optional dependency. `storm-core` uses them for unit tests, while `storm-starter` now uses them as a dependency rather than duplicating them.

Please hold off on merging until I have a chance to better document the changes. I'm putting this pull request up so others can comment on the approach I took.


ptgoetz:
Commenting to trigger JIRA sync. Ignore.

HeartSaVioR:
Your approach is great!
Btw, since each language contains only one source file, how about aggregating these into multilang module? pom.xml seems to be duplicated, and seems like aggregating these makes no effect on testing / using.

ptgoetz:
@HeartSaVioR Thanks for the feedback. I'm open to either approach...

My thought behind keeping them separate was that some languages may require compilation (e.g. go, c++, etc.), or some sort of build process (maybe involving Maven filtering?). In that case keeping them separate makes more sense. We don't have that requirement now, but we could in the future.

HeartSaVioR:
@ptgoetz Yes, you're right. I didn't think about language which needs compilation. 
Thanks for clarification.

harshach:
+1

revans2:
+1 great job.

PR #510, opened 2015-04-07
Add a convenience target to use when debugging topologies locally

l03wn3:
https://issues.apache.org/jira/browse/STORM-759

It would be really nice if there was a storm target like "jar", but one that would pause and wait for a remote java debugger so you could debug your topology while running locally.
Incidentally, here's a more or less trivial fix for this.

### DIFF
--- a/bin/storm.py
+++ b/bin/storm.py
@@ -83,6 +83,7 @@ init_storm_env()
 CONFIG_OPTS = []
 CONFFILE = ""
 JAR_JVM_OPTS = shlex.split(os.getenv('STORM_JAR_JVM_OPTS', ''))
+JAR_DEBUG_JVM_OPTS = '-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=5005'
 JAVA_HOME = os.getenv('JAVA_HOME', None)
 JAVA_CMD = 'java' if not JAVA_HOME else os.path.join(JAVA_HOME, 'bin', 'java')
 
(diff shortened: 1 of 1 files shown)


Parth-Brahmbhatt:
you could just use the storm jar command and provide the debug args as topology.worker.childopts config. I don't think we need a special command for this.

l03wn3:
Doing 
        Config conf = new Config();
        conf.put(Config.TOPOLOGY_WORKER_CHILDOPTS, "-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=5005");

Doesn't make the JVM suspend upon start and wait for a debugger. Would there be another convenient way?

HeartSaVioR:
@Parth-Brahmbhatt 
Maybe @l03wn3 want to attach debugger with main process which submits topology, not deployed workers. But still I agree it doesn't need to add special command.

@l03wn3 
If I know "storm.py" right, for now you can try adding your arguments to system environment STORM_JAR_JVM_OPTS before executing "storm jar".
Sure you can restore STORM_JAR_JVM_OPTS after doing this.

Parth-Brahmbhatt:
in that case you are right , if you want to test in local mode you should use STORM_JAR_JVM_OPTS.

l03wn3:
@HeartSaVioR Yes, precisely. I've been adding stuff to storm.py and then removing it, but since it's annoying and not convenient to edit the storm binary of the system, I decided to add another target instead. I know there's other less convenient ways, but this is much more convenient and does not require modifying system files.

lazyval:
@Parth-Brahmbhatt +1

d2r:
@l03wn3, is STORM_JAR_JVM_OPTS OK for debugging this way?

If so, would you close this pull request?  Otherwise, would you create a Jira issue for this change?

knusbaum:
Closing this. Please reopen if it's still an issue.

PR #515, opened 2015-04-08
STORM-762: uptime for worker heartbeats is lost when converted to thrift

revans2:
I made uptime_secs required because this is a bug where falling back to broken behavior does not feel right, and the current code has not been officially released yet.  If you want me to make it able to do a rolling upgrade I can without too much difficulty.


Parth-Brahmbhatt:
There are some unchanged files in the generated packages that should be removed from the commit.  I  think given we have not released this version yet it is ok to make this required field.

revans2:
Some of the python files had `#/bin/env python` added at the beginning.  This was part of the script that generated them.  I left them in, because it looked like they were dropped out somehow.  Not totally sure how.  If you want me to remove them I can, but then every time someone generates them again using the current scripts they will show up again.

Parth-Brahmbhatt:
Ok, I think we can keep them as is. I am +1 for the change and thanks for catching this.

PR #516, opened 2015-04-08
[STORM-483] extlib for external, extlib-daemon for daemo-only external

zhuoliu:
1. Packaging creates two new directories named "extlib" and "extlib-daemon" for holding external jars, and daemons-only external jars, respectively.
2. "bin/storm" takes the jars from the above two directories and optional external classpaths, and puts into classpath for launching exec_storm_class (for daemons).
3. In supervisor, when launching a worker, worker-classpath is built from scratch (include extlib but not include extlib-daemon), rather than inherited from supervisor's current-classpath . This is for avoiding classpath leakage from daemons to workers.
Thanks Robert and Naga's comments and proposal.


bin/storm.py
```diff
@@ -85,6 +85,10 @@ def init_storm_env():
 JAR_JVM_OPTS = shlex.split(os.getenv('STORM_JAR_JVM_OPTS', ''))
 JAVA_HOME = os.getenv('JAVA_HOME', None)
 JAVA_CMD = 'java' if not JAVA_HOME else os.path.join(JAVA_HOME, 'bin', 'java')
+NIMBUS_EXTRA_CLASSPATHS = os.getenv('NIMBUS_EXTRA_CLASSPATHS', None)
+SUPERVISOR_EXTRA_CLASSPATHS = os.getenv('SUPERVISOR_EXTRA_CLASSPATHS', None)
```
revans2: I don't see a reason to have NIMBUS and SUPERVISOR specific class paths options.  If you have a good use case, please let me know otherwise I would like to see them removed.

ptgoetz:
+1 (and I assume @revans2 feels the same and just forgot to comment as such :) ).

I would suggest that we open a ticket to document this for end users since it is a departure from how things worked in the past, and might trip up some users.

revans2:
Yes +1 thanks @ptgoetz I had the comment written but didn't hit the button.  Boy I feel dumb, and I agree on the documentation as well.

PR #517, opened 2015-04-08
STORM-764: Have option to compress thrift heartbeat

revans2:


Parth-Brahmbhatt:
Do you want to change the defaults.yaml so by default we use the gzipped thrift serialization delegate?

storm-core/src/jvm/backtype/storm/serialization/GzipBridgeThriftSerializationDelegate.java
```diff
@@ -0,0 +1,65 @@
...
+
+import java.util.Map;
+import java.util.zip.GZIPInputStream;
+
+/**
+ * Always writes gzip out, but tests incoming to see if it's gzipped. If it is, deserializes with gzip. If not, uses
+ * {@link backtype.storm.serialization.DefaultSerializationDelegate} to deserialize. Any logic needing to be enabled
+ * via {@link #prepare(java.util.Map)} is passed through to both delegates.
+ */
+@Deprecated
```
HeartSaVioR: I'm curious about @Deprecated annotation since this class is added just now.
If it should be deprecated now, how about explaining to javadoc about cause of Deprecated and alternative way to accomplish same work?

revans2: Great catch, this is a copy paste error.  I'll fix it.

revans2:
I am fine with changing the default or not.  I left it the way it was, because I was not sure if everyone is as concerned about the load placed on ZK like we are, but if you want the default changed I will.

revans2:
I updated the code to remove the derprecated annotation, and updated the comment to be correct.  I also changed the default to be compressed thrift.

Parth-Brahmbhatt:
+1.

PR #275, opened 2014-10-01
STORM-512 KafkaBolt doesn't handle ticks properly

nielsbasjes:
I implemented this by 
1.   Adding an 'boolean isTick()' method to the Tuple interface. 
       This way it is for application builders much easier to  determine if the tuple at hand is a Tick or not.
2.   Letting the KafkaBolt use that method to ignore the ticks.
3.   Updating the other place in the code base that handles ticks differently to use this new method.

PR #522, opened 2015-04-16
STORM-786: KafkaBolt should ack tick tuples

miguno:
[STORM-512](https://issues.apache.org/jira/browse/STORM-512) (KafkaBolt doesn't handle ticks properly) adds special-casing of tick tuples. What is missing in the patch is that the input tuple, when it is a tick tuple, should be properly acked like normal tuples.


miguno:
We already [received a +1](https://github.com/apache/storm/pull/275#issuecomment-94068063) from @nathanmarz.

revans2:
+1 looks good to me.

PR #523, opened 2015-04-16
[STORM-787]: test-ns should announce test failures with 'BUILD FAILURE'

knusbaum:
Slightly more complicated clojure.
The only thing really changing here is that it checks the values for the `:fail` and `:error` keys in the return value of `run-tests`

In case of error, `System.exit(1)`. 
In the python, propagate the exit code.


revans2:
+1, should help me find tests that fail but not all the time.

HeartSaVioR:
Looks good to me.

PR #526, opened 2015-04-16
STORM-773: STORM-772: Transactional test timeout too low on restart, and InterruptedException

revans2:


revans2:
So after much searching and tracing through logs, with some added logs in the CoordinatedBolt I found out that the CoordinatedBolt was timing out the batch in a few cases, if the batch took longer then 300ms to complete because the timeout is set to 30 seconds by default and 10 seconds of simulated time equals 100ms of wall time.  When this would happen the bolts would be confused and the batch would never be fully acked.  I am not sure why the coordinator/spout was not getting a timeout and replaying the batch in simulated time, but because it is a simulated time issue, and only really shows up on this one test, I decided to increase the timeout.  If others think we should dig deeper and understand why the replay is not happening I am happy to hand the JIRA over to them.

knusbaum:
+1

hustfxj:
+1

PR #528, opened 2015-04-17
STORM-791: Storm UI displays maps in the config incorrectly

revans2:


revans2:
jsonformat

knusbaum:
+1
Super cool.

HeartSaVioR:
Cool!
Is it copied to clipboard as same as we're seeing?

harshach:
+1

revans2:
@HeartSaVioR I didn't try it, but I will.

revans2:
@HeartSaVioR yes, copy/paste comes out exactly as you see it.  Even in the case of maps and lists with values in them there are new lines.  I like the JSON because you can copy and paste it on the command line with a -c command.  But the new lines make it a little difficult.  If you think this is a blocker please let me know and I'll look for a work around, like having a copy button that does not include any extra white space.  Of we could file a follow on JIRA for it.

HeartSaVioR:
@revans2 I don't think it's a blocker and it can be fixed when we think it'll be really necessary.
Looks good to me again.

PR #532, opened 2015-04-22
[STORM-756] ShellBolt can delay sending tuples on demand from subprocess

HeartSaVioR:
Please refer https://issues.apache.org/jira/browse/STORM-756 to see more details.

* introduce new multilang protocol: delay [seconds]
  * command: "delay", msg: "seconds(number format)"
  * python: storm.delay(seconds)
  * ruby: Storm::Protocol.delay(seconds)
* expose pending queue size to let users check and request delay
  * python: storm.getPendingQueueSize()
  * ruby: Storm::Protocol.get_pending_queue_size()

### DIFF
--- a/storm-core/src/jvm/backtype/storm/task/ShellBolt.java
+++ b/storm-core/src/jvm/backtype/storm/task/ShellBolt.java
@@ -29,12 +29,14 @@ import backtype.storm.tuple.Tuple;
 import backtype.storm.utils.ShellProcess;
 import clojure.lang.RT;
 import com.google.common.util.concurrent.MoreExecutors;
+import com.google.common.util.concurrent.Uninterruptibles;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.*;
 import java.util.concurrent.*;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 
 import static java.util.concurrent.TimeUnit.SECONDS;
@@ -82,13 +84,14 @@ public class ShellBolt implements IBolt {
 
     private Thread _readerThread;
     private Thread _writerThread;
-    
+
     private TopologyContext _context;
 
     private int workerTimeoutMills;
     private ScheduledExecutorService heartBeatExecutorService;
     private AtomicLong lastHeartbeatTimestamp = new AtomicLong();
     private AtomicBoolean sendHeartbeatFlag = new AtomicBoolean(false);
+    private AtomicInteger delaySeconds = new AtomicInteger(0);
 
     public ShellBolt(ShellComponent component) {
         this(component.get_execution_command(), component.get_script());
@@ -246,7 +249,7 @@ public class ShellBolt implements IBolt {
         if (name.isEmpty()) {
             throw new RuntimeException("Receive Metrics name is empty");
         }
-        
+
         //get metric by name
         IMetric iMetric = _context.getRegisteredMetricByName(name);
         if (iMetric == null) {
@@ -256,7 +259,7 @@ public class ShellBolt implements IBolt {
             throw new RuntimeException("Metric["+name+"] is not IShellMetric, can not call by RPC");
         }
         IShellMetric iShellMetric = (IShellMetric)iMetric;
-        
+
         //call updateMetricFromRPC with params
         Object paramsObj = shellMsg.getMetricParams();
         try {
@@ -265,7 +268,17 @@ public class ShellBolt implements IBolt {
             throw re;
         } catch (Exception e) {
             throw new RuntimeException(e);
-        }       
+        }
+    }
+
+    private void handleDelay(ShellMsg shellMsg) {
+        try {
+            Integer seconds = Integer.valueOf(shellMsg.getMsg());
+            int newDelaySeconds = delaySeconds.addAndGet(seconds);
+            LOG.info("Delay requested from subprocess, new delay seconds : " + newDelaySeconds);
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
     }
 
     private void setHeartbeat() {
@@ -310,8 +323,6 @@ public class ShellBolt implements IBolt {
 
             sendHeartbeatFlag.compareAndSet(false, true);
         }
-
-
     }
 
     private class BoltReaderRunnable implements Runnable {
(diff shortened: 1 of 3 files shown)


itaifrenkel:
AFAIK the spout can just ignore the nextTuple requests. Are you trying to save on stdout/stdin nextTuple traffic ?

HeartSaVioR:
@itaifrenkel 
I think Spout doesn't matter since subprocess has a chance to slow down its flow.
But Bolt does matter cause ShellBolt sends all tuples to subprocess, and ShellBolt actually doesn't trace tuples.
More details are described to STORM-756.

HeartSaVioR:
@itaifrenkel Actually I read paper about Twitter Heron, and it can propagate backpressure. 
Storm currently doesn't have it, so ShellBolt itself should ensure backpressure cause it's better to let machine failure triggered by subprocess (unlimited memory consumption).

HeartSaVioR:
@itaifrenkel 
I see your point. 
Spout can skip sending nextTuple to subprocess so that we can save stdout/stdin traffic and Spout doesn't block or sleep and continue to handle ack / fail.
Two points still exist,
1. When Spout can skip sending nextTuple?
2. Who is better to decide it? ShellSpout? Or subprocess?

revans2:
@itaifrenkel any update on this?  we now have automatic back-pressure like heron so this would be a nice addition to it too.

HeartSaVioR:
@revans2 
This PR improves subprocess to request ShellBolt for back pressure, but I left decision when to request  to users since I thought it is more flexible.
If we want to provide default, sure we can have default condition for requesting back-pressure via high/low watermark (like ABP).

Btw, we now have ABP, which seems not compatible to ShellBolt.
ShellBolt just receives tuples from Disruptor Queue and pass to subprocess whether subprocess is busy or not. In result size of the receive queue doesn't nearly reach high watermark.
One idea to make ShellSpout be compatible to ABP is, changing implementation of ShellBolt to sync (like ShellSpout). 
Expected downside is it may hurt performance, but it is actually more natural flow.

revans2:
@HeartSaVioR I am fine either way you want to approach this.  I would prefer not to change the protocol in a non-backwards compatible way, and adding a required sync feels like an incompatible change.  Having the bolt turn on and off BP feels like a more backwards compatible change to me.

revans2:
I'm not sure I like the delay especially in seconds.  I have been playing around with performance and we could easily process tens of thousands of tuples in a second.  1 second feels like way too long of a time for a delay.

Also reading through the code I think we do get ABP when TOPOLOGY_SHELLBOLT_MAX_PENDING is set.  Unless the bolt is explicitly multi-threaded we only have a single thread of execution. So when the bolt is processing data it is not reading more tuples to process.  On all POSIX systems, and I assume windows as well stdin has a limited buffer size, so when that buffer fills up because the process is doing other things the BoltWriterRunnable thread will block, and then _pendingWrites will fill up, and the emit thread will block, and the input queue will backup.

I am happy to write a test and verify that it works.

HeartSaVioR:
@revans2 
Oh, I have been forgotten TOPOLOGY_SHELLBOLT_MAX_PENDING when creating this PR. 
I agree that we would be fine with TOPOLOGY_SHELLBOLT_MAX_PENDING since it guarantees count of pending tuples with blocking manner, and finally ABP would be triggered when subprocess couldn't handle tuples as fast as it flows.

Btw, do we have to take care of ABP when TOPOLOGY_SHELLBOLT_MAX_PENDING is off?

revans2:
@HeartSaVioR I would like to see ABP work correctly be default on the Shell bolt, and have it work correctly when TOPOLOGY_SHELLBOLT_MAX_PENDING is off, but I don't think a change to the protocol is going work.  The buffer that fills up is in java, not in the external process.  Having the ability for the external process to throttle only really makes since if the external process is buffering tuples somehow and then it can just stop reading to let us know to stop as well.

Perhaps what we can just have a default value for TOPOLOGY_SHELLBOLT_MAX_PENDING.  The only reason we would want to turn it off completely is that we have a deadlock if taskIds are requested.  Here we write a message back into _pendingWrites, which with a hard coded limit could cause the only thread that reads from _pendingWrites to block trying to write to it.  If we can fix this bug then I would say we just set a default value.

HeartSaVioR:
@revans2 
Sorry to respond late.
I agree that we can set default value of TOPOLOGY_SHELLBOLT_MAX_PENDING to reasonable value  (not off) when deadlock bug you stated is fixed. Then ShellBolt could play well with ABP. 
I'm curious that ShellBolt could play with ABP when TOPOLOGY_SHELLBOLT_MAX_PENDING is off cause ShellBolt always tries to buffer pending writes ASAP regardless of pressure of subprocess.

In order to resolve deadlock, I think we can have another queue which only stores pending messages for taskid which is unbounded (not affected to TOPOLOGY_SHELLBOLT_MAX_PENDING). 
And we introduce priorities for kinds of ShellBolt message, heartbeat > taskid > process, and let BoltWriterRunnable always tries to send message for higher priority. 
In result, only Thread which runs execute() could be blocked, and it is unblocked as soon as pending taskids are all sent.

It doesn't end up starvation because heartbeat is triggered for every 1 sec, and taskids messages are never generated while subprocess doesn't process any requests.

Additional points of this approach are,
1. Subprocess can finish emitting new tuple regardless of count of added pending messages before adding taskids. We can reduce waiting time for taskids.
2. It can also greatly reduce size of subprocess's pending_commands queue, which could reduce out of memory issue for subprocess. It is more important than ShellBolt's OOME because subprocess's memory issue could make machine hang or down.
3. 2. also reduce wait time for processing heartbeat message, which has been a headache issue.

I'll close this pull request and work on new approach.

HeartSaVioR:
@revans2 
I thought about deadlock again, and I couldn't see deadlock you stated.

Let's think about these scenario, subprocess emits tuple and waits for taskids while _pendingWrites is full.

- ExecutorThread will be blocked waiting for writing BoltMsg to _pendingWrites (since it's fix size queue).
- ReaderThread will be blocked waiting for writing taskids - which subprocess is waiting for - to _pendingWrites (again, since it's fix size queue).
- WriterThread will pull message from _pendingWrites and send to subprocess.
  - Though message is BoltMsg, not taskids, subprocess can add BoltMsg (tuple) to its pending queue (unbounded) when waiting for taskids.
- _pendingWrites adds message in order, so eventually taskids is written to _pendingWrites and ReaderThread will be unblocked. (Sure, eventually ExecutorThread will be unblocked, too)
- Subprocess can finish emitting tuple, and continue processing.

It seems to be safe because of unbounded queues in subprocess.
Please let me know if I missed something.

HeartSaVioR:
If there's no deadlock when _pendingWrites is bounded queue, we can just set default value of TOPOLOGY_SHELLBOLT_MAX_PENDING to reasonable and deny set it to null. In result ABP will work.

Btw, approach I commented before still have advantages whether there's deadlock issue or not.

PR #534, opened 2015-04-24
[STORM-436] Clean up ns declarations

danielcompton:
[STORM-436](https://issues.apache.org/jira/browse/STORM-436)

The purpose of this change is to remove as many :use's as possible. This will make the codebase codebase easier to understand and maintain because we can see where vars are coming from.

It will also be cleaner and more idiomatic Clojure.

### DIFF
--- a/storm-core/src/clj/backtype/storm/command/activate.clj
+++ b/storm-core/src/clj/backtype/storm/command/activate.clj
@@ -14,11 +14,11 @@
 ;; See the License for the specific language governing permissions and
 ;; limitations under the License.
 (ns backtype.storm.command.activate
-    (:require [backtype.storm.thrift :as t]
-      [backtype.storm.log :as log])
-    (:gen-class))
+  (:require [backtype.storm.thrift :as thrift]
+            [backtype.storm.log :refer [log-message]])
+  (:gen-class))
 
 (defn -main [name]
-  (t/with-configured-nimbus-connection nimbus
+  (thrift/with-configured-nimbus-connection nimbus
     (.activate nimbus name)
-    (log/log-message "Activated topology: " name)))
+    (log-message "Activated topology: " name)))
--- a/storm-core/src/clj/backtype/storm/command/activate.clj
+++ b/storm-core/src/clj/backtype/storm/command/activate.clj
@@ -14,11 +14,11 @@
 ;; See the License for the specific language governing permissions and
 ;; limitations under the License.
 (ns backtype.storm.command.activate
-  (:use [backtype.storm thrift log])
-  (:gen-class))
+    (:require [backtype.storm.thrift :as t]
+      [backtype.storm.log :as log])
+    (:gen-class))
 
-(defn -main [name] 
-  (with-configured-nimbus-connection nimbus
+(defn -main [name]
+  (t/with-configured-nimbus-connection nimbus
     (.activate nimbus name)
-    (log-message "Activated topology: " name)
-    ))
+    (log/log-message "Activated topology: " name)))
--- a/storm-core/src/clj/backtype/storm/command/dev_zookeeper.clj
+++ b/storm-core/src/clj/backtype/storm/command/dev_zookeeper.clj
@@ -15,8 +15,8 @@
 ;; limitations under the License.
 (ns backtype.storm.command.dev-zookeeper
   (:require [backtype.storm.zookeeper :as zoo]
-            [backtype.storm.config :as config])
-  (:use [backtype.storm util ])
+            [backtype.storm.config :as config]
+            [backtype.storm.util :refer [rmr]])
   (:gen-class))
 
 (defn -main [& args]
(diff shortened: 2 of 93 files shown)


storm-core/src/clj/backtype/storm/command/dev_zookeeper.clj
```diff
@@ -14,13 +14,15 @@
 ;; See the License for the specific language governing permissions and
 ;; limitations under the License.
 (ns backtype.storm.command.dev-zookeeper
-  (:use [backtype.storm zookeeper util config])
+  (:require [backtype.storm.zookeeper :as zoo]
+            [backtype.storm.config :as config])
+  (:use [backtype.storm util ])
```
knusbaum: Trailing space after util.

Is there a reason we're keeping util in a use form instead of putting it in a require, unlike the others?

The second item in the jira is:
'2. Order imports use, require, import.'
but here you've put the use after the require.

storm-core/src/clj/backtype/storm/command/activate.clj
```diff
@@ -14,11 +14,11 @@
 ;; See the License for the specific language governing permissions and
 ;; limitations under the License.
 (ns backtype.storm.command.activate
-  (:use [backtype.storm thrift log])
-  (:gen-class))
+    (:require [backtype.storm.thrift :as t]
+      [backtype.storm.log :as log])
```
knusbaum: Indentation is a little off here, and I'd prefer `:as thrift` or somesuch rather than `:as t`.

storm-core/src/clj/backtype/storm/command/rebalance.clj
```diff
@@ -14,19 +14,20 @@
...
 ;; limitations under the License.
 (ns backtype.storm.command.rebalance
-  (:use [clojure.tools.cli :only [cli]])
-  (:use [backtype.storm thrift config log])
+  (:require [clojure.tools.cli :refer [cli]]
+            [backtype.storm.thrift :as thrift]
+            [backtype.storm.log :as log])
   (:import [backtype.storm.generated RebalanceOptions])
   (:gen-class))
+;; TODO: do we need all these gen-class's?
```
knusbaum: For the commands, yes. These are the classes launched by bin/storm.
If you find other `:gen-class` instances you think we can get rid of, feel free to point them out.

storm-core/src/clj/backtype/storm/config.clj
```diff
@@ -20,7 +20,8 @@
   (:import [backtype.storm Config ConfigValidation$FieldValidator])
   (:import [backtype.storm.utils Utils LocalState])
   (:import [org.apache.commons.io FileUtils])
-  (:require [clojure [string :as str]])
+  (:require [clojure.string :as str]
+            [backtype.storm.util :as util])
   (:use [backtype.storm log util]))
```
knusbaum: Leftover `(:use [... util])`

knusbaum:
I'm generally in favor of this. It's something that's bugged me for a while. 
I did a preliminary review on a portion of the changes. Mostly just some minor spacing and stuff.

One thing I'd like is to get rid of `:use` completely, if we're going to refactor the ns forms. Clojure devs discourage its use in favor of `(:require whatever.namespace :refer :all)`.

More consistency with which namespaces we're requiring `:as` something and which ones we're importing wholesale with `:use` or `:refer :all` would be nice. I see sometimes you `:require` `backtype.storm.util` and sometimes you leave it in a `:use` They don't all have to be one way or the other, but I get the sense that the ones you've chosen to change were chosen arbitrarily (Or maybe just accidentally skipped over; there's a lot to change).

We should probably also put a namespace style policy in writing, just to make it easy for others to follow.

storm-core/src/clj/backtype/storm/command/dev_zookeeper.clj
danielcompton: The plan is to get rid of all :use's, I had missed this one. I'll fix the trailing space here too.

storm-core/src/clj/backtype/storm/command/activate.clj
danielcompton: Will do

storm-core/src/clj/backtype/storm/config.clj
danielcompton: Hadn't quite got to squashing all of the :imports, :requires and removing the :uses here yet

danielcompton:
My plan is to get rid of all of the :uses by the time I've finished this patch. There are some spots where I've left :uses because I haven't finished refactoring those namespaces. There's also a few aliasing and :refering inconsistencies between my earlier work and the later namespaces which I need to go back and fix.

A namespace style policy would be good too, would that go in DEVELOPERS.md?

As I work on this, is it going to be easier for me to keep squashing my changes into the same commit, or wait until its finished and squash them before merging?

danielcompton:
I think it makes sense to keep committing incrementally as I work to make things easier to review, and squashing it once it's accepted.

revans2:
@danielcompton any update on this?  It would be a nice change to get in before 0.11, if you have time to do it.

danielcompton:
I've let things lapse a little bit on this. I'll try to get it up and running against master soon though.

knusbaum:
Closing this. I like the idea, but this PR is very stale, and we're translating to Java.

PR #537, opened 2015-04-27
STORM-797: DisruptorQueueTest has some race conditions in it.

revans2:
The test had two race conditions in it.  The first was that the data returned by the Consumer was not volatile.  The second was that run gave no time for the consumer to actually receive anything so under heavy load it was possible for it to not actually complete.


revans2:
The failed CI build is because #486 is not in yet.

kishorvpatil:
LGTM. +1

knusbaum:
+1

PR #541, opened 2015-05-01
STORM-807: quote args correctly

revans2:


HeartSaVioR:
LGTM, I've confirmed with print COMMAND and ARGS.

Before applying
```
$ ./storm list "a b" c d
list
['a', 'b', 'c', 'd']
```

After applying
```
$ ./storm list "a b" c d
list
['a b', 'c', 'd']
```

liaokevin:
What you mean

HeartSaVioR:
LGTM = Looks good to me

revans2:
@liaokevin previously bin/storm was a python script but recently it changed to a shell script that wraps the original python script.  The shell script did not quote arguments passed to the python script correctly so if I ran.

```./storm list "a b" c d```
it would not honor the quotes around "a b" and treat them as separate arguments.  This is a problem if we include space in the name of a topology, or if we try to add a config to the command line with a space in it.

ptgoetz:
+1

revans2:
@ptgoetz I am merging this into 0.10.x-branch as well, because it was introduced there.  If you disagree please let me know and I will revert it.

revans2:
@ptgoetz never mind, JIRA is off STORM-675 said it went into 0.10.0, but it did not.  I will just merge this to trunk.

PR #544, opened 2015-05-05
STORM-809: Topology.message.timeout.secs should not allow null or <= 0

revans2:


HeartSaVioR:
LGTM.
Regarding NotNullPosIntegerValidator, it may be better to add some tests on config_test.clj.

revans2:
@HeartSaVioR great comment.  I added in the tests.

HeartSaVioR:
Great! I looked at b5ace22 and also looks good to me.

harshach:
+1

PR #547, opened 2015-05-06
STORM-811: remove old metastor_db before running tests again.

revans2:


harshach:
@revans2  I am +1 on the patch. Good to remove metastore_db before running tests.

PR #454, opened 2015-03-03
Storm-697: Support for Emitting Kafka Message Offset and Partition

mjtieman:
It would be nice expose the offset and partition of messages consumed from Kafka to the Scheme generating the Tuples. This is useful for auditing/replaying data from arbitrary points on a Kafka topic, saving the partition and offset of each message of a discrete stream instead of persisting the entire message.

* Added new sheme to that accepts a Partition and the message offset in its deserialization method.
* Defined an overload of `KafkaUtils.generateTuples` to accept a `Partition` and offset in addition to the message byte[].
* Added a flag in `KafkaConfig` to indicate if the metadata (partition and offset) should be available during tuple generation.
* Wrote a simple String implementation of the new scheme, `StringMessageAndMetadataScheme`, following the same pattern as `StringKeyValueScheme`.
* Unit tests.

PR #550, opened 2015-05-12
STORM-697: Support for Emitting Kafka Message Offset and Partition

itachi90:
Allow kafkaSpout to emit message offset and partition so that user are able to track tuple using this information.


mjtieman:
STORM-697 is covered by https://github.com/apache/storm/pull/454

PR #551, opened 2015-05-12
[STORM-816] maven-gpg-plugin does not work with gpg 2.1

knusbaum:
GPG 2.1 changed some security things involving password entry, and it looks like our version of the plugin doesn't work with it.

We should upgrade.


harshach:
+1

revans2:
+1

PR #553, opened 2015-05-14
[STORM-710] bin/storm command list out all the classes in the output for a command

oleg-ost:
Added -v flag for all bin/storm commands.
Long commands' outputs are removed.
Using -v flag prints traditional long output.

### DIFF
--- a/bin/storm.py
+++ b/bin/storm.py
@@ -190,7 +190,10 @@ def exec_storm_class(klass, jvmtype="-server", jvmopts=[], extrajars=[], args=[]
         "-Dstorm.conf.file=" + CONFFILE,
         "-cp", get_classpath(extrajars, daemon),
     ] + jvmopts + [klass] + list(args)
-    print("Running: " + " ".join(all_args))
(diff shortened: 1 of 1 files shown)


revans2:
I like the concept, but by having storm.py pull the -v out of the command line, it now makes it impossible for any command to use -v any other way.  This is not a big deal for the typical storm commands, but for storm jar it is backwards incompatible.  This is especially a problem for something that for some reason had a -v with an argument.  I would much rather use an environment variable, or a rather log parameter like '--storm-system-verbose' which is not likely to have any collision.

harshach:
@oleg03 can you send us a new patch with @revans2 feedback . Thanks.

revans2:
@oleg03 I don't really like the idea of adding in a -v option that restricts all other commands, including storm jar from supporting it.  I would much rather have a long parameter that is unlikely to collide with something someone else is already using, or us a config or environment variable to trigger it instead.

knusbaum:

PR #554, opened 2015-05-16
[STORM-820] Aggregate topo stats on nimbus, not ui

d2r:


d2r:
This pull request requires some more scrutiny because of its size and to make sure it does not break functionality.

Goals for this PR

* New Nimbus Thrift calls:
   * getTopologyPageInfo
   * getComponentPageInfo
* Nimbus Thrift call getTopologyInfo remains
* Aggregates metrics on nimbus so that serialized data is much smaller for large, highly-connected topologies
* Moves ui code to stats namespace
* Iterates over heartbeats one time.
* Thrift deserialization code avoids deserializing to symbols, preferring strings, since making a symbol out of a string like "600" results in a symbol 600 and not an integer 600.  This can be very confusing to debug, and breaks key comparisons in places.
* Removes some dead code and does a little refactoring in places.

storm-core/src/clj/backtype/storm/log.clj
```diff
@@ -44,3 +46,11 @@
 (defn log-stream
   [& args]
   (apply log/log-stream args))
+
+(defmacro pprint-message
```
revans2: I would prefer to have the name log somewhere in the function name.  perhaps log-pprint

storm-core/src/clj/backtype/storm/stats.clj
```diff
@@ -322,16 +307,16 @@
   [(.get_componentId global-stream-id) (.get_streamId global-stream-id)])
 
 (defmethod clojurify-specific-stats BoltStats [^BoltStats stats]
-  [(window-set-converter (.get_acked stats) from-global-stream-id symbol)
-   (window-set-converter (.get_failed stats) from-global-stream-id symbol)
-   (window-set-converter (.get_process_ms_avg stats) from-global-stream-id symbol)
-   (window-set-converter (.get_executed stats) from-global-stream-id symbol)
-   (window-set-converter (.get_execute_ms_avg stats) from-global-stream-id symbol)])
+  [(window-set-converter (.get_acked stats) from-global-stream-id identity)
```
revans2: Why exactly are we no longer converting this to a symbol?

storm-core/src/clj/backtype/storm/stats.clj
```diff
@@ -373,4 +360,1248 @@
     (ExecutorStats. (window-set-converter (:emitted stats) str)
       (window-set-converter (:transferred stats) str)
       specific-stats
-      rate)))
\ No newline at end of file
+      rate)))
+
+(defn- agg-bolt-lat-and-count
+  "Aggregates number executed and process & execute latencies across all
```
revans2: This is really minor but it looks a little funny mixing "and" and "&" this way.

revans2: OK I understand it is order of operations like thing.  I'm not sure a good way to explain it without being ambiguous.

storm-core/src/clj/backtype/storm/stats.clj
```diff
@@ -373,4 +360,1248 @@
...
+                                   (if (and avg num-e)
+                                     (* avg num-e)
+                                     0)))]
+    (into {}
+      (for [k (keys idk->comp-avg)]
+        [k {:completeLatencyTotal (weight-avg k (get idk->comp-avg k))
+            :acked (get idk->acked k)}]))))
+
+(defn swap-map-order
+  "{:a {:A 3, :B 5}, :b {:A 1, :B 2}}
```
revans2: This doc string is a bit confusing.  I understand what the method does, but I had to read the code to really get it.

storm-core/src/clj/backtype/storm/stats.clj
```diff
@@ -373,4 +360,1248 @@
...
+      (reduce add-pairs
+              [0. 0]) ;; Combine weighted averages and counts.
+      ((fn [[weighted-avg cnt]]
+        (div weighted-avg (* 1000 (min uptime 600))))))))
+
+(defn- page+comp-dispatch
+  [page-type comp-type & _]
+  [page-type comp-type])
+
+(defmulti agg-pre-merge
```
revans2: I'm not sure a defmulti is the cleanest here.  All the places it is used are hard code the first two parameters which are used for routing.

storm-core/src/clj/backtype/storm/ui/core.clj
```diff
@@ -783,128 +504,180 @@
...
+     "inputStats" (map unpack-bolt-input-stat (.get_gsid_to_input_stats info))
+     "outputStats" (map unpack-comp-output-stat (.get_sid_to_output_stats info))
+     "executorStats" (map (partial unpack-comp-exec-stat topology-id)
+                          (.get_exec_stats info))}
+    (-> info .get_errors (component-errors topology-id))))
+
+(defmethod unpack-component-page-info ComponentType/SPOUT
+  [^ComponentPageInfo info topology-id window include-sys?]
+  (merge
+    {"spoutStats" (map unpack-comp-agg-stat (.get_window_to_stats info))
```
revans2: Is this a backwards incompatible change in the Restful web service?  I see that the mustache tags changed.  I'm not sure we want to do that if someone is relying on this, even if it was not named the best before.

storm-core/src/clj/backtype/storm/stats.clj
d2r: It is not necessary to convert the strings keys coming out from thrift deserialization, since we use them as strings anyway.

It is also confusing to debug, as the symbol `600` and the number `600` do not compare equally, yet they look exactly the same when printed out.

I changed this because in the course of writing the code, I printed the contents of a heartbeat and saw `600`, and when I tried to retrieve the value for the key `600` as a number, I got nothing back.

storm-core/src/clj/backtype/storm/stats.clj
d2r: Yeah, I'll reword it.

storm-core/src/ui/public/templates/topology-page-template.html
```diff
@@ -131,33 +131,37 @@
     </tbody>
   </table>
 </script>
+
+<script id="topology-visualization-container-template" type="text/html">
```
revans2: What is this change for? It looks like it just moved from topology-visualization-template to a new topology-visualization-container-template, that is now loaded through AJAX.

storm-core/src/storm.thrift
```diff
@@ -222,6 +222,86 @@ struct TopologyInfo {
 514: optional string owner;
 }
 
+struct CommonAggregateStats {
+513: optional i32 num_executors;
```
revans2: Can we perhaps make all of these start with a 1?  There were other places that I used a 512 for security to avoid a collision with others possibly making changes at the same time while it was on a different branch, but I don't think we have that problem here.

storm-core/src/clj/backtype/storm/stats.clj
d2r: I'll rewrite.

storm-core/src/clj/backtype/storm/stats.clj
d2r: Agreed. I'll make them unique functions.

storm-core/src/ui/public/templates/topology-page-template.html
d2r: For the visualization to initialize, it needs to call the old nimbus thrift API call getTopologyInfo, which has not changed.  I only wanted the visualization to initialize when the user shows it, and not pull down all the stream data for every executor each time the page loads.  If we did that, we would lose most of the gains from aggregating stats data on nimbus.

storm-core/src/storm.thrift
d2r: Agreed, I'll change this.

storm-core/src/clj/backtype/storm/ui/core.clj
d2r: I had changed it because it was named inconsistently, and I also should have updated STORM-UI-REST-API.md.

However since it is not essential, I'll revert to using spoutSummary so that the REST API does not change.

storm-core/src/clj/backtype/storm/log.clj
d2r: I'll rename it.

storm-core/src/clj/backtype/storm/stats.clj
```diff
@@ -373,4 +360,1248 @@
...
+                      [sid->comp-avg
+                       sid->num-acked]))}
+  (letfn [(weight-avg [[id avg]] (* avg (get sid->num-acked id)))]
+    {:completeLatencyTotal (sum (map weight-avg sid->comp-avg))
+     :acked (sum (vals sid->num-acked))}))
+
+(defn add-pairs
+  ([] [0 0])
+  ([[a1 a2] [b1 b2]]
+   [(+ a1 b1) (+ a2 b2)]))
```
knusbaum: This is essentially a zipWith:
```
(let [a [1 2]
       b [3 4]]
  (map + a b))
=> (4 6)
```

storm-core/src/clj/backtype/storm/stats.clj
```diff
@@ -373,4 +360,1248 @@
...
+                       idk->executed]))}
+  (letfn [(weight-avg [id avg] (let [num-e (idk->executed id)]
+                                   (if (and avg num-e)
+                                     (* avg num-e)
+                                     0)))]
+    (into {}
+      (for [k (keys idk->exec-avg)]
+        [k {:executeLatencyTotal (weight-avg k (idk->exec-avg k))
+            :processLatencyTotal (weight-avg k (idk->proc-avg k))
+            :executed (idk->executed k)}]))))
```
knusbaum: Why not pull the value out in the for's binding vector?
```
(for [[k v] idk->exec-avg]
  [k {:executeLatencyTotal (weight-avg k v)
    :processLatencyTotal (weight-avg k v)
    :executed v}]))))
```

storm-core/src/clj/backtype/storm/stats.clj
```diff
@@ -373,4 +360,1248 @@
...
+                      [idk->comp-avg
+                       idk->acked]))}
+  (letfn [(weight-avg [id avg] (let [num-e (get idk->acked id)]
+                                   (if (and avg num-e)
+                                     (* avg num-e)
+                                     0)))]
+    (into {}
+      (for [k (keys idk->comp-avg)]
+        [k {:completeLatencyTotal (weight-avg k (get idk->comp-avg k))
+            :acked (get idk->acked k)}]))))
```
knusbaum: Same here.

storm-core/src/clj/backtype/storm/stats.clj
```diff
@@ -373,4 +360,1248 @@
...
+  uptime."
+  [m uptime]
+  (when uptime
+    (->>
+      ;; For each stream, create weighted averages and counts.
+      (merge-with (fn weighted-avg+count-fn
+                    [avg cnt]
+                    [(* avg cnt) cnt])
+                  (get (:execute-latencies m) "600")
+                  (get (:executed m) "600"))
```
knusbaum: I'd rather see "600" def'd somewhere.

storm-core/src/clj/backtype/storm/stats.clj
```diff
@@ -373,4 +360,1248 @@
...
+      vals ;; Ignore the stream ids.
+      (reduce add-pairs
+              [0. 0]) ;; Combine weighted averages and counts.
+      ((fn [[weighted-avg cnt]]
+        (div weighted-avg (* 1000 (min uptime 600))))))))
```
knusbaum: Same here.

d2r:
Review comments should be addressed.

d2r:
Addressed review comments

storm-core/src/clj/backtype/storm/stats.clj
knusbaum: Whoops. Nevermind. Totally missed that the maps were different.

knusbaum:
+1

storm-core/src/clj/backtype/storm/stats.clj
```diff
@@ -373,4 +362,1222 @@
...
+
+(defn- aggregate-comp-stats*
+  [window include-sys? data init-val]
+  (-> (partial agg-comp-exec-stats
+               window
+               include-sys?)
+      (reduce init-val data)))
+
+(defmulti aggregate-comp-stats
+  (fn dispatch-fn [& args] (-> args last first :stats :type)))
```
d2r: This could result in a null dispatch value if the component's heartbeat does not contain metrics.  We should be using the `:type` directly instead of accessing it via `:stats`.

d2r:
Debugging a corner case we found today...

d2r:
Handled some corner cases when the worker process crashes such that executor heartbeats have no metrics data.

d2r:
Leaving a comment here for now: the key [here](https://github.com/d2r/storm/blob/6b606cf68f60858bdc9064c1fe50c873e6dd1971/storm-core/src/clj/backtype/storm/stats.clj#L961) should be `:complete-latency` or else each spout's complete latency will show 0.000 on the topology page.  (Component page is not affected.)

d2r:
This needs to be upmerged with the nimbus-ha changes

storm-core/src/jvm/backtype/storm/generated/BoltStats.java
```diff
@@ -51,7 +51,7 @@
 import org.slf4j.LoggerFactory;
 
 @SuppressWarnings({"cast", "rawtypes", "serial", "unchecked"})
-@Generated(value = "Autogenerated by Thrift Compiler (0.9.2)", date = "2015-9-3")
+@Generated(value = "Autogenerated by Thrift Compiler (0.9.2)", date = "2015-9-28")
```
revans2: Please fix, only a time stamp change.

storm-core/src/clj/backtype/storm/log.clj
```diff
@@ -44,3 +46,11 @@
...
   [& args]
   (apply log/log-stream args))
+
+(defmacro log-pprint
+  [& args]
+  `(let [^StringWriter writer# (StringWriter.)]
+     (doall
+       (for [object# [~@args]]
+         (pprint object# writer#)))
+     (log-message "\n" writer#)))
```
revans2: Why is there a "\n" before logging the message?

storm-core/src/jvm/backtype/storm/generated/ClusterSummary.java
```diff
@@ -51,7 +51,7 @@
 import org.slf4j.LoggerFactory;
 
 @SuppressWarnings({"cast", "rawtypes", "serial", "unchecked"})
-@Generated(value = "Autogenerated by Thrift Compiler (0.9.2)", date = "2015-3-2")
+@Generated(value = "Autogenerated by Thrift Compiler (0.9.2)", date = "2015-9-28")
```
revans2: Here is another one that is just a date change

revans2:
For the most part this looks good.  My only comment would be to update the REST API docs to describe what the new APIs are like.  And did you change time to errorTime in the JSON response?

storm-core/src/clj/backtype/storm/log.clj
d2r: I think it was to clean up the output. Otherwise the data structure would not begin on its own line but instead begin after the timestamp, etc., and so that was not "pretty."

d2r:
> And did you change time to errorTime in the JSON response?

Good catch. Updated API doc with a note on the name change.

revans2:
Thanks for the update I am +1 on the change now.  The aggregation code is a bit complex, but I cannot think of any way to make it less complex right now.

revans2:
@d2r please upmerge.  The code still looks great but it looks like upSecs and a debug were added into the results.  I really want to get this in, as it will speed up page loads a lot.

d2r:
ok working on it

d2r:
OK, I upmerged, and then had to upmerge again right after that.  I double-checked that uptimeSeconds appear in the places they should, and that I was able to change the log level for a worker running ExclamationTopology.

revans2:
Still +1

PR #557, opened 2015-05-18
[STORM-737] Check task->node+port with read lock to prevent sending to closed connection

HeartSaVioR:
(Actually it's another approach from #521, but this changeset is cleaner so I'd like to move on this PR.)

It's based on Nathan's comment, please refer to #349 (comment)

861a92e seems to make a regression. 
But it also introduces optimizations of sending, it shouldn't be discarded.

I'm still not familiar with clojure so please review and comment if it can be optimized.

Thanks!

### DIFF
--- a/storm-core/src/jvm/backtype/storm/utils/TransferDrainer.java
+++ b/storm-core/src/jvm/backtype/storm/utils/TransferDrainer.java
@@ -26,88 +26,47 @@ import backtype.storm.messaging.TaskMessage;
 
 public class TransferDrainer {
 
-  private HashMap<String, ArrayList<ArrayList<TaskMessage>>> bundles = new HashMap();
-  
-  public void add(HashMap<String, ArrayList<TaskMessage>> workerTupleSetMap) {
-    for (String key : workerTupleSetMap.keySet()) {
-      
-      ArrayList<ArrayList<TaskMessage>> bundle = bundles.get(key);
-      if (null == bundle) {
-        bundle = new ArrayList<ArrayList<TaskMessage>>();
-        bundles.put(key, bundle);
-      }
-      
-      ArrayList tupleSet = workerTupleSetMap.get(key);
-      if (null != tupleSet && tupleSet.size() > 0) {
-        bundle.add(tupleSet);
-      }
-    } 
+  private ArrayList<TaskMessage> buffer = new ArrayList<TaskMessage>();
+
+  public void addAll(ArrayList<TaskMessage> tuples) {
+    buffer.addAll(tuples);
   }
   
-  public void send(HashMap<String, IConnection> connections) {
-    for (String hostPort : bundles.keySet()) {
+  public void send(HashMap<Integer, String> taskToHostPort, HashMap<String, IConnection> connections) {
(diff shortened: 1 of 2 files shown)


storm-core/src/jvm/backtype/storm/utils/TransferDrainer.java
```diff
@@ -26,88 +26,47 @@
...
-      
-      ArrayList tupleSet = workerTupleSetMap.get(key);
-      if (null != tupleSet && tupleSet.size() > 0) {
-        bundle.add(tupleSet);
-      }
-    } 
+  private ArrayList<TaskMessage> buffer = new ArrayList<TaskMessage>();
+
+  public void addAll(ArrayList<TaskMessage> tuples) {
+    buffer.addAll(tuples);
```
d2r: I know you mentioned this in a comment on the former pull request, but since this is just a list do you want to remove `buffer`, `addAll()`, and `clear()` from this class and instead pass an ArrayList to the `send` method?

HeartSaVioR: @d2r 
Anything is fine with me. 
We may introduce new buffer to outer side, but if it feels clearer I'll be happy to do it.

d2r: Yeah, I liked your suggestion of having the new buffer.  Before TransferDrainer was added, the buffer was just an ArrayList anyway.  I do not think it makes sense to have an ArrayList "inside" the TransferDrainer, because then we are implementing wrapper methods for `addAll` and `clear` that do not add value.

HeartSaVioR:
@d2r Changed to have a buffer outer side of TransferDrainer.

d2r:
FYI: unit tests passed for me.
I plan to take a look at performance using [storm-perf-test](https://github.com/yahoo/storm-perf-test)

ptgoetz:
So far I think I'm seeing a performance regression in terms of throughput, but I want to let all the test cases run to be sure. I will have more information tomorrow.

HeartSaVioR:
I'm also seeing a performance regression. IMO resolving issue could introduce performance drop slightly, but shouldn't too much.

HeartSaVioR:
Since current approach do more works within read lock, it may be necessary to test performance with previous approach, too.

ptgoetz:
I'm definitely seeing a performance regression. In a core storm topology I'm seeing a drop in throughput and slight increase in latency. In a trident topology I'm seeing a significant increase in latency (2x).

I'll try testing with #521.

ptgoetz:
Testing with #521 applied to 0.10.x-branch I'm actually seeing a performance ***improvement***.

With core storm topologies there's an increase in throughput and a small reduction in latency. With trident, throughput is the same and latency is slightly lower.

I'll run my fault tolerance tests next to see if there's any change in behavior

ptgoetz:
@HeartSaVioR,
#521 passed my fault tolerance test (which randomly kills workers and tests for data loss).

I'd suggest closing this pull request, and reopening #521 to address @d2r's comments there.

HeartSaVioR:
@ptgoetz Thanks for testing! I'll move on to #521.

PR #558, opened 2015-05-19
STORM-827: Allow AutoTGT to work with storm-hdfs too.

revans2:


Parth-Brahmbhatt:
+1. I do not have a secure hdfs setup right now so if you can confirm that you were able to test this with AutoTGT that would be great.

revans2:
@Parth-Brahmbhatt I did run with this on one of our secure clusters.  It was not exactly the same code, as our version is older and does not have the AutoHDFS in it.  Which is why I originally filed the JIRA incorrectly and said it should support that too.

PR #564, opened 2015-05-27
Remove unused clj-time dependency from storm-core

emidln:
clj-time was formerly used by UI, but is no longer. This removes
clj-time (and thus joda-time 2.0). This might be a breaking change
if a user assumes joda-time 2.0 or clj-time 0.8.0 is provided by
the worker on the classpath.


HeartSaVioR:
LGTM.

emidln:
Ugh. Withdrawing this. If we don't pin clj-time 0.8.0, we get 0.3.7 from ring-core.

HeartSaVioR:
Oh, I didn't check about dependencies changeset. Sorry!

PR #565, opened 2015-05-29
STORM-839

eshioji:
This fixes the reported deadlock between `disruptor-worker-transfer-queue` thread and `client-worker` thread, which seem to have been introduced by STORM-329.

After reviewing the `v0.9.4` code, my conclusion was that the background flushing task can be removed without real change in the behavior. By doing this, the synchronization that was involved in the deadlock can be removed altogether.

My reasoning is as follows:
 - One has three option to deal with Netty's buffer filling up:
  1. Discard incoming new messages
  2. Block client thread until there is space (back pressure)
  3. Keep buffering up until OOME is thrown
 - My guess is that the `v0.9.4` code attempted to implement option (i), but actually the behavior is option (iii). When Netty's `Channel.isWritable` returns false in `Client.send`, the thread avoids writing to the `Channel` and leaves messages in the `Channel.messageBatch` field, which the background task flushes when the `Channel.isWritable` starts to return true again. However, if `Client.send` is called again before this (which should be common), the content of `Channel.messageBatch` is written and buffered on the `Channel` anyways, because `flushMessages` does not check `Channel.isWritable`.
 - AFAIU we like option (ii), but this requires more work. Between option (i) and option (iii), IMO (iii) is superior because if OOME happens, the user can reduce MAX_PENDING_TUPLES to prevent this. Discarding messages would be harder to diagnose.
 - If we are content with option (ii), we can remove the background flushing task and related synchronization altogether, removing the source of deadlock. We'd simply write and buffer onto the unbounded buffer of `Channel`, and if there are too many pending messages, the worker will die of OOME, and the user should reduce this with indirect means like reducing MAX_PENDING_TUPLES until option (ii) is implemented in the future.

Thoughts @miguno ?


eshioji:
Argh sorry I should have opened this against Storm v0.9.4, I worked off v0.9.4 tag.

PR #566, opened 2015-05-29
[STORM-839] Deadlock hazard in backtype.storm.messaging.netty.Client

eshioji:
(I accidentally did a PR against master with the same content, please ignore that one)

This fixes the reported deadlock between `disruptor-worker-transfer-queue` thread and `client-worker` thread, which seem to have been introduced by STORM-329.

After reviewing the `v0.9.4` code, my conclusion was that the background flushing task can be removed without real change in the behavior. By doing this, the synchronization that was involved in the deadlock can be removed altogether.

My reasoning is as follows:
 - One has three option to deal with Netty's buffer filling up:
  1. Discard incoming new messages
  2. Block client thread until there is space (back pressure)
  3. Keep buffering up until OOME is thrown
 - My guess is that the `v0.9.4` code attempted to implement option (i), but actually the behavior is option (iii). When Netty's `Channel.isWritable` returns false in `Client.send`, the thread avoids writing to the `Channel` and leaves messages in the `Channel.messageBatch` field, which the background task flushes when the `Channel.isWritable` starts to return true again. However, if `Client.send` is called again before this (which should be common), the content of `Channel.messageBatch` is written and buffered on the `Channel` anyways, because `flushMessages` does not check `Channel.isWritable`.
 - AFAIU we like option (ii), but this requires more work. Between option (i) and option (iii), IMO (iii) is superior because if OOME happens, the user can reduce MAX_PENDING_TUPLES to prevent this. Discarding messages would be harder to diagnose.
 - If we are content with option (iii), we can remove the background flushing task and related synchronization altogether, removing the source of deadlock. We'd simply write and buffer onto the unbounded buffer of `Channel`, and if there are too many pending messages, the worker will die of OOME, and the user should reduce this with indirect means like reducing MAX_PENDING_TUPLES until option (ii) is implemented in the future.

Thoughts @miguno ?


eshioji:
@Hailei Thanks for the comment. Could you review my reasoning in the [PR comment above](https://github.com/apache/storm/pull/566#issue-82369808)?

Hailei:
>we can remove the background flushing task and related synchronization altogether, removing the source of deadlock. We'd simply write and buffer onto the unbounded buffer of Channel
+1 I agree with you.flushMessages don't check isWritable,So the last message batch check it is not useful.
Block client thread until there is space (back pressure) is better.It can snap out of OOM.

eshioji:
@Hailei I see. 

> Block client thread until there is space (back pressure) is better.It can snap out of OOM.

Agreed, but my understanding is that this needs quite a bit more work. At the least, the change can't be contained within `Client.java`.

miguno:
> My reasoning is as follows:
> 
> * One has three option to deal with Netty's buffer filling up:
>     1. Discard incoming new messages
>     2. Block client thread until there is space (back pressure)
>     3. Keep buffering up until OOME is thrown
>
> My guess is that the v0.9.4 code attempted to implement option (i), but actually the behavior is option (iii).

The code's intention was actually (iii).  As you described back pressure (ii) was not picked because this will require a significant amount of work, which was thus out of scope for fixing STORM-329.  The reason (iii) was preferred over (i) was also as you described -- if (and only if) users have enabled acking = guaranteed message processing for a topology, they can prevent OOM errors from happening by setting [`topology.max.spout.pending`](https://github.com/apache/storm/blob/master/conf/defaults.yaml#L180).

I'll have to look at your code in more detail before commenting.  I'll also ping @ptgoetz, @clockfly, @tedxia, and @revans2 who were involved in this significant patch (sorry for the spam, folks!).

eshioji:
Just as a heads up, I'm seeing STORM-763 as well (300 ERROR messages per second per machine) in our cluster. My initial gut feeling is that maybe reconnect isn't getting enough run time from the ScheduledExecutor. Depending on the results I might create another PR for 763.

eshioji:
Hi @miguno , thanks for taking the time. I think I have a [fix for STORM-763](https://github.com/apache/storm/pull/568) which builds on this PR. Let me know what you think!

PR #569, opened 2015-06-01
STORM-841 Correct doc that OutputCollector is not thread-safe

HeartSaVioR:
* Concepts.md: Correct document
* Troubleshooting.md: Add stack trace for other NPE case caused by same issue
  * Related to STORM-770


ptgoetz:
+1

harshach:
+1

PR #582, opened 2015-06-06
Feature autotuning for max spout pending

chuanlei:
Inspired by Twitter Sigmod 2014 paper “Storm@Twitter”
I use the number of ack to measure the throughout, and use the automata to tune the parameter of max-spout-pending automatically


revans2:
@chuanlei I like auto-tuning max spout pending, but I am confused what we are tuning it for.  The current code seems to increase the size of max-spout-pending if the number of fully acked tuples is increasing, and reduces it if the number appears to be decreasing.  Why do we bother with this at all?  If we just set the max spout pending to a large enough number we will get the same result.

The reason behind having a max-spout-pending is so we don't blow the java heap, by having too many messages in flight at any point in time, and balance the throughput with the latency.  If we have more messages in flight the latency will go up, but the throughput increases.  If we have fewer in flight the latency is much lower but the throughput is also lower.

I would really prefer to see a solution where a user can set a latency target, ideally 95th percentile or 99th percentile target and the spout adjusts the max-spout-pending to try and hit that latency target, with an upper bound on max-spout-pending so we don't overflow the heap.

HeartSaVioR:
@chuanlei 
We can get a hint from Heron paper that Twitter don't catch up latest Storm version.
Paper mentions ZeroMQ, but Storm changes default transport layer to Netty over one years ago.
It means that we should analyze that it still gives a benefit to Storm, and in my opinion you have to analyze it before posting pull request. 
Attaching performance test results is necessary.

d2r:
@chuanlei,

The source repository for this pull request is gone, and there is no Jira associated with it.
Would you create a Jira and a new pull request for this change? I will close this pull request.

PR #584, opened 2015-06-08
Feature add topology based scheduler

chuanlei:
New Scheduler based on the static information of Topology
Here we convert the schedule problem as a min-cut problem. i.e.
We should partition the graph into some clusters where the number of clusters is the number of workers.
Now I have implemented the algorithm of Karger's Min-Cut algorithm to solve this problem.
And if we take the runtime stats into account, this approach is natural to extend.
Maybe we can have another solution.


revans2:
@chuanlei This is a rather interesting feature.  It would be nice if you could file a JIRA for this.

d2r:
@chuanlei , it looks like the source repository has been deleted. Would you create a Jira for this work, and submit a new pull request?

PR #589, opened 2015-06-10
STORM-862: Pluggable System Metrics

revans2:


storm-core/src/jvm/backtype/storm/Config.java
```diff
@@ -1087,6 +1087,17 @@
     public static final String TOPOLOGY_METRICS_CONSUMER_REGISTER = "topology.metrics.consumer.register";
     public static final Object TOPOLOGY_METRICS_CONSUMER_REGISTER_SCHEMA = ConfigValidation.MapsValidator;
 
+    /**
+     * A map of metric name to class name implementing IMetric that will be created once per worker JVM
+     */
+    public static final String TOPOLOGY_WORKER_METRICS = "topology.worker.metrics";
+    public static final Object TOPOLOGY_WORKER_METRICS_SCHEMA = Map.class;
+
+    /**
```
HeartSaVioR: Seems like TOPOLOGY_WORKER_METRICS and WORKER_METRICS provide the same things. 
(Comment is also same.)

Are semantics of two configurations different?

revans2: They do provide the same thing, but the intended audience is different.  Any config beginning with "topology." is intended for topology owners to set where as worker.metrics is intended for a cluster owner to set.  That way I can force all topologies to collect a specific metric, but still allow end users to easily add in more of their own metrics.

HeartSaVioR: I can see the benefit, thanks for the details. :)

HeartSaVioR:
LGTM.

storm-core/src/jvm/backtype/storm/Config.java
HeartSaVioR: It would be better to explain the intention from source code comment or configuration file, or doc, since users can't know that easily.

HeartSaVioR:
Now I can vote this PR. +1

PR #590, opened 2015-06-10
STORM-860 disable "Activate" button when topology status is not "inactivate"

HeartSaVioR:
When we activate topology while it is rebalancing or killed from UI, UI shows that "Error while communicating to nimbus".
Based on these, "Activate" only makes sense while inactivated.


kishorvpatil:
+1. Good catch.

kishorvpatil:
Thank you @HeartSaVioR . This change is merged to master now.

HeartSaVioR:
@kishorvpatil Thanks for merging!

PR #592, opened 2015-06-12
STORM-866: Use storm.log.dir instead of storm.home in log4j2 config

tedxia:
PR for [STORM-866](https://issues.apache.org/jira/browse/STORM-866)

The config log4j2 use storm.home as log dir, but not storm.log.dir, we should use storm.log.dir instead.

### DIFF
--- a/log4j2/cluster.xml
+++ b/log4j2/cluster.xml
@@ -23,8 +23,8 @@
 </properties>
 <appenders>
     <RollingFile name="A1"
-                 fileName="${sys:storm.log.dir}/logs/${sys:logfile.name}"
-                 filePattern="${sys:storm.log.dir}/logs/${sys:logfile.name}.%i">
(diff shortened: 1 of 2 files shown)


HeartSaVioR:
It's better to remove following "/logs" from all of your changeset since storm.log.dir already points that directory.

tedxia:
@HeartSaVioR Thanks for your advice, I just committed another patch.

HeartSaVioR:
@tedxia LGTM. Great for finding missed spot!

kishorvpatil:
LGTM. +1

HeartSaVioR:
I'd like to include this into 0.10.x-branch since it's a bug, and we're having a chance to fix it before releasing stable version.

HeartSaVioR:
Merged into 0.10.x-branch, too.

PR #595, opened 2015-06-15
STORM-867: fix bug with mk-ssl-connector

tedxia:
PR for [STORM-867](https://issues.apache.org/jira/browse/STORM-867)

There are two error about mk-ssl-connector in storm-core/src/clj/backtype/storm/ui/helpers.clj
1. set trustStore password with setTrustStoreType
2. need-client-auth and want-client-auth are boolen, but use as function


HeartSaVioR:
LGTM.

harshach:
+1

PR #597, opened 2015-06-18
[STORM-874] Add UncaughtExceptionHandler for netty threads

kishorvpatil:
The netty threads need to handle uncaught exceptions. This would help ensure that JVM exits if netty threads hit OOM Error  while allocating new buffer.

### DIFF
--- a/storm-core/src/jvm/backtype/storm/messaging/netty/NettyUncaughtExceptionHandler.java
+++ b/storm-core/src/jvm/backtype/storm/messaging/netty/NettyUncaughtExceptionHandler.java
@@ -1,0 +1,35 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package backtype.storm.messaging.netty;
+
+import backtype.storm.utils.Utils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class NettyUncaughtExceptionHandler implements Thread.UncaughtExceptionHandler {
+  private static final Logger LOG = LoggerFactory.getLogger(NettyUncaughtExceptionHandler.class);
+  @Override
+  public void uncaughtException(Thread t, Throwable e) {
+    try {
+      Utils.handleUncaughtException(e);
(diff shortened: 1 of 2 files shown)


harshach:
@kishorvpatil would it makes sense to log the exception as well? , allows users to know the cause of jvm exit.

storm-core/src/jvm/backtype/storm/messaging/netty/NettyUncaughtExceptionHandler.java
```diff
@@ -0,0 +1,35 @@
...
+import backtype.storm.utils.Utils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class NettyUncaughtExceptionHandler implements Thread.UncaughtExceptionHandler {
+  private static final Logger LOG = LoggerFactory.getLogger(NettyUncaughtExceptionHandler.class);
+  @Override
+  public void uncaughtException(Thread t, Throwable e) {
+    try {
+      Utils.handleUncaughtException(e);
```
HeartSaVioR: @kishorvpatil 
Utils.handleUncaughtException() only takes care of Error, and any kinds of Exception will be silently ignored without leaving any logs. 
Is it intentional? Or do we need to take care?

eshioji:
If the thread is used in `ThreadPoolExecutor` (like the boss & worker thread pool currently used), `UncaughtExceptionHandler` are not invoked because `ThreadPoolExecutor` suppress uncaught exceptions coming from tasks (see e.g. http://stackoverflow.com/questions/1838923/why-is-uncaughtexceptionhandler-not-called-by-executorservice). There are workarounds but they are quite hacky.

Maybe it makes more sense to consider the service unhealthy if messages are unable to be dispatched, rather than based on whether an uncaught exception was encountered?

kishorvpatil:
@eshioji We are catching `Error` not `RunTimeException` which shows more of a JVM  resource issue - hence we do exit from the JVM - its more serious than just considering some service being down.

storm-core/src/jvm/backtype/storm/messaging/netty/NettyUncaughtExceptionHandler.java
kishorvpatil: @HeartSaVioR  We consider the Netty message handlers should take care of all those exceptions. We leave JVM level Error to be handled at executorservice handler level.

kishorvpatil:
@harshach Sorry for so long in responding to it. The `Utils.handleUncaughtException` writes ot STDOUT to avoid further major exception by attempting to log exceptions. But otherwise we log them.

storm-core/src/jvm/backtype/storm/messaging/netty/NettyUncaughtExceptionHandler.java
d2r: I agree with @HeartSaVioR , we should at least log any (unchecked) Exception instead of swallow it.

d2r: Also, the name Utils/handleUncaughtException here implies we are handling exceptions that could not be otherwise be dealt with, and so this is the end of the road.  It does not seem that we should return from this handler expecting some other code to deal with the situation, but we should bring down the JVM or throw.  If the code is not intended to be used this way, then we should rename and/or document what is expected of the caller to do with `t` after we return.

kishorvpatil:
These changes went in as part of #838. So closing the PR and JIRA

PR #599, opened 2015-06-19
Update pom.xml

vesense:
add flux-wrappers scope

### DIFF
--- a/external/flux/flux-core/pom.xml
+++ b/external/flux/flux-core/pom.xml
@@ -35,6 +35,7 @@
             <groupId>org.apache.storm</groupId>
             <artifactId>flux-wrappers</artifactId>
             <version>${project.version}</version>
+            <scope>test</scope>
         </dependency>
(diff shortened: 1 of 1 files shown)


d2r:
@vesense, could you create a Jira issue for this change and add it to the title of this pull request?

What problem is the current scope causing here that needs to be fixed?

ptgoetz:
-1 flux-wrappers add convenience multi-lang spout/bolt implementations. Arguably, they could be included in flux-core. I'd rather go that route than set flux-wrappers to test scope, which will make them unavailable for people who want them.

vesense:
Sorry I took so long to respond. It is convenient for multi-lang user to use FLUX just in one jar.
As a minor issue, now, It is OK for me.
In the past, I wanted to change the flux-wrappers scope because users using java did not need flux-wrappers. 
I will close this PR. Thanks @d2r @ptgoetz

PR #601, opened 2015-06-23
STORM-917 : Ability to emit from the spout to a specific stream, by default it will emit to the default stream.

anishek:
STORM-917: This is to allow kafka spout to emit to a specific stream part from the default stream.

### DIFF
--- a/external/storm-kafka/src/jvm/storm/kafka/PartitionManager.java
+++ b/external/storm-kafka/src/jvm/storm/kafka/PartitionManager.java
@@ -23,6 +23,7 @@ import backtype.storm.metric.api.CountMetric;
 import backtype.storm.metric.api.MeanReducer;
 import backtype.storm.metric.api.ReducedMetric;
 import backtype.storm.spout.SpoutOutputCollector;
+import backtype.storm.utils.Utils;
 import com.google.common.collect.ImmutableMap;
 import kafka.javaapi.consumer.SimpleConsumer;
 import kafka.javaapi.message.ByteBufferMessageSet;
@@ -125,8 +126,12 @@ public class PartitionManager {
         return ret;
     }
 
-    //returns false if it's reached the end of current batch
     public EmitState next(SpoutOutputCollector collector) {
+        return this.next(collector, Utils.DEFAULT_STREAM_ID);
+    }
+
+    //returns false if it's reached the end of current batch
+    public EmitState next(SpoutOutputCollector collector, String emitStreamId) {
         if (_waitingToEmit.isEmpty()) {
             fill();
         }
@@ -138,7 +143,7 @@ public class PartitionManager {
             Iterable<List<Object>> tups = KafkaUtils.generateTuples(_spoutConfig, toEmit.msg);
             if (tups != null) {
                 for (List<Object> tup : tups) {
-                    collector.emit(tup, new KafkaMessageId(_partition, toEmit.offset));
+                    collector.emit(emitStreamId, tup, new KafkaMessageId(_partition, toEmit.offset));
                 }
                 break;
             } else {
--- a/external/storm-kafka/src/jvm/storm/kafka/PartitionManager.java
+++ b/external/storm-kafka/src/jvm/storm/kafka/PartitionManager.java
@@ -126,10 +126,6 @@ public class PartitionManager {
         return ret;
     }
 
-    public EmitState next(SpoutOutputCollector collector) {
-        return this.next(collector, Utils.DEFAULT_STREAM_ID);
-    }
-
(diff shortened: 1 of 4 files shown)


HeartSaVioR:
I like the idea.
Could you file issue to JIRA so that we can record changeset into CHANGELOG?

anishek:
https://issues.apache.org/jira/browse/STORM-917

HeartSaVioR:
And please modify PR title to include "STORM-917" as prefix.
Thanks in advance!

anishek:
done hope this helps.

harshach:
@anishek can you also update the README.md with doc on this config.

external/storm-kafka/src/jvm/storm/kafka/SpoutConfig.java
```diff
@@ -29,6 +29,7 @@
 
     // setting for how often to save the current kafka offset to ZooKeeper
     public long stateUpdateIntervalMs = 2000;
+    public String emitStreamId;
```
harshach: is it a mandatory config now? . What if users doesn't pass a emitStreamId shouldn't we use the default.

anishek: you are right i didnt look at the code too closely, I had a overridden method in PartitionManager with the default stream. I thought anyone else using the old method would automatically go through that. Though thinking now i looked at all the usages and there is only one in the kafkaSpout so i will do another pull request with the changes.

sweetest:
This is a gread idea, though I haven't looked through entire code.

anishek:
All the necessary changes are in this request.  Please do let me know if anything else is required.

harshach:
@anishek can you please add this config to README.md https://github.com/apache/storm/blob/master/external/storm-kafka/README.md 
few lines about what it does.

anishek:
@harshach  I did add the line with the comment in the https://github.com/apache/storm/blob/master/external/storm-kafka/README.md  stating what it does under the set of configurations for the spoutConfig. Do you want a separate section  or a separate line in the read me explaining it ?

harshach:
@anishek my bad missed it.

lujinhong:
what about trident?

abhishekagarwal87:
Thanks for doing this Anishek. My comments
Kafka spout should declare the stream id in output declaration. 
```
declarer.declare(streamID, _spoutConfig.scheme.getOutputFields()
```
I would prefer the streamID to come from the scheme. More than often the stream on which data is to be sent to, depends on the input data. Since scheme knows about the data, it can also determine the stream. This makes more sense when you think of multiple output streams from same spout. The only downside I see is that the scheme interface will change and that might translate into compatibility issues. Nonetheless this PR is a good start.

external/storm-kafka/src/jvm/storm/kafka/PartitionManager.java
```diff
@@ -126,7 +127,7 @@ public Map getMetricsDataMap() {
     }
 
     //returns false if it's reached the end of current batch
-    public EmitState next(SpoutOutputCollector collector) {
+    public EmitState next(SpoutOutputCollector collector, String emitStreamId) {
```
abhishekagarwal87: what is the need of additional argument since spoutConfig is available to this class. stream name can be initialized once in the prepare phase itself.

external/storm-kafka/src/jvm/storm/kafka/SpoutConfig.java
```diff
@@ -17,6 +17,8 @@
  */
 package storm.kafka;
 
+import backtype.storm.utils.Utils;
```
abhishekagarwal87: You need to upmerge your master branch.

anishek: Upmerge ? i thought one of you would pull in the request and merge it.

anishek:
@lujinhong - i have not worked with trident and hence don't know where to make the change.

@abhishekagarwal87  -- this was a quick fix i did for a POC under 30 mins, It might not be the right way to go about this.

external/storm-kafka/src/jvm/storm/kafka/SpoutConfig.java
abhishekagarwal87: One of the committers will do that. But for that to happen, the PR should be ready for merge. Currently it has conflicts with master branch.

d2r:
> this was a quick fix i did for a POC under 30 mins, It might not be the right way to go about this.

@anishek This pull request has gone stale. Do we want to continue with this proposed change or close this pull request?

PR #602, opened 2015-06-23
STORM-871 Change multilang heartbeat mechanism

HeartSaVioR:
* Introduce new way of treating heartbeat
  * subprocess has to update pid file's modified time periodically
    * in default implementation we update pid file every 1 sec
  * ShellSpout / ShellBolt checks pid file's modified time every 1 sec
  * No more heartbeat tuple

It may introduces language constraints, so we need to gather opinions from user/dev mailing list to make sure we can safely change mechanism.

### DIFF
--- a/storm-core/test/clj/backtype/storm/multilang_test.clj
+++ b/storm-core/test/clj/backtype/storm/multilang_test.clj
@@ -45,9 +45,9 @@
                   {"2" (thrift/mk-shell-bolt-spec {"1" :shuffle} [executor (str "tester_bolt." file-extension)] ["word"] :parallelism-hint 1)})]
          (submit-local-topology nimbus
                "test"
-               {TOPOLOGY-WORKERS 20 TOPOLOGY-MESSAGE-TIMEOUT-SECS 3 TOPOLOGY-DEBUG true}
+               {TOPOLOGY-WORKERS 20 TOPOLOGY-MESSAGE-TIMEOUT-SECS 3 TOPOLOGY-DEBUG true SUPERVISOR-WORKER-TIMEOUT-SECS 3}
                topology)
-       (Thread/sleep 10000)
+       (Thread/sleep 20000)
        (.killTopology nimbus "test")
        (Thread/sleep 10000)
        )))
--- a/storm-multilang/python/src/main/resources/resources/storm.py
+++ b/storm-multilang/python/src/main/resources/resources/storm.py
@@ -20,6 +20,7 @@ import sys
 import os
 import traceback
 from collections import deque
+import threading
 
 try:
     import simplejson as json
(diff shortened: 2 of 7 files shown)


HeartSaVioR:
Gathering opinions: http://mail-archives.apache.org/mod_mbox/storm-user/201506.mbox/%3CCAF5108iiB0_%2By_MzReVUF7LWDsQgyi9jFK1QNPAupYs6_xj-jw%40mail.gmail.com%3E

HeartSaVioR:
We need to change document when we apply this and release.
http://storm.apache.org/documentation/Multilang-protocol.html

storm-core/test/clj/backtype/storm/multilang_test.clj
```diff
@@ -45,9 +45,9 @@
                   {"2" (thrift/mk-shell-bolt-spec {"1" :shuffle} [executor (str "tester_bolt." file-extension)] ["word"] :parallelism-hint 1)})]
          (submit-local-topology nimbus
                "test"
-               {TOPOLOGY-WORKERS 20 TOPOLOGY-MESSAGE-TIMEOUT-SECS 3 TOPOLOGY-DEBUG true}
+               {TOPOLOGY-WORKERS 20 TOPOLOGY-MESSAGE-TIMEOUT-SECS 3 TOPOLOGY-DEBUG true SUPERVISOR-WORKER-TIMEOUT-SECS 3}
                topology)
-       (Thread/sleep 10000)
+       (Thread/sleep 20000)
```
HeartSaVioR: Adjust option regarding subprocess heartbeat and make test longer for improving test more accurate.

storm-multilang/python/src/main/resources/resources/storm.py
```diff
@@ -29,6 +30,26 @@
 json_encode = lambda x: json.dumps(x)
 json_decode = lambda x: json.loads(x)
 
+# scheduled timer running job at fixed rate (note: GIL!)
```
dan-blanchard: As you note, the GIL is going to be a major problem with this approach.  To the point, that this will not work as intended and you will still have timeout issues if processing an individual tuple takes too long.

HeartSaVioR: Thanks @dan-blanchard ,
I experimented about python's GIL just now, and with python 2.7.6 in OSX I found that other thread can hold CPU more than 1 sec when timer is expired at that time. 
https://gist.github.com/HeartSaVioR/34d90cdd6af906e72935

Actually I wasn't affected this issue during I was working with Python cause it was I/O intensive job, and seems like it isn't same to CPU intensive job.

Default tick time is somewhat very long. I found one document which says tick time is about ~6.5 secs, which doesn't meet our requirement.

GIL is a major problem, and maybe I should check it with Ruby.

HeartSaVioR:
After discussed with @dan-blanchard , I gave up introducing new multilang heartbeat mechanism.
It would be new requirement to storm multilang libraries, which library developers already struggled with current heartbeat mechanism.
Above all, I can't think about overcoming GIL.

I'll close this issue, but to stick with current mechanism STORM-742 should be merged.

PR #603, opened 2015-06-25
STORM-910: Submit job failed when submitter user is also set in nimbus.supervisor.users

tedxia:
PR for [STORM-910](https://issues.apache.org/jira/browse/STORM-910)

When topology submitter user also set in nimbus.supervisor.users, submit topology will failed.


revans2:
This was actually done on purpose.  If you have run-as-user enabled and you run a topology as the supervisor user, that topology now has the ability to launch things as other users.  We put it here so users would get an error early.  If you want to run this way I really would like to see an explicit configuration to enable it that is off by default.

tedxia:
Thanks for the explain, I will close this patch.

PR #605, opened 2015-06-26
STORM-913:Use Curator's delete().deletingChildrenIfNeeded() instead of zk/delete-recursive

caofangkun:


pom.xml
```diff
@@ -211,7 +211,7 @@
         <guava.version>16.0.1</guava.version>
         <netty.version>3.9.0.Final</netty.version>
         <log4j-over-slf4j.version>1.6.6</log4j-over-slf4j.version>
-        <log4j.version>2.1</log4j.version>
+        <log4j.version>2.3</log4j.version>
```
HeartSaVioR: Seems like it is unneeded change.

caofangkun: Thank you @HeartSaVioR  
I have removed this unnessary change.

storm-core/src/clj/backtype/storm/zookeeper.clj
```diff
@@ -103,11 +103,14 @@
      (catch Exception e (throw (wrap-in-runtime e))))))
 
 (defnk delete-node
```
HeartSaVioR: It silently does nothing when path is not exist, but it was throwing NoNodeException when force is false, and default is false.
Only delete-recursive calls delete-node with force = true, and any other usages are using default.

In short, it doesn't show same behavior between two.

HeartSaVioR:
There're many unrelated commits. Could you get rid of these things?

storm-core/src/clj/backtype/storm/zookeeper.clj
HeartSaVioR: Oh, I got it. Whole usages are calling zk/delete-recursive so your changeset makes sense. My mistake.

caofangkun:
@HeartSaVioR 
Thank you for your advise. I have removed unreleated commits and pushed as https://github.com/apache/storm/pull/615/

PR #606, opened 2015-06-26
STORM-914: AutoHDFS should also update hdfs.config into Configuration

tedxia:
PR for [STORM-914](https://issues.apache.org/jira/browse/STORM-914)

When storm user use a specify hdfs cluster, he will put some config such as namenode.host into hdfs.config when submit storm topology, and when HdfsBolt create HdfsConfiguration, it will use hdfs.config update HDFSConfiguration, then HDFSBolt will visit the right hdfs cluster. But in AutoHDFS's renew and getHadoopCredentials, Configuration not for the HDFS that storm user specify, we should update all the Configuration with hdfs.config.

### DIFF
--- a/external/storm-hdfs/src/test/java/org/apache/storm/hdfs/trident/TridentSequenceTopology.java
+++ b/external/storm-hdfs/src/test/java/org/apache/storm/hdfs/trident/TridentSequenceTopology.java
@@ -24,6 +24,7 @@ import backtype.storm.generated.StormTopology;
 import backtype.storm.tuple.Fields;
 import backtype.storm.tuple.Values;
 import org.apache.storm.hdfs.common.rotation.MoveFileAction;
+import org.apache.storm.hdfs.common.security.AutoHDFS;
 import org.apache.storm.hdfs.trident.format.*;
 import org.apache.storm.hdfs.trident.rotation.FileRotationPolicy;
 import org.apache.storm.hdfs.trident.rotation.FileSizeRotationPolicy;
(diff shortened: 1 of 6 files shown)


tedxia:
@revans2 Will you take a look at this, thanks a lot.

external/storm-hdfs/src/test/java/org/apache/storm/hdfs/trident/TridentSequenceTopology.java
```diff
@@ -60,7 +61,7 @@ public static StormTopology buildTopology(String hdfsUrl){
                 .withSequenceFormat(new DefaultSequenceFormat("key", "sentence"))
                 .withRotationPolicy(rotationPolicy)
                 .withFsUrl(hdfsUrl)
-                .withConfigKey("hdfs.config")
```
caofangkun: @tedxia 
if you change ```hdfs.config``` to ```topology.hdfs.config``` 
should the following several files  change too ?
[HdfsFileTopology.java](https://github.com/apache/storm/blob/master/external/storm-hdfs/src/test/java/org/apache/storm/hdfs/bolt/HdfsFileTopology.java#L82)
[simple_hdfs.yaml](https://github.com/apache/storm/blob/master/external/flux/flux-examples/src/main/resources/simple_hdfs.yaml#L74)
[hdfs_test.yaml](https://github.com/apache/storm/blob/master/external/flux/flux-core/src/test/resources/configs/hdfs_test.yaml#L82)

tedxia: yes，I think we should change to hdfs.config for this reason, I will add a patch later.

d2r:
@tedxia This PR is rather stale. Is it still being worked or should we close it?

PR #607, opened 2015-06-30
STORM-799 Use IErrorReport interface more broadly

dossett:


revans2:
+1.  The only possible issue is if someone is implementing their own OutputCollector, possibly for running storm topologies on a different runtime. And I am not too concerned about breaking them.

PR #609, opened 2015-06-30
STORM-944 storm-hive: Fix apache calcite dependency conflict

dossett:
Hive 0.14 depends on apache calcite 0.9.2-incubating-SNAPSHOT, which causes a conflict with the 0.9.2-incubating dependency further down in this pom.  Excluding it from the Hive dependencies resolves a minor conflict.

### DIFF
--- a/external/storm-hive/pom.xml
+++ b/external/storm-hive/pom.xml
@@ -53,6 +53,14 @@
           <groupId>org.slf4j</groupId>
           <artifactId>slf4j-log4j12</artifactId>
         </exclusion>
+        <exclusion>
+          <groupId>org.apache.calcite</groupId>
+          <artifactId>calcite-core</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>org.apache.calcite</groupId>
+          <artifactId>calcite-avatica</artifactId>
+        </exclusion>
       </exclusions>
     </dependency>
 
@@ -65,6 +73,14 @@
           <groupId>org.slf4j</groupId>
           <artifactId>slf4j-log4j12</artifactId>
         </exclusion>
+        <exclusion>
+          <groupId>org.apache.calcite</groupId>
+          <artifactId>calcite-avatica</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>org.apache.calcite</groupId>
+          <artifactId>calcite-core</artifactId>
+        </exclusion>
       </exclusions>
     </dependency>
     <dependency>
@@ -75,6 +91,14 @@
         <exclusion>
           <groupId>org.slf4j</groupId>
           <artifactId>slf4j-log4j12</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>org.apache.calcite</groupId>
+          <artifactId>calcite-core</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>org.apache.calcite</groupId>
+          <artifactId>calcite-avatica</artifactId>
         </exclusion>
       </exclusions>
     </dependency>


harshach:
@dossett will this work with Hive 0.13?

dossett:
I switched hive.version to 0.13.0 but got some compile errors for unrelated reasons (below).  Basically in 0.13 HiveEndPoint doesn't have any methods that directly take a UGI, that I can tell.

That issue aside, Hive 0.13 doesn't appear to have a direct dependency on calcite, so this exclusion should not affect .13?

```
external/storm-hive/src/main/java/org/apache/storm/hive/common/HiveWriter.java:[229,40] no suitable method found for newConnection(boolean,<nulltype>,org.apache.hadoop.security.UserGroupInformation)
[ERROR] method org.apache.hive.hcatalog.streaming.HiveEndPoint.newConnection(java.lang.String,boolean,org.apache.hadoop.hive.conf.HiveConf) is not applicable
[ERROR] (actual argument boolean cannot be converted to java.lang.String by method invocation conversion)
[ERROR] method org.apache.hive.hcatalog.streaming.HiveEndPoint.newConnection(boolean,org.apache.hadoop.hive.conf.HiveConf) is not applicable
[ERROR] (actual and formal argument lists differ in length)
[ERROR] method org.apache.hive.hcatalog.streaming.HiveEndPoint.newConnection(boolean) is not applicable
[ERROR] (actual and formal argument lists differ in length)
```

HeartSaVioR:
There're some issues regarding using SNAPSHOT version of dependencies.

- Tez: https://issues.apache.org/jira/browse/HIVE-8845
- Calcite: https://issues.apache.org/jira/browse/HIVE-8873

For now, without this PR, builds are failed cause it tries to fetch SNAPSHOT version of calcite but nexus.codehaus.org is unreachable.
Behavior of depending on SNAPSHOT is fixed at 1.0.0.

I didn't have time to use Hive, so I have some questions, 

- Is it safe for users to upgrade hive version to 1.x.y without compatibility issues?
- If it is, is it better to notify users to use Hive version to 1.0.0 and higher while using storm-hive?

IMO we would be better to try avoiding any workarounds unless we can't resolve it without workaround. 

@dossett @caofangkun @harshach @ptgoetz @revans2

dossett:
I think this PR is preferable to stating a requirement for Hive 1.0.0 (of course, as the author of the PR, I would think that :-) )

- I could be wrong, but I think the vast majority of Hive installs are not yet at 1.0.0 since HDP 2.2.X is at 0.14.0 and 2.3 has not had time to be widely adopted.  Other Hadoop distributions are slower to upgrade Hive than HDP.

- This PR doesn't try to change any functional behavior. It just prevents maven from trying to resolve a dependency that will fail anyway.

caofangkun:
1 This PR is outdated now and  should upmerge this with master
2  [ org.apache.calcite:calcite-core:0.9.2-incubating](https://github.com/apache/storm/blob/master/external/storm-hive/pom.xml#L83) also not necessary be brought in .
3 This PR only and should only resolve fails of  fetching SNAPSHOT version of calcite
   I do not think there's compatibility issue right now , and of course we could file another issue if necessary.

HeartSaVioR:
@dossett 
Makes sense.
Last thing I'd like to know is that current patch can work with Hive 1.x.y.
If it isn't, storm-hive is just for 0.14.0 since you showed storm-hive doesn't work with Hive 0.13.0.
It would be better to specify supporting Hive versions to README.

@caofangkun 
If it doesn't affect storm-hive, I also think it would be better to remove explicit dependency about calcite, since higher version of Hive could require higher version of calcite.
But as I stated earlier, I didn't use Hive so verification and confirmation from Committer Sponsors are required.

dossett:
@HeartSaVioR 
This PR works with hive.version=1.0.0   By "works" I mean that it compiles and passes all unit tests.

It does NOT work with hive.version=1.2.1 (the version of Hive that ships with HDP 2.3).  Hive 1.2.1 depends on Calcite 1.2.0 and several tests fail if it's excluded.  I could get past these first problems by updating the calcite version in storm-hive/pom.xml to 1.2.0.  However, one test still fails (see below).  I have not had time to investigate these further, but it looks like some Hive behavior changes unrelated to the calcite dependency issue.

A new JIRA for Hive 1.2.1 support seems appropriate for that issue.

```
Running org.apache.storm.hive.bolt.TestHiveBolt
Tests run: 9, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 15.488 sec <<< FAILURE! - in org.apache.storm.hive.bolt.TestHiveBolt
testWithTimeformat(org.apache.storm.hive.bolt.TestHiveBolt)  Time elapsed: 0.692 sec  <<< FAILURE!
org.mockito.exceptions.verification.WantedButNotInvoked: 
Wanted but not invoked:
collector.ack(
    source: null:1, stream: , id: {}, [100, test-123, null, null]
);
-> at org.apache.storm.hive.bolt.TestHiveBolt.testWithTimeformat(TestHiveBolt.java:235)

However, there were other interactions with this mock:
-> at backtype.storm.task.OutputCollector.reportError(OutputCollector.java:223)
-> at backtype.storm.task.OutputCollector.fail(OutputCollector.java:218)
-> at backtype.storm.task.OutputCollector.reportError(OutputCollector.java:223)
-> at backtype.storm.task.OutputCollector.fail(OutputCollector.java:218)

	at org.apache.storm.hive.bolt.TestHiveBolt.testWithTimeformat(TestHiveBolt.java:235)

Running org.apache.storm.hive.common.TestHiveWriter
Tests run: 3, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 2.387 sec - in org.apache.storm.hive.common.TestHiveWriter

Results :

Failed tests: 
  TestHiveBolt.testWithTimeformat:235 
Wanted but not invoked:
collector.ack(
    source: null:1, stream: , id: {}, [100, test-123, null, null]
);
-> at org.apache.storm.hive.bolt.TestHiveBolt.testWithTimeformat(TestHiveBolt.java:235)

However, there were other interactions with this mock:
-> at backtype.storm.task.OutputCollector.reportError(OutputCollector.java:223)
-> at backtype.storm.task.OutputCollector.fail(OutputCollector.java:218)
-> at backtype.storm.task.OutputCollector.reportError(OutputCollector.java:223)
-> at backtype.storm.task.OutputCollector.fail(OutputCollector.java:218)


Tests run: 12, Failures: 1, Errors: 0, Skipped: 0
```

HeartSaVioR:
For now, all Travis CI builds are failing due to SNAPSHOT issue on storm-hive.
It is already bothering us, so I think we should take this first, and discuss later improvements for this issue.
+1.

@caofangkun 
I'm sorry but removing calcite-core from dependency is minor at now. It would be better to be addressed again when we agree to check hive version compatibilities.

PR #612, opened 2015-07-02
add dotNet adapter.

ziyunhx:


ziyunhx:
the StormSimple.exe source is at https://github.com/ziyunhx/storm-net-adapter/tree/master/StormSimple. I can't find where it should be.

lazyval:
Code aside, I guess storing *compiled* binaries and libraries in git isn't quite a good idea.

HeartSaVioR:
@ziyunhx

Agreed to @lazyval. 

- Project setting files should be removed from repo, which includes PDB, etc. (Maybe csproj too?)
- If we can make DLLs from source code, DLLs should be removed from repo.

Btw, I wonder Storm community can maintain non-JVM .net codes (which is actually a project) within JVM project. It also adds/changes 10000+ lines to storm-core, which is big compared to origin storm-core.

Please note that multilang implementations in storm-core are simple enough to represent just one file.
They don't require additional libraries installed, too.

There're some advanced Storm multilang projects, for example, StreamParse(https://github.com/Parsely/streamparse). 
.Net multilang adapter could be better to be maintained like StreamParse, I mean, outer Storm project.
It seems to be more natural.

How about posting new feature to user / dev mailing list and gather a consensus that we would be better to include this?

ziyunhx:
Okay, I will review my code and post new feature to dev mailing list, Thank you!

PR #613, opened 2015-07-02
STORM-924:Set the file mode of the files included when packaging release packages

caofangkun:


lazyval:
:+1: 

For those who wonders what this octal numbers mean, there is a great [*humanizator*](http://permissions-calculator.org/decode/0644/).

HeartSaVioR:
I like the change. It would be better to set files' mode explicitly whenever it needs.

Only thing that I'm afraid of is unrelated commits, which are having no meaning for Apache Storm repo.
Could you handle it?

caofangkun:
@HeartSaVioR 
I have removed unrelated commits  and pushed  to https://github.com/apache/storm/pull/614

PR #615, opened 2015-07-03
STORM-913:Use Curator's delete().deletingChildrenIfNeeded() instead of zk/delete-recursive

caofangkun:


revans2:
@caofangkun I am a bit nervous about this change.  We are changing code that had the option to auto create parents or auto-delete parents to not have that option any more.  Yes this makes it simpler if you don't know if a directory was created or not, but it becomes a problem in the case of ACLs.  If we are running with security and we create a parent directory it may end up with the incorrect ACLs.

I would also want to see where each of these functions are called and see if they all have :force turned on or not.  If they do the parameter is unneeded.  If they do not I want to understand why they don't

vesense:
-0

caofangkun:
@revans2 
The whole usages  call delete-node with ```:force true```, that is the ```:force ``` parameter is unneeded at all.

revans2:
+1 I would check this in but my laptop Internet connection is really bad right now. I'll try later on to see if things get better but if someone else wants to that is fine with me

caofangkun:
@revans2 
Please don't forget  check this PR in  :）

PR #616, opened 2015-07-03
Storm 763/839 0.11.x

eshioji:
This is a port of PR #568 from 0.9.x to 0.11.x. It fixes STORM-839 (Deadlock) and STORM-763 (Establish Netty reconnects asynchronously and reduce verbosity of error logs)

@revans2 This is the PR for master. I ran the performance test against master's HEAD and the results were in-line with the results I got in #568 . I'll create a separate PR for 0.10.x

PR #617, opened 2015-07-03
Storm 763/839 0.10.x

eshioji:
This is a port of PR #568 from 0.9.x to 0.10.x. It fixes STORM-839 (Deadlock) and STORM-763 (Establish Netty reconnects asynchronously and reduce verbosity of error logs)

The content is equivalent to the PR for 0.11.x (#616)

@revans2 This is the PR for 0.10.x. I also ran the performance test for this and the results were again inline with what I got with 0.9.x. I get build failure on my machine but I get the same failure with 0.10.x HEAD, so I'm guessing it's unrelated to this change.


HeartSaVioR:
FYI, failed build is related to storm-kafka, which is not excluded from 0.10.x-branch.

storm-core/src/jvm/backtype/storm/messaging/netty/Client.java
```diff
@@ -59,20 +59,16 @@
  * - Connecting and reconnecting are performed asynchronously.
  *     - Note: The current implementation drops any messages that are being enqueued for sending if the connection to
  *       the remote destination is currently unavailable.
- * - A background flusher thread is run in the background.  It will, at fixed intervals, check for any pending messages
- *   (i.e. messages buffered in memory) and flush them to the remote destination iff background flushing is currently
- *   enabled.
  */
 public class Client extends ConnectionWithStatus implements IStatefulObject {
+    private static final long PENDING_MESSAGES_FLUSH_TIMEOUT_MS = 600000L;
```
knusbaum: This seems like an incredibly large timeout.

eshioji: Maybe, this value was inherited from the current code tho. Maybe @miguno can shed light on the rationale?

revans2: It is large, but it matches the code that was there before.  This is only used on a close, as a timeout when trying to be sure that we have sent all pending messages.  We should look at potentially lowering it, but I think that is for a different JIRA.

knusbaum: That's fine with me.

revans2:
+1 the code compiles and the tests all pass

storm-core/src/jvm/backtype/storm/messaging/netty/StormClientHandler.java
```diff
@@ -0,0 +1,46 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package backtype.storm.messaging.netty;
+
+import java.net.ConnectException;
+
+import org.jboss.netty.channel.*;
```
knusbaum: Can we not import * in the future.

PR #623, opened 2015-07-07
STORM-927: make storm jar wait for deployment to finish

DevInsanity:
This change makes the "storm jar" command on Windows wait for the topology to deploy before exiting.

Tested with a flux-based topology.

### DIFF
--- a/bin/storm.cmd
+++ b/bin/storm.cmd
@@ -79,6 +79,7 @@
 
     :done
     set storm-command-arguments=%args%
+    set STORM_DEBUG=1
   )
   
   if not defined STORM_LOG_FILE (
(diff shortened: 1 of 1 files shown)


caofangkun:
@DevInsanity 
Do your topology use [StormSubmitter.submitTopologyWithProgressBar](https://github.com/apache/storm/blob/master/storm-core/src/jvm/backtype/storm/StormSubmitter.java#L287) ?
It will wait & show progress for the topology to deploy before exiting

DevInsanity:
@caofangkun 
No, we're using a flux based topology and deploying to a remote cluster (via `storm.cmd jar project.jar org.apache.storm.flux.Flux --remote project.yaml --filter project.properties`).

The problem is that because storm.cmd launches the actual process via `start /b` then it doesn't wait for the topology to deploy before returning to the caller, so you can't get any error codes back from the deployer (for example, if the yaml is invalid).

We're actually invoking this from a maven antrun plugin task as part of build process. In that case, it's even worse, as in the case of an error, not only does maven still report a build successful (as the antrun plugin's exec task doesn't see an error code), but none of the output is captured either. With the change, the output from the deployment is captured by maven properly.

While we could specify the storm_debug option explicitly (which would still need the quotes added around the %JAVA% to work properly when java is installed under Program Files), the jar command doesn't appear to be intended to launch a background server, so should wait for it's action to finish before returning.

ppoulosk:
I'm don't understand why this change is necessary, and why we would want to enable DEBUG in this script for all users.

DevInsanity:
@ppoulosk Without the quotes, debug mode doesn't work if java is in a path with spaces. Debug mode is being forced for the jar subcommand only, since it makes it possible to use it from other scripts (as otherwise, the act of deploying the topology gets backgrounded and you can't check return values from it).

PR #525, opened 2015-04-16
[STORM-789] Send more topology context to Multi-Lang components via initial handshake

dan-blanchard:
This PR adds the following keys to the "context" JSON dictionary sent to multi-lang tasks as part of their handshake with ShellBolt/ShellSpout:

-  `componentid`: the component ID for this task, so you no longer have to look it up in `task->component` by `taskid`
-  `streams`: a list of all of the streams for this task
-  `stream->outputfields`:  a mapping from stream names to output fields for that stream
-  `stream->target->grouping`: a mapping from stream names to the targets for this task to the kind of grouping they use.
-  `source->stream->grouping`:  a mapping from the component IDs of the sources to their streams to their groupings.

Note on groupings:  groupings are represented as a dictionary containing a required `type` key (e.g., `"SHUFFLE"`) and an optional `fields` key (for fields groupings).

PR #624, opened 2015-07-07
[STORM-928] Add source->stream->fields mapping to multi-lang context

dan-blanchard:
This pull request adds a `source->stream->fields` mapping to the topology context sent to multi-lang components as part of the handshake.

It also clears up a few minor type issues in TopologyContext.java, and fixes the Multi-Lang protocol doc to note that the additions I made in #525 are in 0.10.0 and not 0.11.0.

### DIFF
--- a/docs/documentation/Multilang-protocol.md
+++ b/docs/documentation/Multilang-protocol.md
@@ -66,7 +66,7 @@ The initial handshake is the same for both types of shell components:
             "4": "example-bolt2"
         },
         "taskid": 3,
-        // Everything below this line is only available in Storm 0.11.0+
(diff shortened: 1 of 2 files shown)


HeartSaVioR:
LGTM, +1.

Btw, I don't quite know why builds are failed from messaging-test for now. Needs investigation.
But changeset is not related to this, so we can just ignore them.

HeartSaVioR:
Thanks @dan-blanchard , I merged into master and 0.10.x-branch at ASF git repos, but seems like Github mirror is out of sync.
I'll resolve issue when mirror keeps up ASF repo.

PR #625, opened 2015-07-09
[STORM-503] fix issue of high cpu usage when bolt idle

errordaiwa:
### Issue
[STORM-503](https://issues.apache.org/jira/browse/STORM-503#)
### Phenomenon
1. Run a topology with large num of executors.
2. When a small number of tuples, CPU usage becomes abnormally high.

### Reason
1. Most threads wait at

	```
com.lmax.disruptor.BlockingWaitStrategy.waitFor(long, com.lmax.disruptor.Sequence, com.lmax.disruptor.Sequence[], com.lmax.disruptor.SequenceBarrier, long, java.util.concurrent.TimeUnit)
```
2. When there are few tuple transfered, most of executors block on DisruptorQueue.consumeBatchWhenAvailable method.

	```
final long availableSequence = _barrier.waitFor(nextSequence, 10, TimeUnit.MILLISECONDS);
```
    When bolt num is large, since timeout is 10ms, there will be frequently switching thread context lead to abnormally high CPU utilization.
		
### Fix
1. Make wait timeout configurable,
	
### Test
1. ENV
	+ Storm 0.9.3
	+ CPU (E5620, 8 core, 2.40GHz)
        + Centos 6.5, kernel 3.10.5-3.el6.x86_64
	+ Word counter topology, change spout sleep time from 100ms to 10s and num of counter bolt from 12 to 500.
2. DisruptorQueue wait time 10ms
	+ 28% CPU usage, %13 user, %15 sys 
3. DisruptorQueue wait time 1000ms
 	+ 5% CPU usage, %3 user, %2 sys
4. Execution latency do not increase.

### DIFF
--- a/conf/defaults.yaml
+++ b/conf/defaults.yaml
@@ -200,5 +200,6 @@ topology.testing.always.try.serialize: false
 topology.classpath: null
 topology.environment: null
 topology.bolts.outgoing.overflow.buffer.enable: false
+topology.disruptor.wait.timeout.millis: 10
 
 dev.zookeeper.path: "/tmp/dev-storm-zookeeper"
--- a/storm-core/src/clj/backtype/storm/daemon/executor.clj
+++ b/storm-core/src/clj/backtype/storm/daemon/executor.clj
@@ -223,6 +223,7 @@
         batch-transfer->worker (disruptor/disruptor-queue
                                   (str "executor"  executor-id "-send-queue")
                                   (storm-conf TOPOLOGY-EXECUTOR-SEND-BUFFER-SIZE)
+                                  (storm-conf TOPOLOGY-DISRUPTOR-WAIT-TIMEOUT-MILLIS)
                                   :claim-strategy :single-threaded
                                   :wait-strategy (storm-conf TOPOLOGY-DISRUPTOR-WAIT-STRATEGY))
         ]
--- a/storm-core/src/clj/backtype/storm/daemon/worker.clj
+++ b/storm-core/src/clj/backtype/storm/daemon/worker.clj
@@ -153,6 +153,7 @@
        ;; TODO: this depends on the type of executor
        (map (fn [e] [e (disruptor/disruptor-queue (str "receive-queue" e)
                                                   (storm-conf TOPOLOGY-EXECUTOR-RECEIVE-BUFFER-SIZE)
+                                                  (storm-conf TOPOLOGY-DISRUPTOR-WAIT-TIMEOUT-MILLIS)
                                                   :wait-strategy (storm-conf TOPOLOGY-DISRUPTOR-WAIT-STRATEGY))]))
        (into {})
        ))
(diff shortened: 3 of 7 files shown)


HeartSaVioR:
@errordaiwa 
Could you change issue number to STORM-503 from PR? 
I just closed STORM-929 as duplicated.

errordaiwa:
@HeartSaVioR  done!

arunmahadevan:
See the comment @ storm-core/src/clj/backtype/storm/disruptor.clj

`;; :block strategy requires using a timeout on waitFor (implemented in DisruptorQueue), as sometimes the consumer stays blocked even when there's an item on the queue.`

So with this change the bolts might have to wait for up to one sec (in the worst case) before they see the tuple after its added to the queue.

errordaiwa:
@arunmahadevan you are right. But I think at least this value should be configurable to tradeoff latency VS CPU usage.

HeartSaVioR:
Letting users configure disruptor queue wait timeout makes sense for me. Better than origin PR.

HeartSaVioR:
Great, +1.

arunmahadevan:
After digging further I suspect the volatile `numWaiters` @ https://github.com/LMAX-Exchange/disruptor/blob/2.10.1/code/src/main/com/lmax/disruptor/BlockingWaitStrategy.java 
could be the culprit. If a volatile is updated by multiple threads we could be losing updates which would manifest as consumers not being waken up when there is an item in the queue.

@errordaiwa  lmax-disruptor 2.10.4 appear to have a fix for this issue. Can you try with this version and call the waitFor without any timeout ?

HeartSaVioR:
I've found some commits which seems to fix race conditions of BlockingWaitStrategy.

https://github.com/LMAX-Exchange/disruptor/commit/80744acd26767eb382c0b7214587c4d26891541f
https://github.com/LMAX-Exchange/disruptor/commit/e47dfc4ab61373289fa27d98d38d44bfddb4d5e4
https://github.com/LMAX-Exchange/disruptor/commit/8ee839119751156552d31390c0ff559598d954c4

It could be more.
I also think that we can apply lmax-disruptor to 2.10.4 before applying this PR, and check it resolves our issue.

errordaiwa:
@arunmahadevan  I open an issue [Race condition in 2.10.1 release](https://github.com/LMAX-Exchange/disruptor/issues/119) and the author of Disruptor queue confirm this fix. But I can't reproduce this bug using 2.10.1. Do you have any idea on how to reprodue it?

HeartSaVioR:
@errordaiwa @arunmahadevan 
We may create unit test to check,

- make sure that there's no available data from disruptor queue
- subscriber thread waits for new data from disruptor queue with wait time 1 sec
- other thread immediately (or less than 1 sec) publish new data to disruptor queue

Subscriber thread should awake immediately (not waiting whole wait time) when publishing.

errordaiwa:
@HeartSaVioR I do unit test for both ver2.10.1 and 2.10.4, but no latency was found. May be the condition to reproduce the bug is very strict. So should we apply this pr or upgrade version of disruptor queue?

HeartSaVioR:
@errordaiwa 
Did you test it with multi-threaded producers? 

Btw, I think we can take a benefit from adopting 2.10.4, since changeset contains some fixes for race.
https://github.com/LMAX-Exchange/disruptor/compare/2.10.1...2.10.4
It would be better to create new issue for changing Disruptor to 2.10.4 cause we should run some tests (unit tests / performance tests / etc.) so that we'll be sure about its safety.

errordaiwa:
@HeartSaVioR 
No, I test it with just one producer.

And here is the new issue [Storm-935](https://issues.apache.org/jira/browse/STORM-935).

arunmahadevan:
@errordaiwa with 2.10.4 you may try the single argument waitFor(long sequeuce) in DisruptrorQueue.java and observe the CPU usage with a high bolt count like what you mentioned in the issue. Also see if the bolts are receiving the tuples without a delay.

HeartSaVioR:
Since #630 contains #625 now, I think we can close this PR.
@errordaiwa Please close PR, thanks!

PR #626, opened 2015-07-09
Fix links to documentation in Javadoc

alexshadow007:


revans2:
@alexshadow007 thanks for the fix.  I pulled this into master.

revans2:
I forgot to formally say +1 so here it is.

PR #627, opened 2015-07-10
STORM-931: Python Scritps to Produce Formatted JIRA and GitHub Joint …

hmcl:
…Reports


harshach:
+1

revans2:
+1 looks a lot better then my original one.

PR #632, opened 2015-07-15
STORM-918 Storm CLI could validate arguments/print usage

bourneagain:
Storm commands that mandate proper args to be passed would now throw
the function's doc string as help rather than exiting out with error after creating the JVM
process.

Example : 

./storm kill
Syntax: [storm kill topology-name [-w wait-time-secs]]

    Kills the topology with the name topology-name. Storm will
    first deactivate the topology's spouts for the duration of
    the topology's message timeout to allow all messages currently
    being processed to finish processing. Storm will then shutdown
    the workers and clean up their state. You can override the length
    of time Storm waits between deactivation and shutdown with the -w flag.

./storm get-errors
Syntax: [storm get-errors topology-name]

    Get the latest error from the running topology. The returned result contains
    the key value pairs for component-name and component-error for the components in error.
    The result is returned in json format.

### DIFF
--- a/bin/storm.py
+++ b/bin/storm.py
@@ -22,7 +22,6 @@ import random
 import subprocess as sub
 import re
 import shlex
-import inspect
 try:
     # python 3
     from urllib.parse import quote_plus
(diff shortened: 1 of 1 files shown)


revans2:
@bourneagain I really don't like the idea of pulling data out of the call stack to check for arguments.  If we are going to check for arguments lets check for them in each of the functions, instead of in exec_storm_class.  Yes it is more code, but I think it is cleaner, and more maintainable.

bourneagain:
Thanks for the comments, I understand the implication now and have changed the code accordingly.

revans2:
+1 looks good to me

PR #638, opened 2015-07-17
STORM-945:<DefaultRolloverStrategy> element is not a policy,and should not be putted in the <Policies> element

caofangkun:
```
-> storm-current  bin/storm nimbus
Running: /usr/share/jdk-current/bin/java -server -Ddaemon.name=nimbus ... ...
2015-07-17 10:23:36,776 ERROR Policies has no parameter that matches element DefaultRolloverStrategy
2015-07-17 10:23:36,787 ERROR Policies has no parameter that matches element DefaultRolloverStrategy
2015-07-17 10:23:36,795 ERROR Policies has no parameter that matches element DefaultRolloverStrategy
```

### DIFF
--- a/log4j2/cluster.xml
+++ b/log4j2/cluster.xml
@@ -30,8 +30,8 @@
         </PatternLayout>
         <Policies>
             <SizeBasedTriggeringPolicy size="100 MB"/> <!-- Or every 100 MB -->
-            <DefaultRolloverStrategy max="9"/>
         </Policies>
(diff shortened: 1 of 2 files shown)


HeartSaVioR:
+1. I confirmed that works with 0.11.x-SNAPSHOT. 
Since it is misconfiguration of log4j2 and it's introduced to 0.10.0-beta, we'd be better to apply it to 0.10.x-branch, too.

PR #640, opened 2015-07-17
STORM-947:  replace all  `backtype.storm.scheduler.ExecutorDetails`  with `backtype.storm.generated.ExecutorInfo `

caofangkun:


storm-core/src/clj/backtype/storm/scheduler/DefaultScheduler.clj
```diff
@@ -18,7 +18,7 @@
   (:require [backtype.storm.scheduler.EvenScheduler :as EvenScheduler])
   (:import [backtype.storm.scheduler IScheduler Topologies
             Cluster TopologyDetails WorkerSlot SchedulerAssignment
-            EvenScheduler ExecutorDetails])
+            EvenScheduler])
```
chuanlei: So, is the original import unnecessary?

chuanlei:
I think it is good since `ExecutorDetails` does not take more information. And this change make Storm code more concise.

But I'm sure whether `ExecutorDetails` is designed for extension in the future.

storm-core/src/clj/backtype/storm/scheduler/DefaultScheduler.clj
caofangkun: @chuanlei 
Yes, It's unnecessary.

PR #641, opened 2015-07-17
[STORM-139] Correctly hash byte array tuple values

d2r:
* Unit test for correctness is included
* The following tests showed no discernible difference in latency or throughput:
  * word_count, last 10min stats after being running > 10min
  * word_count modified to send byte[] words instead of Strings, same conditions

This should handle the kinds of Objects we care about: reference types, all primitive array types, and reference array types.

Using Java's .hashCode is a problem for distributed systems, as was pointed out in the JIRA.

However, the behavior of [String#hashCode](https://docs.oracle.com/javase/7/docs/api/java/lang/String.html#hashCode%28%29), and the implementation of [java.util.Arrays#hashCode](http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/6-b14/java/util/Arrays.java#Arrays.hashCode%28float[]%29) for all primitive array types is defined and should be consistent across JVMs.  This leaves Objects and Object[] that are not Strings.  In these cases, this change still trusts the hashCode method provided in those classes.

In practice, bad cases should be rare.  In the cases when hashing is inconsistent and some sort of partitioning is used, it could be very difficult to debug.


harshach:
+1

vesense:
+1

storm-core/src/jvm/backtype/storm/grouping/PartialKeyGrouping.java
```diff
@@ -65,7 +66,11 @@ public void prepare(WorkerTopologyContext context, GlobalStreamId stream, List<I
                 List<Object> selectedFields = outFields.select(fields, values);
                 ByteBuffer out = ByteBuffer.allocate(selectedFields.size() * 4);
                 for (Object o: selectedFields) {
-                    out.putInt(o.hashCode());
+                    if (o instanceof Object[]) {
```
HeartSaVioR: @d2r 
Could we add same thing (converting List to Object[] by toArray, and apply Arrays.deepHashCode()) from list-hash-code to this?

d2r: Yes, we need to catch the primitive array types as well somehow.

Here `o` may not be a List.  I think we need to check not only `instanceof Object[]` but all of the other primitive array types as well.  Probably something like this?

```Java
for (Object o: selectedFields) {
    if (o instanceof Object[]) {
        out.putInt(Arrays.deepHashCode((Object[])o));
    } else if (o instanceof byte[]) {
        out.putInt(Arrays.hashCode((byte[]) o));
    } else if (o instanceof short[]) {
        out.putInt(Arrays.hashCode((short[]) o));
    } else if (o instanceof int[]) {
        out.putInt(Arrays.hashCode((int[]) o));
    } else if (o instanceof long[]) {
        out.putInt(Arrays.hashCode((long[]) o));
    } else if (o instanceof char[]) {
        out.putInt(Arrays.hashCode((char[]) o));
    } else if (o instanceof float[]) {
        out.putInt(Arrays.hashCode((float[]) o));
    } else if (o instanceof double[]) {
        out.putInt(Arrays.hashCode((double[]) o));
    } else if (o instanceof boolean[]) {
        out.putInt(Arrays.hashCode((boolean[]) o));
    } else if (o != null) {
        out.putInt(o.hashCode());
    } else {
      out.putInt(0);
    }
```

This seems to be what [Arrays#deepHashCode](http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/6-b14/java/util/Arrays.java#3709) does.

What do you think?

HeartSaVioR: @d2r 
I didn't think about primitive array type.
Actually I mean that,
- Does 'o' have an chance to be instance of List? 
- If then, could we apply ```((List)o).toArray()``` to convert it Object[], and let if statement handle it?

d2r: I see. Yes, I could add this.  However, since we do not currently do this in backtype.storm.tuple it would be inconsistent.

In b.s.tuple, I always get a List<Object>.  I convert it to an array there because the call to Arrays#deepHashCode is simple.  If a tuple is _itself_ a List, then Arrays#hashCode will call List#hashCode on it.

storm-core/src/clj/backtype/storm/tuple.clj
```diff
@@ -15,8 +15,11 @@
...
-  (:import [java.util List]))
+  (:import [java.util Arrays List]))
 
 (defn list-hash-code
   [^List alist]
-  (.hashCode alist))
+  (if (nil? alist)
+    1
+    (let [^"[Ljava.lang.Object;" array (.toArray alist)] ;; Object[]
+      (Arrays/deepHashCode array))))
```
d2r: The type hint here seems to be unnecessary.  I'll remove it.

d2r:
Ok, addressed comments.

@HeartSaVioR, I made your suggested change as well.

HeartSaVioR:
LGTM, +1.

PR #642, opened 2015-07-18
[STORM-643] KafkaUtils repeatedly fetches messages whose offset is out of range

vesense:
STORM-643 fixing based on master branch

### DIFF
--- a/external/storm-kafka/src/jvm/storm/kafka/ExponentialBackoffMsgRetryManager.java
+++ b/external/storm-kafka/src/jvm/storm/kafka/ExponentialBackoffMsgRetryManager.java
@@ -19,9 +19,11 @@ package storm.kafka;
 
 import java.util.Comparator;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Map;
 import java.util.PriorityQueue;
 import java.util.Queue;
+import java.util.Set;
 
 public class ExponentialBackoffMsgRetryManager implements FailedMsgRetryManager {
 
--- a/external/storm-kafka/src/jvm/storm/kafka/ExponentialBackoffMsgRetryManager.java
+++ b/external/storm-kafka/src/jvm/storm/kafka/ExponentialBackoffMsgRetryManager.java
@@ -91,6 +91,21 @@ public class ExponentialBackoffMsgRetryManager implements FailedMsgRetryManager 
                 System.currentTimeMillis() >= record.retryTimeUTC;
     }
 
+    @Override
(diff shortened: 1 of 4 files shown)


harshach:
+1

2new:
+1

HeartSaVioR:
@vesense 
Seems like your patch removes one invalid offset per each fill().

We may apply same optimization here.
When TopicOffsetOutOfRangeException occurs, we can remove invalid offsets from ExponentialBackoffMsgRetryManager at once.

Though we should loop and examine each records to find them, it would be better than current.

vesense:
@HeartSaVioR Thanks. I will fix that.

vesense:
@HeartSaVioR
I have completed the optimization. Could you take a look? Thx.

lazyval:
Minor complain, but [next time]  it would be handy to keep commit messages a bit more descriptive

storm 2015-07-20 14-42-47

vesense:
@lazyval Thank you for pointing out my mistake. I'm sorry for that. It won't happen next time.
Thanks again.

HeartSaVioR:
@vesense Since it contains too many commits compared to changed lines, I will try to squash commits into one and merge.

HeartSaVioR:
Forgot to give +1. :)

HeartSaVioR:
@vesense 
Thanks for quick work! I merged this into master and 0.10.x-branch respectively.

- master: https://github.com/apache/storm/commit/f091743d9ebc935c9bbbc210095884c3d819632b
- 0.10.x-branch: https://github.com/apache/storm/commit/05bd1e36697fe7038284833032da262afb82ac99

I squashed your commits into one before merging, so Github can't recognize this PR to closed.
Could you close this PR?

vesense:
@HeartSaVioR OK. thx.

prokopowicz:
we've seen a related problem when a custom scheme returns an empty list of lists via generateTuples.  The code in partitionManager does not ack the offset in that case.  It should be made clear that a scheme should not return an empty list of lists, but null, if it deserializes no objects.  Or, this code in PartitionManager should be fixed to treat an empty list of lists in the same way it treats a null list.

```java
            if (tups != null) {
                if(_spoutConfig.topicAsStreamId) {
                    for (List<Object> tup : tups) {
                        collector.emit(_spoutConfig.topic, tup, new KafkaMessageId(_partition, toEmit.offset));
                    }
                } else {
                    for (List<Object> tup : tups) {
                        collector.emit(tup, new KafkaMessageId(_partition, toEmit.offset));
                    }
                }
                break;
            } else {
                ack(toEmit.offset);
            }
```

PR #644, opened 2015-07-21
[STORM-837] Support for exactly once semantics in HdfsState

arunmahadevan:
Changes to support exactly once semantics in HdfsState.

1. Moved the file rotation and sync to commit()
2. In pre-commit, if we have previously seen the txnid, recover the data up to that point by copying to a new file and discard the current data file.
3. In pre-commit atomically update [current txid, the datafile path and the current offset] in a (per partition) index file.
4. To keep it simple, automatically turn off exactly once semantics if TimedRotation policy is in use.

Have tested with the normal flow and simulating the recovery scenario, with both regular Hdfs file and sequence files. Appears to work fine.

### DIFF
--- a/external/storm-hdfs/src/main/java/org/apache/storm/hdfs/trident/HdfsState.java
+++ b/external/storm-hdfs/src/main/java/org/apache/storm/hdfs/trident/HdfsState.java
@@ -35,6 +35,7 @@ import org.apache.storm.hdfs.trident.format.FileNameFormat;
 import org.apache.storm.hdfs.trident.format.RecordFormat;
 import org.apache.storm.hdfs.trident.format.SequenceFormat;
 import org.apache.storm.hdfs.trident.rotation.FileRotationPolicy;
+import org.apache.storm.hdfs.trident.rotation.FileSizeRotationPolicy;
 import org.apache.storm.hdfs.trident.rotation.TimedRotationPolicy;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -124,9 +125,16 @@ public class HdfsState implements State {
             this.writeLock = new Object();
             if (this.rotationPolicy == null) {
                 throw new IllegalStateException("RotationPolicy must be specified.");
+            } else if (this.rotationPolicy instanceof FileSizeRotationPolicy) {
+                long limit = FileSizeRotationPolicy.Units.GB.getByteCount();
+                if(((FileSizeRotationPolicy) rotationPolicy).getMaxBytes() > limit) {
+                    LOG.warn("*** Exactly once semantics is not supported for FileSizeRotationPolicy with size > 1 GB ***");
+                    LOG.warn("Turning off exactly once.");
+                    this.exactlyOnce = false;
+                }
             } else if (this.rotationPolicy instanceof TimedRotationPolicy) {
-                LOG.warn("*** Exactly once semantics is not supported with TimedRotationPolicy ***");
-                LOG.warn("*** Turning off exactly once.");
+                LOG.warn("*** Exactly once semantics is not supported for TimedRotationPolicy ***");
+                LOG.warn("Turning off exactly once.");
                 this.exactlyOnce = false;
             }
             if (this.fsUrl == null) {
@@ -220,8 +228,18 @@ public class HdfsState implements State {
             return this;
         }
 
-        public HdfsFileOptions withBufferSize(int size) {
-            this.bufferSize = Math.max(4096, size); // at least 4K
+        /**
+         * <p>Set the size of the buffer used for hdfs file copy in case of recovery. The default
+         * value is 131072.</p>
+         *
+         * <p> Note: The lower limit for the parameter is 4096, below which the
+         * option is ignored. </p>
+         *
+         * @param sizeInBytes the buffer size in bytes
+         * @return {@link HdfsFileOptions}
+         */
+        public HdfsFileOptions withBufferSize(int sizeInBytes) {
+            this.bufferSize = Math.max(4096, sizeInBytes); // at least 4K
             return this;
         }
 
--- a/external/storm-hdfs/src/main/java/org/apache/storm/hdfs/trident/HdfsState.java
+++ b/external/storm-hdfs/src/main/java/org/apache/storm/hdfs/trident/HdfsState.java
@@ -139,9 +139,7 @@ public class HdfsState implements State {
                 throw new RuntimeException("Error preparing HdfsState: " + e.getMessage(), e);
             }
 
-            if (this.rotationPolicy instanceof TimedRotationPolicy) {
-                ((TimedRotationPolicy) this.rotationPolicy).start();
-            }
+            rotationPolicy.start();
         }
 
         /**
@@ -455,6 +453,12 @@ public class HdfsState implements State {
     }
 
     /**
+     * Returns temp file path corresponding to a file name.
+     */
+    private Path tmpFilePath(String filename) {
+        return new Path(filename + ".tmp");
+    }
+    /**
      * Reads the last txn record from index file if it exists, if not
      * from .tmp file if exists.
      *
@@ -463,7 +467,7 @@ public class HdfsState implements State {
      * @throws IOException
      */
     private TxnRecord getTxnRecord(Path indexFilePath) throws IOException {
-        Path tmpPath = new Path(indexFilePath.toString() + ".tmp");
+        Path tmpPath = tmpFilePath(indexFilePath.toString());
         if (this.options.fs.exists(indexFilePath)) {
             return readTxnRecord(indexFilePath);
         } else if (this.options.fs.exists(tmpPath)) {
@@ -489,7 +493,7 @@ public class HdfsState implements State {
         FSDataOutputStream out = null;
         LOG.debug("Starting index update.");
         try {
-            Path tmpPath = new Path(this.indexFilePath.toString() + ".tmp");
+            Path tmpPath = tmpFilePath(indexFilePath.toString());
             out = this.options.fs.create(tmpPath, true);
             BufferedWriter bw = new BufferedWriter(new OutputStreamWriter(out));
             TxnRecord txnRecord = new TxnRecord(txId, options.currentFile.toString(), this.options.getCurrentOffset());
--- a/external/storm-hdfs/src/main/java/org/apache/storm/hdfs/trident/HdfsState.java
+++ b/external/storm-hdfs/src/main/java/org/apache/storm/hdfs/trident/HdfsState.java
@@ -17,14 +17,14 @@
  */
 package org.apache.storm.hdfs.trident;
 
+import backtype.storm.Config;
 import backtype.storm.task.IMetricsContext;
 import backtype.storm.topology.FailedException;
 import org.apache.hadoop.conf.Configuration;
-import org.apache.hadoop.fs.FSDataOutputStream;
-import org.apache.hadoop.fs.FileSystem;
-import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.fs.*;
 import org.apache.hadoop.hdfs.client.HdfsDataOutputStream;
 import org.apache.hadoop.io.SequenceFile;
+import org.apache.hadoop.io.Writable;
 import org.apache.hadoop.io.compress.CompressionCodecFactory;
 import org.apache.storm.hdfs.common.rotation.RotationAction;
 import org.apache.storm.hdfs.common.security.HdfsSecurityUtil;
@@ -40,8 +40,7 @@ import storm.trident.operation.TridentCollector;
 import storm.trident.state.State;
 import storm.trident.tuple.TridentTuple;
 
-import java.io.IOException;
-import java.io.Serializable;
+import java.io.*;
 import java.net.URI;
 import java.util.*;
 
@@ -60,6 +59,10 @@ public class HdfsState implements State {
         protected ArrayList<RotationAction> rotationActions = new ArrayList<RotationAction>();
         protected transient Object writeLock;
         protected transient Timer rotationTimer;
+        /**
+         * This is on by default unless TimedRotationPolicy is in use.
+         */
+        private boolean exactlyOnce = true;
 
         abstract void closeOutputFile() throws IOException;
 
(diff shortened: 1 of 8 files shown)


external/storm-hdfs/src/main/java/org/apache/storm/hdfs/trident/HdfsState.java
```diff
@@ -17,14 +17,14 @@
...
 package org.apache.storm.hdfs.trident;
 
+import backtype.storm.Config;
 import backtype.storm.task.IMetricsContext;
 import backtype.storm.topology.FailedException;
 import org.apache.hadoop.conf.Configuration;
-import org.apache.hadoop.fs.FSDataOutputStream;
-import org.apache.hadoop.fs.FileSystem;
-import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.fs.*;
```
knusbaum: Let's not import *

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/trident/HdfsState.java
```diff
@@ -40,8 +40,7 @@
 import storm.trident.state.State;
 import storm.trident.tuple.TridentTuple;
 
-import java.io.IOException;
-import java.io.Serializable;
+import java.io.*;
```
knusbaum: Again here.

arunmahadevan:
@knusbaum makes sense, have removed wildcard imports.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/trident/HdfsState.java
```diff
@@ -136,33 +193,40 @@ public void run() {
...
 
-        public HdfsFileOptions withRotationPolicy(FileRotationPolicy rotationPolicy){
+        public HdfsFileOptions withRotationPolicy(FileRotationPolicy rotationPolicy) {
             this.rotationPolicy = rotationPolicy;
             return this;
         }
 
-        public HdfsFileOptions addRotationAction(RotationAction action){
+        public HdfsFileOptions withBufferSize(int size) {
+            this.bufferSize = Math.max(4096, size); // at least 4K
```
harshach: 4096 should be substituted with buffersize? or do we need to take max it looks like we are saying the minimum should be default bufferSize. if this is needed we need to document it.

arunmahadevan: The default is 131072 (128 K). Here just ensuring that, if at all user tries to change the default, it does not go below 4 K.

harshach:
@revans2 @ptgoetz can you take a look at this patch. Thanks.

ptgoetz:
@arunmahadevan Quick question (I haven't done a full review yet): Is there a way to make this work for Timed rotation policies? That one of the most widely used rotation policies.

ptgoetz:
It took me some time to follow what's going on with this patch, so I'll document it here for the benefit other reviewers.

It operates by using the concept of a temporary "transaction file" that contains the last successfully written/committed file write offset and the transaction ID. That file is initially read in the `prepare()` method call, and is initialized if it does not already exist. The POJO representing that file is stored in a `lastTxn` instance variable, which represents that last committed transaction.

In the `beginCommit()` method, it checks the transaction ID against the one stored in `lastTxn`. If it is <= (indicating a replay of a batch), it triggers recovery (more on that later), otherwise it continues normally (writing data in the `execute()` method).

In the `commit()` method, it essentially re-writes the transaction file with the new transaction ID and the current file offset after the latest batch data was written. This offset serves as the starting point for recovery.

**Recovery**
When recovery is triggered, it is assumed that there was a failure during the `execute()` method, and the data file now contains data from tuples that will be replayed. If it were to simply append to that file, there would be duplicate records. So it effectively truncates the data file by creating new file, and copying everything from the old data file *up to the last successfully committed offset,* and discarding the remaining data.

I think this is an interesting approach and is certainly more resilient than what we currently have, but I think there's a scaling problem with the recovery functionality. Recovery works by essentially copying the old/orphaned data file into a new file and skipping any extra, uncommitted data it contains. This is fine for small files, but what about the case where the data file is really big and the recovery time would take longer than `topology.message.timeout.secs`?

Another problem I see is the lack of support for time-based file rotation. I understand why it's not supported by this patch, but as I mentioned earlier, that's one of the most common use cases.

arunmahadevan:
@ptgoetz to address the concern of recovery times for large files, can we recommend (or auto enforce) the max size in "size based rotation" to a reasonable threshold (say 1GB). As long as the files get rotated at a reasonable threshold which can be copied over quickly there wont be issues.

Regarding time based rotation: It may be possible to support by guarding all the operations with locks and then have the new file path updated in the index along with rotation. (there might be other corner cases which needs to be thought through). 

However the issue I see with time based rotation is we don't have a control of the file size and if users configure a daily rotation policy and the data rate is high, we wont be able to recover such huge files in case of failure.

harshach:
@ptgoetz @arunmahadevan  can we not make this as a default behaviour instead make this as option. We can document that exactly-once behaviour  comes with limitations around file rotation.

arunmahadevan:
@harshach Are you saying that we don't support time based rotation at all? The current approach automatically turns off exactly once if TimedRotation is in use, so that who want to use time based rotation would still be able to use it with the limitation that they will not get exactly once semantics.

I also think we can do the same with max file size in case of size based rotation policy.

arunmahadevan:
@revans2 @ptgoetz @harshach can we go ahead by automatically disabling exactly once for time based rotation and for file sizes > threshold, support exactly once for the other cases and document this behavior ?

harshach:
I am +1 on the last approach. @arunmahadevan  please add to the README under storm-hdfs.

arunmahadevan:
Hi @harshach, added a note to storm hdfs README.md and modified code to disable exactly once based on file size as well.

revans2:
I did a quick pass through the code and it looks OK, but I have not looked at it in great detail.  I am not very happy with the limitations on which rotation policy you can use, nor on the size limit.  I would rather be correct but slow by default in all cases, even if they set bad configs (> 1GB), and give them the power to make it fast but wrong if they know the risks and can deal with it.  Also a hard coded 1GB limit seems a little strange.  What if we have a 10GigE connection or even infiniband for HDFS and all of the data happens to be in memory.  We could in theory have processed the 1GB in less then a second, still painful but not the end of the world.

Why don't we want to support a time based rotation, that rotates at the end of a batch after the time has passed instead of in the middle of the batch?

arunmahadevan:
I was trying to put a safety net  so that the recovery would always work. The limitation on the timed rotation policy is also on similar lines that we don't know how much data would be written to the files before the timed rotation kicks in.

So,

- For file size rotation, remove the restrictions on the file size and just log warnings.
- For time based rotation, set a flag and do the actual rotation in `doCommit` as you suggested.
- Add a note in the README about the risk that the recovery would fail if files cannot be recovered within timeout and hence should be kept to reasonable sizes (and time interval) or the message timeout should be increased.

Does it sound reasonable ? If we agree, will make the changes.

revans2:
I am fine with those changes in principal.  Although I do want to spend some time reading through the code to convince myself that there are no corner cases that we are missing.

Also I would love to see some unit tests to show we can recover.  I believe that hsync works on the local file system so you would not need to bring up an HDFS mini cluster, just create a file system using ```file:///...``` as the URL and write there.

arunmahadevan:
Removed restrictions on size and time based rotation policy, updated the README and also added unit tests.

arunmahadevan:
@revans2 Made changes as you suggested. can you take a look?

revans2:
For me internally the tests pass and everything builds.  Although the rest of storm seems a bit unstable with tests.

I think the changes look fine +1 for merging this in.

I don't think we have any split brain problems because of how we route messages, so the begin command and commit messages should prevent this.

There are some things that I think we can do better, but are not wrong. 
1) I am scared that the transaction file is going to put more of a load on the Name Node than I like.  I am mostly concerned that if we have thousands of instances of hdfs state creating/deleting transaction files multiple times a second the load is going to noticeable on a large cluster.  But this is a premature optimization so I would say we don't address it unless we see it as an actual problem.
2) The RotationActions don't have a recovery interface, and things like the MoveFile action could potentially fail after successfully rotating the file which might leave the file in the wrong directory.
3) I don't really like the special case for the TimedRotationAction.start.  I would rather see start a part of the RotationAction interface, that can be ignored by those that don't want/need it.

The last two could perhaps be a follow on JIRA.  I am fine with the code as is, but I would like another pair of eyes looking at this too before merging it in.

knusbaum:
+1 
I looked through it and didn't see any obvious flaws. I'm not generally in favor of the whitespace-only changes, but that's not a huge deal.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/trident/HdfsState.java
```diff
@@ -136,44 +174,98 @@ public void run() {
...
+         *
+         * @param sizeInBytes the buffer size in bytes
+         * @return {@link HdfsFileOptions}
+         */
+        public HdfsFileOptions withBufferSize(int sizeInBytes) {
+            this.bufferSize = Math.max(4096, sizeInBytes); // at least 4K
+            return this;
+        }
+
+        @Deprecated
```
d2r: Why do we deprecate this?

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/trident/HdfsState.java
```diff
@@ -69,63 +78,92 @@
...
+            } else if (this.rotationPolicy instanceof FileSizeRotationPolicy) {
+                long rotationBytes = ((FileSizeRotationPolicy) rotationPolicy).getMaxBytes();
+                LOG.warn("FileSizeRotationPolicy specified with {} bytes.", rotationBytes);
+                LOG.warn("Recovery will fail if data files cannot be copied within topology.message.timeout.secs.");
+                LOG.warn("Ensure that the data files does not grow too big with the FileSizeRotationPolicy.");
+            } else if (this.rotationPolicy instanceof TimedRotationPolicy) {
+                LOG.warn("TimedRotationPolicy specified with interval {} ms.", ((TimedRotationPolicy) rotationPolicy).getInterval());
+                LOG.warn("Recovery will fail if data files cannot be copied within topology.message.timeout.secs.");
+                LOG.warn("Ensure that the data files does not grow too big with the TimedRotationPolicy.");
+            }
```
d2r: It might be nicer to move this code into the subclasses own methods: FileSizeRotationPolicy#prepare and TimedRotationPolicy#prepare.  These prepare methods can each call `super.prepare(conf, partitionIndex, numPartitions)`.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/trident/HdfsState.java
```diff
@@ -69,63 +78,92 @@
...
-                            rotateOutputFile();
-                        } catch(IOException e){
-                            LOG.warn("IOException during scheduled file rotation.", e);
-                        }
-                    }
-                };
-                this.rotationTimer.scheduleAtFixedRate(task, interval, interval);
+            if (this.rotationPolicy instanceof TimedRotationPolicy) {
+                ((TimedRotationPolicy) this.rotationPolicy).start();
+            }
```
d2r: Referencing the [other comment above](https://github.com/apache/storm/pull/644/files#r37024551), TimedRotationPolicy#prepare could look like:
```Java
LOG.warn("TimedRotationPolicy specified with interval {} ms.", ((TimedRotationPolicy) rotationPolicy).getInterval());
LOG.warn("Recovery will fail if data files cannot be copied within topology.message.timeout.secs.");
LOG.warn("Ensure that the data files does not grow too big with the TimedRotationPolicy.");

super.prepare();

this.rotationPolicy.start();
```

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/trident/HdfsState.java
```diff
@@ -288,47 +391,168 @@ void closeOutputFile() throws IOException {
...
+    /**
+     * Reads the last txn record from index file if it exists, if not
+     * from .tmp file if exists.
+     *
+     * @param indexFilePath the index file path
+     * @return the txn record from the index file or a default initial record.
+     * @throws IOException
+     */
+    private TxnRecord getTxnRecord(Path indexFilePath) throws IOException {
+        Path tmpPath = new Path(indexFilePath.toString() + ".tmp");
```
d2r: It would be good to have this pulled into its own method, just so that we do not have ".tmp" hard-coded in two spots.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/trident/HdfsState.java
arunmahadevan: As mentioned in the previous comments and also in the jira (STORM-837), rotation actions are too generic and make the recovery difficult if the process crash in the middle of the action. Hence its best to avoid rotation actions in case exactly once semantics is expected. Will add a note in the README.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/trident/HdfsState.java
arunmahadevan: Actually the warnings are specific to HDFS state and by moving it to the rotation policy we make the rotation policy aware of the recovery which might not be ideal if the rotation policy class needs to be reused in some other context.

Moving the start logic to policy specific prepare methods makes sense. I will make this change.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/trident/HdfsState.java
arunmahadevan: will do this.

arunmahadevan:
@d2r Thanks for the review, incorporated your feedback.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/trident/HdfsState.java
d2r: Ah, I hadn't noticed that part of the JIRA description.  That's a good answer; just wanted to make sure the change was intentional.

d2r:
Thanks @arunmahadevan 

+1 It seems fine to me, and the new tests pass for me.  The remaining comment on refactoring is minor, so I am ok merging it.

Regarding Bobby's comments: 1) HDFS name node load, I agree we can open an Issue if we see it is really a problem; 2) no recovery from actions, this could be a problem, but we are marking it as deprecated.  If we want to keep support for actions, we should do follow-on work that provides for some recovery of those operations too.

revans2:
Still +1 looks good.

ptgoetz:
+1

PR #647, opened 2015-07-22
(STORM-956) When the execute() or nextTuple() hang on external resources, stop the Worker's heartbeat

chuanlei:


hustfxj:
Maybe it is not good ，the time-out check about  execute()&nextTuple()  should depends on the user .

bastiliu:
I think it is better to raise a warning intead of stopping to update worker heartbeat. There are some topologys that might require to block at execute()/nextTuple() to wait some essential initialization (e.g. establish connection to database..).
For the trigger mechanism of hanging on problem, it is a potential risk that worker might be restart unexpectedly due to heartbeat timeout. Because we consume a batch of tuples each time. Even through the time of processing one tuple is short, the time of processing the batch could cause the timeout.

kishorvpatil:
I think the spout and bolt should take care of handling hangs ( or use timeouts instead of making blocking calls). Also, the spout/bolt code should guard against creating threads that can cause unhandled exceptions/hang-ups. Forcing worker to not send heart-beats would make killing other components running on that worker - which is not desired.
Secondly, worker should not be killed unless it is certain that is the process issue and not external service issue - e.g.  if kafka spout hangs - killing worker might force it to be relaunched or scheduled may not solve the problem - new worker process still make another blocking call and hang-up.

Thirdly, killing worker will force relaunch/reschedule/ - forcing topology to be un-stabie as all other workers in loop have to reconnect to this new worker. In large topologies that might become a bigger problem and lead to domino effects and take longer to settle the topology.

-1

revans2:
I personally can see both sides of this.  There are situations where a bolt/spout may be hung because of a bug, like a thread deadlock in some external client, and restarting the worker will fix the issue.  But I agree that this should be a very rare situation.  

I don't get the argument that this is expensive if we get it wrong.  We already fail fast if some bolt or spout throws an unexpected exception.  I don't see this being that different.

A bolt or spout not able to process anything for 5 mins seems like an OK time to see if we can restart things, especially for a low latency framework.  I personally am +1 on the concept of having timeouts.  I would like to see some changes to the implementation of this patch, but there is no reason to go into that if @kishorvpatil has a -1 on even the idea of it.  @kishorvpatil and @bastiliu have I swayed you at all with my argument?

kishorvpatil:
@revans2 Yes, you have convinced me on conceptually it option where executor could be truely deadlocked. But implementation wise, I think we need better control around how/when we choose to shoot worker itself.

hustfxj:
@revans2 Now  I personally am indecisive on the concept of having timeouts. Sometimes execute()&nextTuple()  maybe need a long time due to data arrival.  Of course, having timeouts maybe is a good choice. But the restart or other operations should dependency on users themselves. So can we do it by Topology Hooks?

bastiliu:
@revans2 Yes, I agree that this checking is helpful to find the problem spout/bolt. My point here is that the solution could be improved.
1. If timeout, it is better to raise a warning(e.g. give a warning on web UI). Because we have seen some topologys that might require to block at execute()/nextTuple() to wait some essential initialization. e.g. the connection to database in a bolt is down. The user would like to wait untill the reconnection is done.  
2. The triggering mechanism of "last-active-time" timeout should be updated. Current implementation puts a "last-active-time" tuple to receiving queue, then spout/bolt update the "last-active-time" when retrieving the trigger tuple from receiving queue. But it is possible that there already have been many tuples in receiving queue before putting the "last-active-time" trigger tuple. So the spout/bolt must process all the tuples which are put into receiving queue before the trigger tuple. The processing of total topology tuples might take a long time which probably cause the timeout, even if the processing time of a tuple is short. From user's point of view, that is unexpected.

revans2:
To make a feature like this work for storm, we need to provide people a lot of control over it.  I personally would prefer to see something where the Bolt/Spout can indicate life without completely processing a tuple.  Probably an API in OutputCollector that would allow it to heartbeat in saying don't shoot me.  Any other interaction with an OutputCollection should also prevent the component from being shot.

The timeout should be configurable on a per-component basis, including turning them off.  Having all of them set to the same cluster wide setting is not good.  Also if we detect that we are in a bad state there is no reason to stop heart-beating, and wait for a supervisor to shoot us.  If we think we are bad just exit.  It will let recovery happen much faster.

I agree with @bastiliu too that one of the options should be not to shot ourselves, but to alert that we are seeing problems.  I would like to see this done through both the metrics system and through writing an error into zookeeper that would show up on the UI for the component that is stuck.  The metrics would allow for automated alerting and the UI would let users manually see what is happening.

knusbaum:
Closing this.

PR #649, opened 2015-07-22
[STORM-958] adds config for group.mapping.service init params

d2r:
Users can provide optional parameters to the plugin in the form of a Map.

### DIFF
--- a/conf/defaults.yaml
+++ b/conf/defaults.yaml
@@ -39,6 +39,7 @@ storm.local.mode.zmq: false
 storm.thrift.transport: "backtype.storm.security.auth.SimpleTransportPlugin"
 storm.principal.tolocal: "backtype.storm.security.auth.DefaultPrincipalToLocal"
 storm.group.mapping.service: "backtype.storm.security.auth.ShellBasedGroupsMapping"
+storm.group.mapping.service.params: null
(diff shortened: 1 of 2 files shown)


harshach:
+1

vesense:
+1

HeartSaVioR:
+1.

HeartSaVioR:
@d2r 
Btw, could you please write some descriptions to document or conf/defaults.yaml?
If I were just a user of Storm, I couldn't understand what this field is for, and/or how to use it.

d2r:
> Btw, could you please write some descriptions to document or conf/defaults.yaml?
> If I were just a user of Storm, I couldn't understand what this field is for, and/or how to use it.

@HeartSaVioR Thanks, yes, I'll do that and update this PR.  Is it sufficient to expand the Java comment in Config.java?  I think this is typically where we put config documentation.

HeartSaVioR:
@d2r Yes, I also think it is enough to put comment to Config.java for now.
If we agree it's not enough to explain configuration, whole of things should be documented to static html page of web site.

d2r:
@HeartSaVioR: updated

HeartSaVioR:
+1

PR #651, opened 2015-07-23
STORM-1028: Eventhub spout meta data

tandrup:
Add "partition","seq-number" fields to emitted tupples for event sourcing consumers, which need to preserve partition order and be able to replay.

### DIFF
--- a/external/storm-eventhubs/src/main/java/org/apache/storm/eventhubs/spout/EventDataScheme.java
+++ b/external/storm-eventhubs/src/main/java/org/apache/storm/eventhubs/spout/EventDataScheme.java
@@ -22,8 +22,10 @@ import java.util.ArrayList;
 import java.util.List;
 import org.apache.qpid.amqp_1_0.client.Message;
 import org.apache.qpid.amqp_1_0.type.Section;
+import org.apache.qpid.amqp_1_0.type.Symbol;
 import org.apache.qpid.amqp_1_0.type.messaging.AmqpValue;
 import org.apache.qpid.amqp_1_0.type.messaging.Data;
+import org.apache.qpid.amqp_1_0.type.messaging.MessageAnnotations;
 
 public class EventDataScheme implements IEventDataScheme {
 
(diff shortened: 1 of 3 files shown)


rsltrifork:
Event hub (and Kafka) play well into event source architectures as event ingest point for later Storm processing to downstream stateful consumers.

Advanced event stream processing, such as replaying parts of a stream, requires that the downstream consumers can synchronise different "stream runs" to their stateful view, which itself can be seen as an aggregation of all previous events. To set up the right context for re-processing the stream in a deterministic way, they need to sync their view with the incoming old data. To be able to do this, they need knowledge of the event sequenceNumber and partition.

For example, if you have a bolt that calculates total_order_amount for a stream of orders, and emits order tuples with the total_order_amount calculated for all previous orders, replaying an order event should not change total_order_amount. I.e. orders with a higher sequenceNumber than the order being processed should not be included in total_order_amount.

This synchronisation can be achieved if the bolt has access to the parition and sequenceNumber from eventHub.

external/storm-eventhubs/src/main/java/org/apache/storm/eventhubs/spout/EventDataScheme.java
```diff
@@ -42,6 +44,9 @@
         AmqpValue amqpValue = (AmqpValue) section;
         fieldContents.add(amqpValue.getValue().toString());
         return fieldContents;
+      }else if (section instanceof MessageAnnotations) {
```
lazyval: quote minor, but additional space will not hurt

harshach:
@tandrup can you please open a JIRA here https://issues.apache.org/jira/browse/STORM/ and also update the PR with JIRA number in the title.

mooso:
(Not a project maintainer, just someone who ran into the same problem)
I support the idea and encountered the same need myself, but I think they way it's implemented in this pull request isn't correct since it doesn't work for people who implement their own IEventDataScheme class. The way I did it was to change the signature of deserialize() in that interface to add a MessageId parameter, that way implementors of that class can choose to add the partition ID (and sequence number) properly if they wish. Would that work?

ptgoetz:
@tandrup Any update on this? Any thoughts regarding @mooso's comment?

rsltrifork:
Mooso is right, and his suggestion is a cleaner implementation. I suggest we use his implementation of IEventDataScheme and EventDataScheme. 
Mooso can you make a PR for your copy of these?

ptgoetz:
@mooso Any update?

@rsltrifork Would you be willing to make the change/PR based on the earlier comments?

knusbaum:

PR #657, opened 2015-07-29
STORM-964, add config for worker logwriter to restrict its mem usage

zhuoliu:
The log writer process that writes out stderr and stdout should have a config that controls it's memory usage and it should default to a very small value, probably 50MB in most cases.

To address above issue, one new parameter is defined for logwriter's jvmopts: "topology.worker.lw.childopts".
The default heap size of logwriter is set as 64M, or user can override it through "-c topology.worker.lw.childopts='xxxx " when submitting a topology.

### DIFF
--- a/storm-core/src/clj/backtype/storm/daemon/supervisor.clj
+++ b/storm-core/src/clj/backtype/storm/daemon/supervisor.clj
@@ -675,7 +675,7 @@
                         (add-to-classpath topo-classpath))
           top-gc-opts (storm-conf TOPOLOGY-WORKER-GC-CHILDOPTS)
           gc-opts (substitute-childopts (if top-gc-opts top-gc-opts (conf WORKER-GC-CHILDOPTS)) worker-id storm-id port)
-          topo-worker-lw-childopts (conf TOPOLOGY-WORKER-LW-CHILDOPTS)
+          topo-worker-logwriter-childopts (conf TOPOLOGY-WORKER-LOGWRITER-CHILDOPTS)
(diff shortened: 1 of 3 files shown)


knusbaum:
Would it be possible to expand 'lw' to 'logwriter'? 
I know it's longer, but it's much clearer what the config is for.

zhuoliu:
Good suggestion! Updated.

storm-core/src/clj/backtype/storm/daemon/supervisor.clj
```diff
@@ -675,6 +675,7 @@
                         (add-to-classpath topo-classpath))
           top-gc-opts (storm-conf TOPOLOGY-WORKER-GC-CHILDOPTS)
           gc-opts (substitute-childopts (if top-gc-opts top-gc-opts (conf WORKER-GC-CHILDOPTS)) worker-id storm-id port)
+          topo-worker-logwriter-childopts (conf TOPOLOGY-WORKER-LOGWRITER-CHILDOPTS)
```
caofangkun: should use ```storm-conf``` but not ```conf``` here? 
```topo-worker-logwriter-childopts (storm-conf TOPOLOGY-WORKER-LOGWRITER-CHILDOPTS)```

zhuoliu: Good catch. Since user may have configuration for this in topology's stormconf.ser. Updated.

revans2:
The change looks good to me +1, but I want to give @knusbaum some time to follow up on his review too.

knusbaum:
+1

PR #652, opened 2015-07-23
STORM-959:remove unnecessary dependency from storm-hive/pom.xml

caofangkun:
org.apache.calcite:calcite-core:0.9.2-incubating does not take affect at all.

PR #658, opened 2015-07-30
STORM-966 ConfigValidation.DoubleValidator is misleading and not neccessary

jerrypeng:
ConfigValidation.DoubleValidator doesn't really validate whether the type of the object is a double.

ConfigValidation.DoubleValidator code only checks if the object is null whether if the object is a instance of Number which is a parent class of Double.
DoubleValidator is only used once in Config.java and in that instance:
public static final Object TOPOLOGY_STATS_SAMPLE_RATE_SCHEMA = ConfigValidation.DoubleValidator;
can just be set to:
public static final Object TOPOLOGY_STATS_SAMPLE_RATE_SCHEMA = NUMBER.class;
Then we can just get rid of the misleading function ConfigValidation.DoubleValidator since it doesn't really check if a object is of double type thus the validator function doesn't really do anything and the name is misleading. In previous commit https://github.com/apache/storm/commit/214ee7454548b884c591991b1faea770d1478cec Number.Class was used anyway

### DIFF
--- a/storm-core/src/jvm/backtype/storm/ConfigValidation.java
+++ b/storm-core/src/jvm/backtype/storm/ConfigValidation.java
@@ -28,7 +28,6 @@ public class ConfigValidation {
     /**
      * Declares methods for validating configuration values.
      */
-    public static interface FieldValidator {
         /**
          * Validates the given field.
          * @param name the name of the field.
@@ -268,7 +267,7 @@ public class ConfigValidation {
                     return;
                 }
             }
-            throw new IllegalArgumentException("Field " + name + " must be a Number");
+            throw new IllegalArgumentException("Field " + name + " must be a Positive Number");
         }
     };
 
--- a/storm-core/src/jvm/backtype/storm/ConfigValidation.java
+++ b/storm-core/src/jvm/backtype/storm/ConfigValidation.java
@@ -254,6 +254,25 @@ public class ConfigValidation {
     };
 
     /**
+     * Validates a Positive Number
+     */
+    public static Object PositiveNumberValidator = new FieldValidator() {
+        @Override
+        public void validateField(String name, Object o) throws IllegalArgumentException {
+            if (o == null) {
+                // A null value is acceptable.
+                return;
+            }
+            if(o instanceof Number) {
+                if(((Number)o).doubleValue() > 0.0) {
+                    return;
+                }
+            }
+            throw new IllegalArgumentException("Field " + name + " must be a Number");
+        }
+    };
+
+    /**
      * Validates a power of 2.
      */
     public static Object PowerOf2Validator = new FieldValidator() {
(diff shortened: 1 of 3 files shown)


caofangkun:
+1

HeartSaVioR:
You're right about DoubleValidator. DoubleValidator is not properly implemented.

Btw, Number.class seems not the proper thing to rely, too.
http://docs.oracle.com/javase/7/docs/api/java/lang/Number.html

BigInteger and BigDecimal extends Number, so if we don't want to handle big numbers in configuration, we should have properly implemented DoubleValidator (It should check the value is not beyond the range of Double), and don't use Number.class to validate.

HeartSaVioR:
@jerrypeng 
For now I can make new issue to address it, and just vote +1 to this PR if you want.
Let me hear your thought about this. :)

jerrypeng:
Perhaps for the purposes of the config "topology.stats.sample.rate" and future number based configs we should have a PositiveNumberValidator.  There is already a PositiveIntegerValidator but it only deals with positive integers.  It seems to me that "topology.stats.sample.rate" can be set to some positive float/double or integer. So I think we should include the PositiveNumberValidator.  I have added a commit that includes the code for the PositiveNumberValidator.  What do you think?

HeartSaVioR:
@jerrypeng 
Great. It couldn't assert it is not big number or big decimal, but it's better than current.
+1.

HeartSaVioR:
@jerrypeng Sorry, but could you provide unit test for PositiveNumberValidator? Thanks!

jerrypeng:
added unit tests.  Please merge? or do I need to squash some of the commits?

storm-core/src/jvm/backtype/storm/ConfigValidation.java
```diff
@@ -28,7 +28,6 @@
     /**
      * Declares methods for validating configuration values.
      */
-    public static interface FieldValidator {
```
caofangkun: This interface should not be annotated.
Could you please have a check?

HeartSaVioR:
@jerrypeng There's a compile failure. Could you check it?

jerrypeng:
You are right @caofangkun, not sure how that line got deleted, but it shouldn't have been removed.  I added back and I think the tests will pass now

jerrypeng:
I don't know why the build is still failing. I just built and ran all the tests fine on my local machine.  I heard that since we just recently started using travis-ci that there are some bugs with it.

caofangkun:
I checked the CI build log and find 
https://travis-ci.org/apache/storm/jobs/74073153#L4132

and there's a PR  on this , @HeartSaVioR Could you please review #652 ?

HeartSaVioR:
+1. Build failure is unrelated. I did test before merge.

PR #660, opened 2015-07-31
STORM-933:NullPointerException during KafkaSpout deactivation

caofangkun:


HeartSaVioR:
We can add null check from KafkaSpout.commit() to save tiny memory (by lazy allocation), but it is really tiny so I think current approach is acceptable.
+1.

HeartSaVioR:
I merged this into master, 0.10.x-branch, 0.9.x-branch respectively.
Seems like Github mirror has issue again. When it becomes stable, your PR could be automatically marked as merged.

PR #661, opened 2015-07-31
[STORM-968] Adding support to generate the id based on names in Trident

snikhil5:
Spout ids based on names
screen shot 2015-07-31 at 2 30 26 pm


snikhil5:
Earlier it was just spout0..spout1 etc..
screen shot 2015-07-31 at 2 31 12 pm

revans2:
+1 it is better than it was before and the CI failure looks spurious.

PR #664, opened 2015-08-02
STORM-969: HDFS Bolt can end up in an unrecoverable state

dossett:
A few notes about this PR:
- I updated the storm-hdfs pom.xml to align with other external modules.  Most significant change was probably going from hdfs version 2.2 to ${hadoop.version} (i.e. currently 2.6)

- Many errors are recovered by forcing a file rotation which opens a new, valid File.  So the rotation now occurs either according to rotation policy or when a serious error happens.  Work could probably be done to reopen the same file name to reduce the number of rotations.

- Added unittests with MiniDFSCluster

### DIFF
--- a/external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/HdfsBolt.java
+++ b/external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/HdfsBolt.java
@@ -17,9 +17,11 @@
  */
 package org.apache.storm.hdfs.bolt;
 
+import backtype.storm.Config;
 import backtype.storm.task.OutputCollector;
 import backtype.storm.task.TopologyContext;
 import backtype.storm.tuple.Tuple;
+import backtype.storm.utils.TupleUtils;
 import org.apache.hadoop.fs.FSDataOutputStream;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
@@ -37,13 +39,19 @@ import java.io.IOException;
 import java.net.URI;
 import java.util.EnumSet;
 import java.util.Map;
+import java.util.List;
+import java.util.LinkedList;
 
 public class HdfsBolt extends AbstractHdfsBolt{
     private static final Logger LOG = LoggerFactory.getLogger(HdfsBolt.class);
+    private static final Integer DEFAULT_RETRY_COUNT = 3;
 
     private transient FSDataOutputStream out;
     private RecordFormat format;
     private long offset = 0;
+    private List<Tuple> tupleBatch = new LinkedList<>();
+    Integer tickTupleInterval = 0;
+    Integer fileRetryCount = DEFAULT_RETRY_COUNT;
 
     public HdfsBolt withFsUrl(String fsUrl){
         this.fsUrl = fsUrl;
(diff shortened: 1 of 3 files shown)


harshach:
@arunmahadevan since you've made changes to HdfsBolt as part of STORM-837 can you take look at this PR check if its already handled by your changes.

arunmahadevan:
@harshach the changes I made were in the trident implementation (HDFSState) which is independent of this. Anyways I reviewed the changes.

Overall it appears that the exceptions are now handled individually at write, sync and rotate phases. But I see a few issues with the change.
- the tuples are acked only on sync - if the tuple rates are low, they will never be acked and keep timing out and same tuples will be replayed again and again.
- there is an attempt to sync even when the write fails with an IOException. Since write already threw an IOException, chances are high that the sync would also fail with IOException.

I think it might be simpler to keep the existing logic and just rotate the file whenever we see an IOException (or maybe after a few times we repeatedly hit the IOException) and completely fail by propagating the exception up if the situation does not improve after a few rotations.

Also I see that the existing implementation acks the tuples before actually syncing them to disk, which might result in data loss. I think we should change this to ack only after an hsync and have a sync policy that considers both count and time based thresholds.

dossett:
@arunmahadevan Thank you for the feedback!  I have added a tick tuple feature to address your first point (I am already using this locally, I forgot to include it in this PR).

You are right that I attempt to sync even when the write fails.  I agree that this is unlikely to succeed, but it seems worthwhile to try to "save" as many tuples as possible.

With this tick tuple change I think the code is very close to your recommendation.  If the bolt is in a bad state the tick tuples will periodically try to rotate the file and all errors will be reported up through ```this.collector.reportError(e)```

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/HdfsBolt.java
```diff
@@ -80,6 +86,11 @@ public HdfsBolt addRotationAction(RotationAction action){
         return this;
     }
 
+    public HdfsBolt withTickTupleIntervalSeconds(int interval) {
```
arunmahadevan: Could give a more meaningful name to convey the actual usage (e.g `withFlushIntervalSeconds`). I think we need to have a default value for this to ensure a sync is `always` done even if user doesn't specify this option and also ensure the option value is within some lower and upper thresholds so that tuples are acked within TOPOLOGY_MESSAGE_TIMEOUT_SECS and a sync doesn't happen too frequently.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/HdfsBolt.java
```diff
@@ -88,35 +99,94 @@ public void doPrepare(Map conf, TopologyContext topologyContext, OutputCollector
...
         }
     }
 
+    private void rotateAndReset() throws IOException {
+        rotateOutputFile(); // synchronized
+        this.offset = 0;
+        this.rotationPolicy.reset();
+    }
+
+    private void syncAndAckTuples() throws IOException {
```
arunmahadevan: An optimization could be to do to the sync only if some tuples were actually written out since last sync.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/HdfsBolt.java
```diff
@@ -88,35 +99,94 @@ public void doPrepare(Map conf, TopologyContext topologyContext, OutputCollector
...
+                //The next tuple will almost certainly fail to write and/or sync, which force a rotation.  That
+                //will give rotateAndReset() a chance to work which includes creating a fresh file handle.
             }
-        } catch (IOException e) {
-            this.collector.reportError(e);
-            this.collector.fail(tuple);
         }
     }
 
+    private void rotateAndReset() throws IOException {
```
arunmahadevan: What would be the behavior if the file system goes into an irrecoverable state where writes are continuously failing? Wont it end up creating a lot of empty files? You might want to consider cleaning this up and also killing the task by throwing a runtime exception after 'n' failures and hope that the new worker might be able to start writing successfully.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/HdfsBolt.java
dossett: I like that change since it separates the function from the implementation.  A similar change could be made to HiveBolt, to which I also added a tick tuple functionality recently.  If this PR is otherwise ok, I'd like to open a new JIRA to change this bolt and HiveBolt at the same time.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/HdfsBolt.java
dossett: I suppose it's possible that empty files would be created, but if a filesystem is in such a bad state I would expect creating new files to fail as well.

Killing the task after enough failures is an interesting idea... what default value for # of attempts would you consider reasonable?

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/HdfsBolt.java
dossett: Agreed, just pushed a commit to do that.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/HdfsBolt.java
arunmahadevan: Assuming that would address default, min and max values for the interval, should be fine.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/HdfsBolt.java
arunmahadevan: You could have a low default (say 10 continuous failures or so) and also add an option for the users to change this. @harshach can you recommend a default number of retries after which the task can exit so that storm would spawn the task in another worker where it might succeed?

harshach: @dossett 
"I suppose it's possible that empty files would be created, but if a filesystem is in such a bad state I would expect creating new files to fail as well."
This might bring down namenode :)  with too many files.

as @arunmahadevan pointed we should have retryCount I would start with 3 failed attempts and a configurable option

dossett:
Thank you @harshach and @arunmahadevan, I will incorporate this feedback soon.

dossett:
@harshach @arunmahadevan Changes made based on feedback, PR also rebased and squashed to one commit.  Additional feedback welcome, thank you!

arunmahadevan:
+1 overall the patch looks good once the [comment] (https://github.com/apache/storm/pull/664#commitcomment-13003095) is addressed

dossett:
@arunmahadevan Our approach is to set the tick tuple frequency to be half of the message timeout setting for the topology.  Can the bolt get access to that topology setting? prepare() passes a TopologyContext to the bolt but I don't see a way to get configuration information out of it.  However, I am not that familiar with TopologyContext.

If that's not possible, the bolt could just set a flush frequency of 15 seconds since the storm timeout default is 30 seconds and assume that anyone changing the timeout setting should also be adjusting the flush frequency as well

arunmahadevan:
The value of `topology.message.timeout.secs` can be read from the `conf` that gets passed as the first parameter to `doPrepare`

dossett:
@arunmahadevan So it can!  PR updated.

dossett:
@arunmahadevan @harshach Do you have any further comments about my updated PR? Thanks!

arunmahadevan:
Looks good to me.

dossett:
@harshach (or other commiters) do you have feedback about this PR?

Anecdotally, this has been very useful to us in production. We had an HDFS restart, which created the exact situation I tested with (failed writes due to HDFS safe mode) but the bolt recovered without a topology restart.

harshach:
+1

harshach:
Thanks @dossett merged into master.

PR #668, opened 2015-08-06
STORM-828 HdfsBolt takes a lot of configuration, need good defaults

redsanket:
Removing configs from HDFSFileTopology example. I have made neccessary config and constructor changes as per my understanding. It will be nice to know if there is anything I need to look at.

### DIFF
--- a/external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/CSVFileBolt.java
+++ b/external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/CSVFileBolt.java
@@ -1,0 +1,32 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.storm.hdfs.bolt;
+
+import org.apache.storm.hdfs.bolt.format.DefaultFileNameFormat;
+import org.apache.storm.hdfs.bolt.format.RecordFormat;
+import org.apache.storm.hdfs.bolt.rotation.TimedRotationPolicy;
+import org.apache.storm.hdfs.bolt.sync.CountSyncPolicy;
+import org.apache.storm.hdfs.common.rotation.MoveFileAction;
+
+public class CSVFileBolt extends HdfsBolt {
+  private static String fileExtension = ".csv";
+
+  public CSVFileBolt(String sourceDir, String destDir) {
+    super(sourceDir, destDir, fileExtension);
+  }
+}
--- a/external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/HdfsBolt.java
+++ b/external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/HdfsBolt.java
@@ -25,10 +25,15 @@ import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.hdfs.client.HdfsDataOutputStream;
 import org.apache.hadoop.hdfs.client.HdfsDataOutputStream.SyncFlag;
+import org.apache.storm.hdfs.bolt.format.DefaultFileNameFormat;
+import org.apache.storm.hdfs.bolt.format.DelimitedRecordFormat;
 import org.apache.storm.hdfs.bolt.format.FileNameFormat;
 import org.apache.storm.hdfs.bolt.format.RecordFormat;
 import org.apache.storm.hdfs.bolt.rotation.FileRotationPolicy;
+import org.apache.storm.hdfs.bolt.rotation.TimedRotationPolicy;
+import org.apache.storm.hdfs.bolt.sync.CountSyncPolicy;
 import org.apache.storm.hdfs.bolt.sync.SyncPolicy;
+import org.apache.storm.hdfs.common.rotation.MoveFileAction;
 import org.apache.storm.hdfs.common.rotation.RotationAction;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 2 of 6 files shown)


external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/CSVFileBolt.java
```diff
@@ -0,0 +1,32 @@
...
+import org.apache.storm.hdfs.bolt.sync.CountSyncPolicy;
+import org.apache.storm.hdfs.common.rotation.MoveFileAction;
+
+public class CSVFileBolt extends HdfsBolt {
+  private static String fileExtension = ".csv";
+
+  public CSVFileBolt(String sourceDir, String destDir) {
+    super(sourceDir, destDir, fileExtension);
+  }
+}
```
dossett: Are there any guarantees that this bolt actually writes out comma separated values?  I don't see that, which makes the class name somewhat misleading.

Similar observation about the TSV bolt further down.

redsanket: I guess I could set the RecordDefaultDelimiter to ",". Yes, thanks for pointing it out, I will make the changes and it up soon

dossett: I think it would take much more work than that and a clearer definition of what the CSVFileBolt does, IMO.  What should the user expect from the behavior of CSVBolt.emit() ?  Will it take arbitrary values from a tuple and join them together with commas?  Will it escape characters? etc.

Another approach would be to create a separate bolt that does nothing but generate CSV output and then passes the results to any other bolt the user wanted.  I often create pre-processing bolts like that.

redsanket: That's true, I should have made that observation, I could not possibly guarantee, I was just trying to refactor the code. I will make changes to PR soon and post it. Thanks again for taking time to review the code

redsanket: For futher clarification, I was further thinking over this issue and I presume currently the spout emits tuples in the form of fields and the csv or tsv bolt joins them with the specified record format delimiter. If it is a ",", all tuples will be appended by a comma. The HdfsBolt actually does this implementation when the execute is called upon it. The TSV or CSV are just abstractions to get the intended values based on the record delimiter. Can you please let me know what exactly has to be done with an example, I thought I got your point but right now I cannot clearly picture it. It will be great if you can get back with a reply. Thanks a lot.

harshach:
@redsanket why we need CSVBolt and TSVBolt when there is DelimitedRecordFormat which users can configure

redsanket:
For futher clarification, I was further thinking over this issue and I presume currently the spout emits tuples in the form of fields and the csv or tsv bolt joins them with the specified record format delimiter. If it is a ",", all tuples will be appended by a comma. The HdfsBolt actually does this implementation when the execute is called upon it. The TSV or CSV are just abstractions to get the intended values based on the record delimiter. Can you please let me know what exactly has to be done with an example, I thought I got your point but right now I cannot clearly picture it. It will be great if you can get back with a reply. Thanks a lot.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/CSVFileBolt.java
redsanket: For futher clarification, I was further thinking over this issue and I presume currently the spout emits tuples in the form of fields and the csv or tsv bolt joins them with the specified record format delimiter. If it is a ",", all tuples will be appended by a comma. The HdfsBolt actually does this implementation when the execute is called upon it. The TSV or CSV are just abstractions to get the intended values based on the record delimiter. Can you please let me know what exactly has to be done with an example, I thought I got your point but right now I cannot clearly picture it. It will be great if you can get back with a reply. Thanks a lot.

redsanket:
ok that seems to be true, I will change that. Mostly it will be down to refactoring a piece of code, we might not have to make it more specific then. Thanks

harshach:
@redsanket I am not sure why this patch needs to change all these files. Why can't we just add default values to the variables here https://github.com/apache/storm/blob/master/external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/AbstractHdfsBolt.java#L53 similar to what we have in hive connector https://github.com/apache/storm/blob/master/external/storm-hive/src/main/java/org/apache/storm/hive/common/HiveOptions.java#L33

redsanket:
Yes I have made the changes, I will change my pull request soon

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/HdfsBolt.java
```diff
@@ -44,6 +49,25 @@
     private transient FSDataOutputStream out;
     private RecordFormat format;
     private long offset = 0;
+    private static String defaultSourceDir = "/tmp/source";
+    private static String defaultDestDir = "/tmp/dest";
+    private static String defaultFileExtension = ".txt";
+
+
```
revans2: Extra blank line.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/CSVFileBolt.java
revans2: If CSV and TSV formats are a bit controversial perhaps we can file a separate JIRA to implement them properly, with escaping, etc. and remove them from here.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/HdfsBolt.java
```diff
@@ -44,6 +49,25 @@
...
+    public HdfsBolt() {
```
revans2: I'm not sure a no argument constructor makes a lot of since.  I think you need to specify at a minimum the destination directory.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/HdfsBolt.java
```diff
@@ -44,6 +49,25 @@
...
+      this(defaultSourceDir, defaultDestDir, defaultFileExtension);
+    }
+
+    public HdfsBolt(String defaultSourceDir, String defaultDestDir, String fileExtension) {
+      this.withRotationPolicy(new TimedRotationPolicy(1.0f, TimedRotationPolicy.TimeUnit.MINUTES))
+          .withConfigKey("hdfs.config")
```
revans2: I don't really like this as a default I thought null would be a better default.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/HdfsBolt.java
```diff
@@ -44,6 +49,25 @@
...
+    private static String defaultSourceDir = "/tmp/source";
+    private static String defaultDestDir = "/tmp/dest";
+    private static String defaultFileExtension = ".txt";
+
+
+    public HdfsBolt() {
+      this(defaultSourceDir, defaultDestDir, defaultFileExtension);
+    }
+
+    public HdfsBolt(String defaultSourceDir, String defaultDestDir, String fileExtension) {
```
revans2: Can we call sourceDir, a stagingDir instead?  To me it seems more logical.

revans2: And not use "default" in the variable names.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/HdfsBolt.java
```diff
@@ -44,6 +49,25 @@
...
+      this.withRotationPolicy(new TimedRotationPolicy(1.0f, TimedRotationPolicy.TimeUnit.MINUTES))
+          .withConfigKey("hdfs.config")
+          .withRecordFormat(new DelimitedRecordFormat().withRecordDelimiter("|"))
+          .withFileNameFormat(new DefaultFileNameFormat()
+                  .withPath(defaultSourceDir)
+                  .withExtension(fileExtension))
+          .withSyncPolicy(new CountSyncPolicy(1000))
+          .addRotationAction(new MoveFileAction().toDestination(defaultDestDir));
```
revans2: I would like to see it that if there is no stagingDir specified that the destDir is used for the DefaultFileNameFormat path and there is no MoveFileAction.  If there is a  stagingDir then it behaved like this.

revans2:
Sorry it took me so long to review this.  For the most part things look good.  I made a few comments on the HdfBolt that I would like to see reflected in the SequenceFileBolt too.  Also an upmerge would be good.

Thank you for being so patient with us.

dossett:
@redsanket Are you still working on this?

PR #672, opened 2015-08-11
STORM-977 Correct log prefix (-15) when sending -9 signal to worker

gliptak:


storm-core/src/clj/backtype/storm/daemon/supervisor.clj
```diff
@@ -262,7 +262,7 @@
       (psim/kill-process thread-pid))
     (doseq [pid pids]
       (if as-user
-        (worker-launcher-and-wait conf user ["signal" pid "9"] :log-prefix (str "kill -15 " pid))
+        (worker-launcher-and-wait conf user ["signal" pid "9"] :log-prefix (str "kill -9 " pid))
```
caofangkun: I think should change the signal value but not the ```log-prefix```
should be 
 ```(worker-launcher-and-wait conf user ["signal" pid "15"] :log-prefix (str "kill -15 " pid))```

caofangkun:
The travis ci build failed does not make sense ,
Does ``` export STORM_TEST_TIMEOUT_MS=100000``` not long enough ?

storm-core/src/clj/backtype/storm/daemon/supervisor.clj
d2r: @caofangkun You are correct.  The JIRA title is wrong.

We want to send the SIGTERM, then wait, then send SIGKILL.

HeartSaVioR:
@caofangkun Build issue is resolved now, but old PRs still need to be upmerged to take effect.

d2r:
+1

PR #682, opened 2015-08-13
[STORM-990] Refactored TimeCacheMap to extend RotatingMap.

ddebree:
This should allow current users of the TimeCacheMap to progressively migrate off the deprecated TimeCacheMap to the simpler RotatingMap.
 This also improves code reuse since many of the methods were shared.

### DIFF
--- a/storm-core/src/jvm/backtype/storm/utils/RotatingMap.java
+++ b/storm-core/src/jvm/backtype/storm/utils/RotatingMap.java
@@ -32,18 +32,22 @@ import java.util.Map.Entry;
  *
  * The advantage of this design is that the expiration thread only locks the object
  * for O(1) time, meaning the object is essentially always available for gets/puts.
+ *
+ * Note: This class is not thread-safe since it does not protect against changes to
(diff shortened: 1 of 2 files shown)


knusbaum:
For large changes like this and #683, Please file a jira at https://issues.apache.org/jira/browse/STORM/

ddebree:
Issue raised: https://issues.apache.org/jira/browse/STORM-990

knusbaum:
@ddebree Thanks. I should have mentioned it, but we also include the issue name in the title of our pull requests by convention.
```
[STORM-990] Refactored TimeCacheMap to extend RotatingMap.
```

knusbaum:
I'm in favor of using RotatingMap to implement TimeCacheMap, but I'd rather have a RotatingMap as a member of TimeCacheMap than as a superclass.

ddebree:
I have updated the class to rather use the RotatingMap as a member variable. It comes out a lot more similar to the original.

knusbaum:
+1 Looks good to me.

PR #683, opened 2015-08-13
[STORM-991] General cleanup of the generics.

ddebree:
Generally added generics to classes within the storm.trident.spout package. This generally removes some of the casts that were in place, and simplifies the code slightly. 
A typical example of what was done is 

    ITridentSpout _spout;

was converted to:

    ITridentSpout<Object> _spout;

### DIFF
--- a/storm-core/src/jvm/storm/trident/spout/RichSpoutBatchTriggerer.java
+++ b/storm-core/src/jvm/storm/trident/spout/RichSpoutBatchTriggerer.java
@@ -56,7 +56,7 @@ public class RichSpoutBatchTriggerer implements IRichSpout {
     @Override
     public void open(Map conf, TopologyContext context, SpoutOutputCollector collector) {
         _delegate.open(conf, context, new SpoutOutputCollector(new StreamOverrideCollector(collector)));
-        _outputTasks = new ArrayList<Integer>();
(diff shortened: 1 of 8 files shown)


storm-core/src/jvm/storm/trident/spout/RichSpoutBatchTriggerer.java
```diff
@@ -56,7 +56,7 @@ public RichSpoutBatchTriggerer(IRichSpout delegate, String streamName, String ba
     @Override
     public void open(Map conf, TopologyContext context, SpoutOutputCollector collector) {
         _delegate.open(conf, context, new SpoutOutputCollector(new StreamOverrideCollector(collector)));
-        _outputTasks = new ArrayList<Integer>();
+        _outputTasks = new ArrayList<>();
```
dossett: On a different PR someone pointed out that this is syntax required Java 1.7  That seems ok to me since the top level pom.xml specifies 1.7 in the compile plugin, but I wanted to just point that fact out.

I support moving to the diamond notation since it makes the code much cleaner, in my opinion.

ddebree:
Issue raised on Jira: https://issues.apache.org/jira/browse/STORM-991

HeartSaVioR:
+1

PR #579, opened 2015-06-05
STORM-850 Convert storm-core's logback-test.xml to log4j2-test.xml

HeartSaVioR:
In order to prevent Travis CI complaining about log exceeding, we need to apply this.

Please review and comment. Thanks in advance!

PR #684, opened 2015-08-13
STORM-976

YvonneIronberg:


storm-core/src/clj/backtype/storm/daemon/supervisor.clj
```diff
@@ -661,8 +661,8 @@
           storm-conf-file (System/getProperty "storm.conf.file")
           storm-log-dir (or (System/getProperty "storm.log.dir") (str storm-home file-path-separator "logs"))
           storm-conf (read-storm-config)
-          storm-log-conf-dir (storm-conf "storm.logback.conf.dir")
-          storm-logback-conf-dir (or storm-log-conf-dir (str storm-home file-path-separator "log4j2"))
+          storm-log-conf-dir (storm-conf "storm.log4j2.conf.dir")
```
knusbaum: Hey, there. I know you're just updating the existing code, but would you mind adding `"storm.log4j2.conf.dir"` to Config.java and removing the `storm-conf (read-storm-config)` binding?

Both of those lines were added in #373 and went uncaught. Instead of reading from `storm-conf`, just read from `conf`.

YvonneIronberg: Thanks for the review. Sure. Updated the patch.

storm-core/src/jvm/backtype/storm/Config.java
```diff
@@ -145,6 +145,12 @@
     public static final String STORM_LOCAL_DIR = "storm.local.dir";
     public static final Object STORM_LOCAL_DIR_SCHEMA = String.class;
 
+		/**
+		 * A directory that holds configuration files for log4j2.
+		 */
+		public static final String STORM_LOG4J2_CONF_DIR = "storm.log4j2.conf.dir";
+		public static final Object STORM_LOG4J2_CONF_DIR_SCHEMA = String.class;
+
```
d2r: indentation

YvonneIronberg: Thanks. Done.

bin/storm-config.cmd
```diff
@@ -83,10 +83,10 @@ if not defined STORM_LOG_DIR (
 )
 
 @rem
-@rem retrieve storm.logback.conf.dir from conf file
+@rem retrieve storm.log4j2.conf.dir from conf file
 @rem
 
-"%JAVA%" -client -Dstorm.options= -Dstorm.conf.file= -cp "%CLASSPATH%" backtype.storm.command.config_value storm.logback.conf.dir > %CMD_TEMP_FILE%
+"%JAVA%" -client -Dstorm.options= -Dstorm.conf.file= -cp "%CLASSPATH%" backtype.storm.command.config_value storm.log4j2.conf.dir > %CMD_TEMP_FILE%
```
d2r: I was looking around and found other usages of LOCKBACK in the storm-config.cmd file.  May as well change those while we are changing the others.

d2r:
I looked through the branch and found some I hadn't noticed before:

* [logback.xml](https://github.com/apache/storm/blob/aa308e11685b8b7945f2d4963e9f63dfd322087b/external/flux/flux-core/src/test/resources/logback.xml) file under flux is still there.  We could fix this separately in another JIRA since it is external.  If we can fix it now, then that is even better.
* [FAQ.md](https://github.com/apache/storm/blob/aa308e11685b8b7945f2d4963e9f63dfd322087b/docs/documentation/FAQ.md) already mentions moving to logback from log4j (1.x), so we should update that one part since we have moved again to log4j2.
* [SECURITY.md](https://github.com/apache/storm/blob/aa308e11685b8b7945f2d4963e9f63dfd322087b/SECURITY.md) also mentions logback in one spot.  We could change it or make it generic.

It would be nice to take care of these three as part of this work, otherwise we can create another JIRA Issue to track.

YvonneIronberg:
Thanks for the suggestions. Addressed them in the latest commit. logback.xml is converted to an equivalent log4j2.xml according to their docs.
* [logback](http://logback.qos.ch/manual/configuration.html).
* [log4j2 config](https://logging.apache.org/log4j/2.x/manual/configuration.html).

bin/storm-config.cmd
ptgoetz: +1 for fixing these things now. Configuration (logs and otherwise) are a big stumbling block for new users. Anything we can do to alleviate such things would be good.

YvonneIronberg:
Passed Travis CI build.

Changing external/flux/flux-core/src/test/resources/logback.xml to log4j2.xml is in accordance with [STORM-850](https://issues.apache.org/jira/browse/STORM-850) that [changed storm-core/test/resources/logback-test.xml to log4j2-test.xml in a similar way](https://github.com/apache/storm/pull/579/files).

knusbaum:
+1 Looks good to me.

d2r:
Looks good to me, +1
Thanks @YvonneIronberg

d2r:
@YvonneIronberg , I did not see an Apache JIRA user name for you, so I have resolved STORM-976 as Unassigned.

If you would like, share your JIRA user name so that we can assign the STORM-976 to you.

YvonneIronberg:
Thanks, @d2r! I've assigned STORM-976 to me on Apache jira.

PR #685, opened 2015-08-14
ShellBolt should treat ack and fail as heartbeat

itaifrenkel:
Python queues up requests, and it can handle many requests before reaching the __heartbeat signal, so the shellbolt must treat ack/fail as heartbeat too.


HeartSaVioR:
@itaifrenkel
Thanks for the work, but it was already addressed.
https://issues.apache.org/jira/browse/STORM-742

Could you review https://github.com/apache/storm/pull/497 and comment?

itaifrenkel:
@HeartSaVioR  I also recommend this fix https://github.com/forter/storm/commit/70f5689b15ab5c6092ed2310fdde4820a4f7bc0d
Otherwise the bolt can fail when preparing in a busy build server (race condition between reader and heartbeat threads)

d2r:
@itaifrenkel,

If this is still an issue, would you create a Jira for it and fix the conflicts with the master branch?

Otherwise, if the problem is resolved, would you close this pull request?

PR #686, opened 2015-08-14
STORM-965: Fix excessive logging

priyank5485:
The reason for this pull request is that when a non-secure client tries to connect to nimbus host in a secure cluster, an exception was being thrown and a null transport object was being returned. This null object in turn was being used by thrift library class and threw a null pointer exception. Both the classes where exception was thrown and caught were logging this exception, creating unnecessary lengthy log statements. Disabling logging in storm code by changing it from error to debug took care of one part but to work around the issue of exception being logged in the thrift jar a NoOpTTrasport.java had to be created. Two main things to consider during review
1. Change from Logger.error to Logger.debug in KerberosSaslTransportPlugin.java
2. Early return false in SaslTransportPlugin.java which is presumed and tested with one nimbus client to work okay. 

Note: TProcess interface process method does not have documentation saying what the returned boolean indicates.


harshach:
+1

PR #699, opened 2015-08-26
[STORM-1010] each KafkaBolt have its own properties

Renkai:


harshach:
@Renkai whats the need for this? why can't you add multiple bolts with its own parallelism. IMO this goes against the storm api.

Renkai:
@harshach 
Different KafkaBolts  may need to send topic to different kafka clusters,use different serializers,and have different params for performance tunning.
In current version there is no way to do that.

revans2:
The changes look fine to me.  +1

@harshach we did a similar thing for the HDFS Bolt.
https://github.com/apache/storm/blob/ffb1562cb0eb3fbd916423136fcd3c181fbf6212/external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/HdfsBolt.java#L53-L56

Each bolt needs to be able to be configured independently of all other bolts.  Putting it in the conf is a is one way to do it, but I have seen a lot of bolts do it through a setter too. I personally would prefer to support both through a config and through an API.  The only trick is with consistency and ordering in the case of conflicts.

Parth-Brahmbhatt:
+1.

harshach:
@revans2 @Renkai makes sense. Instead of using constructor can we should with Options like other bolts are using. probably "withProducerProperties" method and remove the topology level configs so that there isn't two ways configure the bolt.

harshach:
I am fine with the current approach but we need to standardize one way configuring the bolts as each connector follows its own pattern right now and which can be confusing to the users.

Renkai:
@harshach I will push another merge requests,make it harmonic with current code.

Renkai:
Travis CI shows test fails in with jdk8,but I think it's not problems with my commits,tests all passed on my Mac.

harshach:
@Renkai those failed tests are intermittent test failures in storm-core not related to the patch.

HeartSaVioR:
+1.

PR #700, opened 2015-08-26
[STORM-886] Automatic Back Pressure (ABP)

zhuoliu:
This new feature is aimed for automatic flow control through the topology DAG since different components may have unmatched tuple processing speed. Currently, the tuples may get dropped if the downstream components can not process as quickly, thereby causing a waste of network bandwidth and processing capability. In addition, it is difficult to tune the max.spout.pending parameter for best backpressure performance. Therefore, an automatic back pressure scheme is highly desirable.

Recently, Heron proposed a form of back pressure that does not rely on acking or max spout pending. Instead spouts throttle not only when max.spout.pending is hit, but also if any bolt has gone over a high water mark in their input queue, and has not yet gone below a low water mark again. There is a lot of room for potential improvement here around control theory and having spouts only respond to downstream bolts backing up, but a simple bang-bang controller like this is a great start.

Our ABP scheme implements a light-weight yet efficient back pressure scheme. It monitors certain queues in executors and exploits the callback schemes on ZooKeeper and disruptor queue for a fast-responding (in a push manner) flow control.

Please check the attached figures for more details about the implementation.
https://issues.apache.org/jira/secure/attachment/12752518/an%20simple%20example%20for%20backpressure.png
image

### DIFF
--- a/conf/defaults.yaml
+++ b/conf/defaults.yaml
@@ -145,6 +145,13 @@ task.heartbeat.frequency.secs: 3
 task.refresh.poll.secs: 10
 task.credentials.poll.secs: 30
 
+# now should be null by default
+topology.backpressure.enable: true
+backpressure.worker.high.watermark: 0.9
+backpressure.worker.low.watermark: 0.4
+backpressure.executor.high.watermark: 0.9
+backpressure.executor.low.watermark: 0.4
+
 zmq.threads: 1
 zmq.linger.millis: 5000
 zmq.hwm: 0
--- a/conf/defaults.yaml
+++ b/conf/defaults.yaml
@@ -147,10 +147,8 @@ task.credentials.poll.secs: 30
 
 # now should be null by default
 topology.backpressure.enable: true
-backpressure.worker.high.watermark: 0.9
-backpressure.worker.low.watermark: 0.4
-backpressure.executor.high.watermark: 0.9
-backpressure.executor.low.watermark: 0.4
+backpressure.disruptor.high.watermark: 0.9
+backpressure.disruptor.low.watermark: 0.4
 
 zmq.threads: 1
 zmq.linger.millis: 5000
--- a/storm-core/src/clj/backtype/storm/daemon/executor.clj
+++ b/storm-core/src/clj/backtype/storm/daemon/executor.clj
@@ -28,7 +28,7 @@
   (:import [backtype.storm.grouping CustomStreamGrouping])
   (:import [backtype.storm.task WorkerTopologyContext IBolt OutputCollector IOutputCollector])
   (:import [backtype.storm.generated GlobalStreamId])
-  (:import [backtype.storm.utils Utils MutableObject RotatingMap RotatingMap$ExpiredCallback MutableLong Time DisruptorQueue])
+  (:import [backtype.storm.utils Utils MutableObject RotatingMap RotatingMap$ExpiredCallback MutableLong Time DisruptorQueue WorkerBackpressureThread])
   (:import [com.lmax.disruptor InsufficientCapacityException])
   (:import [backtype.storm.serialization KryoTupleSerializer KryoTupleDeserializer])
   (:import [backtype.storm.daemon Shutdownable])
@@ -278,12 +278,12 @@
       (if (not @(:backpressure executor-data))
         (do (reset! (:backpressure executor-data) true)
             (log-debug "executor " (:executor-id executor-data) " is congested, set backpressure flag true")
-            (disruptor/notify-backpressure-checker (:backpressure-trigger (:worker executor-data))))))
+            (WorkerBackpressureThread/notifyBackpressureChecker (:backpressure-trigger (:worker executor-data))))))
     (fn []
       "When receive queue is below lowWaterMark"
       (if @(:backpressure executor-data)
         (do (reset! (:backpressure executor-data) false)
-            (disruptor/notify-backpressure-checker (:backpressure-trigger (:worker executor-data))))))))
+            (WorkerBackpressureThread/notifyBackpressureChecker (:backpressure-trigger (:worker executor-data))))))))
 
 (defn start-batch-transfer->worker-handler! [worker executor-data]
   (let [worker-transfer-fn (:transfer-fn worker)
@@ -377,8 +377,8 @@
 
         disruptor-handler (mk-disruptor-backpressure-handler executor-data)
         _ (.registerBackpressureCallback (:receive-queue executor-data) disruptor-handler)
-        _ (-> (.setHighWaterMark (:receive-queue executor-data) ((:storm-conf executor-data) BACKPRESSURE-WORKER-HIGH-WATERMARK))
-              (.setLowWaterMark ((:storm-conf executor-data) BACKPRESSURE-WORKER-LOW-WATERMARK))
+        _ (-> (.setHighWaterMark (:receive-queue executor-data) ((:storm-conf executor-data) BACKPRESSURE-DISRUPTOR-HIGH-WATERMARK))
+              (.setLowWaterMark ((:storm-conf executor-data) BACKPRESSURE-DISRUPTOR-LOW-WATERMARK))
               (.setEnableBackpressure ((:storm-conf executor-data) TOPOLOGY-BACKPRESSURE-ENABLE)))
 
         ;; starting the batch-transfer->worker ensures that anything publishing to that queue 
@@ -645,7 +645,6 @@
                     (log-message "Activating spout " component-id ":" (keys task-datas))
                     (fast-list-iter [^ISpout spout spouts] (.activate spout)))
                
-                    ;; (log-message "Spout executor " (:executor-id executor-data) " found throttle-on, now suspends sending tuples")
                   (fast-list-iter [^ISpout spout spouts] (.nextTuple spout)))
                 (do
                   (when @last-active
--- a/storm-core/src/clj/backtype/storm/daemon/executor.clj
+++ b/storm-core/src/clj/backtype/storm/daemon/executor.clj
@@ -28,7 +28,7 @@
   (:import [backtype.storm.grouping CustomStreamGrouping])
   (:import [backtype.storm.task WorkerTopologyContext IBolt OutputCollector IOutputCollector])
   (:import [backtype.storm.generated GlobalStreamId])
-  (:import [backtype.storm.utils Utils MutableObject RotatingMap RotatingMap$ExpiredCallback MutableLong Time])
+  (:import [backtype.storm.utils Utils MutableObject RotatingMap RotatingMap$ExpiredCallback MutableLong Time DisruptorQueue])
   (:import [com.lmax.disruptor InsufficientCapacityException])
   (:import [backtype.storm.serialization KryoTupleSerializer KryoTupleDeserializer])
   (:import [backtype.storm.daemon Shutdownable])
@@ -170,7 +170,8 @@
 (defprotocol RunningExecutor
   (render-stats [this])
   (get-executor-id [this])
-  (credentials-changed [this creds]))
+  (credentials-changed [this creds])
+  (get-backpressure-flag [this]))
 
 (defn throttled-report-error-fn [executor]
   (let [storm-conf (:storm-conf executor)
@@ -261,8 +262,28 @@
                                ((:suicide-fn <>))))
      :deserializer (KryoTupleDeserializer. storm-conf worker-context)
      :sampler (mk-stats-sampler storm-conf)
+     :backpressure (atom false)
+     :spout-throttling-metrics (if (= executor-type :spout) 
+                                (builtin-metrics/make-spout-throttling-data)
+                                nil)
      ;; TODO: add in the executor-specific stuff in a :specific... or make a spout-data, bolt-data function?
      )))
+
+(defn- mk-disruptor-backpressure-handler [executor-data]
+  "make a handler for the executor's receive disruptor queue to
+  check highWaterMark and lowWaterMark for backpressure"
+  (disruptor/disruptor-backpressure-handler
+    (fn []
+      "When receive queue is above highWaterMark"
+      (if (not @(:backpressure executor-data))
+        (do (reset! (:backpressure executor-data) true)
+            (log-debug "executor " (:executor-id executor-data) " is congested, set backpressure flag true")
+            (disruptor/notify-backpressure-checker (:backpressure-trigger (:worker executor-data))))))
+    (fn []
+      "When receive queue is below lowWaterMark"
+      (if @(:backpressure executor-data)
+        (do (reset! (:backpressure executor-data) false)
+            (disruptor/notify-backpressure-checker (:backpressure-trigger (:worker executor-data))))))))
 
 (defn start-batch-transfer->worker-handler! [worker executor-data]
   (let [worker-transfer-fn (:transfer-fn worker)
@@ -353,6 +374,13 @@
         report-error-and-die (:report-error-and-die executor-data)
         component-id (:component-id executor-data)
 
+
+        disruptor-handler (mk-disruptor-backpressure-handler executor-data)
+        _ (.registerBackpressureCallback (:receive-queue executor-data) disruptor-handler)
+        _ (-> (.setHighWaterMark (:receive-queue executor-data) ((:storm-conf executor-data) BACKPRESSURE-WORKER-HIGH-WATERMARK))
+              (.setLowWaterMark ((:storm-conf executor-data) BACKPRESSURE-WORKER-LOW-WATERMARK))
+              (.setEnableBackpressure ((:storm-conf executor-data) TOPOLOGY-BACKPRESSURE-ENABLE)))
+
         ;; starting the batch-transfer->worker ensures that anything publishing to that queue 
         ;; doesn't block (because it's a single threaded queue and the caching/consumer started
         ;; trick isn't thread-safe)
@@ -369,7 +397,7 @@
       (render-stats [this]
         (stats/render-stats! (:stats executor-data)))
       (get-executor-id [this]
-        executor-id )
+        executor-id)
       (credentials-changed [this creds]
         (let [receive-queue (:receive-queue executor-data)
               context (:worker-context executor-data)]
@@ -377,6 +405,8 @@
             receive-queue
             [[nil (TupleImpl. context [creds] Constants/SYSTEM_TASK_ID Constants/CREDENTIALS_CHANGED_STREAM_ID)]]
               )))
+      (get-backpressure-flag [this]
+        @(:backpressure executor-data))
       Shutdownable
       (shutdown
         [this]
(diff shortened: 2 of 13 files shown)


conf/defaults.yaml
```diff
@@ -140,6 +140,14 @@ task.heartbeat.frequency.secs: 3
 task.refresh.poll.secs: 10
 task.credentials.poll.secs: 30
 
+# now should be null by default
+topology.backpressure.enable: true
+backpressure.worker.high.watermark: 0.9
+backpressure.worker.low.watermark: 0.4
+backpressure.executor.high.watermark: 0.9
+backpressure.executor.low.watermark: 0.4
+backpressure.spout.suspend.time.ms: 100
```
revans2: This seems way too high.  It is not that expensive to poll at a 1ms interval in the spout waiting for back-pressure to be turned off.

storm-core/src/clj/backtype/storm/daemon/executor.clj
```diff
@@ -602,7 +607,15 @@
                     (log-message "Activating spout " component-id ":" (keys task-datas))
                     (fast-list-iter [^ISpout spout spouts] (.activate spout)))
                
-                  (fast-list-iter [^ISpout spout spouts] (.nextTuple spout)))
+                  (if (and
+                        ((:storm-conf executor-data) TOPOLOGY-BACKPRESSURE-ENABLE)
+                        @(:throttle-on (:worker executor-data))
+                        suspend-time
+                        (not= suspend-time 0))
+                    (do
```
revans2: Can we move this conditional up to be part of the overflow-buffer/max-spout-pending check?  It feels like it fits better there, and that we don't want to output anything when back-pressure is on, instead of just slowing down how quickly we output.

Also instead of logging something, can we look at having a metric instead that shows how may times we called paused for a given reason.  Logging is not going to really give us the picture we want and is going to be difficult to follow.

storm-core/src/clj/backtype/storm/daemon/executor.clj
```diff
@@ -671,7 +689,19 @@
...
+                                         (> (.population receive-queue) high-watermark) 
+                                         (not @(:backpressure executor-data)))
+                                  (do (reset! (:backpressure executor-data) true)
+                                      (log-debug "executor " (:executor-id executor-data) " is congested, set backpressure flag true")
+                                      (disruptor/notify-backpressure-checker (:backpressure-trigger (:worker executor-data)))))
+                                (if (and ((:storm-conf executor-data) TOPOLOGY-BACKPRESSURE-ENABLE) 
+                                         (< (.population receive-queue) low-watermark) 
+                                         @(:backpressure executor-data))
+                                  (do (reset! (:backpressure executor-data) false)
+                                      (disruptor/notify-backpressure-checker (:backpressure-trigger (:worker executor-data)))))
```
revans2: I am a bit confused.  Why are we doing this as part of the metrics tick?  What does that have to do with back pressure at all?

storm-core/src/clj/backtype/storm/daemon/worker.clj
```diff
@@ -114,12 +114,34 @@
...
+            (reset! (:backpressure worker) false))) ;; no executor has backpressure set
+        ;; update the worker's backpressure flag to zookeeper here
+        (.worker-backpressure! storm-cluster-state storm-id assignment-id port @(:backpressure worker))  
+        ))))
+
 (defn mk-transfer-fn [worker]
   (let [local-tasks (-> worker :task-ids set)
         local-transfer (:transfer-local-fn worker)
-        ^DisruptorQueue transfer-queue (:transfer-queue worker)
+        transfer-queue (:transfer-queue worker)
```
revans2: Why did we drop the type hint?

storm-core/src/clj/backtype/storm/daemon/worker.clj
```diff
@@ -114,12 +114,34 @@
...
+      (let [storm-id (:storm-id worker)
+            assignment-id (:assignment-id worker)
+            port (:port worker)
+            storm-cluster-state (:storm-cluster-state worker)]
+        (if executors 
+          (if (reduce #(or %1 %2) (map #(.get-backpressure-flag %1) executors))
+            (reset! (:backpressure worker) true)   ;; at least one executor has set backpressure
+            (reset! (:backpressure worker) false))) ;; no executor has backpressure set
+        ;; update the worker's backpressure flag to zookeeper here
+        (.worker-backpressure! storm-cluster-state storm-id assignment-id port @(:backpressure worker))
```
revans2: I think we have an opportunity here to not put so much load on ZK.  We should be able to keep cached if we think back-pressure is on or not for this worker, and only update if we think it is changing, instead of reading the state each time and updating it, if it is different.  This is being called at least ever 100 ms, which is going to put more load on ZK then I like.

storm-core/src/clj/backtype/storm/daemon/worker.clj
```diff
@@ -510,15 +554,21 @@
...
+                                                   new-throttle-on (if ((:storm-conf worker) TOPOLOGY-BACKPRESSURE-ENABLE)
+                                                                     (.topology-backpressure (:storm-cluster-state worker) storm-id callback) nil)
+                                                   new-creds (.credentials (:storm-cluster-state worker) storm-id nil)]
+                                               (if ((:storm-conf worker) TOPOLOGY-BACKPRESSURE-ENABLE)
+                                                 (reset! (:throttle-on worker) new-throttle-on))
+                                               (when-not (= new-creds @credentials) ;;This does not have to be atomic, worst case we update when one is not needed
+                                                 (AuthUtils/updateSubject subject auto-creds new-creds)
+                                                 (dofor [e @executors] (.credentials-changed e new-creds))
+                                                 (reset! credentials new-creds))))]
+    (.credentials (:storm-cluster-state worker) storm-id (fn [args] (check-credentials-throttle-changed)))
```
revans2: I think I would prefer to have two functions instead of just one.  Both can be scheduled on the same timer, but keeping them separate seems cleaner.

revans2:
This looks very interesting.  I would really like to see some unit tests, especially around the disruptor queue to show that the callback is working.  Perhaps we can also handle the corner cases for callback in DisruptorQueue itself instead of having special case code in other locations as a backup.

vesense:
+1 Looks very exciting.

arunmahadevan:
It appears that the spouts across all the workers are throttled if an executor queue fills up. For instance if there are multiple spouts, even the ones not causing the executor recv queue to fill would be throttled. Is this correct and if so desirable? 

Wont it be ideal to trickle the back pressure up via the immediately preceding component(s) than slowing down the entire topology ?

revans2:
@arunmahadevan It is the intention to throttle the entire topology, all spouts.  This is what Heron does and is intended to be a last resort, which is better then nothing, but not a truly final solution. The reason for this is that from the level of a single queue it is very difficult to know what is causing that congestion. STORM-907 is intended as a follow on that will analyze the topology, determine where there are loops and provide more of true backpressure.  But in the case of a loop, and storm does support loops, you have no way to determine which upstream part is causing slowness.  And in fact it could be the bolt itself, and it needs to increase it's parallelism.

storm-core/src/clj/backtype/storm/daemon/executor.clj
```diff
@@ -811,9 +841,18 @@
...
+              event-handler (mk-task-receiver executor-data tuple-action-fn)
+              high-watermark ((:storm-conf executor-data) BACKPRESSURE-EXECUTOR-HIGH-WATERMARK)
+              low-watermark  ((:storm-conf executor-data) BACKPRESSURE-EXECUTOR-LOW-WATERMARK)
+              receive-queue-size ((:storm-conf executor-data) TOPOLOGY-EXECUTOR-RECEIVE-BUFFER-SIZE)
+              high-watermark (int (* high-watermark receive-queue-size))
+              low-watermark (int (* low-watermark receive-queue-size))]
           (disruptor/consumer-started! receive-queue)
           (fn []            
+            ;; this additional check is necessary because rec-q can be 0 while the executor backpressure flag is forever set
+            (if (and ((:storm-conf executor-data) TOPOLOGY-BACKPRESSURE-ENABLE) (< (.population receive-queue) low-watermark) @(:backpressure executor-data))
```
knusbaum: Can we move the `and` form's second condition to a new line?

storm-core/src/clj/backtype/storm/daemon/worker.clj
```diff
@@ -137,9 +159,14 @@
                         (.add remote (TaskMessage. task (.serialize serializer tuple)))
                         (log-warn "Can't transfer tuple - task value is nil. tuple type: " (pr-str (type tuple)) " and information: " (pr-str tuple)))
                      ))))
-                (local-transfer local)
-                (disruptor/publish transfer-queue remoteMap)
-              ))]
+              ;; each executor itself will do the self setting for the worker's backpressure tag
+              ;; however, when the backpressure is set, the worker still need to check whether all the executors' flag has cleared to unset worker's backpressure
+              (if (and ((:storm-conf worker) TOPOLOGY-BACKPRESSURE-ENABLE) (> (.population transfer-queue) high-watermark))
```
knusbaum: Here too.

storm-core/src/clj/backtype/storm/daemon/worker.clj
```diff
@@ -114,12 +114,34 @@
...
+  (disruptor/backpressure-handler 
+    (fn [worker]
+      (let [storm-id (:storm-id worker)
+            assignment-id (:assignment-id worker)
+            port (:port worker)
+            storm-cluster-state (:storm-cluster-state worker)]
+        (if executors 
+          (if (reduce #(or %1 %2) (map #(.get-backpressure-flag %1) executors))
+            (reset! (:backpressure worker) true)   ;; at least one executor has set backpressure
+            (reset! (:backpressure worker) false))) ;; no executor has backpressure set
```
knusbaum: Can we change these two `if`s to a single `when`?

zhuoliu:
Addressed comments:
a. Removed the config of suspend time, reuse the empty emit and put the judgement together with max.spout.pending
b. Separate functions for credentials and throttle checks in schedule-recurring thread
c. Update worker flag to ZK only when it has changed (for reducing ZK load)
d. Other comments like type hint and And.


To continue work on the other three comments:
a. Add metrics for executor's suspending times for throttle-on, inactive, spout.spending etc.
b. Concern on the possible corner cases since I put the backpressure checks in tuple-action-fn.
c. Unit tests.

storm-core/src/clj/backtype/storm/daemon/worker.clj
```diff
@@ -114,12 +114,38 @@
...
+  "make a handler that checks and updates worker's backpressure flag"
+  (disruptor/backpressure-handler 
+    (fn [worker]
+      (let [storm-id (:storm-id worker)
+            assignment-id (:assignment-id worker)
+            port (:port worker)
+            storm-cluster-state (:storm-cluster-state worker)
+            prev-backpressure-flag @(:backpressure worker)]
+        (doseq [ed executors] ;; Debug, TODO: delete
+          (log-message "zliu executor" (.get-executor-id ed) " flag is " (.get-backpressure-flag ed)))
```
redsanket: zliu in log message

zhuoliu: Thanks for the reminder, I will remove all those debug messages soon after the unit test commit.

zhuoliu:
Hi, Bobby @revans2 , all comments and concerns have been addressed. Ready for another round of review. Thanks!

zhuoliu:
Initial tests on 5 node openstack cluster (1 zk, 1 nimbus, 3 supervisors). 
1. Test a normal topology (no congestion happen)
WordCountTopology3 (3 times more workers and executors than original WordCount).

Without topology running, the zookeeper node's received packets is 5 pack/second; 
without BP, the zk load is 20.1 pack/second;
with BP, the zk load is 20.9 pack/second.
This makes sense: if a topology is running fluently (no congestion happens), the ZK will almost never be accessed by Backpressure procedures. "almost" is because we have added an additional topo-backpressure recurring checking with credential thread (just for dealing with the very rare case that ZK callback fails to proceed), which reads from ZK every 30 seconds.
So 12 workers cause 12/30 = 0.4 pack / second overheads to ZK.

This shows that Backpressure will have minimal additional overheads for any non-congested topologies.

zhuoliu:
2. Test with a congested topology. 
WordCountTopology2 (3 times more workers and executors than original WordCount; in every minute, for the wordcount bolt: in the first 30 second, there is a 50  ms sleep before each tuple emits"if (time/30000 %2 == 0) Thread.sleep(50);").

Interestingly, without ABP, this 12 worker (87 executors) topology can not run successfully in Storm since workers frequently crash because overflowing of tuples in the bolt executors.
And the ZK receive workload is 20.6 to 21 pack/second.

On the contrast, with ABP enabled, this topology runs pretty well.
And we see about 22.5 pack/sec receive workloads on Zookeeper nodes.

This test shows the great advantage of Backpressure when dealing with topologies that may have congested or slow components. Since (1) ABP makes sure this topology can run successfully; (2) ABP causes small overheads to Zookeeper.

@revans2 @d2r @knusbaum

revans2:
@zhuoliu looks like you missed adding in a file, We also need to upmerge.

storm-core/src/clj/backtype/storm/daemon/executor.clj
```diff
@@ -602,6 +640,7 @@
                     (log-message "Activating spout " component-id ":" (keys task-datas))
                     (fast-list-iter [^ISpout spout spouts] (.activate spout)))
                
+                    ;; (log-message "Spout executor " (:executor-id executor-data) " found throttle-on, now suspends sending tuples")
```
revans2: Can we remove this line? it looks like it is no longer needed.

storm-core/src/jvm/backtype/storm/Config.java
```diff
@@ -1024,6 +1024,48 @@
...
+    public static final Object BACKPRESSURE_EXECUTOR_HIGH_WATERMARK_SCHEMA =ConfigValidation.PositiveNumberValidator;
+
+    /**
+     * This signifies a state that an executor has left the congestion.
+     * If the used ratio of an execuotr's receive queue is lower than the low watermark,
+     * it may notify the worker to check whether all its executors have also left congestion,
+     * if yes, the worker's backpressure flag will be unset on the Zookeeper
+     */
+    public static final String BACKPRESSURE_EXECUTOR_LOW_WATERMARK="backpressure.executor.low.watermark";
+    public static final Object BACKPRESSURE_EXECUTOR_LOW_WATERMARK_SCHEMA =ConfigValidation.PositiveNumberValidator;
```
revans2: Looking through the code it looks like these two configs are never used.

BACKPRESSURE_EXECUTOR_HIGH_WATERMARK and BACKPRESSURE_EXECUTOR_LOW_WATERMARK.  I don't see a reason to have different configs for different queues.  Perhaps we can just remove these configs. and make the description for the other configs more generic.

storm-core/src/jvm/backtype/storm/utils/DisruptorQueue.java
```diff
@@ -138,7 +150,21 @@ private void consumeBatchToCursor(long cursor, EventHandler<Object> handler) {
         //TODO: only set this if the consumer cursor has changed?
         _consumer.set(cursor);
     }
-    
+
+    public void registerBackpressureCallback(DisruptorBackpressureCallback cb) {
+        this._cb = cb;
+    }
+
+    static public void notifyBackpressureChecker(Object trigger) {
```
revans2: I don't think this code really belongs here.  It would make more since to move it to the WorkerBackpressureThread itself.

storm-core/src/jvm/backtype/storm/utils/DisruptorQueue.java
```diff
@@ -35,6 +35,7 @@
 import java.util.HashMap;
 import java.util.Map;
 import backtype.storm.metric.api.IStatefulObject;
+import org.jgrapht.graph.DirectedSubgraph;
```
revans2: Where did this come from?

revans2:
For the most part things look really good.  I would also love to see something added to Examples.  Like a Word Count that just goes as fast as it can.  That way we can see before, with acking disabled it crashes, after with acking disabled it stays up and can run really well.  I can help with code for that if you want.

zhuoliu:
Thanks, Bobby @revans2 . I addressed all the comments. 
Actually I have such examples already written and tested. See:
 https://github.com/zhuoliu/storm/blob/888/examples/storm-starter/src/jvm/storm/starter/WordCountTopology2.java
I can check them in to master if needed.

revans2:
@zhuoliu I looked over the code, and I ran some tests with a word count program that does not sleep, and there are not enough split sentence bolts.  This kept the workers from crashing, which I verified happened without it.  It was able to keep a throughput similar to setting a decent value for max spout pending, but with a larger latency.

I am +1 for merging this in.  Great work.

rsltrifork:
Does this solve the problem of tuple timeout, when a bolt is completely stalled waiting for an external component to get back up?
I believe waiting for too long in a bolt triggers the tuple timeout, which causes the Spout to reemit, which is usually not what we want, when a bolt is waiting in a controlled manner before it can resume computation. 

Ideally, the tuple timeout should be used as a last resort to detect that internal storm components don't respond. And back pressure should ensure that the Spout doesn't reemit to temporarily busy/blocked bolts - regardless of timeout.

revans2:
@rsltrifork, No this does not solve that issue.  The timeout is still a hard coded value.  The backpressure just means that the spout will not be outputting new values.  Old tuples can still time out and be replayed.  The problem here is the halting problem.  How do you know that the bolt is waiting in a controlled manner?  Even if you do know that how do you know that if it is waiting in a controlled manner that it has not lost track of a tuple?  You have to be able to predict the future to truly solve this problem.  A hard coded timeout is a simple solution.  There have been a few other proposals to adjust the timeout dynamically, but that all have potentially serious limitations compared to a static timeout.

rsltrifork:
Thanks for the quick reply, revans2.

> How do you know that the bolt is waiting in a controlled manner?

A Bolt sending to an external system can include a circuit breaker, so it can keep waiting while the CB is tripped, while regularly retrying sends to check if the recipient is up.

While doing this, the Bolt could also tell Storm that processing of the current tuple is currently blocked, and that Storm should reset the tuple timeout for it (and subsequent inflight tuples):
            `collector.resetTupleTimeout(tuple);`

> Even if you do know that how do you know that if it is waiting in a controlled manner that it has not lost track of a tuple?

The bolt is continuously saying that it still has the tuple.

revans2:
@rsltrifork That does sound like a very interesting approach.  I would love to see some numbers on how it would perform.

acommuni:
In case external system is down, it could be interesting to pause the Spout for an amount of time. The state "OPEN" of the CB could be directly linked to the spout. I don't know if the implementation of the back pressure is manageable through public API. But it would be a nice enhancement to be able to implement Circuit Breaker algorithm with bolt and spout. 

Error count and timeout generated by request to external system can be detected by the bolt using those external component and could be propagated to the spout. In case the amount of consecutive error reaches a defined value the spout could be paused by the bolt for an amout of time (CLOSE to OPEN state). After sleeping the spout is considered in HALF OPEN state. If new error occurs spout sleeps for another amount of time else it goes to CLOSE state and continue to read new tuples.

Being able to use Circuit breaker framework like Hystrix could be a nice enhancement of the back pressure feature.

https://github.com/Netflix/Hystrix
https://github.com/Netflix/Hystrix/wiki/How-it-Works

PR #703, opened 2015-08-27
[Storm-650] Added new option to allow Kafka spout to save offset and other state using Kafka's offset management api

hsun-cnnxty:
Current Kafka spout stores the offsets (and some other states) inside ZK with its proprietary format.  This does not work well with other Kafka offset monitoring tools such as Burrow, KafkaOffsetMonitor etc. In addition, the performance does not scale well compared with offsets managed by Kafka's built-in offset management api.  I have added a new option for Kafka to store the same data using Kafka's built-in offset management capability.  The change is completely backward compatible with the current ZK storage option.  The feature can be turned on by a single configuration option.  Hope this will help people who wants to explore the option of using  Kafka's built-in offset management api.

-thanks

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,4 @@ _site
 dependency-reduced-pom.xml
 derby.log
 metastore_db
+*.versionsBackup
--- a/examples/storm-starter/pom.xml
+++ b/examples/storm-starter/pom.xml
@@ -20,7 +20,7 @@
   <parent>
       <artifactId>storm</artifactId>
       <groupId>org.apache.storm</groupId>
-      <version>0.11.0-SNAPSHOT</version>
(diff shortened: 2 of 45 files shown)


hsun-cnnxty:
Not sure whether this PR is related to STORM-590(Use Kafka APIs Instead of Internal ZK Metadata) or 
should be part of STORM-650 (Storm-Kafka Refactoring and Improvements)

HeartSaVioR:
@hsun-cnnxty 
Since STORM-650 is a kind of EPIC / Umbrella issue, and it seems not maintained now, so it would be better to file new issue.
And please do not modify version from pom.xml.

hsun-cnnxty:
Sure, will make the change and resubmit.

Thanks

PR #711, opened 2015-09-01
STORM-996: netty-unit-tests/test-batch demonstrates out-of-order delivery

knusbaum:
Netty calls notifyInterestChanged which causes a race condition in flushMessages.


d2r:
+1 This seems to fix the issue. I ran test-batch ~50 times without a failure.

kishorvpatil:
LGTM. +1

HeartSaVioR:
+1
Great! Missed StormClientHandler so there's only one thread which accesses to Client at any time.

ptgoetz:
+1 Nice work.

PR #713, opened 2015-09-02
[STORM-1005]replace default storm.local.dir value to absolute path.

Renkai:
One of my supervisors shutdown by it self after running for a few weeks.After I restart the supervisor from         another directory(actually I don't remember which directory I started the supervisor at first.Then the workers running on that machine become wild.They running as usual,but the supervisor can't detect them,and I can't shutdown them with `storm kill`.When I submit a new topology,the supervisor will try to start new a worker with the port running workers already used,it will cause failure.

It's hard form a typical user to know why such exception happens,so I think it's better to set storm.local.dir as a absolute path.

If `storm.local.dir` can be set to `$STORM_HOME/storm-local`,it may be best.But it seems yaml file only support constant values.

### DIFF
--- a/docs/documentation/Setting-up-a-Storm-cluster.md
+++ b/docs/documentation/Setting-up-a-Storm-cluster.md
@@ -52,10 +52,13 @@ storm.zookeeper.servers:
 
 If the port that your Zookeeper cluster uses is different than the default, you should set **storm.zookeeper.port** as well.
 
-2) **storm.local.dir**: The Nimbus and Supervisor daemons require a directory on the local disk to store small amounts of state (like jars, confs, and things like that). You should create that directory on each machine, give it proper permissions, and then fill in the directory location using this config. For example:
+2) **storm.local.dir**: The Nimbus and Supervisor daemons require a directory on the local disk to store small amounts of state (like jars, confs, and things like that).
+ You should create that directory on each machine, give it proper permissions, and then fill in the directory location using this config.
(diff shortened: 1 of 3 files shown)


HeartSaVioR:
One thing we should keep in mind is, some users are using Storm in Windows environment.
Using absolute path, especially known *nix path to default value will make a problem for those users.

Renkai:
If make storm.local.dir to a relative path is better,how about make the path  relative to where storm is installed rather than where user start the storm processes.

HeartSaVioR:
Sounds good for me.

docs/documentation/Setting-up-a-Storm-cluster.md
```diff
@@ -52,10 +52,13 @@ storm.zookeeper.servers:
 
 If the port that your Zookeeper cluster uses is different than the default, you should set **storm.zookeeper.port** as well.
 
-2) **storm.local.dir**: The Nimbus and Supervisor daemons require a directory on the local disk to store small amounts of state (like jars, confs, and things like that). You should create that directory on each machine, give it proper permissions, and then fill in the directory location using this config. For example:
+2) **storm.local.dir**: The Nimbus and Supervisor daemons require a directory on the local disk to store small amounts of state (like jars, confs, and things like that).
+ You should create that directory on each machine, give it proper permissions, and then fill in the directory location using this config.
+ You'd better set it to absolute path rather than relative path,otherwise it will cause unexpected behavior when you launch daemons
+ from different directories.For example:
```
erikdw: space between . and For

PR #714, opened 2015-09-02
[STORM-1024] Pass -Dstorm.log.dir arg to LogWriter

arunmahadevan:
The log4j variable `${sys:storm.log.dir}` is used in `worker.xml` to set the fileName however the system property was not being set while launching worker (via LogWriter) causing a directory by the name ${sys:storm.log.dir} to be created under `STORM_HOME` dir.

This fix is to pass the system property while launching LogWriter


harshach:
+1

HeartSaVioR:
+1

HeartSaVioR:
While merging to master, I found it breaks some tests of supervisor_test.clj. 
Since it's minor, I fixed things before merging.

PR #715, opened 2015-09-02
STORM-1026: Expand external classpaths to include any jars

shellbj:


bin/storm.py
```diff
@@ -103,6 +103,12 @@ def get_config_opts():
     sys.exit(1)
 
 def get_jars_full(adir):
+    files = []
+    if os.path.isdir(adir):
+        files = os.listdir(adir)
+    elif os.path.exists(adir):
+        files = [aidr]
+
     files = os.listdir(adir)
```
HeartSaVioR: This line should be removed to accomplish above lines.

shellbj: Good catch; updated.

HeartSaVioR:
Sorry to being late, +1

revans2:
@shellbj and @HeartSaVioR thanks for the patch and the review, sorry it took so long to get this merged in.

PR #722, opened 2015-09-06
[STORM-1005] set storm.local.dir relative values relative to storm.home rather tha…

Renkai:
Solve the same problem with 
https://github.com/apache/storm/pull/713
but another way as @HeartSaVioR advised.

### DIFF
--- a/storm-core/test/clj/backtype/storm/config_test.clj
+++ b/storm-core/test/clj/backtype/storm/config_test.clj
@@ -168,3 +168,24 @@
                       (catch Exception e e)))))
         (is (thrown-cause? java.lang.IllegalArgumentException
           (.validateField validator "test" 42)))))))
+
+(deftest test-absolute-storm-local-dir
+  (let [storm-home-key "storm.home"
+        old-storm-home (System/getProperty storm-home-key)
+        conf-relative {STORM-LOCAL-DIR "storm-local"}
+        conf-absolute {STORM-LOCAL-DIR
+                       (if on-windows?
+                         "C:\\storm-local"
+                         "/var/storm-local")}]
+    (try
+      (System/setProperty storm-home-key (if on-windows? "C:\\storm-home" "/usr/local/storm-home"))
+      (testing
+        "for relative path"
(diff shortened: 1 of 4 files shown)


HeartSaVioR:
@Renkai 
Great. Could you add simple unit test to check ```absolute-storm-local-dir``` works with relative / absolute path?

Renkai:
@HeartSaVioR I hope the unit test is good enough,but I don't have a windows environment to test the code.It seems travis-ci do not have windows environment,too.

HeartSaVioR:
+1
@Renkai I think current unit test is enough.

Renkai:
Would you like to merge these commits or they need someone else to review?

HeartSaVioR:
@Renkai 
Sorry for the delay. I intended to wait for other committers to review, but since no one is reviewing I'll take care of it.

storm-core/test/clj/backtype/storm/config_test.clj
```diff
@@ -168,3 +168,24 @@
...
+(deftest test-absolute-storm-local-dir
+  (let [storm-home-key "storm.home"
+        old-storm-home (System/getProperty storm-home-key)
+        conf-relative {STORM-LOCAL-DIR "storm-local"}
+        conf-absolute {STORM-LOCAL-DIR
+                       (if on-windows?
+                         "C:\\storm-local"
+                         "/var/storm-local")}]
+    (try
+      (System/setProperty storm-home-key (if on-windows? "C:\\storm-home" "/usr/local/storm-home"))
```
HeartSaVioR: @Renkai 
Unfortunately I found this PR affects other tests randomly - supervisor tests are failing. 
Please see https://travis-ci.org/apache/storm/jobs/79231189.
Could we just use current storm.home?

HeartSaVioR: @Renkai If test should change storm.home to work correctly, let's remove test.

Renkai:
@HeartSaVioR It's fine to leave default `storm.home` to test,I manually set it before because I'm not sure if the value is already set when the test code was running.Now I remove the code which may cause environment changing.

PR #723, opened 2015-09-06
[STORM-1032] Add generics to component configuration methods

ddebree:
The getComponentConfiguration methods currently return a simple Map object. In fact they should be returning a Map<String, Object>. 
By changing this, it will make it slightly clear what is being returned by the method and potentially pick up some issues at compile time.

### DIFF
--- a/storm-core/src/jvm/backtype/storm/drpc/LinearDRPCTopologyBuilder.java
+++ b/storm-core/src/jvm/backtype/storm/drpc/LinearDRPCTopologyBuilder.java
@@ -120,7 +120,7 @@ public class LinearDRPCTopologyBuilder {
                     new CoordinatedBolt(component.bolt, source, idSpec),
                     component.parallelism);
             
-            for(Map conf: component.componentConfs) {
+            for(Map<String, Object> conf: component.componentConfs) {
                 declarer.addConfigurations(conf);
             }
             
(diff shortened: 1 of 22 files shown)


storm-core/src/jvm/backtype/storm/drpc/LinearDRPCTopologyBuilder.java
```diff
@@ -174,13 +174,12 @@ private static String boltId(int index) {
...
         public IRichBolt bolt;
         public int parallelism;
-        public List<Map> componentConfs;
+        public List<Map<String, Object>> componentConfs = new ArrayList<>();
         public List<InputDeclaration> declarations = new ArrayList<InputDeclaration>();
         
         public Component(IRichBolt bolt, int parallelism) {
             this.bolt = bolt;
             this.parallelism = parallelism;
-            this.componentConfs = new ArrayList();
```
ddebree: I moved this to line 177 above to make it slightly cleaner

storm-core/src/jvm/storm/trident/topology/TridentTopologyBuilder.java
```diff
@@ -204,7 +204,7 @@ public StormTopology buildTopology() {
         for(String id: _bolts.keySet()) {
             Component c = _bolts.get(id);
             
-            Map<String, CoordSpec> specs = new HashMap();
+            Map<String, CoordSpec> specs = new HashMap<>();
```
ddebree: Another minor generics fix.

HeartSaVioR:
+1

PR #730, opened 2015-09-10
STORM-1039: Remove commons-codec shading, commons-codec is used by ha…

priyank5485:
…doop authentication library which is used during ui authentication in secured environment.


harshach:
+1

revans2:
Can we please hold off on this for a bit.  I have a pull request I have been working on to improve the shading/packaging.  We had messed up the clojure shading and it looks like we were not properly building dependencies.  It looks like this is similar with external.

PR #731, opened 2015-09-11
Implement SQL parser using ANTLRv4.

haohui:
This PR is supposed to go against the STORM-1040 branch which does not exist yet in this mirror yet. This PR is for review only.

This PR implements a SQL parser using ANTLR v4. The parser is based upon https://github.com/antlr/grammars-v4/blob/master/sqlite/SQLite.g4. It closely follows the SQLite grammars. It only implements the subset of SQL statements that are useful for STORM-1040.

### DIFF
--- a/external/storm-sql/README.md
+++ b/external/storm-sql/README.md
@@ -1,0 +1,22 @@
+# Storm SQL
+
+Compile SQL queries to Storm topologies.
+
+## License
+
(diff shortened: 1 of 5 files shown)


harshach:
@haohui branch available here https://github.com/apache/storm/tree/STORM-1040

PR #732, opened 2015-09-14
Update ObjectDef.java for flux

lispking:
update:let constructorArgs support map type.


revans2:
@lispking is there a JIRA for this?  The change looks good to me and I am +1 for merging it in.

@ptgoetz could you also take a look?

ptgoetz:
I'd like to see this change covered by the unit tests. It should be as simple as adding a constructor, config method, etc. that takes a map as input to one of the dummy classes, then reference it in one of the unit test YAML files.

ptgoetz:
+1 tests can be added later

PR #734, opened 2015-09-15
STORM-1045. Parse CREATE TABLE statements.

haohui:
This PR supersedes https://github.com/apache/storm/pull/731.

Based on the discussion in STORM-1040, this PR takes the approach to extend the Calcite SQL parser to parse the CREATE TABLE statement. On a very high level it does the following:

* Extract the template of Parser.jj from the calcite jar
* Instantiate the template with the code in src/codegen
* Invoke javacc to generate the parser

The approach is similar to the one in Apache Drill.

### DIFF
--- a/external/sql/README.md
+++ b/external/sql/README.md
@@ -1,0 +1,22 @@
+# Storm SQL
+
+Compile SQL queries to Storm topologies.
+
+## License
+
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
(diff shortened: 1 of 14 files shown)


harshach:
+1 . @ptgoetz @revans2 would like merge this into feature branch STORM-1040.

external/sql/storm-sql-core/pom.xml~
```diff
@@ -0,0 +1,60 @@
...
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+    <build>
+        <sourceDirectory>src/jvm</sourceDirectory>
+        <testSourceDirectory>src/test</testSourceDirectory>
+    </build>
+</project>
```
d2r: This looks like an editor swap file.  Should this be removed?

PR #621, opened 2015-07-06
STORM-926:change pom to use maven-shade-plugin:2.2 and minimize the size of shade jar and fix duplicated classes warnings of flue-examples

caofangkun:
1: fix external/flux/flux-examples duplicated classes warings 
```
[WARNING] commons-collections-3.2.1.jar, commons-beanutils-core-1.8.0.jar, commons-beanutils-1.7.0.jar define 3 overlappping classes: 
[WARNING]   - org.apache.commons.collections.ArrayStack
[WARNING]   - org.apache.commons.collections.BufferUnderflowException
[WARNING]   - org.apache.commons.collections.Buffer
[WARNING] javax.servlet-3.1.jar, servlet-api-2.5-6.1.14.jar, javax.servlet-api-3.0.1.jar, servlet-api-2.5.jar define 17 overlappping classes: 
[WARNING]   - javax.servlet.ServletInputStream
[WARNING]   - javax.servlet.Filter
[WARNING]   - javax.servlet.http.HttpSession
[WARNING]   - javax.servlet.FilterConfig
[WARNING]   - javax.servlet.FilterChain
[WARNING]   - javax.servlet.http.Cookie
[WARNING]   - javax.servlet.http.HttpServletResponse
[WARNING]   - javax.servlet.ServletConfig
[WARNING]   - javax.servlet.RequestDispatcher
[WARNING]   - javax.servlet.ServletRequest
[WARNING]   - 7 more...
[WARNING] javax.servlet-3.1.jar, javax.servlet-api-3.0.1.jar define 26 overlappping classes: 
[WARNING]   - javax.servlet.Registration
[WARNING]   - javax.servlet.DispatcherType
[WARNING]   - javax.servlet.ServletRegistration$Dynamic
[WARNING]   - javax.servlet.FilterRegistration
[WARNING]   - javax.servlet.Registration$Dynamic
[WARNING]   - javax.servlet.annotation.ServletSecurity$TransportGuarantee
[WARNING]   - javax.servlet.FilterRegistration$Dynamic
[WARNING]   - javax.servlet.descriptor.JspPropertyGroupDescriptor
[WARNING]   - javax.servlet.annotation.ServletSecurity$EmptyRoleSemantic
[WARNING]   - javax.servlet.descriptor.TaglibDescriptor
[WARNING]   - 16 more...
[WARNING] hadoop-yarn-api-2.2.0.jar, hadoop-yarn-common-2.2.0.jar define 1 overlappping classes: 
[...]

PR #736, opened 2015-09-15
STORM-1012 STORM-967 STORM-922 STORM-1042 Shaded everything that was not already shaded, and cleaned up build s…

revans2:
…o the dependency-reduced pom.xml is used correctly.

### DIFF
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -241,17 +241,16 @@ in order to get started as fast as possible. Users can still install a specific 
 
 The following commands must be run from the top-level directory.
 
-    # Build the code and run the tests (requires nodejs, python and ruby installed) 
-    # `mvn clean package` will fail because storm-core requires storm-maven-plugin.
-    # This plugin should be installed before compiling storm-core.
-    $ mvn clean install
+The first step is to build/install the plugins and storm-core
 
-    # Build the code and run the tests, with specifying default test timeout (in millisecond)
-    $ export STORM_TEST_TIMEOUT_MS=10000
-    $ mvn clean install
+`mvn clean install -Pstorm-core`
 
-    # Build the code but skip the tests
-    $ mvn clean install -DskipTests=true
+If you wish to skip the unit tests you can do this by adding `-DskipTests` to the command line. Once this
+completes successfully you may run
+
+`mvn clean install -Pstorm-more`
+
+to build and test all of the external libraries and examples.  Again if you want to include `-DskipTests` you can.
 
 In case you modified `storm.thrift`, you have to regenerate thrift code as java and python code before compiling whole project.
 
@@ -266,19 +265,7 @@ You can also run tests selectively via the Clojure REPL.  The following example 
 [auth_test.clj](storm-core/test/clj/backtype/storm/security/auth/auth_test.clj), which has the namespace
 `backtype.storm.security.auth.auth-test`.
 
-First, start the REPL from within the relevant sub-project (here: `storm-core`):
-
-    $ cd storm-core/
-    $ mvn clojure:repl
-
-Now we run the tests in `auth_test.clj` in the REPL:
-
-```clojure
-;; You can use both absolute as well as relative paths to the .clj file.
-(load-file "test/clj/backtype/storm/security/auth/auth_test.clj")
-(ns backtype.storm.security.auth.auth-test)
-(run-tests)
-```
+You can also run tests selectively with `-Dtest=<test_name>`.  This works for both clojure and junit tests.
 
 > Tip: IDEs such as IntelliJ IDEA support a built-in Clojure REPL, which you can also use to run tests selectively.
 > Sometimes you may find that tests pass/fail depending on which REPL you use, which -- although frustrating --
@@ -295,7 +282,8 @@ You can create a _distribution_ (like what you can download from Apache) as foll
 do not use the Maven release plugin because creating an official release is the task of our release manager.
 
     # First, build the code.
-    $ mvn clean install  # you may skip tests with `-DskipTests=true` to save time
+    $ mvn clean install -Pstorm-core # you may skip tests with `-DskipTests=true` to save time
+    $ mvn clean install -Pstorm-more # you may skip tests with `-DskipTests=true` to save time
 
     # Create the binary distribution.
     $ cd storm-dist/binary && mvn package
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -192,6 +192,7 @@ To pull in a merge request you should generally follow the command line instruct
 3. Merge the pull request into your local test branch.
 
         $ git pull <remote_repo_url> <remote_branch>
+    You can use `./dev-tools/storm-merge.py <pull-number>` to produce the above command most of the time.
 
 4.  Assuming that the pull request merges without any conflicts:
     Update the top-level `CHANGELOG.md`, and add in the JIRA ticket number (example: `STORM-1234`) and ticket
@@ -222,36 +223,10 @@ To pull in a merge request you should generally follow the command line instruct
 
 # Build the code and run the tests
 
-Maven has some issues with maven pugins, shading and multi-module projects which force Storm's build process to be somewhat non-standard.
-
-Essentially what happens is that if you build a multi-module project where one module depends on another module that is a part of the same build, maven tries to be smart and will use the pom.xml and .jar or .class directory from the module that is a dependency.
-
-So if we have a project set up like the following where we have flux depend on storm-core, and storm-core is shaded.  flux will see different versions of the code and pom.xml from storm-core depending on what maven command is run.
-
-- storm.git/
-  - pom.xml
-  - storm-core/
-    - pom.xml
-    - dependency-reduce-pom.xml
-    - target/
-      - classes/
-      - storm-core.jar (shaded)
-      - original.storm-core.jar (no-shaded)
-  - flux/
-    - pom.xml (depends on storm-core)
-    - target/
-      - classes/
-      - flux.uber.jar
-
-If I run `mvn clean compile` the classpath for compiling flux will include `storm.git/storm-core/target/classes` which are not shaded, and if flux is written to use storm-core shaded class it will fail to compile.  flux will also see the full set of non-shaded dependencies from storm-core, so if flux is creating an uber jar, it will assume that all of those dependencies will be provided by storm-core, when in reality they will not be.
-
-If I run `mvn clean package` or `mvn clean install` flux will see the shaded storm-core.jar on its classpath, but it will still see the full non-shaded set of storm-core's dependencies.
-
-The only way I found to make flux use the dependency-reduced-pom.xml from storm-core, is to first build and install storm-core by itself, and then build and install flux without storm-core.  That way storm-core will install the shaded jar and dependency-reduced-pom.xml into the local repository.  Then flux will get its dependencies from the local repo instead of trying to rebuild storm-core.
-
-Because of this we have split the build into two profiles storm-core and storm-more.  By default both build together which works OK for simple testing, but when doing a release it is important to build them in two phases or flux, storm-starter, and possibly others will not be packaged correctly.
-
 ## Prerequisites
+Firt of all you need to make sure you are using maven 3.2.5 or below.  There is a bug in later versions of maven as linked to from https://issues.apache.org/jira/browse/MSHADE-206 that
+cause shaded dependencies to not be packaged correctly.  Also please be aware that because we are shading dependencies mvn dependency:tree will not always show the dependencies correctly. 
+
 In order to build `storm` you need `python`, `ruby` and `nodejs`. In order to avoid an overful page we don't provide platform/OS specific installation instructions for those here. Please refer to you platform's/OS' documentation for support.
 
 The `ruby` package manager `rvm` and `nodejs` package manager `nvm` are for convenience and are used in the tests which run on [travis](https://travis-ci.org/apache/storm). They can be installed using `curl -L https://get.rvm.io | bash -s stable --autolibs=enabled && source ~/.profile` (see the [rvm installation instructions](https://github.com/rvm/rvm) for details) and `wget -qO- https://raw.githubusercontent.com/creationix/nvm/v0.26.1/install.sh | bash && source ~/.bashrc` (see the [nvm installation instructions](https://github.com/creationix/nvm) for details).
@@ -270,16 +245,9 @@ in order to get started as fast as possible. Users can still install a specific 
 
 The following commands must be run from the top-level directory.
 
-The first step is to build/install the plugins and storm-core
+`mvn clean install`
 
-`mvn clean install -Pstorm-core`
-
-If you wish to skip the unit tests you can do this by adding `-DskipTests` to the command line. Once this
-completes successfully you may run
-
-`mvn clean install -Pstorm-more`
-
-to build and test all of the external libraries and examples.  Again if you want to include `-DskipTests` you can.
+If you wish to skip the unit tests you can do this by adding `-DskipTests` to the command line. 
 
 In case you modified `storm.thrift`, you have to regenerate thrift code as java and python code before compiling whole project.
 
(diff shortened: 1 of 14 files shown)


revans2:
I am working on a version for 0.10.0 too,  This is against master

dev-tools/travis/travis-install.sh
```diff
@@ -26,12 +26,23 @@ TRAVIS_SCRIPT_DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
...
 BUILD_RET_VAL=$?
 
 if [[ "$BUILD_RET_VAL" != "0" ]];
 then
-  cat "install.txt"
+  cat "install-storm-core.txt"
+  exit ${BUILD_RET_VAL}
+fi
+
+exit ${BUILD_RET_VAL}
```
HeartSaVioR: This line should be removed to continue building storm-more.

dev-tools/travis/travis-script.sh
```diff
@@ -31,7 +31,7 @@ cd ${STORM_SRC_ROOT_DIR}
 export STORM_TEST_TIMEOUT_MS=100000
 
 # We now lean on Travis CI's implicit behavior, ```mvn clean install -DskipTests``` before running script
-mvn test -fae -Pnative
+mvn test -fae -Pnative -Pstorm-core && mvn test -fae -Pnative -Pstorm-more
```
HeartSaVioR: We could remove -Pnative when running test with -Pstorm-more.

storm-core/pom.xml
```diff
@@ -385,168 +427,199 @@
...
+                            <pattern>clj-time</pattern>
+                            <shadedPattern>org.apache.storm.shade.clj-time</shadedPattern>
+                        </relocation>
+                        <relocation>
+                            <pattern>com.fasterxml</pattern>
+                            <shadedPattern>org.apache.storm.shade.com.fasterxml</shadedPattern>
+                        </relocation>
+                        <relocation>
+                            <pattern>org.apache.thrift</pattern>
+                            <!-- This pattern is inconsistent for backwards compatability purposes. -->
```
HeartSaVioR: Trivial typo here, compatability -> compatibility

Btw, we're always having possibilities of breaking backward compatibility in point of users view when more libraries would be shaded/relocated.
I mean users should include additional dependencies to their topology and rebuild topology package.
So IMO, with explanation we can fix the thrift shaded pattern now.

storm-dist/binary/src/main/assembly/binary.xml
```diff
@@ -30,38 +30,6 @@
             <useProjectArtifact>false</useProjectArtifact>
             <outputDirectory>lib</outputDirectory>
             <unpack>false</unpack>
-            <excludes>
```
HeartSaVioR: That was one thing what I didn't understand, and it was a sign that we didn't shade libraries properly. 
(Because assembly can see those packages...)
Do I understand correctly?

HeartSaVioR:
May be a beginner questions, I have some questions regarding changeset.

1. Seems like we're excluding "clj" files from shaded clojure library dependencies which doesn't have any "class" files in jar. Could you explain how it works?
2. I don't know what splitting build steps changes build result. I'm seeing two dependency-reduced-pom.xml but can't find major differences except additional shading. Could you explain it, too?

DEVELOPER.md
```diff
@@ -274,7 +261,8 @@ You can create a _distribution_ (like what you can download from Apache) as foll
 do not use the Maven release plugin because creating an official release is the task of our release manager.
 
     # First, build the code.
-    $ mvn clean install  # you may skip tests with `-DskipTests=true` to save time
+    $ mvn clean install -Pstorm-core # you may skip tests with `-DskipTests=true` to save time
+    $ mvn clean install -Pstorm-more # you may skip tests with `-DskipTests=true` to save time
```
knusbaum: Duplicate?

storm-core/pom.xml
knusbaum: +1

DEVELOPER.md
HeartSaVioR: No, its profiles are different. It shows that build steps need to be split into two.

knusbaum: Ahh. *more* and *core*

knusbaum:
Compiles/runs fine for me. 

It would be nice to have a makefile or somesuch to implement basic targets like compile, install, and test, since maven's commands have gotten so complicated.

storm-core/pom.xml
revans2: I left it the same because it is a different level of compatibility.  This change just requires a recompile and explicitly calling out dependencies.  If I change the package that changes some APIs that we expose because we expose a TException in a few places.  That is why I didn't shade disruptor or kryo because both of them have pluggable interfaces exposed through configs, so I didn't want to touch anything that was directly exposed to the end user.  If we want to make the change we can, but I would like to see a separate JIRA for that.  This is to fix the builds so shading is working properly.

storm-dist/binary/src/main/assembly/binary.xml
revans2: Yes, the issue was partly that we didn't shade the libraries properly, but also we had excludes in there that were not needed.  The build would print out lots of warning that we were excluding things that were not a part of the dependency list.

d2r:
> It would be nice to have a makefile or somesuch to implement basic targets like compile, install, and test, since maven's commands have gotten so complicated.

Yeah, I agree with the idea that we should be able to issue one command line and have that build everything correctly.  I think layering something else like Make on top of Maven would complicate things.  Maven should be capable of building a complex setup like this correctly without needing to wrap it in some other build tool or some script. Several devs use IDEs and other tools that are Maven-aware, and it would be good to keep things working smoothly.

I think this should be something for another Jira Issue, as it could take a lot of work to figure out how to tell Maven what we want to do.

knusbaum:
Another Jira is fine. I definitely agree we should aim to have Maven behave properly. 

As you said, though, it could take a lot of work to figure out. I'm assuming many are going to write their own build scripts or makefiles around Maven anyway, so an *official* makefile or build script might be a convenient stop-gap.

revans2:
Yes, I would love to see maven do the right thing from the beginning.  But it just is not setup to be able to do this type of thing.  I don't even know if it is possible.  Essentially all we really need to do is to be able to change the storm-core pom.xml dynamically during the build to switch it over to the dependency reduced version.  But I have no idea how to make that happen.

revans2:
I just pushed an update addressing the review comments.  The big thing is an explanation in storm-core/pom.xml explaining about the filtering of the .clj files and what is happening there.

revans2:
The travis failure is unrelated looks like another test did not release the socket fast enough and the test could not bind to it.

knusbaum:
+1

d2r:
+1 Looks OK to me.
Built with an empty maven repo & ran all tests following the instructions:
`mvn clean install -Pstorm-core -DskipTests && mvn clean install -Pstorm-more -DskipTests`

ptgoetz:
@revans2 @d2r @knusbaum Can you guys summarize all the issues this addresses? It looks like it covers a lot. I noticed that at some point an issue with shading was introduced in master (manifest as the fat jar in flux-examples not including its dependencies), but this PR seems to cover a lot more.

I need a little time to review this before we merge it in. I mainly want to make sure it doesn't cause any issues with the maven release plugin (which is a painfully fickle beast ;) ).

revans2:
@ptgoetz There were a number of issues.

We were shading dependencies of dependencies without shading the direct dependency.  This resulted in shaded classes leaking into user facing APIs.

We were not using the storm-core dependency reduced pom to build other components (flux most notably).  This resulted in flux assuming that dependencies would be there when they were not.

We also were partially shading some of our clojure dependencies.  Our dependency clojure code was being compiled into .class files placed in our jar. The packages for these dependencies were being shaded, but the maven dependency was not listed, so it was not removed from the dependency reduced pom.

There are also some issues with the ClojureTransformer where it would not correctly shade clojure code when the package name is split up in the .clj file, so to work around that we removed the majority of .clj files from the uber jar.

storm-core/pom.xml
```diff
@@ -385,168 +427,205 @@
...
-                          <shadedPattern>org.apache.storm.jetty</shadedPattern>
+                          <shadedPattern>org.apache.storm.shade.org.eclipse.jetty</shadedPattern>
                         </relocation>
                     </relocations>
                     <transformers>
                         <transformer implementation="org.apache.storm.maven.shade.clojure.ClojureTransformer" />
                     </transformers>
                     <filters>
+                        <!-- Several of these filters remove the .clj files from the shaded dependencies, even though only .clj files are in these jars.
+                             The reason for this is a bit complex, but intentional.  Durring the build process all of the dependency .clj files are
```
erikdw: typo: durring

storm-core/pom.xml
```diff
@@ -385,168 +427,205 @@
...
+                             compiled down into .class files, and included in storm-core.jar.  The regular shade thransformer handles these in
```
erikdw: typo: thransformer

storm-core/pom.xml
```diff
@@ -385,168 +427,205 @@
...
                     </relocations>
                     <transformers>
                         <transformer implementation="org.apache.storm.maven.shade.clojure.ClojureTransformer" />
                     </transformers>
                     <filters>
+                        <!-- Several of these filters remove the .clj files from the shaded dependencies, even though only .clj files are in these jars.
+                             The reason for this is a bit complex, but intentional.  Durring the build process all of the dependency .clj files are
+                             compiled down into .class files, and included in storm-core.jar.  The regular shade thransformer handles these in 
+                             the majority of cases correctly.  However, the ClojureTransformer does not shade everything correctly all the
+                             time.  Instead of spending a lot of time to get the ClojureTransformer to parse clojrue correctly we opted to remove
```
erikdw: typo: clojrue

kishorvpatil:
+1

d2r:
Still +1

revans2:
@ptgoetz @HeartSaVioR are there any objections to merging this in?  How about #741 for 0.10.X.

ptgoetz:
@revans2 let me quickly check to make sure the release process works with this. Be aware that there will be some commits as a result that I will revert.

ptgoetz:
Unfortunately, I'm -1 at this point since it breaks the release process (Maven release plugin) and I could not find a workaround.

The main problem is the two step build using profiles, since the release plugin expects the project to be built all at once. I can release with the `storm-core` profile, but when I go to release with `storm-more` maven goes into an infinite loop of resolving snapshot dependencies (because the poms for `storm-core` have a new version).

If we eliminate the separate profiles so we can build with one maven command, the release plugin should be happy. But in its current form, this patch would make it impossible to release maven artifacts.

revans2:
@ptgoetz 
What exactly is the release process?  I don't see it documented anywhere?  Our internally release process is mostly.

```
mvn -Pstorm-core,storm-more versions:set -DnewVersion=${NEW_VERSION}
mvn -Pstorm-core,storm-more versions:update-child-modules
mvn -Pnative -Pstorm-core -Dworker-launcher.conf.dir=${EXEC_CONF_DIR} install
mvn -Pstorm-more install
```
we typically will package thins a little differently here because we don't use the .tgz or .zip directly but the following works fine.
```
cd storm-dist/binary
mvn clean package
```
We typically don't check in the version changes so revert them with 
```
mvn versions:revert
mvn versions:update-child-modules
```
and just tag the base.  But checking in the specific version to a new branch is helpful too.

How exactly are you pushing the finished jars/poms to the maven repository?
Is it really that hard to manually use git manually instead of relying on the release plugin?

ptgoetz:
@revans2 The release process for pushing maven artifacts is:

```
mvn release:prepare -P dist
mvn release:perform -P dist
```

That prepares and signs all the maven artifacts (including javadoc, etc.) and pushes them to repository.apache.org. It also tags the release in git and rewrites the version numbers across all poms and commits those changes.

revans2:
@ptgoetz I will play around with it and see if I can come up with a release strategy that works.

revans2:
@ptgoetz If you are OK with doing a manual process instead of using the fancy plugin I think we can make it work.
```
mvn versions:set
mvn clean install -P storm-core -DskipTests
mvn clean install -P storm-more -DskipTests
mvn clean install -P dist -DskipTests
mvn clean deploy -P dist,storm-core,storm-more
```
Then you need to do the git tagging/branching manually as well.

ptgoetz:
@revans2 That seems reasonable to me. Let me make sure that works for deploying to Nexus and get back to you. If it works then I'm fine with these changes.

DEVELOPER.md
```diff
@@ -222,6 +222,11 @@ To pull in a merge request you should generally follow the command line instruct
 
 # Build the code and run the tests
 
+Storm has some oddities with packaging and plugin dependencies.  We build our own plugins and shade storm-core,
+but have sub-modules that depend on storm-core.  Because of this the build needs to be split into two parts.
```
erikdw: Can you please explain in a bit more detail why this split is needed, and how the change accomplishes that?

revans2: Essentially what happens is that if you build a multi-module project where one module depends on another module that is a part of the same build, maven tries to be smart and will use the pom.xml and .jar or .class directory from the module that is a dependency.

So if we have a project set up like the following where we have b depends on a, and a is shaded.  b will see different versions of the code and pom.xml from a depending on what maven command is run.

- base/
  - pom.xml
  - a/
    - pom.xml
    - dependency-reduce-pom.xml
    - target/
      - classes/
      - a.jar (shaded)
      - a.orig.jar (no-shaded)
  - b/
    - pom.xml (depends on a)

If I run `mvn clean compile` the classpath for compiling b will include `base/a/target/classes` which is not shaded, and if b is written to use a shaded class it will fail to compile.  b will also see the full set of non-shaded dependencies from a, so if b is creating an uber jar, it will assume that all of those dependencies will be provided by a, when in reality they will not be.

If I run `mvn clean package` or `mvn clean install` b will see the shaded a.jar on its classpath, but it will still see the full non-shaded set of a's dependencies.

The only way I found to make b use the dependency-reduced-pom.xml from a, is to first build/install a without b, and then build/install b without a.  That way a will install the shaded jar and dependency-reduced-pom.xml into the local repository.  Then b will get its dependencies from the local repo instead of trying to rebuild a.

I'll try to add this to the documentation.

revans2:
@ptgoetz if it does work for you, I'll file a follow on JIRA to look at how we can clean this entire process up.  To me it feels like a bug in the shade plugin, but I don't know for sure.

revans2:
I filed https://issues.apache.org/jira/browse/MSHADE-206 to see if there is something we can do as part of the shade plugin to address this.

revans2:
Good news on MSHADE-206 the author thinks that is a bug in maven itself and thinks they have tracked it down to a specific commit that broke this.  I am going to test this and possibly update things.  We may have top require a very specific version of maven to get the build to work properly though.

ptgoetz:
I did some more digging and figured out that the following commit:

https://github.com/apache/storm/pull/621

is what broke the packaging for flux. Specifically, minimizing the shaded jar for flux-examples excludes classes that Maven doesn't think are used -- when in fact they are used, but used via reflection.

The warnings mentioned in the above pull request can be eliminated by alternate methods, but minimizing a flux-enabled jar will exclude classes that are in fact needed.

revans2:
Turns out that this is a bug in maven itself, along with @ptgoetz's comment.  If you want this to work correctly you need to use version maven 3.2.5 or below.  Anything above that will package dependencies incorrectly.

I will revert my changes to split the build into two, revert #621, update the documentation accordingly, and see what I can do about making sure travis gets the right version of maven.

revans2:
@ptgoetz I have the build updated.  Shading is working if you use maven 3.2.5 or below.  Please take a look again, and I will update the pull request for 0.10.x

ptgoetz:
+1

Thanks for your patience @revans2, and for bearing with me. :)

revans2:
@ptgoetz I totally get it.  the split build was a really ugly hack, I am very glad that I understand now why I was getting messed up jars.

PR #737, opened 2015-09-15
added missing installation instructions for installation of rvm and n…

krichter722:
…vm of DEVELOPER.md (without running \`mvn clean install\`) isn't possible on an average system (like Ubuntu 15.04) and subheadings to increase readability

### DIFF
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -222,6 +222,23 @@ To pull in a merge request you should generally follow the command line instruct
 
 # Build the code and run the tests
 
+## Prerequisites
+In order to build `storm` you need `python`, `ruby` and `nodejs`. In order to avoid an overful page we don't provide platform/OS specific installation instructions for those here. Please refer to you platform's/OS' documentation for support.
+
+The `ruby` package manager `rvm` and `nodejs` package manager `nvm` are for convenience and are used in the tests which run on [travis](https://travis-ci.org/apache/storm). They can be installed using `curl -L https://get.rvm.io | bash -s stable --autolibs=enabled && source ~/.profile` (see the [rvm installation instructions](https://github.com/rvm/rvm) for details) and `wget -qO- https://raw.githubusercontent.com/creationix/nvm/v0.26.1/install.sh | bash && source ~/.bashrc` (see the [nvm installation instructions](https://github.com/creationix/nvm) for details).
+
+With `rvm` and `nvm` installed you can run
+
+```sh
+rvm use 2.1.5 --install
+nvm install 0.12.2
+nvm use 0.12.2
+```
+
+in order to get started as fast as possible. Users can still install a specific version of `ruby` and/or `node` manually.
+
+## Building
+
 The following commands must be run from the top-level directory.
 
     # Build the code and run the tests (requires nodejs, python and ruby installed) 
@@ -242,6 +259,8 @@ In case you modified `storm.thrift`, you have to regenerate thrift code as java 
 cd storm-core/src
 sh genthrift.sh
 ```
+
+## Testing
 
 You can also run tests selectively via the Clojure REPL.  The following example runs the tests in
 [auth_test.clj](storm-core/test/clj/backtype/storm/security/auth/auth_test.clj), which has the namespace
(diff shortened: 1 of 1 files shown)


DEVELOPER.md
```diff
@@ -222,6 +222,19 @@ To pull in a merge request you should generally follow the command line instruct
 
 # Build the code and run the tests
 
+## Prerequisites
+You need the `ruby` package manager `rvm` which can be installed using `curl -L https://get.rvm.io | bash -s stable --autolibs=enabled && source ~/.profile` (see the [rvm installation instructions](https://github.com/rvm/rvm) for details) and the `nodejs` package manager `nvm` which can be installed using `wget -qO- https://raw.githubusercontent.com/creationix/nvm/v0.26.1/install.sh | bash && source ~/.bashrc` (see the [nvm installation instructions](https://github.com/creationix/nvm) for details).
```
HeartSaVioR: @krichter722 
It's better to clarify that Storm actually requires Python / Ruby / Node, not RVM, and NVM.
So we would better to explain below first, 
```
You need the Python, Ruby, Node to build the codes.
Python doesn't have package manager so if you don't have python, you need to install manually. (see the [python download page](https://www.python.org/downloads/) for details)
```
and continue explaining RVM and NVM with clarifying it's for convenience, not requirement. Users can still install Ruby and Node manually.

krichter722:
Thanks for your feedback. I'm just getting started with (building) `storm`.

> Python doesn't have package manager so if you don't have python, you need to install manually.

I don't understand completely. Do you mean that `python` doesn't have a built-in package manager which allows bootstrapping `python.` Is that really necessary to mention? And is it important where people get their `python` from (source installation, package managers)? -> I'd leave users with the information that they need `python` & co and add links for famous OSs if people start to ask too many questions on the mailing list.

> continue explaining RVM and NVM with clarifying it's for convenience, not requirement.

With `python`, `ruby` and `nodejs` installed I'm getting the (extremely unclear) error `Clojure failed` - I can provide more details if you want (I found https://github.com/talios/clojure-maven-plugin/issues/80 which speculates that `Clojure failed` indicates missing binaries) -, so that imo `rvm` and `nvm` are a requirement - if not other build instructions are necessary because `mvn clean install` doesn't work out-of-the-box, like one might think after reading `DEVELOPER.md`. What is your insight on this?

HeartSaVioR:
> Python doesn't have package manager so if you don't have python, you need to install manually.

I agree with you. We don't need that sentence, just providing download page link should be enough.

> With python, ruby and nodejs installed I'm getting the (extremely unclear) error Clojure failed - I can provide more details if you want (I found talios/clojure-maven-plugin#80 which speculates that Clojure failed indicates missing binaries) -, so that imo rvm and nvm are a requirement 

RVM and NVM is just a package manager for giving users convenience, not requirement. 
(I use RVM but don't use NVM on my Mac, and I don't have any issues.)
Users can still install specific version of ruby / node manually, by homebrew in OSX, etc.

Btw, you can get ```Clojure failed``` whenever any clojure tests error occur, so it doesn't tell you have an issue related to multilang.
You should open target/test-reports to see what actual tests are failed.

As you can see (from Travis CI), storm-core suffers from random failures, especially slow machines.
When you provide actual errors from test-reports I'll take a look.

krichter722:
> Btw, you can get Clojure failed whenever any clojure tests error occur, so it doesn't tell you have an issue related to multilang.
You should open target/test-reports to see what actual tests are failed.

> As you can see (from Travis CI), storm-core suffers from random failures, especially slow machines.
When you provide actual errors from test-reports I'll take a look.

Offer appreciated. I'll open a separate issue if necessary. After installating `rvm` and `nvm` everything works fine - except for the random test failures, of course.

What do you think about the new commit?

HeartSaVioR:
Great, +1.

krichter722:
> Btw, you can get Clojure failed whenever any clojure tests error occur, so it doesn't tell you have an issue related to multilang.
You should open target/test-reports to see what actual tests are failed.

I've no insight in clojure, the `maven-clojure-plugin` and its usage in `storm`, but in case that's a generic way to deal with such failures (with really incredibly unhelpful feedback) it should be added to the `Testing` section as well, right?

HeartSaVioR:
@krichter722 Great idea. Could you add it too?

krichter722:
Done.

HeartSaVioR:
+1. Thanks for contributing!

HeartSaVioR:
@krichter722 Thanks for the work, I merged into master.

PR #738, opened 2015-09-16
configurable supervisor log filename (0.9.x-branch)

erikdw:
For the storm-on-mesos project there are multiple Storm Supervisors on
every worker host. Each of these Supervisors is dedicated to a
particular topology. A challenge with this design is that the shared
supervisor log file contains logs from every single topology that
has run on a particular host. We instead want the option to have
separate supervisor log files for each topology, so that topology owners
can more easily debug their own problems.

This change allows having configurable log files for every launch of
the Supervisor, since we will be able to embed the topology ID into the
supervisor log filename.

Notably, storm-0.10 already includes a similar change for the worker
log filenames (came in with Yahoo!'s multi-tenant changes).

### DIFF
--- a/bin/storm
+++ b/bin/storm
@@ -71,6 +71,8 @@ else:
 if (not os.path.isfile(USER_CONF_DIR + "/storm.yaml")):
     USER_CONF_DIR = CLUSTER_CONF_DIR
 
+STORM_SUPERVISOR_LOG_FILE = os.getenv('STORM_SUPERVISOR_LOG_FILE', "supervisor.log")
+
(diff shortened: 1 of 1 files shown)


erikdw:
Seems the 0.9.x-branch isn't set up for travis (above failure is because we don't have a `.travis.yml`, and no `Rakefile`).
I'm assuming that this means I shouldn't be doing PRs for commits into the 0.9.x-branch.

erikdw:
Yah, this is not the proper process.  Closing it.

PR #739, opened 2015-09-16
STORM-1050:Topologies with same name run on one cluster

danny0405:
When a topology is submitted repeatedly very closely, or nimbus is too busy to response the  repeated submit RPC calls from same topology, there is very high probability that nimbus receive the repeatedly calls of the same topology, and finally several topologies with same name will be running on storm cluster.

### DIFF
--- a/storm-core/src/clj/backtype/storm/daemon/nimbus.clj
+++ b/storm-core/src/clj/backtype/storm/daemon/nimbus.clj
@@ -1181,6 +1181,7 @@
             ;; lock protects against multiple topologies being submitted at once and
             ;; cleanup thread killing topology in b/w assignment and starting the topology
             (locking (:submit-lock nimbus)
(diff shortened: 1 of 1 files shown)


HeartSaVioR:
PR's changeset is small (just one line), but since it's a bug I'd like to leave JIRA issue to let us track easier.

Could you file a JIRA issue so that we can track? 
Also please change PR's title to start with JIRA-ISSUE-NUMBER.

Please refer https://github.com/apache/storm/blob/master/DEVELOPER.md#contribute-code to how and where to make JIRA issue, and refer https://github.com/apache/storm/blob/master/DEVELOPER.md#create-a-pull-request to how to create PR's title.

Thanks in advance!

danny0405:
I have filed a JIRA issue, thanks for your nice response.

HeartSaVioR:
+1

PR #740, opened 2015-09-16
[STORM-1044] Setting dop to zero does not raise an error

mjsax:
- added IllegalArgumentException to .setBolt(...), .setSpout(...), and .setStateSpout(...) in TopologyBuilder

### DIFF
--- a/storm-core/test/clj/backtype/storm/nimbus_test.clj
+++ b/storm-core/test/clj/backtype/storm/nimbus_test.clj
@@ -846,16 +846,6 @@
                   NIMBUS-SLOTS-PER-TOPOLOGY 8}]
     (letlocals
       (bind topology (thrift/mk-topology
-                        {"1" (thrift/mk-spout-spec (TestPlannerSpout. true) :parallelism-hint 0 :conf {TOPOLOGY-TASKS 1})}
-                        {}))
-     
-      (is (thrown? InvalidTopologyException
-        (submit-local-topology (:nimbus cluster)
-                               "test"
-                               {}
-                               topology)))
-                               
-      (bind topology (thrift/mk-topology
                         {"1" (thrift/mk-spout-spec (TestPlannerSpout. true) :parallelism-hint 1 :conf {TOPOLOGY-TASKS 1})}
                         {}))
       (is (thrown? InvalidTopologyException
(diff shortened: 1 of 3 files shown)


storm-core/test/clj/backtype/storm/nimbus_test.clj
```diff
@@ -846,16 +846,6 @@
                   NIMBUS-SLOTS-PER-TOPOLOGY 8}]
     (letlocals
       (bind topology (thrift/mk-topology
-                        {"1" (thrift/mk-spout-spec (TestPlannerSpout. true) :parallelism-hint 0 :conf {TOPOLOGY-TASKS 1})}
```
HeartSaVioR: @mjsax 
Is it intended to remove this test?

HeartSaVioR:
@mjsax 
Looks good overall. 
I just want to know why one of nimbus unit test is removed, though TopologyBuilder prevents topology with zero dop.

storm-core/test/clj/backtype/storm/nimbus_test.clj
mjsax: Removed this test because it now throws different exception and this is tested in the new unit test I added for `TopologyBuilder`. Actually, I tried to change the expected exception, but this did not fix this test for me (the test fails if I change `InvalidTopologyException` to `IllegalArgumentException`). Don't know why. But I am still new to Clojure and Clojure tests.

HeartSaVioR: @mjsax 
thrift/mk-spout-spec throws IllegalArgumentException since it actually uses TopologyBuilder.
Removing this test makes sense since we cannot even try to submit topology with dop 0 now.

HeartSaVioR:
+1

storm-core/test/clj/backtype/storm/nimbus_test.clj
mjsax: I agree on this.

However, I still don't understand why changing the expected exception type does not "fix" the test... If I put `thrown? IllegalArgumentException` instead of `thrown? InvalidTopologyException` the test does not pass but fails. Can you explain?

HeartSaVioR: Because thrift/mk-topology actually throws IllegalArgumentException, not submit-local-topology.
We should wrap bind statement to check IllegalArgumentException.

mjsax: Thx. Got it now :)

PR #742, opened 2015-09-17
[STORM-1051] Fix for flushMessagse NPE

schonfeld:
STORM-763 introduced a situation in which flushMessages could receive batch = null, in which case, an NPE is thrown because we weren't validating != null


HeartSaVioR:
@schonfeld 
Could you post a new PR which points to master, too?
Thanks in advance!

schonfeld:
Closing in favor of PR #745

PR #747, opened 2015-09-19
STORM-1054. Excessive logging ShellBasedGroupsMapping if the user doesn't have any groups.

harshach:


Parth-Brahmbhatt:
+1.

revans2:
+1.  I would be nice to log the exit code in the debug message, but this is very minor.

PR #748, opened 2015-09-19
STORM-1055

randerzander:
Fixed README by adding note about com.google.common.collect class usage, also fixed hikariConfigMap reference in example.

PR #749, opened 2015-09-20
STORM-1055

randerzander:
missed a second map->hikariConfigMap.

Please also merge in 0.10.x, which has the same README issue.


HeartSaVioR:
+1
I'll merge into master, and 0.10.x with #748

PR #752, opened 2015-09-22
STORM-981 Wait for shutdown DRPC servers while testing in drpc_auth_test

HeartSaVioR:
SaslTransportPlugin.getServer() returns TThreadPoolServer, and when TThreadPoolServer.stop() is called, it just flip a stop flag.

https://github.com/apache/thrift/blob/0.9.2/lib/java/src/org/apache/thrift/server/TThreadPoolServer.java

Actual shutdown takes some time, so sometimes another unit test runs while TThreadPoolServer is being shutdown.
Fortunately, it calls setServing(false) when shutdown is completed so we can wait.

I ran whole tests from Travis CI five times and drpc_auth_test is always succeed.

### DIFF
--- a/storm-core/test/clj/backtype/storm/security/auth/drpc_auth_test.clj
+++ b/storm-core/test/clj/backtype/storm/security/auth/drpc_auth_test.clj
@@ -61,7 +61,9 @@
       ~@body
       (log-message "Stopping DRPC servers ...")
       (.stop handler-server#)
+      (wait-for-condition #(not (.isServing handler-server#)))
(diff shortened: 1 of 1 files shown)


HeartSaVioR:
adds 2 passes on my fork and PR. :)

HeartSaVioR:
I found that this PR can't resolve drpc_auth_test properly.
https://travis-ci.org/HeartSaVioR/storm/jobs/81493823

I'll see what we can do for this failure. Closing PR for now.

PR #754, opened 2015-09-22
STORM-1017: Ignore ignoreZkOffsets in case of worker crash and restart.

priyank5485:


harshach:
+1. @priyank5485 it will be great if you can add a integration test we can do this in another jira.

priyank5485:
@harshach Created JIRA item for that. @ptgoetz can you please review this as well?

PR #756, opened 2015-09-22
[STORM-1058] create CLI kill_workers to kill workers on a supervisor node

zhuoliu:
Kill all the workers running on this supervisor. 
Currently there is no direct way to kill all the running workers on a supervisor node from a command line. In addition, if the supervisor dies, the workers will not suicide automatically. 
Also, this CLI can be called by future node health-check script to cleanly shutdown a supervisor on an problematic node.

### DIFF
--- a/bin/storm.py
+++ b/bin/storm.py
@@ -352,7 +352,7 @@ def kill_workers(*args):
     """Syntax: [storm kill_workers]
 
     Kill the workers running on this supervisor. This command should be run
-    on a supervisor node. If the cluster is running in secure node, then user needs 
+    on a supervisor node. If the cluster is running in secure mode, then user needs
     to have admin rights on the node to be able to successfully kill all workers.
     """
     exec_storm_class(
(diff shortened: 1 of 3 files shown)


redsanket:
LGTM

bin/storm.py
```diff
@@ -348,6 +348,19 @@ def get_errors(*args):
         jvmtype="-client",
         extrajars=[USER_CONF_DIR, os.path.join(STORM_DIR, "bin")])
 
+def kill_workers(*args):
+    """Syntax: [storm kill_workers]
+
+    Kill the workers running on this supervisor. This command should be run
+    on a supervisor node. If the cluster is running in secure node, then user needs
```
jerrypeng: nit: "secure mode" not "node"

jerrypeng:
other than my nit comment, looks good to me

revans2:
+1 looks good.

PR #757, opened 2015-09-22
STORM-1060. Serialize Calcite plans into JSON format.

haohui:
This PR provides an implementation to serialize the results of Calcite to JSON strings. The JSON strings will be passed down to the optimizers and code generators to generate the Storm bolts or Trident functions.


harshach:
+1

PR #759, opened 2015-09-23
STORM-1062. Establish the basic structure of the code generator.

haohui:
This PR establishes the basic structure of the code generator to provide the essential build blocks for the actual implementation of the code generator.


harshach:
@haohui I am +1 on this but looks like build is broken doesn't look this patch caused it though.

PR #760, opened 2015-09-23
[STORM-1063] support relative log4j conf dir for both daemons and workers

zhuoliu:
An easier fix might be dropping the default configuration of "log4j" and force users to provide complete path, however, we still prefer to support relative path of log4j conf dir.


HeartSaVioR:
+1

jerrypeng:
LGTM +1

revans2:
+1 looks great, thanks for doing this.

PR #762, opened 2015-09-24
[STORM-1068], configure request.required.acks in KafkaUtilsTest

zhuoliu:
configure request.required.acks to be 1 in KafkaUtilsTest to make sure that the send() function gets synchronized to the broker for avoiding potential offset-out-of-range error.


d2r:
Looks OK to me. +1 pending travis CI build passes.

PR #764, opened 2015-09-25
STORM-950: Update website for publishing.

ptgoetz:
https://issues.apache.org/jira/browse/STORM-950

This pull request builds on the previous work:

* adds a people section
* cleaned up/expanded talks and slideshows page
* Added documentation section for "external" components
* removed hard-coded links to news section
* Misc. minor fixes

Once this is in, we can publish via svn one last time, then ask INFRA to turn on git publishing so the site gets updated anytime the `asf-site` branch gets updated.


harshach:
+1

Parth-Brahmbhatt:
+1.

d2r:
https://github.com/apache/storm/blob/1d09012e718091f67e3c8710fae7e6da64d86378/css/style.css#L341-L344

Sorry, but could we tone down the button hover transition to 100 or 200ms from 1s?  It's pretty intense. :)

d2r:
Some of the links referencing code in Message-passing-implementation, storm-jdbc, Transactional-topologies look like they might be broken or pointing to the wrong place.  We can correct this separately.

+1 even with these issues.  I think this looks really good, nice work.

PR #768, opened 2015-09-29
STORM-1071 : Task Message format to include source task id.

bourneagain:
As part of gathering inputs for Resource Aware Scheduler, understanding the tuple network characteristics between tasks would be helpful. 
The present implementation of TaskMessage includes only the destination task id and message. Including the source task id to it would help us understand the task-task (component/component) network bandwidth characteristics. It shall also avoid the deserialization of message to retrieve the source task id from the serialized tuple message.


bourneagain:
There seems to be a problem with workers dying if the topology is left to run for long time. I shall debug and open a new pull request after merging with master. Closing this for now.

PR #769, opened 2015-09-29
[STORM-1066] add current directory for worker process launcher and fix absolute-storm-local-dir bug

zhuoliu:
Currently, when supervisor launches a worker, the current directory of the worker process is not specified. In that case, the worker will be launched with a current directory as ".", which is a random directory. It can cause problems especially when the worker needs to dump something to its current directory.
To fix this, we should enable ProcessBuilder to specify worker's current directory as worker-root/worker-uuid.

Secondly, this pull request fixes one bug in absolute-storm-local-dir function. If STORM-LOCAL is configured as nil or not configured, "(is-absolute-path? path)" can cause NPE.

### DIFF
--- a/storm-core/src/clj/backtype/storm/config.clj
+++ b/storm-core/src/clj/backtype/storm/config.clj
@@ -134,7 +134,9 @@
 (defn absolute-storm-local-dir [conf]
   (let [storm-home (System/getProperty "storm.home")
         path (conf STORM-LOCAL-DIR)]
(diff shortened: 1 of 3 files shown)


revans2:
+1 looks good to me.

HeartSaVioR:
+1 Nice work.

ptgoetz:
+1

HeartSaVioR:
Since PR contains bugfix which is introduced to 0.10.x-branch, I'll backport this to 0.10.x.

PR #772, opened 2015-09-30
STORM-1079. Batch Puts to HBase.

harshach:


Parth-Brahmbhatt:
+1.

dossett:
I applied some of this same batching and tick tuple logic to SequenceFileBolt in STORM-1073.  I was able to abstract much of that code to the AbstractHdfsBolt so that HdfsBolt and SequenceFileBolt don't repeat much code.

I wonder if there's an even higher level of abstraction that could apply to?  All of these bolts are sinks, so something like a BatchedTickSink abstract class?  I will think about that some more...

+1 (non-binding) for this change.

HeartSaVioR:
+1

revans2:
My only issue is that this is not a backwards compatible change.  The code looks fine, but I am a little concerned that someone will upgrade and not realize that if it is a low throughput bolt tuples will start to timeout.  I am OK with merging this in as is, but I want to be sure that everyone is aware that this could cause issues.

dossett:
When I made similar changes to HdfsBolt I added a default tick interval (hat-tip to @Parth-Brahmbhatt for that suggestion) to address that:

https://github.com/apache/storm/blob/master/external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/HdfsBolt.java#L106-L112

external/storm-hbase/src/main/java/org/apache/storm/hbase/bolt/HBaseBolt.java
```diff
@@ -53,21 +61,62 @@ public HBaseBolt withConfigKey(String configKey) {
...
+        if (TupleUtils.isTick(tuple)) {
+            LOG.debug("TICK received! current batch status [" + tupleBatch.size() + "/" + batchSize + "]");
+            flush = true;
+        } else {
+            byte[] rowKey = this.mapper.rowKey(tuple);
+            ColumnList cols = this.mapper.columns(tuple);
+            List<Mutation> mutations = hBaseClient.constructMutationReq(rowKey, cols, writeToWAL? Durability.SYNC_WAL : Durability.SKIP_WAL);
+            batchMutations.addAll(mutations);
+            tupleBatch.add(tuple);
+            if (tupleBatch.size() >= batchSize)
```
ptgoetz: Minor nit, but if/for blocks should include curly braces even if they're only one line.

external/storm-hbase/src/main/java/org/apache/storm/hbase/bolt/HBaseBolt.java
```diff
@@ -53,21 +61,62 @@ public HBaseBolt withConfigKey(String configKey) {
...
+                this.hBaseClient.batchMutate(batchMutations);
+                LOG.debug("acknowledging tuples after batchMutate");
+                for(Tuple t : tupleBatch)
+                    collector.ack(t);
+            }
         } catch(Exception e){
             this.collector.reportError(e);
-            this.collector.fail(tuple);
-            return;
+            for (Tuple t : tupleBatch)
```
ptgoetz: Need curly braces.

ptgoetz:
To @revans2's point about backwards compatibility, I think we should make batching optional and default to the old behavior. That way users won't be surprised after an upgrade.

Parth-Brahmbhatt:
why would tuples start to timeout given before the tuple timeout occurs we have a tick tuple that fires and tries to flush the batch even if it has not reached the desired size? I thought the only difference the users would see now is that their latency may go up in a low throughput system unless they explicitly set the batch size = 1.

external/storm-hbase/src/main/java/org/apache/storm/hbase/bolt/HBaseBolt.java
harshach: @ptgoetz there is another community who doesn't want curly braces :). Will send a updated PR.

external/storm-hbase/src/main/java/org/apache/storm/hbase/bolt/AbstractHBaseBolt.java
```diff
@@ -72,5 +74,6 @@ public void prepare(Map map, TopologyContext topologyContext, OutputCollector co
         Map<String, Object> hbaseConfMap = new HashMap<String, Object>(conf);
         hbaseConfMap.put(Config.TOPOLOGY_AUTO_CREDENTIALS, map.get(Config.TOPOLOGY_AUTO_CREDENTIALS));
         this.hBaseClient = new HBaseClient(hbaseConfMap, hbConfig, tableName);
+
```
erikdw: remove unnecessary empty line?

external/storm-hbase/src/main/java/org/apache/storm/hbase/bolt/HBaseBolt.java
Parth-Brahmbhatt: I can vouch for existence of such a community :-).

HeartSaVioR:
I missed something before voting: seems like we should think about meaning about default values.

If user takes default configuration, tuples could be timed out on low throughput bolt because default will be changed to use batch but don't set tick tuple for flushing. 
(Please note that default value of flushIntervalSecs is 0)

It breaks backward compatibility, and since Storm should also consider low latency requirement, I think there's no reason to change to use batch by default.

HeartSaVioR:
Maybe it also needs to include default tick interval as @dossett stated.

harshach:
@ptgoetz  @HeartSaVioR @revans2 I really don't want to make batching optional. Without batching in my tests I am able to insert 2500 tuples per sec. Thats abysmal performance and I don't see a reason going forward users should be using per tuple insertion into hbase. We can call this out as part of release notes.

revans2:
I am fine with batching not being optional.  But we just have to document how this changes things and I would like to have the flush interval non-zero by default.  My concern really is just around someone with a low throughput bolt that now has to change their code to work.

If we can have a 1 second flush interval as the default I think that would be fine, along with some documentation on what it is doing.

HeartSaVioR:
Tuple timeout really does matter, so default value of flush interval should be considered.
Letting default value of flush interval to 1 sec would be fine to me.

When we agree default value doesn't nearly affect timeout, I'm OK to let batching default.

harshach:
@revans2 @HeartSaVioR I added the default flushIntervalSecs like @dossett did in storm-hdfs. Let me know if thats ok or you still want flushIntervalSecs to 1.

external/storm-hbase/src/main/java/org/apache/storm/hbase/bolt/HBaseBolt.java
```diff
@@ -53,21 +61,69 @@ public HBaseBolt withConfigKey(String configKey) {
...
+    public Map<String, Object> getComponentConfiguration() {
+        Map<String, Object> conf = super.getComponentConfiguration();
+        if (conf == null) {
+            conf = new Config();
+        }
+
+        if (conf.containsKey("topology.message.timeout.secs") && tickTupleInterval == 0) {
+            Integer topologyTimeout = Integer.parseInt(conf.get("topology.message.timeout.secs").toString());
+            flushIntervalSecs = (int)(Math.floor(topologyTimeout / 2));
+            LOG.debug("Setting flush interval to [" + flushIntervalSecs + "] based on topology.message.timeout.secs");
```
revans2: minor nit.  Can we use the "{}" for logging instead of using +.

revans2:
I have one very minor nit about the logging.  I am +1 either way though.

HeartSaVioR:
@harshach 
I'm fine to use message.timeout.sec / 2 as tick secs.

But, there's compile failure from Travis CI.

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.1:compile (default-compile) on project storm-hbase: Compilation failure
[ERROR] /home/travis/build/apache/storm/external/storm-hbase/src/main/java/org/apache/storm/hbase/bolt/HBaseBolt.java:[81,66] cannot find symbol
[ERROR] symbol:   variable tickTupleInterval
[ERROR] location: class org.apache.storm.hbase.bolt.HBaseBolt
```

Should change tickTupleInterval to flushIntervalSecs.

If "errored check" has raised, we should check build logs from Travis CI cause it could be possible to be compile errors.

revans2:
The test failure looks unrelated I am +1

HeartSaVioR:
Test failure is NPE from time simulator, which is known and #798 addressed it.

I've reviewed code again. Nice work, +1

PR #774, opened 2015-09-30
STORM-1081. Import json11 and gmock into the storm-sql project.

haohui:
This PR includes json11 and gmock into the storm-sql project. These two third-party libraries are going to be used in the later jiras.


harshach:
+1

PR #776, opened 2015-10-01
STORM-1078: Updated RateTracker to be thread safe

revans2:
, more accurate, and very fast. Removed sub-sampling of rate calculation in disruptor, because sub-sampling was more expensive than just doing the calculation.

### DIFF
--- a/storm-core/test/jvm/backtype/storm/utils/RateTrackerTest.java
+++ b/storm-core/test/jvm/backtype/storm/utils/RateTrackerTest.java
@@ -17,9 +17,9 @@
  */
 package backtype.storm.utils;
 
-import org.junit.Assert;
 import org.junit.Test;
 import junit.framework.TestCase;
+import static org.junit.Assert.*;
 
 /**
  * Unit test for RateTracker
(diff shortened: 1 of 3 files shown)


storm-core/test/jvm/backtype/storm/utils/RateTrackerTest.java
```diff
@@ -27,36 +27,63 @@
...
+        double [] expected = new double[] {10.0, 10.0, 10.0, 10.0, 10.0, 10.0, 10.0, 10.0, 10.0, 10.0};
+        for (int i = 0; i < expected.length; i++) {
+            double exp = expected[i];
+            rt.notify(10);
+            time += interval;
+            double actual = rt.reportRate(time);
+            rt.forceRotate(1, interval);
+            assertEquals("Expected rate on iteration "+i+" is wrong.", exp, actual, 0.00001);
+        }
+        expected = new double[] {11.0, 12.0, 13.0, 14.0, 15.0, 16.0, 17.0, 18.0, 19.0, 20.0};
```
HeartSaVioR: It would be better to comment that it is continuous test after previous loop, so for each loop earliest bucket (which stores 10) is eclipsed and replaced to 20.
I was misunderstood L43-51 to a new test, so I spent some times to find out why it is not 20.0.

HeartSaVioR:
+1
It's easier to understand. Great work.

storm-core/test/jvm/backtype/storm/utils/RateTrackerTest.java
revans2: Great point I will add this in.

ptgoetz:
+1

PR #778, opened 2015-10-01
STORM-1080. Compiler SQL literals to LLVM constants

haohui:
This PR establishes the basic structure of the expression compiler. It compiles SQL literals to LLVM constants. It provides the building block to compile arithmetic expressions, etc..


harshach:
+1. LGTM. I am not sure why the build is failing though.

PR #780, opened 2015-10-02
STORM-1086. Make FailedMsgRetryManager configurable when setting up KafkaSpout

kleinsch:
We'd like to be able to drop messages when they've been replayed N times. Seems easy to do if we could extend ExponentialBackoffMsgRetryManager, but there's no way to inject a custom FailedMsgRetryManager without changing the Kafka codebase. This PR allows users to specify the FailedMsgRetryManager as part of SpoutConfig.

### DIFF
--- a/external/storm-kafka/src/jvm/storm/kafka/FailedMsgRetryManagerFactory.java
+++ b/external/storm-kafka/src/jvm/storm/kafka/FailedMsgRetryManagerFactory.java
@@ -18,26 +18,7 @@
 package storm.kafka;
 
 import java.io.Serializable;
-import java.util.List;
 
-
-public class SpoutConfig extends KafkaConfig implements Serializable {
-    public List<String> zkServers = null;
-    public Integer zkPort = null;
-    public String zkRoot = null;
-    public String id = null;
-
-    // if set to true, spout will set Kafka topic as the emitted Stream ID
-    public boolean topicAsStreamId = false;
-
-    // setting for how often to save the current kafka offset to ZooKeeper
-    public long stateUpdateIntervalMs = 2000;
-
-    public FailedMsgRetryManagerFactory failedMsgRetryManagerFactory = new ExponentialBackoffMsgRetryManagerFactory();
-
-    public SpoutConfig(BrokerHosts hosts, String topic, String zkRoot, String id) {
-        super(hosts, topic);
-        this.zkRoot = zkRoot;
-        this.id = id;
-    }
+public interface FailedMsgRetryManagerFactory extends Serializable {
+    FailedMsgRetryManager construct();
 }
--- a/external/storm-kafka/src/jvm/storm/kafka/ExponentialBackoffMsgRetryManagerFactory.java
+++ b/external/storm-kafka/src/jvm/storm/kafka/ExponentialBackoffMsgRetryManagerFactory.java
@@ -1,0 +1,42 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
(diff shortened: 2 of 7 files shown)


external/storm-kafka/src/jvm/storm/kafka/FailedMsgRetryManagerConfig.java
```diff
@@ -0,0 +1,7 @@
+package storm.kafka;
+
+import java.io.Serializable;
+
+public interface FailedMsgRetryManagerConfig extends Serializable {
```
ptgoetz: Since this a factory pattern, I would rename this `FailedMsgRetryManagerFactory`, and the impl `ExponentialBackoffMsgRetryManagerFactory`.

external/storm-kafka/src/jvm/storm/kafka/SpoutConfig.java
```diff
@@ -39,6 +39,8 @@
     public double retryDelayMultiplier = 1.0;
     public long retryDelayMaxMs = 60 * 1000;
```
ptgoetz: Since the above values only apply to the ExponentialBackoff implementation, it would be nice if we could factor them out of this class. They could probably just be moved to `ExponentialBackoffMsgRetryManagerFactory`, and in that case we could probably eliminate the factory pattern altogether.

jerrypeng:
non trivial fail in the unit tests that need to be investigated

kleinsch:
I wasn't sure if I could make a non-backwards compatible change to SpoutConfig, but seems like a good idea to move the implementation-specific code out of there.

Renamed to FailedMsgRetryManagerFactory and moved params for the exponential backoff manager into the ExponentialBackoffMsgRetryManagerFactory.

This needs to stay a factory bc the SpoutConfig has to be serialized to be sent out to the nodes, then construct the FailedMsgRetryManager per-node. I didn't want to require that the FailedMsgRetryManager be serializable, so said the Factory would be serializable but the RetryManager doesn't need to be.

kleinsch:
I've run the tests locally using `mvn install` and they pass using JDK7 and JDK8. Any tips for why the Travis build would pass for Java 8 but not for Java 7?

kleinsch:
OK, all feedback addressed and checks passing. Should be ready to merge!

kleinsch:
Ping @harshach @ptgoetz @jerrypeng

external/storm-kafka/src/jvm/storm/kafka/FailedMsgRetryManagerFactory.java
```diff
@@ -0,0 +1,24 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package storm.kafka;
+
+import java.io.Serializable;
+
+public interface FailedMsgRetryManagerFactory extends Serializable {
+    FailedMsgRetryManager construct();
```
harshach: can you rename this to build instead of construct.

harshach:
@kleinsch re-thinking a bit not sure if we need this Factory construct given that this is single class will only be created once in when the spout initializes. Why not try the following
1. add configure method to FailedMsgRetryManager
public interface FailedMsgRetryManager {
     public void configure(Map<String,Object> config)  or public void configure(SpoutConfig config);
}

2. accept fully qualified class name as pluggable instance ( example look for authorizer config in storm-core)
3. by default create instance of ExponentialBackoffMsgRetryManager  . You can use https://github.com/apache/storm/blob/master/storm-core/src/jvm/backtype/storm/utils/Utils.java#L96
4. call configure method with spoutconfig so that it gets all the available options

Crim:
Any chance we can revive this guy from the dead?  Happy to help carry this one across the finish line anyway that I can :)

abhishekagarwal87:
I think its been taken care of by 
https://github.com/apache/storm/pull/1331. can you check?

kleinsch:
@abhishekagarwal87 Yes, #1331 looks good to me. If that gets merged, we can discard this PR.

PR #781, opened 2015-10-03
Add codota search for apache/storm to Readme

drorweiss:
The Storm codebase was semantically indexed in Codota - available at
https://www.codota.com/xref/#/github_apache_storm_560da9ada8cb8703008bbfdc/findUsages

I believe it will be useful for developers who want to learn and explore the code.

### DIFF
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -366,6 +366,8 @@ The Storm JIRA is available at:
 
 If you do not have a JIRA account yet, then you can create one via the link above (registration is free).
 
+<a name="code-search"></a>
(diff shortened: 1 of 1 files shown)


knusbaum:
I'm not sure how useful codota is since it doesn't look like it indexes the clojure code. I'm not against adding it, but I'll wait to see what others think.

My only real concern is that providing an incomplete guide to the code base will result in confusion for anyone that wanted to learn about it.

revans2:
I am +1 for merging this in.  With the JStorm merge I think it will become more useful for developers as it is now, it is mildly useful for users who what to see what the java code is doing.

kishorvpatil:
LGTM. +1

PR #783, opened 2015-10-06
[STORM-1090] Nimbus HA should support `storm.local.hostname`

schonfeld:
Nimbus HA's `NimbusInfo` class relies on each nimbus's hostname for network reachability. This is a show-stopper in situations that utilize Config.STORM_LOCAL_HOSTNAME / `storm.local.hostname`.


HeartSaVioR:
+1
Config.STORM_LOCAL_HOSTNAME is already used for supervisors/workers to report its hostname to nimbus, so it makes sense for nimbuses to report its hostname by STORM_LOCAL_HOSTNAME to each other.

ptgoetz:
+1 Nice catch.

PR #785, opened 2015-10-06
[STORM-1084] - Improve Storm config validation process to use java annotations instead of *_SCHEMA format

jerrypeng:
So currently we specify validators:

public static final String STORM_MESSAGING_NETTY_MIN_SLEEP_MS = "storm.messaging.netty.min_wait_ms";

public static final Object STORM_MESSAGING_NETTY_MIN_SLEEP_MS_SCHEMA = ConfigValidation.IntegerValidator;

A better way to do this is using annotations. Something like:

@isInteger
public static final String STORM_MESSAGING_NETTY_MIN_SLEEP_MS = "storm.messaging.netty.min_wait_ms";

Do this has many advantages. For one you can stack multiple annotations:

@isInteger
@NotNull
public static final String STORM_MESSAGING_NETTY_MIN_SLEEP_MS = "storm.messaging.netty.min_wait_ms";

And we can pass parameters into the annotations:

@isPositiveInteger(includeZero = true)
public static final String DRPC_REQUEST_TIMEOUT_SECS = "drpc.request.timeout.secs";

instead of having to write another validator: ConfigValidation.NonNegativeIntegerValidator for including zero

Also can pass in multiple validators to validate entries in lists or key and values in maps: 

 @isNoDuplicateInList
 @NotNull
 @isListEntryCustom(entryValidatorClasses={IntegerValidator.class,PositiveNumberValidator.class})
 public static final String SUPERVISOR_SLOTS_PORTS = "supervisor.slots.ports";

Users can pass in any number of validators to @isListEntryCustom and validation process I have implemented will validate every entry in the list with every validator.


storm-core/src/jvm/backtype/storm/validation/ConfigValidation.java
```diff
@@ -0,0 +1,359 @@
+package backtype.storm.validation;
```
revans2: Needs that Apache header.  Also not completely sure why we changed the package.  Not important though.

storm-core/src/jvm/backtype/storm/validation/ConfigValidation.java
```diff
@@ -0,0 +1,359 @@
...
+
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Created by jerrypeng on 10/1/15.
```
revans2: Please provide a description of the class instead of your name and date.

storm-core/src/jvm/backtype/storm/validation/ConfigValidation.java
jerrypeng: I created another package because I added a couple other files related to config validation

storm-core/src/jvm/backtype/storm/validation/ConfigValidation.java
jerrypeng: will add

revans2:
@jerrypeng 
I keep going back and forth on how we want the annotations to work. Right now they are more or less a one to one translation of what we had before.  Admittedly it is a lot cleaner than it was before, but I almost want to combine some of the validators together.  Like

```
@IntegerValidator(min=0, nullAllowed=true)
public static final String SOME_CONFIG="some.config";
```

instead of

```
@PositiveIntegerValidator
public static final String SOME_CONFIG="some.config";
```

I'm also not sure about having validator at the end of each annotation.  But I don't think we can have a validator named the same as a common class.  I want to think this through a bit more.

jerrypeng:
Well for some for annotations I call a common validator and just pass in the type that needs to be validated, such as @String @Boolean validators.

jerrypeng:
Writing another validator like IntegerValidator(min=0, nullAllowed=true) is a trivial task, but I think we need to first focus on setting a standard of using annotations in general.  There could be a lot of refactoring done for the validator's but I am not sure if that is in the scope of this Jira

revans2:
Could we use nested annotations for the complex types?

https://blogs.oracle.com/toddfast/entry/creating_nested_complex_java_annotations

```
@MapType(key=@StringType, value=@NumberType)
public static final ...;
```

instead of 

```
@MapOfStringToNumberValidator
```

We might be able to add in an `@Or`  annotation so we could get some of the complex validators to be built up from smaller validators.  So we don't need a once off Validator.

```
@StringOrStringListValidator
```
would become 

```
@OR(@StringType, @ListType(@StringType))
```

To me this just gives users better options so they don't have to create custom validators and annotations whenever a new data structure is needed.

revans2:
@jerrypeng I agree that we want a base set of annotations, and we can adjust what they do.  I am just concerned about making that base set as flexible and complete as possible.

jerrypeng:
@revans2 your suggestion are great, but the parameters passed into java annotations have to declared to be a specific type thus we could not have: 

@MapType(key=@StringType, value=@NumberType)
public static final ...;

but maybe something like:

@MapType(key=@TypeValidator(type = String.class), value=@TypeValidator(type = Number.class))
public static final ...;

revans2:
hmmm.  You are right, and there is no inheritance in annotations.  So we cannot get that to work.  Let me think about it a bit more.

revans2:
@jerrypeng Perhaps we can do a hybrid approach, where we have validation for complex types made up of something that is a bit ugly, but expressive.  And a set of simple annotations for common types. for example

```
@Validate(type=Map.class, nullAllowed=false,
    key=@Validate(type=String.class),
    value=@Validate(type=Number.class, min=0, max=500))
public static final String MY_MAP_CONF="my.map.conf";

@Validate(type=List.class, value=@Validate(type=String.class))
public static final String MY_LIST_OF_STRINGS="my.list.of.strings";

@IntValidator(min=0, nullAllowed=false)
public static final String MY_INT_CONF="my.int.conf";

//And for the really strange types
@Validate(type=List.class, value=@Validate(type=Map.class, validator=RegistryValidator.class))
```

jerrypeng:
@revans2 I don't think you solution will work either.  I don't think we can get recursion and java annotations to work,  since fields in annotations cannot default to a specific non-null type.  To have annotations that contain fields of it self will result in a unending recursive definition:

        ComplexValidator key() default  @ComplexValidator(type=Object.class, key=@ComplexValidator(type=Object.class, key=@ComplexValidator(...), value=ComplexValidator(ype=Object.class, key=@ComplexValidator(...));

revans2:
Why can't the default be null?

jerrypeng:
the java annotation spec was written that way

revans2:
Crap OK.  The JSR has thwarted my plan to take over the universe.  What we have now looks like it is as good as it gets.

jerrypeng:
@revans2 any other suggestions?

jerrypeng:
@knusbaum @d2r @zhuoliu @rfarivar Can review this PR? Thanks!

storm-core/test/jvm/backtype/storm/TestConfigValidate.java
```diff
@@ -0,0 +1,137 @@
...
+import java.lang.reflect.InvocationTargetException;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.LinkedList;
+import java.util.Map;
+
+/**
+ * Created by jerrypeng on 10/1/15.
+ */
```
d2r: remove?

storm-core/src/jvm/backtype/storm/validation/ConfigValidationAnnotations.java
```diff
@@ -0,0 +1,205 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ * <p>
```
d2r: There are several places where it seems an IDE has added these HTML tags to the license header.

d2r:
It seems ConfigValidation was changed on master since the merge base for this branch. Need to check.

storm-core/test/jvm/backtype/storm/TestConfigValidate.java
jerrypeng: will remove

storm-core/src/jvm/backtype/storm/validation/ConfigValidationAnnotations.java
jerrypeng: will remove

jerrypeng:
@d2r upmerged and added functionality.  Ready for review

storm-core/src/jvm/backtype/storm/validation/ConfigValidationAnnotations.java
d2r: I see some in the latest commit. It would be nice to remove them.

jerrypeng: will remove again.  Dane it intellij

storm-core/src/jvm/backtype/storm/validation/ConfigValidation.java
```diff
@@ -0,0 +1,518 @@
...
+     * Validates a positive number
+     */
+    public static class PositiveNumberValidator extends Validator{
+
+        @Override
+        public void validateField(String name, Object o) {
+            validateField(name, false, o);
+        }
+
+            public void validateField(String name, boolean includeZero, Object o) {
```
d2r: check indentation of method declaration

storm-core/src/jvm/backtype/storm/validation/ConfigValidation.java
```diff
@@ -0,0 +1,518 @@
...
+            if (o == null) {
+                return;
+            }
+            if (o instanceof Number) {
+                if(includeZero == true) {
```
d2r: Don't need `== true` here

storm-core/src/jvm/backtype/storm/validation/ConfigValidation.java
jerrypeng: will fix

storm-core/src/jvm/backtype/storm/validation/ConfigValidation.java
jerrypeng: will fix

storm-core/src/jvm/backtype/storm/validation/ConfigValidation.java
```diff
@@ -0,0 +1,518 @@
...
+
+        @Override
+        public void validateField(String name, Class type, Object o) {
+            ConfigValidationUtils.NestableFieldValidator validator = ConfigValidationUtils.listFv(type, false);
+            validator.validateField(name, o);
+        }
+    }
+
+    /**
+     * Validates entry in the a list with a list of custom validator
```
d2r: Maybe this? `Validates each entry in a list against a list of custom Validators`

jerrypeng: will fix

storm-core/src/jvm/backtype/storm/validation/ConfigValidationAnnotations.java
```diff
@@ -0,0 +1,216 @@
...
+    @Target(ElementType.FIELD)
+    public @interface isType {
+        Class validatorClass() default ConfigValidation.SimpleTypeValidator.class;
+
+        Class type();
+    }
+
+    @Retention(RetentionPolicy.RUNTIME)
+    @Target(ElementType.FIELD)
+    public @interface isStrings {
```
revans2: Can we make this areStrings or isListOfStrings.  isStrings is too close to isString for me, and it feels a bit grammatically suspect.

jerrypeng: ok how about isListOfStrings then? so to keep the "is" prefix consistent

storm-core/src/jvm/backtype/storm/validation/ConfigValidation.java
```diff
@@ -0,0 +1,518 @@
...
+        public void validateField(String name, Object o) {
+            if (o == null) {
+                return;
+            }
+            ConfigValidationUtils.NestableFieldValidator validator = ConfigValidationUtils.mapFv(ConfigValidationUtils.fv(String.class, false),
+                    ConfigValidationUtils.mapFv(ConfigValidationUtils.fv(String.class, false),
+                            ConfigValidationUtils.listFv(String.class, false), false), true);
+            validator.validateField(name, o);
+        }
+    }
```
d2r: Realizing this validator name was copied from before, it would be really nice to clean up the name here.  This really validates more than str->str->[str], but it also validates that the values of the nested map are lists of strings.  This is so specialized that maybe we want to rename the validator and its annotation to something like `ImpersonationAclValidator` and `@isImpersonationAcl`.

The reason is, should someone else suppose it is as generic as the name describes, they may find out it is being too strict and rejecting config entries that the javadoc and name do not say should be invalid.

jerrypeng: will change

revans2:
This is nothing you did, but now that the validation is much more readable I am seeing a few things that should probably be fixed, but I am happy to do them myself on a follow on JIRA

`TOPOLOGY_ISOLATED_MACHINES` needs `@isPositiveNumber`

All of the 'ZMQ_` configs should be deprecated.

`TRANSACTIONAL_ZOOKEEPER_PORT`needs `@isPositiveNumber`

It would be great if we could restrict `TOPOLOGY_LOGGING_SENSITIVITY` to one of the allowed values "S0", "S1", "S2", "S3"

`TOPOLOGY_SHELLBOLT_MAX_PENDING` needs `@isPositiveNumber`

`TOPOLOGY_TRIDENT_BATCH_EMIT_INTERVAL_MILLIS` needs `@isPositiveNumber`

`TOPOLOGY_MAX_ERROR_REPORT_PER_INTERVAL` and `TOPOLOGY_ERROR_THROTTLE_INTERVAL_SECS` both seem to need `@isPositiveNumber`

`TOPOLOGY_TRANSFER_BUFFER_SIZE` needs to be `@isPowerOf2`

`TOPOLOGY_ENVIRONMENT` should be `@isMapEntryType(keyType = String.class, valueType = String.class)`.

`TOPOLOGY_SLEEP_SPOUT_WAIT_STRATEGY_TIME_MS` needs `@isPositiveNumber(includeZero = true)`

`TOPOLOGY_MAX_SPOUT_PENDING` needs `@isPositiveNumber`

`TOPOLOGY_MAX_TASK_PARALLELISM` needs `@isPositiveNumber`

`WORKER_METRICS` and `TOPOLOGY_WORKER_METRICS` should be  `@isMapEntryType(keyType = String.class, valueType = String.class)`.

`TOPOLOGY_METRICS_CONSUMER_REGISTER` should have a custom validator (you might not have time to do it, so we might need a follow on JIRA for this). Something like

`@isListEntryCustom(entryValidatorClasses={MetricRegistryValidator.class})`

MetricRegistryValidator.class needs to check that it is a map, with a "class" key that points to a string, a "parallelism.hint" key that points to a positive non-null integer.

`TOPOLOGY_EVENTLOGGER_EXECUTORS` needs `@isPositiveNumber`

`TOPOLOGY_ACKER_EXECUTORS` needs `@isPositiveNumber`

`TOPOLOGY_TASKS` and `TOPOLOGY_WORKERS` need `@isPositiveNumber`

`TASK_CREDENTIALS_POLL_SECS` needs `@isPositiveNumber`

`TASK_REFRESH_POLL_SECS` `TASK_HEARTBEAT_FREQUENCY_SECS` `WORKER_HEARTBEAT_FREQUENCY_SECS` and `WORKER_RECEIVER_THREAD_COUNT` need `@isPositiveNumber`

`SUPERVISOR_MONITOR_FREQUENCY_SECS` and `SUPERVISOR_HEARTBEAT_FREQUENCY_SECS` need `@isPositiveNumber`

`SUPERVISOR_WORKER_SHUTDOWN_SLEEP_SECS` needs `@isPositiveNumber`

`DRPC_HTTP_FILTER_PARAMS` should be `@isMapEntryType(keyType = String.class, valueType = String.class)`.

`DRPC_INVOCATIONS_THREADS` and `DRPC_INVOCATIONS_PORT` need `@isPositiveNumber`

`DRPC_QUEUE_SIZE` `DRPC_MAX_BUFFER_SIZE` and `DRPC_WORKER_THREADS` need `@isPositiveNumber`

`DRPC_AUTHORIZER_ACL` needs to be a Map<String, Map<String, String or List<String>>>.  This too probably needs a custom validator in a follow on JIRA.

`DRPC_PORT` needs `@isPositiveNumber`

`DRPC_HTTPS_PORT` and `DRPC_HTTP_PORT` need  `@isPositiveNumber`

`UI_HTTPS_PORT` and `UI_HEADER_BUFFER_BYTES` need  `@isPositiveNumber`

`UI_FILTER_PARAMS` should be `@isMapEntryType(keyType = String.class, valueType = String.class)`.

`LOGVIEWER_HTTPS_PORT` needs `@isPositiveNumber`

`LOGVIEWER_PORT` and `UI_PORT` need `@isPositiveNumber`

`NIMBUS_CREDENTIAL_RENEW_FREQ_SECS` needs `@isPositiveNumber`

`NIMBUS_IMPERSONATION_ACL` needs to be updated, because I don't think Map of string to map.  It is more complex then that.

`NIMBUS_TASK_LAUNCH_SECS` `NIMBUS_SUPERVISOR_TIMEOUT_SECS` `NIMBUS_INBOX_JAR_EXPIRATION_SECS` `NIMBUS_CLEANUP_INBOX_FREQ_SECS` `NIMBUS_MONITOR_FREQ_SECS` and `NIMBUS_TASK_TIMEOUT_SECS` need `@isPositiveNumber`

`NIMBUS_THRIFT_MAX_BUFFER_SIZE` needs `@isPositiveNumber`

`NIMBUS_THRIFT_THREADS` and `NIMBUS_THRIFT_PORT` need `@isPositiveNumber`

I think there are more too so we probably need a follow on JIRA for this.

storm-core/src/jvm/backtype/storm/validation/ConfigValidation.java
```diff
@@ -0,0 +1,518 @@
...
+
+    /**
+     * validates a list of has no dupicates
```
d2r: "duplicates"

storm-core/src/jvm/backtype/storm/validation/ConfigValidation.java
```diff
@@ -0,0 +1,518 @@
...
+            if (field == null) {
+                return;
+            }
+            //check if iterable
+            SimpleTypeValidator isIterable = new SimpleTypeValidator();
+            isIterable.validateField(name, Iterable.class, field);
+            HashSet<Object> objectSet = new HashSet<Object>();
+            for (Object o : (Iterable) field) {
+                if (objectSet.contains(o)) {
+                    throw new IllegalArgumentException(name + " should contain no duplicate elements");
```
d2r: Let's go ahead and include the value that was duplicated too.

storm-core/src/jvm/backtype/storm/validation/ConfigValidation.java
jerrypeng: will fix

storm-core/src/jvm/backtype/storm/validation/ConfigValidation.java
jerrypeng: will add

storm-core/src/jvm/backtype/storm/validation/ConfigValidationAnnotations.java
revans2: Perhaps isStringList so it matches isStringOrStringList.

jerrypeng: sounds good, will fix

revans2:
Filed https://issues.apache.org/jira/browse/STORM-1111 as a follow on to have the correct validation on each of the configs.

storm-core/src/jvm/backtype/storm/validation/ConfigValidation.java
```diff
@@ -0,0 +1,518 @@
...
+                    Object valueValidator = vv.newInstance();
+                    if (valueValidator instanceof Validator) {
+                        ((Validator) valueValidator).validateField(name + " Map value", entry.getValue());
+                    } else {
+                        LOG.warn("validator: {} cannot be used in MapEntryCustomValidator to validate values.  Individual entry validators must a instance of Validator class", vv.getName());
+                    }
+                }
+            }
+        }
+    }
```
d2r: It might be good to add a concise javadoc for this class as we have done for others; maybe something like `validates each key and each value against the respective arrays of validators`.

jerrypeng: will add

storm-core/src/jvm/backtype/storm/validation/ConfigValidation.java
```diff
@@ -0,0 +1,518 @@
...
+
+    /**
+     * Validates a field given field name as string
+     *
+     * @param fieldName provided as a string
+     * @param conf      map of confs
+     */
+    public static void validateField(String fieldName, Map conf) throws NoSuchFieldException, IllegalAccessException, InstantiationException, NoSuchMethodException, InvocationTargetException {
+        validateField(fieldName, conf, CONFIG_CLASS);
+    }
```
d2r: Is this method needed?

jerrypeng: useful if a developer would like to validate a field in a conf map at any point.  The developer can just call:

 validateField(Config.EXAMPLE_CONF, confMap)

instead of trying to get the corresponding Field object first.  Just make the code cleaner if someone else decides to validates some configs in the future

jerrypeng:
@d2r @revans2 I just pushed some updates based on what we discussed.  Can you guys look at it again? thanks!

storm-core/src/jvm/backtype/storm/validation/ConfigValidation.java
```diff
@@ -0,0 +1,570 @@
...
+            if (o == null) {
+                return;
+            }
+            if (o instanceof String) {
+                return;
+            }
+            //check if iterable
+            try {
+                SimpleTypeValidator.validateField(name, Iterable.class, o);
+            } catch (Exception ex) {
```
revans2: Why are we ignoring the exception?

jerrypeng: that is not suppose to be in there.  Not sure how the try catch statement got in there to begin with.  Will remove

revans2:
It looks really good. I am +1 for merging this in.  I have one question about checking that they type is iterable in one place, and then ignoring the error, to have the other code blow up if it is not iterable.  It should do the right thing, just the code is confusing and/or unnecessary.

jerrypeng:
@revans2 Thanks! You are right that piece of code involving the try catch statement is useless. I have removed it

revans2:
I am +1 pending not related CI failures.

storm-core/src/jvm/backtype/storm/validation/ConfigValidation.java
d2r: Looks good to me.

storm-core/src/jvm/backtype/storm/validation/ConfigValidation.java
d2r: Looks good.

storm-core/src/jvm/backtype/storm/validation/ConfigValidation.java
d2r: OK good answer. Let's leave it.

revans2:
The test failure is an issue with maven trying to download something and getting a failure. I don't see how it is related.

d2r:
JDK8 CI test failed because connections to repo.maven.apache.org were reset, and so dependency resolution did not work.  I ran all the tests on OSX myself and they passed for me.
I am +1, and thanks @jerrypeng for your patience during the review.

jerrypeng:
@revans2 @d2r Thank you for spending the time to review my PR! Much appreciated!

PR #786, opened 2015-10-07
[STORM-1095] Tuple.getSourceGlobalStreamid() has wrong camel-case naming

mjsax:


ptgoetz:
Should we consider overloading and deprecating vs. renaming to preserve backwards compatibility for the short term?

mjsax:
It's up to you. I had this issues in mind, too... API breaking changes might be inconvenient/problematic. Just let me know your decision.

HeartSaVioR:
@ptgoetz @mjsax 
In order to keep backward compatibility, I think we need to decide when to break backward compatibility.

After Storm 0.9.3, seems like we try to respect semver. (Right?)
If I'm right we can say that we'll break backward compatibility for each major version release. 
First version would be 1.0.0.

We can mark @Deprecated to legacy with leaving vaild reason and when to remove.

mjsax:
Just updated this PR.

storm-core/src/jvm/backtype/storm/tuple/Tuple.java
```diff
@@ -37,9 +37,15 @@
     /**
      * Returns the global stream id (component + stream) of this tuple.
```
HeartSaVioR: @mjsax 
Could you append comment to explain 'broken naming convention, use getSourceGlobalStreamId() instead'?

mjsax: Done.

HeartSaVioR:
Thanks for following up, +1

mjsax:
One side question: I linked my own GitHub repo to Travis. However, the build fails very often (even if everything is ok). The Travis build of the Storm repo seems to be much more stable. 

For example this build is green, but my personal build of this branch failed: https://travis-ci.org/mjsax/storm/builds/84478972

Do you have any idea why? It is somewhat annoying that I cannot rely on my personal Travis builds.

HeartSaVioR:
@mjsax 
Storm has some kinds of random failures now.
I've been filing failures on https://issues.apache.org/jira/browse/STORM-915

mjsax:
Shall I add failing builds if I encounter them?

HeartSaVioR:
@mjsax Yeah, it would be great. :)

PR #789, opened 2015-10-08
STORM-1085. Compile comparison, arithmetic, and field reference expressions.

haohui:
This PR augments the compilers to compile the comparison, arithmetic and field references in SQL expressions.


harshach:
+1

PR #790, opened 2015-10-08
[STORM-1093] Launching Workers with resources specified in resource-aware schedulers

zhuoliu:
This feature provides a set of schemes that allow nimbus to put different types of resources to each worker slot, then push the resources within assignment to ZooKeepers; also, at the supervisor side, such resources in each worker slot should be used by supervisor for launching a worker's JVM (initially, the JVM heap size, it can also be used by CGroup or Docker for more extensive resource segregation).

This feature can be used not only by RAS scheduler (STORM-893), but also by any customized scheduler for conducting mem/cpu resource aware scheduling 
(by simply calling "workerSlot.allocateResource(double memOnHeap, double memOffHeap, double cpu)" at the end of their scheduling).


jerrypeng:
Looks good but probably should wait until the resource scheduler actually goes in before we merge this PR

zhuoliu:
Thanks Jerry, nicely STORM-893 (RAS) has just been merged.

zhuoliu:
Just manual test with RAS, it works well, and we can see each worker is being launched with different JVM heap sizes (added up by their tasks' specified memory usage).
May I have your reviews when you have time @revans2  @d2r ?

zhuoliu:
May I also have @revans2 @rfarivar @knusbaum @d2r to review this pull request if you are available?

conf/defaults.yaml
```diff
@@ -140,7 +140,8 @@ supervisor.memory.capacity.mb: 3072.0
 supervisor.cpu.capacity: 400.0
 
 ### worker.* configs are for task workers
-worker.childopts: "-Xmx768m"
+worker.heap.memory.mb: 768
+worker.childopts: "-Xmx%HEAP-MEM%m"
```
knusbaum: Why is this replacement value in defaults.yaml, but none of the others are?

zhuoliu: @revans2 and I think it would be better practice if we put the HEAP-MEM in the substitute-childopts function together with WORKER-ID, TOPOLOGY-ID, WORKER-PORT.

revans2: The reason for this is backwards compatibility + flexibility.  If an admin has set their own worker.childopts for a given supervisor to say 2GB, we want that to have higher priority over what we calculate.  Similarly if someone sets topology.worker.childopts it needs to have higher priority over the other two. But what if the topology wants to test out the resource aware scheduler on just their topology?  With this they can set topology.worker.childopts to "-Xmx%HEAP-MEM%m" and override the hard coded 2GB that was there before.

revans2:
The code looks good the tests all pass I am +1 on this.

knusbaum:
+1

PR #795, opened 2015-10-12
[STORM-1106] Netty should not limit attempts to reconnect

kishorvpatil:


revans2:
The change itself looks OK, but I would like us to remove the maxConnectionAttemtps member variable.  It is not needed anymore, and it scares me that we will regress again with it there.  Additionally we should look at updating the config.

```
    /**
     * Netty based messaging: The max # of retries that a peer will perform when a remote is not accessible
     */
    public static final String STORM_MESSAGING_NETTY_MAX_RETRIES = "storm.messaging.netty.max_retries";
    public static final Object STORM_MESSAGING_NETTY_MAX_RETRIES_SCHEMA = ConfigValidation.IntegerValidator;
```

Describes what the config was doing, but that is fundamentally flawed.  It would be nice to deprecate this and file a follow on JIRA so that we can use a better named/documented config in the backoff calculation

kishorvpatil:
- Updated code to mark the configuration as deprecated
- Remove private variable.
- Created follow-up jira https://issues.apache.org/jira/browse/STORM-1107

revans2:
+1 pending travis

kishorvpatil:
The Travis build failure on 1.8 seems unrelated.

storm-core/src/jvm/backtype/storm/messaging/netty/Client.java
```diff
@@ -134,7 +128,7 @@
         LOG.info("creating Netty Client, connecting to {}:{}, bufferSize: {}", host, port, bufferSize);
         int messageBatchSize = Utils.getInt(stormConf.get(Config.STORM_NETTY_MESSAGE_BATCH_SIZE), 262144);
 
-        maxReconnectionAttempts = Utils.getInt(stormConf.get(Config.STORM_MESSAGING_NETTY_MAX_RETRIES));
+        int maxReconnectionAttempts = Utils.getInt(stormConf.get(Config.STORM_MESSAGING_NETTY_MAX_RETRIES));
```
d2r: If this config does not make sense anymore, should we stop using it in the code?

d2r: > It would be nice to deprecate this and file a follow on JIRA so that we can use a better named/documented config in the backoff calculation

Disregard: I see from the above comment what we intend to do.

d2r:
+1 I would be good to address [STORM-1107](https://issues.apache.org/jira/browse/STORM-1107) before doing a subsequent release.

HeartSaVioR:
@kishorvpatil 
For confirming that I understood this issue well, I'd like you to elaborate this issue some more.

As far as I understand, when retry count exceeds, Connect.run() throws RuntimeException, but worker doesn't be killed since it is a TimerTask.
So it just closes the connection and wait for reassign for such worker.

If Nimbus reassigns dead worker to another after retry limit exceed, another connection is being made and it would be fine.
But some reason if problematic worker is just not able to connect (for example, STW, and so on) to another workers for longer than connection retrying but not forever, and nimbus doesn't reassign problematic worker, another workers cannot connect to problematic worker forever.

Is my assumption right? Or there's other reason?

kishorvpatil:
1. Nimbus would not reassign anything because - workers come-up and heartbeat correctly ( as it is independent thread.)
2. If workerA runs out of attempts to WorkerB ( because workerB was delayed in launching..) The workerA simply hangs..)

This patch simply helps resolve that bottleneck for trying to connect forever.

kishorvpatil:
Also, there is bigger problem with `HashedWheelTimer` has bigger problems when worker has large number of connections to make and the remote workers are not accepting connections. At least for version of netty we are using there have been reports of it cracking over high CPU usage - i.e. once I saw `HashedWheelTimer` cause worker hangups. I am in the process of creating a reproducible test and solution for that. But we don't have to hold back 0.10.x release for that.

HeartSaVioR:
@kishorvpatil 
Thanks for the detail. I didn't think about situation 2. :)
+1

HeartSaVioR:
Seems like default value of storm.messaging.netty.max_wait_ms is very small.

I didn't see the details about StormBoundedExponentialBackoffRetry, but if it is running as intended via class comment, we're sleeping between 100 ~ 1000 ms for each retry by default.
But we're pinning max retry count to 29 from ExponentialBackoffRetry.validateMaxRetries() since we use 1 << retryCount to sleep, resulting that we only sleep about max 29 secs, which is too short.

kishorvpatil:
@HeartSaVioR, your understanding about `StormBoundedExponentialBackoffRetry` is correct, the 29 seconds is decent time - in case where workerB restarts due to any reason - during regular set-up.  But initial startup ( requires supervisor to read assignments/download code/launch workers.) which could mean much more than that. 

The `StormBoundedExponentialBackoffRetry` lets you go beyond max attempts to give you 1sec sleep until you succeed. So, the workerA should keep trying unless nimbus informs workerA that workerB ( is no longer working on that port.)

HeartSaVioR:
@kishorvpatil 
Yeah, after this PR gets merged. :)
Thinking it once more, since default nimbus.task.timeout.secs is 30, retrying for each 1 secs seems not be short.
Just curious about adjusting max_wait_ms may affect HashedWheelTimer's CPU usage as you stated.

PR #796, opened 2015-10-12
STORM-1105. Compile the logical plans into LLVM functions.

haohui:
This PR compiles three logical operators, tablescan, filter and projection to LLVM functions.

### DIFF
--- a/external/sql/storm-sql-core/src/jvm/org/apache/storm/sql/compiler/CalcitePlanSerializer.java
+++ b/external/sql/storm-sql-core/src/jvm/org/apache/storm/sql/compiler/CalcitePlanSerializer.java
@@ -20,6 +20,7 @@ package org.apache.storm.sql.compiler;
 import com.fasterxml.jackson.core.JsonFactory;
 import com.fasterxml.jackson.core.JsonGenerator;
 import com.google.common.base.Joiner;
+import com.google.common.base.Preconditions;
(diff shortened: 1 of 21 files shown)


harshach:
+1

arunmahadevan:
Curious to know why LLVM with C++ was chosen ? Calcite spark adapter appears to be using Janino to compile the generated java code. https://github.com/apache/incubator-calcite/tree/master/spark

Just a thought if we could have avoided native code by doing something similar, however I am not fully aware of the use cases that we are trying to address so I may be wrong.

haohui:
The main motivation is performance -- The code that generated by Calcite has a lot of down casts which can potentially have a visible impact, as the targeted throughput is millions of messages per seconds. 
Bringing LLVM allows StormSQL to leverage the mature optimization and opens up further opportunities including automatic vectorization, etc.

Note that LLVM is used only for optimization -- the next step is to translate LLVM IR back to Java Bytecode (STORM-1097). The approach is quite similar to using Janino to generate Java bytecode. There is no native components in the worker node.

arunmahadevan:
Thanks, I get it now. I think it would be good to leave some hooks to plugin other approaches to bytecode generation in future if needed.

haohui:
Thanks for the feedback. The architecture will be flexible to allow pluggable code generation.

PR #800, opened 2015-10-14
[STORM-1112] add executor-id to executor thread name

zhuoliu:


storm-core/src/clj/backtype/storm/daemon/executor.clj
```diff
@@ -685,7 +685,7 @@
           0))
       :kill-fn (:report-error-and-die executor-data)
       :factory? true
-      :thread-name component-id)]))
+      :thread-name (str component-id "-executor" (:executor-id executor-data)))]))
```
HeartSaVioR: @zhuoliu 
nit: I'd like to add a "-" between "-executor" and executor-id, but it's related to personal style, and really minor.

HeartSaVioR:
@zhuoliu 
Great work, +1

storm-core/src/clj/backtype/storm/daemon/executor.clj
zhuoliu: Thanks @HeartSaVioR 
Currently the executor send queue thread has name like this:
 "Thread-25-disruptor-executor[25 25]-send-queue"

I take similar format to above to be consistent, now we will have executor thread name as:
"Thread-26-spout-executor[25 25]"

HeartSaVioR: @zhuoliu Ah, OK. I prefer current formatting. :)

revans2:
+1

PR #802, opened 2015-10-15
[STORM-1115] Stale leader-lock key effectively bans all nodes from becoming leaders

danielschonfeld:
From the issue:

```
I believe this curator bug is what's in play causing the above described situation.
https://issues.apache.org/jira/browse/CURATOR-202
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -180,7 +180,7 @@
         <commons-fileupload.version>1.2.1</commons-fileupload.version>
         <commons-codec.version>1.6</commons-codec.version>
         <clj-time.version>0.8.0</clj-time.version>
-        <curator.version>2.5.0</curator.version>
+        <curator.version>2.9.0</curator.version>
         <json-simple.version>1.1</json-simple.version>
         <ring.version>1.3.0</ring.version>
         <ring-json.version>0.3.1</ring-json.version>


Parth-Brahmbhatt:
lot of unnecessary log statements, can you remove them?

danielschonfeld:
any idea why the JDK7 test fails? I'm not entirely clear on why DRPC server is having problems starting up

revans2:
@danielschonfeld the test failure is because we have too many tests that don't use ephemeral ports.  In all likelihood the JDK8 test was running on the same box at the same time and got the port before this test could.  JDK8 tends to run through tests faster then JDK7 does.

danielschonfeld:
@revans2 what can I do to make this PR approval ready?

revans2:
@danielschonfeld It looks fine to me.  Personally am +1 on it, but I want to hear back from @Parth-Brahmbhatt.  He has a concern about unnecessary log statements.  I assume that these are coming directly out of curator itself, so I am not really sure how he wants to handle this.

Parth-Brahmbhatt:
@revans2 The log concerns were from the origin PR that @danielschonfeld which he has fixed but I guess he force pushed the branch. I am +1 on this change too. 

@danielschonfeld On a side note, can you provide any steps to reproduce this locally?

danielschonfeld:
@Parth-Brahmbhatt that's a tricky one.  I haven't found a way to reproduce but leaving nimbus work for a day or so with number of nimbuses > 1 and a good load on the system we see the number of ZK nodes/keys go up to (X*nimbuses)+1 under /leader-lock.  When that happens, we have problems trying to do anything as no nimbus thinks it's the leader which is exactly what's described in CURATOR-202.

If you can think of a way to disconnect the ZK connection but reconnect using the same session programmatically you'll have a reproduction of this bug as this always starts showing up after something like the following log lines:

```
2015-10-16 18:16:13 o.a.s.s.o.a.c.f.s.ConnectionStateManager [INFO] State change: RECONNECTED
2015-10-16 18:16:14 o.a.s.s.o.a.z.ClientCnxn [INFO] Client session timed out, have not heard from server in 6668ms for sessionid 0x1506caf14ab005f, closing socket connection and attempting reconnect
2015-10-16 18:16:14 o.a.s.s.o.a.z.ClientCnxn [INFO] Client session timed out, have not heard from server in 6672ms for sessionid 0x1506caf14ab0060, closing socket connection and attempting reconnect
2015-10-16 18:16:15 o.a.s.s.o.a.z.ClientCnxn [INFO] Opening socket connection to server 10.101.1.2/10.101.1.2:2181. Will not attempt to authenticate using SASL (unknown error)
2015-10-16 18:16:15 o.a.s.s.o.a.z.ClientCnxn [INFO] Socket connection established to 10.101.1.2/10.101.1.2:2181, initiating session
2015-10-16 18:16:15 o.a.s.s.o.a.z.ClientCnxn [INFO] Session establishment complete on server 10.101.1.2/10.101.1.2:2181, sessionid = 0x1506caf14ab005f, negotiated timeout = 20000
2015-10-16 18:16:15 o.a.s.s.o.a.c.f.s.ConnectionStateManager [INFO] State change: RECONNECTED
```

PR #803, opened 2015-10-16
STORM-1071 Task Message format to include source task id

bourneagain:
As part of gathering inputs for Resource Aware Scheduler, understanding the tuple network characteristics between tasks would be helpful. 
The present implementation of TaskMessage includes only the destination task id and message. Including the source task id to it would help us understand the task-task (component/component) network bandwidth characteristics. It shall also avoid the deserialization of message to retrieve the source task id from the serialized tuple message.


bourneagain:
I see I should be updating netty_unit_test.clj to prevent the Travis CI build failures. I shall push an update soon.

bourneagain:
I see with present changes, when I performed the unit test, netty ```test-basic``` and ```test-large-msg``` work fine but ```test-batch``` fails if we go over 1000 messages. I could also see ```org.apache.maven.plugin.MojoExecutionException: Clojure failed.```

PR #804, opened 2015-10-16
[STORM-1117] Adds visualization-init route previously missing

d2r:
Also corrects arity of one function, removing an unused parameter.


zhuoliu:
Looks good. +1

revans2:
+1

HeartSaVioR:
+1 Before applying, 'Show Visualization' button doesn't work. After applying it does work.

PR #805, opened 2015-10-17
STORM-1118: Added test to compare latency vs. throughput in storm.

revans2:
Also added a metrics gathering server for testing, an HdrHistogram based metric, and CPU utilization metric using sigar.

This should provide a simple test for comparing different versions of storm to one another or different hardware options.  The goal is to produce a load at a given throughput and measure the latency it takes the system to process that throughput.  Generally if you have a given SLA you want the same hardware to handle the most throughput possible to meet your SLA.

The current test uses word count, but should easily be extended to handle more complex processing.


kishorvpatil:
LGTM. +1. `storm-metrics` would be useful sub-module to keep up our efforts with improving storm metrics.

jerrypeng:
+1 this is going to be very useful when attempting to profile storm performance and underlying hardware performance

PR #810, opened 2015-10-21
[STORM-1121] Remove method call to avoid overhead during topology submission time

kishorvpatil:
Nimbus calls mk-assignments from SubmitTopology within lock is causing it to wait for processing of all heartbeats. This conflicts with recurring mk-assignments call. Topology can be submitted without making assignments, since for new topology the assignments would be made available as part of next scheduling cycle.). This gives more consistent response time as avoid locking within nimbus.

### DIFF
--- a/storm-core/test/clj/backtype/storm/integration_test.clj
+++ b/storm-core/test/clj/backtype/storm/integration_test.clj
@@ -237,9 +237,7 @@
                              "acking-test1"
                              {}
                              (:topology tracked))
-      ;; Instead of sleeping until topology is scheduled, rebalance topology so mk-assignments is called.
-      (.rebalance (:nimbus cluster)
-                             "acking-test1" (doto (RebalanceOptions.) (.set_wait_secs 0)))
+      (advance-cluster-time cluster 11)
       (.feed feeder1 [1])
       (tracked-wait tracked 1)
       (checker1 0)
@@ -282,9 +280,7 @@
                              "test-acking2"
                              {}
                              (:topology tracked))
-      ;; Instead of sleeping until topology is scheduled, rebalance topology so mk-assignments is called.
-      (.rebalance (:nimbus cluster)
-                             "test-acking2" (doto (RebalanceOptions.) (.set_wait_secs 0)))
+      (advance-cluster-time cluster 11)
       (.feed feeder [1])
       (tracked-wait tracked 1)
       (checker 0)
@@ -355,9 +351,7 @@
                              "test"
                              {}
                              (:topology tracked))
-      ;; Instead of sleeping until topology is scheduled, rebalance topology so mk-assignments is called.
-      (.rebalance (:nimbus cluster)
-                             "test" (doto (RebalanceOptions.) (.set_wait_secs 0)))
+      (advance-cluster-time cluster 11)
       (.feed feeder [1])
       (tracked-wait tracked 1)
       (checker 1)
@@ -579,9 +573,7 @@
                                               TOPOLOGY-DEBUG true
                                               }
                                              (:topology tracked))
-            ;; Instead of sleeping until topology is scheduled, rebalance topology so mk-assignments is called.
-            _ (.rebalance (:nimbus cluster)
-                                             "test-errors" (doto (RebalanceOptions.) (.set_wait_secs 0)))
+            _ (advance-cluster-time cluster 11)
             storm-id (get-storm-id state "test-errors")
             errors-count (fn [] (count (.errors state storm-id "2")))]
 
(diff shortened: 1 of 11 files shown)


storm-core/test/clj/backtype/storm/integration_test.clj
```diff
@@ -236,6 +237,7 @@
                              "acking-test1"
                              {}
                              (:topology tracked))
+      (Thread/sleep 11000)
```
revans2: Is there a way we can force mk-assignments to be called?  I don't really like adding around 1 min to the time it takes to run the tests needlessly.

revans2:
I love the concept, but I would like to see the tests updated so we are not adding several mins to the time it takes to run the unit tests.

HeartSaVioR:
@kishorvpatil 
You may want to check Travis build failure, seems like there's a missing spot.

> Exception in thread "main" java.lang.IllegalArgumentException: Unable to resolve classname: RebalanceOptions, compiling:(backtype/storm/transactional_test.clj:417:26)

HeartSaVioR:
I also love the concept.
Btw, in order to rely on recurring mk-assignments, 

- We should remove ```nimbus.reassign``` since it should be true for new topologies to be assigned. We can't change ```when statement``` cause it makes nimbus.reassign ineffective.
- It would be better to let users know that new topology will be assigned by nimbus not at the moment but within ```nimbus.monitor.freq.secs```.

```
    (schedule-recurring (:timer nimbus)
                        0
                        (conf NIMBUS-MONITOR-FREQ-SECS)
                        (fn []
                          (when (conf NIMBUS-REASSIGN)
                            (locking (:submit-lock nimbus)
                              (mk-assignments nimbus)))
                          (do-cleanup nimbus)
                          ))
```

kishorvpatil:
@HeartSaVioR  looking into compilation issue. Somehow it did not get committed. Re-running the build.

HeartSaVioR:
@kishorvpatil 
rebalance seems not work cause it requires topology to be alive at the moment.

kishorvpatil:
@HeartSaVioR Thank you for review and suggestions. I have made changes so that our tests are using simulated-time cluster to avoid `Thread/sleep`. Please revisit the code and let me know if you have any further suggestions/questions.

revans2:
The changes look good I am +1, and time travis took is about the same as before.

HeartSaVioR:
@kishorvpatil 
Test changes look good.

@kishorvpatil @revans2 
I'd like to see my previous comment (https://github.com/apache/storm/pull/810#issuecomment-150047429) makes sense, and if it does, I think we should address these things.

HeartSaVioR:
@kishorvpatil 
Seems like some places in the tests still rely on rebalance. Could we change these to use simulated-time cluster as well when possible?

kishorvpatil:
@HeartSaVioR  I tried that for supervisor_test.clj, somehow mocking of submit topologies etc.methods don't make it easy. It seems straight forward to rebalance and proceed.( usually because moving time forward will also require you to send all necessary heartbeats etc.)

HeartSaVioR:
@kishorvpatil 
Good, I didn't think about mocking. Your additional changes look great.

Just I wish to discuss how we can do with ```nimbus.reassign```. 
If my understand is right, what if user turns off ```nimbus.reassign``` so that recurring is not in place? 
And maybe minor but what if user set ```nimbus.monitor.freq.secs``` a bit higher so that recurring interval is also longer?

kishorvpatil:
@HeartSaVioR Removed rebalance calls at places where I could use simulated time cluster.

kishorvpatil:
@HeartSaVioR I don't think we can remove `NIMBUS-REASSIGN` as it used mostly during supervisor_test to avoid rescheduling with value `false`. The documentation clearly advises users to not set this value to `false`.

HeartSaVioR:
@kishorvpatil 
With this change users have no option to turn off reassign since it makes cluster unusable (cluster cannot run topology unless user calls other operation which triggers mk-assignments) , so I think it should be not selectable to users.

If we need ```nimbus.reassign``` just for testing, we should deprecate this option to 0.10.x and hide this option.

kishorvpatil:
@HeartSaVioR I complete agree with decommissioning this configuration used only for testing. It should be unavailable for users. Let me deprecate this variable and add documentation. 
Are you suggesting we  push this change to 0.10.x as well?

HeartSaVioR:
@kishorvpatil 
No, just marking ```nimbus.reassign``` option as deprecate and leave proper reason from 0.10.x would be fine.

kishorvpatil:
@HeartSaVioR 
  - Modified to make "nimbus.reassign" unavailable for users. 
  - Created #815 to decommission variable for 0.10.x branch.

HeartSaVioR:
@kishorvpatil Amazing! +1. Test failure is unrelated. (Kafka)

PR #813, opened 2015-10-22
[STORM-1125] Adding separate ZK client for read in Nimbus ZK State

kishorvpatil:
Separate reader and writer for nimbus to avoid longer waits for reads to finish.

### DIFF
--- a/storm-core/src/clj/backtype/storm/cluster.clj
+++ b/storm-core/src/clj/backtype/storm/cluster.clj
@@ -54,6 +54,10 @@
     (when (Utils/isZkAuthenticationConfiguredTopology topo-conf)
       [(first ZooDefs$Ids/CREATOR_ALL_ACL)
        (ACL. ZooDefs$Perms/READ (Id. "digest" (DigestAuthenticationProvider/generateDigest payload)))])))
+ 
+(defn is-nimbus?
+  []
+  (= (System/getProperty "daemon.name") "nimbus"))
 
 (defnk mk-distributed-cluster-state
   [conf :auth-conf nil :acls nil]
@@ -62,7 +66,7 @@
     (.close zk))
   (let [callbacks (atom {})
         active (atom true)
-        zk (zk/mk-client conf
+        zk-writer (zk/mk-client conf
                          (conf STORM-ZOOKEEPER-SERVERS)
                          (conf STORM-ZOOKEEPER-PORT)
                          :auth-conf auth-conf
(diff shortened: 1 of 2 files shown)


storm-core/src/clj/backtype/storm/cluster.clj
```diff
@@ -89,68 +107,69 @@
...
        [this path watch?]
-       (zk/exists-node? zk path watch?))
+       (zk/exists-node? zk-reader path watch?))
 
      (close
        [this]
        (reset! active false)
-       (.close zk))
+       (.close zk-writer)
+       (if (is-nimbus?) (.close zk-writer)))
```
zhuoliu: I think here it should be:  (if (is-nimbus?) (.close zk-reader)))

kishorvpatil: good catch!

kishorvpatil: fixed

storm-core/src/clj/backtype/storm/cluster.clj
```diff
@@ -54,6 +54,10 @@
     (when (Utils/isZkAuthenticationConfiguredTopology topo-conf)
       [(first ZooDefs$Ids/CREATOR_ALL_ACL)
        (ACL. ZooDefs$Perms/READ (Id. "digest" (DigestAuthenticationProvider/generateDigest payload)))])))
+ 
+(defn is-nimbus?
```
HeartSaVioR: How about removing this and let the caller of mk-distributed-cluster-state decides whether writer and reader should be separated? (I mean additional argument or multimethod.)

I think it can achieve loose coupling since ClusterState doesn't need to know current daemon is nimbus or not. It just wants to know whether writer and reader should be separated or not.

storm-core/src/clj/backtype/storm/cluster.clj
```diff
@@ -89,68 +107,69 @@
...
        [this path watch?]
-       (zk/exists-node? zk path watch?))
+       (zk/exists-node? zk-reader path watch?))
 
      (close
        [this]
        (reset! active false)
-       (.close zk))
+       (.close zk-writer)
+       (if (is-nimbus?) (.close zk-reader)))
```
HeartSaVioR: nit: If we can replace ```is-nimbus?``` to check object identity between zk-writer and zk-reader, it would be more readable.

HeartSaVioR:
Looks good overall. 
I left some comments to make code cleaner. We can discuss / address these, and after addressing I'll vote.

kishorvpatil:
@HeartSaVioR I have updated the the code to use flag instead of is-nimbus? method. Thank you for suggesting this loose-coupling. Please revisit the changes and comment.

HeartSaVioR:
Nice, +1.

zhuoliu:
LGTM. +1

PR #816, opened 2015-10-24
STORM-1128: Make metrics fast

revans2:
The metrics in my profiling are contributing significantly to the CPU utilization of storm, which slows it down.  With this patch I was able to go from 6,500 sentences/second fully counted to 11,000 on a Mac Book Pro running two workers in the Throughput vs. Latency test.  If I turn off sub-sampling of metrics the performance goes back to about where it was before.  There is still more optimizations that could be done with the metrics, like having the latency metric also report counts, so we don't have to do an extra lookup and inc operation on the critical path, but for now this is good enough, and allows me to start looking at the next bottleneck.

These changes also compliment the batching changes very well.  The batching went from 6,500 sentences/second to 15,000 by itself.  This patch increases it to 25,000 sentences/second.  That is almost a 4x improvement by using both together.

### DIFF
--- a/storm-core/src/jvm/backtype/storm/metric/internal/CountStatAndMetric.java
+++ b/storm-core/src/jvm/backtype/storm/metric/internal/CountStatAndMetric.java
@@ -1,0 +1,211 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ * <p/>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p/>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package backtype.storm.metric.internal;
+
+import java.util.Map;
+import java.util.HashMap;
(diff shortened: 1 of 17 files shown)


revans2:
Just added in some fixes to remove reflection from the critical path.  The patch now is able to support 12,000 sentences/second.  it seems to be just shy of 13,000.  it can do it for a while once the JIT has kicked in, but falls behind very slowly.

harshach:
@revans2 ran some tests with regular topologies. Patch looks good to me. For measurements are you running the ThroughputVsLatency? if so I am getting  15,000 acked/sec.
+1

kishorvpatil:
LGTM. +1

storm-core/src/jvm/backtype/storm/metric/internal/CountStatAndMetric.java
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    private synchronized void rotate(long value, long timeSpent, long targetSize, long [] times, long [] buckets) {
+        times[0] += timeSpent;
+        buckets[0] += value;
+
+        long currentTime = 0;
+        long currentVal = 0;
+        if (times[0] >= targetSize) {
+            for (int i = 0; i < buckets.length; i++) {
```
abhishekagarwal87: @revans2  Instead of making a pass over the array, can we instead keep an index for each array and rotate the index instead?

revans2: Yes we could, but I don't think it is going to make that much of a difference.  I purposely moved rotating the metrics off of the critical path.  The call to rotate happens in a background thread, that is shared by all similar metrics in a worker process.  This allows the critical path, the thing that happens all of the time, to be just an atomic increment.  Rotating only happens every 30 seconds or so too.  So I suspect the best that it is going to do is reduce the overall CPU the worker uses by a tiny amount.

abhishekagarwal87: Right. I missed that it is not in critical path.

PR #822, opened 2015-10-27
[STORM-901] Worker Artifacts Directory for Hierarchical Log Management and Debugging

zhuoliu:
Originally all logs (both for workers and daemons) are under storm-home/logs, this causes name confusion ("wc22-1-1445957152-worker-6700.log") and difficulty for debugging (hard to store and manage heapdump and gc log files for all topologies' workers in a single directory). 
For better management, we present the Worker Artifacts Directory for Hierarchical Log Management and Debugging: 

1. workers' logs and metadata are moved to the new directory "storm-local/workers-artifacts", while daemon logs are still placed in storm-home/logs. Note that each topology has a separate directory, holding one sub-directory for each worker within this topology.

2. Cleanup, view and list implementations are modified accordingly in logviewer and UI. Specifically, we add per-worker quota and total quota for files' size, and the cleanup thread will periodically cleanup the oldest files in each worker directory and the whole artifacts if either quota is exceeded. Note that, the active workers' active logs will be skipped during cleaning.

3. In addition, a symlink called artifacts is created from each worker's directory to its log directory "workers-artifacts/topoId/port/". gc.log and heapdump files can be configured by users to place under this symlink, thus able to be downloaded from UI conveniently. Other useful features can be easily supported by worker artifacts directory, where their resource files can be dumped.

4. Add new web routes in logviewer for people to view, list and download all daemons' logs: e.g., nimbus.log, supervisor.log, web-access logs, metrics logs.

5. gz file is supported for review in UI.

6. A new web route "listLogs" is added to list all logs in a supervisor node (for all topologies, a certain port, or a certain topology).

7. Unit tests are added and the event log are supported under each worker's artifacts directory.

Current log structures under storm-local/workers-artifacts:
--topoId0
--topoId1
------port1
------port2
        ---------worker.log
        ---------worker.log.err 
        ---------worker.log.out
        ---------worker.yaml
        ---------gc.log
        ----------heapdump
        ----------events.log

Worker logs and daemon logs' Views on UI:
image

image

### DIFF
--- a/storm-core/src/clj/backtype/storm/daemon/logviewer.clj
+++ b/storm-core/src/clj/backtype/storm/daemon/logviewer.clj
@@ -190,7 +190,7 @@
       logs)))
 
 (defn per-workerdir-cleanup
-  "Delete the oldest files in overloaded worker log dir"
+  "Delete the oldest files in each overloaded worker log dir"
   [^File root-dir size]
   (dofor [worker-dir (get-all-worker-dirs root-dir)]
     (let [filtered-logs (filter #(not (is-active-log %)) (.listFiles worker-dir))
@@ -367,6 +367,9 @@ Note that if anything goes wrong, this will throw an Error and exit."
 (defn- daemon-download-link [fname]
   [[:p (link-to (url-format "/daemondownload/%s" fname) "Download Full File")]])
 
+(defn- is-txt-file [fname]
+  (re-find #"\.(log.*|txt|yaml|pid)$" fname))
+
 (def default-bytes-per-page 51200)
 
 (defn log-page [fname start length grep user root-dir]
@@ -390,9 +393,8 @@ Note that if anything goes wrong, this will throw an Error and exit."
                length (if length
                        (min 10485760 length)
                        default-bytes-per-page)
-              is-txt-file (re-find #"\.(log.*|txt|yaml|pid)$" fname)
               log-string (escape-html
-                           (if is-txt-file
+                           (if (is-txt-file fname)
                              (if start
                                (page-file path start length)
                                (page-file path length))
@@ -405,7 +407,7 @@ Note that if anything goes wrong, this will throw an Error and exit."
                           (filter #(.contains % grep))
                           (string/join "\n"))
                      log-string)])
-            (let [pager-data (if is-txt-file (pager-links fname start length file-length) nil)]
+            (let [pager-data (if (is-txt-file fname) (pager-links fname start length file-length) nil)]
               (html (concat (log-file-selection-form reordered-files-str "log") ; list all files for this topology
                             pager-data
                             (download-link fname)
@@ -420,7 +422,7 @@ Note that if anything goes wrong, this will throw an Error and exit."
 
 (defn daemonlog-page [fname start length grep user root-dir]
   (if (or (blank? (*STORM-CONF* UI-FILTER))
-        (authorized-log-user? user fname *STORM-CONF*)) ;; how to deal with this???????????
+        (authorized-log-user? user fname *STORM-CONF*))
     (let [file (.getCanonicalFile (File. root-dir fname))
           file-length (.length file)
           path (.getCanonicalPath file)
@@ -436,9 +438,8 @@ Note that if anything goes wrong, this will throw an Error and exit."
               files-str (for [file log-files]
                           (.getName file))
               reordered-files-str (conj (filter #(not= fname %) files-str) fname)
-              is-txt-file (re-find #"\.(log.*|txt|yaml|pid)$" fname)
               log-string (escape-html
-                           (if is-txt-file
+                           (if (is-txt-file fname)
                              (if start
                                (page-file path start length)
                                (page-file path length))
@@ -451,7 +452,7 @@ Note that if anything goes wrong, this will throw an Error and exit."
                        (filter #(.contains % grep))
                        (string/join "\n"))
                      log-string)])
-            (let [pager-data (if is-txt-file (pager-links fname start length file-length) nil)]
+            (let [pager-data (if (is-txt-file fname) (pager-links fname start length file-length) nil)]
               (html (concat (log-file-selection-form reordered-files-str "daemonlog") ; list all daemon logs
                       pager-data
                       (daemon-download-link fname)
@@ -561,9 +562,6 @@ Note that if anything goes wrong, this will throw an Error and exit."
         (log-error ex)
         (ring-response-from-exception ex))))
   (GET "/download/:file" [:as {:keys [servlet-request servlet-response log-root]} file & m]
-    ;; We do not use servlet-response here, but do not remove it from the
-    ;; :keys list, or this rule could stop working when an authentication
-    ;; filter is configured.
     (try
       (let [user (.getUserName http-creds-handler servlet-request)]
         (download-log-file file servlet-request servlet-response user log-root))
@@ -571,9 +569,6 @@ Note that if anything goes wrong, this will throw an Error and exit."
         (log-error ex)
         (ring-response-from-exception ex))))
   (GET "/daemondownload/:file" [:as {:keys [servlet-request servlet-response daemonlog-root]} file & m]
-    ;; We do not use servlet-response here, but do not remove it from the
-    ;; :keys list, or this rule could stop working when an authentication
-    ;; filter is configured.
     (try
       (let [user (.getUserName http-creds-handler servlet-request)]
         (download-log-file file servlet-request servlet-response user daemonlog-root))
--- a/storm-core/src/clj/backtype/storm/daemon/logviewer.clj
+++ b/storm-core/src/clj/backtype/storm/daemon/logviewer.clj
@@ -16,12 +16,14 @@
 (ns backtype.storm.daemon.logviewer
   (:use compojure.core)
   (:use [clojure.set :only [difference intersection]])
-  (:use [clojure.string :only [blank?]])
-  (:use [hiccup core page-helpers])
+  (:use [clojure.string :only [blank? split]])
+  (:use [hiccup core page-helpers form-helpers])
   (:use [backtype.storm config util log timer])
   (:use [backtype.storm.ui helpers])
+  (:import [backtype.storm.utils Utils])
   (:import [org.slf4j LoggerFactory])
-  (:import [java.io File FileFilter FileInputStream])
+  (:import [java.io File FileFilter FileInputStream InputStream])
+  (:import [java.util.zip GZIPInputStream])
   (:import [org.apache.logging.log4j LogManager])
   (:import [org.apache.logging.log4j.core Appender LoggerContext])
   (:import [org.apache.logging.log4j.core.appender RollingFileAppender])
@@ -29,14 +31,12 @@
            [org.yaml.snakeyaml.constructor SafeConstructor])
   (:import [backtype.storm.ui InvalidRequestException]
            [backtype.storm.security.auth AuthUtils])
-  (:require [compojure.route :as route]
-            [compojure.handler :as handler]
-            [ring.middleware.keyword-params]
-            [ring.util.response :as resp])
   (:require [backtype.storm.daemon common [supervisor :as supervisor]])
   (:import [java.io File FileFilter])
   (:require [compojure.route :as route]
             [compojure.handler :as handler]
+            [ring.middleware.keyword-params]
+            [ring.util.codec :as codec]
             [ring.util.response :as resp]
             [clojure.string :as string])
   (:gen-class))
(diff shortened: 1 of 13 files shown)


zhuoliu:
Reopen and make sure it connects with JIRA

storm-core/src/clj/backtype/storm/daemon/logviewer.clj
```diff
@@ -219,7 +321,7 @@ Note that if anything goes wrong, this will throw an Error and exit."
     (if (and appender-name appender (instance? RollingFileAppender appender))
       (.getParent (File. (.getFileName appender)))
       (throw
-       (RuntimeException. "Log viewer could not find configured appender, or the appender is not a FileAppender. Please check that the appender name configured in storm and log4j2 agree.")))))
+       (RuntimeException. "Log viewer could not find configured appender, or the appender is not a FileAppender. Please check that the appender name configured in storm and logback agree.")))))
```
revans2: This exception should be log4j2 not logback

storm-core/src/clj/backtype/storm/util.clj
```diff
@@ -580,6 +582,21 @@
...
+  ([path-dir target-dir file-name]
+    (create-symlink! path-dir target-dir file-name file-name))
+  ([path-dir target-dir from-file-name to-file-name]
+    (let [path (str path-dir file-path-separator from-file-name)
+          target (str target-dir file-path-separator to-file-name)
+          empty-array (make-array String 0)
+          attrs (make-array FileAttribute 0)
+          abs-path (.toAbsolutePath (Paths/get path empty-array))
+          abs-target (.toAbsolutePath (Paths/get target empty-array))]
+      (log-message "Creating symlink [" abs-path "] to [" abs-target "]")
```
revans2: This should probably be debug if we keep it around at all.

storm-core/src/jvm/backtype/storm/utils/Utils.java
```diff
@@ -692,5 +696,65 @@ public static void handleUncaughtException(Throwable t) {
...
+                }
+            }
+        } finally {
+            zipFile.close();
+        }
+    }
+
+    //Note: Only works for zip files whose uncompressed size is less than 4 GB
+    //Otherwise returns the size module 2^32, per gzip specifications
+    //Returns a long, since that's what file lengths in Java/Clojure usually are.
```
revans2: Can we turn this into a javadoc comment?

revans2:
@zhuoliu all of the tests are failing with
```
java.nio.file.NoSuchFileException: /tmp/667af8d4-e62c-4bee-a874-118c019c478d/workers-artifacts/test-1-1445964515/1027/events.log
	at sun.nio.fs.UnixException.translateToIOException(UnixException.java:86) ~[?:1.8.0_31]
	at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:102) ~[?:1.8.0_31]
	at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:107) ~[?:1.8.0_31]
	at sun.nio.fs.UnixFileSystemProvider.newByteChannel(UnixFileSystemProvider.java:214) ~[?:1.8.0_31]
	at java.nio.file.spi.FileSystemProvider.newOutputStream(FileSystemProvider.java:434) ~[?:1.8.0_31]
	at java.nio.file.Files.newOutputStream(Files.java:216) ~[?:1.8.0_31]
	at java.nio.file.Files.newBufferedWriter(Files.java:2860) ~[?:1.8.0_31]
	at backtype.storm.metric.FileBasedEventLogger.initLogWriter(FileBasedEventLogger.java:32) [classes/:?]
	at backtype.storm.metric.FileBasedEventLogger.prepare(FileBasedEventLogger.java:85) [classes/:?]
	at backtype.storm.metric.EventLoggerBolt.prepare(EventLoggerBolt.java:39) [classes/:?]
	at backtype.storm.daemon.executor$fn__7917$fn__7930.invoke(executor.clj:819) [classes/:?]
	at backtype.storm.util$async_loop$fn__1152.invoke(util.clj:480) [classes/:?]
	at java.lang.Thread.run(Thread.java:745) [?:1.8.0_31]
```

It looks like the event logger test is trying to write to a directory that does not exist.

revans2:
The log link on the UI for the nimbus.log is not working.  We probably should do something to special case the daemon log files.

revans2:
Oh and with adding in support for nimbus.log, we need to be careful that the code does not allow for files outside of the logs or the worker artifacts directories to be downloaded.

zhuoliu:
Thanks @revans2. 
New web routes are added for supporting view, list and download of all daemon logs.
The concern of directory access permission have been addressed also (through verification of the root log dir).

storm-core/src/clj/backtype/storm/daemon/logviewer.clj
```diff
@@ -294,9 +418,55 @@ Note that if anything goes wrong, this will throw an Error and exit."
         (resp/status 404))
       (unauthorized-user-html user))))
 
+(defn daemonlog-page [fname start length grep user root-dir]
+  (if (or (blank? (*STORM-CONF* UI-FILTER))
+        (authorized-log-user? user fname *STORM-CONF*)) ;; how to deal with this???????????
```
d2r: I think we can remove this comment.

storm-core/src/clj/backtype/storm/daemon/logviewer.clj
```diff
@@ -294,9 +418,55 @@ Note that if anything goes wrong, this will throw an Error and exit."
...
+            (.exists file))
+        (let [file-length (if zip-file? (Utils/zipFileSize (clojure.java.io/file path)) (.length (clojure.java.io/file path)))
+              length (if length
+                       (min 10485760 length)
+                       default-bytes-per-page)
+              log-files (into [] (filter #(.isFile %) (.listFiles (File. root-dir)))) ;all types of files included
+              files-str (for [file log-files]
+                          (.getName file))
+              reordered-files-str (conj (filter #(not= fname %) files-str) fname)
+              is-txt-file (re-find #"\.(log.*|txt|yaml|pid)$" fname)
```
d2r: We might have pulled this out into its own function, since it is used at least twice now. Not a big deal though.

storm-core/src/clj/backtype/storm/daemon/logviewer.clj
```diff
@@ -333,42 +503,112 @@ Note that if anything goes wrong, this will throw an Error and exit."
...
   (GET "/download/:file" [:as {:keys [servlet-request servlet-response log-root]} file & m]
-       (try
-         (let [user (.getUserName http-creds-handler servlet-request)]
-           (download-log-file file servlet-request servlet-response user log-root))
-         (catch InvalidRequestException ex
-           (log-error ex)
-           (ring-response-from-exception ex))))
+    ;; We do not use servlet-response here, but do not remove it from the
+    ;; :keys list, or this rule could stop working when an authentication
+    ;; filter is configured.
```
d2r: It seems we do use `servlet-response` here, so can we remove this comment?

storm-core/src/clj/backtype/storm/daemon/logviewer.clj
```diff
@@ -333,42 +503,112 @@ Note that if anything goes wrong, this will throw an Error and exit."
...
+    (try
+      (let [user (.getUserName http-creds-handler servlet-request)]
+        (download-log-file file servlet-request servlet-response user log-root))
+      (catch InvalidRequestException ex
+        (log-error ex)
+        (ring-response-from-exception ex))))
+  (GET "/daemondownload/:file" [:as {:keys [servlet-request servlet-response daemonlog-root]} file & m]
+    ;; We do not use servlet-response here, but do not remove it from the
+    ;; :keys list, or this rule could stop working when an authentication
+    ;; filter is configured.
```
d2r: Same comment; remove this text?

d2r:
Just some minor comments.

revans2:
Looks good I am +1 once the comments from @d2r are addressed.

zhuoliu:
Really appreciate the review efforts. @d2r 's comments have been addressed, welcome to check back.

d2r:
+1

harshach:
@zhuoliu  @revans2 @d2r Hi,  @arunmahadevan pointed out this jira changed log location for worker logs. This is actually not acceptable for our users and I don't see any reason for log location to change since we've storm.log.dir and its been used in previous releases . If its necessary for this patch to function to have the logs in storm-local/worker-artifacts than we should put  in storm.log.dir and symlink to storm-local/worker-artifacts.  Checking to see the reason behind moving the logs there before sending a patch. Thanks.

zhuoliu:
Hi @harshach , the main reasons we change log location are two-fold: 1) we want to separate the daemon logs from worker logs for better namespace and permission management; 2) the workers-artifacts directory are beyond "just logs", it appears more like a drop-box for each worker, containing log files, heapdump, gc files, and other possible files users may need to dump. 
It will be one line change to "config/worker-artifacts-root" function if we want to place worker-artifacts under storm-home/logs/. In that way it will look as follows. In our opinion, it would be preferable to let workers-artifacts to stay with storm-local/workers, worker-users.
We can discuss which one is more appropriate.

storm-home/logs/
------supervisor.log
------nimbus.log
------workers-artifacts/
-----------topo1/
------------------port1/
------------------port2/
------------------------worker.log
------------------------worker.log.err
------------------------gc.log

arunmahadevan:
@zhuoliu additionally can we create symlinks from storm-home/logs/worker-port.log -> workers-artifacts/topo/port/worker.log so that all the recent worker logs can be found under a single dir ?

unsleepy22:
I agree with @harshach , it's a big breaking change of log layout, as @arunmahadevan says, maybe it's  better to make a symlink?

zhuoliu:
That is doable though weird, but I would be interested to see why we want all logs to be found under a single directory.

zhuoliu:
On the contrary, say a supervisor has been running for a long period (say a week), 24 workers * 10 restarting* 40 files per worker (10 worker.log.gz files, 10 gc log files, error files, out files and many other files). We will have around 10000 files from all topologies' workers and all daemons in the single storm-home directory, with tedious long file names. This may not be an ideal status.

arunmahadevan:
@zhuoliu i meant only the current worker-port.log file, which would make it a bit easier to switch between files while tailing. Right now we need to change the entire path to switch to a different worker log file. The rotated and log.gz files can be under workers-artifacts.

zhuoliu:
@arunmahadevan , if we do have this application scenario, it would be nice to do that. We can let the supervisor to redirect(create) the worker-port.log symlink to the topo/port/worker.log in launcher-worker.

d2r:
@harshach
Use case:
A cluster admin shelling to a supervisor node and browsing all current logs easily, without having to `cd` to other directories?

Are there other use cases? What makes this unacceptable to users?  Maybe we can come up with something better in general.

> We can let the supervisor to redirect(create) the worker-port.log symlink to the topo/port/worker.log in launch-worker. Thanks for the explain. Btw, we should care that some symlink may become dead link since its pointed file might have been deleted by cleanup.

I am not excited about doing this, because it adds complexity to storm that ought to be handled by either a logging system or shell environment. It seems to me to make more sense to provide a tool for shells to make browsing logs easier where they are, rather than adding complexity to storm to support the shell.

harshach:
@arunmahadevan @d2r @zhuoliu  As Arun said above if the current worker-port.log is in root dir it will be helpful to tail the logs. This preferable but not necessary. My main concern is we are not using storm.log.dir and adding logs to storm-local dir which is supposed to be used to store storm metadata dir and other topology metadata not application logs. And changing log location of topologies while using storm.log.dir is not a desired behavior. If we can store all the logs in user configured storm.log.dir than I am fine with current structure.

erikdw:
+1 to @harshach 's comment.  We rely on being able to modify the `storm.log.dir` location to change where logs go.

revans2:
@harshach and @erikdw can you explain the exact use cases you are concerned about?  From previous interactions it feels like you are concerned about storm on slider on YARN and storm on mesos respectively.  So the fix is to place the logs in a configurable place that other tools can get access to?

I just want to be sure I understand the use case so I can properly address those concerns.  We can move the worker-artifacts dir to be in a configurable location.  That should be totally doable.  I wouldn't really want them all to be under storm.log.dir, as @zhuoliu pointed out there are more than just logs stored here.  As part of this change we essentially upgraded the logs to be application metadata that can have limits placed on them and managed.  But if you really want it based off of storm.log.dir as a default it is not a deal breaker.

erikdw:
@revans2 :  good memory, my use-case is indeed storm-on-mesos.  And in the way my company deploys it, we leverage the `storm.log.dir` to force the logs to be written into a known location where we can forward them to various downstream log-indexing/storing services.  That is the use case we want to be supported *somehow*.  I'm not particularly married to `storm.log.dir`, so your proposal to have the `worker-artifacts` go to some other location seems sufficient for my use case.  I'll explicitly state that I would expect it to be a knob that is configurable from within `storm.yaml`.

BTW, changing to a hierarchical layout from a flat style was concerning at first blush, but on further reflection I believe it will work just fine.  Our tools for log-forwarding can use globs/wildcards so it should be ok.

revans2:
@erikdw Thanks for the feedback, yes the changes should be something that you can do through `storm.yaml`.

@harshach do you have any feedback?

harshach:
@revans2 I want user experience to be consistent . I.e when they configure storm.log.dir , the logs to be appear in configured storm.log.dir , not some of the logs in storm.log.dir and others in storm-local/worker-artifacts. Its hard to understand why they are ending up there. In some cases storm-local dir might only accessed by linux user "storm" who is running the daemons. My request is to keep the current hierarchical dir that we've right now but move all the logs to configured storm.log.dir not have it in storm-local/worker-artifacts. Not sure if this requires lot of changes for this patch.

revans2:
It would be a rather large change to support two different directory structures, one for logs and one for everything else.  It is not impossible to do, but it would add quite a bit of complexity to something that is already more complex then I like.

I guess it comes down to how often do you and your customers access the logs through the storm UI or some other automated log system vs how often you access the logs by going to the box directly and reading them directly.  For us our clusters have grown large enough that very few people try to read the logs directly.  Finding the right node is such a pain that the added directly structure is not something they have complained about.

Would having the worker-artifacts directory structure by default be under `storm.log.dir` but also have a separate config be an acceptable solution?  This is a very simple change to make, vs trying to split the data.

revans2:
@harshach sorry I misread your comment.

Moving them to be under storm-log-dir by default would be fine.  I would want a config to put it some other place if we want to, but the default could be under the log directory.

@zhuoliu could you file a JIRA for this? it should be just a few lines of change to try and read a new config with a default being the value of `storm.log.dir`

zhuoliu:
sure. this conclusion makes sense to me and won't be much change. A JIRA is filed here STORM-1387

unsleepy22:
@revans2 @zhuoliu This solution looks good to me too.
My initial concern was sort of like @erikdw 's use case. And usually files in logs directory do a rolling update/delete(maybe by a cron job), while we may not want a rolling delete for metadata(or at least different delete stragety), so it's good to add an extra config in storm.yaml to specify this directory.

harshach:
Thanks @zhuoliu @revans2

PR #827, opened 2015-10-28
STORM-1075 add external module storm-cassandra

fhussonnois:
https://issues.apache.org/jira/browse/STORM-1075

### DIFF
--- a/external/storm-cassandra/src/main/java/org/apache/storm/cassandra/Murmur3StreamGrouping.java
+++ b/external/storm-cassandra/src/main/java/org/apache/storm/cassandra/Murmur3StreamGrouping.java
@@ -1,0 +1,89 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.storm.cassandra;
+
+import backtype.storm.generated.GlobalStreamId;
+import backtype.storm.grouping.CustomStreamGrouping;
+import backtype.storm.task.WorkerTopologyContext;
+import backtype.storm.topology.FailedException;
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.Lists;
+import com.google.common.hash.Hashing;
+
+import java.io.ByteArrayOutputStream;
+import java.io.DataOutputStream;
+import java.io.IOException;
+import java.util.List;
+
+/**
+ *
+ * Simple {@link backtype.storm.grouping.CustomStreamGrouping} that uses Murmur3 algorithm to choose the target task of a tuple.
+ *
+ * This stream grouping may be used to optimise writes to Apache Cassandra.
+ */
+public class Murmur3StreamGrouping implements CustomStreamGrouping {
+
+    private List<Integer> targetTasks;
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void prepare(WorkerTopologyContext context, GlobalStreamId stream, List<Integer> targetTasks) {
+        this.targetTasks = targetTasks;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public List<Integer> chooseTasks(int taskId, List<Object> values) {
+        try {
+            int n = Math.abs( (int) hashes(values) % targetTasks.size() );
+            return Lists.newArrayList(targetTasks.get(n));
+        } catch (IOException e) {
+            throw new FailedException(e);
+        }
+    }
+
+    /**
+     * Computes the murmur3 hash for the specified values.
+     * http://stackoverflow.com/questions/27212797/cassandra-hashing-algorithm-with-composite-keys
+     * https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/db/marshal/CompositeType.java
+     *
+     * @param values the fields which are part of the (compose) partition key.
+     * @return the computed hash for input values.
+     * @throws java.io.IOException
+     */
+    @VisibleForTesting
+    public static long hashes(List<Object> values) throws IOException {
+        byte[] keyBytes;
+        try(ByteArrayOutputStream bos = new ByteArrayOutputStream(); DataOutputStream out = new DataOutputStream(bos)) {
+            for(Object key : values) {
+                byte[] arr = ((String)key).getBytes("UTF-8");
+                out.writeShort(arr.length);
+                out.write(arr, 0, arr.length);
+                out.writeByte(0);
+            }
+            out.flush();
+            keyBytes = bos.toByteArray();
+        }
+        return Hashing.murmur3_128().hashBytes(keyBytes).asLong();
+    }
+}
--- a/external/storm-cassandra/src/main/java/org/apache/storm/cassandra/Murmur3StreamGrouping.java
+++ b/external/storm-cassandra/src/main/java/org/apache/storm/cassandra/Murmur3StreamGrouping.java
@@ -22,13 +22,18 @@ import backtype.storm.generated.GlobalStreamId;
 import backtype.storm.grouping.CustomStreamGrouping;
 import backtype.storm.task.WorkerTopologyContext;
 import backtype.storm.topology.FailedException;
+import backtype.storm.tuple.Fields;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.Lists;
 import com.google.common.hash.Hashing;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.io.ByteArrayOutputStream;
 import java.io.DataOutputStream;
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 
 /**
@@ -41,12 +46,43 @@ public class Murmur3StreamGrouping implements CustomStreamGrouping {
 
     private List<Integer> targetTasks;
 
+    private List<Integer> partitionKeyIndexes;
+
+    /**
+     * A list of partition key. The order of specified keys will be used to generate the partition key hash.
+     * It should respect the column order defined into the targeted CQL table.
+     */
+    private List<String> partitionKeyNames;
+
+    /**
+     * Creates a new {@link Murmur3StreamGrouping} instance.
+     * @param partitionKeyNames {@link org.apache.storm.cassandra.Murmur3StreamGrouping#partitionKeyNames}.
+     */
+    public Murmur3StreamGrouping(String...partitionKeyNames) {
+        this( Arrays.asList(partitionKeyNames));
+    }
+
+    /**
+     * Creates a new {@link Murmur3StreamGrouping} instance.
+     * @param partitionKeyNames {@link org.apache.storm.cassandra.Murmur3StreamGrouping#partitionKeyNames}.
+     */
+    public Murmur3StreamGrouping(List<String> partitionKeyNames) {
+        this.partitionKeyNames = partitionKeyNames;
+    }
+
+
     /**
      * {@inheritDoc}
      */
     @Override
     public void prepare(WorkerTopologyContext context, GlobalStreamId stream, List<Integer> targetTasks) {
         this.targetTasks = targetTasks;
+
+        this.partitionKeyIndexes = new ArrayList<>();
+        Fields componentOutputFields = context.getComponentOutputFields(stream);
+        for (String partitionKeyName : partitionKeyNames) {
+            partitionKeyIndexes.add(componentOutputFields.fieldIndex(partitionKeyName));
+        }
     }
 
     /**
@@ -55,11 +91,19 @@ public class Murmur3StreamGrouping implements CustomStreamGrouping {
     @Override
     public List<Integer> chooseTasks(int taskId, List<Object> values) {
         try {
-            int n = Math.abs( (int) hashes(values) % targetTasks.size() );
+            int n = Math.abs( (int) hashes(getKeyValues(values)) % targetTasks.size() );
             return Lists.newArrayList(targetTasks.get(n));
         } catch (IOException e) {
             throw new FailedException(e);
         }
+    }
+
+    private List<Object> getKeyValues(List<Object> values) {
+        List<Object> keys = new ArrayList<>();
+        for(Integer idx : partitionKeyIndexes) {
+            keys.add(values.get(idx));
+        }
+        return keys;
     }
 
     /**
--- a/external/storm-cassandra/src/main/java/org/apache/storm/cassandra/bolt/BaseCassandraBolt.java
+++ b/external/storm-cassandra/src/main/java/org/apache/storm/cassandra/bolt/BaseCassandraBolt.java
@@ -26,6 +26,7 @@ import backtype.storm.topology.base.BaseRichBolt;
 import backtype.storm.tuple.Fields;
 import backtype.storm.tuple.Tuple;
 import backtype.storm.utils.TupleUtils;
+import backtype.storm.utils.Utils;
 import com.datastax.driver.core.Session;
 import com.datastax.driver.core.exceptions.NoHostAvailableException;
 import org.apache.storm.cassandra.BaseExecutionResultHandler;
@@ -105,7 +106,7 @@ public abstract class BaseCassandraBolt<T> extends BaseRichBolt {
     }
 
     public BaseCassandraBolt withOutputFields(Fields fields) {
-        this.outputsFields.put(null, fields);
+        this.outputsFields.put(Utils.DEFAULT_STREAM_ID, fields);
         return this;
     }
 
@@ -139,7 +140,7 @@ public abstract class BaseCassandraBolt<T> extends BaseRichBolt {
     public final void execute(Tuple input) {
         getAsyncHandler().flush(outputCollector);
         if (TupleUtils.isTick(input)) {
-            tick();
+            onTickTuple();
             outputCollector.ack(input);
         } else {
             process(input);
@@ -156,14 +157,14 @@ public abstract class BaseCassandraBolt<T> extends BaseRichBolt {
     /**
      * Calls by an input tick tuple.
      */
-    abstract protected void tick();
+    abstract protected void onTickTuple();
 
     /**
      * {@inheritDoc}
      */
     @Override
     public void declareOutputFields(OutputFieldsDeclarer declarer) {
-        Fields fields = this.outputsFields.remove(null);
+        Fields fields = this.outputsFields.remove(Utils.DEFAULT_STREAM_ID);
         if( fields != null) declarer.declare(fields);
         for(Map.Entry<String, Fields> entry : this.outputsFields.entrySet()) {
             declarer.declareStream(entry.getKey(), entry.getValue());
--- a/external/storm-cassandra/src/main/java/org/apache/storm/cassandra/bolt/BaseCassandraBolt.java
+++ b/external/storm-cassandra/src/main/java/org/apache/storm/cassandra/bolt/BaseCassandraBolt.java
@@ -1,0 +1,193 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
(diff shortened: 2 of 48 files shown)


HeartSaVioR:
Before seeing the details, 
1. It would be better to use source/target version to 1.7. Storm changes its base JDK version to 1.7, but not yet 1.8. Did you use specific syntax or set of APIs provided to JDK 1.8? 
2. Please remove author tag ("@author fhussonnois") from all comments.
3. Please add storm-cassandra to storm-dist/binary/src/main/assembly/binary.xml. You can refer other modules to how to add.

fhussonnois:
I have updated the PR with the remarks.

harshach:
@satishd can you please review this.

ptgoetz:
+1 for adding this.

I've done a lot of work with Cassandra in the past (Back when the thrift API was the only game in town)
 and have always wanted to see first class support for it in Storm -- even if it is an optional component.

Moving to the DS Java driver is clearly the only way to go.

I'd be willing to act as a sponsor for this as well.

external/storm-cassandra/src/main/java/org/apache/storm/cassandra/Murmur3StreamGrouping.java
```diff
@@ -0,0 +1,89 @@
...
+import java.io.IOException;
+import java.util.List;
+
+/**
+ *
+ * Simple {@link backtype.storm.grouping.CustomStreamGrouping} that uses Murmur3 algorithm to choose the target task of a tuple.
+ *
+ * This stream grouping may be used to optimise writes to Apache Cassandra.
+ */
+public class Murmur3StreamGrouping implements CustomStreamGrouping {
```
harshach: what happens if the users chose to use shuffle grouping will write perform badly?

external/storm-cassandra/src/main/java/org/apache/storm/cassandra/bolt/BatchCassandraWriterBolt.java
```diff
@@ -0,0 +1,192 @@
...
+            LOG.info(logPrefix() + "The message queue is full, preparing batch statement...");
+            prepareAndExecuteStatement();
+            queue.add(input);
+        }
+    }
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    protected void tick() {
```
harshach: can we rename this method other than tick.

external/storm-cassandra/src/main/java/org/apache/storm/cassandra/bolt/BatchCassandraWriterBolt.java
```diff
@@ -0,0 +1,192 @@
...
+        return this;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public Map<String, Object> getComponentConfiguration() {
+        Config conf = new Config();
+        conf.put(Config.TOPOLOGY_TICK_TUPLE_FREQ_SECS, tickFrequencyInSeconds);
```
harshach: we should also have a default value that is less than topology.message.timeout.secs

harshach:
@fhussonnois sorry for the delay in reviewing this. I've few minor nitpicks overall it looks good to me. 
@satishd also worked on similar connector, I would like to get his reviews in. I am also willing to act as sponsor.

external/storm-cassandra/src/main/java/org/apache/storm/cassandra/client/ClusterFactory.java
```diff
@@ -0,0 +1,71 @@
...
+    @Override
+    protected Cluster make(Map<String, Object> stormConf) {
+        CassandraConf cassandraConf = new CassandraConf(stormConf);
+
+        Cluster.Builder cluster = Cluster.builder()
+                .withoutJMXReporting()
+                .withoutMetrics()
+                .addContactPoints(cassandraConf.getNodes())
+                .withPort(cassandraConf.getPort())
+                .withRetryPolicy(DowngradingConsistencyRetryPolicy.INSTANCE)
```
satishd: DefaultRetryPolicy should be used as default policy which is conservative. DowngradingConsistencyRetryPolicy is kind of aggressive and this should not be the default setting. You may want to take this policy from conf.

external/storm-cassandra/src/main/java/org/apache/storm/cassandra/client/ClusterFactory.java
```diff
@@ -0,0 +1,71 @@
...
+                .withReconnectionPolicy(new ExponentialReconnectionPolicy(100L, TimeUnit.MINUTES.toMillis(1)))
```
satishd: Better not to have hard coded timeout values. Create respective properties in config.

external/storm-cassandra/src/main/java/org/apache/storm/cassandra/client/impl/DefaultClient.java
```diff
@@ -0,0 +1,123 @@
...
+        Thread thread = Thread.currentThread();
+        return thread.getName();
+    }
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public synchronized Session connect() throws NoHostAvailableException {
+        if( isDisconnected() ) {
+            LOG.info(String.format("Connected to cluster: %s", cluster.getClusterName()));
```
satishd: Can you replace String.format with LOG.info(String format, Object... arguments) as it does not build the string if the respective logging level is not enabled. It saves building strings unnecessarily. You may want to refactor it at other log usages.

external/storm-cassandra/src/main/java/org/apache/storm/cassandra/bolt/BaseCassandraBolt.java
```diff
@@ -0,0 +1,193 @@
...
+        }
+    }
+
+    public BaseCassandraBolt withResultHandler(ExecutionResultHandler resultHandler) {
+        this.resultHandler = resultHandler;
+        return this;
+    }
+
+    public BaseCassandraBolt withOutputFields(Fields fields) {
+        this.outputsFields.put(null, fields);
```
satishd: Not a good practice to pass null to a map. It seems this is passed to set fields for default stream. You should use backtype.storm.utils.Utils.DEFAULT_STREAM_ID as key instead of null.

external/storm-cassandra/src/main/java/org/apache/storm/cassandra/executor/AsyncExecutor.java
```diff
@@ -0,0 +1,152 @@
...
+
+    private final static Logger LOG = LoggerFactory.getLogger(AsyncExecutor.class);
+
+    protected Session session;
+
+    protected ExecutorService executorService;
+
+    protected AsyncResultHandler<T> handler;
+
+    private Map<SettableFuture<T>, Boolean> pending = new ConcurrentHashMap<>( );
```
satishd: Do we really need a map for this? AtomicInteger may be sufficient for now. We can add map implementation if we need it.

external/storm-cassandra/src/main/java/org/apache/storm/cassandra/executor/AsyncExecutor.java
```diff
@@ -0,0 +1,152 @@
...
+                handler.failure(t, inputs);
+            }
+        }, executorService);
+        return settableFuture;
+    }
+
+    /**
+     * Returns the number of currently executed tasks which are not yet completed.
+     */
+    public int getPendingExec( ) {
```
satishd: Rename to getPendingTasksSize()

external/storm-cassandra/src/main/java/org/apache/storm/cassandra/Murmur3StreamGrouping.java
fhussonnois: @harshach, there will be no performance issue using shuffle, field or murmur3. This strategy should be used with the BatchCassandraBolt in order to group all tuples that will be written into a same partition to the same storm task.

In fact, to stream the tuples according to cassandra (data location) we need to retrieve ip addresses of each task within the CustomStreamGrouping. I don't know if that is possible ?

satishd:
STORM-1211 and STORM-1212 are filed for trident state/query support and synchronous bolt support. I have partially worked on these as part of this JIRA. I will update those JIRAs with pull requests once I have them ready.

external/storm-cassandra/src/main/java/org/apache/storm/cassandra/bolt/BatchCassandraWriterBolt.java
```diff
@@ -0,0 +1,192 @@
...
+        this.tickFrequencyInSeconds = tickFrequencyInSeconds;
+    }
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void prepare(Map stormConfig, TopologyContext topologyContext, OutputCollector outputCollector) {
+        super.prepare(stormConfig, topologyContext, outputCollector);
+        this.componentID = topologyContext.getThisComponentId();
+        this.queue = new LinkedBlockingQueue<>(QUEUE_MAX_SIZE);
```
satishd: queue size should be configurable at bolt like withQueueSize or withBatchSize.

external/storm-cassandra/src/main/java/org/apache/storm/cassandra/bolt/BatchCassandraWriterBolt.java
fhussonnois: what names should be used ? onTickTuple, onSystemTuple...

external/storm-cassandra/src/main/java/org/apache/storm/cassandra/Murmur3StreamGrouping.java
jnioche: Would it make sense to provide this class as part of the code module? It was mentioned in a [discussion on the user list](https://mail-archives.apache.org/mod_mbox/incubator-storm-user/201510.mbox/%3CCAC19-6AKWc8cH681muaQ1LsFkwiofQhaffjKESfU04LpVoRcow@mail.gmail.com%3E) not long ago and it sounded like it gives better results in same cases.

external/storm-cassandra/src/main/java/org/apache/storm/cassandra/bolt/BatchCassandraWriterBolt.java
```diff
@@ -0,0 +1,192 @@
...
+    public void prepareAndExecuteStatement() {
+        int size = queue.size();
+        if( size > 0 ) {
+            List<Tuple> inputs = new ArrayList<>(size);
+            queue.drainTo(inputs);
+            try {
+                List<PairStatementTuple> psl = buildStatement(inputs);
+
+                int sinceLastModified = updateAndGetSecondsSinceLastModified();
+                LOG.debug(logPrefix() + String.format("Execute cql batches with %s statements after %s seconds", size, sinceLastModified));
```
jnioche: Using SLF4J's {} placeholders would be more elegant and also efficient; see [http://www.slf4j.org/faq.html#logging_performance]

external/storm-cassandra/src/main/java/org/apache/storm/cassandra/client/CassandraConf.java
```diff
@@ -0,0 +1,146 @@
...
+
+    private <T> T get(Map<String, Object> conf, String key) {
+        Object o = conf.get(key);
+        if(o == null) {
+            throw new IllegalArgumentException("No '" + key + "' value found in configuration!");
+        }
+        return (T)o;
+    }
+
+    private <T> T getOrElse(Map<String, Object> conf, String key, T def) {
```
jnioche: That does pretty much the same as [https://github.com/apache/storm/blob/768a85926373355c15cc139fd86268916abc6850/storm-core/src/jvm/backtype/storm/utils/Utils.java#L344], doesn't it?

external/storm-cassandra/src/main/java/org/apache/storm/cassandra/executor/AsyncExecutor.java
```diff
@@ -0,0 +1,152 @@
...
+        this.handler = handler;
+    }
+
+    protected static ExecutorService newSingleThreadExecutor() {
+        return Executors.newSingleThreadExecutor(new ThreadFactoryBuilder().setNameFormat("cassandra-async-handler-%d").build());
+    }
+
+    /**
+     * Asynchronously executes all statements associated to the specified input. The input will be passed to
+     * the {@link #handler} once all queries succeed or failed.
```
satishd: The input will be passed to handler#onSuccess once all queries succeed or to handler#onFailure if any one of them fails. Is that right?

external/storm-cassandra/src/main/java/org/apache/storm/cassandra/Murmur3StreamGrouping.java
satishd: @fhussonnois Cassandra should be partitioning rows based on partition key on a row according to the configured partitioner. But here, the grouping is done based on the hash generated on tuple values which may contain values other than partition key. Can you elaborate on this?  Pl add more details in java doc. 

It maybe helpful for users to have information about this grouping in ReadMe.md.

satishd:
@fhussonnois Can you add an example(or an integration test, which can be disabled for now)? It will be helpful for users to try Cassandra connector.

external/storm-cassandra/src/main/java/org/apache/storm/cassandra/Murmur3StreamGrouping.java
fhussonnois: Yes, that's right! Actually, I didn't even noticed there is no method like this on component : customGrouping(componentId, customStreamGrouping, fields...).

So, one solution could be to pass the indexes of the the partition keys as follows : 
myBolt.customGrouping("comp", new Murmur3StreamGrouping(0,1)) ?

satishd: It is better to take field names instead of indexes. You can get output declared fields by using the API: WorkerTopologyContext#getComponentOutputFields(GlobalStreamId id) in prepare method and store the respective indexes like below.

    public void prepare(WorkerTopologyContext workerTopologyContext, GlobalStreamId globalStreamId, List<Integer> targetTasks) {
        this.targetTasks = targetTasks;

        partitionKeyIndexes = new ArrayList<>();
        Fields componentOutputFields = workerTopologyContext.getComponentOutputFields(globalStreamId);
        for (String partitionKeyName : partitionKeyNames) {
            partitionKeyIndexes.add(componentOutputFields.fieldIndex(partitionKeyName));
        }
    }

external/storm-cassandra/src/main/java/org/apache/storm/cassandra/bolt/BatchCassandraWriterBolt.java
satishd: onTickTuple looks good to me.

external/storm-cassandra/src/main/java/org/apache/storm/cassandra/executor/AsyncExecutor.java
fhussonnois: Yes there is an issue on the comment. This behavior is implemented with 
http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/util/concurrent/Futures.html#allAsList(java.lang.Iterable)

satishd:
@fhussonnois I am working on Select statement and Clause query implementation as part of STORM-1211. You can incorporate remaining comments in your code or let me know if you have need any clarifications.

satishd:
@fhussonnois I think we should accept only cql strings for now instead of giving a fluent API for building queries. This requires implementing all kinds of queries supported by cql just to map column names. Cassandra connector API should be agnostic about the cql by simply using datastax driver. This avoids implementing any new features being added in cql. We should support only simple/prepared statement builder.

fhussonnois:
@satishd I think the fluent API handles common use cases. Its purpose is to ease the cassandra integration because creating a CQL statement from tuple is cumbersome. In case of specifics needs developers can still implement their own CQLStatementTupleMapper. So I think it will be bad to remove that API. In fact, there is a risk that developers that will use the connector have to re-develop a DSL.
I'm not sure, but I think there is not a lot changes into cassandra connector API. DataStax seems to focus their developments onto the cassandra internal system and not onto query language support.

satishd:
@fhussonnois  What I said is building insert or update query with where clause etc can be done by query builder apis or they can give cql statement. But mapping tuples/columns will still be done using fluent APIs. Users do not need to write their own tuple/column mappers and they can use user friendly fluent API.

@harshach What is your opinion on this?

harshach:
@fhussonnois lets use the cql as @satishd suggested and provide a mapper interface like Jdbc https://github.com/apache/storm/tree/master/external/storm-jdbc which can map tuple fields to table column. mapper interface used across all the connectors which will make it easier for users.

satishd:
We can use/refactor the existing fluent API for mappers as I mentioned in my earlier comment. Users can either set their own mappers or use the default mapper implemetation with fluent API.

fhussonnois:
@satishd, @harshach ok perfect. So I will refactor API to remove Insert/Update builder. 
Then I add a simple interface to map tuples to columns : 
```java
public interface CqlMapper  extends Serializable {
    List<Column> getColumns(ITuple tuple);
}
```
In addition, I will implement a SimpleStatementMapper to build query as follows : 

```java
new CassandraWriterBolt(
  async(
    simpleQuery("INSERT INTO weather.temperature (weatherstation_id, event_time, temperature) VALUES(?,?,?)"),with(field("weatherstation_id"), field("event_time").now(), field("temperature")));
  )
);
```
Same Bolt could be write with QueryBuilder : 
```java
new CassandraWriterBolt(
  async(
simple(
                QueryBuilder.insertInto("weather", "temperature")
                        .value("weatherstation_id", "?")
                        .value("event_time", "?")
                        .value("temperature", "?")
                ,
                with(field("weatherstation_id"), field("event_time").now(), field("temperature")));
  )
);
```
What do you think about that ?

ptgoetz:
@fhussonnois I think that sounds like a good approach.

satishd:
@fhussonnois That was the idea, It looks good to me. I assume you still have field("stn_id").as("weatherstation_id") along with the APIs you mentioned.

harshach:
@fhussonnois looks good to me as well.

satishd:
@fhussonnois I already have some of those mappers code and trident related code with me. I can work on mappers and integrate with field-tuple/column fluent api. Can you refactor your API to remove insert/update builder and push those changes? I can take over from there to complete the remaining changes.

ptgoetz:
@satishd I think that work can be done in follow-up JIRAs after this PR is merged.

satishd:
Created https://issues.apache.org/jira/browse/STORM-1348 for refactoring API to remove insert/update builder.

+1, these changes look good to me.

satishd:
@fhussonnois You may want to squash/rebase these 14 commits into 1 or 2 commits.

harshach:
@fhussonnois I am getting following errors in the tests can you take a look
Tests in error: 
  BatchCassandraWriterBoltTest.shouldBatchInsertGivenStaticTableNameAndDynamicQueryBuildFromAllTupleFields:41->executeAndAssertWith:57->BaseTopologyTest.runLocalTopologyAndWait:49 » IllegalState
  CassandraWriterBoltTest.shouldAsyncInsertGivenStaticTableNameAndDynamicQueryBuildFromAllTupleFields:40->executeAndAssertWith:56->BaseTopologyTest.runLocalTopologyAndWait:49 » IllegalState

fhussonnois:
@harshach I can't reproduce the error. All tests pass successfully. Did you run tests after merging the branch ?

satishd:
@harshach I ran multiple times but I could not reproduce the issue. Is your zookeeper already running before tests are run?

harshach:
@fhussonnois @satishd Yes after merging the branch. I am running mvn clean package inside storm-cassandra or at the top level.
@satishd what you mean my zookeeper running before tests ran? is that a requirement?

satishd:
I ran "mvn clean install" in storm root directory and it worked fine. I had seen some intermittent failures when zookeeper was already running while running "mvn clean install" in storm because of unexpected state in zookeeper. It is not mandatory to have zookeeper running but it is better not to be running while tests are run.

harshach:
@fhussonnois @satishd I feel like the tests are flaky and I can't make them all pass
latest run
CassandraWriterBoltTest.shouldAsyncInsertGivenStaticTableNameAndDynamicQueryBuildFromAllTupleFields:40->executeAndAssertWith:56->BaseTopologyTest.runLocalTopologyAndWait:50 expected:<61234> but was:<95083> . 
We can't really expect these to pass all the time when we have sleep(time) in unit tests.  

and another run

 BatchCassandraWriterBoltTest.shouldBatchInsertGivenStaticTableNameAndDynamicQueryBuildFromAllTupleFields:41->executeAndAssertWith:57->BaseTopologyTest.runLocalTopologyAndWait:50 expected:<11627> but was:<15791>
  CassandraWriterBoltTest.shouldAsyncInsertGivenStaticTableNameAndDynamicQueryBuildFromAllTupleFields:40->executeAndAssertWith:56->BaseTopologyTest.runLocalTopologyAndWait:50 expected:<66740> but was:<81138>


@fhussonnois  lets remove these tests for now and we can address in another jira

fhussonnois:
@harshach @satishd I added ignore annotation to skip these tests.

harshach:
Thanks for the contribution @fhussonnois and keeping up with reviews. I merged into master.
Thanks @satishd for reviews.

fhussonnois:
Thank you @harshach @satishd. I will start working on https://issues.apache.org/jira/browse/STORM-1348 as soon as possible. BTW is my name will be add to the contributors list ? ;)

HeartSaVioR:
@fhussonnois Sorry about missing that thing. I just added you as contributor to README.md.

PR #830, opened 2015-10-29
 STORM-1145: Have IConnection push tuples instead of pull them.

revans2:
This pull request is based off of #765 because it moved the overflow buffer into the disruptor queue itself, and we don't want the netty thread to block when doing routing.

This does not improve the maximum throughput much, but because it removes an extra thread and an extra queue that a tuple must go through when being sent to another worker the 99%-ile, 99.9%-ile and worst case latency goes down some.

### DIFF
--- a/storm-core/src/clj/backtype/storm/daemon/executor.clj
+++ b/storm-core/src/clj/backtype/storm/daemon/executor.clj
@@ -21,7 +21,7 @@
   (:import [java.util List Random HashMap ArrayList LinkedList Map])
   (:import [backtype.storm ICredentialsListener])
   (:import [backtype.storm.hooks ITaskHook])
-  (:import [backtype.storm.tuple Tuple Fields TupleImpl MessageId])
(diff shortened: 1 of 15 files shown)


storm-core/src/clj/backtype/storm/disruptor.clj
```diff
@@ -27,10 +27,10 @@
    :single-threaded ProducerType/SINGLE})
 
 (defnk disruptor-queue
-  [^String queue-name buffer-size timeout :producer-type :multi-threaded]
+  [^String queue-name buffer-size timeout :producer-type :multi-threaded :batch-size 100 :batch-timeout 1]
```
harshach: can we make these default batch-size values come from config

revans2:
I just rebased on the latest code in master, so not the only diffs you see should be what this patch is doing.  Before it was based off of other code that was really a part of other diffs.

Please take a look again.

kishorvpatil:
+1

revans2:
@kishorvpatil I just rebased, please take another quick look.

PR #832, opened 2015-10-29
STORM-1127:  allow for boolean arguments (Flux)

ashnazg:


revans2:
To me the code looks good.  +1, but I would like someone more familiar with Flux to also take a look.

@ptgoetz can you take a look, or point out someone who would be good to review this?

ptgoetz:
+1. Looks good to me.

PR #833, opened 2015-10-29
STORM-1126:  allow for configMethods with no args (Flux)

ashnazg:


revans2:
This looks good to me I am +1, but I would like to have someone more familiar with Flux take a look too.

@ptgoetz can you take a look or point out someone who can?

ashnazg:
Do we think this one can get in for 0.11?

ptgoetz:
+1

@ashnazg Yes, I will add to the list.

ashnazg:
Is there a trick to mailing dev@storm.apache.org?  I've only gotten one out of six emails to show up there.  I'm sending them as plaintext, but I'm not a subscriber.  Is it moderated, and perhaps infrequently checked?

PR #835, opened 2015-10-30
[STORM-40] Turn worker garbage collection and heapdump on by default

zhuoliu:
Since STORM-901 worker artifacts directory just gets merged in, it provides a private place for us to turn on the garbage collection and heapdump for each worker by default. 
Since user can easily view and download gc log and heapdump file from logviewer UI, this will greatly improve the debugging of storm. 

Welcome discussion on the setting details (default gc rollover size, number of files, whether we want pid in the gc log file name, etc.),

### DIFF
--- a/conf/defaults.yaml
+++ b/conf/defaults.yaml
@@ -142,7 +142,7 @@ supervisor.cpu.capacity: 400.0
 
 ### worker.* configs are for task workers
 worker.heap.memory.mb: 768
-worker.childopts: "-Xmx%HEAP-MEM%m"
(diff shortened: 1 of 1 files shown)


harshach:
@zhuoliu heapdump will be helpful but I am not sure about the gc.log as it can produce lot of logging.

zhuoliu:
Thanks @harshach . This configuration is similar to what we have in Yahoo cluster. 
Different types of workers may diff, mostly generating less than 100B per second. 24 workers per node might generate 2KB/s gc log data.
In terms of total size, it can be rolled over to be limited as 10 X 1MB files for each worker, which will not take too much disk space.

revans2:
@harshach the reason we have the gc logs on by default is because we have lots of users who ask why was my worker shot?  Most of the time it is due to heavy GC and we wanted to have a smoking gun we could quickly look at to show them.  The bigger problem for us is workers that log way too much.

I am +1 on this change, but if @harshach has concerns we can hold off on it.

harshach:
Thanks for the details @zhuoliu  I am +1 as well.

PR #841, opened 2015-11-02
STORM-1154: SequenceFileBolt needs unit tests

dossett:
These are just a start, I am not terribly knowledgeable of SequenceFiles

### DIFF
--- a/external/storm-hdfs/src/test/java/org/apache/storm/hdfs/bolt/TestSequenceFileBolt.java
+++ b/external/storm-hdfs/src/test/java/org/apache/storm/hdfs/bolt/TestSequenceFileBolt.java
@@ -1,0 +1,186 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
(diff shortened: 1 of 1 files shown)


revans2:
The code looks good and is better then what we have now, which is nothing.  I am +1 on merging this in.

harshach:
+1

external/storm-hdfs/src/test/java/org/apache/storm/hdfs/bolt/TestSequenceFileBolt.java
```diff
@@ -0,0 +1,169 @@
+package org.apache.storm.hdfs.bolt;
```
HeartSaVioR: Please add Apache License header.

HeartSaVioR:
Except minor header issue then it's great! +1.

dossett:
Thanks, everyone! @HeartSaVioR I added the Apache License Header.

Would adding something like the Apache Rat plugin to the build process be helpful to audit issues like this?  I saw at least one other file I contributed recently is also missing the Apache License.  I would be happy to create a JIRA to use Rat.

revans2:
@dossett Yes please do look into it.  I know we use it for releases so I think it is already configured in the pom.xml, but if we could make it is a part of the default build that would catch problems like this sooner, rather then later.

dossett:
@revans2 Opened STORM-1161 and submitted a PR to do just that.

revans2:
Still +1

PR #842, opened 2015-11-02
[STORM-1157] Adding dynamic profiling for worker, restarting worker, jstack, heap dump, and profiling

kishorvpatil:
- Request restarting worker.
- Request jstack, 
- Request heap dump, 
- Start Profiling
   - Dump profiling
   - Stop Profiling


revans2:
It looks good to me.  I am a little bit concerned that there is no way to turn this off, partially or all the way.  So for windows where there is currently no support for this we will end up trying to run a shell script that will fail, and if someone does not have/want commercial java support from Oracle, i.e. using open JDK, or IBM JDK, etc.  They have to make code modifications, not just config modifications to not show users buttons that will not work.

Could we put in something that could let users turn off part or all of this feature?  At a minimum in the UI, but it would be good if nimbus could also let users know that this particular feature is not currently configured on, and if the worker could see the config and know to add in ```-XX:+UnlockCommercialFeatures -XX:+FlightRecorder``` on the command line.

Other then that I am +1.

kishorvpatil:
@revans2 Thank you for the feedback. Following changes should help make this feature completely optional and pluggable.   
   - Added `worker.profiler.enabled` to shutdown use of profilers - 
     - If disabled, UI simply removes the section **Profiling and Debugging** section. 
     - If disabled, the supervisor would not add "worker.profiler.childopts" to worker JVM options - hence avoiding issues with jvm that does not provide commercial JProfiler. 
   - Added `worker.profiler.command` to specify which script to use at supervisor level so that it can be customized for different profiler scheme.

kishorvpatil:
Currently, this feature lacks support for windows and command-line interface.

kishorvpatil:
dynamic_profiling_debugging_1
dynamic_profiling_debugging_2
dynamic_profiling_debugging_3

revans2:
+1 it looks like a great addition.

docs/DYNAMIC_WORKER_PROFILING.md
```diff
@@ -0,0 +1,29 @@
+Dynamic Worker Profiling
```
zhuoliu: Should we move this doc file to docs/documentation directory?
Also the DYNAMIC_LOG_LEVEL_SETTINGS.md file.

kishorvpatil:
@zhuoliu I moved docs under documentation location. Also, added pointers from main page.

storm-core/src/clj/backtype/storm/ui/core.clj
```diff
@@ -986,6 +1021,127 @@
         (log-message "Setting topology " id " log config " new-log-config)
         (.setLogConfig nimbus id new-log-config)
         (json-response (log-config id) (m "callback")))))
+
+  (GET "/api/v1/topology/:id/profiling/start/:host-port/:timeout"
```
zhuoliu: We may need to update the REST API doc for this.

storm-core/src/clj/backtype/storm/ui/core.clj
```diff
@@ -986,6 +1021,127 @@
...
+           (json-response {"status" "ok"
+                           "id" host-port
+                           "timeout" timeout
+                           "dumplink" (worker-dump-link
+                                       host
+                                       port
+                                       id)}
+                          (m "callback")))))
+
+  (GET "/api/v1/topology/:id/profiling/stop/:host-port"
```
zhuoliu: Same here and below.

zhuoliu:
Just some nits about docs. Otherwise looks really good to me.

kishorvpatil:
@zhuoliu I have updated the REST API documentation for all new endpoints.

zhuoliu:
+1

STORM-UI-REST-API.md
```diff
@@ -636,6 +636,201 @@ Sample response:
...
 
+## Profiling and Debugging GET Operations
+
+###  /api/v1/topology/:id/profiling/start/:host-port/:timeout (GET)
+
+Request to start profiler on worker with timeout. Returns status and link to profiler artifacts for worker.
+
+|Parameter |Value   |Description  |
+|----------|--------|-------------|
+|id   	   |String (required)| Topology Id  |
```
revans2: Do we want an `:id` here to be consistent with the other Parameters?

STORM-UI-REST-API.md
```diff
@@ -636,6 +636,201 @@ Sample response:
...
+}
+```
+
+###  /api/v1/topology/:id/profiling/dumpprofile/:host-port (GET)
+
+Request to dump profiler recording on worker. Returns status and worker id for the request.
+
+|Parameter |Value   |Description  |
+|----------|--------|-------------|
+|id   	   |String (required)| Topology Id  |
```
revans2: Here too.

STORM-UI-REST-API.md
```diff
@@ -636,6 +636,201 @@ Sample response:
...
+}
+```
+
+###  /api/v1/topology/:id/profiling/stop/:host-port (GET)
+
+Request to stop profiler on worker. Returns status and worker id for the request.
+
+|Parameter |Value   |Description  |
+|----------|--------|-------------|
+|id   	   |String (required)| Topology Id  |
```
revans2: And here

revans2:
The changes look good and I am +1 on merging this in.  I had one minor comment on the documentation, but it is not critical and I am happy to have this merged in, with or without the change.

kishorvpatil:
@revans2  Fixed REST API documentation for consistency. Moved it to right location and added necessary link for Documentation.md

revans2:
Looks good +1

storm-core/src/clj/backtype/storm/ui/core.clj
hustfxj: I am concern about Thread concurrency. If user is setting the worker's ProfileAction, supervisor is reading the worker's ProfileAction. Then what happened？

kishorvpatil: As rest of the communication between nimbus and supervisor for assignments - going through ZK answers all Thread concurrency issues. As user ProfileAction is written to ZK by nimbus ( which performs 2 phase commit). The supervisor can read old value or new depending on whether it reads is before or after the commit phase.. The supervisor reads it every 30 seconds.

PR #849, opened 2015-11-03
STORM-1155: Supervisor recurring health checks

tgravescs:
https://issues.apache.org/jira/browse/STORM-1155


revans2:
+1 looks good.  It might be nice to mention in the documentation that if you are running the supervisor under supervision that calling `storm health-check` will let you know if you should run the supervisor or not.

zhuoliu:
+1

tgravescs:
Added note about being under supervision.

redsanket:
+1

longdafeng:
the code is fine to me, but this is the first step to do health check:
(1) where put the script, we had better put the script under $STORM_HOME/healthcheck, because when install storm, the script will be also installed in every node.
(2) the script depends on OS, in Windows, they should be xxx, in Linux, they should be XXX;
maybe we need some OS detect action to judge what kind of OS is running. so in the "STORM_HEALTH_CHECK_DIR" dir, there are several directory,  windows/linux/mac, different OS use different dir.
(3) could you please help to implement healthcheck.clj with java. in a short while, the storm will contain two core, one is clojure, the other is java. so could you please help to implement healthcheck.clj.

tgravescs:
I'd way we can convert this to java later when we do the merge.  
1) the directory is already configurable I don't follow what you are asking
2) I don't have windows box on here to test but since its all configuration and user writes the scripts I think it will work on windows already. If not perhaps we can file follow on for folks who have access to test windows to fix.

revans2:
@longdafeng it might be nice in the future to have some "standard" health check scripts come with storm, but honestly even with just Linux, it is difficult to come up with something that will work on all Linux distros, or even on all boxes with the same distro.

The things I see as potential issues would be the network card fell back to 100Mbit.  But that does not work if you are running on old 100Mbit hardware, or what if for some reason you are using infiniband, with a 100Mbit control path.  Or if you are using bonded NICs?  There are too many variables in my opinion to really make a lot of these generic.

The default directory is `$STORM_HOME/healthcheck` already.  Directories are usually relative to $STORM_HOME, but there have been a few cases where we needed to make it explicit.  @zhuoliu you did a lot of work around resolving relative paths in configs.  could you take a look at this and see if we need to change anything to make this work so it is always relative to `$STORM_HOME`/?

zhuoliu:
Hi @revans2 and @tgravescs ,

For checking all cases of relative/absolute directory for healthcheck, I would suggest to add the following function in config.clj, then healthcheck.clj can all this function to get the directory.

(defn absolute-healthcheck-dir [conf]
  (let [storm-home (System/getProperty "storm.home")
        path (conf STORM-HEALTH-CHECK-DIR)]
    (if path
      (if (is-absolute-path? path) path (str storm-home file-path-separator path))
      (str storm-home file-path-separator “healthcheck”))))

tgravescs:
added the absolute path check and upmerged

revans2:
+1

zhuoliu:
+1

tgravescs:
travis failed with disrupterQueueTest failures which is unrelated.

revans2:
@longdafeng do you have any more comments on this pull request?  I would like to merge this in soon if you are OK with it.

longdafeng:
+1 
sorry for late response

storm-core/src/clj/backtype/storm/command/healthcheck.clj
```diff
@@ -0,0 +1,88 @@
...
+    :success))
+
+(defn process-script [conf script]
+  (let [script-proc (. (Runtime/getRuntime) (exec script))
+        curthread (Thread/currentThread)
+        interrupter-thread (interrupter curthread
+                                        (conf STORM-HEALTH-CHECK-TIMEOUT-MS))]
+    (try
+      (.waitFor script-proc)
+      (.interrupt interrupter-thread)
```
hustfxj: @revans2  If script-proc is blocked，then throw InterruptedException and println "Script" script "timed out.".But the script-proc isn't really stop.Like that:
admin    12755     1  0 12:49 pts/0    00:00:00 /bin/sh /home/admin/test/healthCheck.sh
admin    12978     1  0 12:50 pts/0    00:00:00 /bin/sh /home/admin/test/healthCheck.sh
admin    13228     1  0 12:51 pts/0    00:00:00 /bin/sh /home/admin/test/healthCheck.sh
admin    13504     1  0 12:52 pts/0    00:00:00 /bin/sh /home/admin/test/healthCheck.sh
admin    13644 13465  0 12:52 pts/0    00:00:00 /bin/sh /home/admin/test/healthCheck.sh

Maybe we can stop the process ?

(defn interrupter
+  "Interrupt a given thread after ms milliseconds."
+  [script-proc ms]
+  (let [interrupter (Thread.
+                     (fn []
+                       (try
+                         (Thread/sleep ms)
+                         (.destory script-proc)
+                         (catch InterruptedException e))))]
+    (.start interrupter)
+    interrupter))

PR #854, opened 2015-11-04
STORM-1129: Use topology name instead of id in UI calls.

priyank5485:
Note that all-topologies-summary has been used to get topology id from a topology name. That involves a few calls to zookeeper which is not ideal. However, UI does not seem to take any significant performance hit. If needed we can handle it possibly using one of the options below. Since its a separate performance issue we can handle it in a separate JIRA.
1. Have nimbus thrift server cache summary for topologies so it does not hit zookeeper every time we try to get topology id from name.
2. Update nimbus thrift api with a method that takes options and use that to do only the minimal necessary interaction with zookeeper for a given option.

### DIFF
--- a/storm-core/src/ui/public/component.html
+++ b/storm-core/src/ui/public/component.html
@@ -132,12 +132,13 @@ function workerActionSelectedClicked() {
 workerActionSelected = {};
 
 $(document).ready(function() {
+    var topologyId;
     var componentId = $.url("?id");
-    var topologyId = $.url("?topology_id");
-    var tableStateKey = ":".concat(topologyId, ":", componentId);
+    var topologyName = $.url("?topology_name");
+    var tableStateKey = ":".concat(topologyName, ":", componentId);
     var window = $.url("?window");
     var sys = $.cookies.get("sys") || "false";
-    var url = "/api/v1/topology/"+topologyId+"/component/"+componentId+"?sys="+sys;
+    var url = "/api/v1/topology/"+topologyName+"/component/"+componentId+"?sys="+sys;
     if(window) url += "&window="+window;
 
     $.extend( $.fn.dataTable.defaults, {
@@ -186,6 +187,7 @@ $(document).ready(function() {
     }
 
     $.getJSON(url,function(response,status,jqXHR) {
+        topologyId = response["topologyId"];
         var uiUser = $("#ui-user");
         $.get("/templates/user-template.html", function(template) {
             uiUser.append(Mustache.render($(template).filter("#user-template").html(),response));
@@ -194,7 +196,7 @@ $(document).ready(function() {
 
         var componentSummary = $("#component-summary");
         var componentActions = $("#component-actions");
-        var buttonJsonData = componentActionJson(response["encodedTopologyId"], response["encodedId"], response["id"],
+        var buttonJsonData = componentActionJson(response["name"], response["encodedId"], response["id"],
                                                  response["topologyStatus"], response["debug"], response["samplingPct"]);
         var componentStatsDetail = $("#component-stats-detail")
         var inputStats = $("#component-input-stats");
@@ -328,7 +330,7 @@ $(document).ready(function() {
 });
 
 function start_profiling() {
-    var topologyId = $.url("?topology_id");
+    var topologyName = $.url("?topology_name");
     var timeout = $("#timeout").val();
 
     if(timeout == "") { timeout = 10; }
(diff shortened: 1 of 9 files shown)


jerrypeng:
Topology names are not unique.  Users can submit topologies with the same name, but the their auto generated topology ids will be unique.  I am not sure this is the right way to go about this

revans2:
@jerrypeng they can submit topologies with the same name, but not at the same time.  You can only have one "foo" topology running at any point in time.

priyank5485:
@jerrypeng We have a check and it throws the following exception.

Exception in thread "main" java.lang.RuntimeException: Topology with name `wordcount` already exists on cluster
	at backtype.storm.StormSubmitter.submitTopologyAs(StormSubmitter.java:231)
	at backtype.storm.StormSubmitter.submitTopology(StormSubmitter.java:275)
	at backtype.storm.StormSubmitter.submitTopologyWithProgressBar(StormSubmitter.java:311)
	at backtype.storm.StormSubmitter.submitTopologyWithProgressBar(StormSubmitter.java:292)
	at storm.starter.WordCountTopology.main(WordCountTopology.java:94)

@revans2 Thanks for chiming in.

harshach:
@jerrypeng topology names are unique

Parth-Brahmbhatt:
+1.

jerrypeng:
thanks for the explanation and clarification everybody

revans2:
This is a non-backwards compatible change to the REST API and I really would prefer to maintain compatibility if at all possible.  It seems fairly simple to make the REST API work with the name or the id.  We call get-id-from-name all over the place, it seems fairly simple to check to see if there is a topology with that as the ID, if not check to see if there is a topology with it as the name.

storm-core/src/clj/backtype/storm/ui/core.clj
```diff
@@ -851,141 +859,151 @@
...
-        (.activate nimbus name)
-        (log-message "Activating topology '" name "'")))
-    (json-response (topology-op-response id "activate") (m "callback")))
-  (POST "/api/v1/topology/:id/deactivate" [:as {:keys [cookies servlet-request]} id & m]
-    (populate-context! servlet-request)
-    (assert-authorized-user "deactivate" (topology-config id))
-    (thrift/with-configured-nimbus-connection nimbus
+  (GET "/api/v1/topology/:name" [:as {:keys [cookies servlet-request scheme]} name & m]
+    (let [id (get-id-from-name name)]
+      (populate-context! servlet-request)
```
revans2: We have to populate the context before ever talking to nimbus.  `get-id-from-name` talks to nimbus and could potentially authenticate with the wrong user to nimbus.  This needs to be fixed everywhere.

storm-core/src/ui/public/component.html
```diff
@@ -116,7 +116,7 @@
 
         var componentSummary = $("#component-summary");
         var componentActions = $("#component-actions");
-        var buttonJsonData = componentActionJson(response["encodedTopologyId"], response["encodedId"], response["id"],
+        var buttonJsonData = componentActionJson(response["name"], response["encodedId"], response["id"],
```
revans2: We need to encode the name just like we conded the topologyId, or we have the possibility of an incorrect URL for odd characters in a topology name.

revans2:
@priyank5485 for the most part this change looks good to me, but there are a few serious bugs/incompatibilities that need to be fixed before this can go in.

Parth-Brahmbhatt:
I agree with @revans2 , lets maintain backward compatibility.

priyank5485:
@revans2 @Parth-Brahmbhatt @jerrypeng @harshach  Thanks all for the feedback.  Other than the changes here there are some commits to apache/master after creation of this PR which added some more apis in UI using topology id making this PR unmergeable. I will  sort all of that out and update the PR.

revans2:
@priyank5485 thanks for doing this.  I think it is going to make the UI much more usable.  Essentially giving us perma-links to a topology.

wuchong:
+1 

In Alibaba, we implement our monitor system using topology name instead of id. And it is very useful , as users often resubmit topology several times , and they do not need to find the new topology page link.

But we should  maintain backward compatibility. maybe we can deal with  topology-id arg in function  `get-id-from-name`

harshach:
@priyank5485 sorry for the delay. can you please upmerge this patch and open up another PR for 1.x-branch as well.

priyank5485:
@harshach @revans2 I have addressed the comments  and raised a PR https://github.com/apache/storm/pull/1277 against the 1.x branch as I could not test my branch off of master due to another issue. Can you please review the PR against 1.x branch? I will port the same to master by updating this PR.

PR #861, opened 2015-11-05
[STORM-1177] Build fails because DISCLAIMER file is missing

schonfeld:
apache/storm@4a57e500b3e0c3f3256e776357d1e1de1c7f5e49 removed the DISCLAIMER file, which causes `mvn package` to fail (as that file is required in [binary.xml#L320](https://github.com/apache/storm/blob/master/storm-dist/binary/src/main/assembly/binary.xml#L320)).

### DIFF
--- a/storm-dist/binary/src/main/assembly/binary.xml
+++ b/storm-dist/binary/src/main/assembly/binary.xml
@@ -316,10 +316,6 @@
             <source>${project.basedir}/../../README.markdown</source>
             <outputDirectory>/</outputDirectory>
         </file>
-        <file>
-            <source>${project.basedir}/../../DISCLAIMER</source>
-            <outputDirectory>/</outputDirectory>
-        </file>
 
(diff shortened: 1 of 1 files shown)


revans2:
@schonfeld it looks like this was already done on a different pull request.

```
commit d59eaac77ed65b85f6aa2d8ab56ca99b1075d468
Author: Sriharsha Chintalapani <harsha at hortonworks dot com>
Date:   Thu Nov 5 16:45:15 2015 -0800

    Removed DISCLAIMER from storm-dist.
```

If it looks OK to you could you close this pull request? 

@harshach did you do this work as part of a JIRA?  If so could you duple STORM-1177 to that JIRA?

schonfeld:
@revans2 you are correct!

harshach:
@revans2 checked into master without a jira since it was a small fix.

PR #859, opened 2015-11-04
[STORM-1170] Fix the producer alive issue in DisruptorQueueTest

zhuoliu:

PR #864, opened 2015-11-05
STORM-1183:  bump timeout to 10s, since test case isn't really testing speed

ashnazg:


revans2:
@ashnazg I think we fixed the failure in https://github.com/apache/storm/pull/859  Could you please check and see if it worked for you?

ashnazg:
@revans2 yep, that test is consistently passing in the builds I'm watching.

PR #865, opened 2015-11-06
[STORM-1185] replace nimbus.host with nimbus.seeds

vesense:


revans2:
+1

zhuoliu:
+1

conf/storm.yaml.example
```diff
@@ -19,7 +19,7 @@
 #     - "server1"
 #     - "server2"
 # 
-# nimbus.host: "nimbus"
+# nimbus.seeds: ["localhost"]
```
jerrypeng: since for Storm HA there can be multiple nimbus hosts, shouldn't we provide an example like with storm.zookeeper.servers that includes multiple hosts? Like 

nimbus.seeds: ["host1", "host2", "host3"]

vesense: Good catch. Updated.

Parth-Brahmbhatt: Can you instead use values that will be valid in a single node setup? With current values anyone running storm locally will be forced to update this value.

nimbus.seeds:["localhost","127.0.0.1"]

hustfxj:
+1

jerrypeng:
+1

HeartSaVioR:
+1

PR #866, opened 2015-11-06
STORM-1161: Adding ratprint.py

dossett:
@revans2 Apologies, but I just realized that my squash for this JIRA dropped one new file.  This PR adds it back.


revans2:
+1 oops my bad.  I ran the tests but not travis.

PR #869, opened 2015-11-09
STORM-1191:  bump timeout by 50% due to intermittent travis build failures

ashnazg:


kishorvpatil:
+1

HeartSaVioR:
+1

Btw, it could help, but it may not resolve origin issue since 100s should be long enough.

ashnazg:
True, but if the original issue is just that Travis is slow, ;-) ... I just want to minimize the false negatives of test failures that are caused when Travis is bogged down and the test case runs too slow.  I presume that the test will _logically_ pass if given enough time, and that the test case itself is not trying to enforce a performance speed spec.

revans2:
@ashnazg the test "might" "logically" pass.  The code is a bit complex.  When debugging these in the past the tests use simulated time, which is a great thing it lets us run tests that would take a long time very quickly, but to make it work properly there are some critical points where we wait for the simulated cluster to be idle.  Typically we see timeouts happen here, waiting for the entire cluster to stop.  But for some reason on really slow hardware, that I don't totally understand yet, this does not always happen. and increasing the timeout does not always induce it to be in that state.

I am +1 on merging this in, but I am not totally sure that it will "fix" the problem.

knusbaum:
Most all of the spurious test failures I've seen have been *real* failures and not timeouts.
I'm okay with this going in, even though I don't like potentially making the build time longer and I'm not optimistic about this fixing any problems.

PR #870, opened 2015-11-09
STORM-1190: System Load too high after recent changes

revans2:
This helps a fair amount at the low end, but not as much as I had hoped.  The spout seems to sleeping more with batching enabled which is offsetting some of the gains from updating the DisruptorQueue code.  I really have not idea why at this point.

### DIFF
--- a/storm-core/src/jvm/backtype/storm/utils/DisruptorQueue.java
+++ b/storm-core/src/jvm/backtype/storm/utils/DisruptorQueue.java
@@ -65,7 +65,7 @@ public class DisruptorQueue implements IStatefulObject {
 
     private static class FlusherPool { 
         private Timer _timer = new Timer("disruptor-flush-trigger", true);
-        private ThreadPoolExecutor _exec = new ThreadPoolExecutor(1, 100, 10, TimeUnit.SECONDS, new ArrayBlockingQueue<Runnable>(1024));
+        private ThreadPoolExecutor _exec = new ThreadPoolExecutor(1, 100, 10, TimeUnit.SECONDS, new ArrayBlockingQueue<Runnable>(1024), new ThreadPoolExecutor.DiscardPolicy());
         private HashMap<Long, ArrayList<Flusher>> _pendingFlush = new HashMap<>();
         private HashMap<Long, TimerTask> _tt = new HashMap<>();
 
@@ -92,8 +92,8 @@ public class DisruptorQueue implements IStatefulObject {
                 if (tasks != null) {
                     _exec.invokeAll(tasks);
                 }
-            } catch (Exception e) {
-               LOG.error("Could not invoke all ", e); 
+            } catch (InterruptedException e) {
+               //Ignored
             }
         }
 
--- a/storm-core/src/jvm/backtype/storm/utils/DisruptorQueue.java
+++ b/storm-core/src/jvm/backtype/storm/utils/DisruptorQueue.java
@@ -34,11 +34,15 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Timer;
+import java.util.TimerTask;
+import java.util.concurrent.ArrayBlockingQueue;
+import java.util.concurrent.Callable;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentLinkedQueue;
-import java.util.concurrent.ScheduledFuture;
-import java.util.concurrent.ScheduledThreadPoolExecutor;
+import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.concurrent.locks.ReentrantLock;
@@ -57,7 +61,51 @@ public class DisruptorQueue implements IStatefulObject {
     private static final Logger LOG = LoggerFactory.getLogger(DisruptorQueue.class);    
     private static final Object INTERRUPT = new Object();
     private static final String PREFIX = "disruptor-";
-    private static final ScheduledThreadPoolExecutor TIMER = new ScheduledThreadPoolExecutor(0);
+    private static final FlusherPool FLUSHER = new FlusherPool();
+
+    private static class FlusherPool { 
+        private Timer _timer = new Timer("disruptor-flush-trigger", true);
+        private ThreadPoolExecutor _exec = new ThreadPoolExecutor(1, 100, 10, TimeUnit.SECONDS, new ArrayBlockingQueue<Runnable>(1024));
+        private HashMap<Long, ArrayList<Flusher>> _pendingFlush = new HashMap<>();
+        private HashMap<Long, TimerTask> _tt = new HashMap<>();
+
+        public synchronized void start(Flusher flusher, final long flushInterval) {
+            ArrayList<Flusher> pending = _pendingFlush.get(flushInterval);
+            if (pending == null) {
+                pending = new ArrayList<>();
+                TimerTask t = new TimerTask() {
+                    @Override
+                    public void run() {
+                        invokeAll(flushInterval);
+                    }
+                };
+                _pendingFlush.put(flushInterval, pending);
+                _timer.schedule(t, flushInterval, flushInterval);
+                _tt.put(flushInterval, t);
+            }
+            pending.add(flusher);
+        }
+
+        private synchronized void invokeAll(long flushInterval) {
+            try {
+                ArrayList<Flusher> tasks = _pendingFlush.get(flushInterval);
+                if (tasks != null) {
+                    _exec.invokeAll(tasks);
+                }
+            } catch (Exception e) {
+               LOG.error("Could not invoke all ", e); 
+            }
+        }
+
+        public synchronized void stop(Flusher flusher, long flushInterval) {
+            ArrayList<Flusher> pending = _pendingFlush.get(flushInterval);
+            pending.remove(flusher);
+            if (pending.size() == 0) {
+                _pendingFlush.remove(flushInterval);
+                _tt.remove(flushInterval).cancel();
+            }
+        }
+    }
 
     private static class ObjectEventFactory implements EventFactory<AtomicReference<Object>> {
         @Override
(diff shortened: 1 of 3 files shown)


danielschonfeld:
@revans2 what if we built a 'clock service' for each worker.  A single dedicated timer thread to which all bolts/spouts would register 'interest' and would tick or fire a callback on them every X milliseconds?  1 clock thread per worker to serve all the different DisruptorQueue tick needs.

What do you think of that?

revans2:
@danielschonfeld Yes we needs something like that.  We are using a single timer thread for all of the system metrics to be able to do something similar, but there is no possibility of them blocking.  It is all atomic memory operations when the timer goes off so having a single thread shared between them is not a big deal.

With the disruptor queue if the queue is full the thread will block trying to flush the messages.  This could be a very long period of time.  This is why I went with the ScheduledThreadPoolExecutor.  The problem here is that it has no way to spin up new threads under load or tear them down when idle.  I could spend some time and build my own, but this seems to solve a lot, but not all of the load problem.

I am going to try and spend some time to create something that can do all of that, but it is not something I can put together in a few hours and expect it to work.

danielschonfeld:
@revans2 can you explain the load problem to me again please? is the problem here the fact that we have so many threads in a sleeping state? meaning that cutting the amount of those will result in a total performance gain? or is there another problem at play here?

Also if I wrote code, how can I test it?

revans2:
@danielschonfeld I don't know 100% what the cause of the issue is, but I suspect that it is having lots of threads trying to sleep very frequently.

```storm jar ./examples/storm-starter/storm-starter-topologies-0.11.0-SNAPSHOT.jar storm.starter.ThroughputVsLatency 100 1 5```

is the test that I have been using.  It is really just word count, but with some latency and system utilizations statistics added in.  The problem is that with batching the CPU utilization under low load is much higher then before the batching patch went in. 24886eec5c45e7fd30cac804fd080360f17599a0 before batching, a0f3412a8268e75f87a95421a53f6bc4b6af9842 with batching, and then the current patch are tests I ran. all of the measurements are in ms measured over a 30 second period.

| version | user | sys |
|---|---|---|
| no-batch | 4,320 | 4,789 |
| batch | 7,565 | 17,483 |
| this | 8,101 | 10,238 |

So the patch shifted some load from the kernel to user space, and dropped the overall CPU utilization, but it is still much higher than before.

**Comment by revans2** [comment]
I did some micro-benchmark experiments.

```
public class Test extends Thread {
  final long _expectedEnd;
  final long _sleepTime;

  public Test(long ee, long st) {
    _expectedEnd = ee;
    _sleepTime = st;
  }

  public void run() {
    try {
      while (System.currentTimeMillis() < _expectedEnd) {
        Thread.sleep(_sleepTime);
      }
    } catch (Exception e) {
    }
  }

  public static void main(String [] args) throws Exception {
    long sleepTime = 1;
    if (args.length > 0) {
      sleepTime = Long.valueOf(args[0]);
    }
    long totalTimeSec = 100;
    if (args.length > 1) {
      totalTimeSec = Long.valueOf(args[1]);
    }
    int totalThreads = 10;
    if (args.length > 2) {
      totalThreads = Integer.valueOf(args[2]);
    }

    long totalTimeMs = totalTimeSec * 1000;
    long expectedEnd = System.currentTimeMillis() + totalTimeMs;
    int ret = -1;
    try {
      Test [] tests = new Test[totalThreads];
      for (int i = 0; i < totalThreads; i++) {
        tests[i] = new Test(expectedEnd, sleepTime);
        tests[i].start();
      }
      for (int i = 0; i < totalThreads; i++) {
        tests[i].join();
      }
      ret = 0;
    } finally {
      System.exit(ret);
    }
  }
}
```
can make my laptop become unresponsive at 300 threads and a 1 ms sleep time.

```
import java.util.concurrent.*;
import java.util.concurrent.atomic.*;
import java.util.Random;

public class Test extends Thread {
  public static final ScheduledThreadPoolExecutor _timer = new ScheduledThreadPoolExecutor(0);
  public static final AtomicLong _error = new AtomicLong(0);
  private long _expected;
  private final long _sleepTime;
  private final Random _rand = new Random();

  public Test(long sleepTime) {
    _sleepTime = sleepTime;
    _expected = System.currentTimeMillis() + sleepTime;
  }

  public void run() {
    long now = System.currentTimeMillis();
    long err = Math.abs(now - _expected);
    _error.addAndGet(err);
    if (_rand.nextInt(100) >= 99) {
      try {
        Thread.sleep(10);
      } catch (Exception e) {
        //ignored
      }
    }
    _expected = System.currentTimeMillis() + _sleepTime;
  }

  public static void main(String [] args) throws Exception {
    long sleepTime = 1;
    if (args.length > 0) {
      sleepTime = Long.valueOf(args[0]);
    }
    long totalTimeSec = 100;
    if (args.length > 1) {
      totalTimeSec = Long.valueOf(args[1]);
    }
    int totalThreads = 10;
    if (args.length > 2) {
      totalThreads = Integer.valueOf(args[2]);
    }

    long totalTimeMs = totalTimeSec * 1000;
    int ret = -1;
    try {
      ScheduledFuture<?> [] tests = new ScheduledFuture<?>[totalThreads];
      for (int i = 0; i < totalThreads; i++) {
        Test t = new Test(sleepTime);
        //TODO real one will need locking
        _timer.setCorePoolSize(_timer.getCorePoolSize() + 1);
        tests[i] = _timer.scheduleWithFixedDelay(t, sleepTime, sleepTime, TimeUnit.MILLISECONDS);
      }
      Thread.sleep(totalTimeMs);
      for (int i = 0; i < totalThreads; i++) {
        tests[i].cancel(true);
      }
      System.out.println("AVG Error "+(((double)_error.get())/totalThreads/totalTimeSec));
      ret = 0;
    } finally {
      System.exit(ret);
    }
  }
}
```
uses almost no CPU and does more or less the same thing, but with more jitter, and a measurement of how much error there is.  I tried various combinations of thins to improve the performance.  Now I think we need to write a custom SchedulerExecutorService, that can increase and decrease the number of threads based off of how much load there is.  It can purposely run multiple tasks together, even if they would be spread out with a more accurate clock waking them up.

**Comment by revans2** [comment]
Because this patch shows an improvement over the current code, I thought we should merge it in while I play around with a "better" fix.

**Comment by danielschonfeld** [comment]
@revans2 this is what I had in mind.  We call start() on this in worker.clj and we pass this `FlushingService` to `DisruptorQueue` so each queue could register a callable to flush the messages.

It uses a cached thread pool which will increase according to demand or otherwise just re-use threads already created.

```
package backtype.storm.utils;


import com.google.common.collect.Lists;

import java.util.ArrayList;
import java.util.Collection;
import java.util.concurrent.*;

public class FlushingService implements Runnable {
    private static final ScheduledThreadPoolExecutor TIMER = new ScheduledThreadPoolExecutor(0);
    private static final ExecutorService es = Executors.newCachedThreadPool();

    private final Collection<Callable> registeredRunnables = Lists.newArrayList();

    private ScheduledFuture<?> _future;
    private final long _flushInterval;

    public FlushingService(long flushInterval) {
        _flushInterval = flushInterval;
    }

    public void registerCallable(Callable c) {
        registeredRunnables.add(c);
    }
    
    public void run() {
        try {
            es.invokeAll(registeredRunnables);
        } catch (InterruptedException e) {
            //do something?
        }
    }

    public void start() {
        _future = TIMER.scheduleWithFixedDelay(this, _flushInterval, _flushInterval, TimeUnit.MILLISECONDS);
        synchronized(TIMER) {
            TIMER.setCorePoolSize(TIMER.getCorePoolSize() + 1);
        }
    }

    public void close() {
        if (_future != null) {
            _future.cancel(true);
            _future = null;
            synchronized(TIMER) {
                TIMER.setCorePoolSize(TIMER.getCorePoolSize() - 1);
            }
        }
    }
}

```

revans2:
I'll give it a try and see how it works.  I'll probably change the TIMER to be an actual Timer though.

danielschonfeld:
@revans2 sounds good! if you post to this PR we can iterate/brain storm over it together.

revans2:
@danielschonfeld that is the current plan

revans2:
@danielschonfeld I made the changes, I am going to play around with it a bit, but the initial implementation is actually worse.

user: 9,015
sys: 12,982

danielschonfeld:
@revans2 lol what a nice dream I had :) 

can you push your changes so I could take a look too please?

revans2:
I put it at https://github.com/revans2/incubator-storm/tree/STORM-1190-even-better despite the name it does not appear to be so.  I also have some changes on that branch that I was playing with optimizing no batching, but it does not really have much of an impact.

danielschonfeld:
@revans2 i think we might have had different understandings about what I was suggesting or maybe I just didn't understand the problem correctly.

I thought the objective was to reduce the number of threads in total.  In doing so, having only one single thread that goes to sleep constantly why submitting the required flush to a global ExecutorService that will take care of the **all** the flushing needed by all bolts/spouts on a single worker JVM.

In your implementation we still have a thread per DisruptorQueue doing pretty much the same thing as before, but instead of doing the flushes on the same thread, we split the tasks into a multi-threaded approach and try to do it faster that way.  That's not what I had intended with the above example.

I'd love to hear your thoughts on this.

revans2:
No the FlusherPool is a static final member. 

https://github.com/revans2/incubator-storm/blob/STORM-1190-even-better/storm-core/src/jvm/backtype/storm/utils/DisruptorQueue.java#L67

Only one per JVM

danielschonfeld:
@revans2 i take that back, apologies.  I see now you've followed my line of thought exactly.

I wonder what else we can do to tweak this to achieve better performance.

The results you're seeing, are they worse all around or just the CPU still being high?  I'm confused about the numbers you posted before as they seem higher (better?) than the table of comparison you posted earlier in this PR

revans2:
OK Don't sell your idea short just yet.  I did some tuning of the ExecutorService and the numbers are on par with prior to batching.

user:      4,552
sys:      5,015

These are just the CPU ms over a 30 second interval, so smaller is better here.  I will check in the changes so you can have a look.  There is still some code cleanup to do and more testing before I consider it ready for a full pull request but it is looking very promising.

danielschonfeld:
@revans2 Thanks for the kind words! Even more so, great job on tweaking the instantiation of that ExecutorService.

I looked at the code and I believe this is the best I have to offer.  The only other thing I could think about is perhaps adding priority logic to clear batches from spout/bolts that are more backlogged than others in a particular worker.

revans2:
Looks like we have a test that keeps getting stuck.  I am going to do some debugging to see if I can understand what is happening.

revans2:
Turns out that invokeAll waits for all of the processes to finish, and they could block so it was not what I wanted/expected.

revans2:
This time the travis failure is unrelated.  It got past the disruptor tests and failed to bind to a port for a DRPC test.  The code should be good to go now.  Reviews are welcome.

On a side note this dropped the CPU utilization enough that I was now able to do 30,000 sentences/second with ThroughputVsLatency, where as before I was only able to do 27,000.  So it improves the performance as well.

kishorvpatil:
LGTM. +1

PR #872, opened 2015-11-10
STORM-1173. Support string operations in StormSQL.

haohui:
This PR implements the support of string operations in StormSQL.


harshach:
+1

PR #761, opened 2015-09-23
STORM-1067. Run Travis-CI to in containers.

haohui:
This PR switches Travis-CI to run on containers so that the tests can be run faster.

PR #874, opened 2015-11-10
STORM-1197: Migrate Travis-CI to container-based builds.

knusbaum:
http://docs.travis-ci.com/user/migrating-from-legacy/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,6 @@ before_install:
   - nvm use 0.12.2
 install: /bin/bash ./dev-tools/travis/travis-install.sh `pwd` $MODULES
 script: /bin/bash ./dev-tools/travis/travis-script.sh `pwd` $MODULES
-sudo: false
 cache:
   directories:
     - "$HOME/.m2/repository"
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@
 #  limitations under the License.
 
 env:
-  - MODULES=storm-core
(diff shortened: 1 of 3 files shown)


HeartSaVioR:
@knusbaum 
It makes build slower than current now. Please refer #761 for more info.

Btw, if it is more stable than legacy (I mean less random test failures) I'll definitely vote +1.

knusbaum:
@HeartSaVioR 

There. I shaved a little more than 5 minutes off the build time.

HeartSaVioR:
@knusbaum 
Container based build doesn't save build time. 
Seems like splitting build matrix saves build time, so I'd like to see it applied to legacy.

Btw, container based build is triggered faster than legacy, so container based build takes advantage when build time is small enough.

HeartSaVioR:
@knusbaum 
Hey, I see the benefit of splitting build matrix, since it seems faster than current.
If ```legacy + splitting build matrix``` doesn't work, I'd like to vote +1 to ```container based + splitting build matrix```.
I'd like to see reason of closing this PR. Did you see the disadvantage of splitting build matrix?

knusbaum:
@HeartSaVioR 
Yes, I remembered that the external parts I split off the storm-core build are dependent on the storm-core jar, so they must be done after storm-core. There isn't any benefit to doing the external, etc. in parallel with each other either, since they are so small that the startup cost overwhelms the parallelization benefit.

The caching is still valid, but it only shaves a few seconds off.

PR #875, opened 2015-11-10
[STORM-1198] Web UI to show resource usages and Total Resources on al…

zhuoliu:
Calculate, manage and display on UI for
each supervisor's total memory, used memory, total cpu and used CPU (by RAS).

For non-RAS schedulers, the assigned memory for each supervisor and each topology will
be calculated based on the JVM opts settings.

See snapshot on UI:
image

### DIFF
--- a/storm-core/src/jvm/backtype/storm/scheduler/Cluster.java
+++ b/storm-core/src/jvm/backtype/storm/scheduler/Cluster.java
@@ -466,6 +466,75 @@ public class Cluster {
         return networkTopography;
     }
 
+    /*
+    * Get heap memory usage for a worker's main process and logwriter process
+    * */
+    private Double getAssignedMemoryForSlot(Map topConf) {
+        Double totalWorkerMemory = 0.0;
+
+        String topology_worker_childopts = Utils.getString(topConf.get(Config.TOPOLOGY_WORKER_CHILDOPTS), null);
+        String worker_childopts = Utils.getString(topConf.get(Config.WORKER_CHILDOPTS), null);
+        Double mem_topology_worker_childopts = Utils.parseWorkerChildOpts(topology_worker_childopts, null);
+        Double mem_worker_childopts = Utils.parseWorkerChildOpts(worker_childopts, null);
+
+        if (mem_topology_worker_childopts != null) {
+            totalWorkerMemory += mem_topology_worker_childopts;
+        } else if (mem_worker_childopts != null) {
+            totalWorkerMemory += mem_worker_childopts;
+        } else {
+            totalWorkerMemory += Utils.getInt(topConf.get(Config.WORKER_HEAP_MEMORY_MB));
+        }
+
+        String topology_worker_lw_childiopts = Utils.getString(topConf.get(Config.TOPOLOGY_WORKER_LOGWRITER_CHILDOPTS), null);
+        if (topology_worker_lw_childiopts != null) {
+            totalWorkerMemory += Utils.parseWorkerChildOpts(topology_worker_lw_childiopts, 0.0);
+        }
+        return totalWorkerMemory;
+    }
+
+    /*
+    * Update memory usage for each topology and each supervisor node after every round of scheduling
+    * */
+    public void updateAssignedMemoryForTopologyAndSupervisor(Topologies topologies) {
+        Map<String, Double> supervisorToAssignedMem = new HashMap<String, Double>();
+
+        for (Map.Entry<String, SchedulerAssignment> entry : this.getAssignments().entrySet()) {
+            String topId = entry.getValue().getTopologyId();
+            Map topConf = topologies.getById(topId).getConf();
+            Double assignedMemForTopology = 0.0;
+            Double assignedMemPerSlot = getAssignedMemoryForSlot(topConf);
+            for (WorkerSlot ws: entry.getValue().getSlots()) {
+                assignedMemForTopology += assignedMemPerSlot;
+                String nodeId = ws.getNodeId();
+                if (supervisorToAssignedMem.containsKey(nodeId)) {
+                    supervisorToAssignedMem.put(nodeId, supervisorToAssignedMem.get(nodeId) + assignedMemPerSlot);
+                } else {
+                    supervisorToAssignedMem.put(nodeId, assignedMemPerSlot);
+                }
+            }
+            if (this.getResourcesMap().containsKey(topId)) {
+                Double[] topo_resources = getResourcesMap().get(topId);
+                topo_resources[3] = assignedMemForTopology;
+            } else {
+                Double[] topo_resources = {0.0, 0.0, 0.0, 0.0, 0.0, 0.0};
+                topo_resources[3] = assignedMemForTopology;
+                this.setResources(topId, topo_resources);
+            }
+        }
+
+        for (Map.Entry<String, Double> entry : supervisorToAssignedMem.entrySet()) {
+            String nodeId = entry.getKey();
+            if (this.supervisorsResources.containsKey(nodeId)) {
+                Double[] supervisor_resources = supervisorsResources.get(nodeId);
+                supervisor_resources[2] = entry.getValue();
+            } else {
+                Double[] supervisor_resources = {0.0, 0.0, 0.0, 0.0};
+                supervisor_resources[2] = entry.getValue();
+                this.supervisorsResources.put(nodeId, supervisor_resources);
+            }
+        }
+    }
+
     public void setStatus(String topologyId, String status) {
         this.status.put(topologyId, status);
     }
@@ -478,11 +547,15 @@ public class Cluster {
         this.resources.put(topologyId, resources);
     }
 
+    public void setResourcesMap(Map<String, Double[]> topologies_resources) {
+        this.resources.putAll(topologies_resources);
+    }
+
     public Map<String, Double[]> getResourcesMap() {
         return this.resources;
     }
 
-    public void setSupervisorsResources(Map<String, Double[]> supervisors_resources) {
+    public void setSupervisorsResourcesMap(Map<String, Double[]> supervisors_resources) {
         this.supervisorsResources.putAll(supervisors_resources);
     }
 
--- a/storm-core/src/jvm/backtype/storm/scheduler/Cluster.java
+++ b/storm-core/src/jvm/backtype/storm/scheduler/Cluster.java
@@ -507,7 +507,7 @@ public class Cluster {
 
         String topoWorkerLwChildopts = null;
         if (topConf.get(Config.TOPOLOGY_WORKER_LOGWRITER_CHILDOPTS) instanceof List) {
-            topoWorkerLwChildopts = getStringFromStringList(topConf.get(Config.TOPOLOGY_WORKER_LOGWRITER_CHILDOPTS);
+            topoWorkerLwChildopts = getStringFromStringList(topConf.get(Config.TOPOLOGY_WORKER_LOGWRITER_CHILDOPTS));
         } else {
             topoWorkerLwChildopts = Utils.getString(topConf.get(Config.TOPOLOGY_WORKER_LOGWRITER_CHILDOPTS), null);
         }
--- a/storm-core/src/jvm/backtype/storm/scheduler/Cluster.java
+++ b/storm-core/src/jvm/backtype/storm/scheduler/Cluster.java
@@ -466,15 +466,35 @@ public class Cluster {
         return networkTopography;
     }
 
+    private String getStringFromStringList(Object o) {
+        StringBuilder sb = new StringBuilder();
+        for (String s : (List<String>) o) {
+            sb.append(s);
+            sb.append(" ");
+        }
+        return sb.toString();
+    }
+
     /*
     * Get heap memory usage for a worker's main process and logwriter process
     * */
     private Double getAssignedMemoryForSlot(Map topConf) {
         Double totalWorkerMemory = 0.0;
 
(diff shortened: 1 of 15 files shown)


storm-core/src/ui/public/templates/index-page-template.html
```diff
@@ -215,6 +215,26 @@
...
+        </span>
+      </th>
+      <th>
+        <span data-toggle="tooltip" data-placement="left" title="Memory that has been allocated.">
+          Used Mem (MB)
+        </span>
+      </th>
+      <th>
+        <span data-toggle="tooltip" data-placement="above" title="CPU capacity of a supervisor. Every 100 means one core.">
+          Total CPU (%)
```
jerrypeng: Doesn't seem like this should be a percentage

storm-core/src/ui/public/templates/index-page-template.html
```diff
@@ -215,6 +215,26 @@
...
+        </span>
+      </th>
+      <th>
+        <span data-toggle="tooltip" data-placement="left" title="CPU that has been allocated. Every 100 means one core">
+          Used CPU (%)
```
jerrypeng: Same here

storm-core/src/ui/public/templates/index-page-template.html
zhuoliu: Thanks for the comment. This is a percentage to a vCore. So every 1 equals to 1% of a core. E.g., 400 here means 4 CPU vCore. I would appreciate if we can have a more descriptive unit sign here.

jerrypeng: ok this make sense

jerrypeng:
LGTM +1

redsanket:
what would happens to the UI information if it is not using RAS? Just curious if it is turned off then do we want to show this information?

storm-core/src/ui/public/templates/index-page-template.html
```diff
@@ -229,6 +249,10 @@
       <td>{{uptime}}</td>
       <td>{{slotsTotal}}</td>
       <td>{{slotsUsed}}</td>
+      <td>{{totalMem}}</td>
+      <td>{{usedMem}}</td>
+      <td>{{totalCpu}}</td>
+      <td>{{usedCpu}}</td>
```
revans2: Could you update the REST API docs to indicate the new fields

revans2:
I just tested what @redsanket asked and topologies not using the resource aware scheduler show no resources used at all.  It really would be good if we could parse out the heap from the command line of a worker, and do a quick SWAG on the CPU when not using RAS.  That way we could at least see if memory/CPU is being over committed on a node.

storm-core/src/ui/public/templates/index-page-template.html
zhuoliu: Sure, adding it.

For non-RAS scheduler, the totalMem and totalCpu would be default(3000 and 400), and the usedMem and usedCpu will be always 0 and 0, no matter whether there is topology or not.

jerrypeng:
@redsanket good point! the ui should still be compatible if RAS is not used

jerrypeng:
Can we just not show any of the columns that deal with resources if the scheduler used is not RAS

zhuoliu:
@revans2 If we want to show allocated cpu/mem for non-RAS schedulers, we will also need to do that for topology view except for the supervisor view.
Also, it would be also possible that the usedMem might be larger than totalMem (since totalMem may not be correctly configured for non-RAS schedulers, e.g., using default 3GB).

HeartSaVioR:
I'd also like to hide 4 columns for non-RAS scheduler since exposing default value confuses users.
But if we can fill usedCpu / usedMem to actual value it would be much better.

revans2:
I am fine with having a config to hide the 4 columns for the non-RAS case, but please file a follow on JIRA so we can at least show the memory requested for non-RAS schedulers.  We may change a lot of it once the JStorm merger happens and we have numbers for CPU/Memory actually used.  Once RAS is stable we may even migrate the existing schedulers to be strategies for RAS, and have them support not violating resource constraints.

zhuoliu:
Resource columns are hidden for non-RAS schedulers in topology and supervisor views.

For non-RAS schedulers, we can calculate per-topology and per-supervisor total memory assigned by analyzing the topology assignments, MAX-HEAP-MEMORY, worker.childopts and topology.worker.logwriter.childopts and topology.worker.childopts.

storm-core/src/ui/public/index.html
```diff
@@ -143,6 +143,13 @@
             ]
           });
           $('#topology-summary [data-toggle="tooltip"]').tooltip();
+          $.getJSON("/api/v1/cluster/configuration", function(json){
+              var scheduler = json["storm.scheduler"];
+              if (scheduler != "backtype.storm.scheduler.resource.ResourceAwareScheduler"){
+                  $('#topology-summary td:nth-child(9),#topology-summary th:nth-child(9)').hide();
+                  $('#topology-summary td:nth-child(10),#topology-summary th:nth-child(10)').hide();
+              }
+          });
```
revans2: I don't really like having this be configured based off of a hard coded scheduler.  I would prefer to have a separate config, or even better to have the rest API return a null for fields that it knows were not calculated properly.  Then have the template filter out those fields based off of those values.

If we are going to go with a javascript hide, please add a class to all of the fields you want to hide with something like RAS_ONLY and then just do ```$(".RAS_ONLY").hide();```

storm-core/src/ui/public/index.html
```diff
@@ -156,6 +163,15 @@
...
           $('#supervisor-summary [data-toggle="tooltip"]').tooltip();
+          $.getJSON("/api/v1/cluster/configuration", function(json){
+              var scheduler = json["storm.scheduler"];
+              if (scheduler != "backtype.storm.scheduler.resource.ResourceAwareScheduler"){
+                  $('#supervisor-summary td:nth-child(6),#supervisor-summary th:nth-child(6)').hide();
+                  $('#supervisor-summary td:nth-child(7),#supervisor-summary th:nth-child(7)').hide();
+                  $('#supervisor-summary td:nth-child(8),#supervisor-summary th:nth-child(8)').hide();
+                  $('#supervisor-summary td:nth-child(9),#supervisor-summary th:nth-child(9)').hide();
+              }
+          });
```
revans2: Same here as above

revans2:
It also looks like you need to upmerge and regenerate the thrift code with thrift-0.9.3.

storm-core/src/ui/public/index.html
```diff
@@ -143,6 +143,12 @@
             ]
           });
           $('#topology-summary [data-toggle="tooltip"]').tooltip();
+          $.getJSON("/api/v1/cluster/configuration", function(json){
```
kishorvpatil: i suggest we avoid making another ajax call - instead we can have topology summary include boolean as part of topology-summary.

storm-core/src/ui/public/index.html
```diff
@@ -143,6 +143,12 @@
...
+              var displayResource = json["scheduler.display.resource"];
+              if (!displayResource){
+                  $('#topology-summary td:nth-child(10),#topology-summary th:nth-child(10)').hide();
```
kishorvpatil: Let's avoid referring to fields by number which could change in future and make maintenance harder..

kishorvpatil: We could modify the template to decide not list headers and values id `displayResource` is `false`. That is cleaner that populating and hiding certain columns by number.

kishorvpatil:
Two minor suggestions. Otherwise it looks good.

storm-core/src/ui/public/index.html
zhuoliu: I tried to pass "schedulerDisplayResource" to each supervisor-summary and topology-summary in REST and then use this to check whether we want to display in template htmls. However, it seems a little weird to store a cluster-wide configuration as a supervisor-specific or topology-specific field. Also, when there is no topology submitted, it is not feasible to check whether we want to display the assignedCPU in the topology-summary table's header (similarly for supervisor summary table).

We may rather keep it still since later we will have RAS as default scheduler and will merge with JSTORM's UI features.

wuchong: agree with @kishorvpatil , although it's a little odd.

zhuoliu:
Hi @kishorvpatil , your comments have been addressed. Thanks.

zhuoliu:
Add one rest field for topo/id, delete the last REST getJSON call

kishorvpatil:
LGTM. +1. @zhuoliu please rebase the branch.

storm-core/src/jvm/backtype/storm/scheduler/Cluster.java
```diff
@@ -461,6 +466,75 @@ public SupervisorDetails getSupervisorById(String nodeId) {
         return networkTopography;
     }
 
+    /*
+    * Get heap memory usage for a worker's main process and logwriter process
+    * */
+    private Double getAssignedMemoryForSlot(Map topConf) {
+        Double totalWorkerMemory = 0.0;
+
+        String topology_worker_childopts = Utils.getString(topConf.get(Config.TOPOLOGY_WORKER_CHILDOPTS), null);
```
jerrypeng: I think we should use camel case to be consistent with rest of the file

storm-core/src/jvm/backtype/storm/scheduler/Cluster.java
```diff
@@ -461,6 +466,75 @@ public SupervisorDetails getSupervisorById(String nodeId) {
...
+        String worker_childopts = Utils.getString(topConf.get(Config.WORKER_CHILDOPTS), null);
+        Double mem_topology_worker_childopts = Utils.parseWorkerChildOpts(topology_worker_childopts, null);
```
jerrypeng: same here change to camel case

storm-core/src/jvm/backtype/storm/scheduler/Cluster.java
```diff
@@ -461,6 +466,75 @@ public SupervisorDetails getSupervisorById(String nodeId) {
...
+        Double mem_worker_childopts = Utils.parseWorkerChildOpts(worker_childopts, null);
```
jerrypeng: same here change to camel case

jerrypeng:
Just tested the UI, resource numbers look right.  Good Stuff +1

zhuoliu:
Thanks @jerrypeng a lot for the thorough review and test verification! Camel case issue has been addressed.

storm-core/src/jvm/backtype/storm/utils/Utils.java
```diff
@@ -764,5 +778,30 @@ public static long zipFileSize(File myFile) throws IOException{
...
+
+    /**
+     * parses the arguments to extract jvm heap memory size.
+     * @param input
+     * @param defaultValue
+     * @return the value of the JVM heap memory setting in a java command.
+     */
+    public static Double parseWorkerChildOpts(String input, Double defaultValue) {
+        if (input != null) {
+            Pattern optsPattern = Pattern.compile("Xmx[0-9]+m");
```
revans2: This is not good enough.  It is not just an 'm' that is supported.

From http://docs.oracle.com/javase/7/docs/technotes/tools/windows/java.html

```
-Xmxn

    Specifies the maximum size, in bytes, of the memory allocation pool. This value must a multiple of 1024 greater than 2 MB. Append the letter k or K to indicate kilobytes, or m or M to indicate megabytes. The default value is chosen at runtime based on system configuration.
```

I have also seen 'g' and 'G' work for gigabytes too.

storm-core/src/jvm/backtype/storm/scheduler/Cluster.java
```diff
@@ -461,6 +466,75 @@ public SupervisorDetails getSupervisorById(String nodeId) {
...
     }
 
+    /*
+    * Get heap memory usage for a worker's main process and logwriter process
+    * */
+    private Double getAssignedMemoryForSlot(Map topConf) {
+        Double totalWorkerMemory = 0.0;
+
+        String topologyWorkerChildopts = Utils.getString(topConf.get(Config.TOPOLOGY_WORKER_CHILDOPTS), null);
+        String workerChildopts = Utils.getString(topConf.get(Config.WORKER_CHILDOPTS), null);
```
revans2: Both of these configs will return either a string or a list of strings.  We should support both options.

storm-core/src/jvm/backtype/storm/utils/Utils.java
```diff
@@ -764,5 +778,30 @@ public static long zipFileSize(File myFile) throws IOException{
...
         return (Double.isNaN(x) || Double.isInfinite(x)) ? 0.0 : x;
     }
+
+    /**
+     * parses the arguments to extract jvm heap memory size.
+     * @param input
+     * @param defaultValue
+     * @return the value of the JVM heap memory setting in a java command.
+     */
+    public static Double parseWorkerChildOpts(String input, Double defaultValue) {
```
revans2: I would prefer to have the name of this function indicate what it is parsing out of the childopts.

revans2:
Just a few more comments.

storm-core/src/jvm/backtype/storm/utils/Utils.java
```diff
@@ -503,6 +505,24 @@ public static boolean getBoolean(Object o, boolean defaultValue) {
         }
     }
 
+    public static String getString(Object o, String defaultValue) {
```
revans2: Can we inline this or move it someplace else, it does not feel generic enough to be a part of Utils.

revans2:
One more minor comment and then I am +1

revans2:
travis CI failure looks unrelated +1

PR #877, opened 2015-11-12
Added in feature diff 

revans2:
based off of work from Basti Liu


ptgoetz:
+1

I was just going to attach the spreadsheet to JIRA, but this is a much better solution in terms of enabling collaboration.

STORM_JSTORM_FEATURE_DIFF.md
```diff
@@ -0,0 +1,30 @@
...
+Scheduler | Resource aware scheduling	(Work In Progress) | <ol><li>Make the tasks of same component distributed evenly across the nodes in cluster.<li>Make number of tasks in a worker is balanced.<li>Try to assign two tasks which are transferring message directly into a same worker to reduce network cost.<li>Support user defined assignment and  to use the result of last assignment	Different solution between Storm and JStorm.</ol> Actually, JStorm has ever used cpu/mem/disk/network demond for scheduling. But we faced the problem of low utilization of cluster. For example, because the configuration of topologys and resources of hardwares in a cluster are different. After running a period, it is possible that no topology will be assigned to the node with enough other resources but just lacking one resource. | Scheduler interface is pluggable so we should be able to support boht schedulers for the time being |
+Nimbus HA	| Support to configure more than one spare nimbus. When the master nimbus is down, a most appropriate spare nimbus(topologys in disk mostly matches the recoreds in zk) will be chosen to be promoted. | Different solution. | |
```
wuchong: Seems that the Nimbus HA feature of storm and jstorm has been swapped. 
Maybe it should be `Nimbus HA  |	| Support to configure more than one spare nimbus. When the master nimbus is down, a most appropriate spare nimbus(topologys in disk mostly matches the recoreds in zk) will be chosen to be promoted. | Different solution. | `

STORM_JSTORM_FEATURE_DIFF.md
```diff
@@ -0,0 +1,30 @@
...
+metric system | | <ol><li>All levels of metrics, including stream metrics, task metrics, component metrics, topology metrics, even cluster metrics, will be sampled & calculated. Some metrics, e.g. ""tuple life cycle"", are very useful for debug and finding hot spot of a topology.<li>Support full metrics data. previous metric system can only display mean value of meters/histograms, the new metric system can display m1, m5, m15 of meters, and common percentiles of histograms.<li>Use new metrics windows, the min metric window is 1min, thus we can see the metircs data every single minute.<li>Supplies a metric uploader interface, thirty-party companies can easily build their own metric systems based on all historic metric data. | JStorm metric system is completely redesigned.
+jar command | Support progress bar when submitting topology | Does not support progress bar | Low risk for merging to JStorm
+rebalance	command | Basic functionality of rebalance | Besides rebalance, scale-out/in by updating the number of worker, acker, spout & bolt dynamically without stopping topology | How is grouping consistency maintined when changing the number or bolt/spout? |
```
bastiliu: For grouping consistency, the out tasks will be updated accordingly when refreshing connection

revans2: OK we will need some really good documentation explaining how this works, especially with state.  Also if we ever want to go towards a flink or apex like state check-pointing for exactly once processing it will likely not be compatible with this feature, unless we allow everyone to checkpoint state in a lazy key/value like way, which sounds a lot more complex.  This is why I like the task/executor split, because you can set the task level high while keeping the executors low, and if you need to scale up the state does not need to be split/combined.

revans2:
@wuchong and @bastiliu thanks for the feedback and comments.  I updated the pull request based off of them.  Any more feedback is definitely welcome.  By the way great work on jstorm.  My has spent way too much time playing around with http://storm.taobao.org/ being jealous of the features there.

Very excited to see what all of us can accomplish together.

STORM_JSTORM_FEATURE_DIFF.md
bastiliu: Thansk for the explanation. We will try to prepare the documentation, like high level design.
BTW, does community plan to re-design the exactly once processing of Trident? or plan to provide a solution for both normal topology and Trident?

revans2: There are no plans to change exactly once processing right now.  At yahoo we have a decent number of trident topologies running in production, and I don't see it ever going fully away.  But I do see the appeal of having something like apex or flink, where you can have exactly once processing without the requirement for kafka to do the replay.  It would also be nice to be able to have the exact same code work for both exactly once, at least once, and best effort processing.

But that would be a long ways off if we do it.

STORM_JSTORM_FEATURE_DIFF.md
```diff
@@ -0,0 +1,30 @@
...
+Nimbus HA	| Support to configure more than one spare nimbus. When the master nimbus is down, a most appropriate spare nimbus(topologys in disk mostly matches the recoreds in zk) will be chosen to be promoted. | Different solution. | | 
+Topology structure | worker/executor/task | worker/task | It might be the biggest diff between Storm and JStorm. So, not sure if there are some features related to "executor" can not be merged. (Rebalancing allows us to combine multiple tasks on a single executor)  This may go away in the future once RAS is stable |
+Topology Master | (Heartbeat server is similar for scaling on a pull request) | New system bolt "topology master" was added, which is responsible for collecting task heartbeat info of all tasks and reports to nimbus. Besides task hb info, it also can be used to dispatch control messages in topology. Topology master significantly reduce the amout of read/write to zk.	Before this change, zk is the bottleneck to depoly big cluster and topology.
```
hustfxj: TM is mainly used to manage topology later，and  we hope  users can  defines control  stream in the future.

revans2:
This is now captured at https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=61328109

PR #878, opened 2015-11-13
STORM-1203. worker metadata file creation doesn't use storm.log.dir config

harshach:


revans2:
@harshach great find.

+1

ptgoetz:
+1

HeartSaVioR:
Great, +1

HeartSaVioR:
Btw, seems like LOG-DIR is not used anywhere. Could you check it?

PR #879, opened 2015-11-14
[STORM-1204] Fixing attempt to access to directory before authorizing user

kishorvpatil:
Fixing dumps endpoints, if URL request has no file/directory it return 500 instead of page-not-found.

### DIFF
--- a/storm-core/src/clj/backtype/storm/daemon/logviewer.clj
+++ b/storm-core/src/clj/backtype/storm/daemon/logviewer.clj
@@ -950,6 +950,17 @@
       :headers {"Access-Control-Allow-Origin" origin
                 "Access-Control-Allow-Credentials" "true"})))
 
+(defn get-profiler-dump-files
+  [dir]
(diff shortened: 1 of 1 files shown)


storm-core/src/clj/backtype/storm/daemon/logviewer.clj
```diff
@@ -967,15 +978,30 @@
...
+                           topo-id
+                           file-path-separator
+                           port))
+           file (File. (str log-root
+                            file-path-separator
+                            topo-id
+                            file-path-separator
+                            port
+                            file-path-separator
+                            filename))]
```
redsanket: Can we convert this to a utility method, it will be useful

redsanket:
minor suggestion, +1 NB

revans2:
+1

kishorvpatil:
@redsanket We are doing this only are couple of places.. no making it a utility may not be much useful.

zhuoliu:
+1

PR #880, opened 2015-11-15
STORM-1009:Link on http://storm.apache.org/documentation/Kestrel-and-…

jayapriya90:
…Storm.html is wrong

### DIFF
--- a/docs/documentation/Kestrel-and-Storm.md
+++ b/docs/documentation/Kestrel-and-Storm.md
@@ -7,7 +7,7 @@ This page explains how to use to Storm to consume items from a Kestrel cluster.
 
 ## Preliminaries
 ### Storm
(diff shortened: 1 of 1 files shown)


hustfxj:
+1

revans2:
@jayapriya90 sadly the documentation was moved to the asf-site branch, but the old docs were not removed from the master branch.  Could you update your pull request to be against that branch instead?

I am +1 on the change.

jayapriya90:
@hustfxj  and @revans2 Thanks for the review and update. I created new pull request #882 for asf-site branch. 

I've also noticed in asf-site branch that auto-generated html files are not ignored. So when I made this change, there were lots of unrelated changes to html files. In previous master branch, auto generated html files were ignored. Should we do the same for asf-site branch as well? If so, I can create a JIRA for that.

wuchong:
Maybe we should close this PR as the new same pull request #882 have been created.

jerrypeng:
can we close this dup PR now?

PR #881, opened 2015-11-16
[STORM-1208] Guard against NPE, and avoid using NaN values

d2r:
* Removes malformed preconditions that don't do anything, and if they did work would throw Errors that would crash nimbus.
* When reading and writing average-type metrics, use `0` instead of `NaN`
* Avoid NPE accessing some maps by doing `(get m k)` instead of `(m k)`.

### DIFF
--- a/storm-core/src/clj/backtype/storm/stats.clj
+++ b/storm-core/src/clj/backtype/storm/stats.clj
@@ -288,10 +288,11 @@
       rate)))
 
 (defn valid-number?
-  "Returns true if x is a number that is not NaN, false otherwise"
+  "Returns true if x is a number that is not NaN or Infinity, false otherwise"
   [x]
   (and (number? x)
-       (not (Double/isNaN x))))
+       (not (Double/isNaN x))
+       (not (Double/isInfinite x))))
 
 (defn apply-default
   [f defaulting-fn & args]
--- a/storm-core/src/clj/backtype/storm/stats.clj
+++ b/storm-core/src/clj/backtype/storm/stats.clj
@@ -287,18 +287,42 @@
       specific-stats
       rate)))
 
+(defn valid-number?
+  "Returns true if x is a number that is not NaN, false otherwise"
+  [x]
+  (and (number? x)
+       (not (Double/isNaN x))))
+
+(defn apply-default
+  [f defaulting-fn & args]
+  (apply f (map defaulting-fn args)))
+
+(defn apply-or-0
+  [f & args]
+  (apply apply-default
+         f
+         #(if (valid-number? %) % 0)
+         args))
+
+(defn sum-or-0
+  [& args]
+  (apply apply-or-0 + args))
+
+(defn product-or-0
+  [& args]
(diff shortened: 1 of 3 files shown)


redsanket:
+1 NB

zhuoliu:
Nice fix. +1

hustfxj:
it looks good

storm-core/src/jvm/backtype/storm/metric/internal/LatencyStatAndMetric.java
```diff
@@ -242,7 +248,7 @@ private synchronized void rotate(long lat, long count, long timeSpent, long targ
             totalCount += countBuckets[i];
             timeNeeded -= bucketTime[i];
         }
-        return ((double)totalLat)/totalCount;
+        return totalCount > 0 ? ((double)totalLat)/totalCount : 0.0;
```
wuchong: Maybe we can pull the code out into a function named like `div-or-0`, as it appears more than once.

d2r: Yes we should do this.

In fact, while reading the documentation again, it seems we must guard against both values of `NaN` and `Infinity`.

```
user=> (/ 0. 0.)
NaN
user=> (/ 1. 0.)
Infinity
```

We have only seen the NaN case recently, but it would not hurt to handle both cases, since the dividend and divisor could be 0 independently, especially since the metrics are sampled independently.


I will update the code to refactor and to check both.

d2r:
@hustfxj , @kishorvpatil , @zhuoliu , @redsanket,

Added checking for `Infinity` plus some refactoring as per @hustfxj's review.

kishorvpatil:
still +1.

wuchong:
+1 LGTM

storm-core/src/clj/backtype/storm/stats.clj
```diff
@@ -335,30 +358,21 @@
...
-                                     0)))]
+  (letfn [(weight-avg [id avg]
+            (let [num-e (idk->executed id)]
+              (product-or-0 avg num-e)))]
     (into {}
       (for [k (keys idk->exec-avg)]
-        [k {:executeLatencyTotal (weight-avg k (idk->exec-avg k))
-            :processLatencyTotal (weight-avg k (idk->proc-avg k))
+        [k {:executeLatencyTotal (weight-avg k (get idk->exec-avg k))
+            :processLatencyTotal (weight-avg k (get idk->proc-avg k))
```
jerrypeng: what is the point of adding the get especially for (get idk->exec-avg k)? you are getting the keys from the map and then accessing it right? the keys cannot possibly not exist in the map

d2r:
```
user=> (def nilmap nil)
#'user/nilmap
user=> (get nilmap :somekey)
nil
user=> (nilmap :somekey)
NullPointerException   user/eval3 (NO_SOURCE_FILE:4)
```
`get` guards against the map itself is `nil`.

PR #884, opened 2015-11-17
[STORM-126] Add Lifecycle support API for worker nodes

schonfeld:
This PR is prob the 112th attempt at this ticket... But I'd really like support for worker hooks... So, here's my crack at it. Still missing tests -- I'd love some guidance re what exactly I should be testing/where I should be testing it. Thanks!

### DIFF
--- a/storm-core/src/clj/backtype/storm/daemon/worker.clj
+++ b/storm-core/src/clj/backtype/storm/daemon/worker.clj
@@ -551,7 +551,7 @@
 
 (defn run-worker-start-hooks [worker]
   (let [topology (:topology worker)
-        topo-conf (:conf worker)
+        topo-conf (:storm-conf worker)
         worker-topology-context (worker-context worker)
         hooks (.get_worker_hooks topology)]
     (dofor [hook hooks]
--- a/storm-core/src/clj/backtype/storm/daemon/worker.clj
+++ b/storm-core/src/clj/backtype/storm/daemon/worker.clj
@@ -553,12 +553,11 @@
   (let [topology (:topology worker)
         topo-conf (:conf worker)
         worker-topology-context (worker-context worker)
-        task-ids (:task_ids worker)
         hooks (.get_worker_hooks topology)]
     (dofor [hook hooks]
       (let [hook-bytes (Utils/toByteArray hook)
             deser-hook (Utils/javaDeserialize hook-bytes BaseWorkerHook)]
-        (.start deser-hook topo-conf worker-topology-context task-ids)))))
+        (.start deser-hook topo-conf worker-topology-context)))))
 
 (defn run-worker-shutdown-hooks [worker]
   (let [topology (:topology worker)
--- a/storm-core/src/clj/backtype/storm/daemon/worker.clj
+++ b/storm-core/src/clj/backtype/storm/daemon/worker.clj
@@ -680,7 +680,6 @@
 
                     (close-resources worker)
 
-                    ;; TODO: here need to invoke the "shutdown" method of WorkerHook
                     (log-message "Trigger any worker shutdown hooks")
                     (run-worker-shutdown-hooks worker)
 
--- a/storm-core/src/clj/backtype/storm/daemon/worker.clj
+++ b/storm-core/src/clj/backtype/storm/daemon/worker.clj
@@ -22,7 +22,8 @@
   (:require [backtype.storm [disruptor :as disruptor] [cluster :as cluster]])
   (:require [clojure.set :as set])
   (:require [backtype.storm.messaging.loader :as msg-loader])
-  (:import [java.util.concurrent Executors])
+  (:import [java.util.concurrent Executors]
+           [backtype.storm.hooks IWorkerHook BaseWorkerHook])
   (:import [java.util ArrayList HashMap])
   (:import [backtype.storm.utils Utils TransferDrainer ThriftTopologyUtils WorkerBackpressureThread DisruptorQueue])
   (:import [backtype.storm.grouping LoadMapping])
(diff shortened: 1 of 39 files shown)


schonfeld:
Added some tests around `ThriftTopologyUtils`

schonfeld:
ping @revans2 @harshach -- sorry to bug, but, mind taking a look?

storm-core/src/clj/backtype/storm/daemon/worker.clj
```diff
@@ -665,6 +686,8 @@
                     (close-resources worker)
 
                     ;; TODO: here need to invoke the "shutdown" method of WorkerHook
```
revans2: Please remove this TODO, you implemented it.

storm-core/src/jvm/backtype/storm/hooks/BaseWorkerHook.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+package backtype.storm.hooks;
+
+import backtype.storm.task.WorkerTopologyContext;
+
+import java.io.Serializable;
+import java.util.List;
+import java.util.Map;
+
+public class BaseWorkerHook implements IWorkerHook, Serializable {
```
revans2: Could you add in some javadoc comments.

storm-core/src/jvm/backtype/storm/hooks/BaseWorkerHook.java
```diff
@@ -0,0 +1,37 @@
...
+    private static final long serialVersionUID = 2589466485198339529L;
+
+    @Override
+    public void start(Map stormConf, WorkerTopologyContext context, List taskIds) {
+
```
revans2: Could we put a comment in the body here like `//NOOP`

storm-core/src/jvm/backtype/storm/hooks/BaseWorkerHook.java
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    @Override
+    public void shutdown() {
+    }
```
revans2: Similar comment here.

storm-core/src/jvm/backtype/storm/hooks/IWorkerHook.java
```diff
@@ -0,0 +1,29 @@
...
+import backtype.storm.task.WorkerTopologyContext;
+
+import java.io.Serializable;
+import java.util.List;
+import java.util.Map;
+
+public interface IWorkerHook extends Serializable {
+    void start(Map stormConf, WorkerTopologyContext context, List taskIds);
+    void shutdown();
+}
```
revans2: Could you put in some javadoc comments about this interface and the methods in it.

storm-core/test/jvm/backtype/storm/utils/ThriftTopologyUtilsTest.java
```diff
@@ -0,0 +1,94 @@
...
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
+import junit.framework.TestCase;
+import org.junit.Assert;
+import org.junit.Test;
+
+import java.nio.ByteBuffer;
+import java.util.Set;
+
+public class ThriftTopologyUtilsTest extends TestCase {
```
revans2: For JUNIT4 I don't think you should extend TestCase.  Is there a reason you are doing it?

storm-core/src/jvm/backtype/storm/hooks/IWorkerHook.java
revans2: Especially describing the type that taskIds is.  It would be nice to put in generics for it if we could.

Parth-Brahmbhatt:
Thanks for the contribution, any reason you decided to make this hook part of serialized topology. If you see some other examples, like the nimbus hook (though it is not really a great example as its on nimbus side and not worker but still close) we ask the user to just provide the Fully qualified classname as a config option, create an instance of it using reflection and invoke prepare (start in your case) or cleanup on that instance. The only advantage of putting this as part of topology is that users will be able to provide objects that are completely serialized so it can be initialized with constructor args or with any other way that relies on instance variable initialization but I don't see that as a huge upside. On the other hand a consistent way to implement all hooks will make code easy to read and reason about.

storm-core/test/jvm/backtype/storm/utils/ThriftTopologyUtilsTest.java
schonfeld: I did that because all other test classes in `backtype.storm.utils` (DisruptorQueueBackpressureTest, DisruptorQueueTest, StormBoundedExponentialBackoffRetryTest, etc) extend `TestCase`... Should I remove?

revans2:
Overall this looks good.  I have a few minor comments, mostly around code cleanup.  My biggest concern is that we seem to be installing the IWorkerHook on all workers, but not providing a simple way for them to selectively decide if they want to run on a per-bolt/spout basis.  But that can be done later on if we need it.

schonfeld:
@Parth-Brahmbhatt @revans2 to be honest, the reason I implement this as a worker-specific serializable hook is cos that's what @nathanmarz suggested in the discussion on [STORM-126](https://issues.apache.org/jira/browse/STORM-126)... Seemed like a logical approach, since it is bound to a worker and not a specific component..?

storm-core/test/jvm/backtype/storm/utils/ThriftTopologyUtilsTest.java
revans2: It is very minor I am fine with leaving it.

revans2:
@schonfeld we can make it work easily with specific components.  The taskIds are not needed when calling the start method of IWorkerHook.  You can get them out by calling getThisWorkerTasks of the WorkerTopologyContext.  But that too is really minor.  You can also call getComponentId using the task id to get the component ID.

So if we wanted to we could have something like (probably does not compile)
```java
public class ComponentWorkerHook implements IWorkerHook {
    private final IWorkerHook _wrapped;
    private final Set<String> _comps;
    private boolean _active = false;

    public ComponentWorkerHook(IWorkerHook wrapped, Set<String> comps) {
        _wrapped = wrapped;
        _comps = new HashSet<>(comps);
    }

    public void addComp(String comp) {
        _comps.add(comp);
    }

   @Override
    public void start(Map stormConf, WorkerTopologyContext context, List<Integer> taskIds) {
         Set<String> found = new HashSet<String>();
         for (Integer id: taskids) {
             found.put(context.getComponentId(id));
         }
         found.retainAll(_comps);
         _active = !found.isEmpty();
        if (_active) {
            _wrapped.start(stromConf, context, taskIds);
        }
    }

    @Override
    public void shutdown() {
        if (_active) {
           _wrapped.shutdown();
        }
    }
}
```

I am +1 for merging this in, but I want to hear from @Parth-Brahmbhatt if he really wants to change this to be config based instead of through TopologyBuilder.

schonfeld:
@revans2 yah, you're right.. I'll go ahead and remove the taskIds from @start. re ComponentWorkerHook... I can't think of a reason you'd want to run the a WorkerHook only on specific components... And if you do want that, shouldn't we just implement `start` & `shutdown` methods in `IComponent`?

revans2:
@schonfeld the reason for doing it per component is because you have a framework, like spring, that wants to be setup and cleaned up, but only a subset of the components use it.  I supposed having the worker hook run on workers that don't use spring probably is not that bad, but I can also see that if it adds extra overhead some people may want to turn it off.

I am still +1 on this change.

schonfeld:
ping @Parth-Brahmbhatt?

Parth-Brahmbhatt:
I'm +1 too.

schonfeld:
Fantastic! I'll rebase on master real quick...

revans2:
@schonfeld sorry this took so long.  Could you rebase again?  I am going to be out for the Thanksgiving Holiday, so @Parth-Brahmbhatt no need to wait for me to give a +1 again.

schonfeld:
@revans2 all good! we actually caught a bug (in that last commit apache/storm@9bc9350f613a75de566cf670874c6a037ecc5bfd) when we ran this in prod today.

Anyway, rebased & pushed. Just gotta wait for the tests to finish...

PR #885, opened 2015-11-17
[STORM-1210] Set Output Stream id in KafkaSpout

Zhiqiang-He:
topicAsStreamId can only set output stream id to topic name. In some case ,we need to set output stream id to other name.


wuchong:
Looks good to me.

But the condition code is a little reduplicate and verbose , I prefer to check only one variable to set streamId with reseting outputStreamId in SpoutConfig constructor.

```java
this.outputStreamId = topicAsStreamId ? topic : this.outputStreamId;
```

Not a big deal, just something trivial .

wuchong:
Looks good to me.

But the condition code (`topicAsStreamId` and `outputStreamId`) is a little reduplicate as the logic is almost the same.

Zhiqiang-He:
@wuchong 
Yes, you are right. but for compatibility reasons, I add this argument based on topicAsStreamId .

jerrypeng:
just a quick question. where is outputStreamId set?

Zhiqiang-He:
@jerrypeng 
outputStreamId  is used in declareStream as emit method.
It is very useful when bolt after kafkaspout recive two or more stream messages.

revans2:
+1 the code looks good to me.  I agree with @wuchong that the logic is very similar between the two and there is the potential to combine some code, them but it is small enough I don't think it is that important.

Zhiqiang-He:
@revans2 I already removed topicAsStreamId, please review it. thanks.

external/storm-kafka/src/jvm/storm/kafka/SpoutConfig.java
```diff
@@ -27,8 +27,7 @@
     public String zkRoot = null;
     public String id = null;
 
-    // if set to true, spout will set Kafka topic as the emitted Stream ID
-    public boolean topicAsStreamId = false;
+    public String outputStreamId;
```
wuchong: Minor, but we can note  here that if outputStreamId not set, spout will set Kafka topic as the emitted Stream ID

wuchong:
+1 Nice work. It is much clearer now.

ppoulosk:
+1, NB

external/storm-kafka/src/jvm/storm/kafka/PartitionManager.java
```diff
@@ -149,9 +150,9 @@ public EmitState next(SpoutOutputCollector collector) {
             }
             
             if ((tups != null) && tups.iterator().hasNext()) {
-                if(_spoutConfig.topicAsStreamId) {
+               if (!Strings.isNullOrEmpty(_spoutConfig.outputStreamId)) {
```
rfarivar: Isn't it an overkill to import a whole package (com.google.common.base.Strings) to only test for null or empty? Can't we just use: 
if (_spoutConfig.outputStreamId!=null && _spoutConfig.outputStreamId!="") 
or something to that effect instead of importing a new library?

Zhiqiang-He:
@rfarivar Because guava is already import in storm-kafka.
And com.google.common.collect.ImmutableMap is already used in PartitionManager.java.

ptgoetz:
+1

revans2:
Still +1

PR #886, opened 2015-11-17
add storm-id to worker log filename

jessicasco:
add storm-id to worker log filename, so we can view logs group by the topology-id

### DIFF
--- a/storm-core/src/clj/backtype/storm/daemon/supervisor.clj
+++ b/storm-core/src/clj/backtype/storm/daemon/supervisor.clj
@@ -852,7 +852,7 @@
           gc-opts (substitute-childopts (if top-gc-opts top-gc-opts (conf WORKER-GC-CHILDOPTS)) worker-id storm-id port mem-onheap)
           topo-worker-logwriter-childopts (storm-conf TOPOLOGY-WORKER-LOGWRITER-CHILDOPTS)
           user (storm-conf TOPOLOGY-SUBMITTER-USER)
-          logfilename "worker.log"
+          logfilename (logs-filename storm-id port)
           workers-artifacts (worker-artifacts-root conf)
           logging-sensitivity (storm-conf TOPOLOGY-LOGGING-SENSITIVITY "S3")
           worker-childopts (when-let [s (conf WORKER-CHILDOPTS)]


kishorvpatil:
@jessicasco change worker log, requires changes to core ui, logviewer, worker in order to make sure the links and other search functionality don't break.

wuchong:
We have topology-id and port in log directory,  current log structures under storm-local/workers-artifacts:
```
--topoId0
--topoId1
------port1
------port2
---------worker.log
---------worker.log.err 
---------worker.log.out
```

jessicasco:
@wuchong  I got it. Thank you.

jessicasco:
@wuchong    How about 0.9.x version? I checked 0.9.6 still has `worker-{port}.log`, which mix different topology log to one file.  Will the `workers-artifacts` log feature merge to 0.9.x version?

wuchong:
@jessicasco  yes,  you are right.
The `workers-artifacts` feature is merged by @zhuoliu ,  you may ask him.

zhuoliu:
hi jessicasco, though worker-artifacts changes log directory structure and naming, when you download worker.log from UI, the storm id and port will be automatically append to the file name, which is quite convenient.

jessicasco:
@zhuoliu  Thank you for your explanation.   Is there a simple way to split topology log in `worker-{port}.log` for storm 0.9.6?

zhuoliu:
Hi jessicasco, for the versions before workers-artifacts, the following code in util.clj might achieve what you want:
```
(defn- logs-rootname [storm-id port]
  (str storm-id "-worker-" port))

(defn logs-filename [storm-id port]
  (str (logs-rootname storm-id port) ".log"))

(defn logs-metadata-filename [storm-id port]
  (str (logs-rootname storm-id port) ".yaml"))

(def worker-log-filename-pattern #"^((.*-\d+-\d+)-worker-(\d+))\.log")

(defn get-log-metadata-file
  ([fname]
    (if-let [[_ _ id port] (re-matches worker-log-filename-pattern fname)]
      (get-log-metadata-file id port)))
  ([id port]
    (clojure.java.io/file LOG-DIR "metadata" (logs-metadata-filename id port))))
```

jessicasco:
@zhuoliu Thank you for your code snippet.

darionyaphet:
@jessicasco could be named with `topology-name` ?

jessicasco:
@darionyaphet  not `topology-id` but `topology-name`?

zhuoliu:
@darionyaphet, in 0.10 and before, worker log file has topology-id and port in its name. In latest version, the topo-name and port no longer exist in its same (please see [storm-901 | https://github.com/apache/storm/pull/822]). However, when user download log file from ui, topo-id and port will be added to the downloaded file's name automatically.

darionyaphet:
@jessicasco yep 

current log's name looks like `worker-{port}.log` rename to 'topo-name-{port}' maybe better ?

jessicasco:
@darionyaphet My code rename the log file to `{topology-id}-{port}.log`, but as @zhuoliu says they have a better solution since 0.10.

hustfxj:
@zhuoliu  Storm current log structures is:
storm-home/logs/
------supervisor.log
------nimbus.log
------workers-artifacts/
-----------topo1/
------------------port1/
------------------port2/
------------------------worker.log
------------------------worker.log.err
------------------------gc.log

worker's gc log's name maybe like {workers-artifacts}-{topo}-{port}-{worker}.gc. Supervisor will restart the worker if some worker dead. Then the worker's gc name is the same as before. I find the dead worker gc log is   coverd with new gc log

darionyaphet:
@jessicasco sorry  I'm using 0.9.4 ..

zhuoliu:
Hi Fang, @hustfxj. Great question! We also met the same problem. Actually our setting for gc log name is "-Xloggc:artifacts/gc.%p.log". Then gc log won't be overwritten due to worker restating.
Then we will also need to https://issues.apache.org/jira/browse/STORM-1206 to avoid too many gc files' names crashing the logviewer JVM (this might happen in  a very very extreme and rare case, only once in our cluster).

hustfxj:
very good! thank you  @zhuoliu

PR #887, opened 2015-11-17
STORM-1213: Remove sigar binaries from source tree

ptgoetz:
See https://issues.apache.org/jira/browse/STORM-1213 for description.

Also removes the LICENSE entry for sigar, since I believe that is unnecessary since it is Apache licensed.


kishorvpatil:
+1

jerrypeng:
+1

revans2:
+1

revans2:
Actually I just tried to run this and it looks like the libraries are under `resources/resources/` instead of just `resources/`  because of this I am getting linking errors when trying to use the metrics.

external/storm-metrics/pom.xml
```diff
@@ -48,6 +48,13 @@
       </exclusions>
     </dependency>
     <dependency>
+      <groupId>org.fusesource</groupId>
+      <artifactId>sigar</artifactId>
+      <version>1.6.4</version>
```
revans2: We should also make this version a property or set in the parent pom, because there are multiple packages tied to this same version.  And we should remove the comment on line 39, as it is not needed.

external/storm-metrics/pom.xml
```diff
@@ -55,4 +62,28 @@
...
+        <version>2.10</version>
+        <executions>
+          <execution>
+            <id>unpack-dependencies</id>
+            <phase>generate-resources</phase>
+            <goals>
+              <goal>unpack-dependencies</goal>
+            </goals>
+            <configuration>
+              <outputDirectory>${project.build.directory}/classes/resources/resources</outputDirectory>
```
revans2: This is the line causing the issue.  It should just be `<outputDirectory>${project.build.directory}/classes/resources</outputDirectory>`

revans2:
Sorry about the comments after giving it a +1.  I manually ran LatencyVsThroughput and saw that it was not working.

revans2:
Even with changing the directory I am getting 
```
java.lang.UnsatisfiedLinkError: org.hyperic.sigar.Sigar.getPid()J
	at org.hyperic.sigar.Sigar.getPid(Native Method) ~[stormjar.jar:0.11.0-SNAPSHOT]
	at org.apache.storm.metrics.sigar.CPUMetric.<init>(CPUMetric.java:38) ~[stormjar.jar:0.11.0-SNAPSHOT]
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method) ~[?:1.8.0]
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62) ~[?:1.8.0]
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45) ~[?:1.8.0]
	at java.lang.reflect.Constructor.newInstance(Constructor.java:408) ~[?:1.8.0]
	at java.lang.Class.newInstance(Class.java:433) ~[?:1.8.0]
	at backtype.storm.utils.Utils.newInstance(Utils.java:91) ~[storm-core-0.11.0-SNAPSHOT.jar:0.11.0-SNAPSHOT]
	at backtype.storm.metric.SystemBolt.registerMetrics(SystemBolt.java:150) ~[storm-core-0.11.0-SNAPSHOT.jar:0.11.0-SNAPSHOT]
	at backtype.storm.metric.SystemBolt.prepare(SystemBolt.java:143) ~[storm-core-0.11.0-SNAPSHOT.jar:0.11.0-SNAPSHOT]
	at backtype.storm.daemon.executor$fn__5157$fn__5170.invoke(executor.clj:777) ~[storm-core-0.11.0-SNAPSHOT.jar:0.11.0-SNAPSHOT]
	at backtype.storm.util$async_loop$fn__556.invoke(util.clj:480) [storm-core-0.11.0-SNAPSHOT.jar:0.11.0-SNAPSHOT]
	at java.lang.Thread.run(Thread.java:744) [?:1.8.0]
```

We need to do some more work to understand the difference between this patch and what we were doing before.

ptgoetz:
@revans2 Ah... okay. So we need to change `resources/resources` to just `resources`, and remove the `-1.6.4` from the libraries, correct?

That should be doable with the dependency plugin.

revans2:
@ptgoetz Yes you are correct about the changes.

ptgoetz:
@revans2 I couldn't get the maven dependency plugin to do what I wanted, so I changed tactics. I switched to using the maven antrun plugin to download and unpack the sigar native binaries. It downloads the archive to the local maven repository alongside the other sigar dependencies so it only needs to download it once.

The good news is that allowed me to use the full binary archive, so it includes support for windows and other OSes/architectures. On the down side it's somewhat brittle compared to using "pure" maven, and IMO is kind of ugly. It's also hard-coded to use google code, which is supposedly going away at some point, but that's the only place I could find the archive with a published checksum.

revans2:
@ptgoetz 
+1 the changes look good and my manual tests all pass.

I Agree the it kind of ugly.  Ideally what we need to do is to talk to sigar and see if they can update their native jar to do what we want.  It should be fairly simple to have them remove the version number from the jar.

PR #888, opened 2015-11-17
[STORM-1215] Use Asynchronous logger for improving performance.

kishorvpatil:
The changes to loggers for all daemons and workers.
- Turn off immediateFlush.
- Enable async logging.
- Turn off Immediate Fail for log events on syslog logger to avoid making blocking call without timeout.

### DIFF
--- a/bin/storm.py
+++ b/bin/storm.py
@@ -448,6 +448,7 @@ def nimbus(klass="backtype.storm.daemon.nimbus"):
     cppaths = [CLUSTER_CONF_DIR]
     jvmopts = parse_args(confvalue("nimbus.childopts", cppaths)) + [
         "-Dlogfile.name=nimbus.log",
+        "-DLog4jContextSelector=org.apache.logging.log4j.core.async.AsyncLoggerContextSelector",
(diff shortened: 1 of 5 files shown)


d2r:
This also switches to UTF-8 encoding for syslog, for compatibility.  We have seen issues with this in the past.

Evidently we cannot configure the context except via a System property, so looks OK to me.

+1

revans2:
The log level changes on the UI still work, and performance seems fine too, but if I enable topology.debug I am seeing some very odd things that I would like to explore some more.

revans2:
+1 I just tested and the workers are no longer crashing, but it might be good to set immediateFlush="false" in the worker.

kishorvpatil:
@revans2  Turned off immediateFlush for worker.

revans2:
+1 looks good to me

revans2:
@d2r any other comments on the most recent changes?

d2r:
Looks good to me +1

rajeshksv:
You mean Turn ON Immediate Fail right ?

PR #891, opened 2015-11-19
STORM-1218: Use markdown for JavaDoc

ptgoetz:
JIRA: https://issues.apache.org/jira/browse/STORM-1218

This is a first pass at converting the JavaDoc in `storm-core` to markdown, using [pegdown-doclet](https://github.com/Abnaxos/pegdown-doclet). If there is support from the community, I will move on to other components, though any html will just pass through, so it shouldn't garble, and will likely improve any existing javadoc.

The first commit adds the corresponding maven javadoc plugin to the default build. This means it will increase build time when you run `mvn install`, but it will also spit out warnings about javadoc problems and may encourage developers to fix or improve documentation problems.

The second commit is a quick pass over `storm-core` to remove unnecessary HTML and some minor javadoc comment linting. Only javadoc comments were modified.

The third commit adds the `highlight.js` license to both the source and binary LICENSE files, since the generated javadoc includes it. I'm not sure this is necessary, since the JavaDoc is generated, but I will find out if there's support for this patch.

*Note:* The [pegdown-doclet licence](https://github.com/Abnaxos/pegdown-doclet/blob/master/LICENSE.txt) is GPL, but it's only used at build time, so that does not seem to present a licensing issue.


ptgoetz:
Re: Merge conflicts, I'm happy to resolve therm at merge time.

revans2:
I am +1 on the concept.  I don't see where highlight.js is added to the codebase.  It is automatically inserted into the created javadocs?

ptgoetz:
@revans2 Yes, it is inserted into the generated HTML. Since javadoc isn't part of the source release, it probably only needs to be added to the LICENSE for the binary distribution.

ptgoetz:
I removed the highlight.js license from the source release LICENSE file since generated javadoc is not part of a source release.

ptgoetz:
**Note:** This pull request originally included the following commit: https://github.com/apache/storm/commit/e03b28ca6222f61471b4acb1a6086aab9b597b8a

Earlier today I inadvertently pushed that commit to master. I have since reverted it in master. The last commit in this PR re-adds what was reverted, but only in my repository.

harshach:
+1. Thanks this will make doc writing easier.

redsanket:
NB +1

harshach:
Thanks @ptgoetz merged into master.

PR #893, opened 2015-11-19
STORM-1219: Fix HDFS and Hive bolt flush/acking

arunmahadevan:
HDFS and Hive bolts were setting the default tick tuple interval in the
prepare() method, which was not taking effect. Correctly set a default
and return from getComponentConfiguration if the user does not
explicitly set a value.

### DIFF
--- a/external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/AbstractHdfsBolt.java
+++ b/external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/AbstractHdfsBolt.java
@@ -48,6 +48,11 @@ import java.util.TimerTask;
 public abstract class AbstractHdfsBolt extends BaseRichBolt {
     private static final Logger LOG = LoggerFactory.getLogger(AbstractHdfsBolt.class);
     private static final Integer DEFAULT_RETRY_COUNT = 3;
+    /**
+     * A low default value so that tuples are flushed and acked as soon as possible to avoid
+     * replay and spouts can continue emitting without hitting TOPOLOGY_MAX_SPOUT_PENDING.
+     */
+    private static final int DEFAULT_TICK_TUPLE_INTERVAL_SECS = 1;
 
     protected ArrayList<RotationAction> rotationActions = new ArrayList<RotationAction>();
     private Path currentFile;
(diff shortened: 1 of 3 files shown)


harshach:
+1

dossett:
Was the prior approach not working because getComponentConfiguration() is called before prepare() ?

arunmahadevan:
@dossett it appears that the component configuration is merged with the global storm config. This happens when the topology is constructed and serialized. The merged config is then passed to `prepare`, so if the map passed to prepare is updated, it does not seem to have any effect.

dossett:
Interesting, this must have been a regression when that particular code was refactored from HdfsBolt up to AbstractHdfsBolt.

arunmahadevan:
@harshach @dossett any other comments? Can we merge ?

dossett:
It's still not clear to me why the original approach doesn't work, but that's probably a problem with my understanding of storm internals.

+1

harshach:
@arunmahadevan I might be overlooking something but if the default tick tuple is at 1secs and this will cause the tuples to be flushed much more frequently without even reaching to batchsize. This behavior can create lot of small files in hdfs and also for hive.

harshach:
I think the default should be at least 15secs than 1

dossett:
@harshach I don't believe flushing would force early file rotations.  On second thought, I do agree that 15 seconds would be better than 1 second.

harshach:
@dossett It does for hive. It creates small files , which eventually will be merged by compactor. But its not good to have small files especially if your compactor runs like every hour.

dossett:
@harshach Ah, thanks, I was only thinking of the hdfs bolt side.

arunmahadevan:
@harshach with 15 secs, the throughput was very low since the spout could not emit more tuples after hitting max.spout.pending. The low value (1s) default also makes sure that the value is not more than topology.message.timeout.secs in case user changes the default timeout (if the value is more it would cause unnecessary duplicates). There seems to be no way to get the message timeout value in getComponentConfiguration and adjust the tick tuple interval accordingly.

The 1s is only the default, the user can change it via HiveOptions if compaction is a concern.

harshach:
@arunmahadevan throughput should control by batch size not by tickTupleinterval. If the tuples are not appearing the hive table or hdfs file, one should lower the batch size or configure the tickTupleInterval to lower but making this default is not ideal. Most often users won't configure them and can create adverse effects in production.  15 secs will at least guarantee the tuples won't be timing out and not create smaller files.

arunmahadevan:
@harshach yes its a trade off. Updated default to 15s

harshach:
+1

revans2:
+1

PR #894, opened 2015-11-19
STORM-1220. Avoid double copying in the Kafka spout.

haohui:
Currently the Kafka spout performs an extra copy for all payloads in Kafka when integrating with Storm. This PR proposes to avoid the extra copy to improve performance.

### DIFF
--- a/external/storm-kafka/src/jvm/storm/kafka/StringScheme.java
+++ b/external/storm-kafka/src/jvm/storm/kafka/StringScheme.java
@@ -20,23 +20,27 @@ package storm.kafka;
 import backtype.storm.spout.Scheme;
 import backtype.storm.tuple.Fields;
 import backtype.storm.tuple.Values;
+import backtype.storm.utils.Utils;
 
-import java.io.UnsupportedEncodingException;
+import java.nio.ByteBuffer;
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
 import java.util.List;
 
 public class StringScheme implements Scheme {
-
+    private static final Charset UTF8_CHARSET = StandardCharsets.UTF_8;
     public static final String STRING_SCHEME_KEY = "str";
 
-    public List<Object> deserialize(byte[] bytes) {
+    public List<Object> deserialize(ByteBuffer bytes) {
         return new Values(deserializeString(bytes));
     }
 
-    public static String deserializeString(byte[] string) {
-        try {
-            return new String(string, "UTF-8");
-        } catch (UnsupportedEncodingException e) {
-            throw new RuntimeException(e);
+    public static String deserializeString(ByteBuffer string) {
+        if (string.hasArray()) {
+            int base = string.arrayOffset();
+            return new String(string.array(), base + string.position(), string.remaining());
+        } else {
+            return new String(Utils.toByteArray(string), UTF8_CHARSET);
         }
     }
 
--- a/storm-core/src/jvm/backtype/storm/spout/RawScheme.java
+++ b/storm-core/src/jvm/backtype/storm/spout/RawScheme.java
@@ -18,12 +18,17 @@
 package backtype.storm.spout;
 
 import backtype.storm.tuple.Fields;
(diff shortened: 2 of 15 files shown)


storm-core/src/jvm/backtype/storm/spout/RawScheme.java
```diff
@@ -18,11 +18,13 @@
...
 
 import backtype.storm.tuple.Fields;
+
+import java.nio.ByteBuffer;
 import java.util.List;
 import static backtype.storm.utils.Utils.tuple;
 
 public class RawScheme implements Scheme {
-    public List<Object> deserialize(byte[] ser) {
+    public List<Object> deserialize(ByteBuffer ser) {
```
arunmahadevan: Will this break the compatibility at the receiver ?. Earlier the tuple contained a `byte[]` and with the change it would contain a `ByteBuffer`.

external/storm-kafka/src/jvm/storm/kafka/StringScheme.java
```diff
@@ -20,21 +20,23 @@
...
-    public List<Object> deserialize(byte[] bytes) {
+    public List<Object> deserialize(ByteBuffer bytes) {
         return new Values(deserializeString(bytes));
     }
 
-    public static String deserializeString(byte[] string) {
+    public static String deserializeString(ByteBuffer string) {
         try {
-            return new String(string, "UTF-8");
+            return new String(Utils.toByteArray(string), "UTF-8");
```
arunmahadevan: This would create a copy. If the ByteBuffer is array backed, you could use `ByteBuffer.array()` to avoid copy.

storm-core/src/jvm/backtype/storm/spout/RawScheme.java
harshach: @arunmahadevan it might but I don't seen any one using this apart from KafkaSpout

arunmahadevan: @harshach Actually I was referring to the receiver (the bolts) that might be currently doing something like `byte[] bytes = inputTuple.getBinaryByField("bytes");` to get the data emitted from kafka spout. It appears that the `deserialize` method returns a tuple that wraps the input ByteBuffer which gets emitted.

harshach: @arunmahadevan good point. we can use ByteBuffer.array() to have same signature.

external/storm-kafka/src/jvm/storm/kafka/StringScheme.java
```diff
@@ -20,23 +20,27 @@
...
         return new Values(deserializeString(bytes));
     }
 
-    public static String deserializeString(byte[] string) {
-        try {
-            return new String(string, "UTF-8");
-        } catch (UnsupportedEncodingException e) {
-            throw new RuntimeException(e);
+    public static String deserializeString(ByteBuffer string) {
+        if (!string.hasArray()) {
```
arunmahadevan: I think this should be `if (string.hasArray())`

external/storm-kafka/src/jvm/storm/kafka/StringScheme.java
```diff
@@ -20,23 +20,27 @@
...
+            int base = string.arrayOffset();
+            return new String(string.array(), base + string.position(), base + string.limit());
```
arunmahadevan: `String` ctor takes length as the last arg, so you need to pass `limit - pos`. Also the charset is missing.

harshach:
+1. @arunmahadevan can you take a look at it.

arunmahadevan:
Overall it looks good. I think https://github.com/apache/storm/pull/894#discussion_r45434875 should be addressed. Otherwise it may break existing bolts.

haohui:
Addressed in the latest patch.

harshach:
+1

harshach:
Thanks @haohui  merged into master.

PR #901, opened 2015-11-23
STORM-1200. Support collations of primary keys.

haohui:
This PR allows users to specify collations on primary keys so that the query engines can use this as a hint when processing streaming aggregations.


harshach:
+1

harshach:
Thanks @haohui

PR #903, opened 2015-11-25
STORM-1341 Let topology have own heartbeat timeout for multilang subprocess

HeartSaVioR:
* config name: topology.subprocess.timeout.secs
* if it's not specified, supervisor.worker.timeout.secs will be used


knusbaum:
+1
Looks fine to me.

PR #904, opened 2015-11-25
STORM-1353 Update "jstorm-import" branch to the latest JStorm release 2.1.0

bastiliu:
Update the reference banch "jstorm-import" with latest JStorm release 2.1.0, since many JStorm features listed in JStorm merging plan are not included.


hustfxj:
+1

wuchong:
+1

darionyaphet:
+1

ptgoetz:
+1

zhuoliu:
+1

ptgoetz:
Thanks @bastiliu I merged this into the jstorm-import.

**One important thing to note:**

This pull request introduced highchart.js which has a commercial license and can't be included in an Apache release. If we end up using the jstorm-ui code, highcharts.js will have to be removed and replaced with something that's compatible with ASF license policy.

bastiliu:
@ptgoetz Thanks for the information. We find a new solution which is compatible with ASF license policy.

bastiliu:
@ptgoetz Thanks for the information. We will find a new solution which is compatible with ASF license policy.

wuchong:
Hi @ptgoetz , thanks for the note. We have tried a lot of alternative solutions, such as: d3, chart.js, xCharts, chartist.js and ECharts. Finally we find that [ECharts](http://echarts.baidu.com/index-en.html) is best meet our needs and [its license](https://github.com/ecomfe/echarts/blob/master/LICENSE.txt) is much like BSD which is compatible with ASF license I think.

The image below is that we use ECharts to re-implement metric charts. 

image

and this is the version using highcharts

image

ptgoetz:
@wuchong Good to hear. Visually they look nearly identical, and the license looks compatible.

darionyaphet:
maybe need a modify list to display JStorm's update

PR #911, opened 2015-12-02
STORM-1360:  storm.py issues with python 3

ashnazg:
This changes the "print" and "except" syntax usages such that they test fine for me on Python 2.6.6 (RHEL) and 3.5.0 (Win7).

### DIFF
--- a/storm-multilang/python/src/main/resources/resources/storm.py
+++ b/storm-multilang/python/src/main/resources/resources/storm.py
@@ -196,7 +196,7 @@ class Bolt(object):
                     sync()
                 else:
                     self.process(tup)
-        except Exception, e:
+        except Exception as e:
             reportError(traceback.format_exc(e))
 
 class BasicBolt(object):
@@ -222,10 +222,10 @@ class BasicBolt(object):
                     try:
                         self.process(tup)
                         ack(tup)
-                    except Exception, e:
+                    except Exception as e:
                         reportError(traceback.format_exc(e))
                         fail(tup)
-        except Exception, e:
+        except Exception as e:
             reportError(traceback.format_exc(e))
 
 class Spout(object):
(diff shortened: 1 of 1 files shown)


unsleepy22:
@ashnazg 
since python 3.x is officially not fully compatible with 2.x, is it necessary to make this change? By making this PR, we may need to check all python files.

ashnazg:
I intentionally tested it against 2.x to be sure it was still ok, so that a larger scale effort would not necessarily be required.  I simply couldn't get my Win7 builds to run, due to the syntax breakdown as seen in the clj multilang-test.

ashnazg:
As an aside, I had tried to find a reference in the Storm documentation with regard to Python version requirements, but did not see any.  That's why I explored why this file was failing for me on 3.x.

ashnazg:
I'll have to rebase after #912 merges before this one will test good on Travis.

revans2:
@ashnazg the changes look good to me, and I would love to maintain both python2 and python3 compatibility.  I think the best way to do this is to have travis CI handle that for us.

https://docs.travis-ci.com/user/languages/python

Would you mind filing a JIRA to have travis run with different versions of python. Ideally it would be nice to configure it so we only run the multi-lang tests more than once, but that could be an optimization that comes later on.

Also it would be good to update to docs to indicate the versions of python supported.

I am +1 for this change.

ashnazg:
@revans2 , created STORM-1362 for it.  I'll take a stab at it, but someone more familiar with Travis might can knock it out faster.  I don't see an obvious "travis-ish" way to do this, so I'm guessing it'll require its own shellscript.

revans2:
@ashnazg Ya I think we are going to need to do something special to make it work too, but travis is the place we should be doing it if we want to be sure we don't break things in the future.

ashnazg:
Not sure how I had this succeeding previously, but now it consistently does not.  What I'm seeing now implies that only a full-on python3 adjustment over many files solve this.  I'll just have to drop my Win7 env down to python2.

PR #912, opened 2015-12-02
STORM-1361:  Apache License missing from two Cassandra files

ashnazg:


revans2:
+1, and since this is just a comment change I am not going to wait the full 24 hours to merge it in.  No point in keeping master broken.

PR #915, opened 2015-12-03
STORM-1211 Added trident Update/Query support for Cassandra connector.

satishd:
Trident state and query support for Cassandra.

### DIFF
--- a/external/storm-cassandra/pom.xml
+++ b/external/storm-cassandra/pom.xml
@@ -51,6 +51,15 @@
                 <role>developer</role>
             </roles>
         </developer>
+        <developer>
+            <id>satishd</id>
+            <name>Satish Duggana</name>
+            <email>satish.duggana@gmail.com</email>
+            <url>https://github.com/satishd</url>
+            <roles>
+                <role>developer</role>
+            </roles>
+        </developer>
     </developers>
 
     <build>
--- a/external/storm-cassandra/src/main/java/org/apache/storm/cassandra/trident/state/CassandraState.java
+++ b/external/storm-cassandra/src/main/java/org/apache/storm/cassandra/trident/state/CassandraState.java
@@ -1,0 +1,149 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.storm.cassandra.trident.state;
+
+import backtype.storm.task.IMetricsContext;
+import backtype.storm.topology.FailedException;
+import backtype.storm.tuple.Values;
+import com.datastax.driver.core.BatchStatement;
+import com.datastax.driver.core.Session;
+import com.datastax.driver.core.Statement;
+import com.google.common.base.Preconditions;
+import org.apache.storm.cassandra.client.SimpleClient;
+import org.apache.storm.cassandra.client.SimpleClientProvider;
+import org.apache.storm.cassandra.query.CQLResultSetValuesMapper;
+import org.apache.storm.cassandra.query.CQLStatementTupleMapper;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import storm.trident.operation.TridentCollector;
+import storm.trident.state.State;
+import storm.trident.tuple.TridentTuple;
+
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+/**
+ *
+ */
+public class CassandraState implements State {
+
+    private static final Logger LOG = LoggerFactory.getLogger(CassandraState.class);
+
+    private final Map conf;
+    private final Options options;
+
+    private Session session;
+    private SimpleClient client;
+
+    public CassandraState(Map conf, IMetricsContext metrics, int partitionIndex, int numPartitions, Options options) {
+        this.conf = conf;
+        this.options = options;
+    }
+
+    public static final class Options implements Serializable {
+        private final SimpleClientProvider clientProvider;
+        private CQLStatementTupleMapper cqlStatementTupleMapper;
+        private CQLResultSetValuesMapper cqlResultSetValuesMapper;
+        private BatchStatement.Type batchingType = BatchStatement.Type.LOGGED;
+
+
+        public Options(SimpleClientProvider clientProvider) {
+            this.clientProvider = clientProvider;
+        }
+
+        public Options withCQLStatementTupleMapper(CQLStatementTupleMapper cqlStatementTupleMapper) {
+            this.cqlStatementTupleMapper = cqlStatementTupleMapper;
+            return this;
+        }
+
+        public Options withCQLResultSetValuesMapper(CQLResultSetValuesMapper cqlResultSetValuesMapper) {
+            this.cqlResultSetValuesMapper = cqlResultSetValuesMapper;
+            return this;
+        }
+
+        public Options withBatching(BatchStatement.Type batchingType) {
+            this.batchingType = batchingType;
+            return this;
+        }
+
+    }
+
+    @Override
+    public void beginCommit(Long txid) {
+        LOG.debug("beginCommit is noop");
+    }
+
+    @Override
+    public void commit(Long txid) {
+        LOG.debug("commit is noop");
+    }
+
+    public void prepare() {
+        client = options.clientProvider.getClient(conf);
+        session = client.connect();
+    }
+
+    public void cleanup() {
+        session.close();
+        client.close();
+    }
+
+    public void updateState(List<TridentTuple> tuples, final TridentCollector collector) {
+        Preconditions.checkNotNull(options.cqlStatementTupleMapper, "CassandraState.Options should have cqlStatementTupleMapper");
+
+        List<Statement> statements = new ArrayList<>();
+        for (TridentTuple tuple : tuples) {
+            statements.addAll(options.cqlStatementTupleMapper.map(conf, session, tuple));
+        }
+        BatchStatement batchStatement = new BatchStatement(options.batchingType);
+        batchStatement.addAll(statements);
+
+        try {
+            session.execute(batchStatement);
+        } catch (Exception e) {
+            LOG.warn("Batch write operation is failed.");
+            collector.reportError(e);
+            throw new FailedException(e);
+        }
+
+    }
+
+    public List<List<Values>> batchRetrieve(List<TridentTuple> tridentTuples) {
+        Preconditions.checkNotNull(options.cqlStatementTupleMapper, "CassandraState.Options should have cqlStatementTupleMapper");
+        Preconditions.checkNotNull(options.cqlResultSetValuesMapper, "CassandraState.Options should have cqlResultSetValuesMapper");
+
+        List<List<Values>> batchRetrieveResult = new ArrayList<>();
+        try {
+            for (TridentTuple tridentTuple : tridentTuples) {
+                List<Statement> statements = options.cqlStatementTupleMapper.map(conf, session, tridentTuple);
+                for (Statement statement : statements) {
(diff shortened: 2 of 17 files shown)


external/storm-cassandra/pom.xml
```diff
@@ -51,6 +51,15 @@
                 <role>developer</role>
             </roles>
         </developer>
+        <developer>
+            <id>satishd</id>
```
harshach: developer tag shouldn't be there for apache I think. can you remove this.

external/storm-cassandra/src/main/java/org/apache/storm/cassandra/trident/state/CassandraState.java
```diff
@@ -0,0 +1,149 @@
...
+        session = client.connect();
+    }
+
+    public void cleanup() {
+        session.close();
+        client.close();
+    }
+
+    public void updateState(List<TridentTuple> tuples, final TridentCollector collector) {
+        Preconditions.checkNotNull(options.cqlStatementTupleMapper, "CassandraState.Options should have cqlStatementTupleMapper");
```
harshach: do we need to check this in updateState can't we do this in prepare()

external/storm-cassandra/src/main/java/org/apache/storm/cassandra/trident/state/CassandraState.java
```diff
@@ -0,0 +1,149 @@
...
+        } catch (Exception e) {
+            LOG.warn("Batch write operation is failed.");
+            collector.reportError(e);
+            throw new FailedException(e);
+        }
+
+    }
+
+    public List<List<Values>> batchRetrieve(List<TridentTuple> tridentTuples) {
+        Preconditions.checkNotNull(options.cqlStatementTupleMapper, "CassandraState.Options should have cqlStatementTupleMapper");
```
harshach: same as above

external/storm-cassandra/pom.xml
harshach: my bad this is a common practice , thanks for the correction @satishd

external/storm-cassandra/src/main/java/org/apache/storm/cassandra/trident/state/CassandraState.java
```diff
@@ -0,0 +1,149 @@
...
+
+    private static final Logger LOG = LoggerFactory.getLogger(CassandraState.class);
+
+    private final Map conf;
+    private final Options options;
+
+    private Session session;
+    private SimpleClient client;
+
+    public CassandraState(Map conf, IMetricsContext metrics, int partitionIndex, int numPartitions, Options options) {
```
hmcl: It seems that not all constructor parameters are referred. Is it  intended?

external/storm-cassandra/src/main/java/org/apache/storm/cassandra/trident/state/CassandraState.java
```diff
@@ -0,0 +1,149 @@
...
+        public Options withBatching(BatchStatement.Type batchingType) {
+            this.batchingType = batchingType;
+            return this;
+        }
+
+    }
+
+    @Override
+    public void beginCommit(Long txid) {
+        LOG.debug("beginCommit is noop");
```
hmcl: Do you mean no operation? Perhaps the message could avoid acronym.

external/storm-cassandra/src/main/java/org/apache/storm/cassandra/trident/state/CassandraState.java
```diff
@@ -0,0 +1,149 @@
...
+
+    public void prepare() {
+        Preconditions.checkNotNull(options.cqlStatementTupleMapper, "CassandraState.Options should have cqlStatementTupleMapper");
+
+        client = options.clientProvider.getClient(conf);
+        session = client.connect();
+    }
+
+    public void cleanup() {
+        session.close();
```
hmcl: If for a reason an exception occurs here the client won't get closed.

external/storm-cassandra/src/test/java/org/apache/storm/cassandra/DynamicStatementBuilderTest.java
```diff
@@ -116,15 +124,15 @@ private void executeStatementAndAssert(CQLStatementTupleMapper mapper, String...
...
             Assert.assertEquals(listExpected.get(i), map.get(i).toString());
         }
     }
 
     @Test
     public void shouldBuildStaticBoundStatement() {
-        CQLStatementTupleMapper mapper = boundQuery("INSERT INTO weather.temperature(weatherstation_id, event_time, temperature) VALUES(?, ?, ?)")
-                .bind(with(field("weatherstation_id"), field("event_time").now(), field("temperature")));
+        CQLStatementTupleMapper mapper = boundQuery("INSERT INTO weather.temperature(weather_station_id, event_time, temperature) VALUES(?, ?, ?)")
+                .bind(with(field("weather_station_id"), field("event_time").now(), field("temperature")));
```
hmcl: If this constants are expected to be used consistently, perhaps when expanding test cases, we could move them into pub stat fields.

external/storm-cassandra/src/test/java/org/apache/storm/cassandra/trident/TridentTopologyTest.java
```diff
@@ -0,0 +1,125 @@
...
+        CassandraState.Options options = new CassandraState.Options(new CassandraContext());
+        CQLStatementTupleMapper insertTemperatureValues = boundQuery(
+                "INSERT INTO weather.temperature(weather_station_id, weather_station_name, event_time, temperature) VALUES(?, ?, ?, ?)")
+                .bind(with(field("weather_station_id"), field("name").as("weather_station_name"), field("event_time").now(), field("temperature")));
+        options.withCQLStatementTupleMapper(insertTemperatureValues);
+        return new CassandraStateFactory(options);
+    }
+
+    public CassandraStateFactory getSelectWeatherStationStateFactory() {
+        CassandraState.Options options = new CassandraState.Options(new CassandraContext());
```
hmcl: perhaps rename to create... / build...?

external/storm-cassandra/src/test/java/org/apache/storm/cassandra/trident/TridentTopologyTest.java
```diff
@@ -0,0 +1,125 @@
...
+        @Override
+        public void execute(TridentTuple tuple, TridentCollector collector) {
+            System.out.println("####### tuple = " + tuple.getValues());
+            collector.emit(tuple.getValues());
+        }
+    }
+
+    private CassandraStateFactory getInsertTemperatureStateFactory() {
+        CassandraState.Options options = new CassandraState.Options(new CassandraContext());
+        CQLStatementTupleMapper insertTemperatureValues = boundQuery(
```
hmcl: perhaps rename to create... / build...?

external/storm-cassandra/src/main/java/org/apache/storm/cassandra/trident/state/TridentResultSetValuesMapper.java
```diff
@@ -0,0 +1,63 @@
...
+    private Fields outputDeclaredFields;
+
+    public TridentResultSetValuesMapper(Fields outputDeclaredFields) {
+        this.outputDeclaredFields = outputDeclaredFields;
+    }
+
+    @Override
+    public List<List<Values>> map(Session session, Statement statement, ITuple tuple) {
+        List<List<Values>> list = new ArrayList<>();
+        ResultSet resultSet = session.execute(statement);
```
hmcl: can resultset ever be null? Do we need null check?

external/storm-cassandra/src/test/java/org/apache/storm/cassandra/trident/TridentTopologyTest.java
satishd: Cleaning up these test cases is part of another JIRA STORM-1348. The above changes are done as part of refactoring the cassandra table script.

external/storm-cassandra/src/main/java/org/apache/storm/cassandra/trident/state/TridentResultSetValuesMapper.java
satishd: Pl check the javadoc of Session which says "@return the result of the query. That result will never be null but can be empty"

external/storm-cassandra/src/main/java/org/apache/storm/cassandra/trident/state/CassandraState.java
```diff
@@ -0,0 +1,149 @@
...
+        client.close();
+    }
+
+    public void updateState(List<TridentTuple> tuples, final TridentCollector collector) {
```
fhussonnois: I think async writes should be used to update state. In cassandra batch statements must not be used to improved write performance. In addition, Cassandra will warn if the batch size is too large and this may lead to an overloading of the cluster. Batches must only be used to make few writes atomic.

satishd: The reason for using sync write is if it fails we want these Exceptions to be thrown back and retried like the way JDBC connector does. If we do not do batching then update could be partial when some of those statements fail to execute.

arunmahadevan:
You need to provide a `MapState` implementation as well for this to be used with `groupby`. Most of the other connectors have a MapState implementation.

external/storm-cassandra/src/main/java/org/apache/storm/cassandra/trident/state/CassandraState.java
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    @Override
+    public void beginCommit(Long txid) {
+        LOG.debug("beginCommit is noop");
+    }
+
+    @Override
+    public void commit(Long txid) {
+        LOG.debug("commit is noop");
```
arunmahadevan: If you ignore beginCommit and commit it will not provide exactly once semantics and may end up with duplicate records in the db.

external/storm-cassandra/src/main/java/org/apache/storm/cassandra/trident/state/CassandraState.java
fhussonnois: Ok I understand your point but this may lead to some issues at scale. Usually writes are idempotent in cassandra. So even if a write fail you could resend all tuples. Could we imagine two strategies to update the state ?

external/storm-cassandra/src/main/java/org/apache/storm/cassandra/trident/state/CassandraState.java
satishd: It supports atleastonce for now. Most of the connectors do not support exactlyonce except HDFS.

satishd:
Filed https://issues.apache.org/jira/browse/STORM-1369 for implementing MapState for Cassandra. This PR is about making Update and Query support.

external/storm-cassandra/src/main/java/org/apache/storm/cassandra/trident/state/CassandraState.java
satishd: OK. I agree to have an option for not to use batching.

external/storm-cassandra/src/main/java/org/apache/storm/cassandra/trident/state/CassandraState.java
arunmahadevan: The other connectors also provide a Mapstate implementation which gives exactly once. I am not sure why the other connectors ignore the beginCommit and commit in their `State` implementation for trident and if its desirable. May be that should also get fixed.

satishd: It seems storm.trident.state.map.TransactionalMap provides the semantics.  IBackingMap implementation can be wrapped into that. HBase and Redis use these APIs for having transactional semantics.

external/storm-cassandra/src/test/java/org/apache/storm/cassandra/trident/TridentTopologyTest.java
satishd: Cleaning up these test cases is part of another JIRA STORM-1348. The above changes are done as part of refactoring the cassandra table script.

external/storm-cassandra/src/test/java/org/apache/storm/cassandra/DynamicStatementBuilderTest.java
satishd: Cleaning up these test cases is part of another JIRA STORM-1348. The above changes are done as part of refactoring the cassandra table script.

external/storm-cassandra/src/main/java/org/apache/storm/cassandra/trident/state/CassandraState.java
satishd: @fhussonnois By default, it does not use batching but it can be enabled by using the API Options#withBatching(BatchStatement.Type batchingType).

harshach:
+1

harshach:
Thanks @satishd merged into master.

PR #917, opened 2015-12-03
STORM-1362 Travis should run at least the multilang-test using Python2 and 3

ashnazg:
Here's how I was able to get the multilang-test to be run separately with python3 in travis.


ashnazg:
Putting this here for future reference only, because my testing shows that a larger python3 migration effort will be needed to really make this build cleanly.  This PR is just to record how I was able to get Travis to run the test using python3 when the container-based VM also has python2 as its default python installation.

PR #900, opened 2015-11-23
STORM-1187 Support windowing based on tuple ts

arunmahadevan:
Support for doing window calculations based to tuple timestamps and handle
out of order events based on time lag.

PR #925, opened 2015-12-04
STORM-1374 fix random failure on WindowManagerTest

HeartSaVioR:
* Problem: Trigger could be called two times while sleeping
  * which first one has the events which are actually we want
  * and second one resets it, which makes test failed
* Solution: Trigger manually
  * and remove sleep if not necessary

@arunmahadevan Please verify this shouldn't break your intention on WindowManagerTest.

### DIFF
--- a/storm-core/test/jvm/backtype/storm/windowing/WindowManagerTest.java
+++ b/storm-core/test/jvm/backtype/storm/windowing/WindowManagerTest.java
@@ -137,7 +137,10 @@ public class WindowManagerTest {
     @Test
     public void testTimeBasedWindow() throws Exception {
         windowManager.setWindowLength(new Duration(1, TimeUnit.SECONDS));
-        windowManager.setSlidingInterval(new Duration(100, TimeUnit.MILLISECONDS));
+        // don't rely on TimeTriggerPolicy since that easily lead timing-issue test failure
+        // set this to never triggered, and trigger manually
+        windowManager.setSlidingInterval(new Duration(1, TimeUnit.DAYS));
+
         long now = System.currentTimeMillis();
 
         // add with past ts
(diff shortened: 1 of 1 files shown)


arunmahadevan:
@HeartSaVioR thanks for providing the patch. I had overlooked the edge cases where the TimeTrigger might not fire or fires twice causing tests to fail.
The patch you have provided addresses the issue. Only concern is that the existing TimeEvictionPolicy uses System.currentTimeMillis() to evict the events so it relies on the passage of time to expire events and if you manually invoke the trigger, the expected events may not expire.

However I verified that the tests pass with your changes so it can be merged to fix the random test failures.

In https://github.com/apache/storm/pull/900 an api is added to set the reference timestamp for the eviction policy instead of it relying on system timestamp. With that PR I will fix the unit tests by setting the future timestamp to simulate passage of time and invoke the trigger manually.

HeartSaVioR:
@arunmahadevan 
Yes, actually I don't remove sleep at some places since we still need to wait for eviction though we trigger manually.

arunmahadevan:
@HeartSaVioR so I think it will cause trigger to fire two times (during sleep and then manually). For the current test cases it does not matter so your patch would work.

I updated the unit tests as a part of PR#900 to not rely on sleep and trigger manually.
https://github.com/apache/storm/pull/900/files#diff-04ed7847355c5e160efb25a9a5dc8a25R157

HeartSaVioR:
@arunmahadevan 
Yes, it requires changing of trigger policy to never fire.
I'll take a look at #900 and discard this PR when #900 covers this.

d2r:
#900 is a change of much larger scope, and looking at it just now, there are two other new problems preventing tests passing successfully.  Would it be better to merge this in ahead of #900, given that it is a change of much smaller scope, rather than making the test fix wait on other unrelated changes?

arunmahadevan:
@d2r this can get merged. 
I have fixed the issues as a part of PR#900 and will resolve the conflicts once this gets merged.

revans2:
+1 this looks god to me.  Could someone also file a JIRA to use simulated time instead of sleeping for these tests.

HeartSaVioR:
@revans2 
Seems like #900 already got it. @arunmahadevan removes all places which calls Thread.sleep().
Btw, we may want to take some time to review #900 since it is not small patch, so though #900 now covers #925, I think we'd better to get #925 merged.

PR #928, opened 2015-12-07
STORM-1378:  Use File.separator for OS independence in Blobstore LocalizerTest

ashnazg:
This clears up one incidental issue shown on STORM-1378, the one test failure due solely to "/" vs "\" in the path comparison.  This does _not_ solve the larger issue on STORM-1378.

### DIFF
--- a/storm-core/src/jvm/backtype/storm/blobstore/BlobStore.java
+++ b/storm-core/src/jvm/backtype/storm/blobstore/BlobStore.java
@@ -298,7 +298,9 @@ public abstract class BlobStore implements Shutdownable {
     public byte[] readBlob(String key, Subject who) throws IOException, KeyNotFoundException, AuthorizationException {
         ByteArrayOutputStream out = new ByteArrayOutputStream();
         readBlobTo(key, out, who);
-        return out.toByteArray();
+        byte[] bytes = out.toByteArray();
+        out.close();
+        return bytes;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


redsanket:
+1 NB

wuchong:
+1 nice catch

unsleepy22:
I'd prefer to create a util method like `joinPath` which is like:
```
public static String joinPath(String... pathList) {
    return Joiner.on(File.separator).join(pathList);
}
```
to avoid so many duplicate `File.separator`s.

revans2:
The change looks god +1.  I agree with @unsleepy22 though and a joinPath would make the code more readable, but it is optional in my mind.

ashnazg:
Umm... is Joiner only available in Java8?

unsleepy22:
@ashnazg use Joiner in guava (should be already in dependency) :)

revans2:
@ashnazg http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/base/Joiner.html

ashnazg:
Ok, let me doublecheck my build... the dep wasn't found, which is why I thought "surely I don't need to add a new dep for this" :)

revans2:
@ashnazg adding a new dep in this case is not that big of a deal, because it is only for a test.  Just be sure to mark the dep as a test dependency in maven if you do need to add a new one.

ashnazg:
Ahhh... Eclipse gave me the wrong import (org.apache.storm.guava.base)...

unsleepy22:
looks good to me now~

ashnazg:
These last additions are the best that can be done.  The symlink issue can't be solved (see JIRA).

revans2:
+1

PR #933, opened 2015-12-08
[STORM-1376] Performance slowdown due excessive zk connections and log-debugging

redsanket:
Uneccessary creation of client and closing of connections puts load on zookeeper.
Having a single client connection to perform reads helps

### DIFF
--- a/conf/defaults.yaml
+++ b/conf/defaults.yaml
@@ -179,7 +179,7 @@ task.refresh.poll.secs: 10
 task.credentials.poll.secs: 30
 
 # now should be null by default
-topology.backpressure.enable: false
+topology.backpressure.enable: true
 backpressure.disruptor.high.watermark: 0.9
 backpressure.disruptor.low.watermark: 0.4
 
--- a/conf/defaults.yaml
+++ b/conf/defaults.yaml
@@ -179,7 +179,7 @@ task.refresh.poll.secs: 10
 task.credentials.poll.secs: 30
 
 # now should be null by default
-topology.backpressure.enable: true
+topology.backpressure.enable: false
 backpressure.disruptor.high.watermark: 0.9
 backpressure.disruptor.low.watermark: 0.4
 
--- a/storm-core/test/clj/backtype/storm/nimbus_test.clj
+++ b/storm-core/test/clj/backtype/storm/nimbus_test.clj
@@ -1238,10 +1238,11 @@
   (testing "nimbus-data uses correct ACLs"
     (let [scheme "digest"
           digest "storm:thisisapoorpassword"
-          auth-conf {STORM-ZOOKEEPER-AUTH-SCHEME scheme
+          auth-conf (merge (read-storm-config)
+                    {STORM-ZOOKEEPER-AUTH-SCHEME scheme
                      STORM-ZOOKEEPER-AUTH-PAYLOAD digest
                      STORM-PRINCIPAL-TO-LOCAL-PLUGIN "backtype.storm.security.auth.DefaultPrincipalToLocal"
-                     NIMBUS-THRIFT-PORT 6666}
+                     NIMBUS-THRIFT-PORT 6666})
           expected-acls nimbus/NIMBUS-ZK-ACLS
           fake-inimbus (reify INimbus (getForcedScheduler [this] nil))]
       (stubbing [nimbus-topo-history-state nil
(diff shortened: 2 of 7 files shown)


revans2:
@redsanket did you reproduce the original issue?  Do we need any synchronization around zk client now that we don't create a new one each time?

redsanket:
@revans2 I have launched around 5 topologies on 5 supervisor nodes and could see lot of connections being made to zookeeper and event processing slowing down. But I could not observe zookeeper slowing down to a point it could not respond. I will launch more topologies to make sure it really slows down. I realized in this scenario that it is only performing reads and as per zookeeper implementation I presume connections are thread safe and hence one connection to read information would be sufficient I guess. An interesting point came up with synchronization here, I looked at the localizer code and I have seen downloads to the supervisor are performed by locking but updates are not therefore to support nimbus HA checkForUpdate and checkForDownload might have to be synchronized. Will make sure by running it with more topologies. Thanks for the review

redsanket:
@revans2 Would Automatic backpressure make a difference, as it is turned on now by default and do we need to use pace maker along with it as I guess this change improves to remove the number of connections and speed up a little but on the whole it is not doing good when compared to 0.10.x branch when automatic backpressure is on and running with no pace maker

redsanket:
STORM-1215 has done some changes to Async Logging, I removed those commits and tested it again and now 0.10.x and 0.11 current version are doing similar

conf/defaults.yaml
```diff
@@ -179,7 +179,7 @@ task.refresh.poll.secs: 10
 task.credentials.poll.secs: 30
 
 # now should be null by default
-topology.backpressure.enable: true
+topology.backpressure.enable: false
```
revans2: Do we want this off by default?

storm-core/test/clj/backtype/storm/nimbus_test.clj
```diff
@@ -1238,10 +1238,11 @@
   (testing "nimbus-data uses correct ACLs"
     (let [scheme "digest"
           digest "storm:thisisapoorpassword"
-          auth-conf {STORM-ZOOKEEPER-AUTH-SCHEME scheme
+          auth-conf (merge (read-storm-config)
+                    {STORM-ZOOKEEPER-AUTH-SCHEME scheme
                      STORM-ZOOKEEPER-AUTH-PAYLOAD digest
                      STORM-PRINCIPAL-TO-LOCAL-PLUGIN "backtype.storm.security.auth.DefaultPrincipalToLocal"
-                     NIMBUS-THRIFT-PORT 6666}
+                     NIMBUS-THRIFT-PORT 6666})
```
ppoulosk: I know that the code was already there, but is it possible to use an ephemeral port here?

revans2:
Overall the patch looks good, just curious why we turned off backpressure by default.

conf/defaults.yaml
```diff
@@ -179,7 +179,7 @@ task.refresh.poll.secs: 10
...
 backpressure.disruptor.high.watermark: 0.9
 backpressure.disruptor.low.watermark: 0.4
```
rfarivar: How about adding a comment describing why the choice of 0.4 and 0.9?

storm-core/test/clj/backtype/storm/nimbus_test.clj
rfarivar: Ditto.

conf/defaults.yaml
redsanket: Oh Im sorry I was testing few scenarios might have come along with my last commit. Will change it back

hustfxj:
yes, others are good, but why we turned off backpressure by default.

conf/defaults.yaml
zhuoliu: This low/high watermark default values are now more of empirical values based on a limited set of topologies. We may need more experiments and theoretical analysis (based on representative queue draining speed, callback response time etc.) for choices of these watermark percentages.

storm-core/test/clj/backtype/storm/nimbus_test.clj
redsanket: All the test cases in this file have a similar port, will create a JIRA to tackle this issue separately as I feel it is not a part of this change

revans2:
+1

ppoulosk:
@redsanket, Agreed.  thanks for filing the Jira.  +1, NB

knusbaum:
+1

PR #934, opened 2015-12-08
[STORM-1373] Blobstore API sample example usage

redsanket:


examples/storm-starter/src/jvm/storm/starter/BlobStoreAPIWordCountTopology.java
```diff
@@ -0,0 +1,248 @@
...
+    private static String key = "key1";
+    private static final Logger LOG = LoggerFactory.getLogger(BlobStoreAPIWordCountTopology.class);
+    private static final int READ = 0x01;
+    private static final int WRITE = 0x02;
+    private static final int ADMIN = 0x04;
+    private static final List<AccessControl> WORLD_EVERYTHING =
+            Arrays.asList(new AccessControl(AccessControlType.OTHER, READ | WRITE | ADMIN));
+
+    // Spout implementation
+    public static class RandomSentenceSpout extends BaseRichSpout {
```
ptgoetz: You may want to consider renaming this to `BlobStoreSpout` since it just emits the contents of a blob entry (the random logic is elsewhere, i.e. not in the spout).

examples/storm-starter/src/jvm/storm/starter/BlobStoreAPIWordCountTopology.java
```diff
@@ -0,0 +1,248 @@
...
+        public void ack(Object id) {
+        }
+
+        @Override
+        public void fail(Object id) {
+        }
+
+        @Override
+        public void declareOutputFields(OutputFieldsDeclarer declarer) {
+            declarer.declare(new Fields("word"));
```
ptgoetz: This should probably be "sentence" instead of "word".

examples/storm-starter/src/jvm/storm/starter/BlobStoreAPIWordCountTopology.java
```diff
@@ -0,0 +1,248 @@
...
+            LOG.info(getBlobContent(key, store));
+            // Launching word count topology with blob updates
+            wc.buildAndLaunchWordCountTopology(args);
+            // Updating blobs few times
+            for (int i = 0; i < 10; i++) {
+                updateBlobWithContent(key, store);
+                Utils.sleep(100);
+            }
+        } catch (KeyAlreadyExistsException kae) {
+            // Do nothing
```
ptgoetz: We should probably at least log the exception here.

examples/storm-starter/src/jvm/storm/starter/BlobStoreAPIWordCountTopology.java
```diff
@@ -0,0 +1,248 @@
...
+import java.util.Map;
+import java.util.Random;
+
+public class BlobStoreAPIWordCountTopology {
+    private static NimbusBlobStore store = new NimbusBlobStore(); // Client API to invoke blob store API functionality
+    private static String key = "key1";
+    private static final Logger LOG = LoggerFactory.getLogger(BlobStoreAPIWordCountTopology.class);
+    private static final int READ = 0x01;
+    private static final int WRITE = 0x02;
+    private static final int ADMIN = 0x04;
```
ptgoetz: Shouldn't these constants be part of the blobstore API?

examples/storm-starter/src/jvm/storm/starter/BlobStoreAPIWordCountTopology.java
redsanket: I guess you meant to say the spout part to be @Override
+        public void declareOutputFields(OutputFieldsDeclarer declarer) {
+            declarer.declare(new Fields("sentence"));
+        } rather than bolt which splits the sentence. Made the changes thanks for the catch

examples/storm-starter/src/jvm/storm/starter/BlobStoreAPIWordCountTopology.java
```diff
@@ -0,0 +1,246 @@
...
+                Thread.sleep(10000);
+
+                cluster.shutdown();
+            }
+        } catch (InvalidTopologyException | AuthorizationException | AlreadyAliveException | InterruptedException exp) {
+            throw new RuntimeException(exp);
+        }
+    }
+
+    private static void createBlobWithContent(String blobKey, ClientBlobStore clientBlobStore, SettableBlobMeta settableBlobMeta)
```
ppoulosk: Should these helper routines be in some Util class?

ppoulosk:
+1 NB

examples/storm-starter/src/jvm/storm/starter/BlobStoreAPIWordCountTopology.java
```diff
@@ -0,0 +1,246 @@
...
+        @Override
+        public void declareOutputFields(OutputFieldsDeclarer declarer) {
+            declarer.declare(new Fields("word", "count"));
+        }
+    }
+
+    public void buildAndLaunchWordCountTopology(String[] args) {
+
+        TopologyBuilder builder = new TopologyBuilder();
+
```
rfarivar: remove the empty line?

examples/storm-starter/src/jvm/storm/starter/BlobStoreAPIWordCountTopology.java
```diff
@@ -0,0 +1,246 @@
...
+        builder.setSpout("spout", new BlobStoreSpout(), 5);
+
```
rfarivar: remove the empty line?

examples/storm-starter/src/jvm/storm/starter/BlobStoreAPIWordCountTopology.java
```diff
@@ -0,0 +1,246 @@
...
+                conf.setNumWorkers(3);
+                StormSubmitter.submitTopologyWithProgressBar(args[0], conf, builder.createTopology());
+            } else {
+                conf.setMaxTaskParallelism(3);
+
+                LocalCluster cluster = new LocalCluster();
+                cluster.submitTopology("word-count", conf, builder.createTopology());
+
+                Thread.sleep(10000);
+
```
rfarivar: remove the empty lines (a couple lines above too)?

revans2:
@redsanket I don't like this example.   I really would prefer to see an example that is realistic to how we want users to interact with the blob storm.  A blob store spout is something I would discourage people from using.

Could we do something like a blacklist file that is added to the blob store at the beginning of the run.  Then a filter bolt would read in the file every n seconds from off of disk and throw away any words/tuples that appeared in the file.  In the middle of the example we can have code that updates the blob, like you have to show the blacklist changing part way through.

In the examples that interact with the blob store we could show the command line that would do the same thing, so they don't have to write code if they don't want to.

redsanket:
@revans2 Will rewrite it as suggested. I agree the example should look more like what you have suggested, I just thought more on terms of just accessing the API interface

examples/storm-starter/src/jvm/storm/starter/BlobStoreAPIWordCountTopology.java
```diff
@@ -0,0 +1,246 @@
...
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Random;
+
+public class BlobStoreAPIWordCountTopology {
+    private static NimbusBlobStore store = new NimbusBlobStore(); // Client API to invoke blob store API functionality
```
unsleepy22: store -> clientStore?

redsanket:
@revans2 concerns addressed

examples/storm-starter/src/jvm/storm/starter/BlobStoreAPIWordCountTopology.java
```diff
@@ -0,0 +1,288 @@
...
+        }
+
+        @Override
+        public Map<String, Object> getComponentConfiguration() {
+            return null;
+        }
+    }
+
+    public static class FilterWords extends BaseBasicBolt {
+        String fileName = "blacklist.txt";
```
jerrypeng: can we just re use the fileName variable defined at the top?

examples/storm-starter/src/jvm/storm/starter/BlobStoreAPIWordCountTopology.java
```diff
@@ -0,0 +1,288 @@
...
+        File file = new File(fileName);
+        Set<String> wordSet = new HashSet<>();
+        if (!file.exists()) {
+            return wordSet;
+        }
+        StringTokenizer tokens = new StringTokenizer(readFile(file).toString(), "\r\n");
+        while (tokens.hasMoreElements()) {
+            wordSet.add(tokens.nextToken());
+        }
+        LOG.info("parseFile {}", wordSet);
```
jerrypeng: perhaps debug?

examples/storm-starter/src/jvm/storm/starter/BlobStoreAPIWordCountTopology.java
```diff
@@ -0,0 +1,288 @@
...
+        return wordSet;
+    }
+
+    private static StringBuilder readFile(File file) throws IOException {
+        String line;
+        StringBuilder fileContent = new StringBuilder();
+        // Do not use canonical file name here
```
jerrypeng: I don't quite understand how the file is being found when providing the filename and not the path.  Can you please provide me with some clarification.  Thanks!

examples/storm-starter/src/jvm/storm/starter/BlobStoreAPIWordCountTopology.java
```diff
@@ -0,0 +1,288 @@
...
+        TopologyBuilder builder = new TopologyBuilder();
+        builder.setSpout("spout", new RandomSentenceSpout(), 5);
+        builder.setBolt("split", new SplitSentence(), 8).shuffleGrouping("spout");
+        builder.setBolt("filter", new FilterWords(), 6).shuffleGrouping("split");
+
+        Config conf = new Config();
+        conf.setDebug(true);
+        try {
+            conf.setNumWorkers(3);
+            StormSubmitter.submitTopologyWithProgressBar(args[0], conf, builder.createTopology());
```
revans2: Can we have a default topology name if args[0] is not set?

examples/storm-starter/src/jvm/storm/starter/BlobStoreAPIWordCountTopology.java
```diff
@@ -0,0 +1,288 @@
...
+        }
+    }
+
+    public static class FilterWords extends BaseBasicBolt {
+        String fileName = "blacklist.txt";
+        @Override
+        public void execute(Tuple tuple, BasicOutputCollector collector) {
+            try {
+                String word = tuple.getString(0);
+                Set<String> wordSet = parseFile(fileName);
```
revans2: Can we please cache the wordSet and only update it once every 5 seconds?

examples/storm-starter/src/jvm/storm/starter/BlobStoreAPIWordCountTopology.java
```diff
@@ -0,0 +1,288 @@
...
+        bw.close();
+    }
+
+    public static void main(String[] args) {
+        prepare();
+        BlobStoreAPIWordCountTopology wc = new BlobStoreAPIWordCountTopology();
+        try {
+            File file = createFile(fileName);
+            // Creating blob again before launching topology
+            createBlobWithContent(key, store, file);
```
revans2: Are we going to get errors if we run this again?  The key will already exist.  Would it be good to delete the blob at the end of the example so others can also run the example?

revans2:
Looks good to me,  Just a few minor comments.

examples/storm-starter/src/jvm/storm/starter/BlobStoreAPIWordCountTopology.java
redsanket: Yes I am catching the keyAlreadyExists exception, I might well have it deleted but the topology will run with no filtering

examples/storm-starter/src/jvm/storm/starter/BlobStoreAPIWordCountTopology.java
redsanket: I have mentioned how we launch the topology in main, as we want to include topology.blobstore.map config

redsanket:
@revans2 and @jerrypeng Comments Addressed

jerrypeng:
LGTM +1

examples/storm-starter/src/jvm/storm/starter/BlobStoreAPIWordCountTopology.java
revans2: OK

revans2:
+1

PR #936, opened 2015-12-09
STORM-1199 : HDFS Spout Functionally complete. Not well tested. Some UTs

roshannaik:
HDFS Spout  for initial review (no Trident support).

**Features:**
 - Monitors directory. Picks up new files as they show up.
 - Built-in  support for consuming Text and Sequence File formats 
 - Extensible to other file formats. 
 - Works with or without ACKers.
 - Concurrent consumption: Each spout instance consumes a different file in parallel. 
 - Zookeeper free synchronization: HDFS based synchronization mechanism enables each spout to concurrently decide which of the available files to consume next. Spout acquires a lock on a file before consuming it. The lock files are stored in a .lock subdirectory. Lock file is deleted once file is consumed.
 - Archiving: Completed files are moved to a (configurable) 'done' folder 
 - Easy progress tracking: Files being consumed are renamed with a '.inprogress' suffix allowing users to easily check which files are being processed by simply listing the dir. More more detailed info users can rely on dashboard metrics.
 - Read model : Buffered reads & lazy tuple parsing. Reads are buffered, but parsing out a tuple is delayed until tuple is needed. Avoids periodic latency spikes involved with 'eager' parsing (e.g Kafka spout) of the entire  batch of records. 
 - Sliding window model used to track outstanding unACKed tuples (instead of Tumbling Window model used in Kafka spout). Tumbling window model prevents new tuple generation even if one tuple in the batch is not acked.
 - Support for clusters where NTP is not used to synchronize machine clocks (for identifying timed out locks - see *spout failure detection* section below). 
 - Kerberos support

**Fault Tolerance :**
 - Un-parseable files - are moved to a  separate (configurable) directory
 - I/O errors - Retry later if currently experiencing I/O errors during read
 - Duplicate Delivery mitigation
     + Spout tracks its progress on HDFS. So if it dies, another spout can resume from where it left off, rather than reprocessing the whole file.
     + Configurable number of outstanding unACKed tuples - helps contain the amount of duplicate emits when a spout takes over the job of a failed spout.
 - Spout Failure and Failover :
  + Spout failure detection - Each spout acquires a lock on the file it is reading and periodically heartbeats (appends) its progress into the lock file with a timestamp. This helps identify stale/abandoned locks easily (based on configurable lock timeout).  Spouts periodically check for stale locks.
  + Failover to another spout - A spout can take ownership of a stale lock and resume reading the file corresponding to the lock. The latest progress noted in the lock file is used to decide the location from which to resume.

### DIFF
--- a/external/storm-hdfs/README.md
+++ b/external/storm-hdfs/README.md
@@ -452,9 +452,9 @@ before selecting the next file for consumption.
 **Lock on *.lock* Directory**
 Hdfs spout instances create a *DIRLOCK* file in the .lock directory to co-ordinate certain accesses to 
 the .lock dir itself. A spout will try to create it when it needs access to the .lock directory and
-then delete it when done.  In case of a topology crash or force kill, this file may not get deleted.
-In this case it should be deleted manually to allow the new topology instance to regain  full access 
-to the  .lock  directory and resume normal processing. 
+then delete it when done.  In error conditions such as a topology crash, force kill or untimely death 
+of a spout, this file may not get deleted. Future running instances of the spout will eventually recover
+this once the DIRLOCK file becomes stale due to inactivity for hdfsspout.lock.timeout.sec seconds.
 
 ## Usage
 
@@ -515,13 +515,13 @@ Only settings mentioned in **bold** are required.
 |**hdfsspout.source.dir**      |             | HDFS location from where to read.  E.g. /data/inputfiles  |
 |**hdfsspout.archive.dir**     |             | After a file is processed completely it will be moved to this directory. E.g. /data/done|
 |**hdfsspout.badfiles.dir**    |             | if there is an error parsing a file's contents, the file is moved to this location.  E.g. /data/badfiles  |
-|hdfsspout.lock.dir            | '.lock' subdirectory under hdfsspout.source.dir | Dir in which lock files will be created. Concurrent HDFS spout instances synchronize using *lock* files. Before processing a file the spout instance creates a lock file in this directory with same name as input file and deletes this lock file after processing the file. Spout also periodically makes a note of its progress (wrt reading the input file) in the lock file so that another spout instance can resume progress on the same file if the spout dies for any reason. When a toplogy is killed, if a .lock/DIRLOCK file is left behind it can be safely deleted to allow normal resumption of the topology on restart.|
+|hdfsspout.lock.dir            | '.lock' subdirectory under hdfsspout.source.dir | Dir in which lock files will be created. Concurrent HDFS spout instances synchronize using *lock* files. Before processing a file the spout instance creates a lock file in this directory with same name as input file and deletes this lock file after processing the file. Spouts also periodically makes a note of their progress (wrt reading the input file) in the lock file so that another spout instance can resume progress on the same file if the spout dies for any reason.|
 |hdfsspout.ignore.suffix       |   .ignore   | File names with this suffix in the in the hdfsspout.source.dir location will not be processed|
 |hdfsspout.commit.count        |    20000    | Record progress in the lock file after these many records are processed. If set to 0, this criterion will not be used. |
 |hdfsspout.commit.sec          |    10       | Record progress in the lock file after these many seconds have elapsed. Must be greater than 0 |
 |hdfsspout.max.outstanding     |   10000     | Limits the number of unACKed tuples by pausing tuple generation (if ACKers are used in the topology) |
 |hdfsspout.lock.timeout.sec    |  5 minutes  | Duration of inactivity after which a lock file is considered to be abandoned and ready for another spout to take ownership |
-|hdfsspout.clocks.insync       |    true     | Indicates whether clocks on the storm machines are in sync (using services like NTP)       |
+|hdfsspout.clocks.insync       |    true     | Indicates whether clocks on the storm machines are in sync (using services like NTP). Used for detecting stale locks. |
 |hdfs.config (unless changed)  |             | Set it to a Map of Key/value pairs indicating the HDFS settigns to be used. For example, keytab and principle could be set using this. See section **Using keytabs on all worker hosts** under HDFS bolt below.| 
 
 ---
--- a/external/storm-hdfs/README.md
+++ b/external/storm-hdfs/README.md
@@ -1,8 +1,57 @@
 # Storm HDFS
 
 Storm components for interacting with HDFS file systems
+ - HDFS Bolt
+ - HDFS Spout
+ 
+
+# HDFS Spout
+
+## Usage
+
+The following example creates an HDFS spout that reads text files from HDFS path hdfs://localhost:54310/source.
+
+```java
+// Instantiate spout
+HdfsSpout textReaderSpout = new HdfsSpout().withOutputFields("line");
+// HdfsSpout seqFileReaderSpout = new HdfsSpout().withOutputFields("key","value");
+
+// Configure it
+Config conf = new Config();
+conf.put(Configs.SOURCE_DIR, "hdfs://localhost:54310/source");
+conf.put(Configs.ARCHIVE_DIR, "hdfs://localhost:54310/done");
+conf.put(Configs.BAD_DIR, "hdfs://localhost:54310/badfiles");
+conf.put(Configs.READER_TYPE, "text"); // or 'seq' for sequence files
+
+// Create & configure topology
+TopologyBuilder builder = new TopologyBuilder();
+builder.setSpout("hdfsspout", textReaderSpout, SPOUT_NUM);
+
+// Setup bolts and other topology configuration
+     ..snip..
+
+// Submit topology with config
+StormSubmitter.submitTopologyWithProgressBar("topologyName", conf, builder.createTopology());
+```
+
+## HDFS Spout Configuration Settings
+
+| Setting                  | Default     | Description |
+|--------------------------|-------------|-------------|
+|**hdfsspout.reader.type** |             | Indicates the reader for the file format. Set to 'seq' for reading sequence files or 'text' for text files. Set to a fully qualified class name if using a custom type (that implements interface org.apache.storm.hdfs.spout.FileReader)|
+|**hdfsspout.source.dir**  |             | HDFS location from where to read.  E.g. hdfs://localhost:54310/inputfiles       |
+|**hdfsspout.archive.dir** |             | After a file is processed completely it will be moved to this directory. E.g. hdfs://localhost:54310/done|
+|**hdfsspout.badfiles.dir**|             | if there is an error parsing a file's contents, the file is moved to this location.  E.g. hdfs://localhost:54310/badfiles  |
+|hdfsspout.ignore.suffix   |   .ignore   | File names with this suffix in the in the hdfsspout.source.dir location will not be processed|
+|hdfsspout.lock.dir        | '.lock' subdirectory under hdfsspout.source.dir | Dir in which lock files will be created. Concurrent HDFS spout instances synchronize using *lock*. Before processing a file the spout instance creates a lock file in this directory with same name as input file and deletes this lock file after processing the file. Spout also periodically makes a note of its progress (wrt reading the input file) in the lock file so that another spout instance can resume progress on the same file if the spout dies for any reason.|
+|hdfsspout.commit.count    |    20000    | Record progress in the lock file after these many records are processed. If set to 0, this criterion will not be used. |
+|hdfsspout.commit.sec      |    10       | Record progress in the lock file after these many seconds have elapsed. Must be greater than 0 |
+|hdfsspout.max.outstanding |   10000     | Limits the number of unACKed tuples by pausing tuple generation (if ACKers are used in the topology) |
+|hdfsspout.lock.timeout.sec|  5 minutes  | Duration of inactivity after which a lock file is considered to be abandoned and ready for another spout to take ownership |
+|hdfsspout.clocks.insync   |    true     | Indicates whether clocks on the storm machines are in sync (using services like NTP)       |
 
 
+# HDFS Bolt
 ## Usage
 The following example will write pipe("|")-delimited files to the HDFS path hdfs://localhost:54310/foo. After every
 1,000 tuples it will sync filesystem, making that data visible to other HDFS clients. It will rotate files when they
@@ -29,6 +78,7 @@ HdfsBolt bolt = new HdfsBolt()
         .withRotationPolicy(rotationPolicy)
         .withSyncPolicy(syncPolicy);
 ```
+
 
 ### Packaging a Topology
 When packaging your topology, it's important that you use the [maven-shade-plugin]() as opposed to the
@@ -115,7 +165,7 @@ Hadoop client version incompatibilites can manifest as errors like:
 com.google.protobuf.InvalidProtocolBufferException: Protocol message contained an invalid tag (zero)
 ```
 
-## Customization
+## HDFS Bolt Customization
 
 ### Record Formats
 Record format can be controlled by providing an implementation of the `org.apache.storm.hdfs.format.RecordFormat`
@@ -236,7 +286,7 @@ If you are using Trident and sequence files you can do something like this:
 ```
 
 
-## Support for HDFS Sequence Files
+## HDFS Bolt Support for HDFS Sequence Files
 
 The `org.apache.storm.hdfs.bolt.SequenceFileBolt` class allows you to write storm data to HDFS sequence files:
 
@@ -277,7 +327,7 @@ public interface SequenceFormat extends Serializable {
 }
 ```
 
-## Support for Avro Files
+## HDFS Bolt Support for Avro Files
 
 The `org.apache.storm.hdfs.bolt.AvroGenericRecordBolt` class allows you to write Avro objects directly to HDFS:
  
@@ -310,7 +360,7 @@ An `org.apache.avro.Schema` object cannot be directly provided since it does not
 The AvroGenericRecordBolt expects to receive tuples containing an Avro GenericRecord that conforms to the provided
 schema.
 
-## Trident API
+## HDFS Bolt support for Trident API
 storm-hdfs also includes a Trident `state` implementation for writing data to HDFS, with an API that closely mirrors
 that of the bolts.
 
--- a/external/storm-hdfs/README.md
+++ b/external/storm-hdfs/README.md
@@ -1,11 +1,50 @@
 # Storm HDFS
 
 Storm components for interacting with HDFS file systems
- - HDFS Bolt
  - HDFS Spout
- 
+ - HDFS Bolt 
 
 # HDFS Spout
+
+Hdfs spout is intended to allow feeding data into Storm from a HDFS directory. 
+It will actively monitor the directory to consume any new files that appear in the directory.
+
+**Impt**: Hdfs spout assumes that the files being made visible to it in the monitored directory 
+are NOT actively being written to. Only after a files is completely written should it be made
+visible to the spout. This can be achieved by either writing the files out to another directory 
+and once completely written, move it to the monitored directory. Alternatively the file
+can be created with a '.ignore' suffix in the monitored directory and after data is completely 
+written, rename it without the suffix. File names with a '.ignore' suffix are ignored
+by the spout.
+
+When the spout is actively consuming a file, ite renames the file with a '.inprogress' suffix.
+After consuming all the contents in the file, the file will be moved to a configurable *done* 
+directory and the '.inprogress' suffix will be dropped.
+
+**Concurrency** If multiple spout instances are used in the topology, each instance will consume
+a different file. Synchronization among spout instances is done using a lock files created in 
+(by default) a '.lock' subdirectory under the monitored directory. A file with the same name
+as the file being consumed (with the in progress suffix) is created in the lock directory.
+Once the file is completely consumed, the corresponding lock file is deleted.
+
+**Recovery from failure**
+Periodically, the spout also records progress information wrt to how much of the file has been
+consumed in the lock file. In case of an crash of the spout instance (or force kill of topology) 
+another spout can take over the file and resume from the location recorded in the lock file.
+
+Certain error conditions can cause lock files to go stale. Basically the lock file exists, 
+but no spout actively owns and therefore will the file will not be deleted. Usually this indicates 
+that the corresponding input file has also not been completely processed. A configuration
+'hdfsspout.lock.timeout.sec' can be set to specify the duration of inactivity that a lock file
+should be considered stale. Stale lock files are candidates for automatic transfer of ownership to 
+another spout.
+
+**Lock on .lock Directory**
+The .lock directory contains another DIRLOCK file which is used to co-ordinate accesses to the 
+.lock dir itself among spout instances. A spout will try to create it when it needs access to
+the .lock directory and then delete it when done.  In case of a topology crash or force kill,
+if this file still exists, it should be deleted to allow the new topology instance to regain 
+full access to the  .lock directory and resume normal processing. 
 
 ## Usage
 
@@ -13,8 +52,10 @@ The following example creates an HDFS spout that reads text files from HDFS path
 
 ```java
 // Instantiate spout
-HdfsSpout textReaderSpout = new HdfsSpout().withOutputFields("line");
-// HdfsSpout seqFileReaderSpout = new HdfsSpout().withOutputFields("key","value");
+HdfsSpout textReaderSpout = new HdfsSpout().withOutputFields(TextFileReader.defaultFields);
+// HdfsSpout seqFileReaderSpout = new HdfsSpout().withOutputFields(SequenceFileReader.defaultFields);
+
+// textReaderSpout.withConfigKey("custom.keyname"); // Optional. Not required normally unless you need to change the keyname use to provide hds settings. This keyname defaults to 'hdfs.config' 
 
 // Configure it
 Config conf = new Config();
@@ -34,21 +75,34 @@ builder.setSpout("hdfsspout", textReaderSpout, SPOUT_NUM);
 StormSubmitter.submitTopologyWithProgressBar("topologyName", conf, builder.createTopology());
 ```
 
-## HDFS Spout Configuration Settings
+## Configuration Settings
+Class HdfsSpout provided following methods for configuration:
 
-| Setting                  | Default     | Description |
-|--------------------------|-------------|-------------|
-|**hdfsspout.reader.type** |             | Indicates the reader for the file format. Set to 'seq' for reading sequence files or 'text' for text files. Set to a fully qualified class name if using a custom type (that implements interface org.apache.storm.hdfs.spout.FileReader)|
-|**hdfsspout.source.dir**  |             | HDFS location from where to read.  E.g. hdfs://localhost:54310/inputfiles       |
-|**hdfsspout.archive.dir** |             | After a file is processed completely it will be moved to this directory. E.g. hdfs://localhost:54310/done|
-|**hdfsspout.badfiles.dir**|             | if there is an error parsing a file's contents, the file is moved to this location.  E.g. hdfs://localhost:54310/badfiles  |
-|hdfsspout.ignore.suffix   |   .ignore   | File names with this suffix in the in the hdfsspout.source.dir location will not be processed|
-|hdfsspout.lock.dir        | '.lock' subdirectory under hdfsspout.source.dir | Dir in which lock files will be created. Concurrent HDFS spout instances synchronize using *lock*. Before processing a file the spout instance creates a lock file in this directory with same name as input file and deletes this lock file after processing the file. Spout also periodically makes a note of its progress (wrt reading the input file) in the lock file so that another spout instance can resume progress on the same file if the spout dies for any reason.|
-|hdfsspout.commit.count    |    20000    | Record progress in the lock file after these many records are processed. If set to 0, this criterion will not be used. |
-|hdfsspout.commit.sec      |    10       | Record progress in the lock file after these many seconds have elapsed. Must be greater than 0 |
-|hdfsspout.max.outstanding |   10000     | Limits the number of unACKed tuples by pausing tuple generation (if ACKers are used in the topology) |
-|hdfsspout.lock.timeout.sec|  5 minutes  | Duration of inactivity after which a lock file is considered to be abandoned and ready for another spout to take ownership |
-|hdfsspout.clocks.insync   |    true     | Indicates whether clocks on the storm machines are in sync (using services like NTP)       |
+`HdfsSpout withOutputFields(String... fields)` : This sets the names for the output fields. 
+The number of fields depends upon the reader being used. For convenience, built-in reader types 
+expose a static member called `defaultFields` that can be used for this. 
+ 
+ `HdfsSpout withConfigKey(String configKey)`
+Allows overriding the default key name (hdfs.config) with new name for specifying HDFS configs. Typicallly used
+to provide kerberos keytabs.
+
+Only settings mentioned in **bold** are required.
+
+| Setting                      | Default     | Description |
+|------------------------------|-------------|-------------|
+|**hdfsspout.reader.type**     |             | Indicates the reader for the file format. Set to 'seq' for reading sequence files or 'text' for text files. Set to a fully qualified class name if using a custom type (that implements interface org.apache.storm.hdfs.spout.FileReader)|
+|**hdfsspout.hdfs**            |             | HDFS URI. Example:  hdfs://namenodehost:8020
+|**hdfsspout.source.dir**      |             | HDFS location from where to read.  E.g. /data/inputfiles  |
+|**hdfsspout.archive.dir**     |             | After a file is processed completely it will be moved to this directory. E.g. /data/done|
+|**hdfsspout.badfiles.dir**    |             | if there is an error parsing a file's contents, the file is moved to this location.  E.g. /data/badfiles  |
+|hdfsspout.lock.dir            | '.lock' subdirectory under hdfsspout.source.dir | Dir in which lock files will be created. Concurrent HDFS spout instances synchronize using *lock* files. Before processing a file the spout instance creates a lock file in this directory with same name as input file and deletes this lock file after processing the file. Spout also periodically makes a note of its progress (wrt reading the input file) in the lock file so that another spout instance can resume progress on the same file if the spout dies for any reason. When a toplogy is killed, if a .lock/DIRLOCK file is left behind it can be safely deleted to allow normal resumption of the topology on restart.|
+|hdfsspout.ignore.suffix       |   .ignore   | File names with this suffix in the in the hdfsspout.source.dir location will not be processed|
+|hdfsspout.commit.count        |    20000    | Record progress in the lock file after these many records are processed. If set to 0, this criterion will not be used. |
+|hdfsspout.commit.sec          |    10       | Record progress in the lock file after these many seconds have elapsed. Must be greater than 0 |
+|hdfsspout.max.outstanding     |   10000     | Limits the number of unACKed tuples by pausing tuple generation (if ACKers are used in the topology) |
+|hdfsspout.lock.timeout.sec    |  5 minutes  | Duration of inactivity after which a lock file is considered to be abandoned and ready for another spout to take ownership |
+|hdfsspout.clocks.insync       |    true     | Indicates whether clocks on the storm machines are in sync (using services like NTP)       |
+|hdfs.config (unless changed)  |             | Set it to a Map of Key/value pairs indicating the HDFS settigns to be used. For example, keytab and principle could be set using this. See section **Using keytabs on all worker hosts** under HDFS bolt below.| 
 
 
 # HDFS Bolt
--- a/external/storm-hdfs/README.md
+++ b/external/storm-hdfs/README.md
@@ -1,109 +1,10 @@
 # Storm HDFS
 
 Storm components for interacting with HDFS file systems
- - HDFS Spout
  - HDFS Bolt 
+ - HDFS Spout
 
-# HDFS Spout
-
-Hdfs spout is intended to allow feeding data into Storm from a HDFS directory. 
-It will actively monitor the directory to consume any new files that appear in the directory.
-
-**Impt**: Hdfs spout assumes that the files being made visible to it in the monitored directory 
-are NOT actively being written to. Only after a files is completely written should it be made
-visible to the spout. This can be achieved by either writing the files out to another directory 
-and once completely written, move it to the monitored directory. Alternatively the file
-can be created with a '.ignore' suffix in the monitored directory and after data is completely 
-written, rename it without the suffix. File names with a '.ignore' suffix are ignored
-by the spout.
-
-When the spout is actively consuming a file, ite renames the file with a '.inprogress' suffix.
-After consuming all the contents in the file, the file will be moved to a configurable *done* 
-directory and the '.inprogress' suffix will be dropped.
-
-**Concurrency** If multiple spout instances are used in the topology, each instance will consume
-a different file. Synchronization among spout instances is done using a lock files created in 
-(by default) a '.lock' subdirectory under the monitored directory. A file with the same name
-as the file being consumed (with the in progress suffix) is created in the lock directory.
-Once the file is completely consumed, the corresponding lock file is deleted.
-
-**Recovery from failure**
-Periodically, the spout also records progress information wrt to how much of the file has been
-consumed in the lock file. In case of an crash of the spout instance (or force kill of topology) 
-another spout can take over the file and resume from the location recorded in the lock file.
-
-Certain error conditions can cause lock files to go stale. Basically the lock file exists, 
-but no spout actively owns and therefore will the file will not be deleted. Usually this indicates 
-that the corresponding input file has also not been completely processed. A configuration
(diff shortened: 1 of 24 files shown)


external/storm-hdfs/src/main/java/org/apache/storm/hdfs/common/CmpFilesByModificationTime.java
```diff
@@ -0,0 +1,32 @@
...
+import java.util.Comparator;
+
+
+public class CmpFilesByModificationTime
+        implements Comparator<LocatedFileStatus> {
+   @Override
+    public int compare(LocatedFileStatus o1, LocatedFileStatus o2) {
+      return new Long(o1.getModificationTime()).compareTo( o1.getModificationTime() );
+    }
+}
```
dossett: This is a very large PR, so I apologize for starting with something very small, but would using FileStatus instead of LocatedFileStatus work?  It's a higher level of abstraction and that's where getModificationTime() is defined so it seems more appropriate.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/common/HdfsUtils.java
```diff
@@ -0,0 +1,75 @@
...
+    ArrayList<LocatedFileStatus> fstats = new ArrayList<>();
+
+    RemoteIterator<LocatedFileStatus> itr = fs.listFiles(directory, false);
+    while( itr.hasNext() ) {
+      LocatedFileStatus fileStatus = itr.next();
+      if(olderThan>0 && fileStatus.getModificationTime()<olderThan )
+        fstats.add(fileStatus);
+      else
+        fstats.add(fileStatus);
+    }
```
dossett: The if/else branches execute the same statement?
`fstats.add(fileStatus)`

roshannaik: :-) indeed .. thanks for catching.. the first if should be nested conditions

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/common/CmpFilesByModificationTime.java
roshannaik: used it since thats what the API returns. To be more specific are you suggesting the use  FileSystem.listStatus() instead of  FileSystem.listFiles()  ?

dossett: No I meant to implement it like:
`implements Comparator<FileStatus>`
and change the compare() signature appropriately.  Since this is a public class it seems best to use the highest level of abstraction possible.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/HdfsSpout.java
```diff
@@ -0,0 +1,654 @@
...
+    if( ackEnabled  &&  tracker.size()>=maxDuplicates ) {
+      LOG.warn("Waiting for more ACKs before generating new tuples. " +
+               "Progress tracker size has reached limit {}"
+              , maxDuplicates);
+      // Don't emit anything .. allow configured spout wait strategy to kick in
+      return;
+    }
+
+    // 2) If no failed tuples, then send tuples from hdfs
+    while (true) {
```
arunmahadevan: `nextTuple` is already called in a tight loop, so another loop is not required here.

roshannaik: That loop is actually to ensure that when there is a ParseException we don't exit from the method immediately. Exiting from the method w/o emitting can trigger wait strategy. So to avoid that we retry from a new file by going back into the loop.  In other cases we such as successful parse or I/O exception, the loop will not repeat due to the return statements.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/common/CmpFilesByModificationTime.java
roshannaik: oh .. good idea. thanks.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/HdfsSpout.java
```diff
@@ -0,0 +1,654 @@
...
+      } catch (IOException e) {
+        LOG.error("I/O Error processing at file location " + getFileProgress(reader), e);
+        // don't emit anything .. allow configured spout wait strategy to kick in
+        return;
+      } catch (ParseException e) {
+        LOG.error("Parsing error when processing at file location " + getFileProgress(reader) +
+                ". Skipping remainder of file.", e);
+        markFileAsBad(reader.getFilePath());
+        // note: Unfortunately not emitting anything here due to parse error
+        // will trigger the configured spout wait strategy which is unnecessary
```
HeartSaVioR: @roshannaik 
Quick question: Is it safe to let reader as it is? Do we want to make reader as null to trigger finding next file?

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/HdfsSpout.java
HeartSaVioR: Idea is promising, but nextTuple should not be blocked longer cause it also blocks ack / fail.

roshannaik: Not sure if i understand the implied suggestion. It will be blocked for about the same time as the case when spout naturally reaches end of file and switches to next file.  exiting immediately would be a periodic perf hit.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/HdfsSpout.java
roshannaik: yes thats how it is. Reader is null-ed out by the  markFileAsBad() and markFileAsDone() methods.. triggering switch to the next file.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/HdfsSpout.java
HeartSaVioR: Oh, seems like you already handled edge-cases (there's no file to read).
Another edge case I can imagine is that there're too many skewed files and read-parse-discard them takes amount of time, but it shouldn't be normal.
I'm fine to have this loop. Thanks.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/HdfsSpout.java
HeartSaVioR: @roshannaik Thanks for the explanation. I didn't see unlockAndCloseReader().

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/TextFileReader.java
```diff
@@ -0,0 +1,168 @@
...
+
+  public TextFileReader(FileSystem fs, Path file, Map conf, String startOffset) throws IOException {
+    super(fs, file, new Fields(DEFAULT_FIELD_NAME));
+    offset = new TextFileReader.Offset(startOffset);
+    FSDataInputStream in = fs.open(file);
+    in.seek(offset.byteOffset);
+    String charSet = (conf==null || !conf.containsKey(CHARSET) ) ? "UTF-8" : conf.get(CHARSET).toString();
+    int buffSz = (conf==null || !conf.containsKey(BUFFER_SIZE) ) ? DEFAULT_BUFF_SIZE : Integer.parseInt( conf.get(BUFFER_SIZE).toString() );
+    reader = new BufferedReader(new InputStreamReader(in, charSet), buffSz);
+  }
```
satishd: Two constructors share similar logic. You may want to have code like below.


  public TextFileReader(FileSystem fs, Path file, Map conf) throws IOException {
    this(fs, file, conf, null);
  }

  public TextFileReader(FileSystem fs, Path file, Map conf, String startOffset) throws IOException {
    super(fs, file, new Fields(DEFAULT_FIELD_NAME));
    offset = startOffset != null? new TextFileReader.Offset(startOffset) : new TextFileReader.Offset(0,0);
    FSDataInputStream in = fs.open(file);
    in.seek(offset.byteOffset);
    String charSet = (conf==null || !conf.containsKey(CHARSET) ) ? "UTF-8" : conf.get(CHARSET).toString();
    int buffSz = (conf==null || !conf.containsKey(BUFFER_SIZE) ) ? DEFAULT_BUFF_SIZE : Integer.parseInt( conf.get(BUFFER_SIZE).toString() );
    reader = new BufferedReader(new InputStreamReader(in, charSet), buffSz);
  }

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/AbstractFileReader.java
```diff
@@ -0,0 +1,71 @@
...
+import org.apache.hadoop.fs.Path;
+
+
+abstract class AbstractFileReader implements FileReader {
+
+  private final Path file;
+  private final FileSystem fs;
+  private Fields fields;
+
+  public AbstractFileReader(FileSystem fs, Path file, Fields fieldNames) {
```
satishd: fs is never used. You may want to remove that.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/common/HdfsUtils.java
```diff
@@ -0,0 +1,101 @@
...
+    for (LocatedFileStatus fstat : fstats) {
+      result.add(fstat.getPath());
+    }
+    return result;
+  }
+
+  /**
+   * Returns true if succeeded. False if file already exists. throws if there was unexpected problem
+   */
+  public static FSDataOutputStream tryCreateFile(FileSystem fs, Path file) throws IOException {
```
satishd: You may want to update javadoc. It is not aligned with the code.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/common/HdfsUtils.java
```diff
@@ -0,0 +1,101 @@
...
+    try {
+      FSDataOutputStream os = fs.create(file, false);
+      return os;
+    } catch (FileAlreadyExistsException e) {
+      return null;
+    } catch (RemoteException e) {
+      if( e.getClassName().contentEquals(AlreadyBeingCreatedException.class.getName()) ) {
```
satishd: Why contentEquals() is used instead of equals()?  Can we consider using 'instance of' instead of string equals check?

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/FileLock.java
```diff
@@ -0,0 +1,300 @@
...
+ */
+public class FileLock {
+
+  private final FileSystem fs;
+  private final String componentID;
+  private final Path lockFile;
+  private final FSDataOutputStream lockFileStream;
+  private LogEntry lastEntry;
+
+  private static final Logger log = LoggerFactory.getLogger(DirLock.class);
```
satishd: It should have been Logger log = LoggerFactory.getLogger(FileLock.class)

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/FileLock.java
```diff
@@ -0,0 +1,300 @@
...
+    else
+      lockFileStream.writeBytes(line);
+    lockFileStream.hflush();
+
+    lastEntry = entry; // update this only after writing to hdfs
+  }
+
+  public void release() throws IOException {
+    lockFileStream.close();
+    fs.delete(lockFile, false);
```
satishd: fs.delete() returns false if the file is not deleted. You may want to throw an IOException for that. Caller of this method assumes file is not deleted only when IOException is thrown. We may even want to have log-warn when delete is not successful. 
Can we have javadoc for this method to describe the contract?

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/HdfsSpout.java
```diff
@@ -0,0 +1,654 @@
...
+    } catch (IOException e) {
+      LOG.error("Unable to create archive dir ", e);
+      throw new RuntimeException("Unable to create archive directory ", e);
+    }
+
+    // -- bad files dir config
+    if ( !conf.containsKey(Configs.BAD_DIR) ) {
+      LOG.error(Configs.BAD_DIR + " setting is required");
+      throw new RuntimeException(Configs.BAD_DIR + " setting is required");
+    }
```
satishd: Configs.BAD_DIR, Configs.ARCHIVE_DIR and Configs.SOURCE_DIR validation cheks are same. You may want to extract a method for these common validations.

arunmahadevan:
Good to add some notes in the README.md to explain the internals at a high level and the usage with the different config options. May be add an example topology in the storm-starter as well.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/HdfsSpout.java
```diff
@@ -0,0 +1,654 @@
...
+  }
+
+  private void setupCommitElapseTimer() {
+    if(commitFrequencySec<=0)
+      return;
+    TimerTask timerTask = new TimerTask() {
+      @Override
+      public void run() {
+        commitTimeElapsed.set(false);
```
satishd: Should not commitTimeElapsed be set to true?

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/HdfsSpout.java
```diff
@@ -0,0 +1,654 @@
...
+      markFileAsDone(reader.getFilePath());
+      reader = null;
+    }
+    super.ack(msgId);
+  }
+
+  private boolean canCommitNow() {
+    if( acksSinceLastCommit >= commitFrequencyCount )
+      return true;
+    return commitTimeElapsed.get();
```
arunmahadevan: commitTimeelapsed is never set to `true` anywhere?

satishd: commitTimeElapsed would never return true. It seems setupCommitElapseTimer should set that value to true whenever timerTask is run.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/HdfsSpout.java
```diff
@@ -0,0 +1,654 @@
...
+      }
+    }
+
+  }
+
+  // will commit progress into lock file if commit threshold is reached
+  private void commitProgress(FileOffset position) {
+    if ( lock!=null && canCommitNow() ) {
+      try {
+        lock.heartbeat(position.toString());
```
satishd: position parameter can be null, it should be handled.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/HdfsSpout.java
arunmahadevan: Th timer itself is not required since the commit is performed synchronously (whenever nextTuple, ack is invoked). I don't see the commits being triggered asynchronously from the timer.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/HdfsSpout.java
```diff
@@ -0,0 +1,654 @@
...
+            commitProgress(reader.getFileOffset());
+          } else {
+            commitProgress(tracker.getCommitPosition());
+          }
+          return;
+        } else {
+          fileReadCompletely = true;
+          if(!ackEnabled) {
+            markFileAsDone(reader.getFilePath());
+          }
```
satishd: if ackEnabled is true then fileReadCompletely is set as false on line no:142 after reading the first record from the next file. This makes the earlier file is always opened and it is not moved to archive dir.  ack() method has code to markFIleAsDone() when fileReadCompletely is true but it may not get triggered because fileReadCompletely is set as false.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/HdfsSpout.java
```diff
@@ -0,0 +1,654 @@
...
+  public void ack(Object msgId) {
+    MessageId id = (MessageId) msgId;
+    inflight.remove(id);
+    ++acksSinceLastCommit;
+    tracker.recordAckedOffset(id.offset);
+    commitProgress(tracker.getCommitPosition());
+    if(fileReadCompletely) {
+      markFileAsDone(reader.getFilePath());
+      reader = null;
+    }
```
satishd: tracker is not reset for the new file/lock. So, existing tracker may have offsets of earlier file read which may give wrong state of the system.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/common/CmpFilesByModificationTime.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+package org.apache.storm.hdfs.common;
+
+import org.apache.hadoop.fs.LocatedFileStatus;
+
+import java.util.Comparator;
+
+
+public class CmpFilesByModificationTime
```
hmcl: Rename class to ModificationTimeComparator or ModifTimeComparator - I don't think you should abbreviate Comparator

roshannaik: sure. good idea.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/FileLock.java
roshannaik: indeed. thanks.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/common/HdfsUtils.java
roshannaik: right. will update

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/TextFileReader.java
roshannaik: thanks. fixed this.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/AbstractFileReader.java
roshannaik: fixing it.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/FileLock.java
roshannaik: indeed. will do.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/HdfsSpout.java
roshannaik: will do.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/HdfsSpout.java
roshannaik: indeed it can be null. good observation.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/HdfsSpout.java
roshannaik: Ok shall look into this...and also add a Unit Test for it.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/HdfsSpout.java
roshannaik: ok will take a look at this too.. and add an UT for it.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/HdfsSpout.java
roshannaik: fixed. add UT.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/HdfsSpout.java
roshannaik: fixed. thanks.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/DirLock.java
```diff
@@ -0,0 +1,126 @@
...
+
+/**
+ * Facility to synchronize access to HDFS directory. The lock itself is represented
+ * as a file in the same directory. Relies on atomic file creation.
+ */
+public class DirLock {
+  private FileSystem fs;
+  private final Path lockFile;
+  public static final String DIR_LOCK_FILE = "DIRLOCK";
+  private static final Logger log = LoggerFactory.getLogger(DirLock.class);
```
harshach: can you make log as LOG. I understand log is the right naming :) but we've LOG everywhere in storm and connectors and makes it consistent.

roshannaik: done

pom.xml
```diff
@@ -204,7 +204,7 @@
         <clojure.tools.cli.version>0.2.4</clojure.tools.cli.version>
         <disruptor.version>3.3.2</disruptor.version>
         <jgrapht.version>0.9.0</jgrapht.version>
-        <guava.version>16.0.1</guava.version>
+        <guava.version>15.0</guava.version>
```
ptgoetz: Why the downgrade of the guava version?

ptgoetz:
Minor nit, but several instances throughout:

Can you update the style so that all `if` blocks are enclosed in `{}` (even if it is only one line)?

pom.xml
roshannaik: not needed anymore it seems. fixed it.

roshannaik:
@harshach  Can u take a look into committing ? All review comments have been addressed.  Its ready for commit from my perspective.

harshach:
+1

satishd:
+1

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/DirLock.java
```diff
@@ -0,0 +1,133 @@
...
+   * @param fs
+   * @param dir  the dir on which to get a lock
+   * @return The lock object if it the lock was acquired. Returns null if the dir is already locked.
+   * @throws IOException if there were errors
+   */
+  public static DirLock tryLock(FileSystem fs, Path dir) throws IOException {
+    Path lockFile = getDirLockFile(dir);
+
+    try {
+      FSDataOutputStream ostream = HdfsUtils.tryCreateFile(fs, lockFile);
```
arunmahadevan: This relies on `create(path, overwrite)` to atomically create the file if it does not exist. The notes for FSDataOutputStream create from this [link] (https://hadoop.apache.org/docs/stable/hadoop-project-dist/hadoop-common/filesystem/filesystem.html) says `Some implementations split the create into a check for the file existing from the actual creation. This means the operation is NOT atomic`. Can you check?. If create is not atomic you will have to rely on `rename` which must be atomic as per that doc.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/HdfsSpout.java
```diff
@@ -0,0 +1,702 @@
...
+          LOG.debug("Unable to get FileLock for {}, so skipping it.", file);
+          continue; // could not lock, so try another file.
+        }
+        LOG.info("Processing : {} ", file);
+        Path newFile = renameSelectedFile(file);
+        return createFileReader(newFile);
+      }
+
+      return null;
+    } catch (IOException e) {
```
arunmahadevan: Shouldnt the filelock get released in case of IOException and any other Exception?

external/storm-hdfs/README.md
```diff
@@ -405,7 +410,123 @@ On worker hosts the bolt/trident-state code will use the keytab file with princi
...
+processed. When detected, ownership of such stale lock files will be transferred to another spout.   
+The configuration 'hdfsspout.lock.timeout.sec' is used to specify the duration of inactivity after 
+which lock files should be considered stale. For lock file ownership transfer to succeed, the HDFS
+lease on the file (from prev lock owner) should have expired. Spouts scan for stale lock files
+before selecting the next file for consumption.
+
+**Lock on *.lock* Directory**
+Hdfs spout instances create a *DIRLOCK* file in the .lock directory to co-ordinate certain accesses to 
+the .lock dir itself. A spout will try to create it when it needs access to the .lock directory and
+then delete it when done.  In case of a topology crash or force kill, this file may not get deleted.
```
arunmahadevan: Isn't this automatically handled by lock expiry timeout ? So why should one manually delete the dir lock file ?

arunmahadevan:
Looks good overall but have a few questions/comments around the locking. Once addressed I am +1.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/DirLock.java
roshannaik: I believe that is for other file systems (such as AWS) that are supported via that class. I double checked on on HDFS semantics by talking to some HDFS committers. Have also added UTs to check for the requisite Hdfs atomicity/other semantics related to create, append and delete as an added safeguard. We will see failures in these tests if the guarantees are being violated. These are the relevant UT tests:  TestHdfsSemantics.*   TestFileLock.testConcurrentLocking,  TestDirLock.testConcurrentLocking

external/storm-hdfs/README.md
roshannaik: you are right.. it should.  I should reword it. Thanks!

roshannaik:
@arunmahadevan  have also addressed your comment about releasing the lock files on exception with the latest push. Good catch(..) !  ;-)

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/DirLock.java
arunmahadevan: The RawLocalFileSystem implementation does not seem to take care of atomicity - [link](https://github.com/apache/hadoop/blob/trunk/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/RawLocalFileSystem.java#L299). In the distributed mode name node could guarantee atomicity by locking, but I havent checked the details. Did you run the tests in local or distributed mode ? Anyways it would be hard to get the order and timing right for the tests to fail. 
It might be good to add some comments/docs to call out that the locking semantics depends on the underlying implementation guarantees.

arunmahadevan:
+1 Please squash the commits and add some comments/docs on the locking semantics https://github.com/apache/storm/pull/936#discussion_r49158800

external/storm-hdfs/README.md
```diff
@@ -405,7 +410,123 @@ On worker hosts the bolt/trident-state code will use the keytab file with princi
...
+|hdfsspout.lock.dir            | '.lock' subdirectory under hdfsspout.source.dir | Dir in which lock files will be created. Concurrent HDFS spout instances synchronize using *lock* files. Before processing a file the spout instance creates a lock file in this directory with same name as input file and deletes this lock file after processing the file. Spouts also periodically makes a note of their progress (wrt reading the input file) in the lock file so that another spout instance can resume progress on the same file if the spout dies for any reason.|
+|hdfsspout.ignore.suffix       |   .ignore   | File names with this suffix in the in the hdfsspout.source.dir location will not be processed|
+|hdfsspout.commit.count        |    20000    | Record progress in the lock file after these many records are processed. If set to 0, this criterion will not be used. |
+|hdfsspout.commit.sec          |    10       | Record progress in the lock file after these many seconds have elapsed. Must be greater than 0 |
+|hdfsspout.max.outstanding     |   10000     | Limits the number of unACKed tuples by pausing tuple generation (if ACKers are used in the topology) |
```
revans2: Do we want to explain how this works with max.spout.pending, and why we need both?

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/AbstractFileReader.java
```diff
@@ -0,0 +1,60 @@
...
+
+
+abstract class AbstractFileReader implements FileReader {
+
+  private final Path file;
+
+  public AbstractFileReader(FileSystem fs, Path file) {
+    if (fs == null ) {
+      throw new IllegalArgumentException("filesystem arg cannot be null for reader");
+    }
```
revans2: Why do we want fs to not be null, if we never actually use it?

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/DirLock.java
revans2: You are correct local file system does not do anything atomic.  Even rename is not atomic, even though the spec says it should be.  we should not test with the local file system if we want things to be atomic.  I think we mostly want to call out in the docs that if it was not obvious already the FileSystem used needs to support a specific set of atomic operations. (MRv2 has similar constraints for AM recovery, but it is not documented very well)

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/DirLock.java
```diff
@@ -0,0 +1,133 @@
...
+
+    // delete and recreate lock file
+    if( fs.delete(dirLockFile, false) ) { // returns false if somebody else already deleted it (to take ownership)
+      FSDataOutputStream ostream = HdfsUtils.tryCreateFile(fs, dirLockFile);
+      if(ostream!=null) {
+        ostream.close();
+      }
+      return new DirLock(fs, dirLockFile);
+    }
+    return null;
```
revans2: Shuldn't we just replace this part with tryLock?  The code is almost identical except for extra debugging that I think would be good.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/FileLock.java
```diff
@@ -0,0 +1,328 @@
...
+  // For testing only.. invoked via reflection
+  private void forceCloseLockFile() throws IOException {
+    lockFileStream.close();
+  }
+
+  /** returns lock on file or null if file is already locked. throws if unexpected problem */
+  public static FileLock tryLock(FileSystem fs, Path fileToLock, Path lockDirPath, String spoutId)
+          throws IOException {
+    String lockFileName = lockDirPath.toString() + Path.SEPARATOR_CHAR + fileToLock.getName();
+    Path lockFile = new Path(lockFileName);
```
revans2: It is probably cleaner to just do
```
Path lockFile = new Path(lockDirPath, fileToLock.getName());
```

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/FileLock.java
```diff
@@ -0,0 +1,328 @@
...
+   */
+  public static LogEntry getLastEntryIfStale(FileSystem fs, Path lockFile, long olderThan)
+          throws IOException {
+    long modifiedTime = fs.getFileStatus(lockFile).getModificationTime();
+    if( modifiedTime <= olderThan ) { // look
+      //Impt: HDFS timestamp may not reflect recent appends, so we double check the
+      // timestamp in last line of file to see when the last update was made
+      LogEntry lastEntry =  getLastEntry(fs, lockFile);
+      if(lastEntry==null) {
+        throw new RuntimeException(lockFile.getName() + " is empty. this file is invalid.");
```
revans2: Is blowing up the right thing to do when the file may be in the middle of being written. or the writer failed at a bad time?

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/HdfsSpout.java
```diff
@@ -0,0 +1,739 @@
...
+      LOG.debug("Sending tuple from retry list");
+      HdfsUtils.Pair<MessageId, List<Object>> pair = retryList.remove();
+      emitData(pair.getValue(), pair.getKey());
+      return;
+    }
+
+    if( ackEnabled  &&  tracker.size()>= maxOutstanding) {
+      LOG.warn("Waiting for more ACKs before generating new tuples. " +
+              "Progress tracker size has reached limit {}, SpoutID {}"
+              , maxOutstanding, spoutId);
```
revans2: Not sure this is worthy or a warn message.  It feels more like debug.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/HdfsSpout.java
```diff
@@ -0,0 +1,739 @@
...
+        }
+        // 4) Read record from file, emit to collector and record progress
+        List<Object> tuple = reader.next();
+        if (tuple != null) {
+          fileReadCompletely= false;
+          ++tupleCounter;
+          MessageId msgId = new MessageId(tupleCounter, reader.getFilePath(), reader.getFileOffset());
+          emitData(tuple, msgId);
+
+          if(!ackEnabled) {
```
revans2: If acks are disabled by storm ack is called as part of the call to emit.  I think you may ahve extra logic in here that you do not need.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/HdfsSpout.java
```diff
@@ -0,0 +1,739 @@
...
+      LOG.error("Unable to instantiate file system", e);
+      throw new RuntimeException("Unable to instantiate file system", e);
+    }
+
+
+    if ( conf.containsKey(configKey) ) {
+      Map<String, Object> map = (Map<String, Object>)conf.get(configKey);
+        if(map != null) {
+          for(String keyName : map.keySet()){
+            LOG.info("HDFS Config override : " + keyName + " = " + String.valueOf(map.get(keyName)));
```
revans2: minor nit, can we use "{}" for variable substitution.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/HdfsSpout.java
```diff
@@ -0,0 +1,739 @@
...
+      int ackerCount = Integer.parseInt(ackers.toString());
+      this.ackEnabled = (ackerCount>0);
+      LOG.debug("ACKer count = {}", ackerCount);
+    }
+    else {
+      this.ackEnabled = false;
+      LOG.debug("No ACKers config found");
+    }
+
+    LOG.info("ACK mode is {}", ackEnabled ? "enabled" : "disabled");
```
revans2: Again.  If acking is disabled ack is called on the spout as part of the emit.  You should not need all of this extra logic.  Storm should handle all of that for you.

revans2:
Overall things look good, but I did not look too in depth at the code.  I had a few minor nits and potential things for code clean up, but nothing I think would keep this from going in.  Great work.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/FileLock.java
roshannaik: Since these lock files have 1 line appended immediately on creation, if we find a empty file whose modify time is older than lock expiry time, then we have a corrupt file..  cant be case of another in progress writer / bad timing.

revans2: Thanks for the explanation about when this can happen, but I am more concerned about how we can recover from this.  The RuntimeException when it does happen looks like it will bubble all the way up to nextTuple which will trigger the worker to shut down and restart.  But because nothing has happened to the corrupted file all of the spouts as they finish parsing their file will now go into an endless restart loop.

Can we at least think about how we can recover from this situation.  Or at least not crash constantly when it does happen and yell a lot in the logs.

roshannaik: i see. good point. On discovering an empty lock file, we can 
- simply skip over it after printing a log
- or delete it after logging .. as not much can be done with empty files anyway.

i suggest the latter. what do you think ?

harshach:
@roshannaik can you please open a follow-up jira to address any additional comments or suggestions.

roshannaik:
@harshach 
ok i shall open another jira and address all new comments since yesterday through it.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/HdfsSpout.java
```diff
@@ -0,0 +1,739 @@
...
+    LOG.trace("Next Tuple {}", spoutId);
+    // 1) First re-emit any previously failed tuples (from retryList)
+    if (!retryList.isEmpty()) {
+      LOG.debug("Sending tuple from retry list");
+      HdfsUtils.Pair<MessageId, List<Object>> pair = retryList.remove();
+      emitData(pair.getValue(), pair.getKey());
+      return;
+    }
+
+    if( ackEnabled  &&  tracker.size()>= maxOutstanding) {
```
erikdw: nit: whitespace is off:  `if( ackEnabled  &&  ` --> `if (ackEnabled && `

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/HdfsSpout.java
```diff
@@ -0,0 +1,739 @@
...
+        if (reader == null) {
+          reader = pickNextFile();
+          if (reader == null) {
+            LOG.debug("Currently no new files to process under : " + sourceDirPath);
+            return;
+          } else {
+            fileReadCompletely=false;
+          }
+        }
+        if( fileReadCompletely ) { // wait for more ACKs before proceeding
```
erikdw: inconsistent whitespace.
should be:  `if (fileReadCompletely) {`

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/HdfsSpout.java
```diff
@@ -0,0 +1,739 @@
...
+      this.hdfs = FileSystem.get(URI.create(hdfsUri), hdfsConfig);
+    } catch (IOException e) {
+      LOG.error("Unable to instantiate file system", e);
+      throw new RuntimeException("Unable to instantiate file system", e);
+    }
+
+
+    if ( conf.containsKey(configKey) ) {
+      Map<String, Object> map = (Map<String, Object>)conf.get(configKey);
+        if(map != null) {
```
erikdw: indentation is messed up.  the if and its block should be on same level as the map variable declaration.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/HdfsSpout.java
```diff
@@ -0,0 +1,739 @@
...
+
+    try {
+      this.hdfs = FileSystem.get(URI.create(hdfsUri), hdfsConfig);
+    } catch (IOException e) {
+      LOG.error("Unable to instantiate file system", e);
+      throw new RuntimeException("Unable to instantiate file system", e);
+    }
+
+
+    if ( conf.containsKey(configKey) ) {
```
erikdw: please be *consistent* with styling.  All of these variants exist in this code:
```
if(foo
if ( foo
if( foo
if (foo
```

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/HdfsSpout.java
```diff
@@ -0,0 +1,739 @@
...
+    commitProgress(tracker.getCommitPosition());
+    if(fileReadCompletely && inflight.isEmpty()) {
+      markFileAsDone(reader.getFilePath());
+      reader = null;
+    }
+    super.ack(msgId);
+  }
+
+  private boolean canCommitNow() {
+
```
erikdw: extraneous newline

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/common/ModifTimeComparator.java
```diff
@@ -0,0 +1,32 @@
...
+
+import org.apache.hadoop.fs.FileStatus;
+
+import java.util.Comparator;
+
+
+public class ModifTimeComparator
+        implements Comparator<FileStatus> {
+   @Override
+    public int compare(FileStatus o1, FileStatus o2) {
```
erikdw: indentation is messed up.  3 and 4 spaces.  Should be 2.

roshannaik:
As per request from @harshach, I have opened up STORM-1474 for addressing the pending minor review comments. Looks like @harshach is away, @Parth-Brahmbhatt  could you take a look into committing it ?

Parth-Brahmbhatt:
@roshannaik Thanks for the patch. I have merged this request to master.

PR #939, opened 2015-12-10
[STORM-1175] State store for windowing operations

arunmahadevan:
Added IStatefulBolt abstraction that can be implemented by bolts which requires
its state to be checkpointed periodically. State implementations based on
key-value mapping store are added. There is a default in-memory based implementation
and optional redis based implementation that provides state persistence. An internal
CheckpointSpout periodically emits checkpoint tuples which flows through the
topology DAG to take a consistent snapshot of the state across all components.

There is still pending work to capture the evaluated/expired state of the tuples
in the Window and use it to prune duplicate window evaluations during restart. This can
be built on top of the stateful bolt abstractions and will be done as part of STORM-1176

### DIFF
--- a/external/storm-redis/src/main/java/org/apache/storm/redis/state/RedisKeyValueState.java
+++ b/external/storm-redis/src/main/java/org/apache/storm/redis/state/RedisKeyValueState.java
@@ -33,6 +33,7 @@ import java.io.IOException;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 /**
  * A redis based implementation that persists the state in Redis.
@@ -76,7 +77,7 @@ public class RedisKeyValueState<K, V> implements KeyValueState<K, V> {
         this.keySerializer = keySerializer;
         this.valueSerializer = valueSerializer;
         this.jedisContainer = jedisContainer;
-        this.pendingPrepare = new HashMap<>();
+        this.pendingPrepare = new ConcurrentHashMap<>();
         initTxids();
         initPendingCommit();
     }
@@ -101,10 +102,10 @@ public class RedisKeyValueState<K, V> implements KeyValueState<K, V> {
             commands = jedisContainer.getInstance();
             if (commands.exists(prepareNamespace)) {
                 LOG.debug("Loading previously prepared commit from {}", prepareNamespace);
-                pendingCommit = commands.hgetAll(prepareNamespace);
+                pendingCommit = Collections.unmodifiableMap(commands.hgetAll(prepareNamespace));
             } else {
                 LOG.debug("No previously prepared commits.");
-                pendingCommit = new HashMap<>();
+                pendingCommit = Collections.emptyMap();
             }
         } finally {
             jedisContainer.returnInstance(commands);
@@ -169,7 +170,7 @@ public class RedisKeyValueState<K, V> implements KeyValueState<K, V> {
             txIds.put(PREPARE_TXID_KEY, String.valueOf(txid));
             commands.hmset(txidNamespace, txIds);
             pendingCommit = Collections.unmodifiableMap(pendingPrepare);
-            pendingPrepare = new HashMap<>();
+            pendingPrepare = new ConcurrentHashMap<>();
         } finally {
             jedisContainer.returnInstance(commands);
         }
@@ -206,7 +207,7 @@ public class RedisKeyValueState<K, V> implements KeyValueState<K, V> {
             } else {
                 LOG.debug("Nothing to save for commit");
             }
-            pendingPrepare = new HashMap<>();
+            pendingPrepare = new ConcurrentHashMap<>();
         } finally {
             jedisContainer.returnInstance(commands);
         }
@@ -231,7 +232,7 @@ public class RedisKeyValueState<K, V> implements KeyValueState<K, V> {
             }
             commands.hmset(txidNamespace, txIds);
             pendingCommit = Collections.emptyMap();
-            pendingPrepare = new HashMap<>();
+            pendingPrepare = new ConcurrentHashMap<>();
         } finally {
             jedisContainer.returnInstance(commands);
         }
@@ -247,10 +248,6 @@ public class RedisKeyValueState<K, V> implements KeyValueState<K, V> {
             if (txid <= committedTxid) {
                 throw new RuntimeException("Invalid txid '" + txid + "' for prepare. Txid '" + committedTxid +
                                                    "' is already committed");
-            }
-            if (txid > committedTxid + 1) {
-                throw new RuntimeException("Cannot prepare a txn with id '" + txid +
-                                                   "' when last committed txid is '" + committedTxid + "'");
             }
         }
     }
--- a/external/storm-redis/src/main/java/org/apache/storm/redis/state/RedisKeyValueState.java
+++ b/external/storm-redis/src/main/java/org/apache/storm/redis/state/RedisKeyValueState.java
@@ -91,6 +91,7 @@ public class RedisKeyValueState<K, V> implements KeyValueState<K, V> {
             } else {
                 txIds = new HashMap<>();
             }
+            LOG.debug("initTxids, txIds {}", txIds);
         } finally {
             jedisContainer.returnInstance(commands);
         }
@@ -230,7 +231,10 @@ public class RedisKeyValueState<K, V> implements KeyValueState<K, V> {
             } else {
                 txIds.remove(PREPARE_TXID_KEY);
             }
-            commands.hmset(txidNamespace, txIds);
+            if (!txIds.isEmpty()) {
+                LOG.debug("hmset txidNamespace {}, txIds {}", txidNamespace, txIds);
+                commands.hmset(txidNamespace, txIds);
+            }
             pendingCommit = Collections.emptyMap();
             pendingPrepare = new ConcurrentHashMap<>();
         } finally {
--- a/external/storm-redis/src/main/java/org/apache/storm/redis/state/RedisKeyValueState.java
+++ b/external/storm-redis/src/main/java/org/apache/storm/redis/state/RedisKeyValueState.java
@@ -40,8 +40,8 @@ import java.util.concurrent.ConcurrentHashMap;
  */
 public class RedisKeyValueState<K, V> implements KeyValueState<K, V> {
     private static final Logger LOG = LoggerFactory.getLogger(RedisKeyValueState.class);
-    private static final String COMMIT_TXID_KEY = "$commit";
-    private static final String PREPARE_TXID_KEY = "$prepare";
+    private static final String COMMIT_TXID_KEY = "commit";
+    private static final String PREPARE_TXID_KEY = "prepare";
 
     private final BASE64Encoder base64Encoder;
     private final BASE64Decoder base64Decoder;
--- a/external/storm-redis/src/main/java/org/apache/storm/redis/state/RedisKeyValueState.java
+++ b/external/storm-redis/src/main/java/org/apache/storm/redis/state/RedisKeyValueState.java
@@ -1,0 +1,305 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.storm.redis.state;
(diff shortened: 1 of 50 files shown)


storm-core/test/jvm/backtype/storm/state/InMemoryKeyValueStateTest.java
```diff
@@ -0,0 +1,73 @@
...
+        assertEquals(null, keyValueState.get("c"));
+    }
+
+    @Test
+    public void testPrepareCommitRollback() throws Exception {
+        keyValueState.put("a", "1");
+        keyValueState.put("b", "2");
+        keyValueState.prepareCommit(1);
+        keyValueState.put("c", "3");
+        assertEquals("1", keyValueState.get("a"));
```
hmcl: Please put all of this duplicated code in a helper method. All of this copy and paste should be one line helper method call

external/storm-redis/src/main/java/org/apache/storm/redis/state/RedisKeyValueState.java
```diff
@@ -0,0 +1,302 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * A redis based implementation that persists the state in Redis.
+ */
+public class RedisKeyValueState<K, V> implements KeyValueState<K, V> {
+    private static final Logger LOG = LoggerFactory.getLogger(RedisKeyValueState.class);
+    private static final String COMMIT_TXID_KEY = "$commit";
```
harshach: any reason for using $ notation here.

storm-core/src/jvm/backtype/storm/Config.java
```diff
@@ -1491,6 +1491,29 @@
...
+     */
+    @isString
+    public static final String TOPOLOGY_STATE_PROVIDER = "topology.state.provider";
+
+    /**
+     * The configuration specific to the {@link backtype.storm.state.StateProvider} implementation.
+     * This can be overridden at the component level.
+     */
+    @isString
+    public static final String TOPOLOGY_STATE_PROVIDER_CONFIG = "topology.state.provider.config";
```
harshach: whats is this config suppose to do ? will it accept it file that contains config?

storm-core/src/jvm/backtype/storm/spout/CheckPointState.java
```diff
@@ -0,0 +1,78 @@
...
+
+    public enum State {
+        /**
+         * The checkpoint spout has committed the transaction.
+         */
+        COMMITTED,
+        /**
+         * The checkpoint spout has started committing the transaction.
+         */
+        COMMITTING,
```
harshach: should we call this as COMMIT_IN_PROGRESS

storm-core/src/jvm/backtype/storm/spout/CheckPointState.java
```diff
@@ -0,0 +1,78 @@
...
+        /**
+         * The checkpoint spout has started preparing the transaction for commit.
+         */
+        PREPARING
```
harshach: PREPARE probably better

storm-core/src/jvm/backtype/storm/spout/CheckPointState.java
```diff
@@ -0,0 +1,78 @@
...
+ * limitations under the License.
+ */
+package backtype.storm.spout;
+
+/**
+ * Captures the current state of the transaction in
+ * {@link CheckpointSpout}
+ */
+public class CheckPointState {
+    public long txid;
```
Parth-Brahmbhatt: shouldn't these be private fields with getters?

storm-core/src/jvm/backtype/storm/spout/CheckpointSpout.java
```diff
@@ -0,0 +1,280 @@
...
+    public static final String CHECKPOINT_COMPONENT_ID = "$checkpointspout";
+    public static final String CHECKPOINT_FIELD_TXID = "txid";
+    public static final String CHECKPOINT_FIELD_ACTION = "action";
+    public static final String CHECKPOINT_ACTION_PREPARE = "prepare";
+    public static final String CHECKPOINT_ACTION_COMMIT = "commit";
+    public static final String CHECKPOINT_ACTION_ROLLBACK = "rollback";
+    public static final String CHECKPOINT_ACTION_INITSTATE = "initstate";
+
+    private static final String TX_STATE_KEY = "__state";
+    private static final int DEFAULT_CHECKPOINT_INTERVAL = 1000; // every sec
```
harshach: can we have this as part of defaults.yaml

storm-core/src/jvm/backtype/storm/spout/CheckpointSpout.java
```diff
@@ -0,0 +1,280 @@
...
+/**
+ * Emits checkpoint tuples which is used to save the state of the {@link backtype.storm.topology.IStatefulComponent}
+ * across the topology. If a topology contains Stateful bolts, Checkpoint spouts are automatically added
+ * to the topology. There is only one Checkpoint task per topology.
+ * <p/>
+ * Checkpoint spout stores its internal state in a {@link KeyValueState}. The state transitions are as follows.
+ * <p/>
+ * <pre>
+ *                  ROLLBACK(tx2)
+ *               <-------------                  PREPARE(tx2)                     COMMIT(tx2)
```
Parth-Brahmbhatt: why Rollback can only happen from Prepare and not from Committing? in that case why do we have Committing state?

storm-core/src/jvm/backtype/storm/spout/CheckpointSpout.java
```diff
@@ -0,0 +1,280 @@
...
+        checkpointStepInProgress = false;
+        recovering = true;
+    }
+
+    @Override
+    public void nextTuple() {
+        if (shouldRecover()) {
+            LOG.debug("In recovery");
+            handleRecovery();
+            startProgress();
```
Parth-Brahmbhatt: given you do this in both if and else as the last step, probably just move it outside.

storm-core/src/jvm/backtype/storm/spout/CheckpointSpout.java
```diff
@@ -0,0 +1,280 @@
...
+
+    @Override
+    public void declareOutputFields(OutputFieldsDeclarer declarer) {
+        declarer.declareStream(CHECKPOINT_STREAM_ID, new Fields(CHECKPOINT_FIELD_TXID, CHECKPOINT_FIELD_ACTION));
+    }
+
+    @Override
+    public Map<String, Object> getComponentConfiguration() {
+        Config conf = new Config();
+        conf.put(Config.TOPOLOGY_SLEEP_SPOUT_WAIT_STRATEGY_TIME_MS, 100);
```
Parth-Brahmbhatt: anyway to override this?

storm-core/src/jvm/backtype/storm/spout/CheckpointSpout.java
```diff
@@ -0,0 +1,280 @@
...
+            LOG.debug("Emitting commit with txid {}", txState.txid);
+            collector.emit(CHECKPOINT_STREAM_ID, new Values(txState.txid, CHECKPOINT_ACTION_COMMIT), txState.txid);
+        } else if (shouldInitState(txState)) {
+            LOG.debug("Emitting init state with txid {}", txState.txid);
+            collector.emit(CHECKPOINT_STREAM_ID, new Values(txState.txid, CHECKPOINT_ACTION_INITSTATE), txState.txid);
+        }
+        startProgress();
+    }
+
+    private void handleRecoveryAck() {
```
Parth-Brahmbhatt: the state transition code seems to be repeating in handleCheckpointAck, doCheckpoint and handleRecoveryAck. Do you think we can capture the nextState and prevState for each state in the CheckPointState class it self so in all these methods we can just call netxt/prevState followed by saveTxState(txState).

Parth-Brahmbhatt:
@arunmahadevan Thanks for the patch. I think the missing piece is the how-to-use guide. you can file a follow up jira for that. Overall looks good to me and once the comments are addressed , I am +1.

external/storm-redis/src/main/java/org/apache/storm/redis/state/RedisKeyValueState.java
arunmahadevan: Since its internal, just followed a convention to differentiate it (similar to checkpoint stream). But here there is no namespace collision so it could be "commit" as well.

storm-core/src/jvm/backtype/storm/Config.java
arunmahadevan: For now we expect the actual config string (for redis, its a json string with the host, port etc). it could be a file name as well, if the state provider implementation desires so.

storm-core/src/jvm/backtype/storm/spout/CheckPointState.java
arunmahadevan: Named it COMMITTING and PREPARING since its short. We could probably have it as COMMIT_IN_PROGRESS and PREPARE_IN_PROGRESS. Prefer to keep it short and can add more comments.

storm-core/src/jvm/backtype/storm/spout/CheckPointState.java
arunmahadevan: It could be PREPARE_IN_PROGRESS or PREPARING, I think PREPARE conveys a slightly different meaning.

storm-core/src/jvm/backtype/storm/spout/CheckPointState.java
arunmahadevan: will change it

storm-core/src/jvm/backtype/storm/spout/CheckpointSpout.java
arunmahadevan: Its used only if Config.TOPOLOGY_STATE_CHECKPOINT_INTERVAL is not set. Will add comments to make it clear. Will also set the default value for that config in defaults.yaml

storm-core/src/jvm/backtype/storm/spout/CheckpointSpout.java
arunmahadevan: Rollback can happen from 'preparing' state since all bolts might not have prepared the data to commit. Once we get ack for prepare it indicates all bolts have prepared the data to commit, the state changes from 'preparing' to 'committing' and a 'commit' message is sent. We cant rollback from 'committing' state since some bolts might have already committed the data. In case of failures, the commit message is sent again to make sure all bolts commit. If some bolts already committed the prepared data, it ignores the message.

Will add more comments to make it clear.

storm-core/src/jvm/backtype/storm/spout/CheckpointSpout.java
arunmahadevan: Actually it can't be moved out since its in the `elseif`. If nextTuple is invoked while its in the middle of recovering/checkpointing, it does not do anything. I will add a final else and put logs/comment.

storm-core/src/jvm/backtype/storm/spout/CheckpointSpout.java
arunmahadevan: It was added to sleep for slightly longer time rather than spinning in the nextTuple. Instead of hard coding, will change this to 10% of the checkpoint interval time.

storm-core/src/jvm/backtype/storm/spout/CheckpointSpout.java
arunmahadevan: Yes it will make it cleaner. I will try to refactor it and move it to nextAction and nextState in checkpoint state.

arunmahadevan:
@Parth-Brahmbhatt @harshach refactored the state handling code to make it cleaner and addressed other review comments. Also added a followup task for documentation - https://issues.apache.org/jira/browse/STORM-1416

storm-core/src/jvm/backtype/storm/spout/CheckpointSpout.java
harshach: @arunmahadevan well if you add that to defaults.yaml it will be set and thats how the defaults has been used so far in storm. Incase if we need to change default in the future its .yaml file change than the code change.

harshach:
+1 as well. minor nit pick on the default value.

arunmahadevan:
@harshach while addressing review comments I had added the config in defaults.yaml. I further refactored it to remove the constant and made sure that the value is not too less.

arunmahadevan:
@harshach @Parth-Brahmbhatt whilde doing more tests found out a few issues around acking and failure handling and have addressed in https://github.com/apache/storm/commit/c075ffacdf1680ea3792c4e7979228e3a1d65f0c can you take a look?

Parth-Brahmbhatt:
@arunmahadevan I am still +1 but the up merge is failing for storm-core/src/jvm/org/apache/storm/topology/TopologyBuilder.java can you please upmerge the last time and I can merge this into master.

arunmahadevan:
@Parth-Brahmbhatt Fixed the merge conflicts. 

Also had to refactor most of the code since the package name changed from backtype.storm -> org.apache.storm in master. Ran the sample topologies after the package name refactoring and things look good.

ptgoetz:
+1

arunmahadevan:
@kosii Thanks for reporting. Yes the keys in `pendingPrepare` should have priority. Typically this case does not happen since when a prepare fails, the txn is rolled-back and retried (i.e a fresh prepare). 

Do you want to post a JIRA/Patch?

kosii:
@arunmahadevan yes, I'm on it

PR #947, opened 2015-12-14
[STORM-1390] If you find a suitable constructor do NOT need to look for a d…

lispking:
If you find the right constructor, you do not have to look any further.

### DIFF
--- a/external/flux/flux-core/src/main/java/org/apache/storm/flux/FluxBuilder.java
+++ b/external/flux/flux-core/src/main/java/org/apache/storm/flux/FluxBuilder.java
@@ -395,6 +395,7 @@ public class FluxBuilder {
                     eligibleCount++;
                 }
                 LOG.debug("** invokable --> {}", invokable);
+                break;
             } else {
                 LOG.debug("Skipping constructor with wrong number of arguments.");
             }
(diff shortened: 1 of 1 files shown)


unsleepy22:
@lispking regarding the original code, I think it means to give a warning when it meets multiple eligible constructors (sort of like multiple bindings for slf4j), so I don't think we should break on first match.

lispking:
@unsleepy22 i see, but this may make people confused.

external/flux/flux-core/src/main/java/org/apache/storm/flux/FluxBuilder.java
```diff
@@ -395,6 +395,7 @@ private static Constructor findCompatibleConstructor(List<Object> args, Class ta
                     eligibleCount++;
                 }
```
HeartSaVioR: @lispking ```break``` should be here, not there.

HeartSaVioR:
@lispking 
Though I've addressed potential bug point, I'm with @unsleepy22. 
Its role is not that we strongly need to optimize. It doesn't affect performance.
Knowing user's intention correctly is the most important point to Flux so finding out missing point and showing warning is vaild operation, I think.

lispking:
@HeartSaVioR Thank reminder, your proposal is appropriate。

PR #948, opened 2015-12-15
[storm-1150]Fix the authorization of Logviewer in method authorized-log-user?

rujia1019:


storm-core/src/clj/backtype/storm/daemon/logviewer.clj
```diff
@@ -320,7 +320,7 @@
           logs-groups (concat (conf LOGS-GROUPS)
                               group-wl)]
        (or (some #(= % user) logs-users)
-           (< 0 (.size (intersection (set groups) (set group-wl))))))))
+           (< 0 (.size (intersection (set groups) (set logs-groups))))))))
```
rfarivar: Given that logs-groups was sitting there in line 322, and was not being used at all prior to this pull request, this seems like it was intended to be used like exactly you've changed it here. But the fact that it wasn't used makes me think there is more in play here than meets the eyes. 

Isn't group-wl already supposed to extract this information anyway?

revans2:
+1.  great catch @rujia1019

PR #955, opened 2015-12-17
STORM-1399: Blobstore tests should write data to target so it gets re…

ptgoetz:
…moved when running mvn clean

https://issues.apache.org/jira/browse/STORM-1399

### DIFF
--- a/external/storm-hdfs/src/test/java/org/apache/storm/hdfs/blobstore/BlobStoreTest.java
+++ b/external/storm-hdfs/src/test/java/org/apache/storm/hdfs/blobstore/BlobStoreTest.java
@@ -76,6 +76,7 @@ public class BlobStoreTest {
 
   @Before
   public void init() {
+    System.setProperty("test.build.data", "target/test/data");
     initializeConfigs();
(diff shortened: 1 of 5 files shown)


redsanket:
+1 NB

revans2:
+1

external/storm-hdfs/src/test/java/org/apache/storm/hdfs/blobstore/BlobStoreTest.java
```diff
@@ -76,6 +76,7 @@
 
   @Before
   public void init() {
+    System.setProperty("test.build.data", "target/test/data");
     initializeConfigs();
     baseFile = new File("/tmp/blob-store-test-"+UUID.randomUUID());
```
knusbaum: Why are we not also changing here?

ptgoetz: It breaks the blob store tests, and chasing down the root cause was getting to be a pain. At least writing test data to /tmp doesn't clutter up the source tree, but I would also like to avoid that as well. I figured we could address that separately.

PR #959, opened 2015-12-18
[STORM-1403] consider gc childopts in memory calculation for worker JVM

zhuoliu:


revans2:
+1

storm-core/src/jvm/backtype/storm/scheduler/Cluster.java
```diff
@@ -480,6 +480,27 @@ private String getStringFromStringList(Object o) {
     * */
     private Double getAssignedMemoryForSlot(Map topConf) {
         Double totalWorkerMemory = 0.0;
+        final Integer TOPOLOGY_WORKER_DEFAULT_MEMORY_ALLOCATION = 768;
+
+        String topologyWorkerGcChildopts = null;
```
unsleepy22: unnecessary null assignment, same below

storm-core/src/jvm/backtype/storm/scheduler/Cluster.java
```diff
@@ -480,6 +480,27 @@ private String getStringFromStringList(Object o) {
     * */
     private Double getAssignedMemoryForSlot(Map topConf) {
         Double totalWorkerMemory = 0.0;
+        final Integer TOPOLOGY_WORKER_DEFAULT_MEMORY_ALLOCATION = 768;
```
unsleepy22: is it necessary since default worker memory is set in defaults.yaml

zhuoliu: In case someone manually sets it to be nil in the configuration.

hustfxj:
+1

storm-core/src/jvm/backtype/storm/scheduler/Cluster.java
zhuoliu: Do not see any harm of this null assignment, more of a habitual choice. Same below.

PR #961, opened 2015-12-21
[STORM-1407] fix authorize issue for daemon log link in UI

zhuoliu:
The authorize check incorrectly use the worker log manner for daemon log.
Here remove that permission checks for daemon logs.
If necessary, we may later file a separate JIRA to enable different authorize check for daemon logs.


revans2:
+1

vesense:
LGTM +1

harshach:
+1

PR #963, opened 2015-12-22
[STORM-1176] Checkpoint window evaluated/expired state

arunmahadevan:
This uses the stateful bolt abstractions introduced in PR# 939 to checkpoint the last expired and last evaluated tuple message ids. IStatefulWindowedBolt exposes the framework managed state to the windowed bolts so that the state of the windowing operation can be saved. The last expired and last evaluated message ids are saved by `StatefulWindowedBoltExecutor` along with the state of the windowing operation. This last expired/evaluated ids are used during recovery to avoid duplicate window evaluations. This still provides at least once semantics but minimizes the duplicate window evaluations.

Note: This PR includes the commits proposed in https://github.com/apache/storm/pull/939 since it builds on top of that.

### DIFF
--- a/examples/storm-starter/src/jvm/storm/starter/StatefulWindowingTopology.java
+++ b/examples/storm-starter/src/jvm/storm/starter/StatefulWindowingTopology.java
@@ -35,6 +35,8 @@ import org.apache.storm.tuple.Tuple;
 import org.apache.storm.tuple.Values;
 import org.apache.storm.utils.Utils;
 import org.apache.storm.windowing.TupleWindow;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.util.Map;
 
@@ -51,6 +53,7 @@ import static org.apache.storm.topology.base.BaseWindowedBolt.Count;
  * </p>
  */
 public class StatefulWindowingTopology {
+    private static final Logger LOG = LoggerFactory.getLogger(StatefulWindowingTopology.class);
 
     private static class WindowSumBolt extends BaseStatefulWindowedBolt<KeyValueState<String, Long>> {
         private KeyValueState<String, Long> state;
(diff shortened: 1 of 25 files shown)


examples/storm-starter/src/jvm/storm/starter/StatefulWindowingTopology.java
```diff
@@ -0,0 +1,110 @@
...
+        @Override
+        public void prepare(Map stormConf, TopologyContext context, OutputCollector collector) {
+            this.collector = collector;
+        }
+
+        @Override
+        public void initState(KeyValueState<String, Long> state) {
+            this.state = state;
+            sum = state.get("sum", 0L);
+            System.out.println("initState with state [" + state + "] current sum [" + sum + "]");
```
ptgoetz: Could we use slf4j here instead of `System.out`?

storm-core/src/jvm/org/apache/storm/state/StateProvider.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+public interface StateProvider {
+    /**
+     * Returns a new state instance. Each state belongs unique namespace which is typically
+     * the componentid-task of the task, so that each task can have its own unique state.
+     *
+     * @param namespace a namespace of the state
+     * @param stormConf the storm topology configuration
+     * @param context   the {@link TopologyContext}
+     * @return a previously saved state instance
```
ptgoetz: Minor documentation nit: Doesn't it return a previously saved state if one exists, and otherwise return a new instance?

ptgoetz:
+1 aside from 2 very minor nits, but the feature is sizable so I'd like to see additional committer reviews.

I'd also like to see this included in the 1.0 release. It does touch clojure code (`executor.clj`), but in a very minor way.

Parth-Brahmbhatt:
+1.

arunmahadevan:
@ptgoetz @Parth-Brahmbhatt I made a minor fix to exclude checkpoint stream id from the component's input streams. https://github.com/arunmahadevan/storm/commit/596dc3ecafccd4b9e6ec523ef81832f007d1dee6

Could you take a look and merge if things look ok ?

ptgoetz:
Still +1. Since there was a recent code change we should hold off on merging until there's been enough time for review.

PR #966, opened 2015-12-22
[STORM-1401] removes multilang-test

d2r:
This test takes over 2 minutes to run, and it does not validate correctness of the multilang code.  It basically runs the topologies for a while and passes if they do not crash. Since they are now causing repeated failures in automated testing, there is not a lot of value in keeping them as-is.

We should cover multilang code as well as the Shell* classes in unit testing.  So I will create a follow-on issue to cover these with tests.


ptgoetz:
+1

@d2r Did you create a follow-up JIRA for new multi-lang tests?

d2r:
> @d2r Did you create a follow-up JIRA for new multi-lang tests?

Yeah, I linked it in STORM-1401 but not here.  It is [STORM-1410](https://issues.apache.org/jira/browse/STORM-1410)

HeartSaVioR:
+1
It would be more useful to make actual unit tests as you stated.

revans2:
+1

PR #970, opened 2015-12-23
[STORM-1412] null check should be done in the first place

vesense:
https://issues.apache.org/jira/browse/STORM-1412


ptgoetz:
+1

revans2:
+1

PR #978, opened 2015-12-25
[STORM-1420] solr CountBasedCommit should reset count=0 after commited

vesense:
https://issues.apache.org/jira/browse/STORM-1420


hmcl:
@vesense why is this change necessary?

revans2:
@hmcl and @vesense these changes are not necessary.  I pointed out here https://github.com/apache/storm/pull/977#issuecomment-169101930 that the commit API should probably be changed, because if we force a commit, things can get off from each other.  Somehow that turned into this.  I am -1 on this unless we also get changes to the Committer API too.

vesense:
Thanks @revans2 @hmcl for your explaining. These changes are not necessary. I will close this issue as 'Not A Problem'.

PR #979, opened 2015-12-26
STORM-1422 Remove broken example from the tutorial

sethgrid:
The tutorial code was broken giving an error. See http://stackoverflow.com/questions/34467446/apache-storm-classnotfoundexception-on-storm-starter. I did not alter example 2 because I have not verified that alteration.

### DIFF
--- a/examples/storm-starter/README.markdown
+++ b/examples/storm-starter/README.markdown
@@ -72,23 +72,6 @@ the Maven command to build and run storm-starter (see below), Maven will then be
 of Storm in this local Maven repository at `$HOME/.m2/repository`.
 
 
-## Running topologies with Maven
-
-> Note: All following examples require that you run `cd examples/storm-starter` beforehand.
-
-storm-starter topologies which don't use multilang feature can be run with the maven-exec-plugin. 
-For example, to compile and run `ExclamationTopology` in local mode, use the command:
-
-    $ mvn compile exec:java -Dstorm.topology=storm.starter.ExclamationTopology
-
-You can also run clojure topologies with Maven:
-
-    $ mvn compile exec:java -Dstorm.topology=storm.starter.clj.word_count
-
-In Windows parameter should be quoted, like this:
-
-    $ mvn compile exec:java "-Dstorm.topology=storm.starter.clj.word_count"
-
 ## Packaging storm-starter for use on a Storm cluster
 
 You can package a jar suitable for submitting to a Storm cluster with the command:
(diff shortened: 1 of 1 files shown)


examples/storm-starter/README.markdown
```diff
@@ -104,8 +87,8 @@ Example filename of the uberjar:
 
 You can submit (run) a topology contained in this uberjar to Storm via the `storm` CLI tool:
 
-    # Example 1: Run the RollingTopWords in local mode (LocalCluster)
```
sethgrid: `RollingTopWords` does not function either.

examples/storm-starter/README.markdown
```diff
@@ -72,23 +72,6 @@ the Maven command to build and run storm-starter (see below), Maven will then be
 of Storm in this local Maven repository at `$HOME/.m2/repository`.
 
 
-## Running topologies with Maven
```
sethgrid: None of the following examples work. See the linked SO question.

vesense:
Thanks @sethgrid Can you file a JIRA for this?

sethgrid:
@vesense after signing up for a JIRA account, I created the following. I hope it was set up properly: https://issues.apache.org/jira/browse/STORM-1422

d2r:
> @vesense after signing up for a JIRA account, I created the following. I hope it was set up properly: https://issues.apache.org/jira/browse/STORM-1422

Looks good.  Can you edit the title of this pull request to include `STORM-1422` so they are linked?

sethgrid:
@d2r updated title as per your request

examples/storm-starter/README.markdown
d2r: @ptgoetz, what do you think of this change to remove these instructions?

What is the use case for running topologies using maven?

d2r:
@ptgoetz, what do you think of this change to remove these instructions? Asking because I see your name on that section of the README.

What is the use case for running topologies using maven?

It seems since storm-core is not in compile scope, it is not on the classpath for the exec plugin?

ptgoetz:
@d2r I'm fine with updating the instructions. If I remember correctly (it's been a while ;) ), it was to make it very easy for new users to see a topology run in local mode.

+1 for the change.

d2r:
Thanks @ptgoetz.

@sethgrid, I see also that this changes away from RollingTopWords. If RollingTopWords does not work properly, could you create a new Jira issue to fix that separately?  We do not have to fix it with this change.

+1

PR #982, opened 2015-12-29
[STORM-1425] Tick tuples must be acked like normal tuples

vesense:
https://issues.apache.org/jira/browse/STORM-1425


arunmahadevan:
I don't think tick tuples needs to be acked. They are inserted directly into the receive-queue of the executor at specified intervals.

vesense:
@arunmahadevan Some previous comments can be found here https://github.com/apache/storm/pull/275#issuecomment-93690832

ptgoetz:
+1

revans2:
+1

PR #983, opened 2015-12-29
[STORM-1426] keep backtype.storm.tuple.AddressedTuple and delete duplicated backtype.storm.messaging.AddressedTuple

caofangkun:


ptgoetz:
+1 Confirmed that `back type.storm.messaging.AddressedTuple` is not used.

revans2:
+1 my bad.  Thanks for cleaning this up for me.  I guess I accidentally put two in somehow.

PR #987, opened 2016-01-04
STORM-1429: LocalizerTest fix

revans2:


revans2:
The Travis failure in this case is unrelated and is from storm-kafka

HeartSaVioR:
+1

Btw, would 100ms be enough to work with Travis CI?

revans2:
@HeartSaVioR I tested this on my own under-powered linux VM where I was seeing the issue, so I think it is.  The delete has already happened on the same thread, in a method call just above it.  I am not sure why Java and/or the OS is reporting that the file is still there, but it did occasionally.

knusbaum:
+1

PR #989, opened 2016-01-05
AvroGenericRecordBolt instead of SequenceFileBolt

johanwitters:
Changing the **example code** and use **AvroGenericRecordBolt**  instead of
**SequenceFileBolt**…

Section “Support for Avro Files” starts by explained the use of org.apache.storm.hdfs.bolt.AvroGenericRecordBolt for Avro files.
However, the example code uses SequenceFileBolt instead, and that
doesn’t work, obviously because the method withSchemaAsString isn’t
part of SequenceFileBolt, but part of
org.apache.storm.hdfs.bolt.AvroGenericRecordBolt instead. So, the
change is nothing but changing the name of class SequenceFileBolt and
replace it with the correct AvroGenericRecordBolt.

### DIFF
--- a/external/storm-hdfs/README.md
+++ b/external/storm-hdfs/README.md
@@ -295,7 +295,7 @@ The `org.apache.storm.hdfs.bolt.AvroGenericRecordBolt` class allows you to write
         // create sequence format instance.
         DefaultSequenceFormat format = new DefaultSequenceFormat("timestamp", "sentence");
 
-        SequenceFileBolt bolt = new SequenceFileBolt()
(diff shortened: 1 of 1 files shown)


dossett:
I would be +1 on this change if the PR were reopened.  Thank you for catching my sloppy error.

johanwitters:
Always happy to receive +1s :-)

revans2:
+1

revans2:
Since this is a comment change, we don't need to wait the full 24 hours to pull this in.

PR #991, opened 2016-01-05
STORM-1406: Add MQTT Support

ptgoetz:
https://issues.apache.org/jira/browse/STORM-1406


ptgoetz:
Initial commit for code review. Further documentation to follow.

external/storm-mqtt/core/src/main/java/org/apache/storm/mqtt/common/MqttPublisher.java
```diff
@@ -0,0 +1,101 @@
...
+    public MqttPublisher(MqttOptions options, KeyStoreLoader keyStoreLoader, boolean retain){
+        this.retain = retain;
+        this.options = options;
+        this.keyStoreLoader = keyStoreLoader;
+        SslUtils.checkSslConfig(this.options.getUrl(), keyStoreLoader);
+        this.qos = MqttUtils.qosFromInt(this.options.getQos());
+    }
+
+    public void publish(MqttMessage message, boolean retain) throws Exception {
+        this.connection.publish(message.getTopic(), message.getMessage(), this.qos, this.retain);
```
satishd: passed 'retain' argument is ignored. You may want to remove that parameter as it is not used or  
you can introduce one more method without 'retain' argument and pass 'retain' object received in constructor.

external/storm-mqtt/core/src/main/java/org/apache/storm/mqtt/common/MqttPublisher.java
```diff
@@ -0,0 +1,101 @@
...
+    public void connectMqtt(String clientId) throws Exception {
+        MQTT client = new MQTT();
+        URI uri = URI.create(this.options.getUrl());
+
+        client.setHost(uri);
+        if(!uri.getScheme().toLowerCase().equals("tcp")){
+            client.setSslContext(SslUtils.sslContext(uri.getScheme(), this.keyStoreLoader));
+        }
+        client.setClientId(clientId);
+        LOG.info("MQTT ClientID: " + client.getClientId().toString());
```
satishd: Can we use formatted strings for logging instead of stringbuilders?

external/storm-mqtt/core/src/main/java/org/apache/storm/mqtt/MqttLogger.java
```diff
@@ -0,0 +1,45 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger(MqttLogger.class);
+
+    @Override
+    public void debug(String message, Object... args) {
+        LOG.debug(String.format(message, args));
+    }
+
+    @Override
+    public void onSend(MQTTFrame frame) {
+        super.onSend(frame);
```
satishd: You may want to remove this method as it calls only super.

external/storm-mqtt/core/src/main/java/org/apache/storm/mqtt/MqttLogger.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    @Override
+    public void onReceive(MQTTFrame frame) {
+        super.onReceive(frame);
```
satishd: You may want to remove this method as it calls only super.

external/storm-mqtt/core/src/main/java/org/apache/storm/mqtt/common/MqttUtils.java
```diff
@@ -0,0 +1,44 @@
...
+        QoS qos = null;
+        switch(i) {
+            case 0:
+                qos = QoS.AT_MOST_ONCE;
+                break;
+            case 1:
+                qos = QoS.AT_LEAST_ONCE;
+                break;
+            case 2:
+                qos = QoS.EXACTLY_ONCE;
```
satishd: Why are we taking int and converting into Qos enum instance? Can not we use Qos enum directly in Options instead of int? enum is serializable, so it can be passed as argument to spout and bolt constructors.

external/storm-mqtt/core/src/main/java/org/apache/storm/mqtt/common/MqttOptions.java
```diff
@@ -0,0 +1,334 @@
...
+    private long reconnectDelay = 10;
+    private long reconnectDelayMax = 30*1000;
+    private double reconnectBackOffMultiplier = 2.0f;
+    private long reconnectAttemptsMax = -1;
+    private long connectAttemptsMax = -1;
+
+    private String userName = "";
+    private String password = "";
+
+    private int qos = 1;
```
satishd: Why are we avoiding Qos instance and using int? same for willQos.

external/storm-mqtt/core/src/test/java/org/apache/storm/mqtt/StormMqttIntegrationTest.java
```diff
@@ -0,0 +1,132 @@
...
+        MqttOptions options = new MqttOptions();
+        options.setTopics(Arrays.asList("/mqtt-topology"));
+        options.setCleanConnection(false);
+        MqttSpout spout = new MqttSpout(new StringMessageMapper(), options);
+
+        MqttBolt bolt = new MqttBolt(options, new MqttTupleMapper() {
+            @Override
+            public MqttMessage toMessage(ITuple tuple) {
+                LOG.info("Received: " + tuple);
+                return new MqttMessage("/integration-result", "hello mqtt".getBytes());
```
satishd: You may want to extract constants for topic names like "/integration-result", "/mqtt-topology"  as they are used at multiple places.

satishd:
+1 LGTM.

external/storm-mqtt/core/src/main/java/org/apache/storm/mqtt/bolt/MqttBolt.java
```diff
@@ -0,0 +1,108 @@
...
+    public void execute(Tuple input) {
+        //ignore tick tuples
+        if(!TupleUtils.isTick(input)){
+            MqttMessage message = this.mapper.toMessage(input);
+            try {
+                this.publisher.publish(message, this.retain);
+                this.collector.ack(input);
+            } catch (Exception e) {
+                LOG.warn("Error publishing MQTT message. Failing tuple.", e);
+                // should we fail the tuple or kill the worker?
```
HeartSaVioR: Maybe we want to add `collector.reportError(e);`

external/storm-mqtt/core/src/main/java/org/apache/storm/mqtt/spout/MqttSpout.java
```diff
@@ -0,0 +1,291 @@
...
+            connectMqtt();
+        } catch (Exception e) {
+            this.collector.reportError(e);
+            throw new RuntimeException("MQTT Connection failed.", e);
+        }
+
+    }
+
+    private void connectMqtt() throws Exception {
+        MQTT client = new MQTT();
```
HeartSaVioR: Code block from L119 to L148 is duplicated by MqttPublisher.connectMqtt(), L65 to L93, except setting up client id.
We may want to extract duplicated codes to method of another new (util?) class.

external/storm-mqtt/core/pom.xml
```diff
@@ -0,0 +1,153 @@
...
+      <groupId>org.apache.storm</groupId>
+      <artifactId>flux-core</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.fusesource.mqtt-client</groupId>
+      <artifactId>mqtt-client</artifactId>
+      <version>1.10</version>
+    </dependency>
+    <dependency>
```
HeartSaVioR: Maybe we don't need to add storm-hdfs as dependency since it is not used.

external/storm-mqtt/core/pom.xml
```diff
@@ -0,0 +1,153 @@
...
+      </dependency>
+
+      <dependency>
+      <groupId>org.apache.storm</groupId>
+      <artifactId>storm-core</artifactId>
+      <version>${project.version}</version>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.storm</groupId>
```
HeartSaVioR: Moving dependency of flux-core to mqtt-example module would be more sense.

HeartSaVioR:
As we know, flux-example contains other external modules.
In this PR storm-mqtt-example contains flux-core, which direction is opponent.
IMHO making it consistent would be more appreciated.

HeartSaVioR:
Left some comments, but looks great overall.
Actually I don't know about MQTT protocol for now, but mqtt-client abstracts it so I didn't have inconvenience to review this PR. Just I can't test it with real IoT machine.

external/storm-mqtt/core/src/main/java/org/apache/storm/mqtt/common/MqttUtils.java
ptgoetz: I wanted to shield users from the underlying MQTT client implementation. The `QoS` enum is in the `org.fusesource.mqtt.client` package, which users would have to import. If for some reason we switched the MQTT client implementation, it would break existing user code. With this approach we can avoid that.

external/storm-mqtt/core/src/main/java/org/apache/storm/mqtt/common/MqttOptions.java
ptgoetz: See above comment.

external/storm-mqtt/core/src/main/java/org/apache/storm/mqtt/MqttLogger.java
ptgoetz: Will do.

external/storm-mqtt/core/src/main/java/org/apache/storm/mqtt/bolt/MqttBolt.java
ptgoetz: Good call.

external/storm-mqtt/core/pom.xml
ptgoetz: Good catch. That a remnant from the original example that used HDFS.

external/storm-mqtt/core/src/main/java/org/apache/storm/mqtt/common/MqttUtils.java
HeartSaVioR: I'm with the approach what @ptgoetz is intended.
But it is still not intuitive for user side, since that numbers doesn't make sense to match actual meaning of QoS.
If we guarantee that we always support AT_MOST_ONCE, AT_LEAST_ONCE, EXACTLY_ONCE as QoS (I mean we will not drop QoS when switching the MQTT client), we can make our own enum to wrap.
If those numbers are a part of MQTT protocol then I'm fine to leave as it is.

ptgoetz: Those number are part of the MQTT protocol and are well known.

HeartSaVioR: Slightly late. ;) 
http://www.hivemq.com/blog/mqtt-essentials-part-6-mqtt-quality-of-service-levels
Then I'm +1 to MqttUtils.qosFromInt().

satishd: That makes sense. Assuming all other options are also part of mqtt protocol. So, there is no real dependency of underlying client for user APIs.

external/storm-mqtt/core/src/main/java/org/apache/storm/mqtt/trident/MqttPublishFunction.java
```diff
@@ -0,0 +1,84 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import storm.trident.operation.BaseFunction;
+import storm.trident.operation.TridentCollector;
+import storm.trident.operation.TridentOperationContext;
+import storm.trident.tuple.TridentTuple;
+
+import java.util.Map;
+
+public class MqttPublishFunction extends BaseFunction {
```
arunmahadevan: This looks like more of a trident sink than an pure function. So it might be better to follow the `StateUpdater/StateFactory` pattern (like `HdfsState` and `HdfsStatefactory`) and be used with `partitionPersist`.

ptgoetz: I thought about that distinction. Ultimately, I decided upon a function based on the following criteria:

1, There is no bulk publishing mechanism with MQTT, so there wouldn't be any performance benefit in processing tuples in batch.
2. Aside from publishing a message with a QoS of 2, we don't have any way to truly determine if it reached all subscribers, only the broker, and if the broker dies...??
3. Exactly once is virtually impossible with MQTT in a distributed environment. 
4. Most MQTT use cases that need to scale avoid QoS 2 because of the overhead.

That being said, I'm perfectly open to exploring a MQTT State implementation if it would be useful.

arunmahadevan: Since the function does not emit anything, it cannot be chained with further operations like normal functions and it appeared more like a sink. Hence thought it might be more apt for it to be implemented as a state instead of a function.

Agree that its difficult to do exactly once, but many other state implementations also does not guarantee exactly once. Leave it to you to decide the right approach.

ptgoetz:
@satishd, @arunmahadevan, @HeartSaVioR Thanks for your reviews. I believe I've addressed all your comments. I've also added documentation.

arunmahadevan:
+1 you might want to squash the commits. I see one commit from revans2 as well, is it due to the upmerge?

ptgoetz:
@arunmahadevan Thanks. Yes, the revans2 commit is due to upmerge. I'm happy to squash some commits, but some are there to facilitate merging to multiple branches.

We don't currently have a policy regarding squashing commits, managing history, etc.. That's probably worthy of a thread on dev@ to come up with a convention...

Parth-Brahmbhatt:
I think you need to add this module to incubator-storm/storm-dist/binary/src/main/assembly/binary.xml . I am +1 once that is done.

ptgoetz:
@Parth-Brahmbhatt Good catch. Will do.

ptgoetz:
@Parth-Brahmbhatt Done.

Parth-Brahmbhatt:
+1.

PR #994, opened 2016-01-07
[STORM-1499] Fix Kafka spout to maintain backward compatibility

arunmahadevan:
STORM-1220 introduced some changes where in one place it passes ByteBuffer as is instead of byte[].
Existing bolts (0.10.0) expects a byte[] and fails with
"java.lang.RuntimeException: java.lang.ClassCastException: java.nio.HeapByteBuffer cannot be cast to [B "
This patch addresses the issue by emiting byte[] instead of ByteBuffer.

### DIFF
--- a/storm-core/src/jvm/backtype/storm/spout/RawMultiScheme.java
+++ b/storm-core/src/jvm/backtype/storm/spout/RawMultiScheme.java
@@ -21,6 +21,7 @@ import java.nio.ByteBuffer;
 import java.util.List;
 
 import backtype.storm.tuple.Fields;
+import backtype.storm.utils.Utils;
 
 
 import static backtype.storm.utils.Utils.tuple;
@@ -29,7 +30,7 @@ import static java.util.Arrays.asList;
 public class RawMultiScheme implements MultiScheme {
   @Override
   public Iterable<List<Object>> deserialize(ByteBuffer ser) {
-    return asList(tuple(ser));
+    return asList(tuple(Utils.toByteArray(ser)));
   }
 
   @Override


zhuoliu:
+1

revans2:
+1

revans2:
Looks like this is for https://issues.apache.org/jira/browse/STORM-1449 not 1499.

arunmahadevan:
Sorry for the typo, changed the subject.

priyank5485:
Just ran in to an issue where this fix does not work. Correct me if i am wrong. But a fat storm jar that is compiled with older version of storm-kafka will still use byte[] and the newer version of storm now uses ByteBuffer. Hence it throws an exception at runtime if storm is upgraded. I think upgrade of storm should still work with older topology jars.

revans2:
@priyank5485 you are correct.  I didn't look into this in great detail until now.  What is more if you submit using an old client you will get deserialization errors too, because the SerialId is not set.

STORM-1220 seems to fundamentally not be able to maintain that backwards compatibility.  So now the question is are the benefits of STORM-1220 worth the incompatibility?  To me it seems likely that they are, because it is just updating a dependency.

priyank5485:
@revans2 I am fine with the changes except one major concern pointed out by @Parth-Brahmbhatt I confirmed this with him and our main concern would be the rolling upgrade feature for topologies with kafka spout. That would not work anymore.

Parth-Brahmbhatt:
We can either just include this as a warning as part of Release Notes or we can roll this change back. I am fine with including just a Release Note to warn all users trying to upgrade unless others thing Rolling upgradability is important enough that we only brake it when people move between major versions.

arunmahadevan:
@priyank5485 @revans2 @Parth-Brahmbhatt the changes introduced by STORM-1220 has not yet gone out in any releases. It will go out in 1.0 but with the fix in this PR (which is also part of 1.0) the changes are backward compatible. i.e. the KafkaSpout would emit byte[] and not ByteBuffer as values in the tuples. So the old bolts (that expects a byte[]) would continue to work with the new KafkaSpout. 

I assume @priyank5485 got the issue when the jar was built from master branch before this PR got merged in and got bundled with the client. 

Correct me if I am missing something.

revans2:
It sounds like everyone is fine with the current set of incompatibilities, so it comes down to a documentation issue.  I looked at the storm-kafka docs and I don't see anywhere that this incompatability is mentioned.  In fact in the section that talks about the MultiSchema it needs to be updated to reflect the changes from STORM-1220.

Could someone on this list file a JIRA or just update the documentation to explain the incompatibility and how to upgrade?

arunmahadevan:
Filed JIRA and will update the doc - https://issues.apache.org/jira/browse/STORM-1486

There is no incompatibility since the emitted tuple values continue to be byte[]. Only the api signatures of the MultiScheme (and other schemes) have changed. It now accepts a ByteBuffer instead of byte[] but returns tuples with byte[]  as values as before.

revans2:
@arunmahadevan There are several incompatibilities.  Most of them don't matter because this is a part of the 1.x release, so rolling upgrades/etc will not be possible.

The one incompatibility that remains is that you cannot use a pre 1.x kafka spout on a 1.x cluster.  It is always good practice to maintain the versions the same, but in this case it is required.

KafkaUtils.generateTuples, calls the modified MultScheme.deserialize API making this a problem.

arunmahadevan:
@revans2 I see what you are saying. Will document this incompatibility.

HeartSaVioR:
Maintaining backward compatibility is bi-directional, which means just modifying core can break module's backward compatibility what @revans2 stated.
We should bear in mind of backward compatibility with core itself.
Adopting semver is great to go, and in fact preparing 1.0.0 is the chance to break backward compatibility if necessary. If we were deprecated some classes/methods, let's review.
I expect we can not have a chance to release another major version sooner (except 2.0, it is not for such change)

PR #995, opened 2016-01-08
[STORM-1430] ui worker checkboxes

d2r:
This adds new checkboxes to the Executors tables of the component pages. Selecting these will enable the debugging/profiling actions to be performed on multiple workers while making it much easier to choose workers—especially for topologies with hundreds of workers.

Look at the documentation to see what has changed. This is a little hard to test, because of [STORM-1452](https://issues.apache.org/jira/browse/STORM-1452). As a work-around, manually copy flight.bash to somewhere on the supervisor process's PATH and it should work.

This also cleans up some tooltip placement and removes some extra bindings from core.clj, and I put these in their own commits so it is clearer.


d2r:
The one test failure was netty load test. It looks unrelated.

kishorvpatil:
+1. LGTM. It is very helpful to be able to select executor(s) for action while looking at rest of the stats.

storm-core/src/clj/backtype/storm/ui/core.clj
```diff
@@ -1108,12 +1108,8 @@
   (GET "/api/v1/topology/:id/profiling/start/:host-port/:timeout"
        [:as {:keys [servlet-request]} id host-port timeout & m]
        (thrift/with-configured-nimbus-connection nimbus
```
revans2: These all seem to be missing `(populate-context! servlet-request)` We need to add it as the first thing in each of the REST requests

revans2:
Only one thing with security that looks like a merge error, not related to this, but fixing it here would be trivial.

revans2:
Oh +1, even without it, but if you aren't going to do it here let me know and I'll file a JIRA to fix it.

d2r:
> Oh +1, even without it, but if you aren't going to do it here let me know and I'll file a JIRA to fix it.

Good catch. Maybe we should fix this as part of [STORM-1452](https://issues.apache.org/jira/browse/STORM-1452). I made a comment pointing back here on that Jira.

revans2:
+1 for fixing it as a part of STORM-1452 and +1 for this pull request.

PR #997, opened 2016-01-08
[STORM-1457] Avoid collecting pending tuples if topology.debug is off

kishorvpatil:
This would help lower heap memory usage when ```topology.debug``` is off.

### DIFF
--- a/storm-core/src/clj/backtype/storm/daemon/executor.clj
+++ b/storm-core/src/clj/backtype/storm/daemon/executor.clj
@@ -497,6 +497,7 @@
         spouts (ArrayList. (map :object (vals task-datas)))
         rand (Random. (Utils/secureRandomLong))
         ^DisruptorQueue transfer-queue (executor-data :batch-transfer-queue)
+        debug? (= true (storm-conf TOPOLOGY-DEBUG))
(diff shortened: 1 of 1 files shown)


storm-core/src/clj/backtype/storm/daemon/executor.clj
```diff
@@ -572,7 +572,8 @@
                                            (do
                                              (.put pending root-id [task-id
                                                                     message-id
-                                                                    {:stream out-stream-id :values values}
+                                                                    {:stream out-stream-id 
+                                                                     :values (if (= true (storm-conf TOPOLOGY-DEBUG)) values nil)}
```
revans2: I would prefer to avoid calculating if TOPOLOGY-DEBUG is true each time.  This is code on the critical path so if we could do it once at the beginning of the thread function and reuse it, that would be great.

kishorvpatil: Agreed. Let me change it.

kishorvpatil:
@revans2  Moved calculation out of critical path.

revans2:
+1

PR #999, opened 2016-01-08
[STORM-1206] Reduce logviewer memory usage through directory stream

zhuoliu:
Using DirectoryStream to replace File.listFiles will save logviewer memory usage since it does not require loading all files' metadata into memory. This avoids potential memory usage problem in extreme case (e.g., you have millions of small files in your log directory).
Also, a multi-phase PQ-based sorting and cleaning scheme is introduced in DirectoryCleaner for replacing the global in-memory sorting.


revans2:
These changes look rather large, but the tests pass and the code looks good.  I assume you have done some manual testing so I am +1.

PR #1000, opened 2016-01-09
STORM-1160: Add hadoop-auth dependency needed for storm-core

priyank5485:
hadoop-auth is needed for storm ui in a secured cluster for
authentication. Adding the dependency back since it was removed in one
of the commits.


revans2:
+1 hopefully we get true integration tests soon so we can catch this earlier.

PR #1001, opened 2016-01-09
Storm SQL - Storm.py 'storm sql' startup script is not working

hmcl:


bin/storm.py
```diff
@@ -229,7 +231,7 @@ def jar(jarfile, klass, *args):
         daemon=False,
         jvmopts=JAR_JVM_OPTS + ["-Dstorm.jar=" + jarfile])
 
-def sql(sql_file, topo_nam):
```
satishd: minor nit: You may want to rename topo_name to topology_name. We can have JIRA to add a test which uses storm sql command.

bin/storm.py
```diff
@@ -229,7 +231,7 @@ def jar(jarfile, klass, *args):
...
+def sql(sql_file, topo_name):
```
satishd: minor nit: You may want to rename topo_name to topology_name. We can have JIRA to add a test which uses 'storm sql' command.

harshach:
this is a small fix , no need for a jira.

satishd:
@harshach JIRA is for adding a test which uses 'storm sql' command but not for minot nitpick about renaming argument name.

harshach:
@satishd makes sense . It would be good to have tests around storm.py to catch these bugs.

hmcl:
@satishd @harshach Created the following JIRA to track creating unit tests for Storm.py https://issues.apache.org/jira/browse/STORM-1461

satishd:
+1

revans2:
@hmcl the change itself looks OK, but I am not really sure what/how this fixes anything.  You rearranged the order of imports and renamed a variable.  Can you explain what was failing before and what is fixed now?

hmcl:
@revans2 without this fix the command line option 'storm sql' is broken. The reason it is broken is because prior to this change the method parameter 'topo_nam' is referred as 'topo_name' in the method body, hence the script does not run and this change is required. Initially I changed the parameter method parameter to 'topo_name', but following @satishd suggestion, we agreed on renaming 'topology_name'

revans2:
@hmcl ah I missed the one letter diff between them.  Thanks.  +1

PR #1006, opened 2016-01-12
STORM-1467: Switch apache-rat plugin off by default, but enable for Travis-CI

knusbaum:
I added an un-licensed file in to show that the CI will fail. I'll remove it before merge.


revans2:
I really would prefer to see this be a part of a profile.  So we can turn it on with something like `-Prat`

ptgoetz:
+1

PR #1008, opened 2016-01-13
STORM-1468: Merge changes from {master}/docs to {asf-site}

ptgoetz:
JIRA: https://issues.apache.org/jira/browse/STORM-1468


ptgoetz:
Assuming lazy consensus, but would appreciate review in case I missed anything.

ptgoetz:
See also #1009

revans2:
I am +1 on this.  Not having 2 copies is good and a lot better then what we have now, but I really would prefer to see a hybrid approach.  Some documentation is really tied to a release.  The REST API for example, where as other documentation is not, the list of contributors and users for example.  I really would prefer to see what I see on other sites, which is some common things but they point into release specific sections that are generated from the release branch, and are obviously tied to a release.

We could also use symlinks to denote stable releases vs releases that are no longer supported.

PR #1010, opened 2016-01-13
[STORM-1470] Applies shading to hadoop-auth, cleaner exclusions

d2r:
I would also like to see this applied to 1.x-branch, since it is a blocker.


revans2:
+1 pending travis

ptgoetz:
+1

PR #1011, opened 2016-01-13
[STORM-1470] Applies shading to hadoop-auth, cleaner exclusions

d2r:


ptgoetz:
Personally, I don't think we need separate pull requests for each branch yet (master and 1.x have not diverged enough to cause conflicts), as it kind of confuses things in terms of +1s and comments. It might be easier to add a note requesting that it also be applied to another branch, e.g. "I would also like to see this applied to the X branch."

If for some reason the patch couldn't be applied to the other branch(es), the committer doing the merge could either resolve the conflicts or request an additional pull request against the conflicted branch.

Thoughts?

d2r:
See #1010

PR #1012, opened 2016-01-13
[STORM-1452] Fixes profiling/debugging out of the box

d2r:
Ships the flight.bash script
Disables UI display of Profiling and Debugging if on windows
Changes worker.profiler.enabled to control only profiling, not other debugging actions
Changes default of worker.profiler.enabled to false
Adds missing UI REST API data.
Populates user context on profiler UI routes

I would also like to see this applied to the master branch.

### DIFF
--- a/storm-core/src/clj/org/apache/storm/ui/core.clj
+++ b/storm-core/src/clj/org/apache/storm/ui/core.clj
@@ -105,14 +105,6 @@
            (throw (AuthorizationException.
                    (str "UI request '" op "' for '" user "' user is not authorized")))))))))
 
-
-(defn assert-authorized-profiler-action
-  [op]
-  (if-not (*STORM-CONF* WORKER-PROFILER-ENABLED)
-    (throw (AuthorizationException.
-             (str "UI request for profiler action '" op "' is disabled.")))))
-
-
 (defn executor-summary-type
   [topology ^ExecutorSummary s]
   (component-type topology (.get_component_id s)))
@@ -863,6 +855,7 @@
                                       (.get_eventlog_host comp-page-info)
                                       (.get_eventlog_port comp-page-info)
                                       secure?)
+       "profilingAndDebuggingCapable" (not on-windows?)
        "profileActionEnabled" (*STORM-CONF* WORKER-PROFILER-ENABLED)
        "profilerActive" (if (*STORM-CONF* WORKER-PROFILER-ENABLED)
                           (get-active-profile-actions nimbus topology-id component)
@@ -920,6 +913,15 @@
 (defn get-user-name
   [servlet-request]
   (.getUserName http-creds-handler servlet-request))
+
+(defn json-profiling-disabled
+  "Return a JSON response communicating that profiling is disabled and
+  therefore unavailable."
+  [callback]
+  (json-response {"status" "disabled",
+                  "message" "Profiling is not enabled on this server"}
+                 callback
+                 :status 501))
 
 (defroutes main-routes
   (GET "/api/v1/cluster/configuration" [& m]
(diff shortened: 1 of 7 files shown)


ptgoetz:
Not really specific to this pull request, but I missed it before when the profiling functionality was originally added.

If profiling is enabled, we turn on the "Commercial Features" of the JVM:

`worker.profiler.childopts: "-XX:+UnlockCommercialFeatures -XX:+FlightRecorder`

The current Oracle Java license [1] states:

>Some of the packages described in the Installation of Oracle Java SE Product Editions section above install Commercial Features that are restricted to Oracle Java SE Advanced, Oracle Java SE Advanced Desktop and/or Oracle Java SE Suite. **This means that even if you download an Oracle Java SE package for free under the Java BCLA, you must separately license from Oracle (and pay the appropriate license fee) the right to use any Commercial Features, described in Table 1-1 below, included in these packages.**

If my interpretation of the license is correct, that means users must have purchased a license from Orace for each node that has profiling enabled. (Read the license [1] for more information and examples.)

I think that's probably okay, but we may want to document that somewhere (`defaults.yaml` maybe?), so users don't accidentally run afoul of Oracle's licensing terms.

[1] http://www.oracle.com/technetwork/java/javase/terms/products/index.html

d2r:
Yes, I can add something.

d2r:
@ptgoetz , added.  Do you think it is OK?

ptgoetz:
Works for me. Thanks @d2r.

+1

docs/documentation/ui-rest-api.md
```diff
@@ -548,6 +548,10 @@ Response fields:
 |boltStats.processLatency| String (double value returned in String format)  |Average time of the bolt to ack a message after it was received|
 |boltStats.acked| Long |Number of messages acked|
 |boltStats.failed| Long |Number of messages failed|
+|profilingAndDebuggingCapable| Boolean |true if there is support for Profiling and Debugging Actions|
+|profilActionEnabled| Boolean |true if worker profiling (Java Flight Recorder) is enabled|
```
revans2: Should this be profil**e**ActionEnabled?

storm-core/src/clj/org/apache/storm/ui/core.clj
```diff
@@ -863,6 +855,7 @@
                                       (.get_eventlog_host comp-page-info)
                                       (.get_eventlog_port comp-page-info)
                                       secure?)
+       "profilingAndDebuggingCapable" (not on-windows?)
```
revans2: Can we file a follow on JIRA to make this a bit cleaner?  Perhaps a script for different environments that we can check at run time if the script is there?

docs/documentation/ui-rest-api.md
d2r: Good catch! Will fix it.

revans2:
It looks like this needs to be upmerged, because the docs are no longer on master, so the REST API docs needs a separate pull request :(

storm-core/src/clj/org/apache/storm/ui/core.clj
d2r: Sure.  We are using this var in several places, and I did not want to change all of them to fix this.

d2r:
> It looks like this needs to be upmerged, because the docs are no longer on master, so the REST API docs needs a separate pull request :(

OK, I did not realize that.  I'll do so.

storm-core/src/clj/org/apache/storm/ui/core.clj
d2r: [STORM-1489](https://issues.apache.org/jira/browse/STORM-1489) created

ptgoetz:
@d2r You can just delete the .md file in question. I should have a pull request for the docs in a minute.

ptgoetz:
PR for moving the docs to `asf-site`: https://github.com/apache/storm/pull/1032

That includes the typo @revans2 mentioned.

d2r:
> PR for moving the docs to asf-site: #1032
> 
> That includes the typo @revans2 mentioned.

Thanks @ptgoetz !

OK I have rebased the branch to 1.x-branch, and I removed the .md in that process.

PR #1017, opened 2016-01-15
STORM-1476 Filter -c options from args and add them as part of storm.options

satishd:
Filter -c options from args and add them as part of storm.options


ptgoetz:
+1

harshach:
+1

PR #1018, opened 2016-01-15
Improve mention of aggregate() in the partitionAggregate() docs

F30:
This simplifies the mention of `CombinerAggregator` benefits in the `partitionAggregate()` docs and adds a reference to the specific documentation.

Reading the respective paragraph, it took me some time to figure out that `aggregate()` is covered in detail below. I originally thought that it was mostly undocumented and began reading the source.

The actual optimizations are described in the `aggregate()` section much more exhaustively, so I don't think they need to be covered here. Also, in the old wording it was unclear what the benefits should be compared to, because other aggregator types are only introduced afterwards.


knusbaum:
+1 I'm fine with this change. If you want to upmerge, I'll get it in. Otherwise I plan to close this in a few weeks or so.

F30:
It took a while, but I have now rebased the changes onto the current master.

Unfortunately, force-pushing those new commits auto-closed this PR because of missing common history with the old base branch. After adjusting the base branch, the diff has not been recomputed and the PR cannot be reopened. I often contacted GitHub support about this and they claim it's the intended behavior.

The changes now live on in #2362.

PR #1020, opened 2016-01-16
[STORM-1478] make bolt's getComponentConfiguration method cleaner/simpler

vesense:
This PR follow https://github.com/apache/storm/pull/977#discussion_r49160913 @hmcl 's comment:
>I noticed that the body of this method is exactly the same for AbstractHdfsBolt. Furthermore, it's very likely that many Storm connector bolts will use the same code. Can you please write a helper method in TupleUtils and call that method here and AbstractHdfsBolt.


HeartSaVioR:
+1

Btw, there's a typo to method name, `putTickFreqencyIntoComponentConfig` should be `putTickFrequencyIntoComponentConfig`.
We didn't have any official releases including this method and it's used internally, I think we're fine to change its name safely.
Could you please correct a typo? Thanks!

vesense:
Nice catch. Thanks @HeartSaVioR Fixed.

ptgoetz:
+1

HeartSaVioR:
+1 again.

arunmahadevan:
+1

PR #1021, opened 2016-01-16
[STORM-1479] use a simple implemention for IntSerializer

vesense:


arunmahadevan:
Changes looks good. But wonder why isn't there a corresponding IntDeserializer?

vesense:
@arunmahadevan 
This class is used by StormSQL.
>In StormSQL data is represented by external tables. Users can specify data sources using the CREATE EXTERNAL TABLE statement.
* kafka as data source
use XXXSerializer for data ser.
* kafka as data sink
use `org/apache/storm/spout/MultiScheme.java` for data deser.

arunmahadevan:
MulitScheme implementations appear to deserialize raw bytes or String, not Integer.

revans2:
+1 looks OK to me.

PR #1023, opened 2016-01-16
[STORM-1481] avoid Math.abs(Integer) get a negative value

vesense:
https://issues.apache.org/jira/browse/STORM-1481

### DIFF
--- a/storm-core/src/jvm/org/apache/storm/trident/partition/IndexHashGrouping.java
+++ b/storm-core/src/jvm/org/apache/storm/trident/partition/IndexHashGrouping.java
@@ -20,15 +20,17 @@ package org.apache.storm.trident.partition;
 import org.apache.storm.generated.GlobalStreamId;
 import org.apache.storm.grouping.CustomStreamGrouping;
 import org.apache.storm.task.WorkerTopologyContext;
+import org.apache.storm.utils.Utils;
+
 import java.util.Arrays;
 import java.util.List;
 
 public class IndexHashGrouping implements CustomStreamGrouping {
     public static int objectToIndex(Object val, int numPartitions) {
-        if(val==null) return 0;
-        else {
-            return Math.abs(val.hashCode()) % numPartitions;
+        if(val == null) {
+            return 0;
         }
+        return Utils.toPositive(val.hashCode()) % numPartitions;
     }
     
     int _index;
--- a/storm-core/src/jvm/org/apache/storm/utils/Utils.java
+++ b/storm-core/src/jvm/org/apache/storm/utils/Utils.java
@@ -1369,5 +1369,18 @@ public class Utils {
         }
         return topologyInfo;
     }
+
+    /**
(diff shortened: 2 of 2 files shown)


zhuoliu:
+1.
The fix is necessary due to follows, which can cause problem during grouping.
Integer.MAX_VALUE % 1 = 0
Integer.MAX_VALUE % 2 = 0
Integer.MAX_VALUE % 3 = -2
Integer.MAX_VALUE % 4 = 0
Integer.MAX_VALUE % 5 = -3
Integer.MAX_VALUE % 6 = -2
Integer.MAX_VALUE % 7 = -2
Integer.MAX_VALUE % 8 = 0

arunmahadevan:
why not `Math.abs(val.hashCode() % numPartitions)` ?

revans2:
OK so Math.abs(Integer.MIN_VALUE) actually returns Integer.MIN_VALUE because of an integer overflow error.  That is a good corner case to remember in the future.

@arunmahadevan `Math.abs(val.hasCode() % numPartitions)` would work too.  The only reason to do one over the other would be performance vs readability.  In my very unscientific test
```
public class Test {
    public static int toPositive(int number) {
        return number & Integer.MAX_VALUE;
    }

    public static void main(String [] args) {
        int count = Integer.valueOf(args[0]);
        long start = System.nanoTime();
        for (int i = 0; i < count; i++) {
            int ret = Math.abs(i);
            //int ret = toPositive(i);
        }
        long end = System.nanoTime();
        System.out.println(end+" - "+ start+" = "+(end-start)+" rate: "+(((double)count)/(end-start)));
    }
}
```
I get about 291 ops per nanosecond for toPositive

`1453216134803090000 - 1453216134799657000 = 3433000 rate: 291.29041654529567`

and 289 ops per nanosecond for Math.abs
`1453216159561620000 - 1453216159558157000 = 3463000 rate: 288.7669650591972`

To me it is such a small difference About 1% over a part of the code where computing the hash code dominates that it should come down to readability.  Both look equally as readable to me (because of the comments on toPositive).

I am +1 for this change.

revans2:
@vesense I don't see any indication in the JIRA which branches you want this fix merged into.  master and 1.x seem simple enough, but it appears to be a bug in 10.x and 9.x releases too.  I'm not sure it is critical enough to go back to either of them, but I wanted to check with you before merging this in.

ptgoetz:
+1

I wouldn't mind seeing this go back to 1.x and 0.10.x (0.10.x has a few other minor fixes). But I don't think we need to support 0.9.x too much longer.

revans2:
Because the files have change locations porting to 0.10.x is going to need a separate pull request (git is not happy)

I will merge this in for master and 1.x-branch

ptgoetz:
@revans2 I'll take care of the merge to 0.10.x as soon as this is in master.

revans2:
@ptgoetz sounds good just running a sanity check with the tests before merging it in.

PR #1027, opened 2016-01-19
[STORM-1485] DRPC Connectivity Issues

kishorvpatil:
- ZK-Login-From-AutoTGT thread should not die if TGT in `AutoTGT` has expired. ( It gives grace 10 minutes before killing worker.) This required copying and modified `org.apache.zookeeper.Login` class. 
- `DRPCClient` should not give up connecting to DRPC after n-retries.
- `ReturnResults` Bolt should retry connection on failure to send results to DRPC.

### DIFF
--- a/storm-core/src/jvm/org/apache/storm/messaging/netty/KerberosSaslNettyClient.java
+++ b/storm-core/src/jvm/org/apache/storm/messaging/netty/KerberosSaslNettyClient.java
@@ -35,7 +35,6 @@ import javax.security.auth.login.LoginException;
 import javax.security.sasl.Sasl;
 import javax.security.sasl.SaslClient;
 import javax.security.sasl.SaslException;
-import org.apache.zookeeper.Login;
 import org.apache.zookeeper.server.auth.KerberosName;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -84,7 +83,7 @@ public class KerberosSaslNettyClient {
             Configuration.setConfiguration(login_conf);
             //now login
             LOG.debug("Trying to login.");
-            Login login = new Login(jaas_section, ch);
+            ZookeeperSaslLogin login = new ZookeeperSaslLogin(jaas_section, ch);
             subject = login.getSubject();
             LOG.debug("Got Subject: {}", subject.toString());
         } catch (LoginException ex) {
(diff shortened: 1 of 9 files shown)


storm-core/src/jvm/org/apache/storm/messaging/netty/KerberosSaslNettyClient.java
```diff
@@ -84,7 +83,7 @@ public KerberosSaslNettyClient(Map storm_conf, String jaas_section) {
             Configuration.setConfiguration(login_conf);
             //now login
             LOG.debug("Trying to login.");
-            Login login = new Login(jaas_section, ch);
+            ZookeeperSaslLogin login = new ZookeeperSaslLogin(jaas_section, ch);
```
revans2: I don't really like Zookeeper being a part of this classes name.  I also don't really like Sasl in there, because even though we use it for Sasl, there is nothing Sasl specific in it.

revans2:
I just have the one minor comment about the name of the class that we pulled over from Zookeeper.

+1 even with the name being a bit confusing to me.

kishorvpatil:
@revans2 I reverted the name to `Login`.

revans2:
+1

storm-core/src/jvm/org/apache/storm/messaging/netty/Login.java
```diff
@@ -0,0 +1,411 @@
...
+
+    private LoginContext login = null;
+    private String loginContextName = null;
+    private String keytabFile = null;
+    private String principal = null;
+
+    private long lastLogin = 0;
+
+    /**
+     * LoginThread constructor. The constructor starts the thread used
```
ptgoetz: Minor nit: `LoginThread` --> `Login`

ptgoetz:
One very minor nit.

+1

storm-core/src/jvm/org/apache/storm/messaging/netty/Login.java
kishorvpatil: @ptgoetz  I trying not to deviate a lot from org.apache.zookeeer.Login code. If I can not push the changes to zookeeper and this becomes Storm only code then, we may get ride of unwanted comments etc... 
If you still prefer, I am more than willing to make these minor changes in code comments..

revans2: I'm not sure we want to push this back to zookeeper.  Login feels like it was an internal class not meant for others to use.  We probably never should have used it, and forking it is the right thing to do long term.  Even if we do push a fix back to zookeeper.

kishorvpatil: Ok will change it.

ptgoetz:
Still +1

revans2:
Still +1

PR #1030, opened 2016-01-20
[STORM-1227] port org.apache.storm.config to java

zhuoliu:
1. Port "org.apache.storm.config" to java at "org.apache.storm.config.ConfigUtils".
2. Update all the callings to the config functions.
3. Mock the java static functions for test purposes.
4. Address issues in unit and manual tests.
5. Deleted most functions in config.clj except a few: cluster-mode (used in multimethod),  ALL-CONFIGS, clojure-config-name, mk-stats-sampler (depending on util and quite "clojury") and sampling-rate.


storm-core/src/clj/org/apache/storm/config.clj
```diff
@@ -18,14 +18,14 @@
...
+  (:import [org.apache.storm.utils Utils LocalState ConfigUtils])
   (:import [org.apache.storm.validation ConfigValidation])
   (:import [org.apache.commons.io FileUtils])
   (:require [clojure [string :as str]])
   (:use [org.apache.storm log util]))
 
-(def RESOURCES-SUBDIR "resources")
-(def NIMBUS-DO-NOT-REASSIGN "NIMBUS-DO-NOT-REASSIGN")
+;; (def RESOURCES-SUBDIR "resources")
+;; (def NIMBUS-DO-NOT-REASSIGN "NIMBUS-DO-NOT-REASSIGN")
```
revans2: If these are not used we should delete them, not comment them out.

storm-core/src/clj/org/apache/storm/config.clj
```diff
@@ -41,11 +41,30 @@
   (dofor [f (seq (.getFields Config))]
          (.get f nil)))
 
-
+;; TODO this function and its callings will be replace when nimbus and supervisor move to Java
```
revans2: Don't need the TODO.  Please file a separate JIRA to remove the code when it is no longer needed.  You can leave the comment, but preferably with the JIRA number instead.

storm-core/src/clj/org/apache/storm/config.clj
```diff
@@ -41,11 +41,30 @@
...
+  (even-sampler (sampling-rate conf)))
+
+;; TODO this function is to be replaced with Java version when util.clj is in
+(defn read-default-config
+  []
+  (clojurify-structure (Utils/readDefaultConfig)))
+
+
+
+(comment
```
revans2: Why is all of this code commented out and not deleted?

storm-core/src/clj/org/apache/storm/config.clj
```diff
@@ -41,11 +41,30 @@
...
+
+;; TODO this function together with sampling-rate are to be replaced with Java version when util.clj is in
+(defn mk-stats-sampler
+  [conf]
+  (even-sampler (sampling-rate conf)))
+
+;; TODO this function is to be replaced with Java version when util.clj is in
+(defn read-default-config
+  []
+  (clojurify-structure (Utils/readDefaultConfig)))
```
revans2: Are these functions still needed?

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
```diff
@@ -0,0 +1,711 @@
...
+            } else {
+                dir = FILE_SEPARATOR + "logs";
+            }
+        }
+        try {
+            LOG_DIR = new File(dir).getCanonicalPath();
+        } catch (IOException ex) {
+            throw new IllegalArgumentException("Illegal storm.log.dir in conf: " + dir);
+        }
+    }
```
revans2: Exceptions in static code blocks are really confusing to debug.  I personally would prefer to see a static function to get the log dir, that caches the saved result.

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
```diff
@@ -0,0 +1,711 @@
...
+                LOG.error(e.getMessage(), e);
+            }
+        }
+        return ret;
+    }
+
+    public static String clusterMode(Map conf) {
+        String mode = (String) conf.get(Config.STORM_CLUSTER_MODE);
+        return mode;
+
```
revans2: Extra line

storm-core/src/clj/org/apache/storm/config.clj
zhuoliu: Will delete.

storm-core/src/clj/org/apache/storm/config.clj
zhuoliu: Will address.

storm-core/src/clj/org/apache/storm/config.clj
zhuoliu: Will delete.

storm-core/src/clj/org/apache/storm/config.clj
zhuoliu: These 2 functions are actually depending on util.clj. I may just translate those two util functions within this pull request and delete these two.

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
zhuoliu: OK. Will turn it into a static function instead.

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
zhuoliu: Will delete.

storm-core/src/clj/org/apache/storm/config.clj
revans2: Either way is fine with me so long as they are not dropped.

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
```diff
@@ -0,0 +1,711 @@
...
+        if (localDir == null) {
+            return (stormHome + FILE_SEPARATOR + "storm-local");
+        } else {
+            if (new File(localDir).isAbsolute()) {
+                return localDir;
+            } else {
+                return (stormHome + FILE_SEPARATOR + localDir);
+            }
+        }
+
```
revans2: Extra line

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
```diff
@@ -0,0 +1,711 @@
...
+    }
+
+    public static String absoluteHealthCheckDir(Map conf) {
+        String stormHome = System.getProperty("storm.home");
+        String healthCheckDir = String.valueOf(conf.get(Config.STORM_HEALTH_CHECK_DIR));
+        if (healthCheckDir.equals("null")) {
```
revans2: Why the string "null" here and an actual null in absoluteStormLocalDir?

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
```diff
@@ -0,0 +1,711 @@
...
+        try {
+            FileUtils.forceMkdir(new File(ret));
+        } catch (IOException e) {
+            LOG.error("Failed to create dir " + ret, e);
+            throw e;
+        }
+        return ret;
+    }
+
+
```
revans2: Extra line

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
```diff
@@ -0,0 +1,711 @@
...
+        }
+    }
+
+    public static String masterLocalDir(Map conf) throws IOException {
+        String ret = String.valueOf(conf.get(Config.STORM_LOCAL_DIR)) + FILE_SEPARATOR + "nimbus";
+        try {
+            FileUtils.forceMkdir(new File(ret));
+        } catch (IOException e) {
+            LOG.error("Failed to create dir " + ret, e);
+            throw e;
```
revans2: Catching an exception to log it an rethrow it is not typically good practice.  Could we remove the catch/rethrow?  I also am confused why we do a mkdir here, but that is what the original code did so it is a good translation.  Could we file a JIRA to clean this up later?  Creating directories does not feel like something ConfigUtils should be doing.

revans2: Actually can you just make it generic about all of the forceMkdir calls in ConfigUtils?

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
```diff
@@ -0,0 +1,711 @@
...
+        }
+        return ret + FILE_SEPARATOR + "stormdist";
+    }
+
+    public static String stormTmpPath(String stormRoot) {
+        return stormRoot + FILE_SEPARATOR + "tmp";
+    }
+
+    /* Never get used TODO : delete it*/
+    public static String masterTmpDir(Map conf) throws IOException {
```
revans2: Yes lets delete it.

revans2: And the others that are not used.

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
```diff
@@ -0,0 +1,711 @@
...
+        return (stormRoot + FILE_SEPARATOR + "stormconf.ser");
+    }
+
+    public static String masterInbox(Map conf) throws IOException {
+        String ret = masterLocalDir(conf) + FILE_SEPARATOR + "inbox";
+        try {
+            FileUtils.forceMkdir(new File(ret));
+        } catch (IOException e) {
+            LOG.error("Failed to create dir " + ret, e);
+            throw e;
```
revans2: Again lets remove the try/catch/log/rethrow

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
```diff
@@ -0,0 +1,711 @@
...
+        }
+        @Override
+        public void close() {
+            mockedSupervisorLocalDir = null;
+        }
+    }
+    private static String mockedSupervisorLocalDir = null;
+    public static String supervisorLocalDir(Map conf) throws IOException {
+        if (mockedSupervisorLocalDir != null) {
+            return null;
```
revans2: I think this should be `return mockedSupervisorLocalDir`

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
```diff
@@ -0,0 +1,711 @@
...
+        }
+        @Override
+        public void close() {
+            mockedSupervisorStormDistRoot = null;
+        }
+    }
+    private static Map mockedSupervisorStormDistRoot = null;
+    public static String supervisorStormDistRoot(Map conf) throws IOException {
+        if (mockedSupervisorStormDistRoot != null) {
+            return null;
```
revans2: And here too `return  mockedSupervisorStormDistRoot`

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
```diff
@@ -0,0 +1,711 @@
...
+        }
+        return stormDistPath(supervisorLocalDir(conf)); // TODO: no need to forceMake here?, clj does not.
```
revans2: If clj does not, lets not do it either, just remove the TODO.

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
```diff
@@ -0,0 +1,711 @@
...
+    }
+
+    public static String supervisorStormDistRoot(Map conf, String stormId) throws IOException {
+        if (mockedSupervisorStormDistRoot != null) {
+            return null;
+        }
+        return supervisorStormDistRoot(conf) + FILE_SEPARATOR + stormId; // TODO: need to (url-encode storm-id)? Not.
```
revans2: Yes we need to do the url-encode

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
```diff
@@ -0,0 +1,711 @@
...
+        }
+        @Override
+        public void close() {
+            mockedSupervisorState = null;
+        }
+    }
+    private static Map mockedSupervisorState = null;
+    public static LocalState supervisorState(Map conf) throws IOException {
+        if (mockedSupervisorState != null) {
+            return null;
```
revans2: `return mockedSupervisorState`

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
```diff
@@ -0,0 +1,711 @@
...
+        }
+        @Override
+        public void close() {
+            mockedNimbusTopoHistoryState = null;
+        }
+    }
+    private static Map mockedNimbusTopoHistoryState = null;
+    public static LocalState nimbusTopoHistoryState(Map conf) throws IOException {
+        if (mockedNimbusTopoHistoryState != null) {
+            return null;
```
revans2: `return mockedNimbusTopoHistoryState;`

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
```diff
@@ -0,0 +1,711 @@
...
+    public static class SetMockedNimbusTopoHistoryState implements Closeable {
+        public SetMockedNimbusTopoHistoryState(Map conf) {
+            mockedNimbusTopoHistoryState = conf;
+        }
+        @Override
+        public void close() {
+            mockedNimbusTopoHistoryState = null;
+        }
+    }
+    private static Map mockedNimbusTopoHistoryState = null;
```
revans2: This is not a Map.

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
```diff
@@ -0,0 +1,711 @@
...
+    }
+
+    public static StormTopology readSupervisorTopology(Map conf, String stormId) throws IOException {
+        String stormRoot = supervisorStormDistRoot(conf, stormId);
+        String topologyPath = supervisorStormCodePath(stormRoot);
+        return Utils.deserialize(FileUtils.readFileToByteArray(new File(topologyPath)), StormTopology.class);
+    }
+
+    public static String workerUserRoot(Map conf) {
+        return (absoluteStormLocalDir(conf) + FILE_SEPARATOR + "/workers-users");
```
revans2: This matches the code, but I don't think we want the extra "/"  we should file a JIRA to check on that.

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
```diff
@@ -0,0 +1,711 @@
...
+    }
+
+    /* Never get used TODO : may delete it*/
+    public static String workerUserFile(Map conf, String workerId) {
```
revans2: yes delete this.

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
```diff
@@ -0,0 +1,711 @@
...
+            StringBuilder sb = new StringBuilder();
+            int r;
+            while ((r = br.read()) != -1) {
+                char ch = (char)r;
+                sb.append(ch);
+            }
+            String ret = sb.toString().trim();
+            return ret;
+        } catch (IOException e) {
+            LOG.error("Failed to get worker user for " + workerId + ".");
```
revans2: Lest use "{}" replacement here.  Very minor.

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
```diff
@@ -0,0 +1,711 @@
...
+    public static String getIdFromBlobKey(String key) {
+        if (key == null) return null;
+        String ret = null;
+        if (key.endsWith("-stormjar.jar")) {
+            ret = key.substring(0, key.length() - 13);
+        } else if (key.endsWith("-stormcode.ser")) {
+            ret = key.substring(0, key.length() - 14);
+        } else if (key.endsWith("-stormconf.ser")) {
+            ret = key.substring(0, key.length() - 14);
+        }
```
revans2: This works, it is not the RE used in the clojure, but it should be fine.  My only comment would be to have static final strings for "-stormjar.jar", etc. and use the .length() of the static final string instead of the hard coded length in the substring.

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
```diff
@@ -0,0 +1,711 @@
...
+        public SetMockedWorkerUserWSE(Map conf) {
+            mockedWorkerUserWSE = conf;
+        }
+
+        @Override
+        public void close() {
+            mockedWorkerUserWSE = null;
+        }
+    }
+    private static Map mockedWorkerUserWSE = null;
```
revans2: This should be a boolean.

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
```diff
@@ -0,0 +1,711 @@
...
+    //For testing only
+    // for java
+    // try (SetMockedWorkerUserWSE mocked = new SetMockedWorkerUserWSE(conf)) {
+    //    run test ...
+    // }
+    //
+    // for clojure
+    // (with-open [mock (SetMockedWorkerUserWSE. conf)]
+    //     run test ...)
+    public static class SetMockedWorkerUserWSE implements Closeable {
```
revans2: This should be renamed to something like IgnoreMockedWorkerUserWSE

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
```diff
@@ -0,0 +1,711 @@
...
+                return artifactsDir;
+            } else {
+                return (LOG_DIR + FILE_SEPARATOR + artifactsDir);
+            }
+        }
+    }
+
+    public static String workerArtifactsRoot(Map conf, String id) {
+        if (mockedWorkerArtifactsRoot != null) {
+            return mockedWorkerArtifactsRoot;
```
revans2: I'm not sure how the mocking worked in the clj code.  This looks OK, but is confusing because we are not including the ID.  It might be worth a comment here on why that is.

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
zhuoliu: Actually here we should return null. I know it is not straightforward: by default, mockedSupervisorState is null. Only when the test set this variable to be not null then we return "null".

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
```diff
@@ -0,0 +1,711 @@
...
+        }
+        return (workerArtifactsRoot(conf) + FILE_SEPARATOR + id);
+    }
+
+    public static String workerArtifactsRoot(Map conf, String id, Integer port) {
+        if (mockedWorkerArtifactsRoot != null) {
+            return mockedWorkerArtifactsRoot;
```
revans2: Same comment here about why we are not including the port/etc in this.

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
zhuoliu: Or we can do the opposite, by default this "mockedSupervisorState" is not null, only when test set this not be null, we return null in the static function

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
```diff
@@ -0,0 +1,711 @@
...
+        }
+        return (workerArtifactsRoot(conf, id) + FILE_SEPARATOR + port);
+    }
+
+    public static String workerArtifactsPidPath(Map conf, String id, Integer port) {
+        return (workerArtifactsRoot(conf, id, port) + FILE_SEPARATOR +  "worker.pid");
+    }
+
+    public static File getLogMetaDataFile(String fname) {
+        String[] subStrings = fname.split(FILE_SEPARATOR); // TODO: does this work well on windows?
```
revans2: This does not even work all that well off windows.  We should file a JIRA to look at using File for this instead.

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
zhuoliu: This is also why the type of "mockedSupervisorState" does not really matter.

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
```diff
@@ -0,0 +1,711 @@
...
+
+    public static Map overrideLoginConfigWithSystemProperty(Map conf) { // note that we delete the return value
+        String loginConfFile = System.getProperty("java.security.auth.login.config");
+        if (loginConfFile != null) {
+             conf.put("java.security.auth.login.config", loginConfFile);
+        }
+        return conf;
+    }
+
+    /* TODO: make sure test these two functions in manual tests */
```
revans2: What is the result of this TODO?

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
zhuoliu: Yes, this is a direct translation master-local-dir does have a forceMkdir with it. 

So where should we put those forceMkdir, or we require the calling code to check and create by themselves? I can file a JIRA for that if needed.

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
```diff
@@ -0,0 +1,711 @@
...
+        List<String> logsUsers = (List<String>)topologyConf.get(Config.LOGS_USERS);
+        List<String> topologyUsers = (List<String>)topologyConf.get(Config.TOPOLOGY_USERS);
+        Set<String> mergedUsers = new HashSet<String>();
+        if (logsUsers != null) {
+            for (String user : logsUsers) {
+                if (user != null) {
+                    mergedUsers.add(user);
+                }
+            }
+        }
```
revans2: This can be reduced to
```
if (logsUsers != null) {
  mergedUsers.addAll(logsUsers);
}
```

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
```diff
@@ -0,0 +1,711 @@
...
+        if (topologyUsers != null) {
+            for (String user : topologyUsers) {
+                if (user != null) {
+                    mergedUsers.add(user);
+                }
+            }
+        }
```
revans2: And this can similarly be
```
if (topologyUsers != null) {
    mergedUsers.addAll(topologyUsers);
}
```

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
```diff
@@ -0,0 +1,711 @@
...
+        if (topologyGroups != null) {
+            for (String group : topologyGroups) {
+                if (group != null) {
+                    mergedGroups.add(group);
+                }
+            }
+        }
```
revans2: See the above comment for users.  This should apply to groups as well.

revans2:
overall things look fairly good.

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
zhuoliu: A reason for not doing so is that some user might be null or "null" in the configuration?
I see there is "remove nil?" in original clojure code.
``
  (sort (distinct (remove nil?
                    (concat
                      (topology-conf LOGS-GROUPS)
                      (topology-conf TOPOLOGY-GROUPS))))))
``

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
zhuoliu: The second has a valueOf. Will unify as "null", which is clearer.

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
zhuoliu: removed.

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
zhuoliu: Actually we want to return null if the test is setting mockedSupervisorLocalDir as "sth".

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
zhuoliu: Then it will need to depend on util.clj. Or I will translate the url-encode within this pull request.

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
zhuoliu: Mark here: may need to file a JIRA.

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
zhuoliu: Fixed

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
revans2: Then lets take the model and expand things to be a more complete mock, and we can do this everywhere.  I really would like consistency in how this works for all of the static mocking we do, and the real issue is we need to support returning a null value.

```
    public static class SetMockedSupervisorLocalDir implements Closeable {
        public SetMockedSupervisorLocalDir(String dir) {
            mockedSupervisorLocalDir = dir;
            isSupervisorLocalDirMocked = true;
        }
        @Override
        public void close() {
            isSupervisorLocalDirMocked = false;
            mockedSupervisorLocalDir = null;
        }
    }
    private static String mockedSupervisorLocalDir = null;
    private static boolean isSupervisorLocalDirMocked = false;

    public static String supervisorLocalDir(Map conf) throws IOException {
        if (isSupervisorLocalDirMocked) {
            return mockedSupervisorLocalDir;
    ...
```

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
revans2: Good point I missed that, this is fine.

storm-core/src/clj/org/apache/storm/config.clj
```diff
@@ -41,291 +38,18 @@
...
-                      (topology-conf TOPOLOGY-USERS))))))
-
-(defn get-topo-logs-groups
-  [topology-conf]
-  (sort (distinct (remove nil?
-                    (concat
-                      (topology-conf LOGS-GROUPS)
-                      (topology-conf TOPOLOGY-GROUPS))))))
-
+  (even-sampler (sampling-rate conf)))
```
knusbaum: Please don't remove the newline from this file.

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
```diff
@@ -0,0 +1,678 @@
...
+    private final static Logger LOG = LoggerFactory.getLogger(ConfigUtils.class);
+    public final static String RESOURCES_SUBDIR = "resources";
+    public final static String NIMBUS_DO_NOT_REASSIGN = "NIMBUS-DO-NOT-REASSIGN";
+    public static final String FILE_SEPARATOR = File.separator;
+
+    public static String getLogDir() {
+        String dir;
+        Map conf;
+        if (System.getProperty("storm.log.dir") != null) {
+            dir = System.getProperty("storm.log.dir");
```
knusbaum: Can we avoid calling System.getProperty twice by assigning at declaration of dir?

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
```diff
@@ -0,0 +1,678 @@
...
+        } else if ((conf = readStormConfig()).get("storm.log.dir") != null) {
+            dir = String.valueOf(conf.get("storm.log.dir"));
+        } else {
+            if (System.getProperty("storm.home") != null) {
```
knusbaum: No need for nested if here. Make an else-if and an else.

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
```diff
@@ -0,0 +1,678 @@
...
+        public SetMockedStormConfig(Map conf) {
+            mockedStormConfig = conf;
+        }
+
+        @Override
+        public void close() {
+            mockedStormConfig = null;
+        }
+    }
+    private static Map mockedStormConfig = null;
```
knusbaum: Can we move this declaration just above SetMockedStormConfig and give it proper spacing?

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
```diff
@@ -0,0 +1,678 @@
...
+    //For testing only
+    // for java
+    // try (SetMockedStormConfig mocked = new SetMockedStormConfig(conf)) {
+    //    run test ...
+    // }
+    //
+    // for clojure
+    // (with-open [mock (SetMockedStormConfig. conf)]
+    //     run test ...)
+    public static class SetMockedStormConfig implements Closeable {
```
knusbaum: Should we maybe call this MockedStormConfigGuard or something with 'guard' in it? It would be nice to directly indicate that this object is an RAII-style guard.

d2r:
We have a separate implementation for mocking static methods that we think will also allow things like Mockito to track calls, arguments, and things of that sort. I will post a link when I have an example.

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
```diff
@@ -0,0 +1,678 @@
...
+    //For testing only
+    // for java
+    // try (SetMockedSupervisorLocalDir mocked = new SetMockedSupervisorLocalDir(dir)) {
+    //    run test ...
+    // }
+    //
+    // for clojure
+    // (with-open [mock (SetMockedSupervisorLocalDir. dir)]
+    //     run test ...)
+    public static class SetMockedSupervisorLocalDir implements Closeable {
```
knusbaum: Same comments as I made above apply here, and everywhere you use this trick.

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
```diff
@@ -0,0 +1,678 @@
...
+        if (mockedSupervisorLocalDir != null) {
+            return null;
+        }
+        String ret = absoluteStormLocalDir(conf) + FILE_SEPARATOR + "supervisor";
+        FileUtils.forceMkdir(new File(ret));
+        return ret;
+    }
+
+    public static String supervisorIsupervisorDir(Map conf) throws IOException {
+        return ((supervisorLocalDir(conf) + FILE_SEPARATOR + "isupervisor"));
```
knusbaum: Double parens?

d2r:
> I will post a link when I have an example.

new [staticmocking](https://github.com/d2r/storm/tree/56d48faac918123af46f1e021e38549fa6a4cf6c/storm-core/src/jvm/org/apache/storm/testing/staticmocking) package

See example methods under test: [Utils](https://github.com/d2r/storm/blob/56d48faac918123af46f1e021e38549fa6a4cf6c/external/storm-hbase/src/main/java/org/apache/storm/hbase/common/Utils.java)

Here is a (clojure) example of [usage](https://github.com/d2r/storm/blob/56d48faac918123af46f1e021e38549fa6a4cf6c/storm-core/test/clj/org/apache/storm/logviewer_test.clj#L372-L387). This will all eventually be Java, and it would use a try-with-resources syntax in place of the `with-open`.

If this seems like a good pattern to follow, we can add it to the [migration guidelines](https://cwiki.apache.org/confluence/display/STORM/Java+Migration+Guidelines).

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
```diff
@@ -0,0 +1,678 @@
...
+    public static String supervisorStormDistRoot(Map conf, String stormId) throws IOException {
+        if (mockedSupervisorStormDistRoot != null) {
+            return null;
+        }
+        return supervisorStormDistRoot(conf) + FILE_SEPARATOR + URLEncoder.encode(stormId, "UTF-8");
+    }
+
+    public static String supervisorStormJarPath(String stormRoot) {
+        String ret = "";
+        // we do this since to concat a null String will actually concat a "null", which is not the expected: ""
```
knusbaum: This pattern is repeated many, many times in this file. Can we move this to its own method and call that from here?
```
public static String supervisorStormBasePath(String stormRoot) {
        String ret = "";
        if (stormRoot != null) {
            ret = stormRoot;
        }
        return ret + FILE_SEPARATOR;
    }

public static String supervisorStormJarPath(String stormRoot) {
    return supervisorStormBasePath(stormRoot) + "stormjar.jar";
}
...
```

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
```diff
@@ -0,0 +1,678 @@
...
+                if (group != null) {
+                    mergedGroups.add(group);
+                }
+            }
+        }
+        List<String> ret = new ArrayList<String>(mergedGroups);
+        Collections.sort(ret);
+        return ret;
+    }
+}
```
knusbaum: Newline!

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
zhuoliu: Make sense, addressed.

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
zhuoliu: Addressed.

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
zhuoliu: Now go with Derek's proposal for static mocking.

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
zhuoliu: Fixed.

zhuoliu:
Thanks @revans2 @d2r @knusbaum for all your thorough comments, I addressed them. And I've taken Derek's proposed static mocking to replace the previous one.

storm-core/src/jvm/org/apache/storm/testing/staticmocking/MockedConfigUtils.java
```diff
@@ -0,0 +1,31 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.storm.testing.staticmocking;
+
+import org.apache.storm.utils.ConfigUtils;
+
+public class MockedConfigUtils extends ConfigUtils implements AutoCloseable {
```
revans2: Can we add some documentation about how to use this, like we had in the previous mocking code?  This is really minor though.

revans2:
Just one very very minor comment.  I am +1 on this either way.

storm-core/src/jvm/org/apache/storm/testing/staticmocking/MockedConfigUtils.java
zhuoliu: Derek's repository has a file [package-info.java](https://github.com/d2r/storm/blob/56d48faac918123af46f1e021e38549fa6a4cf6c/storm-core/src/jvm/org/apache/storm/testing/staticmocking/package-info.java), which shows how to use it. Do I need to include that within this pull request or wait util JIRA to do that? @d2r @revans2

zhuoliu: Yep, the above documentation file will go in with util translation (STORM-1226)[https://issues.apache.org/jira/browse/STORM-1226].

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
zhuoliu: Manually Tested, passed. I will remove this TODO.

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
zhuoliu: A jira filed for that in [STORM-1498](https://issues.apache.org/jira/browse/STORM-1498).

storm-core/src/jvm/org/apache/storm/testing/staticmocking/MockedConfigUtils.java
d2r: > Derek's repository has a file package-info.java, which shows how to use it. Do I need to include that within this pull request or wait util JIRA to do that? @d2r @revans2

For what it's worth, it is fine with me if you want to include that file in this pull request.

zhuoliu:
Hi @knusbaum , I addressed your comments. Could you check back on this one?

knusbaum:
+1

PR #1031, opened 2016-01-20
[storm-core] Added exception for emit to undeclared stream

mjsax:
`stream->component->grouper` does only contain streams that are not consumed. Thus, I add entries for all streams (with key is `streamID` and `null` as value) to the computed `HashMap` in `outbound-components`.


revans2:
+1 the change looks fine to me.

revans2:
@mjsax is there a JIRA associated with this change?

mjsax:
No JIRA. Should I open one?

revans2:
@mjsax if you want to, if you don't I am happy to merge it in just as a pull request.  I just wanted to check.

revans2:
Thanks @mjsax just merged this into master

mjsax:
Thanks :)

PR #1032, opened 2016-01-20
 STORM-1452: update REST API docs for profiling

ptgoetz:


ptgoetz:
As this is a documentation-only patch, and was previously reviewed as part of #1012, I'm going to merge this. If there are any objections, we can revert the change later.

PR #1038, opened 2016-01-24
[STORM-1483] add storm-mongodb connector

vesense:
Initial commit for code review.

### DIFF
--- a/external/storm-mongodb/src/main/java/org/apache/storm/mongodb/bolt/MongoUpdateBolt.java
+++ b/external/storm-mongodb/src/main/java/org/apache/storm/mongodb/bolt/MongoUpdateBolt.java
@@ -1,0 +1,75 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.storm.mongodb.bolt;
+
+import org.apache.commons.lang.Validate;
+import org.apache.storm.mongodb.common.QueryFilterCreator;
+import org.apache.storm.mongodb.common.mapper.MongoMapper;
+import org.apache.storm.topology.OutputFieldsDeclarer;
+import org.apache.storm.tuple.Tuple;
+import org.bson.Document;
+import org.bson.conversions.Bson;
+
+/**
+ * Basic bolt for updating from MongoDB.
+ *
+ * Note: Each MongoUpdateBolt defined in a topology is tied to a specific collection.
+ *
+ */
+public class MongoUpdateBolt extends AbstractMongoBolt {
+
+    private QueryFilterCreator queryCreator;
+    private MongoMapper mapper;
+
+    private boolean upsert;  //The default is false.
+
+    public MongoUpdateBolt(String url, String collectionName, QueryFilterCreator queryCreator, MongoMapper mapper) {
+        super(url, collectionName);
+
+        Validate.notNull(queryCreator, "QueryFilterCreator can not be null");
+        Validate.notNull(mapper, "MongoMapper can not be null");
+
+        this.queryCreator = queryCreator;
+        this.mapper = mapper;
+    }
+
+    @Override
+    public void execute(Tuple tuple) {
+        try{
+            //get document
+            Document doc = mapper.toDocument(tuple);
+            //get query filter
+            Bson filter = queryCreator.createFilter(tuple);
+            mongoClient.update(filter, doc, upsert);
+            this.collector.ack(tuple);
+        } catch (Exception e) {
+            this.collector.reportError(e);
+            this.collector.fail(tuple);
+        }
+    }
+
+    public void withUpsert(boolean upsert) {
+        this.upsert = upsert;
+    }
+
+    @Override
+    public void declareOutputFields(OutputFieldsDeclarer declarer) {
+        
+    }
+
+}
--- a/external/storm-mongodb/src/main/java/org/apache/storm/mongodb/common/MongoDBClient.java
+++ b/external/storm-mongodb/src/main/java/org/apache/storm/mongodb/common/MongoDBClient.java
@@ -1,0 +1,91 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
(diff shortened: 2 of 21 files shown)


external/storm-mongodb/src/main/java/org/apache/storm/mongodb/bolt/MongoUpdateBolt.java
```diff
@@ -0,0 +1,69 @@
...
+import org.bson.Document;
+import org.bson.conversions.Bson;
+
+/**
+ * Basic bolt for updating from MongoDB.
+ *
+ * Note: Each MongoUpdateBolt defined in a topology is tied to a specific collection.
+ *
+ */
+public class MongoUpdateBolt extends AbstractMongoBolt {
```
arunmahadevan: Why do we need two separate bolts - one for inserting and one for updating ? Can't we merge both of them into one using something like an `upsert` operation ? 'Insert if not exists' could be an option and if specified could insert the document if it does not exist.

external/storm-mongodb/src/main/java/org/apache/storm/mongodb/common/MongoDBClient.java
```diff
@@ -0,0 +1,66 @@
...
+        collection.insertOne(document);
+    }
+
+    public void insert(List<Document> documents) {
+        //This method is equivalent to a call to the bulkWrite method.
+        collection.insertMany(documents);
+    }
+
+    public void update(Bson filter, Bson update) {
+        //Update all documents in the collection
```
arunmahadevan: These could be proper javadocs

external/storm-mongodb/src/main/java/org/apache/storm/mongodb/trident/state/MongoState.java
```diff
@@ -0,0 +1,97 @@
...
+
+public class MongoState implements State {
+
+    private static final Logger LOG = LoggerFactory.getLogger(MongoState.class);
+
+    private Options options;
+    private MongoDBClient mongoClient;
+    private Map map;
+
+    protected MongoState(Map map, int partitionIndex, int numPartitions, Options options) {
```
arunmahadevan: partitionIndex and numPartitions are not used so could be removed.

external/storm-mongodb/src/main/java/org/apache/storm/mongodb/trident/state/MongoState.java
```diff
@@ -0,0 +1,97 @@
...
+        LOG.debug("commit is noop.");
+    }
+
+    public void updateState(List<TridentTuple> tuples, TridentCollector collector) {
+        List<Document> documents = Lists.newArrayList();
+        for (TridentTuple tuple : tuples) {
+            Document document = options.mapper.toDocument(tuple);
+            documents.add(document);
+        }
+        this.mongoClient.insert(documents);
```
arunmahadevan: I think we should use `update` (with upsert) here.

`insert` is always going to create new documents. For instance, if using `MongoState` to store word counts, it will create multiple documents for the same word. Unless thats what we want, the values should be updated.

arunmahadevan:
Github shows `1,058 additions, 1,057 deletions not shown` for pom.xml but does not show any changes. If there are no changes, you could just revert the pom.xml.

arunmahadevan:
May be not as a part of this PR, but would be good to have a `MapState` implementation for supporting trident's exactly once semantics. Take a look at `RedisMapState`, `HbaseMapState` etc.

vesense:
I fixed some issues. Work is going on for `upsert`.
And I will file a new JIRA for the next phase including `MongoMapState`.

external/storm-mongodb/src/main/java/org/apache/storm/mongodb/common/MongoDBClient.java
vesense: OK

external/storm-mongodb/src/main/java/org/apache/storm/mongodb/trident/state/MongoState.java
vesense: OK

external/storm-mongodb/src/main/java/org/apache/storm/mongodb/bolt/MongoUpdateBolt.java
vesense: Since `MongoUpdateBolt` uses `QueryFilterCreator` for creating a Mongo query Filter, maybe it's more clear to the user. @arunmahadevan 's advice is also make sence. I'd like to listen to other people's opinions.

external/storm-mongodb/src/main/java/org/apache/storm/mongodb/trident/state/MongoState.java
vesense: The `MongoStateUpdater` is something like `JdbcUpdater`, just for inserting data to a database. 
But what you state is also meaningful. I'll think about it

revans2: The issue here is with failures.  I am not an mongodb expert so I don't know what happens if you insert in a document that is already there. If you have a distributed state there will be multiple instances of this class spear throughout your topology.  If a single tuple fails as part of the updateState trident will replay the entire batch.  In that case we can try to insert documents that were already inserted.  I am not worried about consistency, because there is no query possible from trident.

revans2:
I am no expert on mongodb but overall it looks rather good.  @vesense if you do need a +1 from me to get this in please let me know, The code is simple enough I feel OK with doing it, but I would rather have someone else who knows the DB better then I do lend their support.

external/storm-mongodb/src/main/java/org/apache/storm/mongodb/bolt/MongoUpdateBolt.java
vesense: In MongoDB, `upsert` is an option of `UpdateRequest`:
>whether this update will insert a new document if no documents match the filter.  The default is false.
I don't think it's a better way to merge insert&update for inserting scenario.

external/storm-mongodb/src/main/java/org/apache/storm/mongodb/trident/state/MongoState.java
vesense: @revans2 Like RDBMS, if no unique index created or having a different primary key(`_id`), the existing documents can be inserted again.

vesense:
@ptgoetz @HeartSaVioR Could you have a chance to take a look?

external/storm-mongodb/src/main/java/org/apache/storm/mongodb/trident/state/MongoState.java
revans2: Then my only comment would be to add that to the documentation.  If there is no primary key provided trident state inserts in the case of failures may result in duplicate documents.

vesense: Updated. Thanks @revans2

external/storm-mongodb/src/main/java/org/apache/storm/mongodb/bolt/MongoUpdateBolt.java
vesense: I add `upsert` option to `MongoUpdateBolt`. Thanks @arunmahadevan

vesense:
@revans2 I think we can merge this to master and 1.x branch. What do you think?

revans2:
+1 the changes look good to me.

ptgoetz:
+1

I've not used MongoDB in ages, so I'm in the same boat as @revans2. Are there any other committers that are willing to sponsor this so it is maintained moving forward?

2new:
What's the status of this PR? which Storm version will include `storm-mongodb`?

vesense:
@2new Maybe Storm 1.0.0 or 2.0.0 will include this.(Storm PMC members will make the final decision).
 Actually, I hope the 1.0 release can include this.

vesense:
ping

vesense:
I filed some follow on JIRAs ([STORM-1573](https://issues.apache.org/jira/browse/STORM-1573) [STORM-1607](https://issues.apache.org/jira/browse/STORM-1607)) for the next phase.

harshach:
@vesense I can sponsor this connector. Please add me as sponsor in README.

harshach:
+1. Can you please squash additional commits.

vesense:
@harshach OK. I'll update it right away.

vesense:
@harshach Done.

vesense:
ping @harshach

harshach:
@vesense thanks for the patch. Merged into master. Can you open up the same PR against 1.x-branch as well.

vesense:
@harshach Done. -> https://github.com/apache/storm/pull/1219

PR #1044, opened 2016-01-25
STORM-1464: Support multiple file outputs

dossett:
This PR will enable storm-hdfs to write to multiple files in a couple of different circumstances:

- If a bolt requires multiple output files.  For example, the avro bolt needs to write different Avro records with different schemas to different files.  Schema evolution is a common use of Avro so I expect this to be common use.

- Partitioning output of the HDFS bolt.  Based on mailing discussion, there is demand for this feature.

It does introduce a couple of incompatible changes.  The most obvious is adding a method to the FileRotationPolicy interface.  I tried to minimize other API changes at the expense of some less than elegant code in a couple of places.

### DIFF
--- a/external/storm-hdfs/README.md
+++ b/external/storm-hdfs/README.md
@@ -184,6 +184,7 @@ Similar to sync policies, file rotation policies allow you to control when data 
 public interface FileRotationPolicy extends Serializable {
     boolean mark(Tuple tuple, long offset);
     void reset();
+    FileRotationPolicy copy();
 }
 ``` 
 
@@ -240,6 +241,23 @@ If you are using Trident and sequence files you can do something like this:
                 .addRotationAction(new MoveFileAction().withDestination("/dest2/"));
 ```
 
+### Data Partitioning
+Data can be partitioned to different HDFS directories based on characteristics of the tuple being processed or purely
+external factors, such as system time.  To partition your your data, write a class that implements the ```Partitioner```
+interface and pass it to the withPartitioner() method of your bolt. The getPartitionPath() method returns a partition
+path for a given tuple.
+
+Here's an example of a Partitioner that operates on a specific field of data:
+
+```java
+
+    Partitioner partitoner = new Partitioner() {
+            @Override
+            public String getPartitionPath(Tuple tuple) {
+                return Path.SEPARATOR + tuple.getStringByField("city");
+            }
+     };
+```
 
 ## HDFS Bolt Support for HDFS Sequence Files
 
@@ -303,16 +321,15 @@ The `org.apache.storm.hdfs.bolt.AvroGenericRecordBolt` class allows you to write
         AvroGenericRecordBolt bolt = new AvroGenericRecordBolt()
                 .withFsUrl("hdfs://localhost:54310")
                 .withFileNameFormat(fileNameFormat)
-                .withSchemaAsString(schema)
                 .withRotationPolicy(rotationPolicy)
                 .withSyncPolicy(syncPolicy);
 ```
-The setup is very similar to the `SequenceFileBolt` example above.  The key difference is that instead of specifying a
-`SequenceFormat` you must provide a string representation of an Avro schema through the `withSchemaAsString()` method.
-An `org.apache.avro.Schema` object cannot be directly provided since it does not implement `Serializable`.
 
-The AvroGenericRecordBolt expects to receive tuples containing an Avro GenericRecord that conforms to the provided
-schema.
+The avro bolt will write records to separate files based on the schema of the record being processed.  In other words,
+if the bolt receives records with two different schemas, it will write to two separate files.  Each file will be rotatated
+in accordance with the specified rotation policy. If a large number of Avro schemas are expected, then the bolt should
+be configured with a maximum number of open files at least equal to the number of schemas expected to prevent excessive
+file open/close/create operations.
 
 To use this bolt you **must** register the appropriate Kryo serializers with your topology configuration.  A convenience
 method is provided for this:
--- a/external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/AbstractHdfsBolt.java
+++ b/external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/AbstractHdfsBolt.java
@@ -17,14 +17,12 @@
  */
 package org.apache.storm.hdfs.bolt;
 
-import org.apache.storm.Config;
 import org.apache.storm.task.OutputCollector;
 import org.apache.storm.task.TopologyContext;
 import org.apache.storm.topology.OutputFieldsDeclarer;
 import org.apache.storm.topology.base.BaseRichBolt;
 import org.apache.storm.tuple.Tuple;
 import org.apache.storm.utils.TupleUtils;
-import org.apache.storm.utils.Utils;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
@@ -32,6 +30,9 @@ import org.apache.storm.hdfs.bolt.format.FileNameFormat;
 import org.apache.storm.hdfs.bolt.rotation.FileRotationPolicy;
 import org.apache.storm.hdfs.bolt.rotation.TimedRotationPolicy;
 import org.apache.storm.hdfs.bolt.sync.SyncPolicy;
+import org.apache.storm.hdfs.common.AbstractHDFSWriter;
+import org.apache.storm.hdfs.common.NullPartitioner;
+import org.apache.storm.hdfs.common.Partitioner;
 import org.apache.storm.hdfs.common.rotation.RotationAction;
 import org.apache.storm.hdfs.common.security.HdfsSecurityUtil;
 import org.slf4j.Logger;
@@ -39,6 +40,8 @@ import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.LinkedList;
 import java.util.Map;
 import java.util.List;
@@ -52,15 +55,16 @@ public abstract class AbstractHdfsBolt extends BaseRichBolt {
      * Half of the default Config.TOPOLOGY_MESSAGE_TIMEOUT_SECS
      */
     private static final int DEFAULT_TICK_TUPLE_INTERVAL_SECS = 15;
+    private static final Integer DEFAULT_MAX_OPEN_FILES = 50;
 
-    protected ArrayList<RotationAction> rotationActions = new ArrayList<RotationAction>();
-    private Path currentFile;
+    protected Map<String, AbstractHDFSWriter> writers;
+    protected Map<String, Integer> rotationCounterMap = new HashMap<>();
+    protected List<RotationAction> rotationActions = new ArrayList<>();
     protected OutputCollector collector;
     protected transient FileSystem fs;
     protected SyncPolicy syncPolicy;
     protected FileRotationPolicy rotationPolicy;
     protected FileNameFormat fileNameFormat;
-    protected int rotation = 0;
     protected String fsUrl;
     protected String configKey;
     protected transient Object writeLock;
@@ -69,22 +73,21 @@ public abstract class AbstractHdfsBolt extends BaseRichBolt {
     protected long offset = 0;
     protected Integer fileRetryCount = DEFAULT_RETRY_COUNT;
     protected Integer tickTupleInterval = DEFAULT_TICK_TUPLE_INTERVAL_SECS;
+    protected Integer maxOpenFiles = DEFAULT_MAX_OPEN_FILES;
+    protected Partitioner partitioner = new NullPartitioner();
 
     protected transient Configuration hdfsConfig;
 
-    protected void rotateOutputFile() throws IOException {
+    protected void rotateOutputFile(AbstractHDFSWriter writer) throws IOException {
         LOG.info("Rotating output file...");
         long start = System.currentTimeMillis();
         synchronized (this.writeLock) {
-            closeOutputFile();
-            this.rotation++;
+            writer.close();
 
-            Path newFile = createOutputFile();
             LOG.info("Performing {} file rotation actions.", this.rotationActions.size());
             for (RotationAction action : this.rotationActions) {
-                action.execute(this.fs, this.currentFile);
+                action.execute(this.fs, writer.getFilePath());
             }
-            this.currentFile = newFile;
         }
         long time = System.currentTimeMillis() - start;
         LOG.info("File rotation took {} ms.", time);
@@ -104,6 +107,8 @@ public abstract class AbstractHdfsBolt extends BaseRichBolt {
             throw new IllegalStateException("File system URL must be specified.");
         }
 
+        writers = new WritersMap(this.maxOpenFiles);
+
         this.collector = collector;
         this.fileNameFormat.prepare(conf, topologyContext);
         this.hdfsConfig = new Configuration();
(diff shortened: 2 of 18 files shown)


dossett:
Title updated to reflect correct JIRA.

dossett:
I up-merged this PR.  Does anyone have feedback on this change?  Are API breaking changes off the table for 2.0?  If so, i can close this PR and come back to it at a later time.

harshach:
@dossett This looks good to me. Can you upmerge the patch. @arunmahadevan can you please review this as well.

external/storm-hdfs/README.md
```diff
@@ -240,6 +240,23 @@ If you are using Trident and sequence files you can do something like this:
...
+path for a given tuple.
+
+Here's an example of a Partitioner that operates on a specific field of data:
+
+```java
+
+    Partitioner partitoner = new Partitioner() {
+            @Override
+            public String getPartitionPath(Tuple tuple) {
+                return Path.SEPARATOR + "city=" + tuple.getStringByField("city");
```
arunmahadevan: Why does the example path contain "city=" ? 
If we are expecting the Partitioner to return a path its better to return a Path (java.nio.file.Path) so that it  does not return any arbitrary string.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/AbstractHdfsBolt.java
```diff
@@ -52,9 +54,10 @@
      * Half of the default Config.TOPOLOGY_MESSAGE_TIMEOUT_SECS
      */
     private static final int DEFAULT_TICK_TUPLE_INTERVAL_SECS = 15;
+    private static final Integer DEFAULT_MAX_OPEN_FILES = 5;
```
arunmahadevan: Isn't the default too low?

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/AbstractHdfsBolt.java
```diff
@@ -145,13 +134,20 @@ public final void execute(Tuple tuple) {
...
+            AbstractHDFSWriter writer = null;
+            String writerKey = null;
+
             if (TupleUtils.isTick(tuple)) {
                 LOG.debug("TICK! forcing a file system flush");
                 this.collector.ack(tuple);
                 forceSync = true;
             } else {
+
+                writerKey = getHashKeyForTuple(tuple);
```
arunmahadevan: Can't the partition path be used as the key?

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/AbstractHdfsBolt.java
```diff
@@ -198,22 +194,62 @@ public final void execute(Tuple tuple) {
...
 
+    private AbstractHDFSWriter getOrCreateWriter(String writerKey, Tuple tuple) throws IOException {
+        AbstractHDFSWriter writer;
+
+        writer = writers.get(writerKey);
+        if (writer == null) {
+            if (writers.size() >= maxOpenFiles)
+            {
+                String keyToOldest = getKeyToOldestWriter();
+                AbstractHDFSWriter oldest = writers.get(keyToOldest);
```
arunmahadevan: Why not have getKeyToOldestWriter return the oldest entry directly and remove the extra call to writers.get ?

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/AbstractHdfsBolt.java
```diff
@@ -198,22 +194,62 @@ public final void execute(Tuple tuple) {
...
+                rotateOutputFile(oldest);
+                writers.remove(keyToOldest);
```
arunmahadevan: Is the oldest entry removed to maintain the max size of the writers hashtable (max open files limit) ? nit: most of the above can be taken care of if the writers map was a LinkedHashMap extension with a LRU logic.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/AbstractHdfsBolt.java
```diff
@@ -198,22 +194,62 @@ public final void execute(Tuple tuple) {
...
+     * A tuple must be mapped to a writer based on two factors:
+     *  - bolt specific logic that must separate tuples into different files in the same directory (see the avro bolt
+     *    for an example of this)
+     *  - the directory the tuple will be partioned into
+     *
+     * @param tuple
+     * @return
+     */
+    private String getHashKeyForTuple(Tuple tuple) {
+        final String boltKey = getWriterKey(tuple);
```
arunmahadevan: Why a separate key instead of the partition path itself as the key ?

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/AbstractHdfsBolt.java
```diff
@@ -223,29 +259,57 @@ public final void execute(Tuple tuple) {
...
+                writers.clear();
+            }
+        };
+        this.rotationTimer.scheduleAtFixedRate(task, interval, interval);
+    }
+
+    protected Path getBasePathForNextFile(Tuple tuple) {
+
+        Path fullPathToFile = new Path(this.fsUrl + this.fileNameFormat.getPath() + this.partitioner.getPartitionPath(tuple),
+                this.fileNameFormat.getName(this.rotation, System.currentTimeMillis()));
```
arunmahadevan: Since `this.rotation` is not per writer, we could end up having non-contiguous files names (like file-1, file-5, file-8 etc) inside a partition path, correct ? If so it should be fixed.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/AbstractHdfsBolt.java
```diff
@@ -223,29 +259,57 @@ public final void execute(Tuple tuple) {
...
+
+        return fullPathToFile;
+    }
 
     abstract void doPrepare(Map conf, TopologyContext topologyContext, OutputCollector collector) throws IOException;
 
+    abstract String getWriterKey(Tuple tuple);
```
arunmahadevan: this method is not required if partition path is used as the key.

arunmahadevan:
@dossett @harshach the changes looks good overall, but I have a few questions/comments and once those are addressed I will review it again.

dossett:
Thanks @arunmahadevan and @harshach.  I'll rebase and review feedback this week.

mazhechao:
Great! I'm looking forward for this feature.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/AbstractHdfsBolt.java
dossett: Fair point.  What seems reasonable.  20? 50?

external/storm-hdfs/README.md
dossett: The "city=" was from a Hive-specific use case I was working on.  The partitioner would return "city=<name>" to create a partitioning scheme that would be usable for a Hive external table.  I will remove that from the standard documentation and maybe add a hive-specific note.

dossett: I thought about having Partitioner returning an actual path but decided against it for two reasons:
- I liked the idea of the "partition" being solely a function of the tuple without reference to anything else
- Since end users implement a Partitioner having it return a complete path would give the user access to details otherwise hidden from their code.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/AbstractHdfsBolt.java
dossett: Great suggestion, thank you.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/AbstractHdfsBolt.java
dossett: Very good observation, thank you.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/AbstractHdfsBolt.java
dossett: I don't believe it can go away entirely.  One of the things it does is track the need for simultaneous multiple open files within the same partition directory.  For example, the Avro bolt must write objects with different schemas to different files regardless of partitioning.  

I think it's best to keep that separate notion and have a key definition that isn't just the partition path.  I will think about that some more though.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/AbstractHdfsBolt.java
rajkonda: How about making spout parallelism (number of spout threads) as default ?

dossett: @rajkonda I like that idea, but there's any sort of fan out could happen between a spout and the HDFS Bolt.  I'll just start with a much larger number like 20.

dossett:
@arunmahadevan I have addressed several of your comments -- thanks for the review, it was helpful. Per my other comments, the key can't be based only on the partition information.

arunmahadevan:
@dossett I went through the patch again and I have one question regarding the writer key. I see that you maintain a separate writer per writer key. In the docs you mention "The avro bolt will write records to separate files based on the schema of the record being processed.  In other words, if the bolt receives records with two different schemas, it will write to two separate files."

Is each writer expected to write to a separate file ? If so I dont see that happening because in `getBasePathForNextFile` the file name is based on the partition path and rotation id alone and then this path is passed to `makeNewWriter`. So there could be multiple writers writing to the same file and infact at overlapping offsets based on that writer's offset and may be corrupt the file. Can you help me understand if each writer can write to the same file or is it always supposed to write to different files?

dossett:
@arunmahadevan Writers should never be writing to the same file.  `getBasePathForNextFile` will increment the rotation value before opening the file.  File name also includes `System.currentTimeMillis()` which should further prevent conflict.

There's also a unit test that attempts to validate this by writing two different schemas and confirming that two files are created: https://github.com/dossett/storm/blob/STORM-1494/external/storm-hdfs/src/test/java/org/apache/storm/hdfs/bolt/AvroGenericRecordBoltTest.java#L156-L168

arunmahadevan:
@dossett thanks for the clarification. So the tuples with different schema are written to the next rotated file. I was assuming some association between the writer key / schema with the file name. May be that could be a follow on enhancement.

+1 for merging this in.

dossett:
Thanks @arunmahadevan and @harshach. I made one final documentation tweak and squashed the commits.

dossett:
@arunmahadevan @harshach Are you both still +1.  I can merge, but want to double check first.  Thanks.

ptgoetz:
+1. We should also merge this into 1.x-branch.

dossett:
Just to reiterate, I tried to minimize public API changes, but a couple were unavoidable.  e.g., FileRotation interface has a new method that all any implementation would have to support.  Just want to make sure those are ok for 1.x (and 2.0 for that matter)

ptgoetz:
@dossett Good point. We may want to hold off on 1.x since that change is backward incompatible.

dossett:
@ptgoetz Thanks for confirming... even though I would have loved to get this is 1.0

dossett:
Merged.  Thank you to everyone who provided feedback!

mazhechao:
1.x does not have the interface Partitioner yet, and 2.x has not released, so the document shouldn't add Data Partitioning section.

dossett:
@mazhechao This change was to the master branch and did not update 1.x documentation.  Were you referring to something else?

mazhechao:
@dossett Oh yes, I read that from the master branch.

matthewdfleming:
Is there any plan to port this to 1.0.x? I saw that there was an api change which I guess makes it more of a 1.1.x kind of change? I guess I'm asking for the best way to add this feature to a project that's being developed right now?

ptgoetz:
@matthewdfleming This will be available in the 1.1 release, which we are getting closer to. You should be able to use it with 1.0.x, but for now you would have to compile it yourself.

dossett:
@ptgoetz This PR was never, to my knowledge, applied to 1.x.  It contained a very small incompatible API change to FileRotation, so users with their own FileRotation implementation would get broken.  I'd certainly be open to applying it to 1.x, but that wouldn't be consistent with semantic versioning, FWIW.  (cc @matthewdfleming)

matthewdfleming:
@ptgoetz it's an interesting point.. "MINOR version when you add functionality in a backwards-compatible manner" (http://semver.org). I wonder if adding a method to a public interface could ever be a minor version then because someone out there could have written their own implementation? This feels pretty limiting, but maybe that's the point.

ryanpersaud:
When the eldest entry in the WritersMap gets removed due to the number of writers exceeding maxWriters, how does the writer get closed and any rotation action get executed?  Thanks for the information.

dossett:
Hi @ryanpersaud -- that is a very good observation, it would seem that doesn't happen.  I admit to being biased against rotation actions and would never use them in production, which is probably why I never noticed it.  But that's not a good enough reason for a bug like that!

Since changing jobs last year I don't have the opportunity to actively work on Storm anymore, but if you have thoughts on a fix I would certainly give a code review my full attention. A failing unit test to confirm the behavior (using a Mockito spy to confirm that close isn't called) would be a very helpful start.

Thank you for pointing this out!

ryanpersaud:
Now you've piqued my curiosity - what is it that you don't like about rotation actions?

Writers are only stored in the writers map, so once they are removed, there is no way for FSDataOutputStream.close() to be invoked by Storm code.  Furthermore, there are no finalize() methods defined in HDFSWriter, AbstractHDFSWiter, FSDataOutputStream, or DataOutputStream (relying on the finalizer to close the writer would be a bad practice anyway).

I used code in the HiveBolt as a starting point for handling retiring writers (see below).  Please let me know what you think.  I'll also create a Jira tonight and a PR once we coalesce on a solution.

     // Setting accessOrder to true to facilitate obtaining the LRU  entry
     writers = new LinkedHashMap<>(this.maxOpenFiles, 0.75f, true);

    private AbstractHDFSWriter getOrCreateWriter(String writerKey, Tuple tuple) throws IOException {
        AbstractHDFSWriter writer = writers.get( writerKey );
        if (writer == null) {
            LOG.debug("Creating Writer for writerKey : " + writerKey);
            Path pathForNextFile = getBasePathForNextFile(tuple);
            writer = makeNewWriter(pathForNextFile, tuple);

            if (writers.size() > (this.maxOpenFiles - 1)) {
                LOG.info("cached writers size {} exceeded maxOpenFiles {} ", writers.size(), this.maxOpenFiles);
                retireEldestWriter();
            }
            this.writers.put(writerKey, writer);
        }
        return writer;
    }

    /**
     * Locate writer that has not been used for longest time and retire it
     */
    private void retireEldestWriter() throws IOException {
        LOG.info("Attempting to close eldest writer");
        Iterator<String> iterator = this.writers.keySet().iterator();
        if (iterator.hasNext()) {
            String eldestKey = iterator.next();
            doRotationAndRemoveWriter(eldestKey, this.writers.get(eldestKey));
        }
    }

dossett:
@ryanpersaud Your basic approach looks right to me, thanks for taking the lead on this.

Regarding rotation actions, my concern is that there's no absolute guarantee they will get executed. If a worker crashes and restarts then the files it had open before will never have rotation actions performed against them.

fescandell:
Hello,

Is there a similar solution like "withPartitioner(new Partitioner()" by using Trident : 

`HdfsState.HdfsFileOptions()`

dossett:
@fescandell I don't believe there is. Because I was unfamiliar with the Trident API, I didn't make any similar changes to it.

PR #1046, opened 2016-01-26
[STORM-1501] launch worker process exception will cause supervisor process exited

caofangkun:
[util.clj/async-loop](https://github.com/apache/storm/blob/master/storm-core/src/clj/org/apache/storm/util.clj#L474) default kill-fn will kill current process  

when supervisor use [util.clj/launch-process](https://github.com/apache/storm/blob/master/storm-core/src/clj/org/apache/storm/util.clj#L546) to launch worker process , if exeception occurs , supervisor process will exit.

### DIFF
--- a/storm-core/src/clj/org/apache/storm/util.clj
+++ b/storm-core/src/clj/org/apache/storm/util.clj
@@ -554,7 +554,7 @@
                  (log-message log-prefix " interrupted.")))
              (exit-code-callback (.exitValue process)))
            nil)
-          :kill-fn (fn [error] (log-error (RuntimeException. ("launch process failed...")))))
+          :kill-fn (fn [error] (log-error (RuntimeException. ("launch process failed..."))))))
       process)))
    
 (defn exists-file?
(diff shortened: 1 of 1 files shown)


revans2:
I am a bit curious why we need this?  What exception/error did you see come from the worker process?

storm-core/src/clj/org/apache/storm/util.clj
```diff
@@ -553,7 +553,8 @@
                (catch InterruptedException e
                  (log-message log-prefix " interrupted.")))
              (exit-code-callback (.exitValue process)))
-           nil)))                    
+           nil)
+          :kill-fn (fn [error] (log-error (RuntimeException. ("launch process failed..."))))))
```
revans2: Why are we throwing away error?

What happens if this does not exit?  I assume that the heartbeat monitoring will detect it and do the right thing, but are there other places in the code besides a worker process that we use this?  Will there be problems especially around the exit-code-callback if they are not called?

PR #1043, opened 2016-01-25
Storm 1226

rfarivar:
This is translating util.clj into java functions mostly in utils/Utils.java. 

Translation is half way done, passing all unit tests. Putting the pull request up to get reviews.

PR #1047, opened 2016-01-26
[STORM-1257] port backtype.storm.zookeeper to java

hustfxj:
1.  Port "org.apache.storm.zookeeper" to java at "org.apache.storm.zookeeper.Zookeeper".
2.  Update all the callings to the config functions.
3.  Deleted most functions in zookeeper.clj except a few: mk-client (because of "watcher" which is derived by clojure in some .clj files).
4.  add the package "Callback", where callback functions put. I am merging clusters and timer which involves lots of callback functions. I put them into package "Callback".


hustfxj:
test

storm-core/src/jvm/org/apache/storm/zookeeper/LeaderElectorImp.java
```diff
@@ -0,0 +1,135 @@
...
+    private AtomicReference<LeaderLatchListener> leaderLatchListener;
+
+    public LeaderElectorImp(Map conf, List<String> servers, CuratorFramework zk, String leaderlockPath, String id, AtomicReference<LeaderLatch> leaderLatch,
+                            AtomicReference<LeaderLatchListener> leaderLatchListener) {
+        this.conf = conf;
+        this.servers = servers;
+        this.zk = zk;
+        this.leaderLatch = leaderLatch;
+        this.id = id;
+        this.leaderLatch = leaderLatch;
```
wuchong: duplicately here

storm-core/src/jvm/org/apache/storm/zookeeper/Zookeeper.java
```diff
@@ -0,0 +1,355 @@
...
+                LOG.info(hostName + " gained leadership");
+            }
+
+            @Override
+            public void notLeader() {
+                LOG.info(hostName + " lost leadership.");
+            }
+        };
+    }
+
```
wuchong: missing comment here? 

  "Zookeeper Implementation of ILeaderElector."

revans2:
Looks like you missed a reference to in-process-zookeeper (from the CI)
```
Exception in thread "main" java.lang.RuntimeException: No such var: org.apache.storm.zookeeper/mk-inprocess-zookeeper, compiling:(org/apache/storm/transactional_test.clj:105:5)
```

storm-core/src/jvm/org/apache/storm/callback/DefaultWatcherCallBack.java
```diff
@@ -0,0 +1,35 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class DefaultWatcherCallBack implements WatcherCallBack {
+
+    private static Logger LOG = LoggerFactory.getLogger(DefaultWatcherCallBack.class);
+
+    @Override
+    public void execute(Watcher.Event.KeeperState state, Watcher.Event.EventType type, String path) {
+        LOG.info("Zookeeper state update:" + ZkKeeperStates.getStateName(state) + "," + ZkEventTypes.getStateName(type) + "," + path);
```
revans2: This is new code that didn't exist in storm proper.  I think it would be cleaner to have this be a debug message instead of an info message.  Also could you use "{}" substitution instead of +?

```
        LOG.info("Zookeeper state update: {}, {}, {}", ZkKeeperStates.getStateName(state),  ZkEventTypes.getStateName(type), path);
```

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1382,5 +1382,43 @@ public static TopologyInfo getTopologyInfo(String name, String asUser, Map storm
...
+            }
+
+        }
+        return buff.toString();
+    }
+
+    public static String normalizePath(String path) {
+        String rtn = toksToPath(tokenizePath(path));
+        return rtn;
+    }
```
revans2: Can we please move tokenizePath, toksToPath, and normalizePath to Zookeeper.java?  and change them from using File.separator, to use "/".  The issue is that they are specific to zookeeper so need the "/", without it on windows it will not work.  They should never have been in util.clj to begin with.  I'll update the util.clj JIRA to indicate this too.

storm-core/src/jvm/org/apache/storm/zookeeper/LeaderElectorImp.java
```diff
@@ -0,0 +1,135 @@
...
+        this.leaderLatchListener = leaderLatchListener;
+    }
+
+    @Override
+    public void prepare(Map conf) {
+        LOG.info("no-op for zookeeper implementation");
```
revans2: Could this be a comment instead of a log message?

storm-core/src/jvm/org/apache/storm/zookeeper/ZkEventTypes.java
```diff
@@ -0,0 +1,43 @@
...
+
+        map.put(Watcher.Event.EventType.None, ":none");
+        map.put(Watcher.Event.EventType.NodeCreated, ":node-created");
+        map.put(Watcher.Event.EventType.NodeDeleted, ":node-deleted");
+        map.put(Watcher.Event.EventType.NodeDataChanged, ":node-data-changed");
+        map.put(Watcher.Event.EventType.NodeChildrenChanged, ":node-children-changed");
+
+    }
+
+    public static String getStateName(Watcher.Event.EventType type) {
```
revans2: I would rather see this called getTypeName instead of getStateName.  Of even just getName.

storm-core/src/jvm/org/apache/storm/callback/DefaultWatcherCallBack.java
revans2: Never mind about making it debug.  I missed original implementation.  I still would like to use the "{}" substitution though.

storm-core/src/jvm/org/apache/storm/zookeeper/Zookeeper.java
```diff
@@ -0,0 +1,355 @@
...
+        fk.start();
+        return fk;
+    }
+
+    /**
+     * connect ZK, register Watch/unhandle Watch
+     *
+     * @return
+     */
+    public static CuratorFramework mkClient(Map conf, List<String> servers, Object port, String root, final WatcherCallBack watcher) {
```
revans2: Could you please file a follow up JIRA to make all of the 'Object port' parameters 'Integer port' instead?  This should be of the java Utils.newCurator too.

storm-core/src/jvm/org/apache/storm/zookeeper/Zookeeper.java
```diff
@@ -0,0 +1,355 @@
...
+    }
+
+    public static void deleteNode(CuratorFramework zk, String path) throws RuntimeException {
+        try {
+            String npath = Utils.normalizePath(path);
+            if (existsNode(zk, npath, false)) {
+                zk.delete().deletingChildrenIfNeeded().forPath(Utils.normalizePath(path));
+            }
+
+        } catch (KeeperException.NoNodeException e) {
```
revans2: In the original code we did a try-cause.  I am not sure if that is needed here or not, I assume not, but I wanted to point it out if we run into issues while testing.

storm-core/src/jvm/org/apache/storm/zookeeper/Zookeeper.java
```diff
@@ -0,0 +1,355 @@
...
+        String npath = Utils.normalizePath(path);
+        if (npath.equals("/")) {
+            return;
+        }
+        if (existsNode(zk, npath, false)) {
+            return;
+        }
+        byte[] byteArray = new byte[1];
+        byteArray[0] = (byte) 7;
+        createNode(zk, npath, byteArray, org.apache.zookeeper.CreateMode.PERSISTENT, acls);
```
revans2: In this place we do need to do something like the try-cause that was in the original code.  There is a race condition where multiple clients can be doing a mkdirs.  If that happens we need to ignore the NodeExistsException from createNode, but createNode will have wrapped it in a RuntimeException.  If you want to avoid the wrapping/unwrapping and just catch the exception yourself without unwrapping it you could do the following instead.

```
try {
    zk.create().withMode(CreateMode.PERSISTENT).withACL(acls).forPath(npath, byteArray);
} catch (KeeperException.NodeExistsException ignored) {
    // this can happen when multiple clients doing mkdir at same time
}
```

storm-core/src/jvm/org/apache/storm/zookeeper/Zookeeper.java
```diff
@@ -0,0 +1,355 @@
...
+     *
+     * @return
+     */
+    public static CuratorFramework mkClient(Map conf, List<String> servers, Object port, String root, final WatcherCallBack watcher) {
+
+        return mkClient(conf, servers, port, root, watcher, null);
+    }
+
+    public static String createNode(CuratorFramework zk, String path, byte[] data, org.apache.zookeeper.CreateMode mode, List<ACL> acls)
+            throws RuntimeException {
```
revans2: In general we probably should not be declaring that we throw a RuntimeException.  This will force other java code to explicitly handle it.  We were wrapping all of the exceptions in Runtime because they were bubbling up through clojure to java code and were violating the java APIs, because apparently clojure can through an exception that is not declared in the API.  We probably want to file a follow on JIRA to clean up the wrapping so the Zookeeper.java APIs don't wrap exceptions, but throw what they need to, and then only when we have to we wrap the exceptions.

storm-core/src/jvm/org/apache/storm/zookeeper/Zookeeper.java
```diff
@@ -0,0 +1,355 @@
...
+        try {
+            String npath = Utils.normalizePath(path);
+            if (existsNode(zk, npath, watch)) {
+                if (watch) {
+                    return zk.getData().watched().forPath(npath);
+                } else {
+                    return zk.getData().forPath(npath);
+                }
+            }
+        } catch (KeeperException e) {
```
revans2: Can we explicitly catch the NoNodeException here instead?  There are lots of other KeeperExceptions that we want to bubble up.

storm-core/src/jvm/org/apache/storm/zookeeper/Zookeeper.java
```diff
@@ -0,0 +1,355 @@
...
+        try {
+            byte[] bytes = null;
+            Stat stats = new Stat();
+            String npath = Utils.normalizePath(path);
+            if (existsNode(zk, npath, watch)) {
+                if (watch) {
+                    bytes = zk.getData().storingStatIn(stats).watched().forPath(npath);
+                } else {
+                    bytes = zk.getData().storingStatIn(stats).forPath(npath);
+
```
revans2: Extra line (Very very minor)

storm-core/src/clj/org/apache/storm/cluster_state/zookeeper_state_factory.clj
```diff
@@ -143,15 +144,15 @@
...
 
      (sync-path
        [this path]
-       (zk/sync-path zk-writer path))
+       (Zookeeper/syncPath zk-writer path))
 
       (delete-node-blobstore
         [this path nimbus-host-port-info]
-        (zk/delete-node-blobstore zk-writer path nimbus-host-port-info))
+        (Zookeeper/deleteDodeBlobstore zk-writer path nimbus-host-port-info))
```
redsanket: must be deleteNodeBlobstore instead of deleteDodeBlobstore "spelling"

storm-core/src/jvm/org/apache/storm/zookeeper/Zookeeper.java
```diff
@@ -0,0 +1,355 @@
...
+                return zk.getChildren().forPath(npath);
+            }
+        } catch (Exception e) {
+            throw Utils.wrapInRuntime(e);
+        }
+    }
+
+    // Deletes the state inside the zookeeper for a key, for which the
+    // contents of the key starts with nimbus host port information
+    public static void deleteDodeBlobstore(CuratorFramework zk, String parentPath, String hostPortInfo) throws RuntimeException {
```
redsanket: Same here deleteNodeBlobstore "spelling"

storm-core/src/jvm/org/apache/storm/zookeeper/Zookeeper.java
```diff
@@ -0,0 +1,355 @@
...
+        String leaderLockPath = conf.get(Config.STORM_ZOOKEEPER_ROOT) + "/leader-lock";
+        String id = NimbusInfo.fromConf(conf).toHostPortString();
+        AtomicReference<LeaderLatch> leaderLatchAtomicReference = new AtomicReference<>(new LeaderLatch(zk, leaderLockPath, id));
+        AtomicReference<LeaderLatchListener> leaderLatchListenerAtomicReference =
+                new AtomicReference<>(leaderLatchListenerImpl(conf, zk, leaderLatchAtomicReference.get()));
+        return new LeaderElectorImp(conf, servers, zk, leaderLockPath, id, leaderLatchAtomicReference, leaderLatchListenerAtomicReference);
+    }
+
+    //To do modify @return once don't need persistentArrayMap
+    public static PersistentArrayMap getDataWithVersion(CuratorFramework zk, String path, boolean watch) {
```
revans2: This does not have to be a PersistentArrayMap.  It can be any Map, but if you want to change it to a specific class right now, and update all of the references to it in the clojure code that is fine with me too.

storm-core/src/jvm/org/apache/storm/zookeeper/Zookeeper.java
```diff
@@ -0,0 +1,355 @@
...
+            }
+        } catch (Exception e) {
+            throw Utils.wrapInRuntime(e);
+        }
+    }
+
+    // Deletes the state inside the zookeeper for a key, for which the
+    // contents of the key starts with nimbus host port information
+    public static void deleteDodeBlobstore(CuratorFramework zk, String parentPath, String hostPortInfo) throws RuntimeException {
+        String parentnPath = Utils.normalizePath(parentPath);
```
redsanket: can we have a better name for than parentnPath, it seems confusing

storm-core/src/jvm/org/apache/storm/zookeeper/Zookeeper.java
```diff
@@ -0,0 +1,355 @@
...
+    public static Integer getVersion(CuratorFramework zk, String path, boolean watch) throws Exception {
+        String npath = Utils.normalizePath(path);
+        Stat stat = null;
+        if (existsNode(zk, npath, watch)) {
+            if (watch) {
+                stat = zk.checkExists().watched().forPath(Utils.normalizePath(path));
+            } else {
+                stat = zk.checkExists().forPath(Utils.normalizePath(path));
+            }
+            return Integer.valueOf(stat.getVersion());
```
revans2: There is potentially a race condition here that is not in the previous code.  I know this is minor but the checkExists() was there to eliminate this race.  I think what we want for this function is something more like.
```
public static Integer getVersion(CuratorFramework zk, String path, boolean watch) throws Exception {
    String npath = Utils.normalizePath(path);
    Stat stat = null;
    if (watch) {
        stat = zk.checkExists().watched().forPath(Utils.normalizePath(path));
    } else {
         stat = zk.checkExists().forPath(Utils.normalizePath(path));
     }
    return stat == null ? null : Integer.valueOf(stat.getVersion());
}
```

storm-core/src/jvm/org/apache/storm/zookeeper/Zookeeper.java
```diff
@@ -0,0 +1,355 @@
...
+        } catch (KeeperException e) {
+            // this is fine b/c we still have a watch from the successful exists call
+        } catch (Exception e) {
+            throw Utils.wrapInRuntime(e);
+        }
+
+        return null;
+    }
+
+    public static Integer getVersion(CuratorFramework zk, String path, boolean watch) throws Exception {
```
revans2: Could we please be explicit in the types of exception thrown, and not just Exception?

storm-core/src/jvm/org/apache/storm/zookeeper/Zookeeper.java
revans2: should be deleteNodeBlobstore, not deleteDodeBlobstore

storm-core/src/jvm/org/apache/storm/zookeeper/Zookeeper.java
revans2: `parentnPath` seems too close to `parentPath` for me to easily distinguish between the two.  Could you change it to `normalizedPatentPath`, or just `nParentPath`.

storm-core/src/jvm/org/apache/storm/zookeeper/Zookeeper.java
```diff
@@ -0,0 +1,355 @@
...
+
+    // Deletes the state inside the zookeeper for a key, for which the
+    // contents of the key starts with nimbus host port information
+    public static void deleteDodeBlobstore(CuratorFramework zk, String parentPath, String hostPortInfo) throws RuntimeException {
+        String parentnPath = Utils.normalizePath(parentPath);
+        List<String> childPathList = null;
+        if (existsNode(zk, parentnPath, false)) {
+            childPathList = getChildren(zk, parentnPath, false);
+        }
+        for (String child : childPathList) {
```
revans2: If the parent path does not exist we will get an NPE here that we would not get before.  Please move this under the existsNode check.

storm-core/src/jvm/org/apache/storm/zookeeper/Zookeeper.java
```diff
@@ -0,0 +1,355 @@
...
+            if (child.startsWith(hostPortInfo)) {
+                LOG.debug("deleteNode child " + child);
```
revans2: `LOG.debug("deleteNode child {}", child);` instead

storm-core/src/jvm/org/apache/storm/zookeeper/Zookeeper.java
```diff
@@ -0,0 +1,355 @@
...
+            throw Utils.wrapInRuntime(e);
+        }
+    }
+
+    public static boolean exists(CuratorFramework zk, String path, boolean watch) throws RuntimeException {
+        return existsNode(zk, path, watch);
+    }
+
+    public static NIOServerCnxnFactory mkInprocessZookeeper(String localdir, Integer port) throws IOException, InterruptedException {
+        LOG.info("Starting inprocess zookeeper at port " + port + " and dir " + localdir);
```
revans2: `LOG.info("Starting inprocess zookeeper at port {} and dir {}", port, localdir);` instead

storm-core/src/jvm/org/apache/storm/zookeeper/Zookeeper.java
```diff
@@ -0,0 +1,355 @@
...
+                break;
+            } catch (BindException e) {
+                report++;
+                if (report > limitPort) {
+                    throw new RuntimeException("No port is available to launch an inprocess zookeeper");
+                }
+            }
+        }
+        factory.startup(zk);
+        return factory;
```
revans2: Are you sure that we don't need the retport?  I see that factory has a getLocalPort method is that good enough?

storm-core/src/jvm/org/apache/storm/zookeeper/Zookeeper.java
```diff
@@ -0,0 +1,355 @@
...
+            }
+
+            @Override
+            public void notLeader() {
+                LOG.info(hostName + " lost leadership.");
+            }
+        };
+    }
+
+    public static ILeaderElector zkLeaderElector(Map conf) throws Exception {
```
revans2: Can we be explicit on the Exceptions being thrown?

storm-core/src/jvm/org/apache/storm/zookeeper/LeaderElectorImp.java
revans2: Never mind I see in the code it does the same thing.

storm-core/src/jvm/org/apache/storm/zookeeper/LeaderElectorImp.java
revans2: I think this was supposed to be the leaderlockPath, that was not set.

storm-core/src/jvm/org/apache/storm/zookeeper/LeaderElectorImp.java
```diff
@@ -0,0 +1,135 @@
...
+
+public class LeaderElectorImp implements ILeaderElector {
+    private static Logger LOG = LoggerFactory.getLogger(LeaderElectorImp.class);
+    private Map conf;
+    private List<String> servers;
+    private CuratorFramework zk;
+    private String leaderlockPath;
+    private String id;
+    private AtomicReference<LeaderLatch> leaderLatch;
+    private AtomicReference<LeaderLatchListener> leaderLatchListener;
```
revans2: Can we please mark all of these member variables as final?  None of them are ever supposed to change and it would have caught the leaderlockPath not being set.

storm-core/src/jvm/org/apache/storm/zookeeper/LeaderElectorImp.java
```diff
@@ -0,0 +1,135 @@
...
+        LOG.info("no-op for zookeeper implementation");
+    }
+
+    @Override
+    public void addToLeaderLockQueue() {
+        // if this latch is already closed, we need to create new instance.
+        try {
+            if (LeaderLatch.State.CLOSED.equals(leaderLatch.get().getState())) {
+                leaderLatch.set(new LeaderLatch(zk, leaderlockPath));
+                leaderLatchListener.set(Zookeeper.leaderLatchListenerImpl(conf, zk, leaderLatch.get()));
```
revans2: missing
`LOG.info("LeaderLatch was in closed state. Resetted the leaderLatch and listeners.");`

storm-core/src/jvm/org/apache/storm/zookeeper/LeaderElectorImp.java
```diff
@@ -0,0 +1,135 @@
...
+            if (LeaderLatch.State.LATENT.equals(leaderLatch.get().getState())) {
+                leaderLatch.get().addListener(leaderLatchListener.get());
+                leaderLatch.get().start();
+                LOG.info("Queued up for leader lock.");
+            } else {
+                LOG.info("Node already in queue for leader lock.");
+            }
+        } catch (Exception e) {
+            LOG.warn("failed to add LeaderLatchListener", e);
+        }
```
revans2: The previous code never caught an exception, and especially did not eat it.  Please either update the Interface to support Exceptions or wrap it in a RuntimeException and rethrow it.

storm-core/src/jvm/org/apache/storm/zookeeper/LeaderElectorImp.java
```diff
@@ -0,0 +1,135 @@
...
+    @Override
+    // Only started latches can be closed.
+    public void removeFromLeaderLockQueue() {
+        if (LeaderLatch.State.STARTED.equals(leaderLatch.get().getState())) {
+            try {
+                leaderLatch.get().close();
+                LOG.info("Removed from leader lock queue.");
+            } catch (IOException e) {
+                LOG.warn("failed to  close leaderLatch!!");
+            }
```
revans2: Here too, lets not eat the exception.  At least wrap it and rethrow it, but preferably update the Interface to allow us to throw it as is.

storm-core/src/jvm/org/apache/storm/zookeeper/LeaderElectorImp.java
```diff
@@ -0,0 +1,135 @@
...
+    public List<NimbusInfo> getAllNimbuses() {
+        List<NimbusInfo> nimbusInfos = new ArrayList<>();
+        try {
+            Collection<Participant> participants = leaderLatch.get().getParticipants();
+            for (Participant participant : participants) {
+                nimbusInfos.add(Zookeeper.toNimbusInfo(participant));
+            }
+        } catch (Exception e) {
+            LOG.warn("failed to get nimbuses", e);
+        }
```
revans2: Again don't eat the exception throw it, and wrap it if you cannot update the interface.

revans2:
I am done with my first pass through the code.  Overall things look good, but there are a few corner cases around error recovery and running on Windows that need to be fixed before this can go in.  Also please be sure to run all of the tests and that they all pass.

storm-core/src/jvm/org/apache/storm/zookeeper/LeaderElectorImp.java
hustfxj: sorry, I made a mistake

storm-core/src/jvm/org/apache/storm/callback/DefaultWatcherCallBack.java
hustfxj: ok

storm-core/src/jvm/org/apache/storm/zookeeper/Zookeeper.java
hustfxj: OK

storm-core/src/jvm/org/apache/storm/zookeeper/Zookeeper.java
hustfxj: yes. I will update APersistentMap， which is the same with " APersistentMap get_data_with_version(String path, boolean watch) ". If I port the zookeeper-state-factory.clj to java, I think it should be a Map.

storm-core/src/jvm/org/apache/storm/zookeeper/Zookeeper.java
hustfxj: but  " forPath "  throw Exception, what do you mean?

storm-core/src/jvm/org/apache/storm/zookeeper/LeaderElectorImp.java
hustfxj: ok

hustfxj:
Thanks @revans2 @wuchong @redsanket  for all your thorough comments,  I update the PR.

storm-core/test/clj/org/apache/storm/nimbus_test.clj
```diff
@@ -30,6 +30,7 @@
   (:import [java.util HashMap])
   (:import [java.io File])
   (:import [org.apache.storm.utils Time Utils])
+  (:import [org.apache.storm.zookeeper Zookeeper])
```
revans2: Not sure why we need this.

storm-core/test/jvm/org/apache/storm/blobstore/BlobStoreTest.java
```diff
@@ -1,3 +1,4 @@
+
```
revans2: Not sure about this one either.

storm-core/test/clj/org/apache/storm/cluster_test.clj
```diff
@@ -308,11 +309,11 @@
 
 (deftest test-cluster-state-default-acls
   (testing "The default ACLs are empty."
-    (stubbing [zk/mkdirs nil
+    (stubbing [Zookeeper/mkdirs nil
```
revans2: You cannot stub a java static function.  It only works on clojure functions.  Please take a look at how we did this in #1030 and #1043.  It is not the cleanest thing, but it does work until we can move to an architecture that is much more java like.

revans2: You have the added issue that you will probably have to use mockito to validate the calls.  If you want me to put together a pull request to your branch to fix this, because it may not be that obvious please let me know and I will.

storm-core/src/clj/org/apache/storm/testing.clj
```diff
@@ -288,11 +289,11 @@
 (defmacro with-inprocess-zookeeper
   [port-sym & body]
   `(with-local-tmp [tmp#]
-                   (let [[~port-sym zks#] (zk/mk-inprocess-zookeeper tmp#)]
+                   (let [[~port-sym zks#] (Zookeeper/mkInprocessZookeeper tmp# nil)]
```
revans2: This is not a drop in replacement because mkInprocessZookeeper is returning just zks# not ~port-sym.  You might be able to make this work by changing it to.

```
   (let [zks# (Zookeeper/mkInprocessZookeeper tmp# nil)
          ~port-sym (.getLocalPort zks#)]
```

storm-core/src/clj/org/apache/storm/testing.clj
```diff
@@ -134,7 +135,7 @@
 (defnk mk-local-storm-cluster [:supervisors 2 :ports-per-supervisor 3 :daemon-conf {} :inimbus nil :supervisor-slot-port-min 1024 :nimbus-daemon false]
   (let [zk-tmp (local-temp-path)
         [zk-port zk-handle] (if-not (contains? daemon-conf STORM-ZOOKEEPER-SERVERS)
-                              (zk/mk-inprocess-zookeeper zk-tmp))
+                              (Zookeeper/mkInprocessZookeeper zk-tmp nil))
```
revans2: This also needs to be updated similarly to below.

```
(let [zk-tmp (local-temp-path)
       zk-handle (if-not (contains? daemon-conf STORM-ZOOKEEPER-SERVERS)
                          (Zookeeper/mkInprocessZookeeper zk-tmp nil))
       zk-port (.getLocalPort zk-handle)
```

storm-core/test/clj/org/apache/storm/nimbus_test.clj
revans2: OK I see that you missed updating zk-leader-elector.

revans2: Again this is for stubbing, so if you want some help with this please let me know.

hustfxj: yes， I don't know how to be better to port zk-leader-elector by java at nimbus_test.clj

storm-core/test/clj/org/apache/storm/cluster_test.clj
hustfxj: ok , thanks.

storm-core/src/clj/org/apache/storm/zookeeper.clj
```diff
@@ -57,15 +46,15 @@
...
-                                (zk-event-types (.getType event))
-                                (.getPath event))))))))
+      (getCuratorListenable)
+      (addListener
+        (reify CuratorListener
+          (^void eventReceived [this ^CuratorFramework _fk ^CuratorEvent e]
+            (when (= (.getType e) CuratorEventType/WATCHED)
+              (let [^WatchedEvent event (.getWatchedEvent e)]
+                (watcher (ZkKeeperStates/getStateName (.getState event))
+                  (ZkEventTypes/getTypeName (.getType event))
```
revans2: These are returning Strings, not symbols. We probably want to just return the enum itself.  Also anyone who is using those symbols needs to be updated.  When I run the tests I am getting NPEs because of this.

It looks like you will mostly need to update 

./storm-core/src/clj/org/apache/storm/cluster_state/zookeeper_state_factory.clj

and

./storm-core/test/clj/org/apache/storm/cluster_test.clj

revans2:
Even with all of those changes to make the tests compile I am seeing test failures.  Please make sure that 

```
mvn clean -Prat -Pall-test integration-test
```

passes before putting up a pull request.  If you run into issues and cannot translate all of zookeeper.clj to java that is fine for a first pass others can help with different parts, but we want to be sure that we never break the build or it will take forever to stabilize later on.

storm-core/test/clj/org/apache/storm/cluster_test.clj
hustfxj:
```
    (with-open [_ (proxy [MockedZookeeper] []
                    (mkdirsImpl [] nil))]
      (stubbing [zk/mk-client (reify CuratorFramework (^void close [this] nil))]
      (mk-distributed-cluster-state {})
    ;  (verify-call-times-for mkdirsImpl 1)  //help
     ; (verify-first-call-args-for-indices mkdirsImpl [2] nil)))  //help
```
@revans2  can u help me? I update the code, but I don't how  update "verify-call-times-for " and verify-first-call-args-for-indices here.

revans2: I'll take a look and see if I can put a pull request up to your branch.

revans2: The pull request for this is up https://github.com/hustfxj/storm/pull/1

storm-core/src/jvm/org/apache/storm/zookeeper/Zookeeper.java
```diff
@@ -0,0 +1,430 @@
...
+    public static void deleteNode(CuratorFramework zk, String path){
+        try {
+            String npath = normalizePath(path);
+            if (existsNode(zk, npath, false)) {
+                zk.delete().deletingChildrenIfNeeded().forPath(normalizePath(path));
+            }
+        } catch (Exception e) {
+            if (exceptionCause(KeeperException.NodeExistsException.class, e)) {
+                // do nothing
+                LOG.info("exception", e);
```
vesense: It's better to change the log message a little bit more descriptive.

storm-core/src/jvm/org/apache/storm/zookeeper/LeaderElectorImp.java
```diff
@@ -0,0 +1,121 @@
...
+
+public class LeaderElectorImp implements ILeaderElector {
+    private static Logger LOG = LoggerFactory.getLogger(LeaderElectorImp.class);
+    final private Map conf;
+    final private List<String> servers;
+    final private CuratorFramework zk;
+    final private String leaderlockPath;
+    final private String id;
+    final private AtomicReference<LeaderLatch> leaderLatch;
+    final private AtomicReference<LeaderLatchListener> leaderLatchListener;
```
vesense: I'd like to see we have the same coding style in Storm. `final private` -> `private final`

hustfxj:
@revans2  Thanks,  I learned a lot about storm&clojure.

storm-core/src/jvm/org/apache/storm/zookeeper/LeaderElectorImp.java
hustfxj: @vesense I get it. thanks

storm-core/src/jvm/org/apache/storm/zookeeper/Zookeeper.java
hustfxj: ok

revans2:
The changes look good to me now and I am +1 on merging them in.

But because I helped contribute some of the code I want at least one other committer to also give a +1.

storm-core/src/jvm/org/apache/storm/callback/WatcherCallBack.java
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.storm.callback;
+
+import org.apache.zookeeper.Watcher;
+
+public interface WatcherCallBack {
+    public void execute(Watcher.Event.KeeperState state, Watcher.Event.EventType type, String path);
+}
```
knusbaum: Please add a newline

storm-core/src/clj/org/apache/storm/testing.clj
knusbaum: How did this make it through unit tests?

revans2: It didn't I fixed it in a different way.  I made Zookeeper/mkInprocessZookeeper match zk/mk-inprocess-zookeeper instead of returning just the factory and trying to pull the port out of the factory.

knusbaum: Ahhhh, okay. This is the part you were talking about. So this is resolved.

knusbaum:
Very minor nit. 

Otherwise, +1

storm-core/src/jvm/org/apache/storm/callback/WatcherCallBack.java
hustfxj: ok

revans2:
A line of white space is simple enough for me to do as I check it in.

PR #1050, opened 2016-01-27
[STORM-1505] Add map, flatMap and filter functions in trident stream

arunmahadevan:
`map` and `flatmap` and `filter` are common stream operations. 

Right now in trident this has to be implemented via each() which also sends the input field values
in addition to the mapped field values, so the map and flatmap should make things slightly
more efficient and easy.

### DIFF
--- a/storm-core/src/jvm/org/apache/storm/trident/operation/FlatMapFunction.java
+++ b/storm-core/src/jvm/org/apache/storm/trident/operation/FlatMapFunction.java
@@ -17,6 +17,7 @@
  */
 package org.apache.storm.trident.operation;
 
+import org.apache.storm.trident.tuple.TridentTuple;
 import org.apache.storm.tuple.Values;
 
 import java.io.Serializable;
@@ -27,10 +28,10 @@ import java.util.List;
  */
 public interface FlatMapFunction extends Serializable {
     /**
-     * Invoked by the framework for each value in a stream.
+     * Invoked by the framework for each trident tuple in a stream.
      *
-     * @param input the input value
+     * @param input the input trident tuple
      * @return an iterable over the resultant values
      */
-    Iterable<Values> execute(Values input);
+    Iterable<Values> execute(TridentTuple input);
 }
--- a/storm-core/src/jvm/org/apache/storm/trident/operation/FlatMapFunction.java
+++ b/storm-core/src/jvm/org/apache/storm/trident/operation/FlatMapFunction.java
@@ -1,0 +1,36 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.storm.trident.operation;
+
+import org.apache.storm.tuple.Values;
+
+import java.io.Serializable;
+import java.util.List;
+
+/**
+ * A one to many transformation function
+ */
+public interface FlatMapFunction extends Serializable {
+    /**
+     * Invoked by the framework for each value in a stream.
+     *
+     * @param input the input value
+     * @return an iterable over the resultant values
+     */
+    Iterable<Values> execute(Values input);
+}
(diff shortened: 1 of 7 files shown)


storm-core/src/jvm/org/apache/storm/trident/planner/processor/MapProcessor.java
```diff
@@ -0,0 +1,87 @@
...
+    public MapProcessor(Fields inputFields, Function function) {
+        _function = function;
+        _inputFields = inputFields;
+    }
+
+    @Override
+    public void prepare(Map conf, TopologyContext context, TridentContext tridentContext) {
+        List<TridentTuple.Factory> parents = tridentContext.getParentTupleFactories();
+        if(parents.size()!=1) {
+            throw new RuntimeException("Each operation can only have one parent");
```
ptgoetz: Should this be "Map operation can have only one parent"?

storm-core/src/jvm/org/apache/storm/trident/operation/FlatMapFunction.java
```diff
@@ -0,0 +1,36 @@
...
+ * A one to many transformation function
+ */
+public interface FlatMapFunction extends Serializable {
+    /**
+     * Invoked by the framework for each value in a stream.
+     *
+     * @param input the input value
+     * @return an iterable over the resultant values
+     */
+    Iterable<Values> execute(Values input);
```
ptgoetz: Why use `Values` as the input parameter and not `ITuple`? With `Values` the user will have to know the index of the value they are after, whereas `Tuple` has convenience methods like `getStringByField()`.

(Same question/comment applies to `MapFunction`.)

arunmahadevan: Had thought about it but made it `Values` to keep the argument and return types same. On rethinking agree that for exposing the convenience methods, the argument type can be changed to `TridentTuple`.

arunmahadevan:
@ptgoetz addressed review comments and also added a `filter` api so that tuples can be filtered by passing only the predicate in the filter api than using each.

ptgoetz:
@arunmahadevan you read my mind with the addition of the `filter()` method, but I'm curious why you didn't use the existing `each()` method for it?

For example:

```java
    public Stream filter(Fields inputFields, Filter filter) {
        return each(inputFields, new FilterExecutor(filter), new Fields());
    }

    public Stream filter(Filter filter) {
        return filter(getOutputFields(), filter);
    }
```

satishd:
Do we need MapProcessor/MapFunction as it looks very similar to each operation (EachProcessor)?

```java 
    public Stream each(Fields inputFields, Function function, Fields functionFields) 
    public Stream each(Function function, Fields functionFields) 
```
Agree with @ptgoetz on filter.

Good to have flatMap operation.

arunmahadevan:
@ptgoetz the current filter implementation on top of each appeared like a kludge. see https://github.com/apache/storm/blob/master/storm-core/src/jvm/org/apache/storm/trident/operation/impl/FilterExecutor.java#L38. It does not appear to emit anything but apparently its taken care at some other level. Also `each` operation always appends the input fields to the output fields, so I directly implemented it via the MapProcessor. I can refactor it as you mentioned to re-use the existing code. 
@satishd `each` emits input fields plus the new fileds, and the Function interface expects the implementor to emit within the function making it less of a pure function. map with the MapFunction interface is closer to the map operation in other streams.

ptgoetz:
@arunmahadevan lets go with the each() based approach for now since that's more familiar to existing users.

Also can you file a follow up jira or update the docs, and add it as a su task for the 1.0 release jira?

arunmahadevan:
@ptgoetz updated `filter` api to use `each` as per your suggestion. Since I wanted to see the tuples flowing through in the example topology to make sure the apis are working as expected, I also added a `peek` api similar to the one available in java8 streams. 

I created subtask for the 1.0 release jira for updating api docs. https://issues.apache.org/jira/browse/STORM-1513. Once the code changes are reviewed, I will update the docs.

ptgoetz:
@arunmahadevan Can we move the peek implementation to a separate discussion so we can discuss it independently?

And can you overload the filter() method with a fields selector parameter as I had suggested?

Once those are done, I'm +1.

arunmahadevan:
@ptgoetz Ok I will move peek implementation to a separate PR.

Are you suggesting we overload filter api with fields parameter to keep it similar to each(inputFields, filter) ? Typically filter method accepts only the predicate so I did not add that api. E.g. [Java8 filter](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html#filter-java.util.function.Predicate-). Let me know if we should add the overload.

arunmahadevan:
Raised a separate JIRA to track peek - https://issues.apache.org/jira/browse/STORM-1517

ptgoetz:
@arunmahadevan Yes, I'm talking about adding the following:

```java
public Stream filter(Fields inputFields, Filter filter)
```

I know it's minor, but I'm thinking of users having existing `Filter` implementations that may depend on specific fields being selected.

arunmahadevan:
@ptgoetz added the overloaded filter api.

ptgoetz:
+1 Thanks @arunmahadevan

PR #1051, opened 2016-01-27
[STORM-1506] It's better to be Integer about port of STORM_ZOOKEEPER_PORT&TRANSACTIONAL_ZOOKEEPER_PORT

hustfxj:
It's better to replace Object by Integer about port of STORM_ZOOKEEPER_PORT&TRANSACTIONAL_ZOOKEEPER_PORT

### DIFF
--- a/storm-core/src/jvm/org/apache/storm/blobstore/BlobStoreUtils.java
+++ b/storm-core/src/jvm/org/apache/storm/blobstore/BlobStoreUtils.java
@@ -47,7 +47,7 @@ public class BlobStoreUtils {
 
     public static CuratorFramework createZKClient(Map conf) {
         List<String> zkServers = (List<String>) conf.get(Config.STORM_ZOOKEEPER_SERVERS);
-        Object port = conf.get(Config.STORM_ZOOKEEPER_PORT);
+        Integer port = (Integer)conf.get(Config.STORM_ZOOKEEPER_PORT);
         ZookeeperAuthInfo zkAuthInfo = new ZookeeperAuthInfo(conf);
         CuratorFramework zkClient = Utils.newCurator(conf, zkServers, port, (String) conf.get(Config.STORM_ZOOKEEPER_ROOT), zkAuthInfo);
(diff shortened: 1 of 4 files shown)


unsleepy22:
LGTM +1

HeartSaVioR:
While Utils.getInt(Object) does provide safe conversion, this PR forces type conversion, which breaks unit tests (Nimbus crashed).
Please see https://travis-ci.org/apache/storm/jobs/105184033 and https://travis-ci.org/apache/storm/jobs/105184034.
We can find the reason of ClassCastException from http://clojuredocs.org/clojure.core/type, type of number literal is actually Long in Clojure.

We may want to consider Clojure unless porting is complete.

HeartSaVioR:
After porting whole things to Java, I think we may have a chance to review type usages (based on strong typing) and try not to use Object when possible. This issue seems to be a bit early.

revans2:
@hustfxj, I probably was not clear enough in my review comments that spawned this JIRA.  I was thinking along the lines of @HeartSaVioR.  Once the majority of the code is in java we can review types everywhere.  I wanted this JIRA filed so it is one that we do not forget about.  If you can make all of the tests pass before the full java conversion I am OK with it, but like @HeartSaVioR said you will need to update the clojure code to go from a long which is the clojure default to a int explicitly.

HeartSaVioR:
Yeah, I'm definitely fine to apply this PR if we can address kinds of type issues from Clojure. 
If we're not sure, let's port first and take a look at STORM-1506 again.

HeartSaVioR:
Btw, you may want to see whole tests succeed at least once from local dev. before submitting a pull request.
Storm has random test failures so you may have to run tests several times, but Storm normally doesn't have consistent test failures.

hustfxj:
@HeartSaVioR @revans2  Thanks for your explanation. In fact I'm not particularly familiar with clojure. So I run into issues with local tests. But I will try my best to port clojure to java for storm. Thanks again!

PR #1059, opened 2016-01-30
STORM-1242: migrate backtype.storm.command.config-value to java

revans2:


storm-core/src/jvm/org/apache/storm/command/ConfigValue.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+package org.apache.storm.command;
+
+import java.util.Map;
+
+import org.apache.storm.utils.ConfigUtils;
+
+public class ConfigValue {
+    public static void main(String [] args) {
+        String name = args[0];
```
redsanket: do we need to check for presence of elements before accessing them.

redsanket:
LGTM, minor clarification

storm-core/src/jvm/org/apache/storm/command/ConfigValue.java
revans2: The old code didn't, and the python code blows up even worse if you don't pass in the parameter.

kishorvpatil:
+1

PR #1060, opened 2016-02-01
[STORM-1245] port backtype.storm.daemon.acker to java

bastiliu:
Increase bucket number of pending map from 2 to 3 to avoid timeout too fast.


bastiliu:
CI was failed. Some clojure test case needs to be updated, so close this PR. The PR will be set up again after the updating of clojure test case.

PR #1061, opened 2016-02-01
STORM-1511 min/max operations support on a trident stream.

satishd:


arunmahadevan:
Haven't gone through the implementation details yet, but from the api

`public <T> Stream max(String inputFieldName, String outputFieldName, Comparator<T> comparator)`

it looks like the comparison is based on a single field and the trident tuples itself are not compared.

Since we are dealing with a stream of `TridentTuple`, can we have an api that just takes a `Comparator<TridentTuple>`  (similar to the Java8 [min/max](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html#max-java.util.Comparator-)) and passes the entire tuple in the compareTo? The comparison can then be based on one or more fields as users wish and the max tuple is finally emitted. 

`Stream max(Comparator<TridentTuple> comparator)`

With this api user need not care about the input or output fields but only the values.

PR #1064, opened 2016-02-01
[STORM-1521] When using Kerberos login from keytab with multiple bolts/executors ticket is not renewed

dbahir:
When using a topology that has multiple executors in same process or multiple HBase bolts the tickets do not get renewed. After the ticket expiration the bolts are no longer able to write/read from HBase. That is caused by a login from keytab for every thread, this fixe ensures that there will only be one login per process.

### DIFF
--- a/external/storm-hbase/src/main/java/org/apache/storm/hbase/security/HBaseSecurityUtil.java
+++ b/external/storm-hbase/src/main/java/org/apache/storm/hbase/security/HBaseSecurityUtil.java
@@ -39,26 +39,34 @@ public class HBaseSecurityUtil {
 
     public static final String STORM_KEYTAB_FILE_KEY = "storm.keytab.file";
     public static final String STORM_USER_NAME_KEY = "storm.kerberos.principal";
+    private static  UserProvider legacyProvider = null;
 
     public static UserProvider login(Map conf, Configuration hbaseConfig) throws IOException {
         //Allowing keytab based login for backward compatibility.
-        UserProvider provider = UserProvider.instantiate(hbaseConfig);
-        if (conf.get(TOPOLOGY_AUTO_CREDENTIALS) == null ||
-                !(((List) conf.get(TOPOLOGY_AUTO_CREDENTIALS)).contains(AutoHBase.class.getName()))) {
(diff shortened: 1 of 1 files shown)


harshach:
@dbahir patch looks good to me. Can you squash multiple commits into single commit. Thanks

harshach:
@dbahir I think this issue will be in storm-hdfs as well https://github.com/apache/storm/blob/master/external/storm-hdfs/src/main/java/org/apache/storm/hdfs/common/security/HdfsSecurityUtil.java . Can you add changes there as well.

dbahir:
Squashed the commits.

Took a look at the HDFS bolt code path there is a somewhat different and requires additional testing. Created https://issues.apache.org/jira/browse/STORM-1535 to handle it, will create a separate pull request

harshach:
+1. @Parth-Brahmbhatt can you review this as well.

Parth-Brahmbhatt:
@dbahir Does the legacy provider take care of logging in only once internally?

external/storm-hbase/src/main/java/org/apache/storm/hbase/security/HBaseSecurityUtil.java
```diff
@@ -39,26 +39,34 @@
...
+                !(((List) conf.get(TOPOLOGY_AUTO_CREDENTIALS)).contains(AutoHBase.class.getName())))) {
             LOG.info("Logging in using keytab as AutoHBase is not specified for " + TOPOLOGY_AUTO_CREDENTIALS);
-            if (UserGroupInformation.isSecurityEnabled()) {
-                String keytab = (String) conf.get(STORM_KEYTAB_FILE_KEY);
-                if (keytab != null) {
-                    hbaseConfig.set(STORM_KEYTAB_FILE_KEY, keytab);
+            //insure that if keytab is used only one login per process executed
+            if(legacyProvider == null) {
+                synchronized (HBaseSecurityUtil.class) {
+                    if(legacyProvider == null) {
```
dbahir: There is a double check pattern here which locks on the class, the lock on the class will ensure that the legacy provider is instantiated and logged in only once per process

Parth-Brahmbhatt:
Not sure how i missed that you were creating a singleton :-). +1.

PR #1067, opened 2016-02-02
STORM-1515: Reset LocalState if corrupted after a hard reboot on Windows

torbiak:
On Windows LocalState IO requests interrupted by a hard reboot can
result in a file full of NULs, similar to the empty-file corruption seen in STORM-307.

I've fixed this for 0.9.x first since I haven't upgraded to 0.10 yet. The fix for 0.10 will be slightly different due to the move to Thrift for LocalState serialization.

It might be desirable to catch EOFException instead of checking serialized.length, since it could cover more cases of corruption, like a partially-written serialization stream.

### DIFF
--- a/storm-core/src/jvm/backtype/storm/utils/LocalState.java
+++ b/storm-core/src/jvm/backtype/storm/utils/LocalState.java
@@ -25,6 +25,7 @@ import java.io.File;
 import java.util.Map;
 import java.util.HashMap;
 import java.io.IOException;
+import java.io.StreamCorruptedException;
 
 
 /**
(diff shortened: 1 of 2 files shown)


frison:
I see no way that CI failed because of this commit.

revans2:
The change looks fine to me, but we need to pull this into master, 1.x-branch and 0.10.x-branch before we can pull it into 0.9.x-branch.

knusbaum:
@torbiak Would you like to do work on this, or should it be closed?

torbiak:
I won't be able to work on it anytime soon.

PR #1068, opened 2016-02-02
[STORM-1519] - Storm syslog logging not confirming to RFC5426 3.1

jerrypeng:


erikdw:
This seems like it could have a *major* impact on performance, no?   It's not clear to me what is so horrible in STORM-1519 that we need to force a disk write for every single log message.

jerrypeng:
Well this only changes that config for the syslog appender.  The syslog appender writes data to a network socket.  This doesn't change the other appenders that will actually write to disk.

jerrypeng:
If a user has a log parser that adheres to RFC5426, then how log4j2 is configured to write to syslog is a problem.

erikdw:
@jerrypeng : ahhh.... thanks for the extra info, both points make sense. To be pedantic, that detail woulda been cool in the (empty :wink:) description.

jerrypeng:
@erikdw the jira i filed contained more detailed information

https://issues.apache.org/jira/browse/STORM-1519

erikdw:
@jerrypeng : heh, yep, I read the whole bug before my first reply.  And it had this note:
* "immediateFlush - When set to true - the default, each write will be followed by a flush. This will guarantee the data is written to disk but could impact performance."

And *that* is what scared me into replying actually!   The log4j2 docs should probably just be updated to "written to disk or network".

revans2:
The change looks fine to me +1.

@erikdw do you have any other concerns about this?

erikdw:
@revans2 : not at all!  I just misunderstood the implications, totally my bad.

PR #1069, opened 2016-02-03
[STORM-1245] port backtype.storm.daemon.acker to java

bastiliu:
1. Increase bucket number of pending map from 2 to 3 to avoid timeout too fast.
2. In clojure testing, creation of acker bolt will be overridden in with-tracked-cluster. After porting the implementation of acker to java, it seems there is no way to do similar override for java method. So, the mk-acker-bolt of acker.clj was kept as temporary solution to make the clojure testing pass. It can be removed after the porting of common.clj and relative testing modules. 
Anyway, if there is a solution or workaround for this issue, please help to comment.


storm-core/src/jvm/org/apache/storm/daemon/AckerBolt.java
```diff
@@ -0,0 +1,129 @@
...
+            if (curr != null) {
+                curr.updateAck(input.getValue(1));
+            } else {
+                curr = new AckObject();
+                curr.val = input.getLong(1);
+                pending.put(id, curr);
+            }
+        } else if (ACKER_FAIL_STREAM_ID.equals(streamId)) {
+            if (curr == null) {
+                // The tuple has been already timeout or failed. So, do nothing
```
abhishekagarwal87: If the bolt explicitly fails an input tuple, is it possible that bolt's fail is received before the init?

bastiliu: Yes, it is possible. But it is difficult to distinguish what the actual root cause is here ("already timeout", "fail" or "receive fail before init").

abhishekagarwal87: It shouldn't matter. If you put ackObject again, ackObject will timeout without any harm. This seems to be same implementation in current acker in clojure. can you cross check? I don't trust my clojure skills.

storm-core/src/jvm/org/apache/storm/daemon/AckerBolt.java
```diff
@@ -0,0 +1,129 @@
...
+                List values = Utils.mkList(id);
+                collector.emitDirect(task, ACKER_ACK_STREAM_ID, values);
+            } else {
+                if (curr.failed) {
+                    pending.remove(id);
+                    List values = Utils.mkList(id);
+                    collector.emitDirect(task, ACKER_FAIL_STREAM_ID, values);
+                }
+            }
+        } else {
```
abhishekagarwal87: No need of else block

storm-core/src/jvm/org/apache/storm/daemon/AckerBolt.java
```diff
@@ -0,0 +1,129 @@
...
+            LOG.warn("Unknown source stream {} from task-{}", streamId, input.getSourceTask());
+            return;
+        }
+
+        Integer task = curr.spoutTask;
+        if (task != null) {
+            if (curr.val == 0) {
+                pending.remove(id);
+                List values = Utils.mkList(id);
```
abhishekagarwal87: you can use Values class.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1374,5 +1374,61 @@ public static RuntimeException wrapInRuntime(Exception e){
...
+
+    public static <T> long bitXorValsSets(java.util.Set<T> vals) {
+        long rtn = 0l;
+        for (T n : vals) {
+            rtn = bitXor(rtn, n);
+        }
+        return rtn;
+    }
+
+    public static long bitXor(Object a, Object b) {
```
abhishekagarwal87: given that bixXor falls in critical path, is it possible to avoid instanceof checks. why is there special treatment for Set type?

storm-core/src/jvm/org/apache/storm/daemon/AckerBolt.java
```diff
@@ -0,0 +1,129 @@
...
+                pending.put(id, curr);
+            }
+        } else if (ACKER_FAIL_STREAM_ID.equals(streamId)) {
+            if (curr == null) {
+                // The tuple has been already timeout or failed. So, do nothing
+                return;
+            }
+            curr.failed = true;
+        } else {
+            LOG.warn("Unknown source stream {} from task-{}", streamId, input.getSourceTask());
```
abhishekagarwal87: this will be called for every tickTuple.

storm-core/src/jvm/org/apache/storm/daemon/AckerBolt.java
```diff
@@ -0,0 +1,129 @@
...
+    public void prepare(Map stormConf, TopologyContext context, OutputCollector collector) {
+        this.collector = collector;
+        this.pending = new RotatingMap<Object, AckObject>(TIMEOUT_BUCKET_NUM);
+    }
+
+    @Override
+    public void execute(Tuple input) {
+        if (TupleUtils.isTick(input)) {
+            Map<Object, AckObject> tmp = pending.rotate();
+            LOG.debug("Number of timeout tuples:{}", tmp.size());
```
abhishekagarwal87: should the function return from here?

storm-core/src/jvm/org/apache/storm/daemon/AckerBolt.java
bastiliu: Addressed. It should return after processing tick tuple.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
bastiliu: This piece of code comes from JStorm. There is some purpose for extension in JStorm. But anyway, I will remove the extra checking.

storm-core/src/clj/org/apache/storm/daemon/acker.clj
```diff
@@ -18,74 +18,24 @@
...
+  (:use [org.apache.storm config util])
   (:gen-class
-   :init init
-   :implements [org.apache.storm.task.IBolt]
-   :constructors {[] []}
-   :state state ))
+    :init init
+    :implements [org.apache.storm.task.IBolt]
+    :constructors {[] []}
+    :state state))
```
redsanket: indentation changed?

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1374,5 +1374,61 @@ public static RuntimeException wrapInRuntime(Exception e){
...
+        ArrayList<V> rtn = new ArrayList<V>();
+        if (args != null) {
+            for (V o : args) {
+                rtn.add(o);
+            }
+        }
+        return rtn;
+    }
+
+    public static <V> List<V> mkList(Collection<V> args) {
```
redsanket: can we call all these methods makeList, more of javaness then clojureness

storm-core/src/clj/org/apache/storm/daemon/acker.clj
```diff
@@ -18,74 +18,24 @@
   (:import [org.apache.storm.tuple Tuple Fields])
```
revans2: Lets just delete this file entirely.  There is not reason to play the games with wrapping the acker bolt and using the Container so that the acker can be serializable.  AckerBolt is already serializable and has the ID set properly so lets just delete this file.

storm-core/src/jvm/org/apache/storm/daemon/AckerBolt.java
```diff
@@ -0,0 +1,128 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger(AckerBolt.class);
+
+    private static final long serialVersionUID = 4430906880683183091L;
+
+    public static final String ACKER_COMPONENT_ID = "__acker";
+    public static final String ACKER_INIT_STREAM_ID = "__ack_init";
+    public static final String ACKER_ACK_STREAM_ID = "__ack_ack";
+    public static final String ACKER_FAIL_STREAM_ID = "__ack_fail";
+
+    public static final int TIMEOUT_BUCKET_NUM = 3;
```
revans2: This is minor but the original code has 2 buckets, not 3.  Is there a reason for the change?

storm-core/src/jvm/org/apache/storm/daemon/AckerBolt.java
```diff
@@ -0,0 +1,128 @@
...
+        AckObject curr = pending.get(id);
+        if (ACKER_INIT_STREAM_ID.equals(streamId)) {
+            if (curr == null) {
+                curr = new AckObject();
+                curr.val = input.getLong(1);
+                curr.spoutTask = input.getInteger(2);
+                pending.put(id, curr);
+            } else {
+                // If receiving bolt's ack before the init message from spout, just update the xor value.
+                curr.updateAck(input.getValue(1));
```
revans2: Can we replace this with 
```
curr.updateAck(input.getLong(1));
```

It didn't really matter before but now that we care about types moving the cast up earlier is cleaner in my opinion.

storm-core/src/jvm/org/apache/storm/daemon/AckerBolt.java
```diff
@@ -0,0 +1,128 @@
...
+    private RotatingMap<Object, AckObject> pending;
+
+    private class AckObject {
+        public long val = 0L;
+        public Integer spoutTask = null;
+        public boolean failed = false;
+
+        // val xor value
+        public void updateAck(Object value) {
+            val = Utils.bitXor(val, value);
```
revans2: Can we inline the xor, and if bitXor is not used anywhere else delete it from Utils.

storm-core/src/jvm/org/apache/storm/daemon/AckerBolt.java
revans2: In the previous code it was an if/else.  So we can either return here, or wrap the rest of execute in an else.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1374,5 +1374,37 @@ public static RuntimeException wrapInRuntime(Exception e){
             return new RuntimeException(e);
         }
     }
+
+    public static long bitXor(Object a, Object b) {
+        return  ((Long) a) ^ ((Long) b);
+    }
+
+    public static <V> List<V> makeList(V... args) {
```
revans2: Can we replace this with `java.utils.Arrays.asList()` and inline it everywhere.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1374,5 +1374,37 @@ public static RuntimeException wrapInRuntime(Exception e){
...
+        ArrayList<V> rtn = new ArrayList<V>();
+        for (V o : args) {
+            rtn.add(o);
+        }
+        return rtn;
+    }
+
+    public static <V> List<V> makeList(java.util.Set<V> args) {
```
revans2: This is not used, lets delete it.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1374,5 +1374,37 @@ public static RuntimeException wrapInRuntime(Exception e){
...
+        ArrayList<V> rtn = new ArrayList<V>();
+        if (args != null) {
+            for (V o : args) {
+                rtn.add(o);
+            }
+        }
+        return rtn;
+    }
+
+    public static <V> List<V> makeList(Collection<V> args) {
```
revans2: This is not used, lets delete it.

storm-core/src/jvm/org/apache/storm/daemon/AckerBolt.java
```diff
@@ -0,0 +1,128 @@
...
+                }
+            }
+        }
+
+        collector.ack(input);
+    }
+
+    @Override
+    public void cleanup() {
+
```
revans2: Can we add a small comment indicating the this is empty on purpose.

revans2:
Overall it looks good to me.  Just a few relatively minor cleanup comments.

storm-core/src/jvm/org/apache/storm/daemon/AckerBolt.java
bastiliu: If there is only 2 buckets, the timeout period is at most equal to the "message timeout secs". But for the most worse case that acker just receives ack_init tuple from spout and then acker process rotate operation, the tuple will be considered as timeout message immediately. For 3 buckets, the timeout period is at least equal to "message timeout secs" while the timeout period is at most 2 times of "message timeout secs" for the most worse case.

storm-core/src/jvm/org/apache/storm/daemon/AckerBolt.java
bastiliu: The bitXor operation is probably required when migrating the ack relative operation for spout or bolt in the future.

storm-core/src/clj/org/apache/storm/daemon/acker.clj
bastiliu: In clojure testing, creation of acker bolt will be overridden in with-tracked-cluster. After porting the implementation of acker to java, it seems there is no way to do a similar override for java method. So, the mk-acker-bolt of acker.clj was kept as temporary solution to make the clojure testing pass. It can be removed after the porting of common.clj and relative testing modules. Is there a solution or workaround for this issue?

storm-core/src/clj/org/apache/storm/daemon/acker.clj
bastiliu: reverted

storm-core/src/jvm/org/apache/storm/daemon/AckerBolt.java
abhishekagarwal87: val is being autoboxed everytime in this call. can the bitXor signature be changed to long bitXor(long a, long b)?

storm-core/src/jvm/org/apache/storm/daemon/AckerBolt.java
revans2: Sounds good.

storm-core/src/jvm/org/apache/storm/daemon/AckerBolt.java
revans2: Yes it is the same behavior I checked in my review.

bastiliu:
@abhishekagarwal87 @redsanket @revans2  Sorry for the late response. I have addressed all your comments. Please help to check. Thanks.

longdafeng:
+1, 

The acker logic is correct, come from JStorm.

In JStorm, it handle two more case:
Acker.ACKER_INIT_STREAM_ID's tuple is after Acker.ACKER_ACK_STREAM_ID's tuple or Acker.ACKER_FAIL_STREAM_ID's tuple.

@bastiliu , the logic handling the above two cases should also port to 1.x branch

storm-core/src/jvm/org/apache/storm/daemon/Acker.java
```diff
@@ -0,0 +1,127 @@
...
+    public static final String ACKER_ACK_STREAM_ID = "__ack_ack";
+    public static final String ACKER_FAIL_STREAM_ID = "__ack_fail";
+
+    public static final int TIMEOUT_BUCKET_NUM = 3;
+
+    private OutputCollector collector;
+    private RotatingMap<Object, AckObject> pending;
+
+    private class AckObject {
+        public long val = 0L;
```
abhishekagarwal87: can you change this initial value to -1. Reason being that 0 indicates a certain status (ack tree completed) and it may not be good idea to start with that default state. In clojure code, default value was nil.

storm-core/src/jvm/org/apache/storm/daemon/Acker.java
```diff
@@ -0,0 +1,127 @@
...
+            if (curr != null) {
+                curr.updateAck(input.getLong(1));
+            } else {
+                curr = new AckObject();
+                curr.val = input.getLong(1);
+                pending.put(id, curr);
+            }
+        } else if (ACKER_FAIL_STREAM_ID.equals(streamId)) {
+            if (curr == null) {
+                // The tuple has been already timeout or failed. So, do nothing
```
abhishekagarwal87: what is final decision for the scenario wherein FAIL occurs before INIT? The clojure code is marking the ack object failed and sends a failure signal to spout. The tuple will get retransmitted nonetheless though it will take longer time to happen. 
@revans2 @longdafeng  - your thoughts?

bastiliu: Just for my understanding, it is better that just return here to ensure only one failure for a certain tuple, rather than sending a failure back to spout. Since it is difficult to tell what the real root cause is here(FAIL occurs before INIT, or tuple has been already timeout or failed), if we just send a failure back here, for the latter scenarios, acker will send more than one message back to spout. It will cause incorrect stats of emitting tuple of acker, and potential duplicated calling of fail() in spout (it depends on the implementation of spout).

storm-core/src/jvm/org/apache/storm/daemon/Acker.java
bastiliu: Actually, in clojure code, it also uses 0 as default value since the val will be xor with the value get from tuple.

abhishekagarwal87: 0 is assigned as default-value only in the update-ack call. let's say FAIL arrives first. `cond (= 0 (:val curr)` will return false in  the clojure code because val is nil. If val was zero instead, acker would have acked a failed message. The reason is value is checked first and failure is checked afterwards. It is not a problem in your changes since code just returns if fail comes before init.

storm-core/src/jvm/org/apache/storm/daemon/Acker.java
abhishekagarwal87: That is correct @bastiliu. It would be rare but it is a possibility. Anyway, spout doesn't differentiate between failures and timeouts so I am fine with it.

longdafeng: @bastiliu  @abhishekagarwal87 , after check the raw clojure code, we had better to add send fail message to spout, in the scenario where FAIL occurs before INIT. Due to in the spout, the spout will only call once fail(), even receive multiple FAIL message from ACKER.

abhishekagarwal87: Becuase failed message is removed from pending map. Good catch @longdafeng

storm-core/src/jvm/org/apache/storm/daemon/Acker.java
bastiliu: Just checked the implementation of clojure code again. Before sending failure to spout, acker wil check if target spout task id is null first. If target spout task is nil, acker will not send out failure. So, for the case of FAIL arriving first, acker will hold on the failure response untill it receives INIT.

storm-core/src/jvm/org/apache/storm/daemon/Acker.java
bastiliu: OK. I will update the code.

PR #1071, opened 2016-02-03
[STORM-1273] port backtype.storm.cluster to java

hustfxj:
Sorry. Some tests don’t pass, and I will check and fix it. I now check in code,  because I want to get some guidance before the Chinese New Year.


revans2:
The travis CI tests failed with

```
13881 [timer] ERROR o.a.s.d.nimbus - Error when processing event
java.lang.NullPointerException
	at org.apache.storm.cluster.Cluster.convertExecutorBeats(Cluster.java:202) ~[classes/:?]
	at org.apache.storm.cluster.StormZkClusterState.executorBeats(StormZkClusterState.java:327) ~[classes/:?]
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[?:1.7.0_76]
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57) ~[?:1.7.0_76]
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[?:1.7.0_76]
	at java.lang.reflect.Method.invoke(Method.java:606) ~[?:1.7.0_76]
	at clojure.lang.Reflector.invokeMatchingMethod(Reflector.java:93) ~[clojure-1.7.0.jar:?]
	at clojure.lang.Reflector.invokeInstanceMethod(Reflector.java:28) ~[clojure-1.7.0.jar:?]
	at org.apache.storm.daemon.nimbus$update_heartbeats_BANG_.invoke(nimbus.clj:589) ~[classes/:?]
```

storm-core/src/clj/org/apache/storm/daemon/common.clj
```diff
@@ -13,14 +13,16 @@
 ;; WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 ;; See the License for the specific language governing permissions and
 ;; limitations under the License.
+;TopologyActionOptions TopologyStatus StormBase RebalanceOptions KillOptions
```
revans2: Not sure what this comment means could you remove it.

storm-core/src/clj/org/apache/storm/pacemaker/pacemaker_state_factory.clj
```diff
@@ -47,7 +43,7 @@
     (reify
       ClusterState
       ;; Let these pass through to the zk-state. We only want to handle heartbeats.
-      (register [this callback] (.register zk-state callback))
+      (register [this callback] (.register zk-state callback))  ; need update callback, have questions?? callback is IFn here
```
revans2: If the callback takes a defined set of parameters that are used consistently lets either create or reuse an interface and then wrap the callback function in a reify call to create an instance of the interface for the callback to talk to.

If it does not take any parameters, it just is a function that is triggered than make it a Runnable.  But I think in this case it is the former, not the latter.

storm-core/src/jvm/org/apache/storm/callback/Callback.java
```diff
@@ -18,6 +18,9 @@
 
 package org.apache.storm.callback;
 
+import clojure.lang.IFn;
+
+// To remove IFn after porting all callbacks to java
 public interface Callback {
     public <T> Object execute(T... args);
```
revans2: I don't like this interface.  The generic <T> for the parameters is just confusing because it implies that all of the parameters should be of the same type, when in reality they are not.  I would much rather see something like

```
public interface ZKStateChangedCallback {
    public void changed(Watcher.Event.EventType type, String path);
}
```

storm-core/src/jvm/org/apache/storm/cluster/Cluster.java
```diff
@@ -0,0 +1,236 @@
...
+        WORKERBEATS_SUBTREE = ZK_SEPERATOR + WORKERBEATS_ROOT;
+        BACKPRESSURE_SUBTREE = ZK_SEPERATOR + BACKPRESSURE_ROOT;
+        ERRORS_SUBTREE = ZK_SEPERATOR + ERRORS_ROOT;
+        BLOBSTORE_SUBTREE = ZK_SEPERATOR + BLOBSTORE_ROOT;
+        BLOBSTORE_MAX_KEY_SEQUENCE_NUMBER_SUBTREE = ZK_SEPERATOR + BLOBSTORE_MAX_KEY_SEQUENCE_NUMBER_ROOT;
+        NIMBUSES_SUBTREE = ZK_SEPERATOR + NIMBUSES_ROOT;
+        CREDENTIALS_SUBTREE = ZK_SEPERATOR + CREDENTIALS_ROOT;
+        LOGCONFIG_SUBTREE = ZK_SEPERATOR + LOGCONFIG_ROOT;
+        PROFILERCONFIG_SUBTREE = ZK_SEPERATOR + PROFILERCONFIG_ROOT;
+    }
```
revans2: I would prefer to see these inlined above, instead of in a static block.

for example

```
public static final String ASSIGNMENTS_SUBTREE = ZK_SEPERATOR + ASSIGNMENTS_ROOT;
```

storm-core/src/jvm/org/apache/storm/cluster/Cluster.java
```diff
@@ -0,0 +1,236 @@
...
+    // tests by subclassing.
+    private static final Cluster INSTANCE = new Cluster();
+    private static Cluster _instance = INSTANCE;
+
+    /**
+     * Provide an instance of this class for delegates to use.  To mock out
+     * delegated methods, provide an instance of a subclass that overrides the
+     * implementation of the delegated method.
+     *
+     * @param u a Zookeeper instance
```
revans2: Please update the comment to be Cluster not Zookeeper

storm-core/src/jvm/org/apache/storm/cluster/Cluster.java
```diff
@@ -0,0 +1,236 @@
...
+     * the class to its original functionality when mocking is no longer
+     * desired.
+     */
+    public static void resetInstance() {
+        _instance = INSTANCE;
+    }
+
+    public static List<ACL> mkTopoOnlyAcls(Map topoConf) throws NoSuchAlgorithmException {
+        List<ACL> aclList = null;
+        String payload = (String)topoConf.get(Config.STORM_ZOOKEEPER_AUTH_PAYLOAD);
```
revans2: This should be STORM_ZOOKEEPER_TOPOLOGY_AUTH_PAYLOAD

storm-core/src/jvm/org/apache/storm/cluster/Cluster.java
```diff
@@ -0,0 +1,236 @@
...
+    public static String backpressurePath(String stormId, String node, Long port) {
+        return backpressureStormRoot(stormId) + ZK_SEPERATOR + node + "-" + port;
+    }
+
+    public static String errorStormRoot(String stormId) {
+        return ERRORS_SUBTREE + ZK_SEPERATOR + stormId;
+    }
+
+    public static String errorPath(String stormId, String componentId) throws UnsupportedEncodingException {
+        return errorStormRoot(stormId) + ZK_SEPERATOR + URLEncoder.encode(componentId, "UTF-8");
```
revans2: I know the last time I told you to let the exceptions bubble up, and for the most part I still agree with that, but in this case the UnsupportedEncodingException should never really be thrown.  If "UTF-8" is not supported by the JVM, which it should always be supported, we have some really big problems.  So I am OK with wrapping this one in a RuntimeException.

storm-core/src/jvm/org/apache/storm/cluster/Cluster.java
```diff
@@ -0,0 +1,236 @@
...
+    }
+
+    public static <T> T maybeDeserialize(byte[] serialized, Class<T> clazz){
+        if (serialized != null){
+            return Utils.deserialize(serialized, clazz);
+        }
+        return null;
+    }
+
+    //Ensures that we only return heartbeats for executors assigned to this worker
```
revans2: Could you make this a javadoc?  That is what the doc string represents in clojure.

storm-core/src/jvm/org/apache/storm/cluster/Cluster.java
```diff
@@ -0,0 +1,236 @@
...
+    public static Map<ExecutorInfo, ClusterWorkerHeartbeat> convertExecutorBeats(List<ExecutorInfo> executors, ClusterWorkerHeartbeat workerHeartbeat){
```
revans2: Lets make this private.  I know it want before but it is only used in this class so it does not need to be public.

storm-core/src/jvm/org/apache/storm/cluster/Cluster.java
```diff
@@ -0,0 +1,236 @@
...
+
+    public static String profilerConfigPath(String stormId) {
+        return PROFILERCONFIG_SUBTREE + ZK_SEPERATOR + stormId;
+    }
+
+    public static String profilerConfigPath(String stormId, String host, Long port, ProfileAction requestType) {
+        return profilerConfigPath(stormId) + ZK_SEPERATOR + host + "_" + port + "_" + requestType;
+    }
+
+    public static <T> T maybeDeserialize(byte[] serialized, Class<T> clazz){
```
revans2: Lets make this private.  a ```(defn- ``` means that the definition is private to the namespace.  We should try and follow that where necessary.

storm-core/src/jvm/org/apache/storm/cluster/Cluster.java
```diff
@@ -0,0 +1,236 @@
...
+
+import java.io.UnsupportedEncodingException;
+import java.net.URLEncoder;
+import java.security.NoSuchAlgorithmException;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class Cluster {
```
revans2: Lets rename this ClusterUtils, because it really is a Utils class.

storm-core/src/jvm/org/apache/storm/cluster/DistributedClusterState.java
```diff
@@ -0,0 +1,268 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+public class DistributedClusterState implements ClusterState {
```
revans2: I am trying to understand how mk-distributed-cluster-state and mk-storm-cluster-state relate to DistributedClusterState and StormZKClusterState.

mk-distributed-cluster-state just instantiates a class pointed to by Config.STORM_CLUSTER_STATE_STORE with a default of "org.apache.storm.cluster_state.zookeeper_state_factory" and then calls mkInstance on it.

That functionality appears to be completely lost in this patch, and the StormZKStateStore is hardcoded everywhere, we need to fix that.
Now DistributedClusterState appears to be the equivalent of what zookeeper_state_factory was doing without the factory pattern, and without any reference to zookeeper in the name, while StormZKClusterState actually looks like the concrete implementation of ClusterState and does not really have any direct dependency on Zookeeper.

So to clean things up a bit lets do the following.

Rename the ClusterState interface to be StateStorage and ClusterStateFactory to be StateStorageFactory, all of them in the org.apache.storm.cluster package.  Along with that please change `:cluster-state` in the worker-data to be `:state-store`. This class, DistributedClusterState, would be renamed to ZKStateStorage and an appropriate ZKStateStorageFactory needs to be created.  pacemaker_state_factory.clj needs to be reverted and updated to extend the newly renamed StateStorageFactory, and to return a StateStorage instance.

ClusterState no longer needs to be an interface and implementation split off from one another.  It should be a single class that uses the StateStore that is passed into it in the constructor.

mk-distributed-cluster-state should be a static method in mkStateStorage that should probably be a part of ClusterUtils.

revans2:
@hustfxj there are some major changes that need to be made to the way you setup your class hierarchy and the naming of those classes.  I stopped looking at the details of the rest of the patch until those changes can be addressed.

storm-core/src/jvm/org/apache/storm/cluster/DistributedClusterState.java
```diff
@@ -0,0 +1,268 @@
...
+
+    @Override
+    public void set_ephemeral_node(String path, byte[] data, List<ACL> acls) {
+        Zookeeper.mkdirs(zkWriter, parentPath(path), acls);
+        if (Zookeeper.exists(zkWriter, path, false)) {
+            try {
+                Zookeeper.setData(zkWriter, path, data);
+            } catch (RuntimeException e) {
+                if (Utils.exceptionCauseIsInstanceOf(KeeperException.NodeExistsException.class, e)) {
+                    Zookeeper.createNode(zkWriter, path, data, CreateMode.EPHEMERAL, acls);
```
abhishekagarwal87: you want to recreate if it already exists?

storm-core/src/jvm/org/apache/storm/cluster/DistributedClusterState.java
```diff
@@ -0,0 +1,268 @@
...
+        return Zookeeper.createNode(zkWriter, path, data, CreateMode.EPHEMERAL_SEQUENTIAL, acls);
+    }
+
+    @Override
+    public void mkdirs(String path, List<ACL> acls) {
+        Zookeeper.mkdirs(zkWriter, path, acls);
+    }
+
+    @Override
+    public void delete_node(String path) {
```
abhishekagarwal87: should the function names be camelCase instead?

storm-core/src/jvm/org/apache/storm/cluster/DistributedClusterState.java
```diff
@@ -0,0 +1,268 @@
...
+    }
+
+    @Override
+    public void set_ephemeral_node(String path, byte[] data, List<ACL> acls) {
+        Zookeeper.mkdirs(zkWriter, parentPath(path), acls);
+        if (Zookeeper.exists(zkWriter, path, false)) {
+            try {
+                Zookeeper.setData(zkWriter, path, data);
+            } catch (RuntimeException e) {
+                if (Utils.exceptionCauseIsInstanceOf(KeeperException.NodeExistsException.class, e)) {
```
abhishekagarwal87: It should be NoNodeException.

storm-core/src/jvm/org/apache/storm/cluster/DistributedClusterState.java
```diff
@@ -0,0 +1,268 @@
...
+        Zookeeper.deleteNode(zkWriter, path);
+    }
+
+    @Override
+    public void set_ephemeral_node(String path, byte[] data, List<ACL> acls) {
+        Zookeeper.mkdirs(zkWriter, parentPath(path), acls);
+        if (Zookeeper.exists(zkWriter, path, false)) {
+            try {
+                Zookeeper.setData(zkWriter, path, data);
+            } catch (RuntimeException e) {
```
abhishekagarwal87: you can catch Exception here instead of just RuntimeException.

storm-core/src/jvm/org/apache/storm/cluster/StormZkClusterState.java
```diff
@@ -0,0 +1,683 @@
...
+                    } else if (root.equals(Cluster.STORMS_ROOT) && size > 1) {
+                        issueMapCallback(stormBaseCallback, toks.get(1));
+                    } else if (root.equals(Cluster.CREDENTIALS_ROOT) && size > 1) {
+                        issueMapCallback(credentialsCallback, toks.get(1));
+                    } else if (root.equals(Cluster.LOGCONFIG_ROOT) && size > 1) {
+                        issueMapCallback(logConfigCallback, toks.get(1));
+                    } else if (root.equals(Cluster.BACKPRESSURE_ROOT) && size > 1) {
+                        issueMapCallback(logConfigCallback, toks.get(1));
+                    } else {
+                        LOG.error("{} Unknown callback for subtree {}", new RuntimeException("Unknown callback for this path"), path);
```
abhishekagarwal87: can you also log that exit is being called due to the error.

storm-core/src/jvm/org/apache/storm/cluster/StormZkClusterState.java
```diff
@@ -0,0 +1,683 @@
...
+        if (callback != null) {
+            stormBaseCallback.put(stormId, callback);
+        }
+        return Cluster.maybeDeserialize(clusterState.get_data(Cluster.stormPath(stormId), callback != null), StormBase.class);
+    }
+
+    @Override
+    public ClusterWorkerHeartbeat getWorkerHeartbeat(String stormId, String node, Long port) {
+        byte[] bytes = clusterState.get_worker_hb(Cluster.workerbeatPath(stormId, node, port), false);
+        if (bytes != null) {
```
abhishekagarwal87: this if block is not needed

storm-core/src/clj/org/apache/storm/daemon/nimbus.clj
```diff
@@ -174,11 +173,11 @@
      :authorization-handler (mk-authorization-handler (conf NIMBUS-AUTHORIZER) conf)
      :impersonation-authorization-handler (mk-authorization-handler (conf NIMBUS-IMPERSONATION-AUTHORIZER) conf)
      :submitted-count (atom 0)
-     :storm-cluster-state (cluster/mk-storm-cluster-state conf :acls (when
+     :storm-cluster-state (StormZkClusterState. conf  (when
                                                                        (Utils/isZkAuthenticationConfiguredStormServer
                                                                          conf)
                                                                        NIMBUS-ZK-ACLS)
```
knusbaum: Please correct the indentation on these lines.

storm-core/src/clj/org/apache/storm/daemon/nimbus.clj
```diff
@@ -275,11 +274,11 @@
...
 (defn do-rebalance [nimbus storm-id status storm-base]
   (let [rebalance-options (:topology-action-options storm-base)]
-    (.update-storm! (:storm-cluster-state nimbus)
+    (.updateStorm (:storm-cluster-state nimbus)
       storm-id
-        (-> {:topology-action-options nil}
+      (thriftify-storm-base (-> {:topology-action-options nil}
           (assoc-non-nil :component->executors (:component->executors rebalance-options))
-          (assoc-non-nil :num-workers (:num-workers rebalance-options)))))
+          (assoc-non-nil :num-workers (:num-workers rebalance-options))))))
```
knusbaum: Can we put `thriftify-storm-base` at the end of the `->` to follow suit here?

storm-core/src/jvm/org/apache/storm/cluster/StormZkClusterState.java
```diff
@@ -0,0 +1,683 @@
...
+            ErrorInfo errorInfo = new ErrorInfo(error, Time.currentTimeSecs());
+            errorInfo.set_host(node);
+            errorInfo.set_port(port.intValue());
+            byte[] serData = Utils.serialize(errorInfo);
+            clusterState.mkdirs(path, acls);
+            clusterState.create_sequential(path + Cluster.ZK_SEPERATOR + "e", serData, acls);
+            clusterState.set_data(lastErrorPath, serData, acls);
+            List<String> childrens = clusterState.get_children(path, false);
+
+            Collections.sort(childrens);
```
abhishekagarwal87: this will sort e1, e2, e12 to e1, e12, e2 while you want it to be e1, e2, e12.

hustfxj:
@revans2 @abhishekagarwal87  Thank you for some advice.

hustfxj:
@revans2 @abhishekagarwal87  I will close this PR to update the codes according to your comments.

PR #1072, opened 2016-02-03
STORM-676 Trident implementation for sliding and tumbling windows.

satishd:
This PR adds windowing support in a trident stream.

- Added support of sliding and tumbling windows based on count of tuples and duration through window operations on stream.
- This functionality uses a WindowsStore implementation which stores tuples, and aggregated results.
- HBase store provider is implemented. It can be enhanced with user configuration or extending the basic implementation. Other providers can be added by implementing the required interfaces.
- More details can be found at

    - https://issues.apache.org/jira/browse/STORM-676
    - https://issues.apache.org/jira/secure/attachment/12780751/StormTrident_windowing_support-676.pdf


satishd:
Upmerged, resolved and rebased.

examples/storm-starter/src/jvm/org/apache/storm/starter/trident/TridentWindowingInmemoryStoreTopology.java
```diff
@@ -0,0 +1,134 @@
...
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+/**
+ *
+ */
+public class TridentWindowingInmemoryStoreTopology {
+    private static final Logger log = LoggerFactory.getLogger(TridentWindowingInmemoryStoreTopology.class);
```
harshach: in storm we are always using capital LOG. can you please make sure this reflected across patch.

examples/storm-starter/src/jvm/org/apache/storm/starter/trident/TridentWindowingInmemoryStoreTopology.java
```diff
@@ -0,0 +1,134 @@
...
+import org.apache.storm.utils.Utils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+/**
```
harshach: nit; please remove empty comments

external/storm-hbase/src/main/java/org/apache/storm/hbase/trident/windowing/HBaseWindowsStore.java
```diff
@@ -0,0 +1,275 @@
...
+public class HBaseWindowsStore implements WindowsStore {
+    private static final Logger log = LoggerFactory.getLogger(HBaseWindowsStore.class);
+    public static final String UTF_8 = "utf-8";
+
+    private final ThreadLocal<HTable> threadLocalHtable;
+    private Queue<HTable> htables = new ConcurrentLinkedQueue<>();
+    private final byte[] family;
+    private final byte[] qualifier;
+
+    public HBaseWindowsStore(final Configuration config, final String tableName, byte[] family, byte[] qualifier) {
```
harshach: this looks like its better have its own config method and take a Map<?,?> config object to configure it.

external/storm-hbase/src/main/java/org/apache/storm/hbase/trident/windowing/HBaseWindowsStore.java
```diff
@@ -0,0 +1,275 @@
...
+        Result[] results = null;
+        try {
+            results = htable().get(gets);
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+
+        Kryo kryo = new Kryo();
+        List<Object> values = new ArrayList<>();
+        for (int i=0; i<results.length; i++) {
```
harshach: nit can you change this to for(int i: results)

external/storm-hbase/src/main/java/org/apache/storm/hbase/trident/windowing/HBaseWindowsStoreFactory.java
```diff
@@ -0,0 +1,51 @@
...
+
+import java.util.Map;
+
+public class HBaseWindowsStoreFactory implements WindowsStoreFactory {
+    private final Map<String, Object> config;
+    private final String tableName;
+    private final byte[] family;
+    private final byte[] qualifier;
+
+    public HBaseWindowsStoreFactory(Map<String, Object> config, String tableName, byte[] family, byte[] qualifier) {
```
harshach: It would be better if we do this via config method than constructor

storm-core/src/jvm/org/apache/storm/trident/operation/builtin/Debug.java
```diff
@@ -40,7 +42,7 @@ public Debug(String name) {
 
     @Override
     public boolean isKeep(TridentTuple tuple) {
-        System.out.println(name + tuple.toString());
+        System.out.println("<"+new Date()+"> "+name + tuple.toString());
```
harshach: shouldn't this be LOG?

storm-core/src/jvm/org/apache/storm/trident/windowing/AbstractTridentWindowManager.java
```diff
@@ -0,0 +1,241 @@
...
+
+    public Queue<TriggerResult> getPendingTriggers() {
+        return pendingTriggers;
+    }
+
+    public void shutdown() {
+        try {
+            log.info("window manager [{}] is being shutdown", windowManager);
+            windowManager.shutdown();
+        } finally {
```
harshach: shouldn't we log an exception why the shutdown didn't happened properly.

storm-core/src/jvm/org/apache/storm/trident/windowing/InMemoryWindowsStoreFactory.java
```diff
@@ -0,0 +1,37 @@
...
+ * under the License.
+ */
+package org.apache.storm.trident.windowing;
+
+/**
+ * InMemoryWindowsStoreFactory contains a single instance of {@code InMemoryWindowsStore} which will be used for
+ * storing tuples and triggers of the window and successfully emitted triggers can be removed from {@code StateUpdater}.
+ *
+ */
+public class InMemoryWindowsStoreFactory implements WindowsStoreFactory {
```
harshach: is this intend to be singleton ?

storm-core/src/jvm/org/apache/storm/trident/windowing/WindowsStore.java
```diff
@@ -0,0 +1,78 @@
...
+/**
+ * Store for storing window related entities like windowed tuples, triggers etc.
+ *
+ */
+public interface WindowsStore extends Serializable {
+
+    /**
+     * This can be used as a separator while generating a key from sequence of strings.
+     */
+    public static final String KEY_SEPARATOR = "|";
```
harshach: Its good to have configure method Map<?, ?> config rather than do it via constructor.

harshach:
Thanks @satishd . Sorry for taking long time to review. I've few minor nits. Can you please address those.

external/storm-hbase/src/main/java/org/apache/storm/hbase/trident/windowing/HBaseWindowsStore.java
satishd: results is an array of Result[] and it can not be reduced to for(int i: results). I did not use for(Result result: results) as I need index to get key as mentioned in the below code. 
@harshach pl let me know if I am missing anything here.

``` java
if(result.isEmpty()) {
                log.error("Got empty result for key [{}]", keys.get(i));
                throw new RuntimeException("Received empty result for key: "+keys.get(i));
            }
```

storm-core/src/jvm/org/apache/storm/trident/operation/builtin/Debug.java
satishd: @harshach If debug logs are always thrown out to system out then we can change this to use log without any backward compatibility issue. I was not sure whether debug logs are always thrown out to system.out and did not change the existing code.

storm-core/src/jvm/org/apache/storm/trident/windowing/AbstractTridentWindowManager.java
satishd: I will add logging an exception in WindowTridentProcessor from which this method is invoked.

storm-core/src/jvm/org/apache/storm/trident/windowing/InMemoryWindowsStoreFactory.java
satishd: @harshach Each  InMemoryWindowsStoreFactory instance will always have the same inmemoryStore instance. The same InMemoryWindowsStoreFactory instance is passed to WindowsStateUpdater which which removes successfully emitted triggers from the same inMemoryWindowsStore instance in WindowsStateUpdater#updateState(WindowsState, List, TridentCollector).

Below code snippet is from Stream#window(WindowConfig windowConfig, WindowsStoreFactory windowStoreFactory, Fields inputFields, Aggregator aggregator,
                          Fields functionFields, boolean storeTuplesInStore)
``` java
// when storeTuplesInStore is false then the given windowStoreFactory is only used to store triggers and
        // that store is passed to WindowStateUpdater to remove them after committing the batch.
        Stream stream = _topology.addSourcedNode(this,
                new ProcessorNode(_topology.getUniqueStreamId(),
                        _name,
                        fields,
                        fields,
                        new WindowTridentProcessor(windowConfig, _topology.getUniqueWindowId(), windowStoreFactory,
                                inputFields, aggregator, storeTuplesInStore)));

        Stream effectiveStream = stream.project(functionFields);

        // create StateUpdater with the given windowStoreFactory to remove triggered aggregation results form store
        // when they are successfully processed.
        StateFactory stateFactory = new WindowsStateFactory();
        StateUpdater stateUpdater = new WindowsStateUpdater(windowStoreFactory);
        stream.partitionPersist(stateFactory, new Fields(WindowTridentProcessor.TRIGGER_FIELD_NAME), stateUpdater, new Fields());
```

Updated the java doc with the below details.

``` java
/**
 * InMemoryWindowsStoreFactory contains a single instance of {@link InMemoryWindowsStore} which will be used for
 * storing tuples and triggers of the window. The same InMemoryWindowsStoreFactory instance is passed to {@link WindowsStateUpdater},
 * which removes successfully emitted triggers from the same {@code inMemoryWindowsStore} instance in
 * {@link WindowsStateUpdater#updateState(WindowsState, List, TridentCollector)}.
 *
 */
public class InMemoryWindowsStoreFactory implements WindowsStoreFactory {

```

examples/storm-starter/src/jvm/org/apache/storm/starter/trident/TridentWindowingInmemoryStoreTopology.java
satishd: Updated with missing java doc.

external/storm-hbase/src/main/java/org/apache/storm/hbase/trident/windowing/HBaseWindowsStoreFactory.java
satishd: @harshach All these fields should be available for windows-store-factory and they are not optional, that is why they are passed with the constructor. We may enclose them in a Configuration class(with immutable fields) and pass it to the constructor like the way it is done for HBaseStateFactory(Options).

external/storm-hbase/src/main/java/org/apache/storm/hbase/trident/windowing/HBaseWindowsStore.java
harshach: @satishd makes sense.

external/storm-hbase/src/main/java/org/apache/storm/hbase/trident/windowing/HBaseWindowsStoreFactory.java
harshach: @satishd isn't table name, cf are specific to Hbase. In that case a config method can take a map and do its validation and this validation will specific to HbaseWindowStoreFactory

storm-core/src/jvm/org/apache/storm/trident/Stream.java
```diff
@@ -565,19 +578,169 @@ public Stream aggregate(Fields inputFields, ReducerAggregator agg, Fields functi
...
+     * Returns a stream of tuples which are aggregated results of a tumbling window with every {@code windowCount} of tuples.
+     *
+     * @param windowCount represents window tuples count
+     * @param inputFields projected fields for aggregator
+     * @param aggregator aggregator to run on the window of tuples to compute the result and emit to the stream.
+     * @param functionFields fields of values to emit with aggregation.
+     *
+     * @return
+     */
+    public Stream tumblingCountWindow(int windowCount, Fields inputFields, Aggregator aggregator, Fields functionFields) {
```
arunmahadevan: The various windowing configurations can be expressed with the `window(WindowConfig windowConfig,...)`. If we are adding wrappers over it, can you please maintain compatibility with the core apis ?

i.e,
```java
tumblingWindow(Count count)
tumblingWindow(Duration duration)
window(Count windowLength, Count slidingInterval)
window(Duration windowLength, Duration slidingInterval)
```

arunmahadevan:
Haven't gone through the code in great detail, but when I ran a simple topology with a tumbling time window of 3 secs,

- If the spout emits a batch and sleeps, the result of the window agregation is never emitted (ideally it should have emitted at 3 secs). 

- If the spout periodically emits batches (one batch every 5 sec), the time when the window output is generated always conincides with the time the spout emits the batches (i.e at 10, 15, 20 secs) whereas one would expect the result to be emitted at 6, 12, 15, 18 etc. Also only one result is emitted at time t = 10 secs, whereas two batches are complete at this time.
 

Can you verify if this is the current behavior and if so fix it?

satishd:
@arunmahadevan What is the batch emit interval in your topology configuration? Each triggered result is emitted as part of the nearest batch completed. Trident runs in batches and the emitted trigger would be part of the latest completed batch. These details are mentioned in the design doc attached in STORM-676. That is the reason it may have emitted results at the nearest completed batches. 
Can you please share your configuration so that I can confirm the behavior?

satishd:
Triggered results piggyback on the incoming tuples to send those results to the downstream. When there are no batch tuples then these triggers would wait. Filed [STORM-1627](https://issues.apache.org/jira/browse/STORM-1627) to enhance this for avoiding piggybacking on incoming tuples.

storm-core/src/jvm/org/apache/storm/trident/Stream.java
satishd: @arunmahadevan Changed API aligned with core windowing API.

harshach:
+1. will wait for @arunmahadevan

arunmahadevan:
@harshach I haven't reviewed the code yet, plan to do that within a couple of days.

examples/storm-starter/pom.xml
```diff
@@ -31,10 +31,13 @@
...
 
   <properties>
-     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-     <!-- see comment below... This fixes an annoyance with intellij -->
-     <provided.scope>provided</provided.scope>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <!-- see comment below... This fixes an annoyance with intellij -->
+    <provided.scope>provided</provided.scope>
+    <hbase.version>0.98.4-hadoop2</hbase.version>
+    <hbase.version>1.1.2</hbase.version>
```
arunmahadevan: hbase.version is defined twice

examples/storm-starter/pom.xml
```diff
@@ -149,6 +152,11 @@
       <artifactId>storm-hdfs</artifactId>
       <version>${project.version}</version>
     </dependency>
+      <dependency>
+          <groupId>org.apache.storm</groupId>
```
arunmahadevan: storm-hbase defines hbase.version, do you still need to define hbase.version above ?

examples/storm-starter/pom.xml
```diff
@@ -177,6 +185,36 @@
       <artifactId>storm-redis</artifactId>
       <version>${project.version}</version>
     </dependency>
+      <dependency>
+          <groupId>org.apache.hbase</groupId>
```
arunmahadevan: why is hbase-server dependency needed ? anyways storm-hbase includes hbase-server dependency so this could be removed.

examples/storm-starter/pom.xml
```diff
@@ -177,6 +185,36 @@
...
+                  <groupId>org.slf4j</groupId>
+                  <artifactId>slf4j-log4j12</artifactId>
+              </exclusion>
+              <exclusion>
+                  <groupId>org.apache.zookeeper</groupId>
+                  <artifactId>zookeeper</artifactId>
+              </exclusion>
+          </exclusions>
+      </dependency>
+      <dependency>
```
arunmahadevan: storm-hbase includes hbase-client dependency so this could be removed.

examples/storm-starter/src/jvm/org/apache/storm/starter/trident/TridentHBaseWindowingStoreTopology.java
```diff
@@ -0,0 +1,122 @@
...
+import org.apache.storm.tuple.Values;
+import org.apache.storm.utils.Utils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ *
```
arunmahadevan: May be explain what this topology does and the various state options (in-memory vs hbase etc)

examples/storm-starter/src/jvm/org/apache/storm/starter/trident/TridentHBaseWindowingStoreTopology.java
```diff
@@ -0,0 +1,122 @@
...
+                new Values("the man went to the store and bought some candy"), new Values("four score and seven years ago"),
+                new Values("how many apples can you eat"), new Values("to be or not to be the person"));
+        spout.setCycle(true);
+
+        TridentTopology topology = new TridentTopology();
+
+        Stream stream = topology.newStream("spout1", spout).parallelismHint(16).each(new Fields("sentence"),
+                new Split(), new Fields("word"))
+                .window(TumblingCountWindow.of(1000), windowsStore, new Fields("word"), new CountAsAggregator(), new Fields("count"))
+//                .tumblingTimeWindow(new BaseWindowedBolt.Duration(3, TimeUnit.SECONDS), windowsStore, new Fields("word"), new CountAsAggregator(), new Fields("count"))
```
arunmahadevan: Remove comment. If you want to illustrate time window its better to write a separate example.

examples/storm-starter/src/jvm/org/apache/storm/starter/trident/TridentHBaseWindowingStoreTopology.java
```diff
@@ -0,0 +1,122 @@
...
+                .each(new Fields("count"), new Debug())
+                .each(new Fields("count"), new Echo(), new Fields("ct"));
```
arunmahadevan: Why both echo and debug ? You could use the `peek` if you want to just print the tuples.

examples/storm-starter/src/jvm/org/apache/storm/starter/trident/TridentHBaseWindowingStoreTopology.java
```diff
@@ -0,0 +1,122 @@
...
+
+        return topology.build();
+    }
+
+    public static class Split extends BaseFunction {
```
arunmahadevan: Why not reuse org.apache.storm.trident.testing.Split ?

examples/storm-starter/src/jvm/org/apache/storm/starter/trident/TridentHBaseWindowingStoreTopology.java
```diff
@@ -0,0 +1,122 @@
...
+        @Override
+        public void cleanup() {
+
+        }
+    }
+
+    public static void main(String[] args) throws Exception {
+        Config conf = new Config();
+        conf.setMaxSpoutPending(20);
+        conf.put(Config.TOPOLOGY_TRIDENT_WINDOWING_INMEMORY_CACHE_LIMIT, 2);
```
arunmahadevan: Is this a reasonable value? May be explain how should users decide what value to provide?

examples/storm-starter/src/jvm/org/apache/storm/starter/trident/TridentHBaseWindowingStoreTopology.java
```diff
@@ -0,0 +1,122 @@
...
+
+        // window-state table should already be created with cf:tuples column
```
arunmahadevan: Is it a pre-req for using HBaseWindowsStoreFactory that the window-state table should be created with cf:tuples ? Is it documented somewhere ? Its better if this can be auto-created.

examples/storm-starter/src/jvm/org/apache/storm/starter/trident/TridentWindowingInmemoryStoreTopology.java
```diff
@@ -0,0 +1,133 @@
...
+
+        TridentTopology topology = new TridentTopology();
+
+        Stream stream = topology.newStream("spout1", spout).parallelismHint(16).each(new Fields("sentence"),
+                new Split(), new Fields("word"))
+                .window(windowConfig, windowStore, new Fields("word"), new CountAsAggregator(), new Fields("count"))
+//                .aggregate(new CountAsAggregator(), new Fields("count"))
+                .each(new Fields("count"), new Debug())
+                .each(new Fields("count"), new Echo(), new Fields("ct"))
+                .each(new Fields("ct"), new Debug());
```
arunmahadevan: Why two echo and one debug? To see the values isnt it enough to just print it once ? You could also use the peek api here.

examples/storm-starter/src/jvm/org/apache/storm/starter/trident/TridentWindowingInmemoryStoreTopology.java
```diff
@@ -0,0 +1,133 @@
...
+
+        return topology.build();
+    }
+
+    public static class Split extends BaseFunction {
```
arunmahadevan: Why not reuse org.apache.storm.trident.testing.Split ?

external/storm-hbase/src/main/java/org/apache/storm/hbase/trident/windowing/HBaseWindowsStore.java
```diff
@@ -0,0 +1,275 @@
...
+    @Override
+    public void put(String key, Object value) {
+        WindowsStore.Entry.nonNullCheckForKey(key);
+        WindowsStore.Entry.nonNullCheckForValue(value);
+
+        if(value == null) {
+            throw new IllegalArgumentException("Invalid value of null with key: "+key);
+        }
+        Put put = new Put(effectiveKey(key));
+        Kryo kryo = new Kryo();
```
arunmahadevan: This creates a new Kryo and Output instance for every invocation which may not be necessary. Also there is no option to register classes so kryo cannot optimize the serialization/deserialization. Its better if you have a serializer interface and provide a default implementation with ability to register classes. You could reuse `org.apache.storm.state.Serializer` and `org.apache.storm.state.DefaultStateSerializer`

storm-core/src/jvm/org/apache/storm/trident/Stream.java
```diff
@@ -6,9 +6,9 @@
  * to you under the Apache License, Version 2.0 (the
  * "License"); you may not use this file except in compliance
  * with the License.  You may obtain a copy of the License at
- *
+ * <p/>
```
arunmahadevan: This is not part of the license, you might want to remove this.

storm-core/src/jvm/org/apache/storm/trident/Stream.java
```diff
@@ -565,19 +578,155 @@ public Stream aggregate(Fields inputFields, ReducerAggregator agg, Fields functi
...
+                                      Fields inputFields, Aggregator aggregator, Fields functionFields) {
+        return window(TumblingCountWindow.of(windowCount), windowStoreFactory, inputFields, aggregator, functionFields);
+    }
+
+    /**
+     * Returns a stream of tuples which are aggregated results of a sliding window with every {@code windowCount} of tuples
+     * and slides the window with {@code slideCount}.
+     *
+     * @param windowCount represents tuples count of a window
+     * @param slideCount represents sliding count window
```
arunmahadevan: Make it in sync with core api - "@param slidingInterval the number of tuples after which the window slides"

storm-core/src/jvm/org/apache/storm/trident/Stream.java
```diff
@@ -565,19 +578,155 @@ public Stream aggregate(Fields inputFields, ReducerAggregator agg, Fields functi
...
+                                     Fields inputFields, Aggregator aggregator, Fields functionFields) {
+        return window(TumblingDurationWindow.of(windowDuration), windowStoreFactory, inputFields, aggregator, functionFields);
+    }
+
+    /**
+     * Returns a stream of tuples which are aggregated results of a window which slides at duration of {@code slideDuration}
+     * and completes a window at {@code windowDuration}
+     *
+     * @param windowDuration represents window duration configuration
+     * @param slideDuration represents sliding duration  configuration
```
arunmahadevan: Make it in sync with core api - "@param slidingInterval the time duration after which the window slides"

storm-core/src/jvm/org/apache/storm/trident/operation/builtin/Debug.java
arunmahadevan: Instead of modifying Debug you could use the peek api where you can specify custom actions.

examples/storm-starter/pom.xml
satishd: This is storm-hbase module version but not hbase.version

examples/storm-starter/src/jvm/org/apache/storm/starter/trident/TridentHBaseWindowingStoreTopology.java
satishd: HBaseWindowsStoreFactory can be configured with whatever column-family and column-name used for storing tuples in this window. So, it is expected that user would have already created them. I do not think we should try to create them as it may require some other credentials for administration etc.

``` java
        HBaseWindowsStoreFactory windowStoreFactory = new HBaseWindowsStoreFactory(new HashMap<String, Object>(), "window-state", "cf".getBytes("UTF-8"), "tuples".getBytes("UTF-8"));
```

examples/storm-starter/src/jvm/org/apache/storm/starter/trident/TridentWindowingInmemoryStoreTopology.java
satishd: Peek API was not there when this code was written. I will use peek API here. :)

storm-core/src/jvm/org/apache/storm/trident/Stream.java
satishd: I mentioned count here as it directly conveys that it is count. I would change javadoc as you suggested to be in sync with core API.

storm-core/src/jvm/org/apache/storm/trident/Stream.java
satishd: I will add that.

external/storm-hbase/src/main/java/org/apache/storm/hbase/trident/windowing/HBaseWindowsStore.java
satishd: Earier, I tried with reusing Kryo instance in these calls but I encountered an issue and fall back to this approach. Filed [STORM-1649](https://issues.apache.org/jira/browse/STORM-1649) to look into this.

external/storm-hbase/src/main/java/org/apache/storm/hbase/trident/windowing/HBaseWindowsStore.java
```diff
@@ -0,0 +1,275 @@
...
+/**
+ * This class stores entries into hbase instance of the given configuration.
+ *
+ */
+public class HBaseWindowsStore implements WindowsStore {
+    private static final Logger LOG = LoggerFactory.getLogger(HBaseWindowsStore.class);
+    public static final String UTF_8 = "utf-8";
+
+    private final ThreadLocal<HTable> threadLocalHtable;
+    private Queue<HTable> htables = new ConcurrentLinkedQueue<>();
```
arunmahadevan: could be final ?

external/storm-hbase/src/main/java/org/apache/storm/hbase/trident/windowing/HBaseWindowsStoreFactory.java
```diff
@@ -0,0 +1,54 @@
...
+
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.hbase.HBaseConfiguration;
+import org.apache.storm.trident.windowing.WindowsStore;
+import org.apache.storm.trident.windowing.WindowsStoreFactory;
+
+import java.util.Map;
+
+/**
+ *
```
arunmahadevan: Add some comments or remove this.

storm-core/src/jvm/org/apache/storm/trident/Stream.java
```diff
@@ -594,19 +607,155 @@ public Stream aggregate(Fields inputFields, ReducerAggregator agg, Fields functi
                 .aggregate(inputFields, agg, functionFields)
                 .chainEnd();
     }
-    
+
+    /**
+     * Returns a stream of tuples which are aggregated results of a tumbling window with every {@code windowCount} of tuples.
+     *
+     * @param windowCount represents no of tuples in the window
```
arunmahadevan: no -> number

storm-core/src/jvm/org/apache/storm/trident/Stream.java
```diff
@@ -594,19 +607,155 @@ public Stream aggregate(Fields inputFields, ReducerAggregator agg, Fields functi
...
+     * @param windowStoreFactory intermediary tuple store for storing windowing tuples
+     * @param inputFields projected fields for aggregator
+     * @param aggregator aggregator to run on the window of tuples to compute the result and emit to the stream.
+     * @param functionFields fields of values to emit with aggregation.
+     *
+     * @return
```
arunmahadevan: `@return the new Stream`

storm-core/src/jvm/org/apache/storm/trident/Stream.java
```diff
@@ -594,19 +607,155 @@ public Stream aggregate(Fields inputFields, ReducerAggregator agg, Fields functi
...
+     */
+    public Stream tumblingWindow(int windowCount, WindowsStoreFactory windowStoreFactory,
+                                      Fields inputFields, Aggregator aggregator, Fields functionFields) {
+        return window(TumblingCountWindow.of(windowCount), windowStoreFactory, inputFields, aggregator, functionFields);
+    }
+
+    /**
+     * Returns a stream of tuples which are aggregated results of a sliding window with every {@code windowCount} of tuples
+     * and slides the window with {@code slideCount}.
```
arunmahadevan: slides the window after slideCount

storm-core/src/jvm/org/apache/storm/trident/Stream.java
```diff
@@ -594,19 +607,155 @@ public Stream aggregate(Fields inputFields, ReducerAggregator agg, Fields functi
...
+     *
+     * @param windowCount represents tuples count of a window
+     * @param slideCount the number of tuples after which the window slides
+     * @param windowStoreFactory intermediary tuple store for storing windowing tuples
+     * @param inputFields projected fields for aggregator
+     * @param aggregator aggregator to run on the window of tuples to compute the result and emit to the stream.
+     * @param functionFields fields of values to emit with aggregation.
+     *
+     * @return
```
arunmahadevan: `@return the new Stream`

storm-core/src/jvm/org/apache/storm/trident/Stream.java
```diff
@@ -594,19 +607,155 @@ public Stream aggregate(Fields inputFields, ReducerAggregator agg, Fields functi
...
+     */
+    public Stream slidingWindow(int windowCount, int slideCount, WindowsStoreFactory windowStoreFactory,
+                                     Fields inputFields, Aggregator aggregator, Fields functionFields) {
+        return window(SlidingCountWindow.of(windowCount, slideCount), windowStoreFactory, inputFields, aggregator, functionFields);
+    }
+
+    /**
+     * Returns a stream of tuples which are aggregated results of a window tumbles at duration of {@code windowDuration}
```
arunmahadevan: of a window that tumbles

storm-core/src/jvm/org/apache/storm/trident/Stream.java
```diff
@@ -594,19 +607,155 @@ public Stream aggregate(Fields inputFields, ReducerAggregator agg, Fields functi
...
+     *
+     * @param windowDuration represents tumbling window duration configuration
+     * @param windowStoreFactory intermediary tuple store for storing windowing tuples
+     * @param inputFields projected fields for aggregator
+     * @param aggregator aggregator to run on the window of tuples to compute the result and emit to the stream.
+     * @param functionFields fields of values to emit with aggregation.
+     *
+     * @return
```
arunmahadevan: `@return the new Stream`

storm-core/src/jvm/org/apache/storm/trident/Stream.java
```diff
@@ -594,19 +607,155 @@ public Stream aggregate(Fields inputFields, ReducerAggregator agg, Fields functi
...
+     */
+    public Stream tumblingWindow(BaseWindowedBolt.Duration windowDuration, WindowsStoreFactory windowStoreFactory,
+                                     Fields inputFields, Aggregator aggregator, Fields functionFields) {
+        return window(TumblingDurationWindow.of(windowDuration), windowStoreFactory, inputFields, aggregator, functionFields);
+    }
+
+    /**
+     * Returns a stream of tuples which are aggregated results of a window which slides at duration of {@code slideDuration}
```
arunmahadevan: `@code slidingInterval`

storm-core/src/jvm/org/apache/storm/trident/Stream.java
```diff
@@ -594,19 +607,155 @@ public Stream aggregate(Fields inputFields, ReducerAggregator agg, Fields functi
...
+     * and completes a window at {@code windowDuration}
+     *
+     * @param windowDuration represents window duration configuration
+     * @param slidingInterval the time duration after which the window slides
+     * @param windowStoreFactory intermediary tuple store for storing windowing tuples
+     * @param inputFields projected fields for aggregator
+     * @param aggregator aggregator to run on the window of tuples to compute the result and emit to the stream.
+     * @param functionFields fields of values to emit with aggregation.
+     *
+     * @return
```
arunmahadevan: `@return the new Stream`

storm-core/src/jvm/org/apache/storm/trident/Stream.java
```diff
@@ -594,19 +607,155 @@ public Stream aggregate(Fields inputFields, ReducerAggregator agg, Fields functi
...
+    }
+
+    /**
+     * Returns a stream of aggregated results based on the given window configuration which uses inmemory windowing tuple store.
+     *
+     * @param windowConfig window configuration like window length and slide length.
+     * @param inputFields input fields
+     * @param aggregator aggregator to run on the window of tuples to compute the result and emit to the stream.
+     * @param functionFields fields of values to emit with aggregation.
+     * @return
```
arunmahadevan: `@return the new Stream`

storm-core/src/jvm/org/apache/storm/trident/Stream.java
```diff
@@ -594,19 +607,155 @@ public Stream aggregate(Fields inputFields, ReducerAggregator agg, Fields functi
...
+
+    /**
+     * Returns stream of aggregated results based on the given window configuration.
+     *
+     * @param windowConfig window configuration like window length and slide length.
+     * @param windowStoreFactory intermediary tuple store for storing tuples for windowing
+     * @param inputFields input fields
+     * @param aggregator aggregator to run on the window of tuples to compute the result and emit to the stream.
+     * @param functionFields fields of values to emit with aggregation.
+     * @return
```
arunmahadevan: `@return the new Stream`

storm-core/src/jvm/org/apache/storm/trident/operation/builtin/Debug.java
```diff
@@ -40,7 +42,7 @@ public Debug(String name) {
 
     @Override
     public boolean isKeep(TridentTuple tuple) {
```
arunmahadevan: Since you replaced Debug with peek, these changes can be reverted.

storm-core/src/jvm/org/apache/storm/trident/windowing/WindowTridentProcessor.java
```diff
@@ -0,0 +1,265 @@
...
+ */
+public class WindowTridentProcessor implements TridentProcessor {
+    private static final Logger LOG = LoggerFactory.getLogger(WindowTridentProcessor.class);
+
+    public static final String TRIGGER_INPROCESS_PREFIX = "tip" + WindowsStore.KEY_SEPARATOR;
+    public static final String TRIGGER_PREFIX = "tr" + WindowsStore.KEY_SEPARATOR;
+    public static final String TRIGGER_COUNT_PREFIX = "tc" + WindowsStore.KEY_SEPARATOR;
+
+    public static final String TRIGGER_FIELD_NAME = "_task_info";
+    public static final long DEFAULT_INMEMORY_TUPLE_CACHE_LIMIT = 100l;
```
arunmahadevan: use capital 'L', with small 'l' it appears like 1001

storm-core/src/jvm/org/apache/storm/trident/windowing/config/WindowConfig.java
```diff
@@ -0,0 +1,55 @@
...
+    /**
+     * Gives the type of windowing. It can be any of {@code Type} values.
+     *
+     * @return
+     */
+    public Type getWindowType();
+
+    public void validate();
+
+    public enum Type {
```
arunmahadevan: The subclasses already capture the type of the WindowConfig so this enum can be avoided. Instead of switching on this type later to return the window strategy, you can add a method here that returns the window strategy and have the subclasses of WindowConfig return the appropriate WindowStrategy.

storm-core/src/jvm/org/apache/storm/trident/windowing/strategy/WindowStrategyFactory.java
```diff
@@ -0,0 +1,60 @@
...
+    /**
+     * Creates a {@code WindowStrategy} instance based on the given {@code windowConfig}.
+     *
+     * @param windowConfig
+     * @return
+     */
+    public static <T> WindowStrategy<T> create(WindowConfig windowConfig) {
+        WindowStrategy<T> windowStrategy = null;
+        WindowConfig.Type windowType = windowConfig.getWindowType();
+        switch(windowType) {
```
arunmahadevan: instead of switching, you can add a getWindowStrategy method in WindowConfig and invoke windowConfig.getWindowStrategy(). That way you don't need to maintain these enum constants and add new constants if new window configs are added in future.

arunmahadevan:
1. The event time based windows are not exposed via trident, can you file a follow up JIRA ?
2. Trident api docs needs to be updated with windowing apis.
3. Would be good to add the implementation details and content from the design doc pdf into a separate .md file under docs.

I will do a final pass after the comments are addressed.

storm-core/src/jvm/org/apache/storm/trident/operation/builtin/Debug.java
satishd: I would like to keep these changes as it is useful to see the timestamp in Debug output. I do not think we need a separate PR for this minor change.

satishd:
- [STORM-1651](https://issues.apache.org/jira/browse/STORM-1651) is created for adding event time based support in trident windowing.
- [STORM-1652](https://issues.apache.org/jira/browse/STORM-1652) is created for adding api/design docs. Design doc content will be retrieved from the doc attached to [STORM-676](https://issues.apache.org/jira/browse/STORM-676) JIRA.

external/storm-hbase/src/main/java/org/apache/storm/hbase/trident/windowing/HBaseWindowsStore.java
ptgoetz: Have you looked at the `ThreadLocal` approach described in the documentation?

https://github.com/EsotericSoftware/kryo/blob/master/README.md#pooling-kryo-instances

The Kryo constructor is fairly heavy and should be limited if possible.

satishd: @ptgoetz I have threadlocals of htable instances. Similarly, I tried with Kryo but got into some issues. It seems DefaultStateSerializer follows very similar approach as KryoValuesSerializer.  Have [STORM-1649](https://issues.apache.org/jira/browse/STORM-1649) for this issue and I will address it there.

harshach:
@arunmahadevan looks like all comments are addressed. Can you please take a look.

arunmahadevan:
+1 for merging this. If we are including this in 1.0, its better to update the trident api docs as well in 1.0.

harshach:
@satishd can you open a PR against 1.x-branch as well.

harshach:
@satishd 
"[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.1:compile (default-compile) on project storm-hbase: Compilation failure: Compilation failure:
[ERROR] /Users/harsha/code/apache/storm/external/storm-hbase/src/main/java/org/apache/storm/hbase/trident/windowing/HBaseWindowsStore.java:[210,25] unreported exception java.io.IOException; must be caught or declared to be thrown
[ERROR] /Users/harsha/code/apache/storm/external/storm-hbase/src/main/java/org/apache/storm/hbase/trident/windowing/HBaseWindowsStore.java:[229,25] unreported exception java.io.IOException; must be caught or declared to be thrown
[ERROR] -> [Help 1]"

satishd:
@harshach Upmerged and refactored HBase put APIs with recent hbase version upgrade in the master branch.

satishd:
@harshach Raised https://github.com/apache/storm/pull/1263 on 1.x-branch.

harshach:
Great work @satishd . Merged into 1.x and master

PR #1073, opened 2016-02-03
[STORM-1523] util.clj available-port conversion to java

redsanket:
Please note this is a small change when compared to a bigger util.clj to java PR.
It is being put first as the work needed to convert the entire file is taken care and will be up soon.
Just did not want to depend on other patches so putting this up first.

### DIFF
--- a/storm-core/src/jvm/org/apache/storm/utils/Utils.java
+++ b/storm-core/src/jvm/org/apache/storm/utils/Utils.java
@@ -100,6 +100,7 @@ import java.net.InetAddress;
 import java.net.URL;
 import java.net.URLDecoder;
 import java.net.UnknownHostException;
+import java.net.ServerSocket;
 import java.nio.ByteBuffer;
 import java.nio.file.FileSystems;
 import java.nio.file.Files;
@@ -1499,6 +1500,22 @@ public class Utils {
         } else {
             return new RuntimeException(e);
         }
+    }
+
+    public static int getAvailablePort(int prefferedPort) {
+        int localPort = -1;
+        try(ServerSocket socket = new ServerSocket(prefferedPort)) {
+            localPort = socket.getLocalPort();
+        } catch(IOException exp) {
+            if (prefferedPort > 0) {
+                return getAvailablePort(0);
+            }
+        }
+        return localPort;
+    }
+
+    public static int getAvailablePort() {
+        return getAvailablePort(0);
     }
 
     /**
--- a/storm-core/src/clj/org/apache/storm/util.clj
+++ b/storm-core/src/clj/org/apache/storm/util.clj
@@ -143,17 +143,6 @@
                true (throw ~error-local)
                )))))
 
-(letfn [(try-port [port]
-                  (with-open [socket (java.net.ServerSocket. port)]
-                    (.getLocalPort socket)))]
-  (defn available-port
-    ([] (try-port 0))
-    ([preferred]
-     (try
-       (try-port preferred)
-       (catch java.io.IOException e
-         (available-port))))))
-
 (defn clojurify-structure
   [s]
   (prewalk (fn [x]
--- a/storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
+++ b/storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
@@ -25,7 +25,15 @@ import org.apache.commons.io.FileUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.io.*;
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.io.Reader;
 import java.lang.reflect.Field;
 import java.util.ArrayList;
 import java.util.HashMap;
--- a/storm-core/test/clj/org/apache/storm/messaging/netty_unit_test.clj
+++ b/storm-core/test/clj/org/apache/storm/messaging/netty_unit_test.clj
@@ -21,7 +21,7 @@
   (:use [org.apache.storm.daemon.worker :only [is-connection-ready]])
   (:import [java.util ArrayList]))
 
-(def port (available-port))
+(def port (Utils/getAvailablePort))
 (def task 1)
 
 ;; In a "real" cluster (or an integration test), Storm itself would ensure that a topology's workers would only be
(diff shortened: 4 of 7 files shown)


storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1374,5 +1393,29 @@ public static RuntimeException wrapInRuntime(Exception e){
...
     }
+
+    public static int getAvailablePort(Integer port) {
+        Integer localPort = null;
+        ServerSocket socket = null;
+        try {
+            socket = new ServerSocket(port);
+            localPort = socket.getLocalPort();
+        } catch (IOException exp) {
+            getAvailablePort(0);
```
abhishekagarwal87: localPort = getAvailablePort(0);

abhishekagarwal87: also this should be done only if the input port is non-zero. similar to what is happening in clojure. since there is no control on recursion depth here.

redsanket: Nice catch thanks, I will do that.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1374,5 +1393,29 @@ public static RuntimeException wrapInRuntime(Exception e){
...
+        } finally {
```
revans2: Can we use the java 7 closable to make this code a little simpler?  And why is port an Integer and not an int?

```
public static int getAvailablePort(int port) throws IOException {
    try (ServerSocket socket = new ServerSocket(port)) {
        return socket.getLocalPort();
    } catch (IOException e) {
        if (port <= 0) {
            throw e;
        }
        return getAvailablePort(0);
    }
}
```

revans2: If it is an int not an Integer we don't have to wrap all of the calls to it with `(Integer. )`

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1374,5 +1393,29 @@ public static RuntimeException wrapInRuntime(Exception e){
...
+                }
+            } catch (IOException exp) {
+                LOG.error("Socket Exception", exp);
+            }
+        }
+        return localPort;
+    }
+
+    public static int getAvailablePort() {
+            return getAvailablePort(0);
```
revans2: Indentation looks off here.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
redsanket: I just thought when we are converting the tests to java auto boxing will help to remove the specific conversion, I just wanted to return a null rather than just returning. Moreover, i might have to wrap all calls for clojure in int as the default is long.

redsanket: I will change it to int

redsanket:
@revans2 @abhishekagarwal87 fixed the issues I presume. Let me know if you have any concerns

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1374,5 +1393,21 @@ public static RuntimeException wrapInRuntime(Exception e){
             return new RuntimeException(e);
         }
     }
+
+    public static int getAvailablePort(int port) {
```
abhishekagarwal87: will be good to rename the port to preferredPort since it only makes a best effort guarantee.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1374,5 +1393,21 @@ public static RuntimeException wrapInRuntime(Exception e){
...
+        int localPort = -1;
+        try(ServerSocket socket = new ServerSocket(port)) {
+            localPort = socket.getLocalPort();
+        } catch(IOException exp) {
+            if (port > 0) {
+                getAvailablePort(0);
```
abhishekagarwal87: return getAvailablePort(0)

redsanket:
@abhishekagarwal87 @knusbaum please take a look at it, it must go into Utils.java

redsanket:
Just a moment, will test it once more

knusbaum:
+1

revans2:
+1 lets squash

redsanket:
@revans squashed

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1446,6 +1448,22 @@ public static RuntimeException wrapInRuntime(Exception e){
...
+    public static int getAvailablePort(int prefferedPort) {
+        int localPort = -1;
+        try(ServerSocket socket = new ServerSocket(prefferedPort)) {
+            localPort = socket.getLocalPort();
+        } catch(IOException exp) {
+            if (prefferedPort > 0) {
+                return getAvailablePort(0);
+            }
+        }
+        return localPort;
```
abhishekagarwal87: one minor comment. previous version never returns -1. can we just bubble up the exception in catch block if preferredPort is not greater than zero and not return a negative port? Other than that +1.

knusbaum:
@abhishekagarwal87 Good catch. I doubt it has any practical effects, but it's good to be consistent.

knusbaum:
Although, if we want to be consistent with the original code, removing the `if` and the final `return` statement is the solution.

abhishekagarwal87:
If there is an exception, tests are going to fail nonetheless. If the exception is bubbled up, test failure will show the right error. Though yes I agree, that in practice binding to port zero is not likely to result in an exception.

knusbaum:
Trying with value 0 on failure was a potentially important behavior of the previous implementation.

Exceptions will still bubble up if it fails if my solution is implemented.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
redsanket: socket.getLocalPort() returns -1 if no port is found as per the internal implementation. No harm in assigning localPort with -1

/**
     * Returns the port number on which this socket is listening.
     * <p>
     * If the socket was bound prior to being {@link #close closed},
     * then this method will continue to return the port number
     * after the socket is closed.
     *
     * @return  the port number to which this socket is listening or
     *          -1 if the socket is not bound yet.
     */
    public int getLocalPort() {
        if (!isBound())
            return -1;
        try {
            return getImpl().getLocalPort();
        } catch (SocketException e) {
            // nothing
            // If we're bound, the impl has been created
            // so we shouldn't get here
        }
        return -1;
    }

redsanket:
The original implementation would return -1 in any case if the port is not bound yet. If we not changing the final outcome and unless we are concerned about throwing an exception when the port is not bound, I guess it would be fine. Moreover, removing if condition would keep trying to find the port recursively until it finds it I presume. The original code only tries once as per my understanding.

abhishekagarwal87:
Yes. Original code tries only once and if condition should definitely be kept. It's fine with me.

redsanket:
@revans2 @knusbaum @abhishekagarwal87 still +1 ? Hope it is languishing too long out here.

revans2:
Yes still +1

abhishekagarwal87:
+1

harshach:
@redsanket can you upmerge.

redsanket:
@harshach upmerged and squashed

PR #1074, opened 2016-02-03
STORM-1226: Port backtype.storm.util to java

knusbaum:
This PR converts nearly the entirety of util.clj into Java.

What's left in util.clj will never be translated, but will disappear when the clojure code using it is rewritten in java. For now it's cleaner to leave these few functions and macros in place instead of trying to inline them everywhere.

The only operation that should need to be performed on util.clj after this PR is merged is `git rm`


d2r:
Test failures appear to be unrelated.

I am +1, however I also worked on much of this.

It would also be good to squash this before it is merged.

knusbaum:
Yep, test failures are unrelated it seems.

@d2r I will squash this if it's very important, but since there are many merges in here, it's going to be a big job to do, and needs to be done **before** review to avoid accidentally messing up the merges while squashing.

knusbaum:
Squashed the commits.
Contributors were @rfarivar, @d2r, and myself.

Test failures are unrelated as far as I can tell. There appear to be test problems affecting all PRs.

storm-core/pom.xml
```diff
@@ -42,6 +42,10 @@
             This is here as a work around to place it at the beginning of the classpath
             even though maven does not officially support ordering of the classpath.-->
         <dependency>
+            <groupId>uk.org.lidalia</groupId>
+            <artifactId>sysout-over-slf4j</artifactId>
+        </dependency>
```
revans2: The new dependency seems OK, and it is unlikely that we will encounter some topology already using this, but could we please shade it anyways?

storm-core/src/clj/org/apache/storm/cluster.clj
```diff
@@ -274,7 +281,9 @@
                          LOGCONFIG-ROOT (issue-map-callback! log-config-callback (first args))
                          BACKPRESSURE-ROOT (issue-map-callback! backpressure-callback (first args))
                          ;; this should never happen
-                         (exit-process! 30 "Unknown callback for subtree " subtree args)))))]
+                         ;(exit-process! 30 "Unknown callback for subtree " subtree args)
```
revans2: Please remove the commented out code.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1469,974 @@ public static int toPositive(int number) {
...
+    public static String containerFilePath (String dir) {
+        return dir + FILE_PATH_SEPARATOR + "launch_container.sh";
+    }
+
+    public static void throwRuntime (Object... strings) {
+        String combinedErrorMessage = "";
+        for (Object oneMessage: strings) {
+            combinedErrorMessage = combinedErrorMessage + oneMessage.toString();
+        }
+        throw new RuntimeException(combinedErrorMessage);
```
revans2: String concatenation is broken here if a nil/null is passed in.  Can at least switch to using StringBuilder, and I would really prefer to just see this inlined everywhere.  Is it really that much simpler to do 

```
(Utils/throwRuntime [a b c])
```

over

```
(throw (RuntimeException. (str a b c))
```

storm-core/src/clj/org/apache/storm/daemon/executor.clj
```diff
@@ -262,13 +266,16 @@
...
-                             ((:report-error <>) error)
-                             (if (or
-                                    (exception-cause? InterruptedException error)
-                                    (exception-cause? java.io.InterruptedIOException error))
-                               (log-message "Got interrupted excpetion shutting thread down...")
-                               ((:suicide-fn <>))))
+     :report-error-and-die (reify
+                             Thread$UncaughtExceptionHandler
+                             (uncaughtException [this _ error]
+                               (fn [error]
```
revans2: uncaughtException returns a void, so this function we are creating is thrown away, and never called.

storm-core/src/clj/org/apache/storm/daemon/nimbus.clj
```diff
@@ -643,10 +650,13 @@
...
-         reverse-map
+         (Utils/reverseMap)
+         clojurify-structure
          (map-val sort)
-         (join-maps component->executors)
-         (map-val (partial apply partition-fixed))
+         ((fn [ & maps ] (Utils/joinMaps (into-array (into [component->executors] maps)))))
+         (clojurify-structure)
+         (map-val (partial apply (fn part-fixed [a b] (Utils/partitionFixed a b))))
+         ((fn [whatever] (log-message (pr-str "after-partition-fixed: " whatever)) whatever))
```
revans2: I think this debugging line should be removed.

storm-core/src/clj/org/apache/storm/daemon/executor.clj
knusbaum: Good catch. Fixing.

storm-core/src/clj/org/apache/storm/command/dev_zookeeper.clj
```diff
@@ -14,6 +14,7 @@
 ;; See the License for the specific language governing permissions and
 ;; limitations under the License.
 (ns org.apache.storm.command.dev-zookeeper
+  (:import [org.apache.storm.utils Utils])
   (:use [org.apache.storm zookeeper util config])
```
abhishekagarwal87: would it better to remove util from this list where ever possible?

storm-core/src/clj/org/apache/storm/converter.clj
```diff
@@ -71,6 +72,8 @@
                                                           (:worker->resources assignment)))))
     thrift-assignment))
 
+;TODO: when translating this function, you should replace the map-val with a proper for loop HERE
+;TODO: when translating this function, you should replace the map-key with a proper for loop HERE
```
abhishekagarwal87: duplicate comment.

storm-core/src/clj/org/apache/storm/converter.clj
```diff
@@ -203,6 +214,8 @@
       (convert-to-symbol-from-status (.get_prev_status storm-base))
       (map-val clojurify-debugoptions (.get_component_debug storm-base)))))
 
+;TODO: when translating this function, you should replace the map-val with a proper for loop HERE
+;TODO: when translating this function, you should replace the map-val with a proper for loop HERE
```
abhishekagarwal87: todos are same.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -106,8 +158,23 @@
...
-            return c.newInstance();
+            LOG.info("Creating new instance for class {}", klass);
+            return newInstance(Class.forName(klass));
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    public static Object newInstance(Class klass) {
+        LOG.info("Inside other newInstance static method.");
```
abhishekagarwal87: this log statement can be changed.

storm-core/src/jvm/org/apache/storm/serialization/SerializationFactory.java
```diff
@@ -141,7 +141,8 @@ public IdDictionary(StormTopology topology) {
                 ComponentCommon common = Utils.getComponentCommon(topology, name);
                 List<String> streams = new ArrayList<>(common.get_streams().keySet());
                 streamNametoId.put(name, idify(streams));
-                streamIdToName.put(name, Utils.reverseMap(streamNametoId.get(name)));
+                //TODO: Can the call to simpleReverseMap be replaced wih Utils.reverseMap ?
```
revans2: Don't know either delete this message or replace the call.

storm-core/src/jvm/org/apache/storm/utils/IPredicate.java
```diff
@@ -15,20 +15,8 @@
...
-public class TestUtils extends Utils {
-
-    public static void testSetupBuilder(CuratorFrameworkFactory.Builder
-            builder, String zkStr, Map conf, ZookeeperAuthInfo auth)
-    {
-        setupBuilder(builder, zkStr, conf, auth);
-    }
-
+public interface IPredicate {
+    Boolean test (Object obj);
```
revans2: Why return a big B Boolean and not a boolean?  When is null something we want to support in a predicate, and what does it even mean to return null?  Please either add javadocs or change it.

storm-core/src/jvm/org/apache/storm/utils/StaticMockable.java
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.storm.utils;
+
+public interface StaticMockable extends AutoCloseable {
```
revans2: Can we have some javadocs to explain what this is doing?  and why?

storm-core/src/jvm/org/apache/storm/utils/Time.java
```diff
@@ -98,10 +104,25 @@ public static long currentTimeMillis() {
             return System.currentTimeMillis();
         }
     }
-    
+
+    public static long toMillis (int secs) {
```
revans2: This is not obvious that it takes seconds.  Can we rename this as sesToMillis?

storm-core/src/jvm/org/apache/storm/utils/Time.java
```diff
@@ -98,10 +104,25 @@ public static long currentTimeMillis() {
...
+        return 1000*(long) secs;
+    }
+    public static long toMillis (String secs) {
```
revans2: This is even less obvious that it is seconds to millis,  and who is calling this anyways?

storm-core/src/jvm/org/apache/storm/utils/Time.java
```diff
@@ -98,10 +104,25 @@ public static long currentTimeMillis() {
...
+        return 1000*Long.parseLong(secs);
+    }
+
     public static int currentTimeSecs() {
         return (int) (currentTimeMillis() / 1000);
     }
+
+    public static int delta(int timeInSeconds) {
```
revans2: Can we rename this to deltaSecs, because it is not consistent if times here are seconds or milliseconds.

storm-core/src/jvm/org/apache/storm/utils/Time.java
```diff
@@ -98,10 +104,25 @@ public static long currentTimeMillis() {
...
+        return Time.currentTimeSecs() - timeInSeconds;
+    }
+
+    public static long deltaMs(long timeInMilliseconds) {
```
revans2: Most other places in this file uses Millis not Ms, or milliseconds is considered the default.  Could we follow and rename this deltaMillis, or rename all of the others to use Ms instead of Millis?

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1469,974 @@ public static int toPositive(int number) {
...
+                          requestId, remoteAddress, principal, operation));
+    }
+
+    /**
+     * Make sure a given key name is valid for the storm config. 
+     * Throw RuntimeException if the key isn't valid.
+     * @param name The name of the config key to check.
+     */
+    public static void validateKeyName(String name) {
+        Set<String> disallowedKeys = new HashSet<>();
```
abhishekagarwal87: This can be declared and initialized statically.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -106,8 +158,23 @@
 
     public static Object newInstance(String klass) {
         try {
-            Class c = Class.forName(klass);
-            return c.newInstance();
+            LOG.info("Creating new instance for class {}", klass);
```
revans2: Please remove or make this debug

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -106,8 +158,23 @@
...
+            return newInstance(Class.forName(klass));
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    public static Object newInstance(Class klass) {
```
revans2: Can we make this generic?

```
public static <T> T newInstance(Class<T> klass) {
```

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -106,8 +158,23 @@
...
+        LOG.info("Inside other newInstance static method.");
+        return _instance.newInstanceImpl(klass);
+    }
+
+    // Non-static impl methods exist for mocking purposes.
+    public Object newInstanceImpl(Class klass) {
+        try {
+            LOG.info("Returning {}.newInstance()", klass);
```
revans2: Please remove or make debug

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -106,8 +158,23 @@
...
+        }
+    }
+
+    public static Object newInstance(Class klass) {
+        LOG.info("Inside other newInstance static method.");
+        return _instance.newInstanceImpl(klass);
+    }
+
+    // Non-static impl methods exist for mocking purposes.
+    public Object newInstanceImpl(Class klass) {
```
revans2: This should be generic too, ideally.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -580,20 +655,23 @@ public static void restrictPermissions(String baseDir) {
...
+        return (clojure.lang.IFn) RT.var(namespace, name).deref();
     }
 
     public static boolean isSystemId(String id) {
         return id.startsWith("__");
     }
 
-    public static <K, V> Map<V, K> reverseMap(Map<K, V> map) {
+    /*
+        TODO: Can this be replaced with reverseMap in this file?
```
revans2: Does not look like it can.  The return types are completely different.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1469,974 @@ public static int toPositive(int number) {
...
+            return "/";
+        }
+        Vector<String> tokens = tokenizePath(path);
+        int length = tokens.size();
+        if (length == 0) {
+            return "/";
+        }
+        String output = "";
+        for (int i = 0; i < length - 1; i++) {  //length - 1 to mimic "butlast" from the old clojure code
+            output = output + "/" + tokens.get(i);
```
abhishekagarwal87: parent path for ab/xyz is ab but here /ab will be returned. is it intended?

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1469,974 @@ public static int toPositive(int number) {
...
+    }
+
+    /**
+     * Extra dir from the jar to destdir
+     *
+     * @param jarpath
+     * @param dir
+     * @param destdir
+     *
+    (with-open [jarpath (ZipFile. jarpath)]
```
abhishekagarwal87: this should be removed

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1469,974 @@ public static int toPositive(int number) {
...
+     * Determines if a zip archive contains a particular directory.
+     *
+     * @param zipfile path to the zipped file
+     * @param target directory being looked for in the zip.
+     * @return boolean whether or not the directory exists in the zip.
+     */
+    public static boolean zipDoesContainDir(String zipfile, String target) throws IOException {
+        List<ZipEntry> entries = (List<ZipEntry>)Collections.list(new ZipFile(zipfile).entries());
+
+        if(entries == null) {
```
revans2: When will this ever be null?

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1469,974 @@ public static int toPositive(int number) {
...
+                }
+            }
+        } catch (IOException e) {
+            LOG.info("Could not extract {} from {}", dir, jarpath);
+        } finally {
+            if (jarFile != null) {
+                try {
+                    jarFile.close();
+                } catch (IOException e) {
+                    throw new RuntimeException(
```
abhishekagarwal87: you should pass parent IO exception in this runtime exception

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1469,974 @@ public static int toPositive(int number) {
...
+        InputStream in = null;
+        try {
+            jarFile = new JarFile(jarpath);
+            Enumeration<JarEntry> jarEnums = jarFile.entries();
+            while (jarEnums.hasMoreElements()) {
+                JarEntry entry = jarEnums.nextElement();
+                if (!entry.isDirectory() && entry.getName().startsWith(dir)) {
+                    File aFile = new File(destdir, entry.getName());
+                    aFile.getParentFile().mkdirs();
+                    out = new FileOutputStream(aFile);
```
abhishekagarwal87: you can use try-with-resource for out/in to simplify code

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1469,974 @@ public static int toPositive(int number) {
...
+    public static void throwRuntime (Object... strings) {
+        String combinedErrorMessage = "";
+        for (Object oneMessage: strings) {
+            combinedErrorMessage = combinedErrorMessage + oneMessage.toString();
+        }
+        throw new RuntimeException(combinedErrorMessage);
+    }
+
+    public static Object nullToZero (Object v) {
+        return (v!=null? v : 0);
```
abhishekagarwal87: space around operators is missing in some places.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1469,974 @@ public static int toPositive(int number) {
...
+        }
+        return null;
+    }
+
+    public static String logsFilename(String stormId, int port) {
+        return stormId + FILE_PATH_SEPARATOR + Integer.toString(port) + FILE_PATH_SEPARATOR + "worker.log";
+    }
+
+    public static String eventLogsFilename(String stormId, int port) {
+        return stormId + FILE_PATH_SEPARATOR + Integer.toString(port) + FILE_PATH_SEPARATOR + "events.log";
```
revans2: In both of these we don't need to do an Integer.toString.  + will handle doing that for you.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1469,974 @@ public static int toPositive(int number) {
...
+    }
+
+    public static Object readYamlFile(String yamlFile) {
+        try (FileReader reader = new FileReader(yamlFile)) {
+            return new Yaml(new SafeConstructor()).load(reader);
+        }
+        catch(Exception ex) {
```
revans2: We usually keep the catch on the same line as the preceding {

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1469,974 @@ public static int toPositive(int number) {
...
+        return null;
+    }
+
+    public static void setupDefaultUncaughtExceptionHandler() {
+        Thread.setDefaultUncaughtExceptionHandler(new Thread.UncaughtExceptionHandler() {
+                public void uncaughtException(Thread thread, Throwable thrown) {
+                    try {
+                        handleUncaughtException(thrown);
+                    }
+                    catch (Error err) {
```
revans2: Same comment here about formatting.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1469,974 @@ public static int toPositive(int number) {
...
+    public static String workerClasspath() {
+        String stormDir = System.getProperty("storm.home");
+        String stormLibDir = Paths.get(stormDir, "lib").toString();
+        String stormConfDir =
+                System.getenv("STORM_CONF_DIR") != null ?
+                System.getenv("STORM_CONF_DIR") :
+                Paths.get(stormDir, "conf").toString();
+        String stormExtlibDir = Paths.get(stormDir, "extlib").toString();
+        String extcp = System.getenv("STORM_EXT_CLASSPATH");
+        if (stormDir == null) {
```
abhishekagarwal87: this statement should be at the top after stormDir is assigned a value

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1469,974 @@ public static int toPositive(int number) {
...
+    }
+
+    /**
+     * Creates a new map with a string value in the map replaced with an 
+     * equivalently-lengthed string of '#'.
+     * @param m The map that a value will be redacted from
+     * @param key The key pointing to the value to be redacted
+     * @return a new map with the value redacted. The original map will not be modified.
+     */
+    public static Map redactValue(Map<Object, String> m, Object key) {
```
revans2: I think we want to preserve the type in the return, but the passed in type is actually wrong, the value does not have to be a String, it can be anything.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1469,974 @@ public static int toPositive(int number) {
...
+            String value = newMap.get(key);
+            String redacted = new String(new char[value.length()]).replace("\0", "#");
+            newMap.put(key, redacted);
+            return newMap;
+        }
+        return m;
+    }
+
+    public static void logThriftAccess(Integer requestId, InetAddress remoteAddress, Principal principal, String operation) {
+        new ThriftAccessLogger().log(
```
revans2: OK So now I see that we had a really bad error in ThriftAccessLogger.  Lets just move this method completely over to ThriftAccessLogger, and if someone else is using the log method there at least make that method static.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1469,974 @@ public static int toPositive(int number) {
...
+     * Make sure a given key name is valid for the storm config. 
+     * Throw RuntimeException if the key isn't valid.
+     * @param name The name of the config key to check.
+     */
+    public static void validateKeyName(String name) {
+        Set<String> disallowedKeys = new HashSet<>();
+        disallowedKeys.add("/");
+        disallowedKeys.add(".");
+        disallowedKeys.add(":");
+        disallowedKeys.add("\\");
```
revans2: The previous version had this as a static Set.  It would be nice to do that here too, but this is minor.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1469,974 @@ public static int toPositive(int number) {
...
+            if( name.contains(key) ) {
+                throw new RuntimeException("Key name cannot contain any of the following: " + disallowedKeys.toString());
+            }
+        }
+        if(name.trim().isEmpty()) {
+            throw new RuntimeException("Key name cannot be blank");
+        }
+    }
+
+    //Everything from here on is translated from the old util.clj (storm-core/src/clj/backtype.storm/util.clj)
```
revans2: Please remove this comment, the code above was also translated.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1469,974 @@ public static int toPositive(int number) {
...
+
+    public static final boolean IS_ON_WINDOWS = "Windows_NT".equals(System.getenv("OS"));
+
+    public static final String FILE_PATH_SEPARATOR = System.getProperty("file.separator");
+
+    public static final String CLASS_PATH_SEPARATOR = System.getProperty("path.separator");
+
+    public static final int SIGKILL = 9;
+    public static final int SIGTERM = 15;
```
revans2: Could we move the static final values up to the top of the file, this is typical.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1469,974 @@ public static int toPositive(int number) {
...
+     * Find the first item of coll for which pred.test(...) returns true.
+     * @param pred The IPredicate to test for
+     * @param coll The Collection of items to search through.
+     * @return The first matching value in coll, or null if nothing matches.
+     */
+    public static Object findFirst (IPredicate pred, Collection coll) {
+        if (coll == null || pred == null) {
+            return null;
+        } else {
+            Iterator<Object> iter = coll.iterator();
```
revans2: Lets not use the iterator directly, and in the old code if pred was null it would NPE.

```
public static <T> T findFirst (IPredicate pred, Iterable<T> coll) {
  if (coll != null) {
    for (T obj: coll) {
      if (pred.test(obj)) {
        return obj;
      }
    }
  }
  return null;
}
```

storm-core/src/clj/org/apache/storm/command/dev_zookeeper.clj
knusbaum: Will do.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1469,974 @@ public static int toPositive(int number) {
...
+            }
+            return null;
+        }
+    }
+
+    public static Object findFirst (IPredicate pred, Map map) {
+        if (map == null || pred == null) {
+            return null;
+        } else {
+            Iterator<Object> iter = map.entrySet().iterator();
```
revans2: Here too using Iterator directly just makes things more complicated.

```
public static <K, V> Map.Entry<K, V> findFirst (IPredicate pred, Map<K, V> map) {
  if (map == null) {
    return null;
  }
  return findFirst(pred, map.entrySet());
}
```

Oh and it would be nice to have IPredicate be a generic as well for the Parameter passed in.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1469,974 @@ public static int toPositive(int number) {
...
+        if (conf == null) {
+            return memoizedLocalHostname();
+        }
+        Object hostnameString = conf.get(Config.STORM_LOCAL_HOSTNAME);
+        if (hostnameString == null ) {
+            return memoizedLocalHostname();
+        }
+        if (hostnameString.equals("")) {
+            return memoizedLocalHostname();
+        }
```
revans2: Can we combine the two ifs together?

```
if (hostnameString == null || hostnameString.equals("")) {
```
```

File: storm-core/src/jvm/org/apache/storm/utils/Utils.java [hunk 41]
```diff
@@ -1370,9 +1469,974 @@ public static int toPositive(int number) {
...
+        return hostnameString.toString();
```
revans2: The actual function casts this to a string, not calling toString on it.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1469,974 @@ public static int toPositive(int number) {
...
+    }
+
+    public static String uuid() {
+        return UUID.randomUUID().toString();
+    }
+
+    public static long secsToMillisLong(double secs) {
```
revans2: This feels like Time is the better place for this function, and we probably want to eventually call it secsToMillis like the other functions there, but I am not sure how clojure will like it, and we we will have to do something with the type casting.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1469,974 @@ public static int toPositive(int number) {
...
+        return (long) (1000 * secs);
+    }
+
+    public static Vector<String> tokenizePath (String path) {
```
revans2: There is already a version of this in Zookeeper.java.  Lets remove this and update everywhere to use the Zookeeper version (which is the right place for it).

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1469,974 @@ public static int toPositive(int number) {
...
+        }
+        for (String tok: tokens) {
+            if (!tok.isEmpty()) {
+                outputs.add(tok);
+            }
+        }
+        return outputs;
+    }
+
+    public static String parentPath(String path) {
```
revans2: Lets move this to Zookeeper.java instead.  This is Zookeeper specific.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1469,974 @@ public static int toPositive(int number) {
...
+        if (path == null) {
+            return "/";
+        }
```
revans2: This would NPE in the previous code.  I would rather see an error instead of trying to fix it in this way.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1469,974 @@ public static int toPositive(int number) {
...
+        Vector<String> tokens = tokenizePath(path);
+        int length = tokens.size();
+        if (length == 0) {
+            return "/";
+        }
+        String output = "";
+        for (int i = 0; i < length - 1; i++) {  //length - 1 to mimic "butlast" from the old clojure code
+            output = output + "/" + tokens.get(i);
+        }
+        return output;
```
revans2: Can we look at using com.google.common.base.Joiner; from guava or at least a StringBuilder instead here?

storm-core/src/jvm/org/apache/storm/utils/Utils.java
knusbaum: Yep. I've made a note of this in the code and moved simpleReverseMap next to the other reverseMaps.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1469,974 @@ public static int toPositive(int number) {
...
+            return "/";
+        }
+
+        String output = "";
+        for (int i = 0; i < toks.size(); i++) {
+            output = output + "/" + toks.get(i);
+        }
+        return output;
+    }
+    public static String normalizePath (String path) {
```
revans2: There are versions of this and toksToPath already in Zookeeper.java we should just use those instead.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1469,974 @@ public static int toPositive(int number) {
...
+        return toksToPath(tokenizePath(path));
+    }
+
+    public static void exitProcess (int val, Object... msg) {
+        StringBuilder errorMessage = new StringBuilder();
+        errorMessage.append("halting process: ");
```
revans2: very minor nit the original had a capitol H "Halting process:"

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1469,974 @@ public static int toPositive(int number) {
...
+        Runtime.getRuntime().exit(val);
+    }
+
+    public static Object defaulted(Object val, Object defaultObj) {
+        if (val != null) {
+            return val;
+        } else {
+            return defaultObj;
+        }
+    }
```
revans2: This is so simple we really should just inline this.

storm-core/pom.xml
knusbaum: Done.

storm-core/src/clj/org/apache/storm/command/dev_zookeeper.clj
knusbaum: Done.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
knusbaum: Yes. That's how the clojure version worked/

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1469,974 @@ public static int toPositive(int number) {
...
+        return rtn;
+    }
+
+
+    /**
+     * Gets the pid of this JVM, because Java doesn't provide a real way to do this.
+     *
+     * @return
+     */
+    public static String processPid() throws RuntimeException {
```
revans2: RuntimeExceptions we don't need to declare that they are thrown.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1469,974 @@ public static int toPositive(int number) {
...
+            list.add(key);
+        }
+        return rtn;
+    }
+
+
+    /**
+     * Gets the pid of this JVM, because Java doesn't provide a real way to do this.
+     *
+     * @return
```
revans2: Probably should just be 

```
@return the pid of this JVM process. Hacky because Java doesn't provide a real way to do this.
```

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1469,974 @@ public static int toPositive(int number) {
...
+    public static String processPid() throws RuntimeException {
+        String name = ManagementFactory.getRuntimeMXBean().getName();
+        String[] split = name.split("@");
+        if (split.length != 2) {
+            throw new RuntimeException("Got unexpected process name: " + name);
+        }
+        return split[0];
+    }
+
+    public static int execCommand(String command) throws ExecuteException, IOException {
```
revans2: Please please file a JIRA to make this be `String... command` instead.  We should not be splitting things on space, especially in some places where we join on space to send to this so we can split on space again.  Come on why are we doing this???

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1469,974 @@ public static int toPositive(int number) {
...
+     * @param jarpath
+     * @param dir
+     * @param destdir
+     *
+    (with-open [jarpath (ZipFile. jarpath)]
+    (let [entries (enumeration-seq (.entries jarpath))]
+    (doseq [file (filter (fn [entry](and (not (.isDirectory entry)) (.startsWith (.getName entry) dir))) entries)]
+    (.mkdirs (.getParentFile (File. destdir (.getName file))))
+    (with-open [out (FileOutputStream. (File. destdir (.getName file)))]
+    (io/copy (.getInputStream jarpath file) out)))))
```
revans2: Please remove the clojure code from here.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1469,974 @@ public static int toPositive(int number) {
...
+            }
+            if (in != null) {
+                try {
+                    in.close();
+                } catch (IOException e) {
+                    throw new RuntimeException(
+                            "Something really strange happened when trying to close the input for jar file" + jarpath);
+                }
+            }
+        }
```
revans2: The try/catch everything can be made a lot simpler with the java7 AutoClosable and make it match the clojure code better.

```
    public static void extractDirFromJar(String jarpath, String dir, String destdir) {
        try (JarFile jarFile = new JarFile(jarpath);) {
            Enumeration<JarEntry> jarEnums = jarFile.entries();
            while (jarEnums.hasMoreElements()) {
                JarEntry entry = jarEnums.nextElement();
                if (!entry.isDirectory() && entry.getName().startsWith(dir)) {
                    File aFile = new File(destdir, entry.getName());
                    aFile.getParentFile().mkdirs();
                    try(
                        out = new FileOutputStream(aFile);
                        in = jarFile.getInputStream(entry); ) {
                        IOUtils.copy(in, out);
                    }
                }
            }
        } catch (IOException e) {
            LOG.info("Could not extract {} from {}", dir, jarpath);
        }
    }
```

storm-core/src/jvm/org/apache/storm/utils/Utils.java
revans2: Or better yet lets just do it now.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1469,974 @@ public static int toPositive(int number) {
...
+                killString = "kill -" + signum + " ";
+            }
+            killString = killString + pid;
+            retval = execCommand(killString);
+        } catch (ExecuteException e) {
+            LOG.info("Error when trying to kill " + pid + ". Process is probably already dead.");
+        } catch (IOException e) {
+            LOG.info("IOException Error when trying to kill " + pid + ".");
+        } finally {
+            return retval;
```
revans2: The original code would throw an IOException, but catch and ExecuteException.  We should follow the same pattern for now.

Also lets just remove retval.  The ExecuteException should already be handling checking for errors, so returning things is just going to cause confusion.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1469,974 @@ public static int toPositive(int number) {
...
+        }
+    }
+
+    public static int forceKillProcess (long pid) {
+        return sendSignalToProcess(pid, SIGKILL);
+    }
+
+    public static int forceKillProcess (String pid) {
```
revans2: When is this ever a string?  We only call it from one place.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1469,974 @@ public static int toPositive(int number) {
...
+        return sendSignalToProcess(Long.parseLong(pid), SIGKILL);
+    }
+
+    public static int killProcessWithSigTerm (long pid) {
+        return sendSignalToProcess(pid, SIGTERM);
+    }
+    public static int killProcessWithSigTerm (String pid) {
```
revans2: Here too when is this ever a String?

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1469,974 @@ public static int toPositive(int number) {
...
+        return sendSignalToProcess(Long.parseLong(pid), SIGTERM);
+    }
+
+    /*
+        Adds the user supplied function as a shutdown hook for cleanup.
+        Also adds a function that sleeps for a second and then sends kill -9
+        to process to avoid any zombie process in case cleanup function hangs.
+     */
```
revans2: Lets make this a java doc.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1469,974 @@ public static int toPositive(int number) {
...
+    public static void addShutdownHookWithForceKillIn1Sec (Runnable func) {
+        Runnable sleepKill = new Runnable() {
+            @Override
+            public void run() {
+                try {
+                    Time.sleepSecs(1);
+                    Runtime.getRuntime().halt(20);
+                } catch (Exception e) {
+                    LOG.warn("Exception in the ShutDownHook: " + e);
```
revans2: Lets use a , not a + here, so we can get the entire stack trace for the exception.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1469,974 @@ public static int toPositive(int number) {
...
+        Runtime.getRuntime().addShutdownHook(new Thread(sleepKill));
+    }
+
+    /**
+     * Returns the combined string, escaped for posix shell.
+     * @param command the list of strings to be combined
+     * @return the resulting command string
+     */
+    public static String shellCmd (List<String> command) {
+        List<String> changedCommands = new LinkedList<>();
```
revans2: Can we make this an ArrayList and pre allocate the same number of arguments as command?  Very minor nit.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1469,974 @@ public static int toPositive(int number) {
...
+    }
+
+    public static Object containerGet (Container container) {
+        return container.object;
+    }
+
+    public static Container containerSet (Container container, Object obj) {
+        container.object = obj;
+        return container;
+    }
```
revans2: Why the crap do we even have a Container class?  I guess it is OK, but lets move these to be set and get methods on Container itself, and then use the proper type hints.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1469,974 @@ public static int toPositive(int number) {
...
+    /**
+     * Creates a symbolic link to the target
+     * @param dir the parent directory of the link
+     * @param targetDir the parent directory of the link's target
+     * @param targetFilename the file name of the links target
+     * @param filename the file name of the link
+     * @return the path of the link if it did not exist, otherwise null
+     * @throws IOException
+     */
+    public static Path createSymlink(String dir, String targetDir,
```
revans2: The return value appears to be ignored everywhere it is used.  For now it is probably better to remove the return value.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1469,974 @@ public static int toPositive(int number) {
...
+     * @param dir the directory to search
+     * @return the jar file names
+     */
+    private static List<String> getFullJars(String dir) {
+        File[] files = new File(dir).listFiles(new FilenameFilter() {
+            @Override
+            public boolean accept(File dir, String name) {
+                return name.endsWith(".jar");
+            }
+        });
```
revans2: Lets make a static FilenameFilter instead of creating a new instance each time.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
knusbaum: Well, the method is only used on values with type String, and the value type has to have a length() method. I don't know of any interface that describes having a length().

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1469,974 @@ public static int toPositive(int number) {
...
+                    out.write(Utils.shellCmd(
+                            Arrays.asList(
+                                    "export",k+"="+v)));
+                    out.write(";");
+                    out.newLine();
+                }
+            }
+            out.newLine();
+            out.write("exec "+Utils.shellCmd(command)+";");
+        } catch (IOException io) {
```
revans2: The original code did not catch/wrap the I/O Exception.  It just let it bubble up.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1469,974 @@ public static int toPositive(int number) {
...
+     * interval of seconds equal to the return value of the previous call.
+     *
+     * The given afn may be a callable that returns the number of seconds to
+     * sleep, or it may be a Callable that returns another Callable that in turn
+     * returns the number of seconds to sleep. In the latter case isFactory.
+     *
+     * @param afn the code to call on each iteration
+     * @param isDaemon whether the new thread should be a daemon thread
+     * @param eh code to call when afn throws an exception
+     * @param priority the new thread's priority see
```
revans2: see what?

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1469,974 @@ public static int toPositive(int number) {
...
+        });
+        if (eh != null) {
+            thread.setUncaughtExceptionHandler(eh);
+        } else {
+            thread.setUncaughtExceptionHandler(new Thread.UncaughtExceptionHandler() {
+                public void uncaughtException(Thread t, Throwable e) {
+                    Utils.exitProcess(1, "Async loop died!");
+                }
+            });
+        }
```
revans2: I'm not sure how much we care about this, but we are wrapping the exceptions thrown by the original code in a RuntimeException that was not happening before.  If all of the code passes we are probably OK.  I just wanted to point out the difference.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
knusbaum: Yep! already done.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
knusbaum: It's actually **always** a string.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
knusbaum: Do we probably at least want to log stuff here?

revans2: The uncaught exception handlers are not the problem.  The code in them is fine.  The issue is that we are installing them on the thread, that because of the java's Exception Checking a Runnable cannot throw arbitrary exceptions.  As such we are catching all Throwables and things that are not caused by InterruptedException are wrapped in a RuntimeException before being rethrown.  This last part of wrapping the Throwable in a RuntimeException was not happening before.  This means that any code that was looking for a FooExcption would now find a RuntimeExeption with Foo as the caused by.  We usually are fairly good about checking the caused by so we probably don't care, but then again this is error code so it is not tested as well as the other code, so it might cause an issue that we are not aware of.

knusbaum:
@abhishekagarwal87 @revans2 
I believe I've addressed all of the concerns.

storm-core/src/clj/org/apache/storm/daemon/executor.clj
```diff
@@ -262,13 +266,17 @@
...
-                                    (exception-cause? InterruptedException error)
-                                    (exception-cause? java.io.InterruptedIOException error))
-                               (log-message "Got interrupted excpetion shutting thread down...")
-                               ((:suicide-fn <>))))
+     :report-error-and-die (reify
+                             Thread$UncaughtExceptionHandler
+                             (uncaughtException [this _ error]
+                               ((:report-error <>) error)
+                               (when (Utils/exceptionCauseIsInstanceOf ClassCastException error)
+                                 (log-message "CLASS CAST EXCEPTION WOOOOOOOOOOOOOOOO!"))
```
abhishekagarwal87: this is probably leftover from debugging.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1444,821 @@ public static int toPositive(int number) {
...
+    }
+
+    public static void exitProcess (int val, Object... msg) {
+        StringBuilder errorMessage = new StringBuilder();
+        errorMessage.append("Halting process: ");
+        for (Object oneMessage: msg) {
+            errorMessage.append(oneMessage);
+        }
+        String combinedErrorMessage = errorMessage.toString();
+        LOG.error(combinedErrorMessage, new RuntimeException(combinedErrorMessage));
```
abhishekagarwal87: can you also log "exiting the process" alongside error message? It helps a lot in debugging.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1444,821 @@ public static int toPositive(int number) {
...
+        String name = ManagementFactory.getRuntimeMXBean().getName();
+        String[] split = name.split("@");
+        if (split.length != 2) {
+            throw new RuntimeException("Got unexpected process name: " + name);
+        }
+        return split[0];
+    }
+
+    public static int execCommand(String... command) throws ExecuteException, IOException {
+        //String[] cmdlist = command.split(" ");
```
abhishekagarwal87: remove this

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1444,821 @@ public static int toPositive(int number) {
...
+        for (int i = 1; i < command.length; i++) {
+            cmd.addArgument(command[i]);
+        }
+
+        DefaultExecutor exec = new DefaultExecutor();
+        return exec.execute(cmd);
+    }
+
+    /**
+     * Extra dir from the jar to destdir
```
abhishekagarwal87: type. Extract

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1444,821 @@ public static int toPositive(int number) {
...
+        sendSignalToProcess(Long.parseLong(pid), SIGKILL);
+    }
+
+    public static void killProcessWithSigTerm (String pid) throws IOException {
+        sendSignalToProcess(Long.parseLong(pid), SIGTERM);
+    }
+
+    /**
+     * Adds the user supplied function as a shutdown hook for cleanup.
+     * Also adds a function that sleeps for a second and then sends kill -9
```
abhishekagarwal87: It doesn't send kill -9 signal.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
knusbaum: What do you mean? "Halting process: " is already in the error message.

abhishekagarwal87: Ohh. Missed that.

storm-core/src/jvm/org/apache/storm/logging/ThriftAccessLogger.java
```diff
@@ -24,4 +26,10 @@
   public void log(String logMessage) {
     LOG.info(logMessage);
   }
+
+  public static void logAccess(Integer requestId, InetAddress remoteAddress, Principal principal, String operation) {
+    new ThriftAccessLogger().log(
+      String.format("Request ID: {} access from: {} principal: {} operation: {}",
+                    requestId, remoteAddress, principal, operation));
+  }
```
revans2: I thought it would be something like
```
LOG.info("Request ID: {} access from: {} principal: {} operation: {}",
                 requestId, remoteAddress, principal, operation);
```

and ThriftAccessLogger.log would just go away

storm-core/src/jvm/org/apache/storm/utils/Time.java
```diff
@@ -98,10 +104,26 @@ public static long currentTimeMillis() {
...
+
     public static int currentTimeSecs() {
         return (int) (currentTimeMillis() / 1000);
     }
+
+    public static int deltaSecs(int timeInSeconds) {
+        return Time.currentTimeSecs() - timeInSeconds;
+    }
+
+    public static long deltaMs(long timeInMilliseconds) {
```
revans2: This one we need a comment on too indicating that this is from System.currentTimeMillis and not Simulated Time.  Not sure if we can use Simulated Time here or not.  If we can it makes things simpler.

revans2:
One minor nit with the logging and then I am +1

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1444,820 @@ public static int toPositive(int number) {
...
+                if (signum == SIGKILL) {
+                    execCommand("taskkill", "/f", "/pid", pid);
+                } else {
+                    execCommand("taskkill", "/pid", pid);
+                }
+            } else {
+                execCommand("kill", "-" + signum, pid);
+            }
+        } catch (ExecuteException e) {
+            LOG.info("Error when trying to kill " + pid + ". Process is probably already dead.");
```
d2r: Minor, but we could use `{}` format strings like elsewhere.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1444,820 @@ public static int toPositive(int number) {
...
+     *       Which key wins is indeterminate:
+     * "{:a 1  :b 1} -> {1 :a} *or* {1 :b}"
+     */
+    public static <K, V> Map<V, K> simpleReverseMap(Map<K, V> map) {
+        Map<V, K> ret = new HashMap<V, K>();
+        for (Map.Entry<K, V> entry : map.entrySet()) {
+            ret.put(entry.getValue(), entry.getKey());
+        }
+        return ret;
+    }
```
d2r: If we encounter duplicate values in the input `map`, would it be better to throw an exception here rather than allow an indeterminate result?

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1444,820 @@ public static int toPositive(int number) {
...
+            V val = entry.getValue();
+            List<K> list = rtn.get(val);
+            if (list == null) {
+                list = new ArrayList<K>();
+                rtn.put(entry.getValue(), list);
+            }
+            list.add(key);
+        }
+        return rtn;
+    }
```
d2r: Should this return `Map<V, Set<K>>` instead?

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1444,820 @@ public static int toPositive(int number) {
...
+            Object val = listEntry.get(1);
+            List list = rtn.get(val);
+            if (list == null) {
+                list = new ArrayList<Object>();
+                rtn.put(val, list);
+            }
+            list.add(key);
+        }
+        return rtn;
+    }
```
d2r: Similar question.  Should this return a `Map` with key collisions resolved by an unordered `Set`?

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1444,820 @@ public static int toPositive(int number) {
...
+
+    /**
+     * "{:a 1 :b 1 :c 2} -> {1 [:a :b] 2 :c}"
+     *
+     * Example usage in java:
+     *  Map<Integer, String> tasks;
+     *  Map<String, List<Integer>> componentTasks = Utils.reverse_map(tasks);
+     *
+     * @param map
+     * @return
```
d2r: We could either remove the javadoc annotations or add them consistently to the other two methods in this group.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1444,820 @@ public static int toPositive(int number) {
...
+        }
+        return null;
+    }
+
+    public static <T, U> T findFirst (IPredicate<T> pred, Map<U, T> map) {
+        if(map == null) {
+            return null;
+        }
+        return findFirst(pred, (Set<T>)map.entrySet());
+    }
```
d2r: These two `findFirst` functions operate on input that is not necessarily sorted. So they might be better named `findAny` or else changed to accept only sorted input. What was the original code doing here?

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -73,11 +102,15 @@
...
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Set;
 import java.util.TreeMap;
 import java.util.UUID;
+import java.util.Vector;
```
d2r: Remove Vector import

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -86,8 +119,27 @@
 import java.util.zip.GZIPOutputStream;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipFile;
+import java.security.Principal;
+import org.apache.storm.logging.ThriftAccessLogger;
```
d2r: Remove Principal and ThriftAccessLogger imports?

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -884,26 +955,8 @@ private static void unTarUsingJava(File inFile, File untarDir,
...
-            if (c != null) {
-                try {
-                    c.close();
-                } catch (IOException e) {
-                    LOG.debug("Exception in closing " + c, e);
-
-                }
-            }
+            tis.close();
+            inputStream.close();
```
d2r: Do we need null checks here when we call close? I think something could go wrong before they are assigned.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1354,6 +1427,7 @@ public static TopologyInfo getTopologyInfo(String name, String asUser, Map storm
         return topologyInfo;
     }
 
+
```
d2r: blank line

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1370,9 +1444,820 @@ public static int toPositive(int number) {
...
+        }
+    }
+
+    /**
+     * Creates a symbolic link to the target
+     * @param dir the parent directory of the link
+     * @param targetDir the parent directory of the link's target
+     * @param targetFilename the file name of the links target
+     * @param filename the file name of the link
+     * @return the path of the link if it did not exist, otherwise null
```
d2r: recently this was changed not to return anything

d2r:
OK, looked again and noticed some things that might be good to address now.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
revans2: The original code did not even thing about maps but someone used it for maps.

```
(defn find-first
  "Returns the first item of coll for which (pred item) returns logical true.
  Consumes sequences up to the first match, will consume the entire sequence
  and return nil if no match is found."
  [pred coll]
  (first (filter pred coll)))
```

storm-core/src/jvm/org/apache/storm/utils/Utils.java
knusbaum: Maybe in the future. We're not trying to change functionality as part of the translation, though. This is how the Clojure version behaved also.

revans2: This function was just renamed/moved it already existed. See line 596 in the original file.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
revans2: It is a List because that is what the clojure returned.
```
(defn reverse-map
  "{:a 1 :b 1 :c 2} -> {1 [:a :b] 2 :c}"
  [amap]
  (reduce (fn [m [k v]]
            (let [existing (get m v [])]
              (assoc m v (conj existing k))))
          {} amap))
```

storm-core/src/jvm/org/apache/storm/utils/Utils.java
revans2: Order matters here, because the List input is ordered the output is expected to be ordered too.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
revans2: And technically maps can be ordered, they don't have to be, but they can be, like a TreeMap.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
knusbaum: Some callers rely on the value being a sequence type.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
d2r: OK, we have added a comment to the effect of "caller is responsible for order" and that is fine for now.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
d2r: ok, not something we should change now.  Comment about ordering is fine.

d2r:
OK looks good I am +1 again.  I would like other +1s since I contributed some of this.

storm-core/src/jvm/org/apache/storm/utils/Time.java
knusbaum: Modified so it uses `Time.currentTimeSecs()` instead.

PR #1076, opened 2016-02-04
STORM-1522 should create error worker log location only when error-host and error-port are available

satishd:
STORM-1522 should create error worker log location only when error-host and error-port are available

### DIFF
--- a/storm-core/src/clj/org/apache/storm/ui/core.clj
+++ b/storm-core/src/clj/org/apache/storm/ui/core.clj
@@ -148,8 +148,10 @@
   (logviewer-link host (Utils/eventLogsFilename topology-id port) secure?))
 
 (defn worker-log-link [host port topology-id secure?]
-  (let [fname (Utils/logsFilename topology-id port)]
-    (logviewer-link host fname secure?)))
+  (if (or (empty? host) (let [port_str (str port "")] (or (empty? port_str) (= "0" port_str))))
+    ""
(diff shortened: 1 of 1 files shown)


revans2:
The code looks OK to me, but I am a little confused about when would we get an invalid host or port for the error?  The only time I can see is when we are doing a rolling upgrade and there are really old errors still in ZooKeeper.  Or you have not upgraded all of the hosts, but have upgraded nimbus + the UI.

I am +1 for the change, but I still would like to understand it a bit better.

satishd:
when REST API (http://localhost:8080/api/v1/topology/{topology-name}) is used for getting topology details. It throws up spouts/bolts details like below which has wrong 'errorWrokerLogLink'. 

errorWorkerLogLink values are created with the error info available. But the current code is generating an error link even though it does not have host/port info which comes from errors. When there are no errors, host/port values are empty and generating an invalid link.

HeartSaVioR:
+1
It doesn't affect UI (since no error occurs) but API still shows broken link, which let us feel odd.

revans2:
Sounds good +1

harshach:
@satishd can you open this up against 1.x-branch as well.

satishd:
@harshach Raised https://github.com/apache/storm/pull/1116 on 1.x branch .

ptgoetz:
+1

PR #1077, opened 2016-02-04
STORM-1520 [for master branch] Nimbus Clojure/Zookeeper issue ("stateChanged" method not found)

HeartSaVioR:
Please take a look at the mapping between Curator ConnectionState and Storm ConnectionState.

When pull request is merged, I'll craft the pull request which can be applied to 1.x-branch.
(It would be nearly same to this pull request but I've seen small conflict since we ported ZK to Java.)

### DIFF
--- a/storm-core/src/clj/org/apache/storm/cluster_state/zookeeper_state_factory.clj
+++ b/storm-core/src/clj/org/apache/storm/cluster_state/zookeeper_state_factory.clj
@@ -20,6 +20,7 @@
   (:import [org.apache.zookeeper KeeperException$NoNodeException CreateMode
              Watcher$Event$EventType Watcher$Event$KeeperState]
            [org.apache.storm.cluster ClusterState DaemonType])
+  (:import [org.apache.storm.utils StormConnectionStateConverter])
   (:use [org.apache.storm cluster config log util])
   (:require [org.apache.storm [zookeeper :as zk]])
   (:gen-class
@@ -144,7 +145,7 @@
        (let [curator-listener (reify ConnectionStateListener
                                 (stateChanged
                                   [this client newState]
-                                  (.stateChanged listener client newState)))]
+                                  (.stateChanged listener (StormConnectionStateConverter/convert newState))))]
          (Zookeeper/addListener zk-reader curator-listener)))
 
      (sync-path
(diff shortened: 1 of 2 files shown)


revans2:
+1

storm-core/src/clj/org/apache/storm/cluster_state/zookeeper_state_factory.clj
```diff
@@ -144,7 +145,7 @@
        (let [curator-listener (reify ConnectionStateListener
                                 (stateChanged
                                   [this client newState]
-                                  (.stateChanged listener client newState)))]
+                                  (.stateChanged listener (StormConnectionStateConverter/convert newState))))]
```
HeartSaVioR: In this line, listener is ClusterStateListener which was introduced to unify interfaces between Zookeeper and Pacemaker.

Method signature of ClusterStateLister.stateChanged() is 
`void stateChanged(org.apache.storm.cluster.ConnectionState newState)`,
 while ConnectionStateListener.stateChanged() is 
`void stateChanged(org.apache.curator.framework.CuratorFramework client, org.apache.curator.framework.state.ConnectionState newState)`

So it should remove client from parameter, and convert ConnectionState from Curator to ConnectionState from Storm.

HeartSaVioR:
Seems like I didn't explain why this occurs and how to fix. Explained inline.

storm-core/src/clj/org/apache/storm/cluster_state/zookeeper_state_factory.clj
erikdw: Nice explanation!  Helps a lot to have such "whys" explained out, thanks!

HeartSaVioR: Thanks for kind words. :) I hope that Storm will get rid of many kinds of type / signature / typo problems when ported to Java, static typed language.

knusbaum:
+1 Nice clean fix.

HeartSaVioR:
It doesn't make any conflicts with 1.x-branch, so I merged to 1.x-branch without new pull request.
Thanks all for the review!

PR #1078, opened 2016-02-04
[STORM-1368] change heapdump file permissions so that UI download wil…

zhuoliu:
Change heapdump file permissions so that UI download wil work in secure clusters for them.
a. For automatically generated heapdump files by system
b. For heapdump generated by user from UI

### DIFF
--- a/storm-core/src/jvm/org/apache/storm/Config.java
+++ b/storm-core/src/jvm/org/apache/storm/Config.java
@@ -231,7 +231,7 @@ public class Config extends HashMap<String, Object> {
 
     /**
      * Whether we want to display all the resource capacity and scheduled usage on the UI page.
-     * We suggest to have this variable set if you are using any kind of resource-related scheduler.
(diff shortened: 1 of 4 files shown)


storm-core/src/jvm/org/apache/storm/Config.java
```diff
@@ -231,7 +231,7 @@
 
     /**
      * Whether we want to display all the resource capacity and scheduled usage on the UI page.
-     * We suggest to have this variable set if you are using any kind of resource-related scheduler.
+     * You msut have this variable set if you are using any kind of resource-related scheduler.
```
knusbaum: msut

knusbaum:
Minor nit, then +1 pending travis.

revans2:
This breaks windows by default, and is only needed when security is turned on.  Why don't we have the logviewer + worker_launcher change the permissions instead of using the odd, and I would say scary, feature to change the permissions.

PR #1080, opened 2016-02-05
STORM-1526   Improve Storm core performance

roshannaik:
Benchmarking with this simple Speed of Light topology shows that this fix improved throughput by over 60% 
https://github.com/roshannaik/storm-benchmark-sol/blob/master/src/main/java/storm/benchmark/benchmarks/SOL.java 

Call Tree Profile
**Before:**
https://github.com/roshannaik/storm-benchmark-sol/blob/master/profiler/storm%20core%20-%20sol%20-%200%20acker/storm%20core%20-%20call%20tree.png

**After:**
https://github.com/roshannaik/storm-benchmark-sol/blob/master/profiler/storm%20core%20-%20sol%20-%200%20acker/after/storm%20core%20-%20sol%20-%20call%20tree%20-%20after.png


HeartSaVioR:
+1 
Fixed line passes *reflection-on-warning*

harshach:
+1

satishd:
+1

dossett:
@roshannaik Can you update the title of this PR to begin with "STORM-1526" so that it gets associated with the JIRA?  Thanks!

storm-core/src/clj/org/apache/storm/daemon/executor.clj
```diff
@@ -58,7 +58,7 @@
   (.prepare grouping context (GlobalStreamId. component-id stream-id) target-tasks)
   (if (instance? LoadAwareCustomStreamGrouping grouping)
     (fn [task-id ^List values load]
-      (.chooseTasks grouping task-id values load))
+      (.chooseTasks ^LoadAwareCustomStreamGrouping grouping task-id values load))
```
abhishekagarwal87: can you also give the type hint for target-tasks as List? I suspect there is no reflection warning for that, because no method is being invoked on target-tasks. but it is better to follow the practice nonetheless for safeguarding.

roshannaik: @abhishekagarwal87  i see several places in that file which are missing type hints. Might be safer to limit this PR to just this perf issue.

abhishekagarwal87: Sure. looks good to me.

PR #1082, opened 2016-02-05
STORM-1492 With nimbus.seeds set to default, a nimbus for localhost may appear "Offline"

HeartSaVioR:
First of all, AFAIK, it's just an issue of API output, and it doesn't hurt functionality although we don't fix it.
- when determining nimbus addresses from nimbus.seeds, filter out localhost addresses
  - localhost address: 127.0.0.1, localhost, 0:0:0:0:0:0:0:1
- there may be a chance that nimbus (localhost) is not reachable so we can't get info from alive-nimbuses
  - but it means ui process itself can't reach nimbus so we can ignore this case
- there may be another change that user specifies nimbus.seeds to both unreachable localhost address and other reachable nimbuses fqdn
  - ui process can reach other nimbus and localhost itself is lost
  - but I think we can treat it as configuration error

### DIFF
--- a/storm-client/src/jvm/org/apache/storm/utils/Utils.java
+++ b/storm-client/src/jvm/org/apache/storm/utils/Utils.java
@@ -63,6 +63,7 @@ import java.util.regex.Pattern;
 import java.util.zip.GZIPInputStream;
 import java.util.zip.GZIPOutputStream;
 
+import com.google.common.collect.Lists;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.input.ClassLoaderObjectInputStream;
 import org.apache.storm.Config;
@@ -106,6 +107,7 @@ public class Utils {
     public static final String DEFAULT_STREAM_ID = "default";
     private static final Set<Class> defaultAllowedExceptions = new HashSet<>();
     public static final String FILE_PATH_SEPARATOR = System.getProperty("file.separator");
+    private static final List<String> LOCALHOST_ADDRESSES = Lists.newArrayList("localhost", "127.0.0.1", "0:0:0:0:0:0:0:1");
 
     private static ThreadLocal<TSerializer> threadSer = new ThreadLocal<TSerializer>();
     private static ThreadLocal<TDeserializer> threadDes = new ThreadLocal<TDeserializer>();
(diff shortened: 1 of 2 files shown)


HeartSaVioR:
Three builds are failing, which are all intermittent.

* storm-core test failure: nimbus-test could be not run completely
  * https://travis-ci.org/apache/storm/jobs/107172981
* !storm-core test failures: storm-hdfs, OS kills org.apache.storm.hdfs.spout.TestDirLock and org.apache.storm.hdfs.blobstore.BlobStoreTest
  * https://travis-ci.org/apache/storm/jobs/107172983
  * https://travis-ci.org/apache/storm/jobs/107172984

Close and reopen to retrigger.

HeartSaVioR:
Probability of random test failures becomes very high, though I don't know why.
Anyway, I ran test which is same to travis CI but without native profile (it doesn't run on my Mac) and build succeed on storm-core. So test failure is not related to this patch.

Btw, build fails on storm-hdfs, throwing `java.lang.NoClassDefFoundError: org/apache/storm/thrift/TBase` which I don't know why it occurs.

HeartSaVioR:
@ptgoetz @revans2 
Could you please take a look? Or Bobby you want to craft your pull request with your suggestion on issue comment?
Otherwise we may want to define STORM-1492 as 'Won't fix'?

wuchong:
I find that when we set `nimbus.seeds` as ip, such as `nimbus.seeds: [10.0.0.10]`. Then a nimbus for 10.0.0.10 will be 'Offline'.

That's because the nimbus information saved in zk is CanonicalHostName. But the nimbus seeds we set maybe ip. At last, the nimbus with ip will be 'Offline'. 

```
offline-nimbuses (clojure.set/difference nimbus-seeds alive-nimbuses)
```



Maybe we need to change all nimbus-seeds to hostname, before do set/difference.

HeartSaVioR:
@wuchong 
You're right. Actually this PR just tries to address weird behavior when users leave the setting to default.

Btw, there're more to cover if we want to address this to more ideal way.
For example, canonical host name for each machine can be overwritten by setting storm.local.hostname.
Please refer NimbusConf#fromConf for details.

In result, every nimbuses should know what's the return value of NimbusConf#fromConf for other nimbuses, and every nimbuses should match nimbus.seeds and corresponding values by 1 on 1.
I don't have an idea how to do it for now, and I really appreciate ideas which address this.

Otherwise users should describe the same value which NimbusConf#fromConf will return, for nimbus.seeds.

HeartSaVioR:
@revans2 Thanks for reviewing. Just got rebased against current master.

revans2:
Still +1

PR #1084, opened 2016-02-05
STORM-1516 Fixed issue in writing pids with distributed cluster mode.

satishd:
Whenever a topology is submitted, it creates respective workers on supervisor/s. These worker pids are stored as files in ${storm-localdir}/workers/{worker-id}/pids/ on supervisor. But there is an issue in storing worker pids. So, supervisor could not find respective worker pids when a topology is killed. Subsequent topology deployment workers are failed because of earlier workers are still alive and bound to the respective ports. 

Fixed worker.clj to have right checks while writing pids to respective locations.

### DIFF
--- a/storm-core/src/clj/org/apache/storm/daemon/worker.clj
+++ b/storm-core/src/clj/org/apache/storm/daemon/worker.clj
@@ -603,11 +603,11 @@
 (defserverfn mk-worker [conf shared-mq-context storm-id assignment-id port worker-id]
   (log-message "Launching worker for " storm-id " on " assignment-id ":" port " with id " worker-id
                " and conf " conf)
-  (if-not (ConfigUtils/isLocalMode conf)
-    (SysOutOverSLF4J/sendSystemOutAndErrToSLF4J))
   ;; because in local mode, its not a separate
   ;; process. supervisor will register it in this case
-  (when (= :distributed (ConfigUtils/clusterMode conf))
+  ;; if (ConfigUtils/isLocalMode conf) returns false then it is in distributed mode.
+  (when-not (ConfigUtils/isLocalMode conf)
+    (SysOutOverSLF4J/sendSystemOutAndErrToSLF4J)
     (let [pid (Utils/processPid)]
       (FileUtils/touch (ConfigUtils/workerPidPath conf worker-id pid))
       (spit (ConfigUtils/workerArtifactsPidPath conf storm-id port) pid)))


storm-core/src/clj/org/apache/storm/daemon/worker.clj
```diff
@@ -577,11 +577,11 @@
 (defserverfn mk-worker [conf shared-mq-context storm-id assignment-id port worker-id]
   (log-message "Launching worker for " storm-id " on " assignment-id ":" port " with id " worker-id
                " and conf " conf)
-  (if-not (ConfigUtils/isLocalMode conf)
-    (redirect-stdio-to-slf4j!))
   ;; because in local mode, its not a separate
   ;; process. supervisor will register it in this case
-  (when (= :distributed (ConfigUtils/clusterMode conf))
+  ;; if (ConfigUtils/isLocalMode conf) returns false then it is in distributed mode.
+  (when-not (ConfigUtils/isLocalMode conf)
```
abhishekagarwal87: how is this different from checking state == "distributed"

HeartSaVioR: https://issues.apache.org/jira/browse/STORM-1516?focusedCommentId=15134067&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-15134067

Satish already explains why it occurs via JIRA comment.

HeartSaVioR: Easy explanation: 
```
=> (= :dist "dist")
false
```

arunmahadevan:
@satishd Is it happening with all topologies ?

storm-core/src/clj/org/apache/storm/daemon/worker.clj
abhishekagarwal87: Thanks @HeartSaVioR . Got it now.

satishd:
@arunmahadevan Right, this issue happens with all topologies. There is nothing specific about the mentioned topology in JIRA.

arunmahadevan:
+1

revans2:
The code looks fine to me, but for some reason nimbus-test is failing fairly consistently.  Also storm-hdfs is very unstable right now.

revans2:
I just did some poking around and it looks like these failures are likely unrelated.

harshach:
@satishd can you upmerge this and also open another PR for 1.x-branch if we want this there as well. Thanks.

satishd:
@harshach Upmerged and resolved conflicts. We do not need this on 1.x.

harshach:
+1

ndtreviv:
Which version was this fixed in? I'm seeing the same thing in 1.0.1

HeartSaVioR:
@ndtreviv This patch is only for 2.0.0. You might hit [STORM-1934|https://issues.apache.org/jira/browse/STORM-1934] which is fixed for 1.0.2. There're lots of things fixed from 1.0.2 so you're encouraged to give it a try.

ndtreviv:
@HeartSaVioR I'm not sure that's the one. I'm pretty sure that I'm seeing this issue. I can see the supervisor.log saying that it can't find the worker file in workers-users. As a result, the worker processes don't get shut down. It's not a race condition, either, as I've killed the topology and waited for it all to settle before re-deploying, and done this three times over, but the worker processes are still related to the very first topology being run.

HeartSaVioR:
@ndtreviv 
This bug was from ported code which only resides on master (2.0.0) so not relate to 1.x.

What you're explaining seems to [STORM-1879|https://issues.apache.org/jira/browse/STORM-1879] which is occurred after supervisor hits STORM-1934 and deletes worker's' (worker root) directory instead of one of them.

ndtreviv:
@HeartSaVioR Perfect. Thanks

PR #1089, opened 2016-02-08
Added scope to dependencies in storm-elasticsearch pom.xml

shoebamer:


shoebamer:
JIRA Ticket: STORM-1531 - Junit and mockito dependencies need to have correct scope defined in storm-elasticsearch pom.xml

PR #1092, opened 2016-02-09
STORM-1536: Remove Java use of TimeCacheMap

dossett:
This removes java use of TimeCacheMap.  I attempted to replace TimeCacheMap with RotationMap in nimbus.clj but consistently get error messages such as:

```
111654 [heartbeat-timer] ERROR o.a.s.d.worker - Error when processing event
java.lang.RuntimeException: java.io.FileNotFoundException: /var/folders/g6/x6w1f66n16z_pf7qyqq58d_swvmvh5/T/0d68bae5-27b5-46ae-b0ba-40b2f252b98d/workers/2bedd279-f18c-4d5f-9712-dc689f98d8dd/heartbeats/1455040724667 (Too many open files in system)
	at org.apache.storm.utils.LocalState.persistInternal(LocalState.java:178) ~[classes/:?]
	at org.apache.storm.utils.LocalState.put(LocalState.java:143) ~[classes/:?]
	at org.apache.storm.local_state$ls_worker_heartbeat_BANG_.invoke(local_state.clj:122) ~[classes/:?]
	at org.apache.storm.daemon.worker$do_heartbeat.invoke(worker.clj:84) ~[classes/:?]
	at org.apache.storm.daemon.worker$fn__9722$exec_fn__2613__auto__$reify__9724$heartbeat_fn__9725.invoke(worker.clj:606) ~[classes/:?]
	at org.apache.storm.timer$schedule_recurring$this__1305.invoke(timer.clj:105) ~[classes/:?]
	at org.apache.storm.timer$mk_timer$fn__1288$fn__1289.invoke(timer.clj:50) [classes/:?]
	at org.apache.storm.timer$mk_timer$fn__1288.invoke(timer.clj:42) [classes/:?]
	at java.lang.Thread.run(Thread.java:745) [?:1.7.0_75]
```

### DIFF
--- a/examples/storm-starter/src/jvm/org/apache/storm/starter/bolt/SingleJoinBolt.java
+++ b/examples/storm-starter/src/jvm/org/apache/storm/starter/bolt/SingleJoinBolt.java
@@ -25,7 +25,7 @@ import org.apache.storm.topology.OutputFieldsDeclarer;
 import org.apache.storm.topology.base.BaseRichBolt;
 import org.apache.storm.tuple.Fields;
 import org.apache.storm.tuple.Tuple;
-import org.apache.storm.utils.TimeCacheMap;
+import org.apache.storm.utils.RotatingMap;
 
 import java.util.*;
 
@@ -34,7 +34,7 @@ public class SingleJoinBolt extends BaseRichBolt {
   Fields _idFields;
   Fields _outFields;
   int _numSources;
-  TimeCacheMap<List<Object>, Map<GlobalStreamId, Tuple>> _pending;
+  RotatingMap<List<Object>, Map<GlobalStreamId, Tuple>> _pending;
   Map<String, GlobalStreamId> _fieldLocations;
 
   public SingleJoinBolt(Fields outFields) {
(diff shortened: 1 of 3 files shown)


knusbaum:
-1
`RotationMap` is not a drop-in replacement for TimeCacheMap. Unless you take care of rotating the map, it doesn't actually do anything more than a normal map. The documentation in `RotatingMap` is actually incorrect, and I'd support changing it in this PR. 

Why is it exactly that we want to remove `TimeCacheMap`? All it is is a thread-safe, self-rotating wrapper around `RotatingMap`, which looks convenient.

Given it's not anywhere near the critical path, I don't see a problem with having a synchronized data structure.

dossett:
Ah, thanks. TimeCacheMap was marked as deprecated, so this was a learning exercise as much as anything, that would hopefully have some code hygiene benefit.  Much more to be done before it's that apparently, so I'll close the PR.  Thanks again!

knusbaum:
@dossett no problem. And if you wanted to submit a PR altering the incorrect documentation, I'd give it a +1.

dossett:
@knusbaum Just to be clear, you're suggesting that the `@deprecated` annotation and explanation be removed from TimeCacheMap?

knusbaum:
@dossett Yes, that in addition to explaining that it is essentially a wrapper around `RotatingMap`. 

`RotatingMap`'s documentation should also be changed, since it doesn't actually "Expire keys that have not been updated in the configured number of seconds." It has no concept of seconds and needs to be manually rotated.

PR #1093, opened 2016-02-09
STORM-1436: Set Travis Heap size to fit in memory limits.

revans2:


revans2:
There is a separate test on travis that consistently fails on JDK7.  I am looking into it, but it should probably be on a separate pull request.

redsanket:
LGTM +1

dev-tools/travis/travis-script.sh
```diff
@@ -24,8 +24,10 @@ cd ${STORM_SRC_ROOT_DIR}
 
 # We should be concerned that Travis CI could be very slow because it uses VM
 export STORM_TEST_TIMEOUT_MS=150000
+# Travis only has 3GB of memory, lets use 1GB for build, and 1.5GB for forked JVMs
+#export MAVEN_OPTS="-Xmx1024m"
```
HeartSaVioR: Why do you explain with commenting at all?

HeartSaVioR:
+1 though I'd like to see the reason about comments in travis-script.sh.

dev-tools/travis/travis-script.sh
revans2: oops I should not have commented that out.  Looks like it works without it, but I meant to not have it in a comment.  will fix it.

revans2:
I fixed my issue with having the HEAP commented out, and I added in a heap setting to the clojure tests in storm-core.  I suspect that it is the same issue that is causing failures on JDK7

revans2:
On my travis build all of the tests passed, except for one test with kafka that is not related.  It looks like all of the failures were the OOM killer kicking in.

revans2:
The test failure is unrelated.

knusbaum:
+1 Great Jorb!

HeartSaVioR:
Great, thanks @revans2

PR #1094, opened 2016-02-10
[STORM-1532]: Fix readCommandLineOpts to parse JSON correctly in windows

arunmahadevan:
In Windows env, the storm.options are not url-encoded. The parsing logic needs to be fixed to not split in the middle of raw JSON objects.

### DIFF
--- a/storm-core/src/jvm/org/apache/storm/utils/Utils.java
+++ b/storm-core/src/jvm/org/apache/storm/utils/Utils.java
@@ -330,7 +330,17 @@ public class Utils {
         Map ret = new HashMap();
         String commandOptions = System.getProperty("storm.options");
         if (commandOptions != null) {
-            String[] configs = commandOptions.split(",");
+            /*
+             Below regex uses negative lookahead to not split in the middle of json objects '{}'
+             or json arrays '[]'. This is needed to parse valid json object/arrays passed as options
+             via 'storm.cmd' in windows. This is not an issue while using 'storm.py' since it url-encodes
+             the options and the below regex just does a split on the commas that separates each option.
+
+             Note:- This regex handles only valid json strings and could produce invalid results
+             if the options contain un-encoded invalid json or strings with unmatched '[, ], { or }'. We can
+             replace below code with split(",") once 'storm.cmd' is fixed to send url-encoded options.
+              */
+            String[] configs = commandOptions.split(",(?![^\\[\\]{}]*(]|}))");
             for (String config : configs) {
                 config = URLDecoder.decode(config);
                 String[] options = config.split("=", 2);
--- a/bin/storm-config.cmd
+++ b/bin/storm-config.cmd
@@ -86,6 +86,10 @@ if not defined STORM_LOG_DIR (
 @rem retrieve storm.log4j2.conf.dir from conf file
 @rem
 
+if not defined CMD_TEMP_FILE (
+  set CMD_TEMP_FILE=tmpfile
+)
+
 "%JAVA%" -client -Dstorm.options= -Dstorm.conf.file= -cp "%CLASSPATH%" org.apache.storm.command.config_value storm.log4j2.conf.dir > %CMD_TEMP_FILE%
 
 FOR /F "delims=" %%i in (%CMD_TEMP_FILE%) do (
--- a/bin/storm.cmd
+++ b/bin/storm.cmd
@@ -90,7 +90,7 @@
     )
 
     if "%c-opt%"=="second" (
-      set config-options=%config-options%=%1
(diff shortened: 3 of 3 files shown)


storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -330,7 +330,11 @@ public static Map readCommandLineOpts() {
         Map ret = new HashMap();
         String commandOptions = System.getProperty("storm.options");
         if (commandOptions != null) {
-            String[] configs = commandOptions.split(",");
+            /*
+             below regex uses negative lookahead to not split in the middle of
+             json objects '{}' or json arrays '[]'
+              */
+            String[] configs = commandOptions.split(",(?![^\\[\\]{}]*(]|}))");
```
revans2: So first of all this is a really complicated REGEX and even if it worked correctly I would be nervous moving from something simple to something so complex. To break down the REGEX (Just so I am sure I understand what it is doing) `,(?![^\\[\\]{}]*(]|}))` will split on a `,` so long as `(?!` it is not followed by anything that matches `[^\\[\\]{}]*(]|})` which matches any character that is not one of `[`, `]`, `{`, or `}` zero or more times followed by either a `]` or `}` character.

In my testing that does not maintain backwards compatibility with what we had before nor what we currently have.  The python code will URL encode the key/value so no `,` appears in it.  That lets us provide backwards compatibility where I could type odd values that didn't parse to JSON, or other odd things like having a string in the JSON that had JSON like characters in it.  This code does nothing to handle quoting of strings, and in some cases with bad JSON it will split things very differently from before, which is at lest an incompatibility between the windows code and the python code.

On Windows `-c 'foo=["A", "B", "C"' -c bar=B` I forgot to close the array for foo
```
(clojure.string/split "foo=[\"A\", \"B\", \"C\",bar=B" #",(?![^\\[\\]{}]*(]|}))")
["foo=[\"A\"" " \"B\"" " \"C\"" "bar=B"]
```
Because we silently skip things that don't make since to us (in keeping with what happened before but is probably not ideal. we would end up with foo set to `[\"A\"` and bar set to `B`.

They python code works just fine here, but if for some reason I wanted bar to be an array or a map It would combine the two options together.

`-c 'foo=["A", "B", "C"' -c 'bar=["B"]'` I forgot to close the array for foo again

```
(clojure.string/split "foo=[\"A\"%2C \"B\" %2C \"C\",bar=[\"B\"]" #",(?![^\\[\\]{}]*(]|}))")
["foo=[\"A\"%2C \"B\" %2C \"C\",bar=[\"B\"]"]
```

Now they are all just one parameter and foo will be set to a really ugly complex string value, and bar will just disappear.

This code cannot go in as is, and making the REGEX more complicated is not going to solve it.  Not to mention the case where for some reason I may want a value that is URL encoded, and if windows does not URL encode it before sending it over, we will URL decode it, and ...

arunmahadevan: The regex works only for valid JSON strings.

With the current storm.cmd script, valid json strings were not parsed properly so one could not pass a list (for e.g. drpc.servers) from windows. I couldn't figure out a way to url-encode from storm.cmd so I came up with the regex that works for valid JSONs. 

The change should not affect the parsing of the storm.options sent via the python script since the url-encoded values would not contain `[, {, ] or }` and will not be touched by the regex.

Agree this is not the ideal fix that handles all the cases but lets one pass valid json in storm.cmd in windows which is currently broken. If there are any alternatives to get the options working in windows l can try to explore.

revans2: Sorry about the long pause.  I thought I had replied already, but I guess I didn't hit the merge comment button.  I forgot that the python was escaping all of the JSON characters, so like you said this works.

I am OK with merging this in, but I would like an explanation by the REGEX explaining about the escaping in most cases and that this is not perfect for windows.

It would also be good to have a follow on JIRA to solve this correctly with url encoding the parameters, but I don't know enough about windows to suggest how to do that.

arunmahadevan:
@revans2 updated comment to mention the gotchas in windows and raised [JIRA](https://issues.apache.org/jira/browse/STORM-1550) to fix storm.cmd.

revans2:
OK I am +1 for this then.

harshach:
+1. @arunmahadevan this should go into 1.x-branch as well.

harshach:
Thanks @arunmahadevan merged into master. Can you please open another PR for 1.x-branch.

arunmahadevan:
@harshach raised https://github.com/apache/storm/pull/1115 for 1.x-branch.

PR #1097, opened 2016-02-11
STORM-1248: port backtype.storm.messaging.loader to java

abhishekagarwal87:


storm-core/src/clj/org/apache/storm/daemon/worker.clj
```diff
@@ -461,11 +460,11 @@
             )))))
 
 (defn register-callbacks [worker]
-  (log-message "Registering IConnectionCallbacks for " (:assignment-id worker) ":" (:port worker))
-  (msg-loader/register-callback (:transfer-local-fn worker)
-                                (:receiver worker)
-                                (:storm-conf worker)
-                                (worker-context worker)))
+  (let [transfer-local-fn (:transfer-local-fn worker) ^IConnection socket (:receiver worker)]
```
knusbaum: `let` bindings should be one per line.

abhishekagarwal87: fixed

storm-core/src/clj/org/apache/storm/daemon/worker.clj
```diff
@@ -487,11 +486,12 @@
...
 (defn register-callbacks [worker]
-  (log-message "Registering IConnectionCallbacks for " (:assignment-id worker) ":" (:port worker))
-  (msg-loader/register-callback (:transfer-local-fn worker)
-                                (:receiver worker)
-                                (:storm-conf worker)
-                                (worker-context worker)))
+  (let [transfer-local-fn (:transfer-local-fn worker)
+        ^IConnection socket (:receiver worker)]
+    (log-message "Registering IConnectionCallbacks for " (:assignment-id worker) ":" (:port worker))
+    (.registerRecv socket (DeserializingConnectionCallback. (:storm-conf worker)
```
knusbaum: Could we just inline `DeserializingConnectionCallback`? It's only used here and it'd be only a couple lines of Clojure. We're going to have to get rid of Java code that references Clojure like `DeserializingConnectionCallback` does anyway.

abhishekagarwal87: It wouldn't be that simple I suppose. There is a thread local KryoTupleDeserializer in the callback. I don't know how that will be implemented in clojure. I might end up porting more java code to clojure than other way around.

knusbaum:
```
(reify ThreadLocal
  (initialValue [this] (KryoTupleDeserializer. (:storm-conf worker) (worker-context worker)))
```
That's it for the thread-local.

abhishekagarwal87: I am still not very convinced. I will port DeserializationCallBack to clojure, then we will again port it back to java. IMO it will be better to make it inline later, if needed. 
@revans2 Since you wrote DeserializationCallback, was there any particular reason for writing it outside worker.clj e.g. intending to use it in other places?

knusbaum: @abhishekagarwal87 That's fine with me. It's true we'd have to translate it back to java, even if inlined.

knusbaum: @abhishekagarwal87 That's fine with me. It's true we'd have to translate it back to java, even if inlined.

knusbaum:
+1

revans2:
+1 pending travis

abhishekagarwal87:
Travis check passed. Ready for merge

PR #1098, opened 2016-02-11
[STORM-1253] - port backtype.storm.timer to java

jerrypeng:


storm-core/src/clj/org/apache/storm/daemon/supervisor.clj
```diff
@@ -332,19 +333,24 @@
...
-                                         (log-error t "Error when processing event")
-                                         (Utils/exitProcess 20 "Error when processing an event")
-                                         ))
-   :blob-update-timer (mk-timer :kill-fn (defn blob-update-timer
-                                           [t]
-                                           (log-error t "Error when processing event")
-                                           (Utils/exitProcess 20 "Error when processing a event"))
-                                :timer-name "blob-update-timer")
+   :heartbeat-timer (StormTimer/mkTimer nil
+                      (reify StormTimer$TimerFunc
```
abhishekagarwal87: This implementation of TimerFunc is repeated in many places. Does it make sense to move it to java based implementation of TimerFunc and use everywhere?

storm-core/src/jvm/org/apache/storm/StormTimer.java
```diff
@@ -0,0 +1,231 @@
...
+    }
+    public static void schedule(StormTimerTask task, int delaySecs, TimerFunc afn) {
+        schedule(task, delaySecs, afn, true, 0);
+    }
+
+    public static void scheduleRecurring(final StormTimerTask task, int delaySecs, final int recurSecs, final TimerFunc afn) {
+        schedule(task, delaySecs, new TimerFunc() {
+            @Override
+            public void run(Object o) {
+                afn.run(null);
```
abhishekagarwal87: can you rename afn since it is a java object

storm-core/src/jvm/org/apache/storm/StormTimer.java
```diff
@@ -0,0 +1,231 @@
...
+        }
+        synchronized (task.lock) {
+            task.add(new QueueEntry(endTimeMs, afn, id));
+        }
+    }
+    public static void schedule(StormTimerTask task, int delaySecs, TimerFunc afn) {
+        schedule(task, delaySecs, afn, true, 0);
+    }
+
+    public static void scheduleRecurring(final StormTimerTask task, int delaySecs, final int recurSecs, final TimerFunc afn) {
```
abhishekagarwal87: can you keep the order of arguments consistent with above scheduleFunction? 
`
StormTimerTask task, int delaySecs, TimerFunc afn, final int recurSecs
`

storm-core/src/jvm/org/apache/storm/StormTimer.java
```diff
@@ -0,0 +1,231 @@
...
+            @Override
+            public void run(Object o) {
+                afn.run(null);
+                // This avoids a race condition with cancel-timer.
+                schedule(task, recurSecs, this, false, 0);
+            }
+        });
+    }
+
+    public static void scheduleRecurringWithJitter(final StormTimerTask task, int delaySecs, final int recurSecs, final int jitterMs, final TimerFunc afn) {
```
abhishekagarwal87: same comment regarding the method parameters ordering.

storm-core/src/jvm/org/apache/storm/StormTimer.java
abhishekagarwal87: I see now that it will lead to changes in parameter order everywhere so I am fine with this order.

storm-core/src/jvm/org/apache/storm/StormTimer.java
```diff
@@ -0,0 +1,231 @@
...
+        public final String id;
+
+        public QueueEntry(Long endTimeMs, TimerFunc afn, String id) {
+            this.endTimeMs = endTimeMs;
+            this.afn = afn;
+            this.id = id;
+        }
+
+        @Override
+        public String toString() {
```
abhishekagarwal87: is it needed?

storm-core/src/jvm/org/apache/storm/StormTimer.java
```diff
@@ -0,0 +1,231 @@
...
+        }
+        task.setOnKillFunc(onKill);
+        task.setActive(true);
+
+        task.setDaemon(true);
+        task.setPriority(Thread.MAX_PRIORITY);
+        task.start();
+        return task;
+    }
+    public static void schedule(StormTimerTask task, int delaySecs, TimerFunc afn, boolean checkActive, int jitterMs) {
```
abhishekagarwal87: Timer active check is missing.

storm-core/src/clj/org/apache/storm/daemon/supervisor.clj
jerrypeng: Perhaps it does, but lets try to keep things as close to the same as before during this initial round of translation.

storm-core/src/jvm/org/apache/storm/StormTimer.java
jerrypeng: will refactor

storm-core/src/jvm/org/apache/storm/StormTimer.java
jerrypeng: @abhishekagarwal87 So I guess you are ok with this?

storm-core/src/jvm/org/apache/storm/StormTimer.java
jerrypeng: only for debugging but we can remove it

storm-core/src/jvm/org/apache/storm/StormTimer.java
jerrypeng: will add

storm-core/src/jvm/org/apache/storm/StormTimer.java
abhishekagarwal87: Yes.

storm-core/src/jvm/org/apache/storm/StormTimer.java
```diff
@@ -0,0 +1,274 @@
...
+            public int compare(Object o1, Object o2) {
+                return ((QueueEntry)o1).endTimeMs.intValue() - ((QueueEntry)o2).endTimeMs.intValue();
+            }
+        });
+
+        // boolean to indicate whether timer is active
+        private AtomicBoolean active = new AtomicBoolean(false);
+
+        // function to call when timer is killed
+        private TimerFunc onKill;
```
revans2: In other places we have translated the killFunc to be a Thread.UncaughtExceptionHandler, and I think in this case we should stay with that.  It makes things more java compatible.

storm-core/src/jvm/org/apache/storm/StormTimer.java
```diff
@@ -0,0 +1,274 @@
...
+public class StormTimer {
+    private static final Logger LOG = LoggerFactory.getLogger(StormTimer.class);
+
+    public interface TimerFunc {
+        public void run(Object o);
+    }
+
+    public static class QueueEntry {
+        public final Long endTimeMs;
+        public final TimerFunc afn;
```
revans2: For these timer funcs they don't take any parameters, and in other places for these we have used a Runnable for this.  Using a Runnable also makes integration with clojure very nice because a function with no arguments is already a Runnable.

storm-core/src/jvm/org/apache/storm/StormTimer.java
```diff
@@ -0,0 +1,274 @@
...
+        }
+    }
+
+    /**
+     * Makes a Timer in the form of a StormTimerTask Object
+     * @param name name of the timer
+     * @param onKill function to call when timer is killed unexpectedly
+     * @return StormTimerTask object that was initialized
+     */
+    public static StormTimerTask mkTimer(String name, TimerFunc onKill) {
```
revans2: Why not just make this a constructor?

revans2: Why not make this a Constructor, and hide StormTimerTask as an implementation detail?

storm-core/src/jvm/org/apache/storm/StormTimer.java
```diff
@@ -0,0 +1,274 @@
...
+        public QueueEntry(Long endTimeMs, TimerFunc afn, String id) {
+            this.endTimeMs = endTimeMs;
+            this.afn = afn;
+            this.id = id;
+        }
+    }
+
+    public static class StormTimerTask extends Thread {
+
+        private PriorityBlockingQueue<QueueEntry> queue = new PriorityBlockingQueue<QueueEntry>(10, new Comparator() {
```
revans2: `s/Comparator/Comparator<QueueEntry>/`

storm-core/src/jvm/org/apache/storm/StormTimer.java
```diff
@@ -0,0 +1,274 @@
...
+        private TimerFunc onKill;
+
+        //random number generator
+        private Random random = new Random();
+
+        // synchronization for cancelling the timer
+        private Semaphore cancelNotifier = new Semaphore(0);
+
+        //used for synchronization
+        private Object lock = new Object();
```
revans2: The lock in the original code is used for access to the PrioirtyQueue and only covered critical sections that were single operations.  In this code we are using a PriorityBlockingQueue which is thread safe so there is no reason for this lock.

For some reason we also held the lock when trying to shut down the thread, which I just don't understand and we should be able to ignore.

storm-core/src/jvm/org/apache/storm/StormTimer.java
```diff
@@ -0,0 +1,274 @@
...
+        private AtomicBoolean active = new AtomicBoolean(false);
+
+        // function to call when timer is killed
+        private TimerFunc onKill;
+
+        //random number generator
+        private Random random = new Random();
+
+        // synchronization for cancelling the timer
+        private Semaphore cancelNotifier = new Semaphore(0);
```
revans2: I personally would rather use Thread.join instead of rewriting that same functionality using a Semaphore.

storm-core/src/jvm/org/apache/storm/StormTimer.java
```diff
@@ -0,0 +1,274 @@
...
+                    synchronized (this.lock) {
+                        queueEntry = this.queue.peek();
+                    }
+                    if ((queueEntry != null) && (Time.currentTimeMillis() >= queueEntry.endTimeMs)) {
+                        // It is imperative to not run the function
+                        // inside the timer lock. Otherwise, it is
+                        // possible to deadlock if the fn deals with
+                        // other locks, like the submit lock.
+                        synchronized (this.lock) {
+                            this.queue.poll();
```
revans2: Crud.  There is actually a race here.  It is in the original code too  We cannot blindly do a poll and throw away the next thing on the queue, because something else may have been prioritized ahead of what we just peeked at.

We don't run at a high enough load that I think this is critical, but we should fix it.  Just replacing the queueEntry with the polled one is probably enough.

```
queueEntry = queue.peek();
if ((queueEntry != null) && (Time.currentTimeMillis() >= queueEntry.endTimeMs)) {
  queueEntry = queue.poll();
  queueEntry.afn.run();
}
```

storm-core/src/jvm/org/apache/storm/StormTimer.java
```diff
@@ -0,0 +1,274 @@
...
+                    } else if (queueEntry != null) {
+                        //  If any events are scheduled, sleep until
+                        // event generation. If any recurring events
+                        // are scheduled then we will always go
+                        // through this branch, sleeping only the
+                        // exact necessary amount of time. We give
+                        // an upper bound, e.g. 1000 millis, to the
+                        // sleeping time, to limit the response time
+                        // for detecting any new event within 1 secs.
+                        Time.sleep(Math.min(1000, (queueEntry.endTimeMs - Time.currentTimeMillis())));
```
revans2: Could you file a follow on JIRA so we can figure out a way to do a notify with simulate time, so we don't need this 1 second polling.

storm-core/src/jvm/org/apache/storm/StormTimer.java
```diff
@@ -0,0 +1,274 @@
...
+                        // response time for detecting any new event
+                        // schedulings when there are no scheduled
+                        // events.
+                        Time.sleep(1000);
+                    }
+                } catch (Throwable t) {
+                    if (!(Utils.exceptionCauseIsInstanceOf(InterruptedException.class, t))) {
+                        this.onKill.run(t);
+                        this.setActive(false);
+                        throw new RuntimeException(t);
```
revans2: We can just return here instead of throwing the exception that we just handled.

revans2:
I didn't realize how messed up timer.clj really was.  Besides fixing the locking issues in the code, my biggest suggestion would be to make it fully object oriented and make all of the static methods regular methods/constructors.  It is already object oriented, we just need to take it the rest of the way.  And then we can make it AutoClosable too.

storm-core/src/jvm/org/apache/storm/StormTimer.java
jerrypeng: The current Timer API allows users to set a "onKill" function to be executed if an exception is thrown in the timer thread.  The "onKill" function take one argument i.e. the throwable that was caught.  So should we use Runnable interface in conjunction with TimerFuncs interface since the "onKill" function cannot be a Runnable as it takes one parameter.  Seems kind of messy if we use the both

storm-core/src/jvm/org/apache/storm/StormTimer.java
jerrypeng: will change

hustfxj:
I would agree with revans2. It is better to  make it fully object oriented and make all of the static methods regular methods/constructors.

storm-core/src/jvm/org/apache/storm/StormTimer.java
jerrypeng: Ya I noticed that too when I first translated the code but I forgot the mention it.  Instead of a poll can we just use a remove? Such as:

queueEntry = queue.peek();
if ((queueEntry != null) && (Time.currentTimeMillis() >= queueEntry.endTimeMs)) {
  queueEntry = queue.remove(queueEntry);
  queueEntry.afn.run();
}

So that we will alway remove what we "peek"ed even in the instance when the head of the queue changed because a new entry got added into the priority queue between peek and poll

storm-core/src/jvm/org/apache/storm/StormTimer.java
jerrypeng: ok we will remove

jerrypeng:
@revans2 thank you for your review!  I have addressed all your comments.

storm-core/src/jvm/org/apache/storm/StormTimer.java
jerrypeng: https://issues.apache.org/jira/browse/STORM-1554

storm-core/src/jvm/org/apache/storm/StormTimer.java
```diff
@@ -0,0 +1,246 @@
...
+ * code that does asynchronous work on the timer thread
+ */
+
+public class StormTimer implements AutoCloseable{
+    private static final Logger LOG = LoggerFactory.getLogger(StormTimer.class);
+
+    public static class QueueEntry {
+        public final Long endTimeMs;
+        public final Runnable func;
+        public final String id;
```
revans2: What is this ID used for?  Seems like it is a UUID that is generated but never read.

storm-core/src/jvm/org/apache/storm/StormTimer.java
```diff
@@ -0,0 +1,246 @@
...
+                // This avoids a race condition with cancel-timer.
+                schedule(recurSecs, this, false, jitterMs);
+            }
+        });
+    }
+
+    /**
+     * check if timer is active
+     */
+    public void checkActive() {
```
revans2: This should probably not be public.  It was private in the original code.

storm-core/src/jvm/org/apache/storm/StormTimer.java
```diff
@@ -0,0 +1,246 @@
...
+            return this.active.get();
+        }
+
+        public void add(QueueEntry queueEntry) {
+            this.queue.add(queueEntry);
+        }
+    }
+
+    //task to run
+    StormTimerTask task = new StormTimerTask();
```
revans2: Can we make this final? and ideally private too.  No reason to change it currently.

storm-core/src/jvm/org/apache/storm/StormTimer.java
jerrypeng: I am not sure. It was included in the original clojure code.  Though I did find it easier to debug the timer code when each timer event is tagged with a unique id

storm-core/src/jvm/org/apache/storm/StormTimer.java
```diff
@@ -0,0 +1,246 @@
...
+    /**
+     * Schedule a function to be executed in the timer
+     * @param delaySecs the number of seconds to delay before running the function
+     * @param func the function to run
+     * @param checkActive whether to check is the timer is active
+     * @param jitterMs add jitter to the run
+     */
+    public void schedule(int delaySecs, Runnable func, boolean checkActive, int jitterMs) {
+        if (this.task == null) {
+            throw new RuntimeException("task is null!");
```
revans2: If task is set to final there is no need to check for null everywhere.

storm-core/src/jvm/org/apache/storm/StormTimer.java
jerrypeng: will change

revans2:
A few minor comment but overall it looks really good.

storm-core/src/jvm/org/apache/storm/StormTimer.java
jerrypeng: will change

storm-core/src/jvm/org/apache/storm/StormTimer.java
jerrypeng: will fix

jerrypeng:
@revans2 thanks again for the review.  I have addressed your comments

storm-core/src/jvm/org/apache/storm/StormTimer.java
```diff
@@ -0,0 +1,240 @@
...
+import java.util.concurrent.PriorityBlockingQueue;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+/**
+ * The timer defined in this file is very similar to java.util.Timer, except
+ * it integrates with Storm's time simulation capabilities. This lets us test
+ * code that does asynchronous work on the timer thread
+ */
+
+public class StormTimer implements AutoCloseable{
```
harshach: space after AutoCloseable

storm-core/src/jvm/org/apache/storm/StormTimer.java
```diff
@@ -0,0 +1,240 @@
...
+        public QueueEntry(Long endTimeMs, Runnable func, String id) {
+            this.endTimeMs = endTimeMs;
+            this.func = func;
+            this.id = id;
+        }
+    }
+
+    public static class StormTimerTask extends Thread {
+
+        private PriorityBlockingQueue<QueueEntry> queue = new PriorityBlockingQueue<QueueEntry>(10, new Comparator<QueueEntry>() {
```
harshach: can you add comment about why the initialCapacity is 10

storm-core/src/jvm/org/apache/storm/StormTimer.java
jerrypeng: will fix

storm-core/src/jvm/org/apache/storm/StormTimer.java
jerrypeng: not sure why... this was the initial capacity set in the original clojure code

jerrypeng: lets remove it

jerrypeng: lets actually set the initialCapacity to the default (11).

jerrypeng:
@harshach thanks for your review, I have addressed your comments

revans2:
+1

revans2:
It looks like there is a minor merge conflict with some imports in two files.

jerrypeng:
@revans2 I just upmerged

revans2:
Still +1 I built and ran the tests myself so I am not waiting for travis.

PR #1100, opened 2016-02-11
[STORM-1230] port backtype.storm.process-simulator to java.

zhuoliu:
From storm-core/src/clj/org/apache/storm/process-simulator.clj to storm-core/src/jvm/org/apache/storm/ProcessSimulator.java


storm-core/src/jvm/org/apache/storm/ProcessSimulator.java
```diff
@@ -0,0 +1,89 @@
...
+     * @param pid
+     */
+    public static void killProcess(String pid) {
+        synchronized (lock) {
+            LOG.info("Begin killing process " + pid);
+            Shutdownable shutdownHandle = getProcessHandle(pid);
+            if (shutdownHandle != null) {
+                shutdownHandle.shutdown();
+            }
+            processMap.remove(pid);
```
abhishekagarwal87: I think this should be moved inside the if block and before the shutdown call. Once process is marked for shutdown, it should be removed from map (similar to clojure code) 
``` 
Shutdownable shutdownHandle = processMap.remove(pid)
```

storm-core/src/jvm/org/apache/storm/ProcessSimulator.java
```diff
@@ -0,0 +1,89 @@
...
+            LOG.info("Successfully killing process " + pid);
```
abhishekagarwal87: typo: killed.

abhishekagarwal87: and should this be logged only if pid is present?

storm-core/src/jvm/org/apache/storm/ProcessSimulator.java
```diff
@@ -0,0 +1,89 @@
...
+import java.util.Collection;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ProcessSimulator {
+    private static Logger LOG = LoggerFactory.getLogger(ProcessSimulator.class);
+    protected static Object lock = new Object();
```
abhishekagarwal87: why are these declared protected?

storm-core/src/jvm/org/apache/storm/ProcessSimulator.java
```diff
@@ -0,0 +1,89 @@
...
+        processMap.put(pid, shutdownable);
+    }
+
+    /**
+     * Get a process' handle
+     * 
+     * @param pid
+     * @return
+     */
+    protected static Shutdownable getProcessHandle(String pid) {
```
abhishekagarwal87: This function is only used in this class. It can be inlined.

storm-core/src/jvm/org/apache/storm/ProcessSimulator.java
```diff
@@ -0,0 +1,89 @@
...
+
+    /**
+     * kill all processes
+     */
+    public static void killAllProcesses() {
+        Set<String> pids = processMap.keySet();
+        for (String pid : pids) {
+            killProcess(pid);
+        }
+        LOG.info("Successfully kill all processes");
```
abhishekagarwal87: typo: killed. You can add a begin info message as well.

storm-core/src/jvm/org/apache/storm/ProcessSimulator.java
```diff
@@ -0,0 +1,89 @@
...
+import org.apache.storm.daemon.Shutdownable;
+
+import java.util.Collection;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ProcessSimulator {
```
abhishekagarwal87: Also ProcessSimulator name is misleading judging from how it is being used in supervisor. ProcessManager sounds a better name to me.

abhishekagarwal87: Ahh. Ignore that. code is doing something else.

storm-core/src/jvm/org/apache/storm/ProcessSimulator.java
zhuoliu: Here I would prefer to keep it unchanged since It would be possible that <pid, null> be wrongly put in the map by registerPorcess. Place the "remove" outside of the "if" would help and not hurt, also this translation conforms the original clj code logic.

storm-core/src/jvm/org/apache/storm/ProcessSimulator.java
revans2: Can we put in some javadocs to explain that the ProcessSimulator keeps track of Shutdownable objects in place of actual processes.

Also can we file a JIRA to replace Shutdownable with AutoClosable?

storm-core/src/jvm/org/apache/storm/ProcessSimulator.java
```diff
@@ -0,0 +1,80 @@
...
+    /**
+     * kill all processes
+     */
+    public static void killAllProcesses() {
+        LOG.info("Begin killing all processes");
+        Set<String> pids = processMap.keySet();
+        for (String pid : pids) {
+            killProcess(pid);
+        }
+        LOG.info("Successfully killed all processes");
```
revans2: We have extra logging here.  In the original code the only log message was once for each process that was "Killed".

revans2:
Only two very minor comments, other even without them I am a +1

storm-core/src/jvm/org/apache/storm/ProcessSimulator.java
abhishekagarwal87: I meant it something such as follows 
```
Shutdownable shutdownHandle = processMap.remove(pid)
if (shutdownHandle != null) {
                shutdownHandle.shutdown();
            }
```
Sorry for the confusion.

zhuoliu:
Hi @revans2, really appreciate your comments! I addressed your comment in [Storm-1563](https://issues.apache.org/jira/browse/STORM-1563) to change all the Shutdownable to be AutoClosable in storm.

redsanket:
upmerge

zhuoliu:
@redsanket thanks, upmerged.

storm-core/src/jvm/org/apache/storm/ProcessSimulator.java
```diff
@@ -0,0 +1,82 @@
...
+            if (shutdownHandle != null) {
+                shutdownHandle.shutdown();
+            }
+            processMap.remove(pid);
+            LOG.info("Successfully killed process " + pid);
+        }
+    }
+
+    /**
+     * kill all processes
```
redsanket: capital "Kill" just for consistency

redsanket:
Very minor comment, not significant LGTM +1

PR #1105, opened 2016-02-12
Storm 1539 - Improve Storm ACK-ing performance

roshannaik:
PR for 1.x-branch

### DIFF
--- a/storm-core/src/clj/org/apache/storm/daemon/executor.clj
+++ b/storm-core/src/clj/org/apache/storm/daemon/executor.clj
@@ -581,7 +581,7 @@
                                                                     (if (sampler) (System/currentTimeMillis))])
                                              (task/send-unanchored task-data
                                                                    ACKER-INIT-STREAM-ID
-                                                                   [root-id (bit-xor-vals out-ids) task-id]))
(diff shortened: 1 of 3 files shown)


knusbaum:
Should remove the corresponding function from `util.clj`

revans2:
+1

HeartSaVioR:
+1

vesense:
LGTM

knusbaum:
+1

ptgoetz:
+1

zkendall:
I found this as a performance update in change log. I'm curious why the new implementation is so much better than the old?

One consideration is [bit-xor contains `reduce` inside it already](https://github.com/clojure/clojure/blob/clojure-1.7.0/src/clj/clojure/core.clj#L1276). So maybe removing the extra reduce in the closure would have helped without resulting to Java alternative.

roshannaik:
The clojure reduce operation was much slower than a Java for loop. Also Storm is moving away from clojure so holding on to clojure was not important.

zkendall:
Did not know that. Is there a roadmap/timeline on the java effort?

roshannaik:
happening incrementally, most of it happening in 2.0 branch   ... look for jiras with titles starting with  "port .. "

PR #1109, opened 2016-02-15
[STORM-1552] Fix topology event sampling log dir

arunmahadevan:
Currently the events are logged under "storm-local/workers-artifacts/{storm-id}/{port}/events.log".
and the "events" link in UI does not display the log file.

The events.log should be kept under "logs/workers-artifacts/{storm-id}/{port}/events.log"
so that its viewable via logviewer.

### DIFF
--- a/storm-core/src/jvm/org/apache/storm/metric/FileBasedEventLogger.java
+++ b/storm-core/src/jvm/org/apache/storm/metric/FileBasedEventLogger.java
@@ -17,6 +17,7 @@
  */
 package org.apache.storm.metric;
 
+import org.apache.storm.Config;
 import org.apache.storm.task.TopologyContext;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -45,7 +46,7 @@ public class FileBasedEventLogger implements IEventLogger {
 
     private void initLogWriter(Path logFilePath) {
         try {
-            LOG.info("logFilePath {}", logFilePath);
+            LOG.info("FileBasedEventLogger log path {}", logFilePath);
             eventLogPath = logFilePath;
             eventLogWriter = Files.newBufferedWriter(eventLogPath, StandardCharsets.UTF_8, StandardOpenOption.CREATE,
                                                      StandardOpenOption.WRITE, StandardOpenOption.APPEND);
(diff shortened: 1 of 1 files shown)


storm-core/src/jvm/org/apache/storm/metric/FileBasedEventLogger.java
```diff
@@ -76,33 +77,35 @@ public void run() {
...
         /*
          * Include the topology name & worker port in the file name so that
          * multiple event loggers can log independently.
          */
-        Path path = Paths.get(logDir, "workers-artifacts", stormId, Integer.toString(port), "events.log");
+        Path path = Paths.get(workersArtifactDir, stormId, Integer.toString(port), "events.log");
         if (!path.isAbsolute()) {
-            path = Paths.get(System.getProperty("storm.home"), logDir, "workers-artifacts",
-                    stormId, Integer.toString(port), "events.log");
+            path = Paths.get(getLogDir(stormConf), workersArtifactDir,
```
abhishekagarwal87: This is very similar to https://github.com/apache/storm/blob/master/storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java#L407

storm-core/src/jvm/org/apache/storm/metric/FileBasedEventLogger.java
```diff
@@ -76,33 +77,35 @@ public void run() {
         scheduler.scheduleAtFixedRate(task, FLUSH_INTERVAL_MILLIS, FLUSH_INTERVAL_MILLIS, TimeUnit.MILLISECONDS);
     }
 
+    private String getLogDir(Map stormConf) {
```
abhishekagarwal87: This function is already present in ConfigUtils.java

storm-core/src/jvm/org/apache/storm/metric/FileBasedEventLogger.java
arunmahadevan: ConfigUtils.java is not available in 1.x-branch. I wanted the fix to be applied to both 1.x and master so took 1.x as the base and submitted the PR. Someone can let me know in such cases one need to raise two separate PRs with different code to fix the same issue.

arunmahadevan:
@ptgoetz can you take a look at this?. Figured out that the event log link is broken while testing STORM-1540. This also needs to go into 1.x branch.

storm-core/src/jvm/org/apache/storm/metric/FileBasedEventLogger.java
abhishekagarwal87: I would be fine with different code. But if it is decided to keep the code same, please add a to-do here.

arunmahadevan:
https://github.com/apache/storm/pull/1123 addresses the issue for master branch and make use of Configutils. I will close this and raise it against 1.x-branch.

PR #1111, opened 2016-02-16
[STORM-1243] port HealthCheck to java

hustfxj:


storm-core/src/jvm/org/apache/storm/command/HealthCheck.java
```diff
@@ -0,0 +1,124 @@
...
+            final Thread curThread = Thread.currentThread();
+            // kill process when timeout
+            interruptThread = new Thread(new Runnable() {
+                @Override
+                public void run() {
+                    try {
+                        Thread.sleep(timeout);
+                        curThread.interrupt();
+                    } catch (InterruptedException e) {
+
```
revans2: can we have a comment like `//Ignored`

storm-core/src/jvm/org/apache/storm/command/HealthCheck.java
```diff
@@ -0,0 +1,124 @@
...
+                    }
+                }
+            });
+            interruptThread.start();
+            process.waitFor();
+            interruptThread.interrupt();
+
```
revans2: I know that this existed in the original code, but there is a race condition here.  If interruptThread interrupts us before we interrupt it, this thread will have it's interrupt status set, and we should clear it before going on, or future I/O could get interrupted before it starts.

This is really minor because the race is so rare, but I think if we add the following here we will be good.

```
curThread.interrupted(); //clearing status 
```

I think we just

storm-core/src/jvm/org/apache/storm/command/HealthCheck.java
```diff
@@ -0,0 +1,124 @@
...
+                    }
+                    if (str.startsWith("ERROR")) {
+                        return FAILED;
+                    }
+                }
+                return SUCCESS;
+            }
+            return FAILED_WITH_EXIT_CODE;
+        } catch (InterruptedException e) {
+            System.out.println("Script " + script + "timed out.");
```
revans2: Need a space before `timed out.`

storm-core/src/jvm/org/apache/storm/command/HealthCheck.java
```diff
@@ -0,0 +1,124 @@
...
+            interruptThread.interrupt();
+
+            if (process.exitValue() != 0) {
+                String str;
+                InputStream stdin = process.getInputStream();
+                BufferedReader reader = new BufferedReader(new InputStreamReader(stdin));
+                while ((str = reader.readLine()) != null) {
+                    if (StringUtils.isBlank(str)) {
+                        continue;
+                    }
```
revans2: I'm not sure this is needed.  It doesn't hurt things, but str should never be null because of the while, and if it is the empty string it will not start with "ERROR"

revans2:
+1 the translation looks really good.  I found one minor issue with the original code, and a few very minor style issues.

hustfxj:
@revans2  Thank you

ppoulosk:
+1, NB

longdafeng:
+1

revans2:
+1.  The code needs to be upmerged, but it is a minor issue in supervisor.clj

hustfxj:
@longdafeng @revans2 @ppoulosk  thank you for your review! I have addressed all your comments.

PR #1116, opened 2016-02-17
STORM-1522 should create error worker log location only when error-host and error-port are available

satishd:
STORM-1522 should create error worker log location only when error-hOst and error-port are available

### DIFF
--- a/storm-core/src/clj/org/apache/storm/ui/core.clj
+++ b/storm-core/src/clj/org/apache/storm/ui/core.clj
@@ -146,8 +146,10 @@
   (logviewer-link host (event-logs-filename topology-id port) secure?))
 
 (defn worker-log-link [host port topology-id secure?]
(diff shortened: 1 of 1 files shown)


harshach:
+1

storm-core/src/clj/org/apache/storm/ui/core.clj
```diff
@@ -146,8 +146,10 @@
   (logviewer-link host (event-logs-filename topology-id port) secure?))
 
 (defn worker-log-link [host port topology-id secure?]
-  (let [fname (logs-filename topology-id port)]
-    (logviewer-link host fname secure?)))
+  (if (or (empty? host) (let [port_str (str port "")] (or (empty? port_str) (= "0" port_str))))
```
redsanket: Can we have proper indentation instead of having a single long line. Also, Can we have if condition inside a single let binding instead of have two lets under an if

PR #1120, opened 2016-02-18
[STORM-1559] Get surefire-reports after travis-ci build failure

vesense:


vesense:
With these changes, many .txt files will be output. So I closed this.

PR #1121, opened 2016-02-18
[STORM-1556]: nimbus.clj/wait-for-desired-code-replication wrong reset for current-replication-count-jar in local mode

wuchong:
We do not go to count the number of jar-replication in local mode, but will count the number of conf-replication.


redsanket:
+1

ppoulosk:
+1, NB

revans2:
+1

PR #1122, opened 2016-02-18
storm-hdfs : change visibility of create and closeOutputFile methods to protected

jnioche:
When extending HdfsBolt it can be useful to be able to override the methods createOutputFile() or closeOutputFile(), for instance to add a header or footer to the files created. This is currently not possible when the class is outside the `org.apache.storm.hdfs.bolt` package. This patch makes it possible.

### DIFF
--- a/external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/AbstractHdfsBolt.java
+++ b/external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/AbstractHdfsBolt.java
@@ -231,7 +231,7 @@ public abstract class AbstractHdfsBolt extends BaseRichBolt {
      * @param tuple
      * @throws IOException
      */
-    abstract void writeTuple(Tuple tuple) throws IOException;
+    abstract protected void writeTuple(Tuple tuple) throws IOException;
 
     /**
      * Make the best effort to sync written data to the underlying file system.  Concrete classes should very clearly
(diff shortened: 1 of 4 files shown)


revans2:
+1 looks fine to me.

dossett:
Would there be a drawback to making all of the abstract methods protected if we want to support end user extensions?

ptgoetz:
+1

jnioche:
@dossett good idea - I have just committed your suggestion
@revans2 and @ptgoetz \: will the content of the master branch become 2.x and if so could this  be committed to the next 1.x release?

ptgoetz:
@jnioche yes, this can go into 1.0, all you need to do is request so (which you have). No need to open a second pull request unless it doesn't apply cleanly to the 1.x branch or a committer requests it.

I generally will try to resolve conflicts when merging, and punt back to the contributor when necessary if the conflicts are too ugly.

jnioche:
@ptgoetz perfect, thanks. BTW I really like the HDFS module, great work! Would be good to have compression for HdfsBolt, I might have a go at implementing that.

jnioche:
Anything blocking this from being committed (apart from committers' availability)?

revans2:
@jnioche just me not getting to it.  Will merge it in now...

revans2:
Thanks @jnioche I merged this into master and branch 1.

jnioche:
thanks for having taken the time to do it @revans2, much appreciated

PR #1124, opened 2016-02-18
STORM-1246: port backtype.storm.local-state to java

abhishekagarwal87:


harshach:
+1. Thanks for the updated patch.

PR #1125, opened 2016-02-18
[STORM-1561] Supervisor should kill/restart if existing worker has changed assignments

kishorvpatil:
Currently, supervisor validates new assignments against existing assignments by port. It should also check on the same port - if executors have changed. The new executors won't have any threads associated with them in the worker.


kishorvpatil:
@longdafeng I have refactored the changes into  separate method.

jerrypeng:
+1

ppoulosk:
+1, NB

kishorvpatil:
I merged this changed in both master and 1.x-branch.

PR #1123, opened 2016-02-18
closes STORM-1545 and STORM-1552

mproch:
Use ConfigUtils.workerArtifactsRoot to determine correct path

PR #1127, opened 2016-02-19
[STORM-1552] Fix topology event sampling log dir

arunmahadevan:
Currently the events are logged under "storm-local/workers-artifacts/{storm-id}/{port}/events.log".
and the "events" link in UI does not display the log file.

The events.log should be kept under "logs/workers-artifacts/{storm-id}/{port}/events.log"
so that its viewable via logviewer.

This fix is for 1.x-branch. For master this would be addressed by https://github.com/apache/storm/pull/1123

### DIFF
--- a/storm-core/src/jvm/org/apache/storm/metric/FileBasedEventLogger.java
+++ b/storm-core/src/jvm/org/apache/storm/metric/FileBasedEventLogger.java
@@ -17,6 +17,7 @@
  */
 package org.apache.storm.metric;
 
+import org.apache.storm.Config;
 import org.apache.storm.task.TopologyContext;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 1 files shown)


arunmahadevan:
@ptgoetz can you take a look ?

storm-core/src/jvm/org/apache/storm/metric/FileBasedEventLogger.java
```diff
@@ -45,7 +46,7 @@
 
     private void initLogWriter(Path logFilePath) {
         try {
-            LOG.info("logFilePath {}", logFilePath);
+            LOG.info("FileBasedEventLogger log path {}", logFilePath);
```
ptgoetz: Minor nit: we probably don't need the classname repeated here, since it is part of the Logger configuration.

ptgoetz:
+1 one very minor comment

ppoulosk:
+1, NB

arunmahadevan:
@ptgoetz updated log message.

harshach:
@arunmahadevan when I run unit tests with the patch its leaving storm-core/null dir with worker logs. Can you check that

arunmahadevan:
@harshach `storm.home` is not set during unit tests so the logs were created under wrong directory. Commit https://github.com/apache/storm/commit/f3baed0d4bb9c415657dfa97f33519994e991e17 fixes it. Please take a look.

harshach:
+1

PR #1129, opened 2016-02-19
[STORM-1488] fix broken component error times

d2r:
Note that the errorTime reported for the component/ path was previously in milliseconds, but the errorTime reported for the topology/ path was in seconds.  This inconsistency fixed, but we should also update the REST API example. The REST API doc currently does not specify the units.

This should also go to master branch.

### DIFF
--- a/storm-core/src/clj/org/apache/storm/ui/core.clj
+++ b/storm-core/src/clj/org/apache/storm/ui/core.clj
@@ -157,11 +157,6 @@
 (defn supervisor-log-link [host]
   (url-format "http://%s:%s/daemonlog?file=supervisor.log" host (*STORM-CONF* LOGVIEWER-PORT)))
 
-(defn get-error-time
-  [error]
-  (if error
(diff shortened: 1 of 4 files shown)


revans2:
+1 the travis failure looks unrelated.  Do you have a corresponding fix for master?

d2r:
> +1 the travis failure looks unrelated. Do you have a corresponding fix for master?

I do not have a branch ready for master, but I can do that.  The only conflict—last I checked—was that util/time-delta was translated to Time#deltaSecs or some such.

d2r:
I saw an epoc date reported on the master branch when no errors had yet been reported.  Defaulting the error time to `0` instead of the empty string may not be handled correctly.  I need to investigate this first.

d2r:
The last error time on the topology page for bolts and spouts is once again blank when no errors have been reported.

revans2:
+1 looks good to me.

HeartSaVioR:
+1

PR #1130, opened 2016-02-20
Update doc for rebalance command

jnioche:
Trivial update for the rebalance command - now similar to the output of 'storm help rebalance'


hustfxj:
+1

jnioche:
Sorry for bumping it, but it is a trivial commit and I think it would be good to update the doc - especially prior to the new releases.

ptgoetz:
+1

PR #1132, opened 2016-02-22
[STORM-1065] kafka-partition can not find leader in zookeeper

lujinhong:
STORM-1065. storm-kafka : kafka-partition can not find leader in zookeeper

### DIFF
--- a/external/storm-kafka/src/jvm/org/apache/storm/kafka/DynamicBrokersReader.java
+++ b/external/storm-kafka/src/jvm/org/apache/storm/kafka/DynamicBrokersReader.java
@@ -85,7 +85,7 @@ public class DynamicBrokersReader {
               String brokerInfoPath = brokerPath();
               for (int partition = 0; partition < numPartitionsForTopic; partition++) {
                   int leader = getLeaderFor(topic,partition);
-                  if(leader == -1){
+                  if(leader == -1) {
                 	      LOG.warn("No leader found for partition " + topic + " : " + partition);
                 	      continue;
                   }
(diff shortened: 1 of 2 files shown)


lujinhong:
In KafkaUtil.java, NOT_LEADER_FOR_PARTITION happend,  catching the NOT_LEADER_FOR_PARTITION execption will be a good idea.

This exception should be catch and then log & ignored Or invole the ZkCoordinator#refresh().

external/storm-kafka/src/jvm/org/apache/storm/kafka/DynamicBrokersReader.java
```diff
@@ -85,6 +85,10 @@ public DynamicBrokersReader(Map conf, String zkStr, String zkPath, String topic)
               String brokerInfoPath = brokerPath();
               for (int partition = 0; partition < numPartitionsForTopic; partition++) {
                   int leader = getLeaderFor(topic,partition);
+                  if(leader == -1){
```
abhishekagarwal87: indentation is off here. Please add space around curly brace.

abhishekagarwal87:
I am thinking of the side effects that this change can lead to -
1. Number of tasks could be more than number of partitions - Acceptable and will be corrected in next refresh
2. GlobalPartitionMapping may be different for multiple tasks - Depending on when the spout task reads the kafka cluster zookeeper, it is possible that task A reads 15 partitions and task B reads 16 partitions for a topic with total 16 partitions. This can be problematic as it will lead to two tasks assigning same partitions to themselves. It breaks the contract that same partition won't be read by two different tasks.

lujinhong:
1. when refreshFreqSecs expired, it will refresh.
2. for the second, I think it is about the task-partition mapping mechanism，is it in the clojure code of Task? And any suggestion?
thanks.

abhishekagarwal87:
https://github.com/apache/storm/blob/master/external/storm-kafka/src/jvm/org/apache/storm/kafka/KafkaUtils.java#L250
You can add the partitions to GlobalPartitionInfo even if leader is -1. That way, it is ensured that all tasks have the same view of partitions. PartitionManager for unassigned leaders can be just inactive and do nothing. 
FYI That may make the changes a bit more complex.

PR #1119, opened 2016-02-18
[STORM-1558] Utils in java breaks component page due to illegal type cast

unsleepy22:

PR #1139, opened 2016-02-22
[STORM-1254]: port ui.helper to java

wuchong:
port ui.helper to java

1.  I find that `defelem table` is never be called, so I  removed it.
2.  I haven't translated `defn requests-middleware`, as it's hard to translate and we will move ring to other REST framework. At that time , we can drop it directly.
3. a minor change for `config-filter` signature


storm-core/src/jvm/org/apache/storm/ui/UIHelpers.java
```diff
@@ -0,0 +1,301 @@
...
+ * <p/>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.storm.ui;
+
+import clojure.lang.*;
```
abhishekagarwal87: can you remove wild card imports in line with other java classes

storm-core/src/jvm/org/apache/storm/ui/UIHelpers.java
```diff
@@ -0,0 +1,301 @@
...
+                keyword("headers"), new PersistentArrayMap(new Object[]{}),
+                keyword("status"), 400,
+                keyword("body"), ex.getMessage()
+        });
+    }
+
+    private static Server removeNonSslConnector(Server server) {
+        for (Connector c : server.getConnectors()) {
+            if (c != null && c instanceof SslSocketConnector) {
+                server.removeConnector(c);
```
abhishekagarwal87: connector should be removed if it is not a SslSocketConnector

storm-core/src/jvm/org/apache/storm/ui/UIHelpers.java
```diff
@@ -0,0 +1,301 @@
...
+                    context.addFilter(filterHolder, "/*", FilterMapping.ALL);
+                }
+            }
+            context.addFilter(mkAccessLoggingFilterHandle(), "/*", EnumSet.allOf(DispatcherType.class));
+            server.setHandler(context);
+        }
+    }
+
+    public static APersistentMap ringResponseFromException(Exception ex) {
+        return new PersistentArrayMap(new Object[]{
```
abhishekagarwal87: You need not use Persistent* classes. Instead you can return lists/map etc. and in the clojure code call clojurify-structure on the result e.g.
```
(clojurify-structure (UIHelpers/ringResponseFromException <arguments>))
```

abhishekagarwal87:
Thanks @wuchong . My comments 
Let's just completely do away with clojure classes in UIHelpers. I have given an example in inline comments. 
commits should preferably have a JIRA id. 
You might have to squash your commits after the code review is over.

storm-core/src/jvm/org/apache/storm/ui/UIHelpers.java
```diff
@@ -0,0 +1,301 @@
...
+                    tmp.addFirst(val % div + dividers[0]);
+                    val = val / div;
+                } else {
+                    tmp.addFirst(val + dividers[0]);
+                }
+            }
+        }
+        return Joiner.on(" ").join(tmp);
+    }
+
```
hustfxj: I know "pretty-uptime-ms" is never be called. But it is API function which can be called by others.

hustfxj:
+1

wuchong:
@abhishekagarwal87 @hustfxj  Thanks for your suggestion, I have addressed all your comments. And fix the commits.

Let me know if you find anything else.

storm-core/src/clj/org/apache/storm/ui/helpers.clj
```diff
@@ -46,197 +46,3 @@
   (fn [req]
```
revans2: Should we move this someplace else?  Or perhaps inline it into the servers that use it?

storm-core/src/jvm/org/apache/storm/ui/UIHelpers.java
```diff
@@ -0,0 +1,319 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.storm.ui;
+
+import clojure.lang.Keyword;
+import clojure.lang.RT;
```
revans2: We should not depend on clojure internals.  For keywords we should have something that can translate clojure maps/keywords into java objects, or just use those java object directly.

storm-core/src/jvm/org/apache/storm/ui/UIHelpers.java
```diff
@@ -0,0 +1,319 @@
...
+            new String[]{"h", "24"},
+            new String[]{"d", null}};
+
+    public static String prettyUptimeStr(String val, String[][] dividers) {
+        int uptime = Integer.parseInt(val);
+        LinkedList<String> tmp = new LinkedList<>();
+        for (String[] divider : dividers) {
+            if (uptime > 0) {
+                if (divider[1] != null) {
+                    int div = Integer.parseInt(divider[1]);
```
revans2: It seems really odd to parse an int out of a Constant that we are setting.  I think this is OK, but just really odd.

storm-core/src/jvm/org/apache/storm/ui/UIHelpers.java
```diff
@@ -0,0 +1,319 @@
...
+    }
+
+    public static Map unauthorizedUserJson(String user) {
+        return ImmutableMap.of(
+                "error", "No Authorization",
+                "errorMessage", String.format("User %s is not authorized.", user)
+        );
+    }
+
+    public static List unauthorizedUserHtml(String user) {
```
revans2: This is very hiccup specific and is now only used by logviewer.  I would prefer to see this removed and the original function placed in logviewer so when we redo the logviewer code and replace hiccup we can remove this code entirely.

storm-core/src/jvm/org/apache/storm/ui/UIHelpers.java
```diff
@@ -0,0 +1,319 @@
...
+
+        if (tsPath != null && tsPassword != null && tsType != null) {
+            factory.setTrustStore(tsPath);
+            factory.setTrustStoreType(tsType);
+            factory.setTrustStorePassword(tsPassword);
+        }
+
+        if (needClientAuth != null && needClientAuth) {
+            factory.setNeedClientAuth(true);
+        }
```
revans2: This should be an `else if`.  cond stops as soon as a condition is true.

storm-core/src/jvm/org/apache/storm/ui/UIHelpers.java
```diff
@@ -0,0 +1,319 @@
...
+                    }
+                    context.addFilter(filterHolder, "/*", FilterMapping.ALL);
+                }
+            }
+            context.addFilter(mkAccessLoggingFilterHandle(), "/*", EnumSet.allOf(DispatcherType.class));
+            server.setHandler(context);
+        }
+    }
+
+    public static Map ringResponseFromException(Exception ex) {
```
revans2: We should probably leave this in clojure and inline it in the files that use it, so when they move to new technology it can go away or the appropriate replacement code can be written.

storm-core/src/jvm/org/apache/storm/ui/UIHelpers.java
```diff
@@ -0,0 +1,319 @@
...
+
+    public static Map jsonResponse(Object data, String callback, Long status) {
+        return jsonResponse(data, callback, true, status, null);
+    }
+
+    public static Map jsonResponse(Object data, String callback, Map headers) {
+        return jsonResponse(data, callback, true, null, headers);
+    }
+
+    public static Map jsonResponse(Object data, String callback, boolean needSerialize, Long status, Map headers) {
```
revans2: Can we please split some of this up?

The final result is ring specific and not really what we want.  Can we have one function for getting the headers, and another function for creating the body.  Then creating the final map can still be in clojure and placed in each of the files that use it.

revans2:
A lot of this looks good, But it needs a little bit of work to move the clojure/ring/hiccup specific parts where they need to be.

storm-core/src/clj/org/apache/storm/ui/helpers.clj
wuchong: I prefer to leave it here, so when we move to new REST framework we can remove this code or the appropriate replacement code can be written.

I will add a comment here.

wuchong:
@revans2 @abhishekagarwal87  @hustfxj  Thanks for your review, I have addressed all the comments. 

Can you have a look again?

revans2:
+1 looks good.  I am still seeing some issues with 

```
java.lang.ClassCastException: java.lang.Integer cannot be cast to java.lang.String
	at org.apache.storm.ui.core$event_log_link.invoke(core.clj:148)
```

But it seems to be unrelated to this.

hustfxj:
+1. It looks fairly good. Thank you @wuchong

wuchong:
@revans2 It's a bug which has been fixed in #1119

revans2:
@wuchong I am still seeing it even after #1119 was merged in.

storm-core/src/clj/org/apache/storm/ui/core.clj
```diff
@@ -134,14 +136,16 @@
...
+    (UIHelpers/urlFormat "https://%s:%s/log?file=%s"
+      (to-array
+        [host
+        (*STORM-CONF* LOGVIEWER-HTTPS-PORT)
+        fname]))
+    (UIHelpers/urlFormat "http://%s:%s/log?file=%s"
+      (to-array
+        [host
+        (*STORM-CONF* LOGVIEWER-PORT)
+        fname]))))
```
revans2: This is the line that I was getting the exception on.  I will investigate a bit more to see if I can understand what is happening.

revans2: Never mind I found my bug.  I was running integration-test and not install.  My bad

PR #1141, opened 2016-02-23
[STORM-1570] Storm SQL support for nested fields and array

arunmahadevan:
Added support to handle array, map and nested field lookups and user defined functions
in Storm SQL.

### DIFF
--- a/external/sql/storm-sql-core/src/jvm/org/apache/storm/sql/compiler/ExprCompiler.java
+++ b/external/sql/storm-sql-core/src/jvm/org/apache/storm/sql/compiler/ExprCompiler.java
@@ -26,8 +26,12 @@ import org.apache.calcite.linq4j.tree.Primitive;
 import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rex.*;
 import org.apache.calcite.runtime.SqlFunctions;
+import org.apache.calcite.schema.Function;
+import org.apache.calcite.schema.impl.ReflectiveFunctionBase;
+import org.apache.calcite.schema.impl.ScalarFunctionImpl;
 import org.apache.calcite.sql.SqlOperator;
 import org.apache.calcite.sql.type.SqlTypeName;
+import org.apache.calcite.sql.validate.SqlUserDefinedFunction;
 import org.apache.calcite.util.BuiltInMethod;
 import org.apache.calcite.util.NlsString;
 import org.apache.calcite.util.Util;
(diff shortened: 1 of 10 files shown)


external/sql/storm-sql-core/src/jvm/org/apache/storm/sql/compiler/ExprCompiler.java
```diff
@@ -234,14 +253,20 @@ public String translate(ExprCompiler compiler, RexCall call) {
               pw.print(String.format("else if (%2$s == null) { %1$s = null; }\n", val, arg));
             }
           }
-          String calc = printMethodCall(method.method, args);
-          pw.print(String.format("else { %1$s = %2$s; }\n", val, calc));
+          String calc = printMethodCall(method, args);
+          String valCast = compiler.javaTypeName(call);
+          pw.print(String.format("else { %1$s = (%2$s)%3$s; }\n", val, valCast, calc));
```
satishd: It seems there is no need to cast here. Tried with supported functions and UDF and it is working fine without casting. You may want to remove this as casting in generating code looks redundant.

satishd:
Overall LGTM. +1 once the comment is addressed.

arunmahadevan:
@haohui In some cases like `true AND null` the query was still returning results, hence I had changed `StormSqlFunctions.eq` and `StormSqlFunctions.ne` to return false if either arg was null. The actual issue was in AND_EXPR and OR_EXPR, so I fixed it and reverted the earlier change. I have also split the change into two commits to review easily. 

Currently the UDF schema definition is not exposed via SQL CREATE. Let me know if you have any inputs on how this could be done.

haohui:
Thanks for updating the patch!

> In some cases like true AND null the query was still returning results, hence I had changed StormSqlFunctions.eq and StormSqlFunctions.ne to return false if either arg was null. 

That sounds like a bug to me. Can you please create a jira and post the query?

Posted inlined comment. I think the first commit is pretty good. Let's separate the second commit to another jira.

There are some pending tasks. I think we need to expose array / UDF in the DDL as well, my suggestion is to closely follow the syntax of Hive. I think both work can be done in separate jiras.

arunmahadevan:
- Removed the second commit. I will raise follow up JIRAs to support UDF and expose it via DDL.
- Removed guava test dep from storm-sql-runtime (not sure why this was a concern since storm-sql-core tests already uses guava classes).
- Added test cases to test the semantics for AND and OR with nullable on both sides.

harshach:
+1. @haohui can you check the final commit. Thanks

haohui:
The patch looks good. There are a few nits in the latest commit, +1 once they are addressed.

Thanks for the work!

arunmahadevan:
Addressed latest comments.

PR #1143, opened 2016-02-23
[STORM-1571]Improvment Kafka Spout Time Metric

darionyaphet:
[STORM-1571](https://issues.apache.org/jira/browse/STORM-1571) System.currentTimeMillis to calculation time interval is better than System.nanoTime :)

### DIFF
--- a/external/storm-kafka/src/jvm/org/apache/storm/kafka/PartitionManager.java
+++ b/external/storm-kafka/src/jvm/org/apache/storm/kafka/PartitionManager.java
@@ -170,7 +170,7 @@ public class PartitionManager {
 
 
     private void fill() {
-        long start = System.nanoTime();
+        long start = System.currentTimeMillis();
         Long offset;
 
         // Are there failed tuples? If so, fetch those first.
@@ -205,8 +205,7 @@ public class PartitionManager {
             
             return;
         }
-        long end = System.nanoTime();
-        long millis = (end - start) / 1000000;
+        long millis = System.currentTimeMillis() - start;
         _fetchAPILatencyMax.update(millis);
         _fetchAPILatencyMean.update(millis);
         _fetchAPICallCount.incr();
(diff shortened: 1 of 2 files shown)


unsleepy22:
I agree that use System.currentTimeMillis() instead of System.nanoTime() will improve performance, but if the code between the timing part is very fast, say, mostly < 1ms, then we possibly won't get useful metrics using System.currentTimeMillis.

darionyaphet:
hi @unsleepy22 this metric is use to count time interval when fetch messages from Kafka Broker . In fetchMessage() , it's building a `FetchResponse` , consumer fetch messages from broker and pick out message set . I think it won't be return in 1ms :)

roshannaik:
Also people may not care much about numbers less than 1 ms.

unsleepy22:
+1, but there're 2 occurrences of System.nanoTime in storm-kafka, @darionyaphet would you please change the other?

darionyaphet:
sure . I push a new commit to update TridentKafkaEmitter's time interval calculation . Should merge them into one commit ?

abhishekagarwal87:
Given that the timing is measured only in the fill call, which happens per batch of messages, the relative performance improvement is very low. By the way, call could return in 1ms if fetch size is configured at a lower value for better latency.

darionyaphet:
This question doesn't focus on performance improvement . 

Using System.currentTimeMillis to replace System.nanoTime is more reasonable

unsleepy22:
I agree with @darionyaphet . Actually I think all System.nanoTime callings in external modules should be replaced with System.currentTimeMillis since they use System.nanoTime to get a ns value and then convert the time back to ms, which is unnecessary.

System.nanoTime should only be used in such cases that we do need high precision timing, e.g., timing certain code snippets in metrics system. If called on a critical path, this method is 5+ times slower than System.currentTimeMillis.

knusbaum:
+1

PR #1147, opened 2016-02-24
[STORM-1252] port backtype.storm.stats to java

unsleepy22:
This PR is quite large but because the defrecords are changed to java POJO's and they're used in many places, I could not break it into small PR's.

Almost all code in stats.clj is translated to java except only one point:
StatsUtil.aggTopoExecsStats (stats/agg-topo-execs-stats) needs to pass an error function, which depends on the translation of cluster.clj, so I added comments and will upmerge when cluster.clj is finished.


abhishekagarwal87:
PR is indeed quite large. StatsUtil file is too large right now. I think while we are porting from clojure, we should also shift from functional paradigm to Object oriented paradigm. In that spirit, many utility functions can go directly inside the classes on which they are invoked upon. e.g.
```
public static void emittedTuple(CommonStats stats, String stream) {
        stats.getEmitted().incBy(stream, stats.rate);
    }
```
can go to CommonStats itself
```
public void emittedTuple(String stream) {
        getEmitted().incBy(stream, rate);
    }
```

Similarly `public static Map renderStats(SpoutExecutorStats stats)` can go to SpoutExecutorStats class itself. These are few examples I could think of in a quick look. But I hope you got the gist of it. It would be good to get the opinion of the @revans2  and and other committers on this as well.

unsleepy22:
@abhishekagarwal87 thanks, good point! I'll address your comments~

unsleepy22:
I've upmerged and added get-last-error, also moved renderStats & update tuple related methods to corresponding classes.

unsleepy22:
@revans2 could you please help review this PR?

storm-core/src/jvm/org/apache/storm/stats/BoltExecutorStats.java
```diff
@@ -0,0 +1,107 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.storm.stats;
+
+import clojure.lang.PersistentVector;
```
revans2: We should not be adding in new java code that depends on clojure.  We want to start using the thrift object where ever possible in the java code, and pushing the layer that translates them to to/from clojure up until it is no longer needed.

storm-core/src/jvm/org/apache/storm/stats/BoltExecutorStats.java
```diff
@@ -0,0 +1,107 @@
...
+    public MultiLatencyStatAndMetric getProcessLatencies() {
+        return (MultiLatencyStatAndMetric) this.get(PROCESS_LATENCIES);
+    }
+
+    public MultiLatencyStatAndMetric getExecuteLatencies() {
+        return (MultiLatencyStatAndMetric) this.get(EXECUTE_LATENCIES);
+    }
+
+    public void boltExecuteTuple(String component, String stream, long latencyMs) {
+        Object key = PersistentVector.create(component, stream);
```
revans2: In this case we can probably just create a new List and use that as the key.  Any List will probably do.

storm-core/src/clj/org/apache/storm/daemon/executor.clj
```diff
@@ -257,8 +257,8 @@
      :batch-transfer-queue batch-transfer->worker
      :transfer-fn (mk-executor-transfer-fn batch-transfer->worker storm-conf)
      :suicide-fn (:suicide-fn worker)
-     :storm-cluster-state (ClusterUtils/mkStormClusterState (:state-store worker) (Utils/getWorkerACL storm-conf)
-                            (ClusterStateContext. DaemonType/WORKER))
+      :storm-cluster-state (ClusterUtils/mkStormClusterState (:state-store worker) (Utils/getWorkerACL storm-conf)
+                             (ClusterStateContext. DaemonType/WORKER))
```
revans2: Indentation is off here.

storm-core/src/clj/org/apache/storm/daemon/nimbus.clj
```diff
@@ -916,7 +916,7 @@
         storm-cluster-state (:storm-cluster-state nimbus)
         ^INimbus inimbus (:inimbus nimbus)
         ;; read all the topologies
-        topology-ids (.activeStorms storm-cluster-state)
+          topology-ids (.activeStorms storm-cluster-state)
```
revans2: Indentation is off

storm-core/src/clj/org/apache/storm/daemon/nimbus.clj
```diff
@@ -1800,7 +1800,7 @@
               storm-name (topology-conf TOPOLOGY-NAME)
               _ (check-authorization! nimbus storm-name topology-conf "getLogConfig")
              storm-cluster-state (:storm-cluster-state nimbus)
-             log-config (.topologyLogConfig storm-cluster-state id nil)]
+              log-config (.topologyLogConfig storm-cluster-state id nil)]
```
revans2: Indentation is off

storm-core/src/clj/org/apache/storm/daemon/supervisor.clj
```diff
@@ -80,7 +79,6 @@
           new-profiler-actions
           (->>
             (dofor [sid (distinct storm-ids)]
-
```
revans2: Changes in this file are not needed.  Please revert them.

storm-core/src/clj/org/apache/storm/ui/core.clj
```diff
@@ -1215,7 +1222,7 @@
            (json-response {"status" "ok"
                            "id" host-port}
                           (m "callback")))))
-
+
```
revans2: White space change not needed

storm-core/src/clj/org/apache/storm/ui/core.clj
```diff
@@ -1231,7 +1238,7 @@
...
```
revans2: White space change not needed

storm-core/src/jvm/org/apache/storm/stats/CommonStats.java
```diff
@@ -0,0 +1,105 @@
...
+public class CommonStats {
+    public static final int NUM_STAT_BUCKETS = 20;
+
+    public static final String RATE = "rate";
+
+    public static final String EMITTED = "emitted";
+    public static final String TRANSFERRED = "transferred";
+    public static final String[] COMMON_FIELDS = {EMITTED, TRANSFERRED};
+
+    protected int rate;
```
revans2: I would prefer this to be a final int and set in the constructor.  Changing it in the middle of operation would require a coordinated effort with external code that does the sampling.

storm-core/src/jvm/org/apache/storm/stats/CommonStats.java
```diff
@@ -0,0 +1,105 @@
...
+    public CommonStats() {
+        put(EMITTED, new MultiCountStatAndMetric(NUM_STAT_BUCKETS));
+        put(TRANSFERRED, new MultiCountStatAndMetric(NUM_STAT_BUCKETS));
+    }
+
+    public int getRate() {
+        return this.rate;
+    }
+
+    public void setRate(int rate) {
```
revans2: This should be removed.

storm-core/src/jvm/org/apache/storm/stats/CommonStats.java
```diff
@@ -0,0 +1,105 @@
...
+
+    public void emittedTuple(String stream) {
+        this.getEmitted().incBy(stream, this.rate);
+    }
+
+    public void transferredTuples(String stream, int amount) {
+        this.getTransferred().incBy(stream, this.rate * amount);
+    }
+
+    protected void cleanupStats() {
```
revans2: This could be updated to just go off of all of the keys in metricMap.  There would be no need for subclasses to override this.

storm-core/src/jvm/org/apache/storm/stats/SpoutExecutorStats.java
```diff
@@ -0,0 +1,84 @@
...
+    public void spoutFailedTuple(String stream, long latencyMs) {
+        this.getFailed().incBy(stream, this.rate);
+    }
+
+    public Map renderStats() {
+        cleanupStats();
+        Map ret = new HashMap();
+        ret.putAll(valueStats(CommonStats.COMMON_FIELDS));
+        ret.putAll(valueStats(SpoutExecutorStats.SPOUT_FIELDS));
+        StatsUtil.putRawKV(ret, StatsUtil.TYPE, StatsUtil.KW_SPOUT);
```
revans2: Again we cannot have new java code depending on clojure.  The return of this is not touched except in thriftify-zk-worker-hb, which is trying to turn it into a thrift object anyways.  Lets just return the thrift ExecutorStats object we ultimately care about, and update the thriftify code to deal with it.

storm-core/src/jvm/org/apache/storm/stats/SpoutExecutorStats.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    public void cleanupStats() {
+        super.cleanupStats();
+        for (String field : SpoutExecutorStats.SPOUT_FIELDS) {
+            cleanupStat(this.get(field));
+        }
+    }
+
+    public static SpoutExecutorStats mkSpoutStats(int rate) {
```
revans2: Can we just have the constructor have the rate, and remove this entirely?

storm-core/src/jvm/org/apache/storm/stats/BoltExecutorStats.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    public void cleanupStats() {
+        super.cleanupStats();
+        for (String field : BOLT_FIELDS) {
+            cleanupStat(this.get(field));
+        }
+    }
+
+    public static BoltExecutorStats mkBoltStats(int rate) {
```
revans2: Similar here the constructor should take the rate and this should be removed.

storm-core/src/jvm/org/apache/storm/stats/BoltExecutorStats.java
```diff
@@ -0,0 +1,107 @@
...
+        this.getFailed().incBy(key, this.rate);
+
+    }
+
+    public Map renderStats() {
+        cleanupStats();
+        Map ret = new HashMap();
+        ret.putAll(valueStats(CommonStats.COMMON_FIELDS));
+        ret.putAll(valueStats(BoltExecutorStats.BOLT_FIELDS));
+        StatsUtil.putRawKV(ret, StatsUtil.TYPE, StatsUtil.KW_BOLT);
```
revans2: And similar here too.  This should return an ExecutorStats, not a Map with clojure specific things in it.

unsleepy22:
Thanks @revans2 , I'll address all your comments~

storm-core/src/jvm/org/apache/storm/stats/BoltExecutorStats.java
unsleepy22: @revans2 I've changed all the others according to your comments, my only problem is renderStats, do you mean that I only need to maintain ExecutorStats in memory without any stats Maps while running? Although renderStats is only used in thriftify-zk-worker-hb, when workers do heartbeats(see worker/do-executor-heartbeats method), it will use a clojure PersistentMap data structure, which messes with ExecutorStats, should I turn it into ExecutorStats too?

revans2:
I started testing this and started to run into some issues.

I created a new test, but the test is crap, because it takes way too long to run, but does give some good results.

https://github.com/apache/storm/compare/master...revans2:stats-test?expand=1

I plan to expand this more, but for now I am running into the following exception.

```
java.lang.ClassCastException: java.lang.Long cannot be cast to java.util.Map
        at org.apache.storm.stats.StatsUtil.filterSysStreams(StatsUtil.java:1343)
        at org.apache.storm.stats.StatsUtil.aggPreMergeTopoPageSpout(StatsUtil.java:314)
        at org.apache.storm.stats.StatsUtil.aggTopoExecStats(StatsUtil.java:481)
        at org.apache.storm.stats.StatsUtil.aggregateTopoStats(StatsUtil.java:570)
        at org.apache.storm.stats.StatsUtil.aggTopoExecsStats(StatsUtil.java:550)
        at org.apache.storm.daemon.nimbus$fn__5066$exec_fn__884__auto__$reify__5095.getTopologyPageInfo(nimbus.clj:2117)
        at sun.reflect.GeneratedMethodAccessor104.invoke(Unknown Source)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:483)
        at clojure.lang.Reflector.invokeMatchingMethod(Reflector.java:93)
        at clojure.lang.Reflector.invokeInstanceMethod(Reflector.java:28)
        at org.apache.storm.LocalCluster$_getTopologyPageInfo.invoke(LocalCluster.clj:110)
        at org.apache.storm.LocalCluster.getTopologyPageInfo(Unknown Source)
        at org.apache.storm.TestStats.testBasic(TestStats.java:150)
```

unsleepy22:
@revans2 yes I see the exception too when running your test, but looks odd because I've tested all storm ui pages in distributed mode with java stats and everything looks good. The exception is as follows, StatsUtil.filterSysStreams expects a map of {window -> stream id -> value}, while in your test it gets a map of {stream id -> value}. I suppose this happens in local mode only? Though I can manage to make through your test by changing this method a little.

revans2:
@unsleepy22 thanks for humoring me with the test.  That is really really odd.  I have no idea why it is doing this in local mode differently.  This is one advantage of strongly typed languages.

I'll try to spend some more time today to look through the code again.

storm-core/src/jvm/org/apache/storm/stats/BoltExecutorStats.java
revans2: `CommonStats.renderStats` (this function) was originally `stats/render-stats!` it is only called from executor.clj as the implementation of the `render-stats` method in an instance of `RunningExecutor` that it creates.

https://github.com/apache/storm/blob/cc1f6d77962ec96c69cf5af85e4016a9e08ccdcc/storm-core/src/clj/org/apache/storm/daemon/executor.clj#L408-L409

This in turn is only called by the `do-executor-heartbeat` function in worker.clj

https://github.com/apache/storm/blob/cc1f6d77962ec96c69cf5af85e4016a9e08ccdcc/storm-core/src/clj/org/apache/storm/daemon/worker.clj#L65-L81

The code there takes all of the executors from the worker calls `render-stats` on them, puts them into a map with the key being the executor id and adds that map into `zk-hb` under the `:executor-stats` keyword.

`zk-hb` is passed into `thriftify-zk-worker-hb` which pulls out `:executor-stats` filters out any executor that does not have stats and passes it to `thriftify-stats`

https://github.com/apache/storm/blob/cc1f6d77962ec96c69cf5af85e4016a9e08ccdcc/storm-core/src/clj/org/apache/storm/converter.clj#L255

which turns the map keys into `ExecutorInfo` instances and the map values into `ExecutorStats` instances. 

Why go to a clojure specific data structure that no one reads except to turn it into the java class we actually care about.

If we modify `renderStats` to return an `ExecutorStats` instance.  Then modify `thriftify-stats` to not modify the values in the map, just the keys, we should be good to go.  We will also need to update nimbus_test.clj because it too calls into `thriftify-zk-worker-hb` in a test.

revans2:
I am going through StatsUtil.java and it is more of the same thing everywhere.  The code is a literal translation of the clojure code.  The input APIs in most cases are clojure specific, but many of these functions actually return thrift object already.  For example `stats/agg-comp-execs-stats` returns an instance of `ComponentPageInfo`.  I would prefer to see us have a shim that will translate the clojure maps, etc into java objects that would be a better java API/state when we translate the clojure code that is calling into it.

For example 
```
(defn agg-comp-execs-stats
  "Aggregate various executor statistics for a component from the given
  heartbeats."
  [exec->host+port
   task->component
   beats
   window
   include-sys?
   topology-id
   topology
   component-id]

```

was translated into

```
public static TopologyPageInfo aggTopoExecsStats(
  String topologyId,
  Map exec2nodePort,
  Map task2component,
  Map beats,
  StormTopology topology,
  String window,
  boolean includeSys,
  IStormClusterState clusterState);
```

but if we expand the signature a bit more it really looks like

```
public static TopologyPageInfo aggTopoExecsStats(
  String topologyId,
  Map<List<Number>/*low exec, high exec*/,List<Object>/*String host, Number port*/ > exec2nodePort,
  Map<Integer, String> task2component,
  Map beats<List<Number>/*low exec, high exec*/, Map<Keyword, Object>/*Clojure map keys :stats, :uptime, :time-secs, good luck*/ >,
  StormTopology topology,
  String window,
  boolean includeSys,
  IStormClusterState clusterState);

```

If we want a more java like API with classes instead of maps/Lists it should probably be something more like

```
public class HostPort {
    public final String host;
    public final int port;
    ...
};

public static TopologyPageInfo aggTopoExecsStats(
    String topologyId, 
    Map<ExecutorInfo, HostPort> exec2nodePort,
    Map<Integer, String> task2component,
    Map beats<ExecutorInfo, ExecutorBeat>, 
    StormTopology topology,
    String window,
    boolean includeSys,
    IStormClusterState clusterState);
```

All of the public facing APIs need this type of a translation to happen.  If you want me to go through all of them and list what I think they should be I can.

unsleepy22:
@revans2 I get your idea and you don't have to go through all of them, my initial thought was to keep it compatible with original clojure data structures, but I'm ok to change all of them, I will update in a day or two~

revans2:
@unsleepy22 I understand the desire to keep them compatible, and that is a decent first step, but one of the goals of this translation is to remove the clojure entirely as a dependency for storm-core.  If we have java code still using it we cannot do that.

unsleepy22:
@revans2 I've changed according to your opinion, still, the code in StatsUtil might be quite rough and clojure-way, I just want to make sure it's the right move, hope you can take a quick look.
My changes are:
1. changed executor heartbeat structure to java HashMap instead of clojure map (and yes there's still a Map anyway)
2. moved update-heartbeat-cache method into StatsUtil.java, because of the change to clojure structure to HashMap, they should be moved back when translating nimbus.clj, I added TODO in methods.
3. because the HashMap structure is quite mixed, e.g., it may store "rate" of stats which is a double, and store "emitted" which is a HashMap, so in many methods, I used Map<String, Object>.
4. for structures like "executor->host+port", I still used the original clojure structure because I think they might be used somewhere else, I only want to change stats related with this PR 

BTW. about your test last time, it's actually a bug of my code, I fixed this, sorry for the confusion.

revans2:
@unsleepy22 It looks a lot better.  I don't care as much about the exact API.  we can change that in the future the important part is removing the dependency on clojure.  I have one minor issue, but overall it looks good and I am +1 for merging it in.

StatsUtils.java 2045-2046: in weightAvgAndSum
The reason for having the productOr0 function is because the inputs could be null, and if they were null we didn't want to blow up with an NPE.  The current code will throw an NPE.  it should probably look more like.

```
ret += productOr0(entry.getValue(), id2num.get(k));
```

I plan to do a bit of testing myself, but great work.

unsleepy22:
@revans2 you're right about the possible NPE, I've updated code.
Also I added a lot of method comments.
I have tested web ui with new code and compared stats data and all looks good.

unsleepy22:
@revans2 any update on this?

revans2:
@unsleepy22 sorry I had a minor personal emergency come up last Friday.  It is fixed now, but I am still trying to catch up on everything that happened Friday and over the weekend.  I'll try to get back to looking at this soon.  If I don't say anything in a day or so please ping me again.

revans2:
+1

PR #1149, opened 2016-02-24
[STORM-1574] Better handle backpressure thread exception & clear dir.

zhuoliu:
1. Better handle exception.
2. Remove topo dir during kill.
3. Add a unit test.


hustfxj:
+1  It looks good to me

hustfxj:
Some Tests failed. It prompts tests can't find remove-backpressure!. I guess it generate some questions when it merge with master.

storm-core/test/jvm/org/apache/storm/utils/WorkerBackpressureThreadTest.java
```diff
@@ -0,0 +1,68 @@
...
+            @Override
+            public void onEvent(Object obj) {
+                throw new RuntimeException();
+            }
+        };
+        WorkerBackpressureThread workerBackpressureThread = new WorkerBackpressureThread(trigger, workerData, callback);
+        workerBackpressureThread.start();
+        Thread.sleep(100);
+        WorkerBackpressureThread.notifyBackpressureChecker(trigger);
+        Thread.sleep(100);
```
wuchong: The CI failed maybe because the sleep time is too short.

I set the sleep time > 500 ,  and the workerBackpressureThread is dead as expect.

wuchong:
+1

storm-core/src/jvm/org/apache/storm/utils/WorkerBackpressureThread.java
```diff
@@ -43,17 +48,33 @@ static public void notifyBackpressureChecker(Object trigger) {
...
 }
 
+class BackpressureUncaughtExceptionHandler implements Thread.UncaughtExceptionHandler {
+    private static final Logger LOG = LoggerFactory.getLogger(BackpressureUncaughtExceptionHandler.class);
+    @Override
+    public void uncaughtException(Thread t, Throwable e) {
+        try {
+            Utils.handleUncaughtException(e);
+        } catch (Error error) {
+            LOG.info("Received error in WorkerBackpressureThread.. terminating the worker...");
```
abhishekagarwal87: This should be LOG.error and error can be passed as second argument.

storm-core/src/jvm/org/apache/storm/utils/WorkerBackpressureThread.java
```diff
@@ -43,17 +48,33 @@ static public void notifyBackpressureChecker(Object trigger) {
         }
     }
 
+    public void terminate() {
+        running = false;
+    }
```
revans2: Could we do an interrupt and a join after this?

revans2: We could then ignore the interrupted exception in run as it would just be used as a trigger to wake up the running thread.

storm-core/test/jvm/org/apache/storm/utils/WorkerBackpressureThreadTest.java
revans2: I really hate hard coded sleeps in tests.  Could we please replace this with something like

```
        workerBackpressureThread.start();
        long start = System.currentTimeMillis();
        while (!workerBackpressureThread.isAlive()) {
            assertTrue("Timeout", (System.currentTimeMillis() - start) < 1000);
            Thread.sleep(100);
        }
        WorkerBackpressureThread.notifyBackpressureChecker(trigger);
        long start = System.currentTimeMillis();
        while (workerBackpressureThread.isAlive()) {
            assertTrue("Timeout", (System.currentTimeMillis() - start) < 1000);
            Thread.sleep(100);
        }
```

storm-core/src/jvm/org/apache/storm/utils/WorkerBackpressureThread.java
```diff
@@ -43,17 +48,33 @@ static public void notifyBackpressureChecker(Object trigger) {
...
         }
     }
 }
 
+class BackpressureUncaughtExceptionHandler implements Thread.UncaughtExceptionHandler {
+    private static final Logger LOG = LoggerFactory.getLogger(BackpressureUncaughtExceptionHandler.class);
+    @Override
+    public void uncaughtException(Thread t, Throwable e) {
+        try {
+            Utils.handleUncaughtException(e);
```
revans2: This ignores all RuntimeExceptions.  This is fine for what it was originally written for, which was around thread pools.  But in this case any RuntimeException is going to still cause the backpressure thread to exit.

@kishorvpatil would you be OK if we renamed handeUncaughtException to be more descriptive, because it is deceptive right now, and I think will cause bug while we translate code to java.  Can we call it eatRuntimeExitOnOOM.

zhuoliu:
@revans2 Updated the exception handling.
Filed a JIRA for the Utils.uncaughtExceptionHandler renaming here:
https://issues.apache.org/jira/browse/STORM-1577

revans2:
The changes look good, but I would like to see the tests updated still.  I am not sure if they will still pass without some changes.  Specifically around the uncaught exception handler.  It would be nice to have an API so we can mock that out.

zhuoliu:
Since we will ignore all exception with onEvent in worker.clj, we should not need to test RuntimeException here. So delete it.

revans2:
I still would like to see the hard coded sleeps replaced by polling with a timeout.

storm-core/src/clj/org/apache/storm/daemon/worker.clj
```diff
@@ -155,7 +155,10 @@
         ;; update the worker's backpressure flag to zookeeper only when it has changed
         (log-debug "BP " @(:backpressure worker) " WAS " prev-backpressure-flag)
         (when (not= prev-backpressure-flag @(:backpressure worker))
-          (.workerBackpressure storm-cluster-state storm-id assignment-id port @(:backpressure worker)))
+          (try
+            (.workerBackpressure storm-cluster-state storm-id assignment-id (long port) @(:backpressure worker))
```
zhuoliu: (:port worker) returns an Integer. Will cause an exception if not turning to long. See also Storm-1578.

zhuoliu:
@revans2 updated the test with timeout, and fixed the Cast exception (Int/Long) in Storm-1578, also manual test works fine.

zhuoliu:
Also, I need to put up another pull request for 1.x.

revans2:
+1

revans2:
@zhuoliu yes you will need a separate pull request for 1.x

kishorvpatil:
LGTM. +1

PR #1155, opened 2016-02-26
[Storm-1579] Fix NoSuchFileException when running tests in storm-core

unsleepy22:
The old code takes "storm.local.dir" property as logDir while changes in STORM-1552 used `ConfigUtils.getLogDir`, which doesn't take this property, causing the ultimate event log directory to be "/logs/xxxx" and resulting in NoSuchFileException.

Note that the only effective change is in `getLogDir` method, I just formatted the whole class.


unsleepy22:
There're test failures, I'll look into it first.

abhishekagarwal87:
I didn't see such error on my system while running the tests though

unsleepy22:
@abhishekagarwal87 they're in test-reports, when running tests, you'll get messages like "Exception: java.lang.ClassCastException thrown from the UncaughtExceptionHandler in thread "Thread-35-__eventlogger-executor[2 2]", by looking into the detailed test-reports, you'll get NoSuchFileException.

abhishekagarwal87:
ohk. I see. Once the commit is ready, you can squash the commits. Looks good to me otherwise.

unsleepy22:
My change was quite simple: just adding back the logic to check "storm.local.dir" property/conf, but there're 3 test failures in "org.apache.storm.supervisor-test" (all the others pass without exceptions), I've no idea why, can someone please help take a look?

PR #1156, opened 2016-02-26
[STORM-1270]  [STORM-1274] port drpc&LocalDRPC  to java 

hustfxj:
1. Port "org.apache.storm.daemon.drpc" to java at "org.apache.storm.daemon.DrpcServer".
2. Port "org.apache.storm.LocalDRPC" to java at "org.apache.storm.LocalDRPC".
3. I keep the "webapp" of drpc.clj , as it's hard to translate and we will move ring to other REST framework.


storm-core/src/jvm/org/apache/storm/daemon/DrpcServer.java
```diff
@@ -0,0 +1,390 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.servlet.Servlet;
+import java.security.Principal;
+import java.util.*;
+import java.util.concurrent.*;
+import java.util.concurrent.atomic.AtomicInteger;
+
+public class DrpcServer implements DistributedRPC.Iface, DistributedRPCInvocations.Iface, Shutdownable {
```
abhishekagarwal87: you can rename it to DRPCServer for consistency.

storm-core/src/jvm/org/apache/storm/daemon/DrpcServer.java
```diff
@@ -0,0 +1,390 @@
...
+                if (handlerServer != null) {
+                    handlerServer.stop();
+                } else {
+                    invokeServer.stop();
+                }
+            }
+        });
+        LOG.info("Starting Distributed RPC servers...");
+
+        LOG.info("Starting Distributed RPC servers...");
```
abhishekagarwal87: Duplicate log statement

storm-core/src/jvm/org/apache/storm/daemon/DrpcServer.java
```diff
@@ -0,0 +1,390 @@
...
+        LOG.debug("authorization class name: {} class: {} handler: {}", klassname, aznClass, authorizer);
+        return authorizer;
+    }
+
+    public Map getConf() {
+        return conf;
+    }
+
+    public static void main(String[] args) throws Exception {
+
```
abhishekagarwal87: There are extra lines in some places which you can remove.

storm-core/src/jvm/org/apache/storm/daemon/DrpcServer.java
```diff
@@ -0,0 +1,390 @@
...
+                }
+            });
+        }
+
+        // To be replaced by Common.StartMetricsReporters
+        List<PreparableReporter> reporters = MetricsUtils.getPreparableReporters(conf);
+        for (PreparableReporter reporter : reporters) {
+            reporter.prepare(new MetricRegistry(), conf);
+            reporter.start();
+            LOG.info("Started statistics report plugin...");
```
abhishekagarwal87: This will be logged multiple times or you can add the reporter class

storm-core/src/jvm/org/apache/storm/daemon/DrpcServer.java
```diff
@@ -0,0 +1,390 @@
...
+                @Override
+                public void execute(Server s) {
+                    UIHelpers.configSsl(s, httpsPort, httpsKsPath, httpsKsPassword, httpsKsType, httpsKeyPassword, httpsTsPath, httpsTsPassword, httpsTsType,
+                            httpsNeedClientAuth, httpsWantClientAuth);
+                    UIHelpers.configFilter(s, httpServlet, filterConfigurations);
+                }
+            });
+        }
+
+        // To be replaced by Common.StartMetricsReporters
```
abhishekagarwal87: you can add a TODO so that it is not missed

storm-core/src/jvm/org/apache/storm/daemon/DrpcServer.java
abhishekagarwal87: You can separate out the implementation of DistributedRPC.Iface, DistributedRPCInvocations.Iface, Shutdownable to a class DRPCService. This service class will only contain the business logic and server logic can be inside DRPCServer. Also AutoCloseable should be the interface than Shutdownable.

storm-core/src/jvm/org/apache/storm/daemon/DrpcServer.java
```diff
@@ -0,0 +1,390 @@
...
+        int newid = idinc % maxvalue;
+        if (idinc != newid) {
+            this.ctr.compareAndSet(idinc, newid);
+        }
+
+        String strid = String.valueOf(newid);
+        Semaphore sem = new Semaphore(0);
+
+        DRPCRequest req = new DRPCRequest(funcArgs, strid);
+        this.idtoStart.put(strid, Time.currentTimeSecs());
```
abhishekagarwal87: minor. can you rename idtoStart to idToStart and similarly other idto* variables.

storm-core/src/jvm/org/apache/storm/daemon/DrpcServer.java
```diff
@@ -0,0 +1,390 @@
...
+        try {
+            sem.acquire();
+        } catch (InterruptedException e) {
+            LOG.error("acquire fail ", e);
+        }
+        LOG.debug("Acquired for DRPC request for {} {} at {}", functionName, funcArgs, System.currentTimeMillis());
+
+        Object result = this.idtoResult.get(strid);
+
+        LOG.info("Returning for DRPC request for " + functionName + " " + funcArgs + " at " + (System.currentTimeMillis()));
```
abhishekagarwal87: this should be LOG.debug

storm-core/src/jvm/org/apache/storm/daemon/DrpcServer.java
```diff
@@ -0,0 +1,390 @@
...
+
+    // TO be replaced by Common.mkAuthorizationHandler
+    private IAuthorizer mkAuthorizationHandler(String klassname, Map conf) {
+        IAuthorizer authorizer = null;
+        Class aznClass = null;
+        if (StringUtils.isNotBlank(klassname)) {
+            try {
+                aznClass = Class.forName(klassname);
+                authorizer = (IAuthorizer) aznClass.newInstance();
+                if (authorizer != null) {
```
abhishekagarwal87: newInstance shouldn't return null

storm-core/src/jvm/org/apache/storm/daemon/DrpcServer.java
```diff
@@ -0,0 +1,390 @@
...
+
+    private AtomicInteger ctr = new AtomicInteger(0);
+    private ConcurrentHashMap<String, Semaphore> idtoSem = new ConcurrentHashMap<String, Semaphore>();
+    private ConcurrentHashMap<String, Object> idtoResult = new ConcurrentHashMap<String, Object>();
+    private ConcurrentHashMap<String, Integer> idtoStart = new ConcurrentHashMap<String, Integer>();
+    private ConcurrentHashMap<String, String> idtoFunction = new ConcurrentHashMap<String, String>();
+    private ConcurrentHashMap<String, DRPCRequest> idtoRequest = new ConcurrentHashMap<String, DRPCRequest>();
+    private ConcurrentHashMap<String, ConcurrentLinkedQueue<DRPCRequest>> requestQueues = new ConcurrentHashMap<String, ConcurrentLinkedQueue<DRPCRequest>>();
+
+    private final Meter meterHttpRequests = new MetricRegistry().meter("drpc:num-execute-http-requests");
```
abhishekagarwal87: this variable is not used

storm-core/src/jvm/org/apache/storm/daemon/DrpcServer.java
```diff
@@ -0,0 +1,390 @@
...
+        Integer drpcHttpPort = (Integer) conf.get(Config.DRPC_HTTP_PORT);
+        handlerServer = initHandlerServer(conf, this);
+        invokeServer = initInvokeServer(conf, this);
+        httpCredsHandler = AuthUtils.GetDrpcHttpCredentialsPlugin(conf);
+        Utils.addShutdownHookWithForceKillIn1Sec(new Runnable() {
+            @Override
+            public void run() {
+                if (handlerServer != null) {
+                    handlerServer.stop();
+                } else {
```
abhishekagarwal87: shouldn't it try to stop both of them

examples/storm-starter/src/jvm/org/apache/storm/starter/ManualDRPC.java
```diff
@@ -30,39 +30,37 @@
...
-    System.out.println(drpc.execute("exclamation", "bbb"));
+        LocalCluster cluster = new LocalCluster();
+        Config conf = new Config();
+        cluster.submitTopology("exclaim", conf, builder.createTopology());
 
-  }
+        System.out.println(drpc.execute("exclamation", "aaa"));
+        System.out.println(drpc.execute("exclamation", "bbb"));
+    }
 }
```
revans2: Please don't change the spacing/formatting for files you didn't touch otherwise.

storm-core/src/clj/org/apache/storm/daemon/drpc.clj
```diff
@@ -210,67 +63,19 @@
...
-                                     https-ts-path
-                                     https-ts-password
-                                     https-ts-type
-                                     https-need-client-auth
-                                     https-want-client-auth)
-                                   (UIHelpers/configFilter server (ring.util.servlet/servlet app) filters-confs))))))
-      (start-metrics-reporters conf)
-      (when handler-server
-        (.serve handler-server)))))
+        (let [app (-> (webapp drpc-server http-creds-handler)
```
revans2: Can we please have a follow on JIRA to port the webapp portion over to java too?  Until it is done I don't consider DRPC moved to java.

storm-core/src/jvm/org/apache/storm/daemon/DrpcServer.java
```diff
@@ -0,0 +1,390 @@
...
+        if (idinc != newid) {
+            this.ctr.compareAndSet(idinc, newid);
+        }
+
+        String strid = String.valueOf(newid);
+        Semaphore sem = new Semaphore(0);
+
+        DRPCRequest req = new DRPCRequest(funcArgs, strid);
+        this.idtoStart.put(strid, Time.currentTimeSecs());
+        this.idtoSem.put(strid, sem);
```
abhishekagarwal87: To make code more clean, can we have methods like below
```
String registerRequest(functionName, functionArgs) - Creates a new id with the function name
awaitOnResult(id) - blocks on the semaphore associated with the id
notifyOnResult(id) - releases the semaphore associated with the id
updateResult(id, Object result) - update the result 
removeRequest(id) - Cleanup the request
```
The above four methods can go into a class named DRPCState/DRPCRequestRegistry. what do you think?

storm-core/src/jvm/org/apache/storm/daemon/DrpcServer.java
```diff
@@ -0,0 +1,390 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.storm.daemon;
+
+import com.codahale.metrics.Meter;
+import com.codahale.metrics.MetricRegistry;
+import com.sun.net.httpserver.HttpsServer;
+import com.sun.org.apache.bcel.internal.generic.ARRAYLENGTH;
```
revans2: We should not be using com.sun internal classes.  We didn't use them before we should not use them now.

storm-core/src/jvm/org/apache/storm/daemon/DrpcServer.java
abhishekagarwal87: All the semaphore/queue related logic will sit together in one class and not dispersed across.

storm-core/src/jvm/org/apache/storm/daemon/DrpcServer.java
```diff
@@ -0,0 +1,390 @@
...
+    // To be removed after porting drpc.clj
+    private Servlet httpServlet;
+
+    private AtomicInteger ctr = new AtomicInteger(0);
+    private ConcurrentHashMap<String, Semaphore> idtoSem = new ConcurrentHashMap<String, Semaphore>();
+    private ConcurrentHashMap<String, Object> idtoResult = new ConcurrentHashMap<String, Object>();
+    private ConcurrentHashMap<String, Integer> idtoStart = new ConcurrentHashMap<String, Integer>();
+    private ConcurrentHashMap<String, String> idtoFunction = new ConcurrentHashMap<String, String>();
+    private ConcurrentHashMap<String, DRPCRequest> idtoRequest = new ConcurrentHashMap<String, DRPCRequest>();
+    private ConcurrentHashMap<String, ConcurrentLinkedQueue<DRPCRequest>> requestQueues = new ConcurrentHashMap<String, ConcurrentLinkedQueue<DRPCRequest>>();
```
revans2: This is a perfect place to actually improve things to make them more object oriented.  The key to almost all of these maps is the same thing.  Why not remove all of the special ordering that we have to do on setup and cleanup.

```
private static class InternalRequest {
    public final Semaphore sem;
    public final long startTimeSecs;
    public final string function;
    public final DRPCRequest request;
    public volatile Object result;

    public InternalRequest(String function, DRPCRequest request) {
        sem = new Semaphore(0);
        startTimeSecs = Time.currentTimeSecs();
        this.function = function;
        this.request = request;
    }
}

private ConcurrentHashMap<String, InternalRequest> outstandingRequests = new ConcurrentHashMap<>();
```

storm-core/src/jvm/org/apache/storm/daemon/DrpcServer.java
```diff
@@ -0,0 +1,390 @@
...
+
+    private final Meter meterHttpRequests = new MetricRegistry().meter("drpc:num-execute-http-requests");
+    private final Meter meterExecuteCalls = new MetricRegistry().meter("drpc:num-execute-calls");
+    private final Meter meterResultCalls = new MetricRegistry().meter("drpc:num-result-calls");
+    private final Meter meterFailRequestCalls = new MetricRegistry().meter("drpc:num-failRequest-calls");
+    private final Meter meterFetchRequestCalls = new MetricRegistry().meter("drpc:num-fetchRequest-calls");
+    private final Meter meterShutdownCalls = new MetricRegistry().meter("drpc:num-shutdown-calls");
```
revans2: We cannot just create a new MetricRegistry each time.  The metrics will be lost.  In the clojure code there is a default registry that is used by all of the metrics.  We need a similar one everywhere for all of our code to use (including the clojure code)

storm-core/src/jvm/org/apache/storm/daemon/DrpcServer.java
```diff
@@ -0,0 +1,390 @@
...
+        return httpCredsHandler;
+    }
+
+    public void setHttpCredsHandler(IHttpCredentialsPlugin httpCredsHandler) {
+        this.httpCredsHandler = httpCredsHandler;
+    }
+
+    public Servlet getHttpServlet() {
+        return httpServlet;
+    }
```
revans2: This is never used and should be removed.

storm-core/src/jvm/org/apache/storm/daemon/DrpcServer.java
```diff
@@ -0,0 +1,390 @@
...
+
+    }
+
+    public IHttpCredentialsPlugin getHttpCredsHandler() {
+        return httpCredsHandler;
+    }
+
+    public void setHttpCredsHandler(IHttpCredentialsPlugin httpCredsHandler) {
+        this.httpCredsHandler = httpCredsHandler;
+    }
```
revans2: the set and get of HttpCredsHandler is never used.

storm-core/src/jvm/org/apache/storm/daemon/DrpcServer.java
```diff
@@ -0,0 +1,390 @@
...
+        }
+        LOG.debug("authorization class name: {} class: {} handler: {}", klassname, aznClass, authorizer);
+        return authorizer;
+    }
+
+    public Map getConf() {
+        return conf;
+    }
+
+    public static void main(String[] args) throws Exception {
```
revans2: This main does not work if we are using HTTP.  The servlet is not setup.  We should probably remove this until it actually works.

storm-core/src/jvm/org/apache/storm/daemon/DrpcServer.java
```diff
@@ -0,0 +1,390 @@
...
+                }
+            } catch (Exception e) {
+                LOG.error("mkAuthorizationHandler failed!", e);
+            }
+        }
+        LOG.debug("authorization class name: {} class: {} handler: {}", klassname, aznClass, authorizer);
+        return authorizer;
+    }
+
+    public Map getConf() {
```
revans2: This is never used.

storm-core/src/jvm/org/apache/storm/daemon/DrpcServer.java
```diff
@@ -0,0 +1,390 @@
...
+                return getTimeoutCheckSecs();
+            }
+        });
+    }
+
+    public Long getTimeoutCheckSecs() {
+        return timeoutCheckSecs;
+    }
+
+    public void launchServer(boolean isLocal, Map conf) throws Exception {
```
revans2: Why did we combine service-handler and launch-server! into a single launchServer method?  The only thing that was missing in local mode that this adds is the clearThread to time out missed tuples.  Did you run into issues with testing when this was not included?

revans2: Never mind the old code launched the clear-thread in service-handler.  To me it feels cleaner to have the Constructor act like service-handler and launchServer to start up the thrift servers and HTTP Server.

storm-core/src/jvm/org/apache/storm/daemon/DrpcServer.java
```diff
@@ -0,0 +1,390 @@
...
+        idtoResult.remove(id);
+        idtoStart.remove(id);
+        idtoFunction.remove(id);
+        idtoRequest.remove(id);
+    }
+
+    @Override
+    public String execute(String functionName, String funcArgs) throws DRPCExecutionException, AuthorizationException, org.apache.thrift.TException {
+        meterExecuteCalls.mark();
+        LOG.debug("Received DRPC request for {} {} at {} ", functionName, funcArgs, System.currentTimeMillis());
```
revans2: minor nit. There were parens around the funcArgs

storm-core/src/jvm/org/apache/storm/daemon/DrpcServer.java
```diff
@@ -0,0 +1,390 @@
...
+        Map<String, String> map = new HashMap<>();
+        map.put(DRPCAuthorizerBase.FUNCTION_NAME, functionName);
```
revans2: Does Guava have a way to create this map cleaner?

storm-core/src/jvm/org/apache/storm/daemon/DrpcServer.java
```diff
@@ -0,0 +1,390 @@
...
+        checkAuthorization(authorizer, map, "execute");
+
+        int idinc = this.ctr.incrementAndGet();
+        int maxvalue = 1000000000;
+        int newid = idinc % maxvalue;
+        if (idinc != newid) {
+            this.ctr.compareAndSet(idinc, newid);
+        }
```
revans2: This does not actually do the same thing.  what happens if the compareAndSet fails?  If you want it the same we have to loop around and try again until it succeeds.

```
int newid = 0;
do {
  int orig = ctr.get();
  newid = (orig + 1) % 1000000000;
} while (!ctr.compareAndSet(orig, newid));
String strid = String.valueOf(newid);
```

storm-core/src/jvm/org/apache/storm/daemon/DrpcServer.java
abhishekagarwal87: ImmutableMap.of(DRPCAuthorizerBase.FUNCTION_NAME, functionName)

storm-core/src/jvm/org/apache/storm/daemon/DrpcServer.java
```diff
@@ -0,0 +1,390 @@
...
+
+        this.cleanup(strid);
+
+        if (result instanceof DRPCExecutionException) {
+            throw (DRPCExecutionException) result;
+        }
+        if (result == null) {
+            throw new DRPCExecutionException("Request timed out");
+        }
+        return String.valueOf(result);
```
revans2: Just cast result to a String, if it is not a String we have an internal error that we need to fix.

Or better yet, in that class I was suggesting we create.  Instead of having an Object for the result lets split it into an Exception and a String result.

storm-core/src/jvm/org/apache/storm/daemon/DrpcServer.java
```diff
@@ -0,0 +1,390 @@
...
+                sem.release();
+            }
+        }
+    }
+
+    protected ConcurrentLinkedQueue<DRPCRequest> acquireQueue(String function) {
+        ConcurrentLinkedQueue<DRPCRequest> reqQueue = requestQueues.get(function);
+        if (reqQueue == null) {
+            reqQueue = new ConcurrentLinkedQueue<DRPCRequest>();
+            requestQueues.put(function, reqQueue);
```
revans2: This should be a putIfAbsent.

```
protected ConcurrentLinkedQueue<DRPCRequest> acquireQueue(String function) {
  ConcurrentLinkedQueue<DRPCRequest> reqQueue = requestQueues.get(function);
  if (reqQueue == null) {
    reqQueue = new ConcurrentLinkedQueue<>();
    ConcurrentLinkedQueue<DRPCRequest> old = requestQueues.putIfAbsent(function, reqQueue);
    if (old != null) {
      reqQueue = old;
    }
  }
  return reqQueue;
}
```

storm-core/src/jvm/org/apache/storm/daemon/DrpcServer.java
```diff
@@ -0,0 +1,390 @@
...
+                throw new AuthorizationException("DRPC request '" + operation + "' for '" + user + "' user is not authorized");
+            }
+        }
+    }
+
+    private void checkAuthorization(IAuthorizer aclHandler, Map mapping, String operation) throws AuthorizationException {
+        checkAuthorization(aclHandler, mapping, operation, ReqContext.context());
+    }
+
+    // TO be replaced by Common.mkAuthorizationHandler
```
revans2: Then lets create the file and just put it in there now, so the next person does not have to do it later.

revans2:
Done with my first pass through the code.

storm-core/src/jvm/org/apache/storm/daemon/DrpcServer.java
hustfxj: I think we can rename DRPCServer to DRPC later

storm-core/src/jvm/org/apache/storm/daemon/DrpcServer.java
hustfxj: meterHttpRequests is not used as drpc.clj has meter for httpRequests. But we can use this meter after we remove drpc.clj

examples/storm-starter/src/jvm/org/apache/storm/starter/ManualDRPC.java
hustfxj: I will pay attention to this，Thank you for reminding me

storm-core/src/clj/org/apache/storm/daemon/drpc.clj
hustfxj: ok

storm-core/src/jvm/org/apache/storm/daemon/DrpcServer.java
hustfxj: @abhishekagarwal87  DRPC have already result fetchRequest execute failRequest， and these are override. I don't want to add some other methods.

storm-core/src/jvm/org/apache/storm/daemon/DrpcServer.java
hustfxj: perfect   thank you

storm-core/src/jvm/org/apache/storm/daemon/DrpcServer.java
hustfxj: Can we  follow up a JIRA to port a default registry over to java too? I will do it.

storm-core/src/jvm/org/apache/storm/daemon/DrpcServer.java
abhishekagarwal87: You can add them in the InternalRequest class suggested by Robert.

storm-core/src/jvm/org/apache/storm/daemon/DrpcServer.java
hustfxj: common.clj maybe submit soon

hustfxj:
@revans2 @abhishekagarwal87 Than you ver much. I are addressing your comments. But I may not understand what you mean about launchServer method, although I update the method. Can you help me review again?  I want to follow up a JIRA to port the webapp portion over to java about webapp,
And I want to follow up a JIRA to port a default registry over to java for all of our code to use;

storm-core/src/jvm/org/apache/storm/LocalDRPC.java
```diff
@@ -0,0 +1,80 @@
...
+import org.apache.storm.utils.Utils;
+import org.apache.thrift.TException;
+
+import java.util.Map;
+
+public class LocalDRPC implements ILocalDRPC {
+    private static final Logger LOG = Logger.getLogger(LocalDRPC.class);
+
+    private DrpcServer handler = new DrpcServer();
+    private Thread thread;
```
revans2: This is never used and should be removed.

storm-core/src/jvm/org/apache/storm/daemon/DrpcServer.java
```diff
@@ -0,0 +1,383 @@
...
+
+    private ThriftServer handlerServer;
+    private ThriftServer invokeServer;
+    private IHttpCredentialsPlugin httpCredsHandler;
+
+    private Thread clearThread;
+
+    private IAuthorizer authorizer;
+
+    //TODO: To be removed after porting drpc.clj
```
revans2: Just delete the comment, the JIRA you filed should cover this.

storm-core/src/jvm/org/apache/storm/daemon/DrpcServer.java
```diff
@@ -0,0 +1,383 @@
...
+            sem = new Semaphore(0);
+            startTimeSecs = Time.currentTimeSecs();
+            this.function = function;
+            this.request = request;
+        }
+    }
+    private ConcurrentHashMap<String, InternalRequest> outstandingRequests = new ConcurrentHashMap<>();
+
+
+    //TODO: to be replaced by a common registry
```
revans2: I don't want to merge in code with broken functionality.  Lets pull in the metrics changes you made first before we pull this JIRA in.

storm-core/src/jvm/org/apache/storm/daemon/DrpcServer.java
```diff
@@ -0,0 +1,383 @@
...
+    private final static Meter meterResultCalls = new MetricRegistry().meter("drpc:num-result-calls");
+    private final static Meter meterFailRequestCalls = new MetricRegistry().meter("drpc:num-failRequest-calls");
+    private final static Meter meterFetchRequestCalls = new MetricRegistry().meter("drpc:num-fetchRequest-calls");
+    private final static Meter meterShutdownCalls = new MetricRegistry().meter("drpc:num-shutdown-calls");
+    
+    public DrpcServer() {
+
+    }
+
+    //TODO: to be removed
```
revans2: Please remove this comment.

storm-core/src/jvm/org/apache/storm/daemon/DrpcServer.java
```diff
@@ -0,0 +1,383 @@
...
+    //TODO: to be replaced by a common registry
+    private final static Meter meterHttpRequests = new MetricRegistry().meter("drpc:num-execute-http-requests");
+    private final static Meter meterExecuteCalls = new MetricRegistry().meter("drpc:num-execute-calls");
+    private final static Meter meterResultCalls = new MetricRegistry().meter("drpc:num-result-calls");
+    private final static Meter meterFailRequestCalls = new MetricRegistry().meter("drpc:num-failRequest-calls");
+    private final static Meter meterFetchRequestCalls = new MetricRegistry().meter("drpc:num-fetchRequest-calls");
+    private final static Meter meterShutdownCalls = new MetricRegistry().meter("drpc:num-shutdown-calls");
+    
+    public DrpcServer() {
+
```
revans2: In the original code clear-thread was started in the service-handler (which is the equivalent of the constructor).  I would like to see it started here too, but that will require the constructor to take the Config.

storm-core/src/jvm/org/apache/storm/daemon/DrpcServer.java
```diff
@@ -0,0 +1,383 @@
...
+    }
+
+    private void initClearThread() {
+        clearThread = Utils.asyncLoop(new Callable() {
+
+            @Override
+            public Object call() throws Exception {
+                for (Map.Entry<String, InternalRequest> e : outstandingRequests.entrySet()) {
+                    InternalRequest internalRequest = e.getValue();
+                    if (Time.deltaSecs(internalRequest.startTimeSecs) > Utils.getInt(conf.get(Config.DRPC_REQUEST_TIMEOUT_SECS), 0)) {
```
revans2: There was no default of 0 in the original code and we should probably not have one here either.

storm-core/src/jvm/org/apache/storm/daemon/DrpcServer.java
```diff
@@ -0,0 +1,383 @@
...
+                        String id = e.getKey();
+                        Semaphore sem = internalRequest.sem;
+                        if (sem != null) {
+                            String func = internalRequest.function;
+                            acquireQueue(func).remove(internalRequest.request);
+                            LOG.warn("Timeout DRPC request id: {} start at {}", id, e.getValue());
+                            sem.release();
+                        }
+                        cleanup(id);
+                        LOG.info("Clear request " + id);
```
revans2: This log did not exist in the original code.

storm-core/src/jvm/org/apache/storm/daemon/DrpcServer.java
revans2: Yes it should the original code was more of

```
if (handlerServer != null) handlerServer.stop();
invokeServer.stop();
```

storm-core/src/jvm/org/apache/storm/daemon/DrpcServer.java
```diff
@@ -0,0 +1,383 @@
...
+        LOG.info("Starting Distributed RPC servers...");
+        new Thread(new Runnable() {
+
+            @Override
+            public void run() {
+                invokeServer.serve();
+            }
+        }).start();
+
+        //TODO: To be replaced by Common.StartMetricsReporters
```
revans2: This should probably be moved into #1171 and just do it there.

revans2: That way it can use the same MetricsRegistry as the others, which is the key to make this all work.

storm-core/src/jvm/org/apache/storm/daemon/DrpcServer.java
revans2: If we move the initClearThread to be called from the constructor, and have the constructor take a conf we can then remove the need for any parameters top be passed in here, and simply have localmode not call this method.

storm-core/src/jvm/org/apache/storm/daemon/DrpcServer.java
```diff
@@ -0,0 +1,383 @@
...
+            throw (DRPCExecutionException)result;
+        }
+        if (result == null) {
+            throw new DRPCExecutionException("Request timed out");
+        }
+        try {
+            return String.valueOf(result);
+        }catch (Exception e){
+            throw new DRPCExecutionException(e.getMessage());
+        }
```
revans2: Lets replace this the entire try/catch with.
```
return (String) result;
```

If we ever have anything but a DRPCExecutorionException or a String in there we have an internal error that we need to fix.

storm-core/src/jvm/org/apache/storm/daemon/DrpcServer.java
revans2: This is still an issue to be fixed.

storm-core/test/clj/org/apache/storm/security/auth/auth_test.clj
```diff
@@ -27,6 +27,8 @@
   (:import [javax.security.auth Subject])
   (:import [java.net InetAddress])
   (:import [org.apache.storm Config])
+  (:import [org.mockito Mockito])
+  (:import [org.mockito.exceptions.base MockitoAssertionError])
```
revans2: Are these needed?  If not please revert.  If so how did this work before?

revans2:
Done with another pass

storm-core/src/jvm/org/apache/storm/daemon/DrpcServer.java
hustfxj: sorry. I make mistake

hustfxj:
@revans2  Thank you.I won't wait updating the metric codes util the #1171  about the metrics is merged master.

hustfxj:
@revans2 @abhishekagarwal87 could you take a look?

storm-core/src/clj/org/apache/storm/daemon/supervisor.clj
```diff
@@ -1297,7 +1297,7 @@
       (.readBlobTo blob-store (ConfigUtils/masterStormConfKey storm-id) (FileOutputStream. (ConfigUtils/supervisorStormConfPath tmproot)) nil)
       (finally
         (.shutdown blob-store)))
-    (FileUtils/moveDirectory (File. tmproot) (File. stormroot))
```
abhishekagarwal87: this file can be reverted.

storm-core/src/jvm/org/apache/storm/daemon/DrpcServer.java
```diff
@@ -0,0 +1,356 @@
...
+            reqQueue = new ConcurrentLinkedQueue<>();
+            ConcurrentLinkedQueue<DRPCRequest> old = requestQueues.putIfAbsent(function, reqQueue);
+            if (old != null) {
+                reqQueue = old;
+            }
+        }
+        return reqQueue;
+    }
+
+    private void checkAuthorization(IAuthorizer aclHandler, Map mapping, String operation, ReqContext reqContext) throws AuthorizationException {
```
abhishekagarwal87: This is not called in this class and has been marked private. So it can be removed I believe.

storm-core/src/jvm/org/apache/storm/daemon/DrpcServer.java
```diff
@@ -0,0 +1,356 @@
...
+    }
+
+    @Override
+    public void failRequest(String id) throws AuthorizationException, TException {
+        meterFailRequestCalls.mark();
+        InternalRequest internalRequest = this.outstandingRequests.get(id);
+        if (internalRequest != null) {
+            Map<String, String> map = new HashMap<>();
+            map.put(DRPCAuthorizerBase.FUNCTION_NAME, internalRequest.function);
+            checkAuthorization(authorizer, map, "failRequest");
```
abhishekagarwal87: It would be good to have operation names declared as public static variables.

abhishekagarwal87:
Thanks @hustfxj . Looks good to me.

storm-core/src/jvm/org/apache/storm/daemon/DrpcServer.java
revans2: It is called, on line 337
```
 private void checkAuthorization(IAuthorizer aclHandler, Map mapping, String operation) throws AuthorizationException {
    checkAuthorization(aclHandler, mapping, operation, ReqContext.context());
}
```

storm-core/src/jvm/org/apache/storm/daemon/DrpcServer.java
revans2: OK

revans2:
@hustfxj looks good to me I am +1 on this.  @abhishekagarwal87 are you OK with merging this in as is.  The comments you had looked relatively minor.

abhishekagarwal87:
Yes. +1

storm-core/src/jvm/org/apache/storm/daemon/DrpcServer.java
abhishekagarwal87: @revans2 I meant there is no need of overloading function as only one signature is being called. But its fine to merge still.

storm-core/src/jvm/org/apache/storm/daemon/DrpcServer.java
```diff
@@ -0,0 +1,356 @@
...
+            @Override
+            public void run() {
+                invokeServer.serve();
+            }
+        }).start();
+
+        StormMetricsRegistry.startMetricsReporters(conf);
+
+        if (handlerServer != null)
+            handlerServer.serve();
```
revans2: Sorry I missed this before.  In the previous code this was the very last thing that happened on the main thread, because this is a blocking call.  With this happening before initHttp the HTTP server is never initialized even when it is configured.

revans2: Looks like you can just switch the order of doing the init and put initHttp() first in launchServer, but I am getting some problems with using HTTP with the following.

```
$ cat ./conf/storm.yaml
drpc.http.port: 9090
drpc.servers:
    - "localhost"
$ ./bin/storm jar ./examples/storm-starter/storm-starter-topologies-2.0.0-SNAPSHOT.jar org.apache.storm.starter.BasicDRPCTopology drpc-test
... wait for to come all the way up
$ curl http://localhost:9090/exclamation/hi
... Error ...
```

from the logs I see
```
2016-03-08 15:17:49.528 o.a.s.s.o.e.j.s.ServletHandler [WARN] /exclamation/hi
java.lang.NullPointerException: Handler returned nil
        at org.apache.storm.shade.ring.util.servlet$make_service_method$fn__7738.invoke(servlet.clj:128)
        at org.apache.storm.shade.ring.util.servlet$servlet$fn__7742.invoke(servlet.clj:136)
        at org.apache.storm.shade.ring.util.servlet.proxy$javax.servlet.http.HttpServlet$ff19274a.service(Unknown Source)
        at org.apache.storm.shade.org.eclipse.jetty.servlet.ServletHolder.handle(ServletHolder.java:654)
        at org.apache.storm.shade.org.eclipse.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1320)
        at org.apache.storm.logging.filters.AccessLoggingFilter.handle(AccessLoggingFilter.java:47)
        at org.apache.storm.logging.filters.AccessLoggingFilter.doFilter(AccessLoggingFilter.java:39)
        at org.apache.storm.shade.org.eclipse.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1291)
        at org.apache.storm.shade.org.eclipse.jetty.servlets.CrossOriginFilter.handle(CrossOriginFilter.java:247)
```

revans2: OK So the issue with http not working is that the handler is never set for the http server to have access to.

hustfxj: @revans2  I can see the same errors when I run Drpc of master branch.  Later I find we maybe misunderstand it. As follows:
$ curl http://localhost:3774/drpc/exclamation/hi
hi!
It's ok. So the URL should like ""/drpc/:func/*"

hustfxj: @revans2  I have adjust the order of initHttp(). Thank you again.

revans2: Crap you are right, I totally messed up the URL.  Great work!

revans2:
+1

PR #1157, opened 2016-02-26
STORM-1581: Repair github links in storm documentation

abhishekagarwal87:


abhishekagarwal87:
It turned out to be more complex than I thought it to be. Many links in the documentation are already broken. some github paths point to a particular commit and some point to an external repository. Many github links will still be broken as the classes are being ported over to java. Though at least many links pointing to old backtype.storm package will get corrected with this PR.

dossett:
+1

harshach:
@abhishekagarwal87 can you upmerge this patch.

abhishekagarwal87:
@harshach will do

abhishekagarwal87:
@harshach This PR is moot now after the documentation changes @revans2 has made. @revans2 can confirm?

HeartSaVioR:
Release specific docs are in their branch, but other docs are only in svn.
I believe `how to expose other docs to contributors` is very important to lead active contributions, so we may have to discuss about this.

abhishekagarwal87:
@HeartSaVioR - At the moment, what will be the procedure of modifying a doc and publishing it? I am not familiar with storm site generation and hence this question.

revans2:
@abhishekagarwal87 yes I modified most of the docs so they are release specific and under the docs directory again.

I should add some meta documentation about the docs themselves (I'll try to take this and update the README.md in all of them.

In subversion release specific documentation is placed under the releases directory, in a sub directory named after the release.  The "current" release should be pointed to by a symlink in that directory called current.

http://svn.apache.org/viewvc/storm/site/releases/

Inside the git docs directory all of the markdown files are supposed to be release specific.

I added a jekyll plugin that automates generating data for and about the release pages.

https://github.com/apache/storm/blob/master/docs/_plugins/releases.rb

It sets three things for all of the release documentation pages.
version - the version number of the  
git-tree-base - a link to a directory in github that this version is on
git-blob-base - a link to to where on github that this version is on, but should be used when pointing to files.

It also sets/augments a dataset for each release.  `site.data.releases` 
each release in the list includes the following, and each can be set in `_data/releases.yml` to override what is automatically generated by the plugin.

git-tag-or-branch - tag or branch name on github/apache/storm
git-tree-base - a link to a directory in github that this version is on
git-blob-base - a link to to where on github that this version is on, but should be used when pointing to files.
base-name - name of the release files to download, without the .tar.gz
has-download - if this is an official release and a download link should be created.

Also I think I have updated all of the javadoc and github links to point to the proper place for each release.  So 1.0.0-SNAPSHOT is pointing to org.apache.storm, and 2.0.0-SNAPSHOT is updated to point to java files where they have moved.

abhishekagarwal87:
Thanks @revans2 for detailed explanation. I checked randomly in few places and the links look good. closing this PR.

PR #1160, opened 2016-02-27
STORM-1228: port fields_test to java

abellina:
Fields::select can currently throw a null pointer exception. Apparently this happens when _index.get("x") cannot find the field "x" returning null. Since it's being used as an integer in tuple.get, the null value attempts but can't unbox to integer. 

Should it call Fields::fieldIndex, which will throw an IllegalArgumentException if the field does not exist?


storm-core/test/jvm/org/apache/storm/tuple/FieldsTest.java
```diff
@@ -0,0 +1,122 @@
...
+        new Fields("foo", "bar", "foo");
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void duplicateFieldsNotAllowedTestWhenConstructingFromListTest() {
+        new Fields(new String[] {"foo", "bar", "foo"});
+    }
+
+    private Fields getFields() {
+        return new Fields("foo", "bar");
```
revans2: There are a lot of places where I would prefer to create the Fields instance once and then manipulate it, instead of creating a new one each time.  Also I find it cleaner to see the class being instantiated right by where they are tested.

revans2:
Looks good to me, just one minor comment.

abellina:
@revans2 How does this look? https://github.com/abellina/storm/commit/daec40455b61f6ae84ab1ef9f1514e15e45dc4e4, do you want to see the type being instantiated, or the actual instantiation happening near the assertion?

Also, I had an initial comment on a potential NPE in .select (first comment in PR), thoughts?

revans2:
I would prefer to see getFields inlined everywhere.  We are not saving enough code to really make it worth it.

```
    @Test(expected = IndexOutOfBoundsException.class)
    public void getThrowsWhenOutOfBoundsTest() {
        Fields fields = getFields(); // only has two items
        fields.get(2);
    }
```
requires the comment `// only has two items` but

```
    @Test(expected = IndexOutOfBoundsException.class)
    public void getThrowsWhenOutOfBoundsTest() {
        Fields fields = new Fields("a", "b");
        fields.get(2);
    }
```

is much more readable and needs no comment.  If getFields were a large method that required a lot of setup then it would be OK, but for this it is small enough I think it should be inlined.

As for the NPE.  I don't really want to touch that code just yet.  If you want to file a JIRA with a reproducible use case we can look at it and see what ramifications a change might have.

abellina:
@revans2 sounds good, addressed. Here's the JIRA for the NPE issue: https://issues.apache.org/jira/browse/STORM-1594

abellina:
Can this go in? It's a pretty simple PR.

abellina:
@revans2 just checking on this one. Anything else that needs done?

HeartSaVioR:
Sorry to be late. +1 on me. Merging...

PR #1161, opened 2016-02-28
[Storm-1579] Fix NoSuchFileException when running tests in storm-core

unsleepy22:
The old code takes "storm.local.dir" property as logDir while changes in STORM-1552 used ConfigUtils.getLogDir, which doesn't take this property, causing the ultimate event log directory to be "/logs/xxxx" and resulting in NoSuchFileException because permission is denied to create /logs directory.

The change of code causes 3 test cases to fail in supervisor_test, so I changed them accordingly to make all tests pass.


unsleepy22:
Build failure looks odd, I have all tests pass locally.

abhishekagarwal87:
I suspected that the logs directory is being created inside the project itself and failing rat check. I pulled in your changes to check the same. 
```find storm-core/storm-local/logs -type f | wc -l
26
``` 
This directory is not excluded from the RAT check and thus build is failing. can you try excluding this directory from RAT check and run the build on your local end (mvn clean install) to see if the build succeeds?
Sorry to tell this late but I see a problem with these changes I didn't think of before. Earlier the default log location was under logs directory. With these changes, logs will start going in storm-local directory which is meant for internal use. logs directory becomes redundant since storm.loca.dir always has a default value. In fact this PR will break STORM-1552 which @arunmahadevan should confirm.

The exceptions in test classes may be fixed without changing the getLogDir implementation (by passing the `storm.local.dir/logs` as storm.log.dir property value instead of hardcoded logs directory)

unsleepy22:
@abhishekagarwal87 meaning that add `storm.log.dir: /logs` to defaults.yaml?

abhishekagarwal87:
/logs is an absolute path. Setting `storm.log.dir: logs` may work if it resolves to {storm.home}/logs in worker and other daemons. However, correct me if I am wrong, this will not remove the test exceptions. Instead of hard-coded "/logs", passing a temporary path such as /tmp/storm-logs as the value of storm.log.dir should remove the exceptions.

unsleepy22:
@abhishekagarwal87 , I get you. My opinion would be:
1. revert changes to ConfigUtils class
2. set "storm.log.dir: logs" in defaults.yaml
3. add "**/logs/**" to apache-rat-plugin excludes
4. remain supervisor-test the same with my commit because settting "storm.log.dir = logs OR /tmp/logs" will break supervisor-test anyway.

What do you think?

abhishekagarwal87:
1.Not necessary though redundant after 2.
2. If it is same as storm.log.dir = `concatIfNotNull(System.getProperty("storm.home")) + FILE_SEPARATOR + "logs"` then I don't see any problem. 
3. +1
4. +1
I feel the test errors can still be solved with just 3 and 4.. can you tell me which tests throw NoSuchFileException? I will also try out something.

unsleepy22:
There're pretty a lot:
org.apache.storm.transactional-test
org.apache.storm.drpc-test
org.apache.storm.grouping-test
org.apache.storm.clojure-test
org.apache.storm.metrics-test
org.apache.storm.messaging.netty-integration-test
org.apache.storm.messaging-test

I cannot see where to explicitly set storm.log.dir when running these tests, that's why I set it in defaults.yaml.

unsleepy22:
@abhishekagarwal87 changed according to your comments, please help review again, rat check also passes.

abhishekagarwal87:
the logs directory resolves to current path (from where the command runs) and not the storm home e.g. I set the `storm.log.dir: logs` in my defaults.yaml and ran it from my home directory. the logs were created inside the home directory. You will have to do something like this in getLogDir() function in ConfigUtils 
```
if (new File(dir).isAbsolute()) {
                return dir;
            } else {
                return concatIfNotNull(System.getProperty("storm.home")) + FILE_SEPARATOR + dir);
            }
```

unsleepy22:
Probably won't work since storm.home is not set when running tests.

abhishekagarwal87:
In ConfigUtils.getLogDir() replace
```
} else  {
            dir = concatIfNotNull(System.getProperty("storm.home")) + FILE_SEPARATOR + "logs";
```
with
```
 } else  if (System.getProperty("storm.home") != null) {
            dir = System.getProperty("storm.home") + FILE_SEPARATOR + "logs";
        } else {
            dir = "logs";
```
Revert the changes in defaults.yaml and keep the changes in pom.xml and supervisor test. 

storm.home is not set and thus log dir is being set to /logs. With the above changes in test, it will be set to logs. So it should work.

unsleepy22:
@abhishekagarwal87 your change is more careful & considerate, but changing defaults.yaml would be simpler and it's pretty reasonable to set storm.log.dir to "logs" by default, what do you think? Also, @revans2 what's your opinion on this?

abhishekagarwal87:
Thanks @unsleepy22 . Logs directory should not be created from the path where the daemon is started from. If storm.log.dir had resolved to ${storm.home}/logs, it would have been fine. But that is not  happening here. Different daemons (worker, logviewer, supervisor) can be started from different paths. That means their logs will go into different locations. Each daemon must see the same path of base log dir.

unsleepy22:
Ah, you're right, I get you!

abhishekagarwal87:
If you are taking my suggestions into consideration, you also need to remove these from getLogDir() func
```
} else if (System.getProperty("storm.local.dir") != null) {
 +            dir = System.getProperty("storm.local.dir") + FILE_SEPARATOR + "logs";
 +        } else if (conf.get("storm.local.dir") != null) {
 +            dir = conf.get("storm.local.dir") + FILE_SEPARATOR + "logs";
 +        }
```

revans2:
The changes I see look fine to me I am +1 for merging them in.  If we are going to go with this patch I would like to see the logs dir deleted as part `mvn clean` and I would like to see it added to `.gitignore` so we don't add log files by accident.

I personally think that a final solution should include changing how we do logging when in local mode.  I would disable the dynamic log configuration for workers when in local mode, and I would disable the event logger configuration when in local mode.  I have seen errors when I try to run in local mode and the permissions for log files in `storm.home` or `storm.log.dir` are not quite set properly.  If I am running in local mode I should not be writing anything to `storm.home`.

unsleepy22:
@abhishekagarwal87 changed according to your comments
@revans2 added logs to .gitignore and maven-antrun-plugin to delete logs directory after tests
please help review again

abhishekagarwal87:
Thanks @unsleepy22 +1

pom.xml
```diff
@@ -914,6 +916,26 @@
...
+                            <configuration>
+                                <tasks>
+                                    <delete quiet="true" includeEmptyDirs="true" failonerror="false">
+                                        <fileset dir="logs"/>
+                                    </delete>
+                                </tasks>
+                            </configuration>
+                        </execution>
+                    </executions>
+                </plugin>
```
revans2: I have had a lot of issues in the past with trying to use antrun to clean.  I really would prefer to change it to be something like

```
            <plugin>
                <artifactId>maven-clean-plugin</artifactId>
                <version>2.5</version>
                <executions>
                    <execution>
                        <id>cleanup</id>
                        <phase>clean</phase>
                        <goals>
                            <goal>clean</goal>
                        </goals>
                        <configuration>
                            <excludeDefaultDirectories>true</excludeDefaultDirectories>
                            <filesets>
                                <fileset>
                                    <directory>./logs/</directory>
                                </fileset>
                            </filesets>
                        </configuration>
                    </execution>
                </executions>
            </plugin>
```

revans2:
One minor comment, but it is important that we get this and #1172 in, so I am fine with making the change myself as a part of the merge.  +1

PR #1168, opened 2016-02-29
Fixed incorrect storm-kafka documentation.

dsimmons:
Removed an incorrect (outdated?) constructor of `SpoutConfig` from the README.

Just threw me for a quick loop until I [looked at the source](https://github.com/apache/storm/blob/master/external/storm-kafka/src/jvm/org/apache/storm/kafka/SpoutConfig.java#L41) and confirmed that, sure enough, `zkRoot` isn't optional (the constructor isn't ***actually*** overloaded like the README claims).


ptgoetz:
+1

dossett:
+1

PR #1169, opened 2016-03-01
[Storm-1585]  Add DDL support for UDFs in storm-sql

arunmahadevan:
This patch proposes to expose the user defined function support added
in STORM-1586 via DDL statements. 

Note: This includes changes from https://github.com/apache/storm/pull/1162

### DIFF
--- a/external/sql/storm-sql-core/src/codegen/includes/parserImpls.ftl
+++ b/external/sql/storm-sql-core/src/codegen/includes/parserImpls.ftl
@@ -93,13 +93,19 @@ SqlNode SqlCreateFunction() :
     SqlParserPos pos;
     SqlIdentifier functionName;
     SqlNode className;
+    SqlNode jarName = null;
 }
 {
     <CREATE> { pos = getPos(); }
     <FUNCTION>
-    functionName = CompoundIdentifier()
+        functionName = CompoundIdentifier()
     <AS>
-    className = StringLiteral() {
-        return new SqlCreateFunction(pos, functionName, className);
+        className = StringLiteral()
+    [
+      <USING> <JAR>
+      jarName = StringLiteral()
+    ]
+    {
+      return new SqlCreateFunction(pos, functionName, className, jarName);
     }
 }
--- a/external/sql/storm-sql-core/src/codegen/includes/parserImpls.ftl
+++ b/external/sql/storm-sql-core/src/codegen/includes/parserImpls.ftl
@@ -84,3 +84,22 @@ SqlNode SqlCreateTable() :
         tbl_properties, select);
     }
 }
+
+/**
+ * CREATE FUNCTION functionname AS 'classname'
+ */
+SqlNode SqlCreateFunction() :
+{
+    SqlParserPos pos;
+    SqlIdentifier functionName;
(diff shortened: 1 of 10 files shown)


external/sql/storm-sql-core/src/codegen/includes/parserImpls.ftl
```diff
@@ -83,4 +83,23 @@ SqlNode SqlCreateTable() :
...
+    SqlParserPos pos;
+    SqlIdentifier functionName;
+    SqlNode className;
+}
+{
+    <CREATE> { pos = getPos(); }
+    <FUNCTION>
+    functionName = CompoundIdentifier()
+    <AS>
+    className = StringLiteral() {
```
satishd: CREATE FUNCTION functionname AS 'classname' does not take any jar location or name. Where should user put UDF libs? Is it mandatory to put each UDF class in a separate jar? 
Can you please describe how should UDF libraries packaged and placed in storm environment?

arunmahadevan: The syntax closely follows the hive syntax for defining UDFs. Whats required is each UDF should map to a unique class with an `evaluate` method. Multiple such classes can be in the same jar and added to the classpath. (for storm, it could be placed under extlib directory). Will update the docs with this info.

Hive has an additional syntax to specify jar file in its DDL syntax which is not supported as of now.

harshach:
@haohui can you take a look

external/sql/storm-sql-core/src/jvm/org/apache/storm/sql/parser/SqlCreateFunction.java
```diff
@@ -0,0 +1,68 @@
+package org.apache.storm.sql.parser;
```
haohui: Need license header

external/sql/storm-sql-core/src/test/org/apache/storm/sql/TestStormSql.java
```diff
@@ -129,4 +134,20 @@ public void testExternalNestedInvalidAccess() throws Exception {
     sql.execute(stmt, h);
     Assert.assertEquals(0, values.size());
   }
+
+  @Test
+  public void testExternalUdf() throws Exception {
```
haohui: It would be a good idea to have unit tests for UDF + type conversions.

external/sql/storm-sql-core/src/codegen/includes/parserImpls.ftl
haohui: That way the clauses might result into a parsing error. Instead of reporting a parsing error, it might be more user friendly to throw an UnimplementedException when compiling the SQL statements. It is your call.

arunmahadevan:
@haohui addressed comments, can you take a look again?

arunmahadevan:
@harshach @haohui this has been pending for a long time. Could you take a look and merge if it looks fine?

harshach:
+1

harshach:
Thanks for your patience @arunmahadevan . Merged into master and 1.x-branch.

PR #1170, opened 2016-03-01
STORM-1588: Do not add event logger details if number of event loggers is zero

abhishekagarwal87:


abhishekagarwal87:
(count (sort nil)) returns zero hence put an explicit condition for checking if event logger is enabled or not.

revans2:
+1 looks good to me.

abhishekagarwal87:
https://github.com/apache/storm/pull/1175 for 1.x branch

PR #1171, opened 2016-03-01
[STORM-1590] port defmeters/defgauge/defhistogram... to java for all of our code to use

unsleepy22:
This PR is under STORM-1590, I changed nimbus.clj only currently and would like to hear your opinion first before moving on.


storm-core/src/jvm/org/apache/storm/metric/StormMetricsRegistry.java
```diff
@@ -0,0 +1,69 @@
...
+import com.codahale.metrics.Meter;
+import com.codahale.metrics.Metric;
+import com.codahale.metrics.MetricRegistry;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@SuppressWarnings("unchecked")
+public class StormMetricsRegistry {
+    private static final Logger LOG = LoggerFactory.getLogger(StormMetricsRegistry.class);
+    private static final MetricRegistry metrics = new MetricRegistry();
```
revans2: Can we just make this public and rename it to DEFAULT_REGISTRY.  We can still use the helper methods, but they are not as critical.

unsleepy22: Sure, will change

unsleepy22:
@revans2 changed according to your comments, also changed all the other files. start-metrics-reporters in common.clj is changed to java too. Please help review again.

abhishekagarwal87:
@revans2 @unsleepy22 - can we use (.) as separator instead of (:) in the metric names? Reason being that graphite and possibly other systems recognize dot as a separator but not (:)? It need not be done in this PR itself. I can make those changes in a separate PR.

revans2:
@abhishekagarwal87 I would be fine with a change like that.  Please file a JIRA for it and we can make the change everywhere.

hustfxj:
@unsleepy22  thank you +1

unsleepy22:
@revans2 could you take a look?

revans2:
+1 looks good to me

PR #1173, opened 2016-03-01
[STORM-1469] Adding Plain Sasl Transport Plugin

kishorvpatil:
Since `org.apache.storm.security.auth.SimpleTransportPlugin` creates `TFramedTransport` with maxBufferSize, but that conflicts with serialized topology of size exceeded 16MB - default `THsHaServer` allowed max buffer size. So the better option is to implement `PlainSaslTransportPlugin` which avoids using any default values for max-buffer-size.

### DIFF
--- a/conf/defaults.yaml
+++ b/conf/defaults.yaml
@@ -39,7 +39,7 @@ storm.exhibitor.port: 8080
 storm.exhibitor.poll.uripath: "/exhibitor/v1/cluster/list"
 storm.cluster.mode: "distributed" # can be distributed or local
 storm.local.mode.zmq: false
-storm.thrift.transport: "org.apache.storm.security.auth.plain.PlainSaslTransportPlugin"
+storm.thrift.transport: "org.apache.storm.security.auth.SimpleTransportPlugin"
 storm.principal.tolocal: "org.apache.storm.security.auth.DefaultPrincipalToLocal"
 storm.group.mapping.service: "org.apache.storm.security.auth.ShellBasedGroupsMapping"
 storm.group.mapping.service.params: null
--- a/conf/defaults.yaml
+++ b/conf/defaults.yaml
@@ -39,7 +39,7 @@ storm.exhibitor.port: 8080
 storm.exhibitor.poll.uripath: "/exhibitor/v1/cluster/list"
 storm.cluster.mode: "distributed" # can be distributed or local
 storm.local.mode.zmq: false
-storm.thrift.transport: "org.apache.storm.security.auth.SimpleTransportPlugin"
+storm.thrift.transport: "org.apache.storm.security.auth.plain.PlainSaslTransportPlugin"
 storm.principal.tolocal: "org.apache.storm.security.auth.DefaultPrincipalToLocal"
 storm.group.mapping.service: "org.apache.storm.security.auth.ShellBasedGroupsMapping"
 storm.group.mapping.service.params: null
--- a/storm-core/src/jvm/org/apache/storm/security/auth/AbstractSaslClientCallbackHandler.java
+++ b/storm-core/src/jvm/org/apache/storm/security/auth/AbstractSaslClientCallbackHandler.java
@@ -1,0 +1,76 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.storm.security.auth;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.security.auth.callback.Callback;
+import javax.security.auth.callback.CallbackHandler;
+import javax.security.auth.callback.NameCallback;
+import javax.security.auth.callback.PasswordCallback;
+import javax.security.auth.callback.UnsupportedCallbackException;
+import javax.security.sasl.AuthorizeCallback;
+import javax.security.sasl.RealmCallback;
+import java.io.IOException;
+
+public abstract class AbstractSaslClientCallbackHandler implements CallbackHandler {
+    protected static final String USERNAME = "username";
+    protected static final String PASSWORD = "password";
+    private static final Logger LOG = LoggerFactory.getLogger(AbstractSaslClientCallbackHandler.class);
+    protected String _username = null;
+    protected String _password = null;
+
+    /**
+     * This method is invoked by SASL for authentication challenges
+     * @param callbacks a collection of challenge callbacks
+     */
+    public void handle(Callback[] callbacks) throws IOException, UnsupportedCallbackException {
+        for (Callback c : callbacks) {
+            if (c instanceof NameCallback) {
+                LOG.debug("name callback");
+                NameCallback nc = (NameCallback) c;
+                nc.setName(_username);
+            } else if (c instanceof PasswordCallback) {
+                LOG.debug("password callback");
+                PasswordCallback pc = (PasswordCallback)c;
+                if (_password != null) {
+                    pc.setPassword(_password.toCharArray());
+                }
+            } else if (c instanceof AuthorizeCallback) {
+                LOG.debug("authorization callback");
+                AuthorizeCallback ac = (AuthorizeCallback) c;
+                String authid = ac.getAuthenticationID();
+                String authzid = ac.getAuthorizationID();
+                if (authid.equals(authzid)) {
+                    ac.setAuthorized(true);
+                } else {
+                    ac.setAuthorized(false);
+                }
+                if (ac.isAuthorized()) {
+                    ac.setAuthorizedID(authzid);
+                }
+            } else if (c instanceof RealmCallback) {
+                RealmCallback rc = (RealmCallback) c;
+                ((RealmCallback) c).setText(rc.getDefaultText());
+            } else {
+                throw new UnsupportedCallbackException(c);
+            }
+        }
+    }
+}
--- a/storm-core/src/jvm/org/apache/storm/security/auth/AbstractSaslServerCallbackHandler.java
+++ b/storm-core/src/jvm/org/apache/storm/security/auth/AbstractSaslServerCallbackHandler.java
@@ -1,0 +1,77 @@
+package org.apache.storm.security.auth;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.security.auth.callback.Callback;
+import javax.security.auth.callback.CallbackHandler;
+import javax.security.auth.callback.NameCallback;
+import javax.security.auth.callback.PasswordCallback;
+import javax.security.auth.callback.UnsupportedCallbackException;
+import javax.security.sasl.AuthorizeCallback;
+import javax.security.sasl.RealmCallback;
+import java.util.HashMap;
+import java.util.Map;
+
+public abstract class AbstractSaslServerCallbackHandler implements CallbackHandler {
+    private static final Logger LOG = LoggerFactory.getLogger(AbstractSaslServerCallbackHandler.class);
+    protected final Map<String,String> credentials = new HashMap<>();
+    protected String userName;
+
+    public void handle(Callback[] callbacks) throws UnsupportedCallbackException {
+        for (Callback callback : callbacks) {
+            if (callback instanceof NameCallback) {
+                handleNameCallback((NameCallback) callback);
+            } else if (callback instanceof PasswordCallback) {
+                handlePasswordCallback((PasswordCallback) callback);
+            } else if (callback instanceof RealmCallback) {
+                handleRealmCallback((RealmCallback) callback);
+            } else if (callback instanceof AuthorizeCallback) {
+                handleAuthorizeCallback((AuthorizeCallback) callback);
+            }
+        }
+    }
+
+    private void handleNameCallback(NameCallback nc) {
+        LOG.debug("handleNameCallback");
+        userName = nc.getDefaultName();
+        nc.setName(nc.getDefaultName());
+    }
+
+    protected void handlePasswordCallback(PasswordCallback pc) {
+        LOG.debug("handlePasswordCallback");
+        if (credentials.containsKey(userName) ) {
+            pc.setPassword(credentials.get(userName).toCharArray());
+        } else {
+            LOG.warn("No password found for user: " + userName);
+        }
+    }
+
+    private void handleRealmCallback(RealmCallback rc) {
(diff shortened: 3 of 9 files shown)


conf/defaults.yaml
```diff
@@ -39,7 +39,7 @@ storm.exhibitor.port: 8080
 storm.exhibitor.poll.uripath: "/exhibitor/v1/cluster/list"
 storm.cluster.mode: "distributed" # can be distributed or local
 storm.local.mode.zmq: false
-storm.thrift.transport: "org.apache.storm.security.auth.SimpleTransportPlugin"
+storm.thrift.transport: "org.apache.storm.security.auth.plain.PlainSaslTransportPlugin"
```
d2r: What is the rationale for the `.plain` portion of this config? The class already says it is Plain.

storm-core/src/jvm/org/apache/storm/security/auth/plain/PlainClientCallbackHandler.java
```diff
@@ -0,0 +1,80 @@
...
+                String authid = ac.getAuthenticationID();
+                String authzid = ac.getAuthorizationID();
+                if (authid.equals(authzid)) {
+                    ac.setAuthorized(true);
+                } else {
+                    ac.setAuthorized(false);
+                }
+                if (ac.isAuthorized()) {
+                    ac.setAuthorizedID(authzid);
+                }
```
d2r: We can just `setAuthorizedID` above when we `setAuthorized`.

storm-core/src/jvm/org/apache/storm/security/auth/plain/PlainClientCallbackHandler.java
```diff
@@ -0,0 +1,80 @@
...
+            } else if (c instanceof RealmCallback) {
```
d2r: Minor: there is no LOG.debug here as with the other callbacks. Is that intentional?

storm-core/src/jvm/org/apache/storm/security/auth/plain/PlainClientCallbackHandler.java
```diff
@@ -0,0 +1,80 @@
...
+                RealmCallback rc = (RealmCallback) c;
+                ((RealmCallback) c).setText(rc.getDefaultText());
```
d2r: Can we not reuse `rc` instead of casting twice?

storm-core/src/jvm/org/apache/storm/security/auth/plain/PlainClientCallbackHandler.java
```diff
@@ -0,0 +1,80 @@
...
+
+/**
+ *  client side callback handler.
+ */
+public class PlainClientCallbackHandler implements CallbackHandler {
+    private static final String USERNAME = "username";
+    private static final String PASSWORD = "password";
+    private static final Logger LOG = LoggerFactory.getLogger(PlainClientCallbackHandler.class);
+    private String _username = "username";
+    private String _password = "password";
```
d2r: Let's get rid of the private members and just use the constants already defined above.

We also want to reference these hard-coded user/pass in the other classes.

storm-core/src/jvm/org/apache/storm/security/auth/plain/PlainSaslTransportPlugin.java
```diff
@@ -0,0 +1,80 @@
...
+            client_callback_handler,
+            transport);
+
+        wrapper_transport.open();
+        LOG.debug("SASL PLAIN client transport has been established");
+
+        return wrapper_transport;
+
+    }
+
```
d2r: nit: blank lines

storm-core/src/jvm/org/apache/storm/security/auth/plain/PlainServerCallbackHandler.java
```diff
@@ -0,0 +1,108 @@
...
+import javax.security.auth.callback.UnsupportedCallbackException;
+import javax.security.sasl.AuthorizeCallback;
+import javax.security.sasl.RealmCallback;
+
+/**
+ * SASL server side callback handler
+ */
+public class PlainServerCallbackHandler implements CallbackHandler {
+    private static final Logger LOG = LoggerFactory.getLogger(PlainServerCallbackHandler.class);
+    private static final String SYSPROP_SUPER_PASSWORD = "storm.SASLAuthenticationProvider.superPassword";
```
d2r: Do we need a superuser password?  I thought this was all hard-coded to use constants?

storm-core/src/jvm/org/apache/storm/security/auth/plain/PlainServerCallbackHandler.java
```diff
@@ -0,0 +1,108 @@
...
+
+    private String userName="username";
+    private final Map<String,String> credentials = new HashMap<>();
+
+    public PlainServerCallbackHandler() throws IOException {
+        credentials.put("username", "password");
+    }
```
d2r: Here we probably want to reference constants.

storm-core/src/jvm/org/apache/storm/security/auth/plain/PlainServerCallbackHandler.java
```diff
@@ -0,0 +1,108 @@
...
+        LOG.debug("handlePasswordCallback");
+        if ("super".equals(this.userName) && System.getProperty(SYSPROP_SUPER_PASSWORD) != null) {
+            // superuser: use Java system property for password, if available.
+            pc.setPassword(System.getProperty(SYSPROP_SUPER_PASSWORD).toCharArray());
+        } else if (credentials.containsKey(userName) ) {
+            pc.setPassword(credentials.get(userName).toCharArray());
+        } else {
+            LOG.warn("No password found for user: " + userName);
+        }
+    }
```
d2r: Same question here.  Not sure we want the concept of a superuser.

storm-core/src/jvm/org/apache/storm/security/auth/plain/PlainServerCallbackHandler.java
```diff
@@ -0,0 +1,108 @@
...
+
+        //if authorizationId is not set, set it to authenticationId.
+        if(ac.getAuthorizationID() == null) {
+            ac.setAuthorizedID(authenticationID);
+        }
+
+        //When authNid and authZid are not equal , authNId is attempting to impersonate authZid, We
+        //add the authNid as the real user in reqContext's subject which will be used during authorization.
+        if(!authenticationID.equals(ac.getAuthorizationID())) {
+            LOG.info("Impersonation attempt  authenticationID = " + ac.getAuthenticationID() + " authorizationID = " + ac.getAuthorizationID());
```
d2r: minor: use format strings `{}` like in other log statements.

storm-core/src/jvm/org/apache/storm/security/auth/plain/SaslPlainServer.java
```diff
@@ -0,0 +1,154 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.storm.security.auth.plain;
+
+import java.security.Provider;
+import java.util.Map;
+
+import javax.security.auth.callback.*;
```
d2r: I would be nice to expand all imports.

conf/defaults.yaml
kishorvpatil: Under `org.apache.storm.security.auth` package we have sub-package by mechanism e.g. `digest`, `kerberos`. Hence the package name `plain`

storm-core/src/jvm/org/apache/storm/security/auth/plain/SaslPlainServer.java
```diff
@@ -0,0 +1,154 @@
...
+import javax.security.sasl.SaslServer;
+import javax.security.sasl.SaslServerFactory;
+
+public class SaslPlainServer implements SaslServer {
+  @SuppressWarnings("serial")
+  public static class SecurityProvider extends Provider {
+    public SecurityProvider() {
+      super("SaslPlainServer", 1.0, "SASL PLAIN Authentication Server");
+      put("SaslServerFactory.PLAIN",
+          SaslPlainServerFactory.class.getName());
```
d2r: Just `"SaslPlainServerFactory"` ?

d2r:
@kishorvpatil Some of this looks like borrowed code that is not tailored specifically to storm. Is that the case?

kishorvpatil:
@d2r That's right. The `SaslPlainServer` is implementation borrowed from [hadoop](https://github.com/apache/hadoop/blob/trunk/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/security/SaslPlainServer.java). And `CallBackHandler` implementations are mostly clones from digest package with minor changes.

kishorvpatil:
@d2r i have tried to address all your comments, including refactoring two abstract classed `AbstractSaslClientCallbackHandler.java` and `AbstractSaslServerCallbackHandler.java` to place common methods for DIGEST and PLAIN in one place.

storm-core/src/jvm/org/apache/storm/security/auth/plain/PlainSaslTransportPlugin.java
```diff
@@ -0,0 +1,69 @@
...
+import java.security.Security;
+
+public class PlainSaslTransportPlugin extends SaslTransportPlugin {
+    public static final String PLAIN = "PLAIN";
+    private static final Logger LOG = LoggerFactory.getLogger(PlainSaslTransportPlugin.class);
+
+    @Override
+    protected TTransportFactory getServerTransportFactory() throws IOException {
+        //create an authentication callback handler
+        CallbackHandler server_callback_handler = new PlainServerCallbackHandler();
```
abhishekagarwal87: minor: could use camelCase convention.

conf/defaults.yaml
revans2: This breaks wire compatibility, so it will not be a rolling upgrade.  I am OK with that, but I want to be sure that it is documented.

storm-core/src/jvm/org/apache/storm/security/auth/AbstractSaslClientCallbackHandler.java
```diff
@@ -0,0 +1,76 @@
...
+    /**
+     * This method is invoked by SASL for authentication challenges
+     * @param callbacks a collection of challenge callbacks
+     */
+    public void handle(Callback[] callbacks) throws IOException, UnsupportedCallbackException {
+        for (Callback c : callbacks) {
+            if (c instanceof NameCallback) {
+                LOG.debug("name callback");
+                NameCallback nc = (NameCallback) c;
+                nc.setName(_username);
```
revans2: So we are always going to set the name to null???

storm-core/src/jvm/org/apache/storm/security/auth/AbstractSaslServerCallbackHandler.java
```diff
@@ -0,0 +1,77 @@
...
+        LOG.debug("handlePasswordCallback");
+        if (credentials.containsKey(userName) ) {
+            pc.setPassword(credentials.get(userName).toCharArray());
+        } else {
+            LOG.warn("No password found for user: " + userName);
+        }
+    }
+
+    private void handleRealmCallback(RealmCallback rc) {
+        LOG.debug("handleRealmCallback: "+ rc.getDefaultText());
```
revans2: Lets use the "{}" syntax here.

storm-core/src/jvm/org/apache/storm/security/auth/AbstractSaslServerCallbackHandler.java
```diff
@@ -0,0 +1,77 @@
...
+        userName = nc.getDefaultName();
+        nc.setName(nc.getDefaultName());
+    }
+
+    protected void handlePasswordCallback(PasswordCallback pc) {
+        LOG.debug("handlePasswordCallback");
+        if (credentials.containsKey(userName) ) {
+            pc.setPassword(credentials.get(userName).toCharArray());
+        } else {
+            LOG.warn("No password found for user: " + userName);
```
revans2: "{}" syntax

storm-core/src/jvm/org/apache/storm/security/auth/plain/PlainSaslTransportPlugin.java
```diff
@@ -0,0 +1,69 @@
...
+        Security.addProvider(new SaslPlainServer.SecurityProvider());
```
revans2: What happens if we add this multiple times?

storm-core/src/jvm/org/apache/storm/security/auth/plain/SaslPlainServer.java
revans2: This ensures that if the package changes or something else like it changes we get a compile error, not a runtime error.

revans2:
Overall it looks fairly good, but you have a problem with RAT somewhere.

storm-core/src/jvm/org/apache/storm/security/auth/AbstractSaslServerCallbackHandler.java
```diff
@@ -0,0 +1,77 @@
+package org.apache.storm.security.auth;
```
revans2: Needs an Apache header.

storm-core/src/jvm/org/apache/storm/security/auth/plain/PlainClientCallbackHandler.java
```diff
@@ -0,0 +1,31 @@
...
+
+import org.apache.storm.security.auth.AbstractSaslClientCallbackHandler;
+
+public class PlainClientCallbackHandler extends AbstractSaslClientCallbackHandler {
+
+    /*
+     * For plain, using constants for a pair of user name and password.
+     */
+    public PlainClientCallbackHandler() {
+        _username = USERNAME;
```
revans2: I forgot to ask before is the user this always reports to nimbus/etc now "username"?  If so I really would prefer to pass in the user name of the current user on this system, or even better have the server set the user to null like we were doing before.

storm-core/src/jvm/org/apache/storm/security/auth/plain/PlainSaslTransportPlugin.java
kishorvpatil: I will add it to constructor with additional check on if provider is already registered.

conf/defaults.yaml
kishorvpatil: Since there is not good place to document this. I am simply going without any chances to default on "storm.thrift.tranport" so we can pull this in 1.x-branch as well. I will create two separate pull requests - 
1. to change default and deprecate  `SimpleTransportPlugin`  and configuration on 1.x-branch.
2. To remove `SimpleTransportPlugin` and configuration in master.

storm-core/src/jvm/org/apache/storm/security/auth/AbstractSaslServerCallbackHandler.java
kishorvpatil: added

storm-core/src/jvm/org/apache/storm/security/auth/plain/PlainClientCallbackHandler.java
kishorvpatil: Fixed by using System.getProperty("user.name").

revans2:
Looks good to me.  The test failures look unrelated to this.  +1

d2r:
Seems good to me. +1

kishorvpatil:
Created two pull requests as follow-up

#1178  Remove SimpleTransportPlugin and configuration and 
#1179 Decommissioning SimpleTransportPlugin and configuration ( on 1.x-branch.)

PR #1107, opened 2016-02-14
STORM-1549: Add support for resetting tuple timeout from bolts via the OutputCollector

srdo:
See https://issues.apache.org/jira/browse/STORM-1549.

Other than implementing tuple timeout reset, I made a minor change to assert-loop in integration_test.clj so it doesn't hang on failing tests. It should now fail the test after 10 seconds. I hope that's okay.

PR #1174, opened 2016-03-01
STORM-1549: [master] Add support for resetting tuple timeout from bolts via the OutputCollector

srdo:
This is a port of https://github.com/apache/storm/pull/1107 to master.


storm-core/src/jvm/org/apache/storm/daemon/Acker.java
```diff
@@ -100,6 +101,8 @@ public void execute(Tuple input) {
             }
             curr.failed = true;
             pending.put(id, curr);
+        } else if(ACKER_RESET_TIMEOUT_STREAM_ID.equals(streamId)) {
```
abhishekagarwal87: minor space around if

abhishekagarwal87: This introduces a NPE scenario because earlier in all the if blocks, if curr was null, it was getting initialized. Now curr.spoutTask will throw NPE if reset timeout comes before init or after init times out. This can be avoided by adding below just before `pending.put(id, curr)`
```
if (curr == null) {
     curr = new AckObject();
}
```

abhishekagarwal87:
Looks good to me after the possible NPE is fixed.

storm-core/src/jvm/org/apache/storm/task/OutputCollector.java
```diff
@@ -218,6 +218,16 @@ public void fail(Tuple input) {
         _delegate.fail(input);
     }
 
+    /**
+    * Resets the message timeout for any tuple trees to which the given tuple belongs.
+    * The timeout is reset to Config.TOPOLOGY_MESSAGE_TIMEOUT_SECS.
+    * @param input the tuple to reset timeout for
+    */
```
revans2: Can we add to the javadoc that this should be used sparingly?  This is not a cheap operation, so having someone do it for each tuple they receive is very expensive.

storm-core/src/jvm/org/apache/storm/topology/BasicOutputCollector.java
```diff
@@ -52,6 +52,10 @@ public void emitDirect(int taskId, List<Object> tuple) {
         emitDirect(taskId, Utils.DEFAULT_STREAM_ID, tuple);
     }
 
+    public void resetTimeout(Tuple tuple){
```
revans2: Can we put the javadoc here too for resetTimeout from outputCollector?

revans2:
Overall things look good.  Just a few minor nits about javadocs and what @abhishekagarwal87 pointed out.

storm-core/src/jvm/org/apache/storm/daemon/Acker.java
revans2: Yes this is very true, in the clj code the entire processing was wrapped in a null check for curr.

srdo:
The NPE should be fixed, I've added the missing space to the if, and added a comment to the javadoc about reset timeout being an expensive operation. The comment has also been added to the 1.x-branch version

rsltrifork:
As all review comments are addressed, will it be merged soon? :-)
We need this feature in the IoT storm topology we're building, where bolts send to unstable external systems and we need at-least-once consistency, so dropping tuples on tuple timeout is bad for us.

srdo:
@revans2 @abhishekagarwal87 Sorry, didn't notice this had merge conflicts. It should be good to go now though :)

abhishekagarwal87:
Looks good to me.

vesense:
LGTM

revans2:
+1

PR #1176, opened 2016-03-02
STORM-1283: port backtype.storm.MockAutoCred to java

abhishekagarwal87:


redsanket:
+1

unsleepy22:
+1

hustfxj:
+1

revans2:
Please move MockAutoCred from `src/jvm/org/apache/storm` to `test/jvm/org/apache/storm`.  The test still passes, and it was only put in the src directory because of problems with compiling clojure vs interpretation.

After that I am +1

revans2:
@abhishekagarwal87 actually I'll just do it as part of the check-in.  There is no code change so it should be fine.

abhishekagarwal87:
Thanks @revans2

PR #1177, opened 2016-03-03
[STORM-1596] Do not use single Kerberos TGT instance between multiple threads.

kishorvpatil:
This needs to be ported to 1.x-branch as well.


kishorvpatil:
The Travis CI test failure seems unrelated to this issue.

revans2:
+1 we might want to look at porting this to 0.10.x also.

kishorvpatil:
Here #1180  is pull request to port these changes to 0.10.x-branch

PR #1180, opened 2016-03-03
[STORM-1596] Do not use single Kerberos TGT instance between multiple threads 0.10x

kishorvpatil:


revans2:
+1

revans2:
The code looks more or less identical to #1177 except for the java package names so I think I can merge it in now.

PR #1004, opened 2016-01-12
STORM:1463 added file scehma to log4j config files for windows env

satishd:

PR #1183, opened 2016-03-04
STORM-1603 Storm UT fails on supervisor test in Windows

HeartSaVioR:
* supervisor UTs seem not handle file path delimeter properly
* due to absolute-storm-local-dir, we should care Windows specific
  * if worker dir path is relative and storm.home is empty, path will be invalid

I think it would affect 1.x and even master, but other tests disturb me verifying test result of supervisor-test.
For example, blobstore UTs are failed on Windows, and it even makes other UTs crashed (so that running test is stopped at the moment).

If I can see the test result for other branches, I'll craft pull request for each branches.

### DIFF
--- a/storm-core/test/clj/backtype/storm/supervisor_test.clj
+++ b/storm-core/test/clj/backtype/storm/supervisor_test.clj
@@ -253,7 +253,6 @@
           mock-worker-id "fake-worker-id"
           mock-cp (str file-path-separator "base" class-path-separator file-path-separator "stormjar.jar")
           mock-sensitivity "S3"
-          mock-cp "/base:/stormjar.jar"
           exp-args-fn (fn [opts topo-opts classpath]
                        (concat [(supervisor/java-cmd) "-cp" classpath 
                                (str "-Dlogfile.name=" mock-storm-id "-worker-" mock-port ".log")
@@ -261,8 +260,8 @@
                                 (str "-Dstorm.id=" mock-storm-id)
                                 (str "-Dworker.id=" mock-worker-id)
                                 (str "-Dworker.port=" mock-port)
-                               "-Dstorm.log.dir=/logs"
-                               "-Dlog4j.configurationFile=/log4j2/worker.xml"
+                                (str "-Dstorm.log.dir=" file-path-separator "logs")
+                                (str "-Dlog4j.configurationFile=" file-path-separator "log4j2" file-path-separator "worker.xml")
                                "backtype.storm.LogWriter"]
                                [(supervisor/java-cmd) "-server"]
                                opts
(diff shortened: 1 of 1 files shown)


storm-core/test/clj/backtype/storm/supervisor_test.clj
```diff
@@ -261,8 +260,8 @@
                                 (str "-Dstorm.id=" mock-storm-id)
                                 (str "-Dworker.id=" mock-worker-id)
                                 (str "-Dworker.port=" mock-port)
-                               "-Dstorm.log.dir=/logs"
-                               "-Dlog4j.configurationFile=/log4j2/worker.xml"
+                                (str "-Dstorm.log.dir=" file-path-separator "logs")
+                                (str "-Dlog4j.configurationFile=" file-path-separator "log4j2" file-path-separator "worker.xml")
```
hustfxj: I think you maybe need  fix -Dlog4j.configurationFile as supervisor.clj’s:

          log4j-configuration-file (str (if (.startsWith (System/getProperty "os.name") "Windows")
                                          (if (.startsWith storm-log4j2-conf-dir "file:")
                                            storm-log4j2-conf-dir
                                            (str "file:///" storm-log4j2-conf-dir))
                                          storm-log4j2-conf-dir)
                                     Utils/FILE_PATH_SEPARATOR "worker.xml")

HeartSaVioR: @hustfxj 
STORM-1463 is merged into master (2.x) and 1.x, and this pull request is for 0.10.x.
While this PR can still make unit tests succeed, we may want to backport STORM-1463 to 0.10.x-branch and reflect change to supervisor_test.clj.

HeartSaVioR: Seems like there're many pull requests regarding log4j configuration applied without backporting to 0.10.x. STORM-1463 is just one of these.

I'm wondering we would want to find whole of things and backport to 0.10.x. If we don't want, this PR is just good to go for 0.10.x, and I'll create pull request for 1.x and master.

What do you think?

hustfxj: @HeartSaVioR  sorry, I didn't see STORM-1463 before.  I think we should find whole of things and fix it.  Thank you @HeartSaVioR

knusbaum:
+1. I'm in favor of backporting STORM-1463 (#1004) as well.

HeartSaVioR:
@knusbaum STORM-1463 is just one of them.
I can't find other issues which affect this since history for supervisor.clj is gone when moving package.
Do we want to just sync up with supervisor.clj in 1.x-branch (only log related) disregarding the history (finding whole issues and change fix versions)?

knusbaum:
@HeartSaVioR 
That depends on how big the diff is. If it's a large diff, I don't think we want to do that. I would much prefer to pull individual patches in to ensure we're not adding features or creating regressions. 

This conversation probably belongs on the mailing list, though, so others can weigh in. For now, I think we can merge this and treat backporting fixes to 10.x as a separate issue.

HeartSaVioR:
@knusbaum 
Yes, I'll send the mail to dev mailing list. For now, I'm also in favor of separating issue as `fixing UT` and `catching up change with backporting`.

HeartSaVioR:
OK, sent the mail to dev. I'll merge this...

PR #1184, opened 2016-03-04
[STORM-1279] port backtype.storm.daemon.supervisor to java

hustfxj:
1 port supervisor.clj to java;
2 Update all the callings to supervisor;
3 Supervisor's class hierarchy as follows:
      ``SupervisorManger is supervisor' manger which can clean and shutdown supervisor;``
      ``SyncSupervisorEvent is responsible for downloading/removing assignments and topologys' files;``
      ``SyncProcessEvent is responsible for starting/killing workers;``
      ``SupervisorUtils have common methods;``
4 create local-supervisor.clj for local mode.
5. fix the fails on supervisor test in Windows


hustfxj:
success on supervisor test in my computer of  windows, but CI have 2 failures about supervisor-test.clj. I can't see the supervisor-test.xml. So I don't kown where failures come ? I didi some manual testing and things look good. I hope everyone can review it and give me some suggests. With my pressure!

abhishekagarwal87:
WorkerManager - shutting down worker, launching worker, cleanup worker, reading worker heartbeat (basically the pieces of code which deals with individual worker) can go here. 
Code which calls the WorkerManager for each worker e.g. shutdownAllWorkers can go to SupervisorManager as you have done.

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SupervisorUtils.java
```diff
@@ -0,0 +1,273 @@
...
+    }
+
+    /**
+     * Given the blob information returns the value of the uncompress field, handling it either being a string or a boolean value, or if it's not specified then
+     * returns false
+     * 
+     * @param blobInfo
+     * @return
+     */
+    public static Boolean isShouldUncompressBlob(Map<String, Object> blobInfo) {
```
abhishekagarwal87: can be renamed to shouldUncompressBlob

storm-core/src/jvm/org/apache/storm/daemon/supervisor/ShutdownWork.java
```diff
@@ -0,0 +1,125 @@
...
+import org.apache.storm.utils.Time;
+import org.apache.storm.utils.Utils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.*;
+
+public  class ShutdownWork implements Shutdownable {
```
zhuoliu: It is OK to put the shutWorker function in a separate class but we not think it should implement Shutdownable. 

Or we might just go with the original implementation to put these two functions in SupervisorUtils.java as static. Then SupervisorManger and SyncProcessEvent will not need to extend ShutdownWork (they seem not natural subclasses of ShutdownWork.)

storm-core/src/jvm/org/apache/storm/daemon/supervisor/StandaloneSupervisor.java
```diff
@@ -0,0 +1,87 @@
...
+
+import java.io.IOException;
+import java.util.Collection;
+import java.util.Map;
+import java.util.UUID;
+
+public class StandaloneSupervisor implements ISupervisor {
+
+    private String supervisorId;
+
```
zhuoliu: extra line

hustfxj:
@zhuoliu @abhishekagarwal87  Thank you. I will consider your suggestions, and adjust class hierarchy. Of course, I hope others can help review the supervisor and give me some comments. Then I will revise the codes together in order to avoid to different suggestions and repeat modifications. Thank you again.

abhishekagarwal87:
Sure @hustfxj  Makes sense

storm-core/src/clj/org/apache/storm/daemon/local_supervisor.clj
```diff
@@ -0,0 +1,61 @@
...
+;; limitations under the License.
+(ns org.apache.storm.daemon.local-supervisor
+  (:import [org.apache.storm.daemon.supervisor SyncProcessEvent SupervisorData ShutdownWork Supervisor]
+           [org.apache.storm.utils Utils ConfigUtils]
+           [org.apache.storm ProcessSimulator])
+  (:use [org.apache.storm.daemon common]
+        [org.apache.storm log])
+  (:require [org.apache.storm.daemon [worker :as worker] ])
+  (:require [clojure.string :as str])
+  (:gen-class))
```
revans2: If we are going to leave this part of the supervisor in clojure for now that is fine, but we need a follow on JIRA to move it to java wonce worker is moved to java.

storm-core/src/jvm/org/apache/storm/daemon/supervisor/State.java
```diff
@@ -0,0 +1,22 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.storm.daemon.supervisor;
+
+public enum State {
+    valid, disallowed, notStarted, timedOut;
```
revans2: minor nit. Can we name these more like java const Enums values?

```
VALID, DISALLOWED, NOT_STARTED, TIMED_OUT;
```

storm-core/src/jvm/org/apache/storm/daemon/supervisor/timer/SupervisorHeartbeat.java
```diff
@@ -0,0 +1,89 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class SupervisorHeartbeat implements Runnable {
+
+    private IStormClusterState stormClusterState;
+    private String supervisorId;
+    private Map conf;
+    private SupervisorInfo supervisorInfo;
```
revans2: This can be removed.  It is only needed inside `update`.

storm-core/src/jvm/org/apache/storm/daemon/supervisor/timer/SupervisorHeartbeat.java
```diff
@@ -0,0 +1,89 @@
...
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class SupervisorHeartbeat implements Runnable {
+
+    private IStormClusterState stormClusterState;
+    private String supervisorId;
+    private Map conf;
```
revans2: Can we please mark stormClusterState, supervisorId, conf and supervisorData as final?

storm-core/src/jvm/org/apache/storm/daemon/supervisor/Supervisor.java
```diff
@@ -0,0 +1,196 @@
...
+                Utils.exitProcess(13, "Error on initialization");
+            }
+        }
+        return supervisorManger;
+    }
+
+    /**
+     * start distribute supervisor
+     */
+    private void launch() {
```
revans2: In the short term `launch` should be static and take an instance of ISupervisor as a parameter.  This is required to continue to support some projects like storm on mesos.

https://github.com/mesos/storm/blob/master/storm/src/main/storm/mesos/MesosSupervisor.java#L58

storm-core/src/jvm/org/apache/storm/daemon/supervisor/Supervisor.java
```diff
@@ -0,0 +1,196 @@
...
+            startMetricsReporters(conf);
+        } catch (Exception e) {
+            LOG.error("Failed to start supervisor\n", e);
+            System.exit(1);
+        }
+    }
+
+    // To be removed
+    private void registerWorkerNumGauge(String name, final Map conf) {
+        MetricRegistry metricRegistry = new MetricRegistry();
```
revans2: Please use the StormMericsRegistry instead.

storm-core/src/jvm/org/apache/storm/daemon/supervisor/Supervisor.java
```diff
@@ -0,0 +1,196 @@
...
+            public Integer getValue() {
+                Collection<String> pids = SupervisorUtils.myWorkerIds(conf);
+                return pids.size();
+            }
+        });
+    }
+
+    // To be removed
+    private void startMetricsReporters(Map conf) {
+        List<PreparableReporter> preparableReporters = MetricsUtils.getPreparableReporters(conf);
```
revans2: Here to please use the code we already merged in.

storm-core/src/jvm/org/apache/storm/daemon/supervisor/Supervisor.java
revans2: If we cannot make it static, that is fine, but it still needs to take an ISupervisor instance as a parameter somewhere.

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SupervisorData.java
```diff
@@ -0,0 +1,339 @@
...
+    public void setStormIdToProfileActions(Map<String, List<ProfileRequest>> stormIdToProfileActions) {
+        this.stormIdToProfileActions.clear();
+        this.stormIdToProfileActions.putAll(stormIdToProfileActions);
+    }
+
+    public Map getConf() {
+        return conf;
+    }
+
+    public void setConf(Map conf) {
```
revans2: This is not used and I am not sure of a time we would want to do this.  Can we remove setConf and make conf final?

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SupervisorData.java
```diff
@@ -0,0 +1,339 @@
...
+        this.conf = conf;
+    }
+
+    public IContext getSharedContext() {
+        return sharedContext;
+    }
+
+    public void setSharedContext(IContext sharedContext) {
```
revans2: Here too this is not used so it would probably be best to remove this method and mark sharedContext as final.  At least for the time being.

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SupervisorData.java
```diff
@@ -0,0 +1,339 @@
...
+
+    public void setActive(boolean active) {
+        this.active = active;
+    }
+
+    public ISupervisor getiSupervisor() {
+        return iSupervisor;
+    }
+
+    public void setiSupervisor(ISupervisor iSupervisor) {
```
revans2: Not used.

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SupervisorData.java
```diff
@@ -0,0 +1,339 @@
...
+        this.iSupervisor = iSupervisor;
+    }
+
+    public Utils.UptimeComputer getUpTime() {
+        return upTime;
+    }
+
+    public void setUpTime(Utils.UptimeComputer upTime) {
```
revans2: Not used.

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SupervisorData.java
```diff
@@ -0,0 +1,339 @@
...
+        this.upTime = upTime;
+    }
+
+    public String getStormVersion() {
+        return stormVersion;
+    }
+
+    public void setStormVersion(String stormVersion) {
```
revans2: Not used

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SupervisorData.java
```diff
@@ -0,0 +1,339 @@
...
+        this.stormVersion = stormVersion;
+    }
+
+    public ConcurrentHashMap<String, String> getWorkerThreadPidsAtom() {
+        return workerThreadPidsAtom;
+    }
+
+    public void setWorkerThreadPidsAtom(ConcurrentHashMap<String, String> workerThreadPidsAtom) {
```
revans2: Not used.

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SupervisorData.java
```diff
@@ -0,0 +1,339 @@
...
+
+    public String getStormVersion() {
+        return stormVersion;
+    }
+
+    public void setStormVersion(String stormVersion) {
+        this.stormVersion = stormVersion;
+    }
+
+    public ConcurrentHashMap<String, String> getWorkerThreadPidsAtom() {
```
revans2: This is not really an atom any more, so lets drop atom from the name.

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SupervisorData.java
```diff
@@ -0,0 +1,339 @@
...
+
+    public void setWorkerThreadPidsAtom(ConcurrentHashMap<String, String> workerThreadPidsAtom) {
+        this.workerThreadPidsAtom = workerThreadPidsAtom;
+    }
+
+    public IStormClusterState getStormClusterState() {
+        return stormClusterState;
+    }
+
+    public void setStormClusterState(IStormClusterState stormClusterState) {
```
revans2: Not used

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SupervisorData.java
```diff
@@ -0,0 +1,339 @@
...
+        this.stormClusterState = stormClusterState;
+    }
+
+    public LocalState getLocalState() {
+        return localState;
+    }
+
+    public void setLocalState(LocalState localState) {
```
revans2: Not used

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SupervisorData.java
```diff
@@ -0,0 +1,339 @@
...
+        this.localState = localState;
+    }
+
+    public String getSupervisorId() {
+        return supervisorId;
+    }
+
+    public void setSupervisorId(String supervisorId) {
```
revans2: Not used.

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SupervisorData.java
```diff
@@ -0,0 +1,339 @@
...
+        this.supervisorId = supervisorId;
+    }
+
+    public String getAssignmentId() {
+        return assignmentId;
+    }
+
+    public void setAssignmentId(String assignmentId) {
```
revans2: Not used

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SupervisorData.java
```diff
@@ -0,0 +1,339 @@
...
+        this.assignmentId = assignmentId;
+    }
+
+    public String getHostName() {
+        return hostName;
+    }
+
+    public void setHostName(String hostName) {
```
revans2: Not Used

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SupervisorData.java
```diff
@@ -0,0 +1,339 @@
...
+        this.hostName = hostName;
+    }
+
+    public ConcurrentHashMap<Long, LocalAssignment> getCurrAssignment() {
+        return currAssignment;
+    }
+
+    public void setCurrAssignment(Map<Long, LocalAssignment> currAssignment) {
+        this.currAssignment.clear();
+        this.currAssignment.putAll(currAssignment);
```
revans2: The clear with the add should probably be atomic.  The race is very small, but I would rather see us have an AtomicReference<Map<Long, LocalAssignment>> instead of a ConcurrentHashMap.

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SupervisorData.java
```diff
@@ -0,0 +1,339 @@
...
+    }
+
+    public StormTimer getHeartbeatTimer() {
+        return heartbeatTimer;
+    }
+
+    public void setHeartbeatTimer(StormTimer heartbeatTimer) {
```
revans2: Not used

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SupervisorData.java
```diff
@@ -0,0 +1,339 @@
...
+        this.heartbeatTimer = heartbeatTimer;
+    }
+
+    public StormTimer getEventTimer() {
+        return eventTimer;
+    }
+
+    public void setEventTimer(StormTimer eventTimer) {
```
revans2: Not used

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SupervisorData.java
```diff
@@ -0,0 +1,339 @@
...
+        this.eventTimer = eventTimer;
+    }
+
+    public StormTimer getBlobUpdateTimer() {
+        return blobUpdateTimer;
+    }
+
+    public void setBlobUpdateTimer(StormTimer blobUpdateTimer) {
```
revans2: Not used

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SupervisorData.java
```diff
@@ -0,0 +1,339 @@
...
+        this.blobUpdateTimer = blobUpdateTimer;
+    }
+
+    public Localizer getLocalizer() {
+        return localizer;
+    }
+
+    public void setLocalizer(Localizer localizer) {
```
revans2: Not used

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SupervisorData.java
```diff
@@ -0,0 +1,339 @@
...
+        this.localizer = localizer;
+    }
+
+    public AtomicInteger getSyncRetry() {
+        return syncRetry;
+    }
+
+    public void setSyncRetry(AtomicInteger syncRetry) {
```
revans2: Not Used.

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SupervisorData.java
```diff
@@ -0,0 +1,339 @@
...
+        this.syncRetry = syncRetry;
+    }
+
+    public ConcurrentHashMap<String, Map<String, Object>> getAssignmentVersions() {
+        return assignmentVersions;
+    }
+
+    public void setAssignmentVersions(Map<String, Map<String, Object>> assignmentVersions) {
+        this.assignmentVersions.clear();
+        this.assignmentVersions.putAll(assignmentVersions);
```
revans2: Here too this operation should be atomic and I would rather see assignmentVersions be an `AtomicReference<Map<String, Map<String, Object>>>`

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SupervisorData.java
```diff
@@ -0,0 +1,339 @@
...
+    }
+
+    public CgroupManager getResourceIsolationManager() {
+        return resourceIsolationManager;
+    }
+
+    public void setResourceIsolationManager(CgroupManager resourceIsolationManager) {
```
revans2: Not used

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SupervisorData.java
```diff
@@ -0,0 +1,339 @@
...
+
+    public Object getDownloadLock() {
+        return downloadLock;
+    }
+
+    public ConcurrentHashSet getDeadWorkers() {
+        return deadWorkers;
+    }
+
+    public void setDeadWorkers(ConcurrentHashSet deadWorkers) {
```
revans2: Not used.

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SupervisorUtils.java
```diff
@@ -0,0 +1,273 @@
...
+        if (StringUtils.isNotBlank(wlinitial)) {
+            wl = wlinitial;
+        } else {
+            wl = stormHome + "/bin/worker-launcher";
+        }
+        List<String> commands = new ArrayList<>();
+        commands.add(wl);
+        commands.add(user);
+        commands.addAll(args);
+        return Utils.launchProcess(commands, environment, logPreFix, exitCodeCallback, dir);
```
revans2: We are missing a log message just before this.
```
(log-message "Running as user:" user " command:" (pr-str command))
```

it can be very helpful for debugging

```
LOG.info("Running as user: {} command: {}", user, command);
```

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SupervisorUtils.java
```diff
@@ -0,0 +1,273 @@
...
+        }
+        return stormIds;
+    }
+
+    public static Collection<String> supervisorWorkerIds(Map conf) {
+        String workerRoot = ConfigUtils.workerRoot(conf);
+        return Utils.readDirContents(workerRoot);
+    }
+
+    public static boolean checkTopoFilesExist(Map conf, String stormId) throws IOException {
```
revans2: check implies to me that we will do something if they don't exist.  I would prefer to see this be a question like `doTopoFilesExist`, or `doRequiredTopoFilesExist`.

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SupervisorUtils.java
```diff
@@ -0,0 +1,273 @@
...
+        if (!Utils.checkFileExists(stormcodepath))
+            return false;
+        if (!Utils.checkFileExists(stormconfpath))
+            return false;
+        if (ConfigUtils.isLocalMode(conf) || Utils.checkFileExists(stormjarpath))
+            return true;
+        return false;
+    }
+
+    public static Collection<String> myWorkerIds(Map conf){
```
revans2: This is the same as supervisorWorkerIds.  We should combine the two together.

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SupervisorUtils.java
```diff
@@ -0,0 +1,273 @@
...
+        String javaHome = System.getenv().get("JAVA_HOME");
+        if (StringUtils.isNotBlank(javaHome)) {
+            ret = javaHome + Utils.FILE_PATH_SEPARATOR + "bin" + Utils.FILE_PATH_SEPARATOR + cmd;
+        } else {
+            ret = cmd;
+        }
+        return ret;
+    }
+    
+    public static List<ACL> supervisorZkAcls() {
```
revans2: In the original code this was a static final const value.  It is probably not that big of a deal, but it would be nice to keep it the same if possible.

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SyncProcessEvent.java
```diff
@@ -0,0 +1,669 @@
...
+        for (Map.Entry<String, LSWorkerHeartbeat> entry : idToHeartbeat.entrySet()) {
+            String workerId = entry.getKey();
+            LSWorkerHeartbeat whb = entry.getValue();
+            State state;
+            if (whb == null) {
+                state = State.notStarted;
+            } else if (!approvedIds.contains(workerId) || !matchesAssignment(whb, assignedExecutors)) {
+                state = State.disallowed;
+            } else if (supervisorData.getDeadWorkers().contains(workerId)) {
+                LOG.info("Worker Process {}as died", workerId);
```
revans2: should be
```
                LOG.info("Worker Process {} has died", workerId);
```

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SyncProcessEvent.java
```diff
@@ -0,0 +1,669 @@
...
+
+    protected boolean matchesAssignment(LSWorkerHeartbeat whb, Map<Integer, LocalAssignment> assignedExecutors) {
+        LocalAssignment localAssignment = assignedExecutors.get(whb.get_port());
+        if (localAssignment == null || !localAssignment.get_topology_id().equals(whb.get_topology_id())) {
+            return false;
+        }
+        List<ExecutorInfo> executorInfos = new ArrayList<>();
+        executorInfos.addAll(whb.get_executors());
+        // remove SYSTEM_EXECUTOR_ID
+        executorInfos.remove(new ExecutorInfo(-1, -1));
```
revans2: Very minor nit. Can we move this ExecutorInfo to Contants as something like
```
public static final SYSTEM_EXECUTOR_INFO = new ExecutorInfo(-1, -1);
```

so when worker.clj goes to java we can drop SYSTEM-EXECUTOR_ID from Constants?

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SyncProcessEvent.java
```diff
@@ -0,0 +1,669 @@
...
+    protected String getWorkerClassPath(String stormJar, Map stormConf) {
+        List<String> topoClasspath = new ArrayList<>();
+        Object object = stormConf.get(Config.TOPOLOGY_CLASSPATH);
+
+        if (object instanceof List) {
+            topoClasspath.addAll((List<String>) object);
+        } else if (object instanceof String){
+            topoClasspath.add((String)object);
+        }else {
+            //ignore
```
revans2: Can we throw an exception here if object is not null?  If TOPOLOGY_CLASSPATH is something else we have an error in our code somewhere.

storm-core/src/jvm/org/apache/storm/daemon/supervisor/ShutdownWork.java
```diff
@@ -0,0 +1,125 @@
...
+            ProcessSimulator.killProcess(threadPid);
+        }
+
+        for (String pid : pids) {
+            if (asUser) {
+                List<String> commands = new ArrayList<>();
+                commands.add("signal");
+                commands.add(pid);
+                commands.add("15");
+                String logPrefix = "kill - 15 " + pid;
```
abhishekagarwal87: nit. -15 (no space)

storm-core/src/jvm/org/apache/storm/daemon/supervisor/ShutdownWork.java
```diff
@@ -0,0 +1,125 @@
...
+            Time.sleepSecs(shutdownSleepSecs);
+        }
+
+        for (String pid : pids) {
+            if (asUser) {
+                List<String> commands = new ArrayList<>();
+                commands.add("signal");
+                commands.add(pid);
+                commands.add("9");
+                String logPrefix = "kill - 9 " + pid;
```
abhishekagarwal87: nit. -9

storm-core/src/jvm/org/apache/storm/daemon/supervisor/Supervisor.java
```diff
@@ -0,0 +1,196 @@
...
+        try {
+            Map<Object, Object> conf = Utils.readStormConfig();
+            if (ConfigUtils.isLocalMode(conf)) {
+                throw new IllegalArgumentException("Cannot start server in local mode!");
+            }
+            ISupervisor iSupervisor = new StandaloneSupervisor();
+            supervisorManager = mkSupervisor(conf, null, iSupervisor);
+            if (supervisorManager != null)
+                Utils.addShutdownHookWithForceKillIn1Sec(supervisorManager);
+            registerWorkerNumGauge("drpc:num-execute-http-requests", conf);
```
abhishekagarwal87: drpc !!.. name should be corrected

storm-core/src/clj/org/apache/storm/daemon/local_supervisor.clj
hustfxj: ok! I will follow up the JIRA to move local_supervisor.clj to java after this PR is merged into master.

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SupervisorUtils.java
```diff
@@ -0,0 +1,273 @@
...
+
+    public static void setInstance(SupervisorUtils u) {
+        _instance = u;
+    }
+
+    public static void resetInstance() {
+        _instance = INSTANCE;
+    }
+
+    public static Process workerLauncher(Map conf, String user, List<String> args, Map<String, String> environment, final String logPreFix,
```
abhishekagarwal87: I never understood why is it called worker launcher. One may think that it launches a worker but it doesn't. I wish if it could be renamed to something else.

hustfxj: Yes. I am confused with the name in fact.

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SyncProcessEvent.java
hustfxj: @revans2 the exception of topology will kill supervisor if we throw an exception. Do you think it is ok?

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SupervisorUtils.java
revans2: It does launch the worker when run-as-user is set, that was what it was originally for, but other operations needed privileges too, and as such it has expanded its role.

I am fine with changing it, just thought I would give you some background on it.  FYI we took most of the code from the hadoop `container-launcher`, and they have the same naming problem.

Ideally I would ultimately like to see a  WorkerManager java API that would call into a native linux-worker-manager binary and would look something like the following, NOTE: I have not thought through all of the APIs and I have really not thought about exactly what parameters would need to be passed to it.  This is just a general concept that would allow us to eventually support docker, local-workers, run-as-user, cgroups and possibly other things much more cleanly.

```
public interface WorkerManager {
  Future<WorkerPrepareResult> prepareWorker(String workerId, LocalisedResources lr, Resources resources);
  Future<LaunchWorkerResult> launchWorker(String workerId, WorkerExitCallback, ...);
  Future<KillWorkerResult> killWorker(String workerId);
  Future<CleanupWorkerResult> cleanupWorker(String workerId,...);
  List<WorkerInfo> runningWorkers();
  //Possibly in the  future
  //Future<ResizeWorkerResult> resizeWorker(String workerId, Resources newResources);
}
```

hustfxj: @revans2  agree.

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SyncProcessEvent.java
revans2: The original code was

```
(let [...
          topo-classpath (if-let [cp (storm-conf TOPOLOGY-CLASSPATH)]
                           [cp]
                           [])
          ... (Utils/addToClasspath something topo-classpath) ...
```

The Signature for Utils.addToClasspath is

```
public static String addToClasspath(String classpath, Collection<String> paths);
```

if `cp` was anything but a `String` in that code we would get a `ClassCastException` that would bring down the supervisor.  This code is even more lenient and supports both `String` and `List<String>`.  Which fixes a bug in our translation of the Utils code.

You are right that we don't want a bad user config to bring down a supervisor, but if nimbus is doing the right thing and validating the config before it goes to the supervisor we should never see this happen.  `TOPOLOGY_CLASSPATH` is annotated with ` @isStringOrStringList`.  So if we ever see `object` as something that is not a `String`, `List<String>`, or `null` it means we have a bug in our code somewhere, and silently ignoring it feels wrong to me.

I would be fine with us logging an error if it is something different, but not totally ignoring it.

storm-core/src/jvm/org/apache/storm/daemon/supervisor/DaemonCommon.java
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.storm.daemon.supervisor;
+
+public interface DaemonCommon {
```
jerrypeng: is DaemonCommon going to be a interface that needs to be implemented by all Storm daemons? If so we need to move this file one directory up

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SupervisorUtils.java
abhishekagarwal87: Thank you for the clarification. runAsUser sounds like a good name for this API. It is also possible to always pass user in the API signatures. If SUPERVISOR_WORKER_LAUNCHER is not enabled, the user argument will be local user and ignored in the implementation. Otherwise, commands will be launched with that user's permission. The implementation can be prepared when supervisor starts up.

storm-core/src/jvm/org/apache/storm/daemon/supervisor/StandaloneSupervisor.java
```diff
@@ -0,0 +1,84 @@
...
+        return supervisorId;
+    }
+
+    @Override
+    public String getAssignmentId() {
+        return supervisorId;
+    }
+
+    @Override
+    // @return is vector which need be converted to be int
```
abhishekagarwal87: this comment can be removed

storm-core/src/jvm/org/apache/storm/daemon/supervisor/timer/SupervisorHeartbeat.java
```diff
@@ -0,0 +1,87 @@
...
+import org.apache.storm.generated.SupervisorInfo;
+import org.apache.storm.utils.Time;
+import org.apache.storm.utils.Utils;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class SupervisorHeartbeat implements Runnable {
```
abhishekagarwal87: does it need to be Runnable?

storm-core/src/jvm/org/apache/storm/daemon/supervisor/timer/SupervisorHeartbeat.java
```diff
@@ -0,0 +1,87 @@
...
+     private final SupervisorData supervisorData;
+
+    public SupervisorHeartbeat(Map conf, SupervisorData supervisorData) {
+        this.stormClusterState = supervisorData.getStormClusterState();
+        this.supervisorId = supervisorData.getSupervisorId();
+        this.supervisorData = supervisorData;
+        this.conf = conf;
+    }
+
+    private SupervisorInfo update(Map conf, SupervisorData supervisorData) {
```
abhishekagarwal87: can this be renamed to buildSupervisorInfo or constructSupervisorInfo?

storm-core/src/jvm/org/apache/storm/daemon/supervisor/timer/SupervisorHeartbeat.java
abhishekagarwal87: Please ignore my above comment. I see that it is being scheduled.

storm-core/src/jvm/org/apache/storm/daemon/supervisor/Supervisor.java
```diff
@@ -0,0 +1,178 @@
...
+            registerWorkerNumGauge("supervisor:num-slots-used-gauge", conf);
+            StormMetricsRegistry.startMetricsReporters(conf);
+        } catch (Exception e) {
+            LOG.error("Failed to start supervisor\n", e);
+            System.exit(1);
+        }
+    }
+
+    private void registerWorkerNumGauge(String name, final Map conf) {
+        StormMetricsRegistry.registerGauge("supervisor:num-slots-used-gauge", new Callable<Integer>() {
```
abhishekagarwal87: name argument is not used.

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SupervisorManger.java
```diff
@@ -0,0 +1,97 @@
...
+    private SupervisorData supervisorData;
+
+    public SupervisorManger(SupervisorData supervisorData, EventManager eventManager, EventManager processesEventManager) {
+        this.eventManager = eventManager;
+        this.supervisorData = supervisorData;
+        this.processesEventManager = processesEventManager;
+    }
+
+    public void shutdown() {
+        LOG.info("Shutting down supervisor{}", supervisorData.getSupervisorId());
```
abhishekagarwal87: minor. space after supervisor.

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SupervisorManger.java
```diff
@@ -0,0 +1,97 @@
...
+
+import org.apache.storm.event.EventManager;
+import org.apache.storm.utils.Utils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Collection;
+import java.util.Map;
+
+public class SupervisorManger implements SupervisorDaemon, DaemonCommon, Runnable {
```
abhishekagarwal87: nit. class name should be SupervisorManager.

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SupervisorManger.java
```diff
@@ -0,0 +1,97 @@
...
+        }
+
+        if (supervisorData.getHeartbeatTimer().isTimerWaiting() && supervisorData.getEventTimer().isTimerWaiting() && eventManager.waiting()
+                && processesEventManager.waiting()) {
+            return true;
+        }
+        return false;
+    }
+
+    public void run() {
```
abhishekagarwal87: it is not very intuitive that SupervisorManager is also a shutdown hook. You can remove this from here and in the use an anonymous shutdown hook

storm-core/src/jvm/org/apache/storm/daemon/supervisor/Supervisor.java
```diff
@@ -0,0 +1,178 @@
...
+        LOG.info("Starting supervisor for storm version '{}'.", VersionInfo.getVersion());
+        SupervisorManger supervisorManager;
+        try {
+            Map<Object, Object> conf = Utils.readStormConfig();
+            if (ConfigUtils.isLocalMode(conf)) {
+                throw new IllegalArgumentException("Cannot start server in local mode!");
+            }
+            supervisorManager = mkSupervisor(conf, null, iSupervisor);
+            if (supervisorManager != null)
+                Utils.addShutdownHookWithForceKillIn1Sec(supervisorManager);
```
abhishekagarwal87: when you remove runnable interface from supervisorManager, this will become 
```
Utils.addShutdownHookWithForceKillIn1Sec(new Runnable() { 
  public void run() {
     supervisorManager.shutdown()
}
});
```

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SupervisorUtils.java
```diff
@@ -0,0 +1,353 @@
...
+
+    /**
+     * Given the blob information returns the value of the uncompress field, handling it either being a string or a boolean value, or if it's not specified then
+     * returns false
+     * 
+     * @param blobInfo
+     * @return
+     */
+    public static Boolean shouldUncompressBlob(Map<String, Object> blobInfo) {
+        return new Boolean((String) blobInfo.get("uncompress"));
```
abhishekagarwal87: It only handles string values and not boolean values. Correct definition would be
` return new Boolean(String.valueOf(blobInfo.get("uncompress")))`

storm-core/src/jvm/org/apache/storm/daemon/supervisor/Supervisor.java
```diff
@@ -0,0 +1,178 @@
...
+            final SupervisorData supervisorData = new SupervisorData(conf, sharedContext, iSupervisor);
+            Localizer localizer = supervisorData.getLocalizer();
+
+            SupervisorHeartbeat hb = new SupervisorHeartbeat(conf, supervisorData);
+            hb.run();
+            // should synchronize supervisor so it doesn't launch anything after being down (optimization)
+            Integer heartbeatFrequency = Utils.getInt(conf.get(Config.SUPERVISOR_HEARTBEAT_FREQUENCY_SECS));
+            supervisorData.getHeartbeatTimer().scheduleRecurring(0, heartbeatFrequency, hb);
+
+            Set<String> downdedStormId = SupervisorUtils.readDownLoadedStormIds(conf);
```
abhishekagarwal87: nit. downloadedStormIds

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SupervisorUtils.java
```diff
@@ -0,0 +1,353 @@
...
+    }
+
+    /**
+     * Remove a reference to a blob when its no longer needed
```
abhishekagarwal87: Correct doc comment is 
`Returns a list of LocalResources based on the blobstore-map passed in.`

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SupervisorUtils.java
```diff
@@ -0,0 +1,353 @@
...
+            return false;
+        if (!Utils.checkFileExists(stormconfpath))
+            return false;
+        if (ConfigUtils.isLocalMode(conf) || Utils.checkFileExists(stormjarpath))
+            return true;
+        return false;
+    }
+
+    /**
+     * Returns map from worr id to heartbeat
```
abhishekagarwal87: nit. map from worker id to heartbeat

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SyncProcessEvent.java
```diff
@@ -0,0 +1,672 @@
...
+    }
+
+    //TODO: initData is intended to local supervisor, so we will remove them after porting worker.clj to java
+    public void init(SupervisorData supervisorData){
+        this.supervisorData = supervisorData;
+        this.localState = supervisorData.getLocalState();
+    }
+
+
+    /**
```
abhishekagarwal87: same comment is at the class level also which can be removed.

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SyncProcessEvent.java
```diff
@@ -0,0 +1,672 @@
...
+    }
+
+    public SyncProcessEvent(){
+
+    }
+    public SyncProcessEvent(SupervisorData supervisorData) {
+        init(supervisorData);
+    }
+
+    //TODO: initData is intended to local supervisor, so we will remove them after porting worker.clj to java
```
abhishekagarwal87: is this a valid to-do?

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SyncProcessEvent.java
```diff
@@ -0,0 +1,672 @@
...
+
+    /**
+     * 1. to kill are those in allocated that are dead or disallowed 2. kill the ones that should be dead - read pids, kill -9 and individually remove file -
+     * rmr heartbeat dir, rmdir pid dir, rmdir id dir (catch exception and log) 3. of the rest, figure out what assignments aren't yet satisfied 4. generate new
+     * worker ids, write new "approved workers" to LS 5. create local dir for worker id 5. launch new workers (give worker-id, port, and supervisor-id) 6. wait
+     * for workers launch
+     */
+    @Override
+    public void run() {
+        LOG.debug("Syncing processes");
```
abhishekagarwal87: this can be removed.

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SyncProcessEvent.java
```diff
@@ -0,0 +1,672 @@
...
+        String jlp = jlp(stormRoot, conf);
+
+        String stormJar = ConfigUtils.supervisorStormJarPath(stormRoot);
+
+        Map stormConf = ConfigUtils.readSupervisorStormConf(conf, stormId);
+
+        String workerClassPath = getWorkerClassPath(stormJar, stormConf);
+
+        Object topGcOptsObject = stormConf.get(Config.TOPOLOGY_WORKER_GC_CHILDOPTS);
+        List<String> topGcOpts = new ArrayList<>();
```
abhishekagarwal87: This should be done only if topoGcOpts Object is not null.

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SyncProcessEvent.java
```diff
@@ -0,0 +1,672 @@
...
+        int memoffheap = (int) Math.ceil(resources.get_mem_off_heap());
+
+        int cpu = (int) Math.ceil(resources.get_cpu());
+
+        List<String> gcOpts = null;
+
+        if (topGcOpts != null) {
+            gcOpts = substituteChildopts(topGcOpts, workerId, stormId, port, memOnheap);
+        } else {
+            gcOpts = substituteChildopts(conf.get(Config.WORKER_GC_CHILDOPTS), workerId, stormId, port, memOnheap);
```
abhishekagarwal87: this will never be called. Please refer to my comment above.

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SyncProcessEvent.java
```diff
@@ -0,0 +1,672 @@
...
+        }
+
+        Object topoWorkerLogwriterObject = stormConf.get(Config.TOPOLOGY_WORKER_LOGWRITER_CHILDOPTS);
+        List<String> topoWorkerLogwriterChildopts = new ArrayList<>();
+        if (topoWorkerLogwriterObject instanceof String) {
+            topoWorkerLogwriterChildopts.add((String) topoWorkerLogwriterObject);
+        } else if (topoWorkerLogwriterObject instanceof List) {
```
abhishekagarwal87: checking instanceof List is not sufficient here as any collection is acceptable. You can do `instanceof Collection` and cast it to `Collection<String>` instead. Same can be done in other places as well.

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SyncProcessEvent.java
```diff
@@ -0,0 +1,672 @@
...
+     * "Generates runtime childopts by replacing keys with topology-id, worker-id, port, mem-onheap"
+     * 
+     * @param value
+     * @param workerId
+     * @param stormId
+     * @param port
+     * @param memOnheap
+     */
+    public List<String> substituteChildopts(Object value, String workerId, String stormId, Long port, int memOnheap) {
+        List<String> rets = new ArrayList<>();
```
abhishekagarwal87: clojure code would have returned null if value is null. Also default type was assumed to be string.

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SyncProcessEvent.java
```diff
@@ -0,0 +1,672 @@
...
+        if (value instanceof String) {
+            String string = (String) value;
+            string = string.replace("%ID%", String.valueOf(port));
+            string = string.replace("%WORKER-ID%", workerId);
+            string = string.replace("%TOPOLOGY-ID%", stormId);
+            string = string.replace("%WORKER-PORT%", String.valueOf(port));
+            string = string.replace("%HEAP-MEM%", String.valueOf(memOnheap));
+            String[] strings = string.split("\\s+");
+            rets.addAll(Arrays.asList(strings));
+        } else if (value instanceof List) {
```
abhishekagarwal87: this can be collection instead.

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SyncProcessEvent.java
```diff
@@ -0,0 +1,672 @@
...
+            List<Object> objects = (List<Object>) value;
+            for (Object object : objects) {
+                String str = (String)object;
```
abhishekagarwal87: nit. space before object

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SupervisorUtils.java
```diff
@@ -0,0 +1,353 @@
...
+
+    public  boolean  isWorkerHbTimedOutImpl(int now, LSWorkerHeartbeat whb, Map conf) {
+        boolean result = false;
+        if ((now - whb.get_time_secs()) > Utils.getInt(conf.get(Config.SUPERVISOR_WORKER_TIMEOUT_SECS))) {
+            result = true;
+        }
+        return result;
+    }
+
+    public static String javaCmd(String cmd) {
```
abhishekagarwal87: since cmd is always going to be "java", can the parameter be removed?

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SyncProcessEvent.java
```diff
@@ -0,0 +1,672 @@
...
+        commandList.add(workerClassPath);
+        commandList.add("org.apache.storm.daemon.worker");
+        commandList.add(stormId);
+        commandList.add(assignmentId);
+        commandList.add(String.valueOf(port));
+        commandList.add(workerId);
+
+        // {"cpu" cpu "memory" (+ mem-onheap mem-offheap (int (Math/ceil (conf STORM-CGROUP-MEMORY-LIMIT-TOLERANCE-MARGIN-MB))))
+        if (Utils.getBoolean(conf.get(Config.STORM_RESOURCE_ISOLATION_PLUGIN_ENABLE), false)) {
+            int cgRoupMem = (int) (Math.ceil((double) conf.get(Config.STORM_CGROUP_MEMORY_LIMIT_TOLERANCE_MARGIN_MB)));
```
abhishekagarwal87: nit. you can rename the variable to cGroupMem

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SyncProcessEvent.java
```diff
@@ -0,0 +1,672 @@
...
+        //for supervisor-test
+        if (stormconf.get(Config.LOGS_GROUPS) != null) {
+            List<String> groups = (List<String>) stormconf.get(Config.LOGS_GROUPS);
+            for (String group : groups){
+                logsGroups.add(group);
+            }
+        }
+        if (stormconf.get(Config.TOPOLOGY_GROUPS) != null) {
+            List<String> topGroups = (List<String>) stormconf.get(Config.TOPOLOGY_GROUPS);
+            for (String group : topGroups){
```
abhishekagarwal87: you can use logsGroups.addAll

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SyncProcessEvent.java
```diff
@@ -0,0 +1,672 @@
...
+                SupervisorUtils.setupStormCodeDir(conf, ConfigUtils.readSupervisorStormConf(conf, stormId), file.getParentFile().getCanonicalPath());
+            } else {
+                file.getParentFile().mkdirs();
+            }
+        }
+        FileWriter writer = new FileWriter(file);
+        Yaml yaml = new Yaml();
+        try {
+            yaml.dump(data, writer);
+        }finally {
```
abhishekagarwal87: nit. space after }

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SyncSupervisorEvent.java
```diff
@@ -0,0 +1,631 @@
...
+                setupBlobPermission(conf, user, userDir.toString());
+                for (LocalizedResource localizedResource : localizedResources) {
+                    File rsrcFilePath = new File(localizedResource.getFilePath());
+                    String keyName = rsrcFilePath.getName();
+                    String blobSymlinkTargetName = new File(localizedResource.getCurrentSymlinkPath()).getName();
+
+                    String symlinkName = null;
+                    if (blobstoreMap != null) {
+                        Map<String, Object> blobInfo = blobstoreMap.get(keyName);
+                        if (blobInfo != null && blobInfo.containsKey("localname")) {
```
abhishekagarwal87: this logic was wrapped inside get-blob-localname in clojure. In java, it is duplicated in three places.

jerrypeng:
I have a bug fix for cgroups that involve a couple lines of modification in supervisor.clj.  

https://github.com/apache/storm/pull/1216

If that pull request gets merged first please remember to also implement that change here.

hustfxj:
@jerrypeng  Of course

hustfxj:
@revans2  can you look at it again ?

jerrypeng:
Found another bug in supervisor:

https://github.com/apache/storm/pull/1226

just another note here to make sure we remember to get in the fix whether in clojure or java

hustfxj:
@jerrypeng  I have merge your bug fix.Thank you.

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SyncSupervisorEvent.java
```diff
@@ -0,0 +1,632 @@
...
+            Integer port = entry.getKey();
+            if (existingAssignment.containsKey(port)) {
+                intersectAssignment.put(port, entry.getValue());
+            }
+        }
+
+        for (Integer port : intersectAssignment.keySet()) {
+            List<ExecutorInfo> existExecutors = existingAssignment.get(port).get_executors();
+            List<ExecutorInfo> newExecutors = newAssignment.get(port).get_executors();
+            if (newExecutors.size() != existExecutors.size()) {
```
jerrypeng: @hustfxj did you also add the bug fix that was merged in STORM-1636?  If it is, the logic the check whether there is a scheduling change in a worker is not correct.  If there is not a scheduling change for a worker than the assignment for that executor will contain the same executors (regardless of ordering) as the "newAssignment".  Thus we are comparing two sets (without considering ordering) if they contain the same elements.

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SyncProcessEvent.java
```diff
@@ -0,0 +1,428 @@
...
+        resourceFileNames.add(ConfigUtils.RESOURCES_SUBDIR);
+        resourceFileNames.addAll(blobFileNames);
+        LOG.info("Creating symlinks for worker-id: {} storm-id: {} for files({}): {}", workerId, stormId, resourceFileNames.size(), resourceFileNames);
+        Utils.createSymlink(workerRoot, stormRoot, ConfigUtils.RESOURCES_SUBDIR);
+        for (String fileName : blobFileNames) {
+            Utils.createSymlink(workerRoot, stormRoot, fileName, fileName);
+        }
+    }
+
+    public void shutWorker(SupervisorData supervisorData, IWorkerManager workerManager, String workerId) throws IOException, InterruptedException{
```
jerrypeng: can we name this function to something other than shutWorker since there is already a shutdownWorker function and the similarities the names may cause confusion

jerrypeng: And why do we need this function? Why are we calling the cleanupWorker function in here and not in shutdownWorker?

hustfxj: @jerrypeng  for integration test, I can not explain it clearly. You can refer to the code as follows 
```
(defmacro capture-changed-workers
  [& body]
  `(let [launch-captured# (atom {})
         shutdown-captured# (atom {})]
     (with-var-roots [local-supervisor/launch-local-worker (mk-capture-launch-fn launch-captured#)
                      local-supervisor/shutdown-local-worker (mk-capture-shutdown-fn shutdown-captured#)]
                     ~@body
                     {:launched @launch-captured#
                      :shutdown @shutdown-captured#})))
```
```
(defn local-process []
  "Create a local process event"
  (proxy [SyncProcessEvent] []
    (launchLocalWorker [supervisorData stormId port workerId resources]
      (launch-local-worker supervisorData stormId port workerId resources))
    (shutWorker [supervisorData worker-manager workerId] (shutdown-local-worker supervisorData worker-manager workerId))))
```

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SyncSupervisorEvent.java
hustfxj: @jerrypeng  I may not follow up your means. If they contain the same elements  (without considering ordering) , it don't  shut down the worker.
···
           for (ExecutorInfo executorInfo : newExecutors) {
                if (!existExecutors.contains(executorInfo)) {
                    syncProcesses.shutWorker(supervisorData, supervisorData.getWorkerManager(), vaildPortToWorkerIds.get(port));
                    break;
                }
            }
···

jerrypeng: I think this logic will work just wasn't clear to me

hustfxj: ok, I will update it. Thank you.

PR #1185, opened 2016-03-04
[STORM-1269] port backtype.storm.daemon.common to java

bastiliu:
1. Following clojure defined records in common, "Assignment, StormBase, SupervisorInfo, ExecutorStats", were not migrated here. There are corresponding classes of Thrift. So, we can just use the thrift class in the dependency modules after each module's migration.
2. worker-context was not migrated. There is much more worker to convert the worker data between java and clojure now. But it is easy to do this after the worker data is migrated to java. So I recommend to leave the mirgation of worker context as a part of  worker module.


storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -2302,4 +2302,54 @@ public Object call() {
...
+            long value = (Long) o;
+            return (int) value;
+        } else if (o instanceof Integer) {
+            return (Integer) o;
+        } else {
+            throw new RuntimeException("Invalid value " + o.getClass().getName() + " " + o);
+        }
+    }
+
+    public static Integer parseInt(Object o, int defaultValue) {
```
hustfxj: Utils has the methods . The name is getInt()

bastiliu: Addressed

storm-core/src/jvm/org/apache/storm/daemon/StormCommon.java
```diff
@@ -0,0 +1,605 @@
...
+        }
+        return inputs;
+    }
+
+    public static String clusterId = null;
+    public static IBolt makeAckerBolt() {
+        return _instance.makeAckerBoltImpl();
+    }
+    public IBolt makeAckerBoltImpl() {
+        return new Acker();
```
abhishekagarwal87: This can be inlined.

storm-core/src/jvm/org/apache/storm/daemon/StormCommon.java
```diff
@@ -0,0 +1,605 @@
...
+     */
+    public static StormCommon setInstance(StormCommon common) {
+        StormCommon oldInstance = _instance;
+        _instance = common;
+        return oldInstance;
+    }
+
+    private static final Logger LOG = LoggerFactory.getLogger(StormCommon.class);
+
+    public static final String ACKER_COMPONENT_ID = Acker.ACKER_COMPONENT_ID;
```
abhishekagarwal87: We can directly use Acker/ACKER_COMPONENT_ID everywhere. Similarly for other stream names.

storm-core/src/jvm/org/apache/storm/daemon/StormCommon.java
```diff
@@ -0,0 +1,605 @@
...
+        return topologyName;
+    }
+
+    /**
+     * Convert topologyId to topologyName. TopologyId = topoloygName-counter-timeStamp
+     *
+     * @param topologyId
+     * @return
+     */
+    public static String topologyIdToName(String topologyId) throws InvalidTopologyException {
```
abhishekagarwal87: This and the function above can be merged.

storm-core/src/jvm/org/apache/storm/daemon/StormCommon.java
```diff
@@ -0,0 +1,605 @@
...
+                ret = topologyId.substring(0, index);
+            else
+                throw new InvalidTopologyException(topologyId + " is not a valid topologyId");
+        } else
+            throw new InvalidTopologyException(topologyId + " is not a valid topologyId");
+        return ret;
+    }
+
+    public static String getStormId(IStormClusterState stormClusterState, final String topologyName) {
+        List<String> activeTopologys = stormClusterState.activeStorms();
```
abhishekagarwal87: nit. activeTopologies

storm-core/src/jvm/org/apache/storm/daemon/StormCommon.java
abhishekagarwal87: In fact the above function can be removed since an invalid name is never expected.

storm-core/src/jvm/org/apache/storm/daemon/StormCommon.java
```diff
@@ -0,0 +1,605 @@
...
+
+        Bolt systemBoltSpec = Thrift.prepareSerializedBoltDetails(null, new SystemBolt(), outputStreams, 0, boltConf);
+        topology.put_to_bolts(Constants.SYSTEM_COMPONENT_ID, systemBoltSpec);
+    }
+
+    public static StormTopology systemTopology(Map stormConf, StormTopology topology) throws InvalidTopologyException {
+        return _instance.systemTopologyImpl(stormConf, topology);
+    }
+
+    protected StormTopology systemTopologyImpl(Map stormConf, StormTopology topology) throws InvalidTopologyException {
```
abhishekagarwal87: this function and the functions called by it can be moved to a different class SystemTopologyBuilder

storm-core/src/jvm/org/apache/storm/daemon/StormCommon.java
```diff
@@ -0,0 +1,605 @@
...
+        addSystemComponents(stormConf, ret);
+        addMetricStreams(ret);
+        addSystemStreams(ret);
+
+        validateStructure(ret);
+
+        return ret;
+    }
+
+    public static boolean hasAckers(Map stormConf) {
```
abhishekagarwal87: hasAckers and hasEventLoggers is only called in executor so they can be moved there or you can move them to ConfigUtils.java since these functions solely operate on conf.

redsanket:
upmerge

storm-core/src/clj/org/apache/storm/daemon/common.clj
```diff
@@ -86,26 +40,6 @@
 (defn new-executor-stats []
   (ExecutorStats. 0 0 0 0 0))
```
revans2: Why are we not moving ExecutorStats, SupervisorInfo, StormBase and Assignment to java?

I know we have thrift equivalents of these already.  Is the plan to move them to java as we change the daemon code to use java?  If so I am fine with it, I just want to understand for sure what is happening.

storm-core/src/clj/org/apache/storm/daemon/common.clj
```diff
@@ -408,19 +69,3 @@
                           (:default-shared-resources worker)
                           (:user-shared-resources worker)
                           ))
```
revans2: If we are leaving some methods in clojure, could you file a follow on JIRA to remove the file once we have the other code ported over to java?

revans2:
Please upmerge the metrics code has been moved to StormMetricsRegistry.java, and is causing conflicts now.

storm-core/src/jvm/org/apache/storm/daemon/StormCommon.java
```diff
@@ -0,0 +1,604 @@
...
+
+import com.codahale.metrics.MetricRegistry;
+import org.apache.storm.Config;
+import org.apache.storm.Constants;
+import org.apache.storm.Thrift;
+import org.apache.storm.cluster.IStormClusterState;
+import org.apache.storm.daemon.metrics.MetricsUtils;
+import org.apache.storm.daemon.metrics.reporters.PreparableReporter;
+import org.apache.storm.generated.*;
+import org.apache.storm.generated.StormBase;
```
revans2: Don't need both the .* import and the StormBase import.  Usually we prefer to not use .* imports, but for me it is not that big of a deal.

storm-core/src/jvm/org/apache/storm/daemon/StormCommon.java
```diff
@@ -0,0 +1,604 @@
...
+import org.apache.storm.testing.NonRichBoltTracker;
+import org.apache.storm.utils.ConfigUtils;
+import org.apache.storm.utils.IPredicate;
+import org.apache.storm.utils.ThriftTopologyUtils;
+import org.apache.storm.utils.Utils;
+import org.json.simple.JSONValue;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.*;
```
revans2: Here we don't want to use a .* import.  We can pull in the wrong thing if we move to a newer version of java.

storm-core/src/jvm/org/apache/storm/daemon/StormCommon.java
revans2: +1

storm-core/src/jvm/org/apache/storm/daemon/StormCommon.java
```diff
@@ -0,0 +1,604 @@
...
+        report.start();
+        LOG.info("Started statistics report plugin...");
+    }
+
+    public static void startMetricsReporters(Map conf) {
+        List<PreparableReporter> reporters = MetricsUtils.getPreparableReporters(conf);
+        for (PreparableReporter reporter : reporters) {
+            startMetricsReporter(reporter, conf);
+        }
+    }
```
revans2: The metrics code has moved to another file already.  these are not needed any more.

storm-core/src/jvm/org/apache/storm/daemon/StormCommon.java
revans2: In fact the above function and this function are replacing 
```
(.get_name (.stormBase storm-cluster-state  x nil)
```

where `x` is the topology id.

I get that this code is more efficient, but also more error prone, and so I would prefer we stick with the original code until we see a real reason to switch.

```
for (String topologyId: stormClusterState.activeStorms()) {
    if (topologyName.equals(stormClusterState.stormBase(topologyId, null).get_name())) {
        return topologyId;
    }
    return null;
}
```

storm-core/src/jvm/org/apache/storm/daemon/StormCommon.java
```diff
@@ -0,0 +1,604 @@
...
+        return _instance.topologyBasesImpl(stormClusterState);
+    }
+
+    protected Map<String, StormBase> topologyBasesImpl(IStormClusterState stormClusterState) {
+        List<String> activeTopologys = stormClusterState.activeStorms();
+        Map<String, StormBase> stormBases = new HashMap<String, StormBase>();
+        if (activeTopologys != null) {
+            for (String topologyId : activeTopologys) {
+                StormBase base = stormClusterState.stormBase(topologyId, null);
+                if (base != null) {
```
revans2: The original code didn't check for null here.  Is this needed?  Extra checks that fix things up like this make me nervous that be broke something somewhere and we are using this to work around it.

storm-core/src/jvm/org/apache/storm/daemon/StormCommon.java
```diff
@@ -0,0 +1,604 @@
...
+    }
+
+    public static Map<String, StormBase> topologyBases(IStormClusterState stormClusterState) {
+        return _instance.topologyBasesImpl(stormClusterState);
+    }
+
+    protected Map<String, StormBase> topologyBasesImpl(IStormClusterState stormClusterState) {
+        List<String> activeTopologys = stormClusterState.activeStorms();
+        Map<String, StormBase> stormBases = new HashMap<String, StormBase>();
+        if (activeTopologys != null) {
```
revans2: The original code didn't check for null.

storm-core/src/jvm/org/apache/storm/daemon/StormCommon.java
```diff
@@ -0,0 +1,604 @@
...
+        }
+    }
+
+    private static void validateIds(StormTopology topology) throws InvalidTopologyException {
+        List<String> componentIds = new ArrayList<String>();
+
+        for (StormTopology._Fields field : Thrift.getTopologyFields()) {
+            if (ThriftTopologyUtils.isWorkerHook(field) == false) {
+                Object value = topology.getFieldValue(field);
+                if (value != null) {
```
revans2: Is this ever really null?  I don't see the null check even implicitly in the clojure code, if it is ever null I the clojure code would throw an NPE.

storm-core/src/jvm/org/apache/storm/daemon/StormCommon.java
```diff
@@ -0,0 +1,604 @@
...
+                components.putAll(((Map) topology.getFieldValue(field)));
+            }
+        }
+        return components;
+    }
+
+    public static Map componentConf(Object component) {
+        Map<Object, Object> conf = new HashMap<Object, Object>();
+        ComponentCommon common = getComponentCommon(component);
+        if (common != null) {
```
revans2: The original code didn't check for null and would throw an NPE if we passed in a null.  We should do the same.

storm-core/src/jvm/org/apache/storm/daemon/StormCommon.java
```diff
@@ -0,0 +1,604 @@
...
+            }
+        }
+        return conf;
+    }
+
+    public static void validateBasic(StormTopology topology) throws InvalidTopologyException {
+        validateIds(topology);
+
+        List<StormTopology._Fields> spoutFields = Arrays.asList(Thrift.getSpoutFields());
+        for (StormTopology._Fields field : spoutFields) {
```
revans2: There is no reason to convert the code to a list before iterating.

```
for (StormTopology._Fields field: Thrift.getSpoutFields()) {
```

storm-core/src/jvm/org/apache/storm/daemon/StormCommon.java
```diff
@@ -0,0 +1,604 @@
...
+
+        List<String> offending = Utils.getRepeat(componentIds);
+        if (offending.isEmpty() == false) {
+            throw new InvalidTopologyException("Duplicate component ids: " + offending);
+        }
+    }
+
+    private static boolean isEmptyInputs(ComponentCommon common) {
+        if (common == null) {
+            return true;
```
revans2: The original code would throw an NPE if this was null.  In this case I can see it being a minor nit, but I still would prefer to keep the code the same if possible.

storm-core/src/jvm/org/apache/storm/daemon/StormCommon.java
```diff
@@ -0,0 +1,604 @@
...
+                    }
+                }
+            }
+        }
+
+        Map<String, Object> componentMap = allComponents(topology);
+        for (Object componentObj : componentMap.values()) {
+            Map conf = componentConf(componentObj);
+            ComponentCommon common = getComponentCommon(componentObj);
+            if (common != null) {
```
revans2: This null check was not in the original code.

storm-core/src/jvm/org/apache/storm/daemon/StormCommon.java
```diff
@@ -0,0 +1,604 @@
...
+        }
+        return outputFields;
+    }
+
+    public static void validateStructure(StormTopology topology) throws InvalidTopologyException {
+        Map<String, Object> componentMap = allComponents(topology);
+        for (Map.Entry<String, Object> entry : componentMap.entrySet()) {
+            String componentId = entry.getKey();
+            ComponentCommon common = getComponentCommon(entry.getValue());
+            if (common != null) {
```
revans2: This null check was not in the original code.

storm-core/src/jvm/org/apache/storm/daemon/StormCommon.java
```diff
@@ -0,0 +1,604 @@
...
+                Map<GlobalStreamId, Grouping> inputs = common.get_inputs();
+                for (Map.Entry<GlobalStreamId, Grouping> input : inputs.entrySet()) {
+                    String sourceStreamId = input.getKey().get_streamId();
+                    String sourceComponentId = input.getKey().get_componentId();
+                    if(componentMap.keySet().contains(sourceComponentId) == false) {
+                        throw new InvalidTopologyException("Component: [" + componentId + "] subscribes from non-existent component [" + sourceComponentId + "]");
+                    }
+
+                    ComponentCommon sourceComponent = getComponentCommon(componentMap.get(sourceComponentId));
+                    if (sourceComponent == null || sourceComponent.get_streams().containsKey(sourceStreamId) == false) {
```
revans2: The null check for sourceComponent was not in the original code, and if it were null it would not mean that we subscribed to a non-existent stream.  It would mean that our topology builder didn't do its job correctly.

storm-core/src/jvm/org/apache/storm/daemon/StormCommon.java
```diff
@@ -0,0 +1,604 @@
...
+                    }
+
+                    Grouping grouping = input.getValue();
+                    if (Thrift.groupingType(grouping) == Grouping._Fields.FIELDS) {
+                        List<String> fields = grouping.get_fields();
+                        Map<String, StreamInfo> streams = sourceComponent.get_streams();
+                        Set<String> sourceOutputFields = getStreamOutputFields(streams);
+                        if (sourceOutputFields.containsAll(fields) == false) {
+                            throw new InvalidTopologyException("Component: [" + componentId + "] subscribes from stream: [" + sourceStreamId  +"] of component " +
+                                    "[" + sourceComponentId + "] + with non-existent fields: " + fields);
```
revans2: In the original code the error message was just for the diff-fields.  The ones that were a problem, not all of the fields.

storm-core/src/jvm/org/apache/storm/daemon/StormCommon.java
```diff
@@ -0,0 +1,604 @@
...
+                if (taskNum > 0 && parallelismHintNum <= 0) {
+                    throw new InvalidTopologyException("Number of executors must be greater than 0 when number of tasks is greater than 0");
+                }
+            }
+        }
+    }
+
+    private static Set<String> getStreamOutputFields(Map<String, StreamInfo> streams) {
+        Set<String> outputFields = new HashSet<String>();
+        if (streams != null) {
```
revans2: In the original code if streams was null the return value was null.

storm-core/src/jvm/org/apache/storm/daemon/StormCommon.java
revans2: No because it is mocked out in some cases.

storm-core/src/jvm/org/apache/storm/daemon/StormCommon.java
```diff
@@ -0,0 +1,604 @@
...
+            common.put_to_streams(ACKER_INIT_STREAM_ID, Thrift.outputFields(Arrays.asList("id", "init-val", "spout-task")));
+            common.put_to_inputs(Utils.getGlobalStreamId(ACKER_COMPONENT_ID, ACKER_ACK_STREAM_ID), Thrift.prepareDirectGrouping());
+            common.put_to_inputs(Utils.getGlobalStreamId(ACKER_COMPONENT_ID, ACKER_FAIL_STREAM_ID), Thrift.prepareDirectGrouping());
+        }
+
+        topology.put_to_bolts(ACKER_COMPONENT_ID, acker);
+    }
+
+    public static ComponentCommon getComponentCommon(Object component) {
+        if (component == null) {
```
revans2: This null check was not in the original code, but it is a minor nit here.

storm-core/src/jvm/org/apache/storm/daemon/StormCommon.java
```diff
@@ -0,0 +1,604 @@
...
+        } else if (component instanceof Bolt) {
+            common = ((Bolt) component).get_common();
+        }
+        return common;
+    }
+
+    public static void addMetricStreams(StormTopology topology) {
+        for (Object component : allComponents(topology).values()) {
+            ComponentCommon common = getComponentCommon(component);
+            if (common != null) {
```
revans2: This null check was not in the original code.

storm-core/src/jvm/org/apache/storm/daemon/StormCommon.java
```diff
@@ -0,0 +1,604 @@
...
+                StreamInfo streamInfo = Thrift.outputFields(Arrays.asList("task-info", "data-points"));
+                common.put_to_streams(Constants.METRICS_STREAM_ID, streamInfo);
+            }
+        }
+    }
+
+    public static void addSystemStreams(StormTopology topology) {
+        for (Object component : allComponents(topology).values()) {
+            ComponentCommon common = getComponentCommon(component);
+            if (common != null) {
```
revans2: Again no null check in original code.

storm-core/src/jvm/org/apache/storm/daemon/StormCommon.java
```diff
@@ -0,0 +1,604 @@
...
+    public static void addEventLogger(Map conf, StormTopology topology) {
+        Integer numExecutors = Utils.getInt(conf.get(Config.TOPOLOGY_EVENTLOGGER_EXECUTORS), Utils.getInt(conf.get(Config.TOPOLOGY_WORKERS)));
+        HashMap<String, Object> componentConf = new HashMap<String, Object>();
+        componentConf.put(Config.TOPOLOGY_TASKS, numExecutors);
+        componentConf.put(Config.TOPOLOGY_TICK_TUPLE_FREQ_SECS, Utils.getInt(conf.get(Config.TOPOLOGY_MESSAGE_TIMEOUT_SECS)));
+        Bolt eventLoggerBolt = Thrift.prepareSerializedBoltDetails(eventLoggerInputs(topology), new EventLoggerBolt(), null, numExecutors, componentConf);
+
+        for(Object component : allComponents(topology).values()) {
+            ComponentCommon common = getComponentCommon(component);
+            if (common != null) {
```
revans2: No null check in the original code.

storm-core/src/jvm/org/apache/storm/daemon/StormCommon.java
```diff
@@ -0,0 +1,604 @@
...
+        } else {
+            return false;
+        }
+    }
+
+    public static int numStartExecutors(Object component) throws InvalidTopologyException {
+        ComponentCommon common = getComponentCommon(component);
+        if (common == null) {
+            throw new InvalidTopologyException("unknown component type " + component.getClass().getName());
+        }
```
revans2: This null check didn't exist in the original code.

storm-core/src/jvm/org/apache/storm/daemon/StormCommon.java
```diff
@@ -0,0 +1,604 @@
...
+        return parallelismHintNum;
+    }
+
+    public static Map stormTaskInfo(StormTopology userTopology, Map stormConf) throws InvalidTopologyException {
+        return _instance.stormTaskInfoImpl(userTopology, stormConf);
+    }
+    /*
+     * Returns map from task -> componentId
+     */
+    protected Map stormTaskInfoImpl(StormTopology userTopology, Map stormConf) throws InvalidTopologyException {
```
revans2: Can we add the generics to the return type?
```
Map<Integer, String>
```

storm-core/src/jvm/org/apache/storm/daemon/StormCommon.java
```diff
@@ -0,0 +1,604 @@
...
+        Map<Integer, String> taskIdToComponentId = new HashMap<Integer, String>();
+
+        StormTopology systemTopology = systemTopology(stormConf, userTopology);
+        Map<String, Object> components = allComponents(systemTopology);
+        Map<String, Integer> componentIdToTaskNum = new TreeMap<String, Integer>();
+        for (Map.Entry<String, Object> entry : components.entrySet()) {
+            Map conf = componentConf(entry.getValue());
+            Object taskNum = conf.get(Config.TOPOLOGY_TASKS);
+            if (taskNum != null) {
```
revans2: This null check was not in the original code.

storm-core/src/jvm/org/apache/storm/daemon/StormCommon.java
```diff
@@ -0,0 +1,604 @@
...
+                    aznHandler = (IAuthorizer) aznClass.newInstance();
+                    if (aznHandler != null) {
+                        aznHandler.prepare(conf);
+                    }
+                    LOG.debug("authorization class name:{}, class:{}, handler:{}",klassName, aznClass, aznHandler);
+                }
+            }
+        } catch (Exception e) {
+            LOG.error("Failed to make authorization handler, klassName:{}", klassName);
+        }
```
revans2: The original code did not catch any exceptions it just threw them.  We should do the same.

storm-core/test/clj/org/apache/storm/messaging/netty_integration_test.clj
```diff
@@ -1,4 +1,3 @@
-
```
revans2: Please revert the change to this file.  Not needed.

storm-core/test/jvm/org/apache/storm/utils/staticmocking/CommonInstaller.java
```diff
@@ -0,0 +1,38 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.storm.utils.staticmocking;
+
+import org.apache.storm.daemon.StormCommon;
+
+public class CommonInstaller implements AutoCloseable {
```
revans2: Do we need this and StormCommonInstaller???

storm-core/src/clj/org/apache/storm/daemon/executor.clj
```diff
@@ -536,17 +536,17 @@
...
                                                                            spout-id tuple-finished-info time-delta "FAIL-STREAM" id)
                                       )))
                                 ;; TODO: on failure, emit tuple to failure stream
                                 ))))
         receive-queue (:receive-queue executor-data)
         event-handler (mk-task-receiver executor-data tuple-action-fn)
-        has-ackers? (has-ackers? storm-conf)
-        has-eventloggers? (has-eventloggers? storm-conf)
+        has-ackers? (clojurify-structure (StormCommon/hasAckers storm-conf))
+        has-eventloggers? (clojurify-structure (StormCommon/hasEventLoggers storm-conf))
```
revans2: I don't think these should need the `clojurify-structure` calls.  Were you seeing errors without it?

storm-core/src/clj/org/apache/storm/daemon/executor.clj
```diff
@@ -742,7 +742,7 @@
                                                                (.getSourceComponent tuple)
                                                                (.getSourceStreamId tuple)
                                                                delta)))))))
-        has-eventloggers? (has-eventloggers? storm-conf)
+        has-eventloggers? (clojurify-structure (StormCommon/hasEventLoggers storm-conf))
```
revans2: Here too I don't think `clojurify-structure` is needed.  Were you seeing errors without it?

storm-core/src/clj/org/apache/storm/daemon/worker.clj
```diff
@@ -254,6 +254,9 @@
         (log-error e "Error when processing event")
         (Utils/exitProcess 20 "Error when processing an event")))))
 
+(defn executor->tasks [executor-id]
+  clojurify-structure (StormCommon/executorIdToTasks executor-id))
```
revans2: Either we need to actually call `clojurify-structure` or we need to remove it.

revans2:
For the most part things look really good.  Just a lots of null checks that were not in the original code.  In these cases where we are validating the structure of the topology I really would prefer to see the code blow up with an NPE in those cases instead of silently skip over them.

storm-core/src/jvm/org/apache/storm/daemon/StormCommon.java
bastiliu: Yes, it is for mocking in with-tracked-cluster.

storm-core/src/jvm/org/apache/storm/daemon/StormCommon.java
bastiliu: OK. I will update it.

storm-core/src/jvm/org/apache/storm/daemon/StormCommon.java
bastiliu: @revans2  Yes, the major purpose here is to reduce the access to zk. Anyway, I am OK for keeping consistent with the original code.

storm-core/src/clj/org/apache/storm/daemon/common.clj
bastiliu: Yes, I recommend to just use the thrift classes directly when we work on relative modules. It can save much effort now for the converting between clojure and jave. Please refer to my comments at the beginning of this PR.

storm-core/src/clj/org/apache/storm/daemon/common.clj
bastiliu: OK, I will create one.

storm-core/src/jvm/org/apache/storm/daemon/StormCommon.java
bastiliu: I prefer to just keep the same with origin code since these functions are related to "topology component", not a utils function. It is better to keep these kind functions in common.

storm-core/src/jvm/org/apache/storm/daemon/StormCommon.java
bastiliu: Addressed

storm-core/test/jvm/org/apache/storm/utils/staticmocking/CommonInstaller.java
bastiliu: CommonInstaller is used for mocking in clojure testing. It is the same way of mocking as Utils and ConfigUtils. I am not sure if there is a better solution. Any suggestion?

bastiliu:
I have updated the code according to your comments, and remove the unnecessary null checks.

storm-core/src/jvm/org/apache/storm/daemon/StormCommon.java
revans2: That sounds fine, but could we do it in another follow on JIRA?  This is really close to being done.

storm-core/test/jvm/org/apache/storm/utils/staticmocking/CommonInstaller.java
revans2: I just thought that we didn't need both.  We could update the tests to use StormCommonInstaller and remove this.

revans2:
The code looks really good an I am +1 for merging it in as is.  There are still a few nits though. one with CommonInstaller and another that @arunmahadevan pointed out for SystemTopologyBuilder that we need to come to a resolution on.

Also you need to upmerge the drpc changes that went in have caused a minor conflict.

bastiliu:
@revans2 @abhishekagarwal87 Thanks for your review. CommonInstaller was removed, and the drpc changes have been merged.  For SystemTopologyBuilder, maybe it is better to follow up in a new JIRA for enhancement or code restructuring, since this JIRA mainly focus on migration.

abhishekagarwal87:
I am fine with refactoring system topology builder in separate JIRA.

revans2:
+1 looks good to me

PR #1186, opened 2016-03-04
add private constructor to KafkaUtils

lujinhong:
Since all of the methods and fields are static, I think it is a good idea to add a private constructor to KafkaUtils to  avoid been constructed.


knusbaum:
+1. Not really necessary, but fine with me.

PR #1189, opened 2016-03-06
[STORM-1606] print the  information of testcase which is on failure

hustfxj:
the api "find(match)" of Python ElementTree finds the first sub element matching match. So the match is sub element, but not sub string

### DIFF
--- a/dev-tools/travis/print-errors-from-test-reports.py
+++ b/dev-tools/travis/print-errors-from-test-reports.py
@@ -55,6 +55,10 @@ def print_error_reports_from_report_file(file_path):
         if fail is not None:
             print_detail_information(testcase, fail)
 
+        failure = testcase.find("failure")
+        if failure is not None:
+            print_detail_information(testcase, failure)
+
 
 def main(report_dir_path):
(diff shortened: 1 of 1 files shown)


dev-tools/travis/print-errors-from-test-reports.py
```diff
@@ -55,6 +55,10 @@ def print_error_reports_from_report_file(file_path):
         if fail is not None:
             print_detail_information(testcase, fail)
 
+        failure = testcase.find("failure")
```
unsleepy22: I would suspect the validity of this change since "fail" is a substring of "failure", the script should not miss "failure"s by searching "fail".

hustfxj: @unsleepy22  the  api "find(match)" of Python ElementTree finds the first sub element matching match. So the match is sub element, but not sub string

unsleepy22:
Sorry, I didn't take a deep look. +1

revans2:
+1 good catch.  I was wondering why it seemed to work on the clojure code but not on junit.  Thanks for doing this.

PR #1193, opened 2016-03-08
[STORM-1612] Remove hardcoded port, and unreferenced global from netty_unit_test

ppoulosk:
Just remove the unnecessary hardcoded port, so that the call will return an ephemeral port, and remove unused global "port".


ppoulosk:
It falls back to ephemeral if preferred is bound, so this is not needed.

PR #1194, opened 2016-03-08
[STORM-1609] Netty Client is not best effort delivery on failed Connection

redsanket:
If Worker-A has connection to Worker-B that is unused ( and if Worker-B restarted), we drop messages because Channel is not in good state. Can we avoid message drop until we succeed in making new connection or a timeout?

### DIFF
--- a/storm-core/src/jvm/org/apache/storm/messaging/netty/Client.java
+++ b/storm-core/src/jvm/org/apache/storm/messaging/netty/Client.java
@@ -75,7 +75,7 @@ public class Client extends ConnectionWithStatus implements IStatefulObject, ISa
     private static final Logger LOG = LoggerFactory.getLogger(Client.class);
     private static final String PREFIX = "Netty-Client-";
     private static final long NO_DELAY_MS = 0L;
-    private static Timer timer;
+    private static final Timer timer = new Timer("Netty-ChannelAlive-Timer", true);
 
     private final Map stormConf;
     private final StormBoundedExponentialBackoffRetry retryPolicy;
@@ -173,13 +173,6 @@ public class Client extends ConnectionWithStatus implements IStatefulObject, ISa
     private void launchChannelAliveThread() {
         // netty TimerTask is already defined and hence a fully
         // qualified name
-        if (timer == null) {
-            synchronized (Client.class) {
-                if (timer == null) {
-                    timer = new Timer("Netty-ChannelAlive-Timer", true);
-                }
-            }
-        }
         timer.schedule(new java.util.TimerTask() {
             public void run() {
                 try {
--- a/storm-core/src/jvm/org/apache/storm/messaging/netty/Client.java
+++ b/storm-core/src/jvm/org/apache/storm/messaging/netty/Client.java
@@ -23,6 +23,7 @@ import java.util.Iterator;
 import java.util.Collection;
 import java.util.Map;
 import java.util.HashMap;
+import java.util.Timer;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
@@ -74,6 +75,7 @@ public class Client extends ConnectionWithStatus implements IStatefulObject, ISa
     private static final Logger LOG = LoggerFactory.getLogger(Client.class);
     private static final String PREFIX = "Netty-Client-";
     private static final long NO_DELAY_MS = 0L;
+    private static Timer timer;
 
     private final Map stormConf;
     private final StormBoundedExponentialBackoffRetry retryPolicy;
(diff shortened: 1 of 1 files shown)


revans2:
+1 looks good to me.

ptgoetz:
+1. We should also apply this to 1.x-branch.

storm-core/src/jvm/org/apache/storm/messaging/netty/Client.java
```diff
@@ -151,10 +164,35 @@
...
         batcher = new MessageBuffer(messageBatchSize);
     }
 
+    /**
+     * This thread helps us to check for channel connection periodically.
+     * This is performed just to know whether the destination address
+     * is alive or attempts to refresh connections if not alive. This
+     * solution is better than what we have now in case of a bad channel.
+     */
+    private void launchChannelAliveThread() {
```
d2r: Let's move the code in the static block above down here at the beginning of `launchChannelAliveThread`, so that as @revans2 suggests, the stack trace will not be so confusing.

```Java
if (timer == null) {
  synchronized (Client.class) {
    if (timer == null) {
      timer = new Timer("Netty-ChannelAlive-Timer", true);
    }
  }
}
```

storm-core/src/jvm/org/apache/storm/messaging/netty/Client.java
```diff
@@ -151,10 +164,35 @@
...
+        // netty TimerTask is already defined and hence a fully
+        // qualified name
+        timer.schedule(new java.util.TimerTask() {
+            public void run() {
+                try {
+                    LOG.debug("running timer task, address {}", dstAddress);
+                    if(closing) {
+                        this.cancel();
```
d2r: Just to be cleaner and avoid a possible race, let's return here so we don't call `getConnectedChannel`.

storm-core/src/jvm/org/apache/storm/messaging/netty/Client.java
```diff
@@ -108,6 +110,13 @@
     private final AtomicInteger messagesLost = new AtomicInteger(0);
 
     /**
+     * Periodically checks for connected channel in order to avoid loss
+     * of messages
+     */
+    private final long CHANNEL_ALIVE_INTERVAL_MS = 30000L;
+
```
d2r: minor: extra line

d2r:
Had some usability concerns, but otherwise looks good.

storm-core/src/jvm/org/apache/storm/messaging/netty/Client.java
redsanket: Good suggestion, I will do that thank you d2r

storm-core/src/jvm/org/apache/storm/messaging/netty/Client.java
redsanket: fixed

storm-core/src/jvm/org/apache/storm/messaging/netty/Client.java
redsanket: fixed

storm-core/src/jvm/org/apache/storm/messaging/netty/Client.java
redsanket: fixed

redsanket:
@d2r Addressed your comments

d2r:
Looks good +1.
I see unrelated failures in supervisor-test and transactional-test.

storm-core/src/jvm/org/apache/storm/messaging/netty/Client.java
```diff
@@ -151,10 +159,43 @@
...
+    /**
+     * This thread helps us to check for channel connection periodically.
+     * This is performed just to know whether the destination address
+     * is alive or attempts to refresh connections if not alive. This
+     * solution is better than what we have now in case of a bad channel.
+     */
+    private void launchChannelAliveThread() {
+        // netty TimerTask is already defined and hence a fully
+        // qualified name
+        if (timer == null) {
```
bugzmanov: This looks a lot like racy(incorrect) version of:
https://en.wikipedia.org/wiki/Double-checked_locking#Usage_in_Java

So you might end up with two instances of timer running 1 job each

redsanket: @d2r, @revans2  looking at it it seems a valid explanation. Would like to know whether it is better to revert to earlier static initialization

redsanket: @d2r @revan2 reverted back to static initialization

hustfxj:
+1. I have a suggestion.  we drop messages because Channel is not in good state. I hope we should put the messages into buffer when Channel is not in good state. Of course, we should drop the buffer if the channel is still not in good state.

redsanket:
Agree, that is a bigger change and I will have a follow up jira for it. Thanks

vesense:
LGTM

harshach:
+1

PR #1196, opened 2016-03-09
[STORM-1605] use '/usr/bin/env python' to check python version

unsleepy22:
Current python version check is hard-coded and it cannot detect python 2.7.x versions. Changed to /usr/bin/env to keep consistent with storm.py

### DIFF
--- a/bin/storm
+++ b/bin/storm
@@ -30,16 +30,8 @@ while [ -h "${PRG}" ]; do
   fi
 done
 
-# find python >= 2.6
-if [ -a /usr/bin/python2.6 ]; then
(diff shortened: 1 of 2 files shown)


hustfxj:
@unsleepy22  It loos good to me.  Many *.py files start with "#!/usr/bin/python" in storm . So do you think we should update them together or not?

unsleepy22:
@hustfxj thanks for pointing out, I added storm.py too. But for the other files, I prefer to leave them alone since they're not part of storm-core.

revans2:
+1 looks good to me.

harshach:
+1

PR #1199, opened 2016-03-09
STORM-1616: Add RAS API for Trident

knusbaum:


storm-core/src/jvm/org/apache/storm/trident/Stream.java
```diff
@@ -124,6 +124,31 @@ public Stream parallelismHint(int hint) {
     }
 
     /**
+     * Sets the CPU Load resource for the current node
+     */
+    public Stream setCPULoad(Number load) {
```
arunmahadevan: A `Stream` should ideally have only streaming operations but it seems we are bringing the topology configuration into the stream api which does not look very clean. Cant this be part of the `TridentTopology` or passed during `submitTopology` ? If this should really be part of stream, can we have a single method like `nodeConfig(NodeConfig config)` and have all these methods as part of the `NodeConfig` so that we have only one extra method which is not really a stream operation?

storm-core/src/jvm/org/apache/storm/trident/TridentTopology.java
```diff
@@ -409,16 +427,26 @@ public StormTopology build() {
...
                 }
-                builder.setSpout(spoutIds.get(sn), sn.streamId, sn.txId, s, parallelism, batchGroupMap.get(sn));
+                builder.setSpout(spoutIds.get(sn), sn.streamId, sn.txId, s, parallelism, batchGroupMap.get(sn))
+                    .setMemoryLoad(onHeap, offHeap)
+                    .setCPULoad(cpuLoad);
             }
         }
-        
+
         for(Group g: mergedGroups) {
```
jerrypeng: can we use a different variable name then "g"? The name was kind of confusing we I was looking through the code

storm-core/src/jvm/org/apache/storm/trident/Stream.java
knusbaum: I can look at adding a method for including a component -> resources map when creating a trident topology. 

As for the others:
There's not really a clean way to add this to TridentTopology, and we wanted some way to specify resources similar to RAS's API. Stream is the cleanest place for it, and it works just like `parallelismHint(...)` does.

Having one method, 'nodeConfig(...)' with a `NodeConfig` object would reduce the number of new methods in Stream, like you said. I don't really see that as looking any better for the end-user, though. Consider this:
```java
TridentTopology topo = new TridentTopology();
TridentState wordCounts = 
    topology.newStream("spout1", spout)
        .setCPULoad(50)
        .setMemoryLoad(1024)
        .each(new Fields("sentence"), new Split(), new Fields("word"))
        .setCPULoad(100)
        .setMemoryLoad(256)
        .groupBy(new Fields("word"))
        .persistentAggregate(new MemorymapState.Factory(), new Count(), new Fields("count"))
        .parallelismHint(6);
```

Would become this:
```java
TridentTopology topo = new TridentTopology();
TridentState wordCounts = 
    topology.newStream("spout1", spout)
        .nodeConfig(new NodeConfig()
                        .setCPULoad(50)
                        .setMemoryLoad(1024))
        .each(new Fields("sentence"), new Split(), new Fields("word"))
        .nodeConfig(new NodeConfig()
                        .setCPULoad(100)
                        .setMemoryLoad(256))
        .groupBy(new Fields("word"))
        .persistentAggregate(new MemorymapState.Factory(), new Count(), new Fields("count"))
        .parallelismHint(6);
```

I may be missing the benefit here, but as far as I can tell, users shouldn't really need to be aware of Nodes at all. They're supposed to be internal to Trident.

storm-core/src/jvm/org/apache/storm/trident/TridentTopology.java
knusbaum: I would be very happy to change this terrible name if it were in new code, but I don't want to start refactoring unrelated things in this PR.

storm-core/src/jvm/org/apache/storm/topology/ResourceDeclarer.java
```diff
@@ -0,0 +1,24 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.storm.topology;
+
+public interface ResourceDeclarer <T extends ResourceDeclarer> {
+    T setMemoryLoad(Number onHeap);
```
knusbaum: Number is just the class RAS uses to hold resource counts. I'm not sure what led to that decision. This interface was pulled out of `ComponentConfigurationDeclarer` (which now inherits this), so the method names and signatures should remain the same.

storm-core/src/jvm/org/apache/storm/trident/TridentTopology.java
```diff
@@ -431,6 +459,52 @@ public StormTopology build() {
...
+
+        Number  onHeap = res.get(Config.TOPOLOGY_COMPONENT_RESOURCES_ONHEAP_MEMORY_MB);
+        Number offHeap = res.get(Config.TOPOLOGY_COMPONENT_RESOURCES_OFFHEAP_MEMORY_MB);
+        Number cpuLoad = res.get(Config.TOPOLOGY_COMPONENT_CPU_PCORE_PERCENT);
+
+        if(onHeap == null) {
+            onHeap = onHeapDefault;
+        }
+        else {
+            onHeap = Math.max(onHeap.doubleValue(), onHeapDefault.doubleValue());
```
jerrypeng: Why are we taking the max of those two values here? Is it not ok for a user to set memory onheap memory requirement to be less than the default

storm-core/src/jvm/org/apache/storm/trident/TridentTopology.java
```diff
@@ -431,6 +459,52 @@ public StormTopology build() {
...
+        }
+
+        if(offHeap == null) {
+            offHeap = offHeapDefault;
+        }
+        else {
+            offHeap = Math.max(offHeap.doubleValue(), offHeapDefault.doubleValue());
```
jerrypeng: same question here

storm-core/src/jvm/org/apache/storm/trident/TridentTopology.java
```diff
@@ -431,6 +459,52 @@ public StormTopology build() {
...
+        }
+
+        if(cpuLoad == null) {
+            cpuLoad = cpuLoadDefault;
+        }
+        else {
+            cpuLoad = Math.max(cpuLoad.doubleValue(), cpuLoadDefault.doubleValue());
```
jerrypeng: same question here

storm-core/test/clj/integration/org/apache/storm/trident/integration_test.clj
```diff
@@ -281,6 +285,86 @@
...
+              (each (fields "word") add-bang (fields "word!"))
+              (parallelismHint 10)
+              (setCPULoad 50)
+              (setMemoryLoad 1024)
+              (groupBy (fields "word!"))
+              (persistentAggregate (memory-map-state) (Count.) (fields "count"))
+              (setCPULoad 100)
+              (setMemoryLoad 2048)))
+        (with-topology [cluster topo storm-topo]
+;          (log-message "\n")
```
jerrypeng: delete commented out code?

storm-core/test/clj/integration/org/apache/storm/trident/integration_test.clj
```diff
@@ -281,6 +285,86 @@
...
+                     20.0))
+
+              (is (= (-> (json-confs "$spoutcoord-spout-words")
+                       (get TOPOLOGY-COMPONENT-CPU-PCORE-PERCENT))
+                     20.0)))
+
+            (testing "bolt combinations"
+              (is (= (-> (json-confs "b-1")
+                         (get TOPOLOGY-COMPONENT-RESOURCES-ONHEAP-MEMORY-MB))
+                     1536.0))
```
jerrypeng: can you put a comment on how this number of calculated (i.e. 1024 + 512)

storm-core/src/jvm/org/apache/storm/trident/operation/DefaultResourceDeclarer.java
```diff
@@ -0,0 +1,62 @@
...
+    public DefaultResourceDeclarer setCPULoad(Number amount) {
+        if(amount != null) {
+            amount = amount.doubleValue();
+            resources.put(Config.TOPOLOGY_COMPONENT_CPU_PCORE_PERCENT, amount);
+        }
+        return this;
+    }
+
+    @Override
+    public Map<String, Number> getResources() {
```
jerrypeng: Why do we need getResources()? thus why DefaultResourceDeclarer needs to implement ITridentResource?

storm-core/src/jvm/org/apache/storm/trident/Stream.java
```diff
@@ -124,6 +124,31 @@ public Stream parallelismHint(int hint) {
     }
 
     /**
+     * Sets the CPU Load resource for the current node
```
satishd: minor nit: You may want to say this configuration is for the current operation instead of Node. Node representation is internal as you have mentioned in other comment. 

"Sets the CPU Load resource for the current operation"

storm-core/src/jvm/org/apache/storm/trident/TridentTopology.java
```diff
@@ -394,11 +395,28 @@ public StormTopology build() {
...
         for(SpoutNode sn: spoutNodes) {
             Integer parallelism = parallelisms.get(grouper.nodeGroup(sn));
+
+            Map<String, Number> spoutRes = null;
+            if(sn instanceof ITridentResource) {
+                spoutRes = mergeDefaultResources(((ITridentResource)sn).getResources(), defaults);
+            }
+            else {
+                spoutRes = mergeDefaultResources(null, defaults);
+            }
```
satishd: sn is always an instance of ITridentResource and you can have simple null check over here. You may want to replace if/else block with the below.
``` java
            Map<String, Number> spoutRes = mergeDefaultResources((sn != null ? sn.getResources() : null), defaults);
```

storm-core/src/jvm/org/apache/storm/trident/TridentTopology.java
satishd:
``` java
    /**
     * The maximum amount of memory an instance of a spout/bolt will take on heap. This enables the scheduler
     * to allocate slots on machines with enough available memory. A default value will be set for this config if user does not override
     */
    @isPositiveNumber(includeZero = true)
    public static final String TOPOLOGY_COMPONENT_RESOURCES_ONHEAP_MEMORY_MB = "topology.component.resources.onheap.memory.mb";
```
Default value should be taken into account if load configuration is not set. It can be considered default value is over ridden when these resource load configuration values are set on stream operations. So, it should take that value instead of taking cap from the default value.

storm-core/src/jvm/org/apache/storm/trident/Stream.java
arunmahadevan: > I can look at adding a method for including a component -> resources map when creating a trident topology.
It would be best if we could have these settings as a part of the topology than the stream.

> I may be missing the benefit here, but as far as I can tell, it doesn't read any cleaner for users, The Stream api would have only this one extra non-stream-operation method. E.g. when the user browses the javadocs or when the IDE pops up the methods of the stream.

>and IMO, users shouldn't really need to be aware of Nodes at all. They're supposed to be internal to Trident.
Yes, its ideal if its part of the topology configuration. In my opinion if we are exposing config via the stream api, its better to let it standout than attempting to hide it in with the other streaming operations because the user needs to be aware of what he is doing. `nodeConfig` was only a suggestion, it could be named appropriately.

storm-core/src/jvm/org/apache/storm/trident/TridentTopology.java
knusbaum: Good catch. I think we can ignore the null check even, though. We don't insert nulls into the spoutNodes collection, and if there were some in there somehow, lots of stuff would blow up.

storm-core/src/jvm/org/apache/storm/trident/operation/DefaultResourceDeclarer.java
knusbaum: it provides an interface to get the resources back out of something extending DefaultResourceDeclarer. This is used in TridentTopology.

storm-core/src/jvm/org/apache/storm/trident/Stream.java
satishd: It is better to have these configs to be set at respective operation level in a stream with intuitive API than setting at topology level. Below API is suggested which may look more verbose than setXXXLoad method but it clearly conveys what it is for.   
But we should not have API for this configuration at topology level as it requires to name these operations and use those names with load configs in topology looks to be clunky. 

``` java
TridentTopology topo = new TridentTopology();
TridentState wordCounts = 
    topology.newStream("spout1", spout)
        .withLoadConfig(new LoadConfigBuilder().
                        .withCPULoad(50)
                        .withMemoryLoad(1024).build())
        .each(new Fields("sentence"), new Split(), new Fields("word"))
         .withLoadConfig(new LoadConfigBuilder().
                        .withCPULoad(20)
                        .withMemoryLoad(256).build())
        .groupBy(new Fields("word"))
        .persistentAggregate(new MemorymapState.Factory(), new Count(), new Fields("count"))
        .parallelismHint(6);
```

storm-core/src/jvm/org/apache/storm/trident/operation/DefaultResourceDeclarer.java
jerrypeng: Sorry mis read the code.  Thought the code returned an empty map

storm-core/src/jvm/org/apache/storm/trident/Stream.java
arunmahadevan: Actually naming the operations wherever specific config needs to be applied looks cleaner and it separates out the streaming operations from the config specification.

```java
TridentTopology topo = new TridentTopology();
TridentState wordCounts = 
    topology.newStream("spout1", spout)
        .name("spout1")
        .each(new Fields("sentence"), new Split(), new Fields("word"))
        .name("split_operation")
        .groupBy(new Fields("word"))
        .persistentAggregate(new MemorymapState.Factory(), new Count(), new Fields("count"))
        .name("aggregate");
```
And then
```java
topology.setConfig("spout1", Config.cpuLoad(20).memoryLoad(1024))
             .setConfig("split_operation", Config.cpuLoad(20).memoryLoad(256))
             .setConfig("aggregate", Config.parallelismHint(6));
```
The concern with directly adding each config api in Stream is that we might want to add more configs in future and then it would mess up the Stream api.

storm-core/src/jvm/org/apache/storm/trident/TridentTopology.java
knusbaum: Since each component can be the sum of a bunch of operations, the simplest thing to do is take the max.

What we don't want to happen is for the user to set low resources on one operation, and have that be set for the compound. If *any* component isn't set, we want to use the default. 

Right now the code doesn't check that. It just sums them up and takes the max for simplicity's sake. We could do something to check if every node had set a low value and if the sum of those is less than the default, keep it, but that doesn't seem like it has great benefits for the added complexity. The defaults are very low as it is, and I'm not sure how many people are setting their components with lower than default resource values.

storm-core/src/jvm/org/apache/storm/trident/Stream.java
knusbaum: +1 arunmahadevan. I agree.

knusbaum:
arunmahadevan added a note 3 days ago

> Actually naming the operations wherever specific config needs to be applied looks cleaner and it separates out the streaming operations from the config specification.

>
```
TridentTopology topo = new TridentTopology();
TridentState wordCounts = 
    topology.newStream("spout1", spout)
        .name("spout1")
        .each(new Fields("sentence"), new Split(), new Fields("word"))
        .name("split_operation")
        .groupBy(new Fields("word"))
        .persistentAggregate(new MemorymapState.Factory(), new Count(), new Fields("count"))
        .name("aggregate");
```
>And then
```
topology.setConfig("spout1", Config.cpuLoad(20).memoryLoad(1024))
             .setConfig("split_operation", Config.cpuLoad(20).memoryLoad(256))
             .setConfig("aggregate", Config.parallelismHint(6));
```
>The concern with directly adding each config api in Stream is that we might want to add more configs in future and then it would mess up the Stream api.'


This does not work, and would require an overhaul of the way naming works right now. First off, your example won't compile. `persistentAggregate` returns a `TridentState`, which doesn't have a `name` method. Okay, well we can move the name call up to just below the groupBy. It's a little confusing being right in the middle, but groupBy returns a `GroupedStream`, which has a name method. Unfortunately, you end up with all your bolts being named "b-*-aggregate" because 'GroupedStream' just renames the stream before it. In fact, except for differentiating between the spout and body, I don't see any existing way to give different sections different names, and doing so is not a simple change, and not one that is going to be backwards-compatible.

I agree with your point in principle, but how much do we want to change the existing API to accommodate it?

ptgoetz:
I lean toward making this part of the Stream API, just like parallelismHint(). In both cases it is about requesting cluster resources, (I.e. Threads/CPU/Memory).

I'm strongly against (though open to new ideas) of making this a topology-level API as proposed earlier. That approach involves a lot of "magic strings" and feels to me like a violation of the Trident API. If there's a better way to do it, I'm open to it, but let's not use strings as identifiers.

arunmahadevan:
@knusbaum @ptgoetz in the proposed name based approach, the naming needs to be done only when one wants to pass specific configs to the node executing the streaming operation, not always. Agree it has magic strings which needs to be matched, but it keeps the configuration and streaming operations separate. The main concern with adding the apis to the Stream is that as we add new config apis to the stream, the stream api gets messy. 

Understand the current limitations, so for now I think we may make it as a part of the stream api, but do evaluate directly adding all the methods to the stream vs adding a single config method taking params.

knusbaum:
@arunmahadevan @ptgoetz Okay, good. We'll leave it in the Stream API then. 

I don't really agree that it's better to have one method and a LoadConfig object. It seems like over-engineering just in case someday we add more things.

It seems like a reasonable compromise, though, so I'll add that stuff in.

revans2:
I am not really sure that we are going to need a Builder interface for setting resources, and for me consistency between the trident and regular storm APIs is more important than the number of methods in Stream.  Right now we only support CPU and Memory as resources.  If we ever add in more resources, or especially generic resources we can add in the Builder API or some other more Generic API to both trident and storm.  For now I think this current API is relatively simple and consistent with the rest of trident/storm.

I am +1 for the current API, but I do have a few comments about the code.

storm-core/src/jvm/org/apache/storm/trident/Stream.java
```diff
@@ -124,6 +124,31 @@ public Stream parallelismHint(int hint) {
     }
 
     /**
+     * Sets the CPU Load resource for the current operation
+     */
+    public Stream setCPULoad(Number load) {
```
revans2: Can we make Stream implement `ResourceDeclarer<Stream>` and have these marked with @Override.

storm-core/src/jvm/org/apache/storm/trident/TridentState.java
```diff
@@ -23,18 +23,33 @@
...
         return new Stream(_topology, _node.name, _node);
     }
-    
+
     public TridentState parallelismHint(int parallelism) {
         _node.parallelismHint = parallelism;
         return this;
     }
+
+    public TridentState setCPULoad(Number load) {
```
revans2: Can we do the same here too?  Have TridentState implement `ResourceDeclarer<TridentState>` and have these marked with `@Override`.

jerrypeng:
As the original author for RAS, I am +1 for the current API.  The most important thing for me is to maintain consistency between the API to set resource load in Trident and Storm

storm-core/src/jvm/org/apache/storm/trident/operation/DefaultResourceDeclarer.java
```diff
@@ -0,0 +1,62 @@
...
+ */
+package org.apache.storm.trident.operation;
+
+import java.util.HashMap;
+import java.util.Map;
+import org.apache.storm.Config;
+import org.apache.storm.utils.Utils;
+import org.apache.storm.topology.ResourceDeclarer;
+
+public class DefaultResourceDeclarer implements ResourceDeclarer, ITridentResource {
```
revans2: Can we make this a generic like

```
public class DefaultResourceDeclarer<T extends DefaultResourceDeclarer> implements ResourceDeclarer<T>, ITridentResource {
```

storm-core/src/jvm/org/apache/storm/trident/planner/Node.java
```diff
@@ -25,7 +26,7 @@
 import org.apache.commons.lang.builder.ToStringStyle;
 
 
-public class Node implements Serializable {
+public class Node extends DefaultResourceDeclarer implements Serializable {
```
revans2: And if DefaultResoureDeclarer is Generic this would change to.

```
public class Node extends DefaultResourceDeclarer<Node> implements Serializable {
```

storm-core/src/jvm/org/apache/storm/trident/TridentTopology.java
```diff
@@ -431,6 +454,52 @@ public StormTopology build() {
         
         return builder.buildTopology();
     }
+
+    private static Map<String, Number> mergeDefaultResources(Map<String, Number> res, Map defaultConfig) {
+        Map<String, Number> ret = new HashMap<String, Number>();
+
+        Number  onHeapDefault = (Number)defaultConfig.get(Config.TOPOLOGY_COMPONENT_RESOURCES_ONHEAP_MEMORY_MB);
```
d2r: minor, extra space after `Number`

storm-core/src/jvm/org/apache/storm/trident/TridentTopology.java
d2r: I agree it would add some complexity.  It might be good to write a brief comment explaining—as you did above—the decision to take the max.

knusbaum:
@arunmahadevan @satishd @ptgoetz, 
Would like your comments on the builder pattern. There are +1's but I believe there were some concerns.

storm-core/src/jvm/org/apache/storm/trident/TridentTopology.java
```diff
@@ -431,6 +454,64 @@ public StormTopology build() {
...
+           thing to do is get the max.
+
+           The situation we want to avoid is that the user sets low resources on one
+           node, and when that node is combined with a bunch of others, the sum is still
+           that low resource count. If any component isn't set, we want to use the default.
+
+           Right now, this code does not check that. It just takes the max of the summed
+           up resource counts for simplicity's sake. We could perform some more complicated
+           logic to be more accurate, but the benefits are very small, and only apply to some
+           very odd corner cases. */g
```
d2r: Comment is OK, but there is a dangling `g` after comment.

storm-core/src/jvm/org/apache/storm/trident/graph/Group.java
```diff
@@ -65,6 +68,25 @@ public Group(Group g1, Group g2) {
     }
 
     @Override
+    public Map<String, Number> getResources() {
+        Map<String, Number> ret = new HashMap<>();
+        for(Node n: nodes) {
+            if(n instanceof ITridentResource) {
+                Map<String, Number> res = ((ITridentResource)n).getResources();
```
satishd: minor: n is always an instance of `ITridentResource`, no need of if check and type casting.

arunmahadevan:
@knusbaum I guess the main concern against builder is that we want to maintain compatibility between core and trident. If so I am fine to change both the apis together and/or explore some other way to pass the config (load, parallelism etc) in trident as a followup. In my opinion, separating the config and the streaming operations would make the trident stream api less cluttered.

knusbaum:
@arunmahadevan Thanks, that sounds good to me.

ptgoetz:
+1

ptgoetz:
@knusbaum Technically this should have waited another ~24 hrs. after the last code change, but I'm not worried because the last change was minor.

Are you planning to pull this into 1.x-branch as well?

knusbaum:
@ptgoetz Bylaws say "1 day from initial patch." I think it makes sense wait longer in general.

I can pull this into 1.x if you want. I thought we were freezing features on that branch, though. I've been a little unclear on what should go in and what shouldn't.

ptgoetz:
@knusbaum Yeah, you're right. :) I always try to wait based on last patch. We should think about changing that.

As far as pulling it into 1.x, I am +1 for it. Although it is a new feature, it builds on top of the core storm RAS feature and I feel it would round out the release. I'd rather 1.x have RAS support for both APIs than just core storm.

knusbaum:
@ptgoetz Makes sense to me. I'll merge it into 1.x.

I'd be +1 for changing the wording in the bylaws, too, if you want to start a [DISCUSSION] or [VOTE], etc.

ptgoetz:
@knusbaum Looks like the clojure tests diverged, I haven't looked at how much. When I think about trident I usually think "not clojure" and forget about the integration tests. :)

I'll start a discussion for the bylaw change.

knusbaum:
@ptgoetz The clojure tests diverged, but it's just import statements. If you want me to create a PR for that, I will, or if you trust me to resolve the imports, I'll just fix them and merge.

ptgoetz:
@knusbaum Yeah, just looked at the conflicts. Not to bad.

Go for it.

PR #1200, opened 2016-03-10
Tests - STORM-1235, STORM-1236, STORM-1237, STORM-1238, STORM-1249

abhishekagarwal87:


storm-core/pom.xml
```diff
@@ -287,13 +287,13 @@
             <artifactId>metrics-clojure</artifactId>
         </dependency>
         <dependency>
-            <groupId>org.mockito</groupId>
```
abhishekagarwal87: hamcrest-core dependency is shaded inside the mockito-all and junit brings in another version of hamcrest-core which are incompatible. Excluding hamcrest-core from junit didn't work. This reordering worked. hamcrest-core dependency is needed by ThrowableNestedCauseMapper class.

abhishekagarwal87:
@revans2 can you take a look? Travis failure looks unrelated.

harshach:
@abhishekagarwal87 Can you file one JIRA for this and list out all the tests that fixed in the description. We shouldn't filing individual jira for each test.

storm-core/pom.xml
harshach: can you add thee above comment in the pom.xml

abhishekagarwal87:
Thanks @harshach. Right now, we have one JIRA for each clojure file that is to be ported into Java. All the JIRA tasks have already been created. Do you want me to file a single JIRA and resolve others as duplicate? The tests, by the way, aren't strictly related to each other.

harshach:
@abhishekagarwal87 ok makes sense confused by the PR title.

abhishekagarwal87:
@revans2  ping

abhishekagarwal87:
@harshach - can you take a look at this PR? It's been pending for a while. Your comments have already been addressed.

abhishekagarwal87:
Can someone review this?

hustfxj:
Thank you @abhishekagarwal87     +1

HeartSaVioR:
+1

PR #1203, opened 2016-03-10
!!! DO NOT MERGE !!! STORM-1617 preview !!! DO NOT MERGE !!!

revans2:
This is intended to be a preview of STORM-1617.  It is by no means done, but I wanted to get some feedback on this.

Adding in all of the javadocs added in about 88MB.  So I am thinking that asf-site itself might need to be in the subversion repo, just so we don't impact all developers with all of the extra files, but I am open to most suggestions.

At this point I have only brought over the javadocs and only put in the 0.10.0 and 0.9.6 releases.

I added in a data file `releases.yaml` that is used to create links to all of the different release sub-directories.  If others think this looks OK at a high level, I will start to go through and add in the rest of the release specific docs.  Put in the rest of the missing releases.  and see if I can put in a permalink, or a sym-link for our current stable release.


kishorvpatil:
I think this is a good idea. A quick looks suggests, we need to move document index and other items under it to general non-release specific folder. (e.g what we see in the first look at the bottom.) 
Documentation index can be at the top link as general documentation.
Also, I see some duplication in http://storm.apache.org/documentation.html#integration and http://storm.apache.org/about/integrates.html

kishorvpatil:
We need to get the links at the bottom fixed.

hustfxj:
I really support  the subversion repo. As far as I am concerned，now the asf-site itself is not convenient for users to browse and learn. And it  is  lack of detailed documentation .

revans2:
Sounds good I will go ahead with doing all of this in subversion, and put up pull requests in git for release specific docs that we would copy to subversion when we do a release.

PR #1206, opened 2016-03-11
STORM-1614: backpressure init and cleanup changes for 1.x branch

abellina:


knusbaum:
+1 LGTM

storm-core/src/clj/org/apache/storm/cluster.clj
```diff
@@ -73,6 +73,7 @@
   (teardown-topology-errors! [this storm-id])
   (heartbeat-storms [this])
   (error-topologies [this])
+  (backpressure-topologies [this])
```
zhuoliu: The name - "backpressure-topologies" is a little confusing (people may think "backpressure" as verb) since it just returns the current topologies' zk nodes (ids) under backpressure root. We might rename it as something like "get-backpressure-topologies-ids".

zhuoliu:
Just minor nit, mostly looks good to me.

storm-core/src/clj/org/apache/storm/cluster.clj
abellina: I agree. I think, however, it is *somewhat* consistent with the interface (e.g. heartbeat-storms, error-topologies don't say get-X-topologies-id). Additionally, I think we should call them one thing throughout (is it storms? is it topologies? is it keys?) Happy to make that change, but I think it's bigger than this PR. Thoughts?

vesense:
LGTM

ptgoetz:
+1

storm-core/src/clj/org/apache/storm/cluster.clj
zhuoliu: Sounds OK.

zhuoliu:
Looks good. +1

ptgoetz:
Merged to 1.x-branch.

PR #1209, opened 2016-03-13
STORM-956: When the execute() or nextTuple() hang on external resources, stop the Worker's heartbeat

srdo:
The previous PR at https://github.com/apache/storm/pull/647 doesn't look active anymore. Having Storm tell you which components are backing up would still be a nice feature to have.

I've taken a look at implementing the suggestions from the previous PR, but I have a few questions.

The previous discussion seemed to point toward shutting down the worker when an executor is hanging. I'm guessing there's no nice way to just restart the hanging executors? Is it sufficient to call shutdown on the worker object from do-executor-heartbeats?

I'm not really sure what Constants/SYSTEM_EXECUTOR_ID is for? Should it be ignored when checking for hanging executors?

I'm hoping to add the zookeeper/metrics logging and shutdown functionality soon if the idea of this PR is sound.


srdo:
I don't want this merged yet, just posted to get feedback :)

hustfxj:
It looks to good.  I also hope we should see this done through both the metrics system and through writing an error into zookeeper that would show up on the UI for the component that is stuck as @revans2  @bastiliu  said. Then let users manually see what is happening.

hustfxj:
Spout itself emits messages by SpoutOutputCollector 's emit().  If lots of messages failed, then acker will trigger SpoutOutputCollector emits those failed messages. It may happen dead lock. Because down bolts may slow to handle messsages and it will block emit(),  then spout/acker thread will block.  Thus others messages which is send by those can't be handled by acker. So the bolts will block. The scene may be called "loop dead lock".  I want say that this PR is sound to this scene. Because It can make us find the dead lock in time.

srdo:
I'll add some tests that errors are going into Zookeeper when executors hang soon/that workers are getting killed only if the option is enabled, and will take a look at adding a function to OutputCollector for manually updating the last hang check timestamp. Could you elaborate on what you'd like this code to do regarding metrics? Is it just a counter of how many times a hanging executor has been found?

hustfxj:
@srdo  I mean that we can report errors into Zookeeper whether the option is enabled. And the metrisc is not only record the counter, but also the timeout time.

srdo:
Will reopen this when the PR is closer to done

srdo:
The hang checks should now support writing errors to Zookeeper, extending the timeout by interacting with an OutputCollector, setting different time limits per component, and disabling the checks entirely by setting the timelimit/check frequency to null. I took a quick look at the metrics system, but can't really see a nice way of logging to it if we're potentially shutting down the worker when this system is triggered.

I'm not sure the automatic/manual hang timeout resets are really necessary on SpoutOutputCollector, since I don't see a case where a user would want to hang in nextTuple while still emitting tuples. Let me know if they should be removed.

I think this PR is ready for re-review.

srdo:
The test failure looks to be the same as on master https://travis-ci.org/apache/storm/jobs/117200843. It ran locally for me with all-tests.

srdo:
Sorry, pressed the close button by accident.

srdo:
@hustfxj Please take another look when you have time, I think this feature is done :)

The test failure seems to occur intermittently in storm-kafka, looks unrelated. It ran for me locally with all-tests.

srdo:
I think this feature is too fiddly for the user, and it doesn't really seem like there's any demand for it. Closing.

PR #1211, opened 2016-03-14
[STORM-1623] fix bug about nimbus.clj

hustfxj:


redsanket:
+1

zhuoliu:
Good catch. +1. 
Checked, it was introduced only to master. 1.x is OK.

jerrypeng:
+1

PR #1213, opened 2016-03-14
Rename README.markdown to README.md.

jumarko:
.md extension is more standard one and commonly used.
There are already CHANGELOG.md, DEVELOPER.md, and SECURITY.md in this repo so it's also better to be consistent.
Moreover some tools don't recognize .markdown extension.


dossett:
Seems like a reasonable change.  Perhaps leave a README.markdown with a pointer to README.md though?

jumarko:
Sure, I can add the symlink README.markdown referencing the README.md.
However, is there some reason for this?
If not, I'd rather remove it completely - the less duplicate files, the better.

dossett:
My thought was something like a "The README is now found <here>" since there are surely some links to README.markdown out there.

PR #1214, opened 2016-03-15
STORM-1629 Files/move doesn't work properly with non-empty directory in Windows

HeartSaVioR:
Please refer https://issues.apache.org/jira/browse/STORM-1629 to see why Files/move with non-empty directory fails on Windows.

* Use FileUtils/moveDirectory on Windows
  * It copies whole contents inside directory, and delete directory
* Keep using Files/move on non-Windows
  * it's still better option since doesn't require copying contents inside directory

Since `on-windows?` is ported to Java, I'll create separate pull request for 1.x-branch.


ptgoetz:
+1

PR #1216, opened 2016-03-15
[STORM-1631] - Storm CGroup bugs 1) when launching workers as the user that submitted the topology 2) when initial cleanup of cgroup fails

jerrypeng:


hustfxj:
It looks good to me, +1

revans2:
+1

caishize:
is any one using cgroup to limit the cpu usage? these days i found the system will be halt when the pressure is growing up, and i found that much of the cpu was wasted by the system, above 25%
CentOS 6.5 
64 bit 
128 G 
32 core
31 workers

vesense:
@caishize too many threads will cause high sys cpu usage. much worse situation, the node will be halt.
 I see there are 31 workers in your machine, what's the total thread number? reducing the number of threads should bring a improvement.

yimapingchuanto:
i want to know if there is a rule about workers num and threads num. In production, i always set the thread num is 10 times than worker num.  oh, i use storm 0.9.5.

vesense:
@yegortokmakov in general, for one topology: 
i:woker number
 worker number < cluster node number
 worker/per node
ii:thread number
 this depends on your logic. for example, the parallelism of KafkaSpout should be equal to the number of partitions, in order to get a good performance.

btw, I really recommend you to upgrade your storm to 1.0.1.

jerrypeng:
At yahoo, cgroups are used to limit both memory and cpu of every worker so
that topologies can have more resource isolation

caishize:
I’ve found that the system will be halt when there are some workers are reboot frequently, this may be a bug of the OS，and i found this https://bugs.centos.org/view.php?id=7538 may be the reason, i hadn’t  confirm this yet..


Darwin

On May 21, 2016, at 11:25 PM, Boyang Jerry Peng <notifications@github.com<mailto:notifications@github.com>> wrote:

At yahoo, cgroups are used to limit both memory and cpu of every worker so
that topologies can have more resource isolation

PR #1220, opened 2016-03-16
STORM-1630 Add guide page for Windows users

HeartSaVioR:
* starting at 1.0.0, Windows users should take care of privilege on account
  * "create symbolic link"

@hustfxj 
I made this page before seeing your comment. 
While technet pages cover how to set up privilege, we could be more kind to providing tl;dr. by pasting your comment on JIRA issue.

```
1) Launch secpol.msc via Start or Start → Run.
2) Open Security Settings → Local Policies → User Rights Assignment.
3) In the list, find the "Create symbolic links" item, which represents SeCreateSymbolicLinkPrivilege.
4) Double-click on the item and add yourself (or the whole Users group) to the list.
```

If you think it would be better to add it, I'll take care of it.

### DIFF
--- a/documentation.html
+++ b/documentation.html
@@ -46,6 +46,7 @@ title: Documentation
                                 <li><a href="documentation/Troubleshooting.html">Troubleshooting</a></li>
                                 <li><a href="documentation/Running-topologies-on-a-production-cluster.html">Running topologies on a production cluster</a></li>
                                 <li><a href="documentation/Maven.html">Building Storm</a> with Maven</li>
+                                <li><a href="documentation/windows-users-guide.html">Windows users guide</li>
                             </ul>
                         </div>
                         <div role="tabpanel" class="tab-pane" id="integration">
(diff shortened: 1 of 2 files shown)


HeartSaVioR:
Btw, I don't see the reason why we keep `_site` directory. 
Since it can be built anytime, and we don't serve website from that directory, I guess it could be removed.

hustfxj:
@HeartSaVioR  sorry I think we shouldn't add it now. It is superfluous, so that's more better to a official guide from MS, linking that page, as you do.

HeartSaVioR:
@hustfxj Good. Thanks for the comment. :)

ptgoetz:
+1

HeartSaVioR:
After STORM-1617, this PR should be rebased to master and 1.x-branch respectively.
I'll create new two pull requests, and ping again.
Thanks for reviewing!

HeartSaVioR:
@hustfxj @ptgoetz 
Please review #1247 and #1248 to continue. Thanks!

PR #1221, opened 2016-03-16
Fix logging for LoggingMetricsConsumer STORM-584

jnioche:
Currently the metrics.log file is always empty and it declared in the cluster.xml file whereas it should be in the worker one. The metrics end up in the log file which is not a huge issue but it would be cleaner to have them in a separate file, as originally intended. 

This patch addresses this and makes so that there is one metrics file per worker, as per the other log files, which should address the issues mentioned in [https://issues.apache.org/jira/browse/STORM-584].


harshach:
+1

unsleepy22:
+1

abhishekagarwal87:
LGTM

hustfxj:
+1

ptgoetz:
+1

revans2:
+1

PR #1223, opened 2016-03-16
STORM-1537: Upgrade to kryo 3

abhishekagarwal87:


abhishekagarwal87:
This is not to be merged yet. Right now, I only wanted to showcase all the changes that are to be made for the upgrade. Kryo 3 apparently has some performance improvements. Since the kryo serialization is used only for intra topology transfer, there should be no compatibility issues. Let me know if this change has other side effects.

ptgoetz:
+1

Tested on a real cluster and compared to 1.x-branch before and after this patch. Throughput increased and latency decreased.

abhishekagarwal87:
@ptgoetz  This should be merged into master as well.

PR #1226, opened 2016-03-17
[STORM-1636] - Supervisor shutdown with worker id pass in being nil

jerrypeng:
In function kill-existing-workers-with-change-in-components in supervisor.clj:
The function tries to detect whether there is a change in assignment. The bug in this function is that the ordering of the assignment matters but it shouldn't. For example, if a worker assignment is [[1 1] [2 2]] and it changed to [[2 2] [1 1]] it will cause the supervisor to restart the worker


jerrypeng:
Also needs to be pushed back into 1.x branch

hustfxj:
nice catch. I have considered this when I port supervisor.clj to java. Than you @jerrypeng

redsanket:
+1

revans2:
+1

PR #1227, opened 2016-03-17
STORM-1602 Blobstore UTs are failed on Windows

HeartSaVioR:
* ensures objects of InputStream / OutputStream are closed after used
  * clojure: with-open
  * java: try-with-resource
* skip checking symbolic link in LocalizerTest when on Windows
  * Windows seems not handle symbolic link in compressed file properly

Some integration tests are still failing on Windows, but it should be handled with another issues.

I'll also address this based on master immediately.

Please have a look. Thanks!

### DIFF
--- a/storm-core/src/clj/org/apache/storm/daemon/supervisor.clj
+++ b/storm-core/src/clj/org/apache/storm/daemon/supervisor.clj
@@ -1168,8 +1168,10 @@
         blob-store (Utils/getNimbusBlobStore conf master-code-dir nil)]
     (try
       (FileUtils/forceMkdir (File. tmproot))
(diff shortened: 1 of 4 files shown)


HeartSaVioR:
#1229 is based on master branch.

revans2:
I assume they all pass on windows so +1

The test failure looks like it is a bug with the mongodb pom.xml that JDK8 does not like but JDK7 seems OK with.  Very odd.  Filed a JIRA for it though.

redsanket:
+1

HeartSaVioR:
@revans2 Very odd indeed... It should give us error because of storm-mongo module.

PR #1229, opened 2016-03-17
STORM-1602 Blobstore UTs are failed on Windows

HeartSaVioR:
* ensures objects of InputStream / OutputStream are closed after using
  * clojure: with-open
  * java: try-with-resource
* skip checking symbolic link in LocalizerTest when on Windows
  * Windows seems not handle symbolic link in compressed file properly


HeartSaVioR:
#1227 is based on 1.x-branch.

redsanket:
+1

PR #1230, opened 2016-03-17
[STORM-1573] Add batch support for MongoInsertBolt

vesense:
https://issues.apache.org/jira/browse/STORM-1573

### DIFF
--- a/external/storm-mongodb/src/main/java/org/apache/storm/mongodb/bolt/MongoInsertBolt.java
+++ b/external/storm-mongodb/src/main/java/org/apache/storm/mongodb/bolt/MongoInsertBolt.java
@@ -17,11 +17,18 @@
  */
 package org.apache.storm.mongodb.bolt;
 
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+
 import org.apache.commons.lang.Validate;
 import org.apache.storm.mongodb.common.mapper.MongoMapper;
 import org.apache.storm.topology.OutputFieldsDeclarer;
 import org.apache.storm.tuple.Tuple;
+import org.apache.storm.utils.TupleUtils;
 import org.bson.Document;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Basic bolt for writing to MongoDB.
@@ -30,8 +37,19 @@ import org.bson.Document;
  *
  */
 public class MongoInsertBolt extends AbstractMongoBolt {
+    private static final Logger LOG = LoggerFactory.getLogger(MongoInsertBolt.class);
+
+    private static final int DEFAULT_FLUSH_INTERVAL_SECS = 1;
 
     private MongoMapper mapper;
+
+    private boolean ordered = true;  //default is ordered.
+
+    private int batchSize = 15000;
+
+    private List<Tuple> tupleBatch;
+
+    private int flushIntervalSecs = DEFAULT_FLUSH_INTERVAL_SECS;
 
     public MongoInsertBolt(String url, String collectionName, MongoMapper mapper) {
         super(url, collectionName);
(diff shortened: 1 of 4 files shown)


vesense:
This PR can be merged to both master and 1.x branch.

vesense:
Could you take a look at this please, @harshach ?

vesense:
ping @harshach

HeartSaVioR:
@vesense 
Could we apply standard batching behavior among external modules?
I mean, this PR doesn't flush when batch size is set to 100 but there're 99 tuples in several hours, while storm-hbase, storm-hive, and more flushes batch when tick tuple is received, and it sets tick tuple interval.

vesense:
@HeartSaVioR Thanks for your reply. I planed to handle tick tuples in a follow up JIRA. But I'm OK for adding these changes in one JIRA, in order to provide standard batching behavior.

Furthermore, I think it's a great job if we can abstract batch processing to an common api among external modules.

HeartSaVioR:
@vesense 
I'd really like to add that change in this issue, since it's not hard to address.
Btw, totally agreed about abstracting batch process to something common.

vesense:
@HeartSaVioR OK. I will add the change later.

vesense:
@HeartSaVioR Updated. also 1.x-branch: https://github.com/apache/storm/pull/1266

vesense:
@HeartSaVioR Could you take a look?

external/storm-mongodb/src/main/java/org/apache/storm/mongodb/bolt/MongoInsertBolt.java
```diff
@@ -99,6 +107,29 @@ public MongoInsertBolt withOrdered(boolean ordered) {
...
+    }
+
+    @Override
+    public Map<String, Object> getComponentConfiguration() {
+        Map<String, Object> conf = super.getComponentConfiguration();
+        if (conf == null) {
+            conf = new Config();
+        }
+
+        if (conf.containsKey(Config.TOPOLOGY_MESSAGE_TIMEOUT_SECS) && flushIntervalSecs == 0) {
```
HeartSaVioR: Turned out it simply doesn't work, since topology config is not propagated to here.
Please refer #893 for more details.
We just have to have default value of flushIntervalSecs, which value should make sense with MongoDB.
(For me I set 1 sec as a default value of HBaseBolt. Please refer #1251.)

vesense: Thanks @HeartSaVioR Fixed the issue.

vesense: also updated 1.x-branch

vesense: @HeartSaVioR I find that `storm-solr` has the same bug as you stated. And I filed a JIRA(https://issues.apache.org/jira/browse/STORM-1669) for that. Maybe this fix should be included in Storm 1.0 release.

HeartSaVioR:
+1

HeartSaVioR:
Could you rebase these three commits? After rebasing I'll merge.

vesense:
@HeartSaVioR Done. rebased both master and 1.x-branch.

HeartSaVioR:
+1 Thanks for the patch.

vesense:
>Btw, totally agreed about abstracting batch process to something common.
@HeartSaVioR I have created a PR for abstracting batch process. see https://github.com/apache/storm/pull/1303

vesense:
@HeartSaVioR Could you take a look at https://github.com/apache/storm/pull/1303 ?

PR #1233, opened 2016-03-18
Optimized wipeZeros() in SlotBasedCounter.

hukun01:
objToCounts.keySet() already returns a set view, so we can avoid creating an unnecessary new set `objToBeRemoved`.


hukun01:
It was failing before I did git clone.

PR #1238, opened 2016-03-19
STORM-1229: port backtype.storm.metric.testing to java

abhishekagarwal87:


storm-core/test/jvm/org/apache/storm/metric/FakeMetricConsumer.java
```diff
@@ -0,0 +1,88 @@
...
+import org.apache.storm.task.IErrorReporter;
+import org.apache.storm.task.TopologyContext;
+
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
+
+public class FakeMetricConsumer implements IMetricsConsumer {
+
+    public static final Table<String, String, Multimap<Integer, Object>> buffer = HashBasedTable.create();
```
unsleepy22: why is this static? can't we make it an instance field?

hustfxj:
@abhishekagarwal87  thank you. It have better use AtomicReference<T> to replace the Map, then we can avoid to use  the "synchronized". And  do HashBasedTable.create() generate the struct "HashMap<R, HashMap<C, V>>", but not "Multimap<R, HashMap<C, V>>" ?

storm-core/test/jvm/org/apache/storm/metric/FakeMetricConsumer.java
abhishekagarwal87: It is used in testing. The tests don't have access to the instance which are prepared inside the storm. That is why it is declared `public static` so that tests can access the metrics.

abhishekagarwal87:
@hustfxj the implementation of handleDataPoints may not be straightforward with AtomicReferences. I found emulating clojure atom operators verbatim tricky so I did it in-place in java fashion. Result code was not that verbose so I just left it there.

abhishekagarwal87:
@revans2 / @knusbaum  - can you review this?

abhishekagarwal87:
This PR has been open for quite some time now. Hoping someone can review and merge this.

HeartSaVioR:
Since it's just a test code, it doesn't need to be optimized unless it is needed.
+1 on me.

PR #1241, opened 2016-03-21
STORM-1641: make subtree node creation consistent

abellina:
This doesn't appear to make a difference. When starting nimbus, I get 9 znodes inside of /storm either way. However, it was inconsistent.


redsanket:
+1

revans2:
+1 looks like this was translated differently for 2.x, so it is not needed on master.

abellina:
@revans2 that is correct.

PR #1242, opened 2016-03-21
Storm 1643 - Performance Fix: Optimize clojure lookups related to throttling and stats

roshannaik:
For a simple Speed of light topology with 1 spout (generating random nos)  and 1 bolt (which just multiplies the number by 2),  Ackers=0 and  Event Loggers=0 ...  this change showed approx 15%  improvement in emits/sec.

### DIFF
--- a/storm-core/src/clj/org/apache/storm/daemon/executor.clj
+++ b/storm-core/src/clj/org/apache/storm/daemon/executor.clj
@@ -365,6 +365,7 @@
 
 (defn mk-executor [worker executor-id initial-credentials]
   (let [executor-data (mk-executor-data worker executor-id)
+        storm-conf (:storm-conf executor-data)
         _ (log-message "Loading executor " (:component-id executor-data) ":" (pr-str executor-id))
         task-datas (->> executor-data
                         :task-ids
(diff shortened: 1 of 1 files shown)


storm-core/src/clj/org/apache/storm/daemon/executor.clj
```diff
@@ -491,7 +493,7 @@
           EVENTLOGGER-STREAM-ID
           [component-id message-id (System/currentTimeMillis) values]))))
 
-(defmethod mk-threads :spout [executor-data task-datas initial-credentials]
+(defmethod mk-threads :spout [executor-data task-datas initial-credentials throttle-on]
```
zhuoliu: I think this is logically wrong. Throttle-on flag is a boolean that you have to obtain dynamically from the executor-data in run time, not that you can pass in as a constant when making the spout thread at the begining.

zhuoliu:
I would suggest testing a special topology with backpressure enabled. E.g., [this one.](https://github.com/zhuoliu/storm/blob/1643/examples/storm-starter/src/jvm/org/apache/storm/starter/WordCountTopology2.java), in which you can intermittently see a throttle-on flag (like "6703") under /backpressure/wc2-topo/ directory of Zookeeper.

storm-core/src/clj/org/apache/storm/daemon/executor.clj
HeartSaVioR: Yes, this change will effectively disable backpressure. 

@roshannaik 
We still need to look up that value in spout async loop. 
One sketching idea for workaround is that we may not need to look up that value for each iteration, which is based on precondition that it may also not hurt when spout notices about throttle slightly later.

Btw, I guess when nextTuple() takes several time (say, about to 1 ms, not speed of light but still not slow), looking up that value may be not hurt overall performance. When we don't have data for emit, we sleep 1ms in loop for default configuration.

```
topology.spout.wait.strategy: "org.apache.storm.spout.SleepSpoutWaitStrategy"
topology.sleep.spout.wait.strategy.time.ms: 1
```

knusbaum:
This hasn't been updated for ~7 months, and I'm -1 on the concept anyhow, so I'm going to close it.

If you'd like to discuss, feel free to reopen.

PR #1244, opened 2016-03-22
Make StormTimer join task thread on close

srdo:
Most tests creating a cluster can be flaky because the supervisor threads can terminate out of order. The StormTimer's close should block until the thread is dead.


hustfxj:
good, +1

knusbaum:
+1

PR #1131, opened 2016-02-21
STORM-822: Kafka Spout New Consumer API

hmcl:
This patch is still under development and was uploaded at this moment for early testing. Please read README.

There may be a bug in the offsets management, because of diff o 1. I am looking into it. 
Currently polling from an arbitrary offset is possible but it will come in the next patch, today or tomorrow
I refactored the code a bit and left, maybe, some unnecessary locking. I am also looking into it.

@connieyang @jianbzhou @tgravescs please let me know of any other requirements you may have and I will address them soon.

Thanks

PR #1245, opened 2016-03-22
DO NOT MERGE [STORM-1646]  Fix intermittent failures in storm-kafka unit tests

ppoulosk:
This PR is dependent on the pr 1131, for STORM-822.  

https://github.com/apache/storm/pull/1131

We have run into a few issues running storm-kafka unit tests on lightly resourced VMs, especially after moving to kafka-9.

This moves the exponential backoff manager to use simulated time in testing.  We had been seeing races when using Thread.sleep with low sleep periods.

It also switches a config no longer used in kafka 9 to the recommended config parameter, and bumps up the timeout.

49394 [main] WARN o.a.k.c.p.KafkaProducer - metadata.fetch.timeout.ms config is deprecated and will be removed soon. Please use max.block.ms

### DIFF
--- a/examples/storm-starter/pom.xml
+++ b/examples/storm-starter/pom.xml
@@ -151,26 +151,12 @@
     </dependency>
     <dependency>
       <groupId>org.apache.kafka</groupId>
-      <artifactId>kafka_2.10</artifactId>
-      <version>0.8.2.1</version>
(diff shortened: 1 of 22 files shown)


ppoulosk:
Closing this PR and putting up another one since STORM-822 is resolved.

PR #1246, opened 2016-03-23
STORM-1648: drpc spout reconnect on failure

abellina:


hustfxj:
It may not be necessary， but I still +1.

revans2:
+1 we have seen issues around rolling upgrades after we kill a DRPC server and bring it back up, subsequent DRPC request drop a few messages before everything starts to work again.  It is minor but a pain for some users.

PR #1247, opened 2016-03-23
STORM-1630 Add guide page for Windows users

HeartSaVioR:
It supersedes #1220 since docs are back to versioned branch.

* Add document to guide how to setup account to be able to 'create symbolic link' on Windows


HeartSaVioR:
It's just forward port (actually backport, but vote is there) and I got +1 from #1220, too.
I'll merge this.

PR #1249, opened 2016-03-23
STORM-1271: Port backtype.storm.daemon.task to java

abhishekagarwal87:


storm-core/src/jvm/org/apache/storm/daemon/Task.java
```diff
@@ -0,0 +1,247 @@
...
+    private WorkerTopologyContext workerTopologyContext;
+    private LoadMapping loadMapping;
+    private Integer taskId;
+    private String componentId;
+    private Object taskObject;
+    private Map stormConf;
+    private Callable<Boolean> emitSampler;
+    private CommonStats executorStats;
+    private Map<String, Map<String, Grouper>> streamComponentToGrouper;
+    private Object builtInMetrics;      //TODO
```
abhishekagarwal87: TODO is for the change after PR for porting builtin-metrics is merged.

knusbaum: I don't think we want any TODO comments in the code. I'm guessing builtin-metrics should be merged before this?

revans2: It already is.  This needs to be upmerged now.

storm-core/src/jvm/org/apache/storm/daemon/Task.java
```diff
@@ -0,0 +1,247 @@
...
+    private boolean debug;
+
+    public Task(Map executorData, Integer taskId) throws IOException {
+        this.taskId = taskId;
+        this.executorData = executorData;
+        this.workerData = (Map) executorData.get("worker");
```
revans2: I assume ExecutorData is going to become a java object at some point in the future, and when that happens these will change.

hustfxj: @abhishekagarwal87  You can make ExecutorData and WorkerData  become a java objects now, In fact I have already create them when I port the executor.clj to java.

abhishekagarwal87: @revans2 - That is correct. Constructor of Task object remains the same this way (only input types change). 
@hustfxj - I guess task should go before executor is ported.

abhishekagarwal87:
up-merged with builtin-metrics changes.

storm-core/src/jvm/org/apache/storm/daemon/Task.java
abhishekagarwal87: @knusbaum - Yes. I have removed it.

storm-core/src/jvm/org/apache/storm/daemon/Task.java
```diff
@@ -0,0 +1,246 @@
...
+
+    private Map executorData;
+    private Map workerData;
+    private TopologyContext systemTopologyContext;
+    private TopologyContext userTopologyContext;
+    private WorkerTopologyContext workerTopologyContext;
+    private LoadMapping loadMapping;
+    private Integer taskId;
+    private String componentId;
+    private Object taskObject;
```
revans2: It might be nice to have some javadocs explaining that this is the actual bolt/spout instance that is being executed.

storm-core/src/clj/org/apache/storm/daemon/executor.clj
```diff
@@ -15,6 +15,7 @@
 ;; limitations under the License.
 (ns org.apache.storm.daemon.executor
   (:use [org.apache.storm.daemon common])
+  (:use [clojure.walk])
```
revans2: I don't see where this import is used.  Is it needed?

abhishekagarwal87: https://github.com/apache/storm/pull/1249/files#diff-6e352b8cfa5485b01db59718464c2dd1R324 `stringify-keys`

storm-core/src/jvm/org/apache/storm/daemon/Task.java
abhishekagarwal87: I agree, the name is a bit confusing. I was thinking of renaming it to componentObject but that may again be confusing. Any suggestions? I will add the javadoc nonetheless

storm-core/src/jvm/org/apache/storm/daemon/Grouper.java
```diff
@@ -0,0 +1,111 @@
...
+import org.apache.storm.grouping.ShuffleGrouping;
+import org.apache.storm.task.WorkerTopologyContext;
+import org.apache.storm.tuple.Fields;
+import org.apache.storm.utils.TupleUtils;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+public abstract class Grouper {
```
revans2: Why is this not a `LoadAwareCustomStreamGrouping`?  Or perhaps better a different interface that LoadAwareCustomStreamGrouping can extend?

That way we would not need as much glue code to make custom groupings work.  It also would make it more consistent in the naming of methods with `chooseTasks` instead of `outTasks`.  This is minor, and a bit for my curiosity if you had a reason because adding in more class hierarchy has its drawbacks as well.

storm-core/src/jvm/org/apache/storm/daemon/Grouper.java
```diff
@@ -0,0 +1,111 @@
...
+
+        @Override
+        public List<Integer> outTasks(Integer taskId, List<Object> values, LoadMapping load) {
+            return customStreamGrouping.chooseTasks(taskId, values);
+        }
+    }
+
+    public static class LoadAwareCustomGrouper extends CustomGrouper {
+
+        public LoadAwareCustomGrouper(LoadAwareShuffleGrouping customStreamGrouping, WorkerTopologyContext context, String componentId,
```
revans2: `LoadAwareShuffleGrouping` should be `LoadAwareCustomStreamGrouping` instead.

storm-core/src/jvm/org/apache/storm/daemon/GrouperFactory.java
```diff
@@ -0,0 +1,110 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Random;
+import java.util.Set;
+
+public class GrouperFactory {
+
+    // A no-op grouper
+    public static Grouper DIRECT = new Grouper() {
```
revans2: final

storm-core/src/jvm/org/apache/storm/daemon/GrouperFactory.java
```diff
@@ -0,0 +1,110 @@
...
+                };
+            case CUSTOM_OBJECT:
+                CustomStreamGrouping customObject =
+                    (CustomStreamGrouping) Thrift.instantiateJavaObject(thriftGrouping.get_custom_object());
+                return new Grouper.CustomGrouper(customObject, context, componentId, streamId, targetTasks);
+            case CUSTOM_SERIALIZED:
+                CustomStreamGrouping
+                    customSerialized =
+                    Utils.javaDeserialize(thriftGrouping.get_custom_serialized(), CustomStreamGrouping.class);
+                return new Grouper.CustomGrouper(customSerialized, context, componentId, streamId, targetTasks);
```
revans2: If someone created a CustormStreamGrouping that was also LoadAwareCustomStreamGrouping the LoadAware portion would never be called, because we are now using constructors instead of the builder pattern.  This really should be.

```
case CUSTOM_SERIALIZED:
  CustomStreamGrouping
    customGrouping =
      Utils.javaDeserialize(thriftGrouping.get_custom_serialized(), CustomStreamGrouping.class);
    if (customGrouping instanceof CustomLoadAwareStreamGrouping) {
      return new Grouper.LoadAwareCustomGrouper(customSerialized, context, componentId, streamId, targetTasks);
    }
    return new Grouper.CustomGrouper(customSerialized, context, componentId, streamId, targetTasks);
```

storm-core/src/jvm/org/apache/storm/daemon/GrouperFactory.java
```diff
@@ -0,0 +1,110 @@
...
+                    @Override
+                    public List<Integer> outTasks(Integer taskId, List<Object> values, LoadMapping load) {
+                        int index = random.nextInt(numTasks);
+                        return Collections.singletonList(targetTasks.get(index));
+                    }
+                };
+            case CUSTOM_OBJECT:
+                CustomStreamGrouping customObject =
+                    (CustomStreamGrouping) Thrift.instantiateJavaObject(thriftGrouping.get_custom_object());
+                return new Grouper.CustomGrouper(customObject, context, componentId, streamId, targetTasks);
```
revans2: If someone created a CustormStreamGrouping that was also LoadAwareCustomStreamGrouping the LoadAware portion would never be called, because we are now using constructors instead of the builder pattern.  This really should be.

```
case CUSTOM_OBJECT:
  CustomStreamGrouping
    customGrouping =
      (CustomStreamGrouping) Thrift.instantiateJavaObject(thriftGrouping.get_custom_object());
    if (customGrouping instanceof CustomLoadAwareStreamGrouping) {
      return new Grouper.LoadAwareCustomGrouper(customSerialized, context, componentId, streamId, targetTasks);
    }
    return new Grouper.CustomGrouper(customSerialized, context, componentId, streamId, targetTasks);
```

storm-core/src/jvm/org/apache/storm/daemon/Task.java
```diff
@@ -0,0 +1,246 @@
...
+
+    public Object getTaskObject() {
+        return taskObject;
+    }
+
+    public BuiltinMetrics getBuiltInMetrics() {
+        return builtInMetrics;
+    }
+
+    private TopologyContext mkTopologyContextBuilder(StormTopology topology) throws IOException {
```
revans2: It looks like `mk-topology-context-builder`, `system-topology-context`, and `user-topology-context` all got merged into this.  It probably would be best to just call it `mkTopologyContext`, and eventually have a Constructor on `TopologyContext` that just takes a StormTopology.

storm-core/src/jvm/org/apache/storm/daemon/Task.java
```diff
@@ -0,0 +1,246 @@
...
+            ConfigUtils.supervisorStormResourcesPath(ConfigUtils.supervisorStormDistRoot(conf, (String) workerData.get("storm-id"))),
+            ConfigUtils.workerPidsRoot(conf, (String) workerData.get("worker-id")),
+            taskId,
+            (Integer) workerData.get("port"),
+            (List<Integer>) workerData.get("task-ids"),
+            (Map<String, Object>) workerData.get("default-shared-resources"),
+            (Map<String, Object>) workerData.get("user-shared-resources"),
+            (Map<String, Object>) executorData.get("shared-executor-data"),
+            (Map<Integer, Map<Integer, Map<String, IMetric>>>) executorData.get("interval->task->metric-registry"),
+            (clojure.lang.Atom) executorData.get("open-or-prepare-was-called?")
```
revans2: I assume this will go to AtomicBoolean when Executor goes to java.

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
```diff
@@ -135,7 +137,28 @@ public static int samplingRate(Map conf) {
         throw new IllegalArgumentException("Illegal topology.stats.sample.rate in conf: " + rate);
     }
 
-    // public static mkStatsSampler // depends on Utils.evenSampler() TODO, this is sth we need to do after util
+    public static Callable<Boolean> evenSampler(final int samplingFreq) {
+        final int start = 0;
+        final Random random = new Random();
+        final MutableInt curr = new MutableInt(-1);
+        final MutableInt target = new MutableInt(random.nextInt(samplingFreq));
```
revans2: The entire reason for the MutableInt was so clojure could ack more like java.  It is going to be a lot faster to just use int instead.

```
    public static Callable<Boolean> evenSampler(final int samplingFreq) {
        final Random random = new Random();
        return new Callable<Boolean>() {
            private int curr = -1;
            private int target = random.nextInt(samplingFreq);

            @Override
            public Boolean call() throws Exception {
                curr++;
                if (curr >= samplingFreq) {
                    curr = 0;
                    target = random.nextInt(samplingFreq);
                }
                return curr == target;
            }
        };
```

storm-core/src/clj/org/apache/storm/daemon/executor.clj
revans2: Ah OK.

revans2:
Overall it looks really good.  Only one real bug that I could see, but then some nits and design choice questions, mostly around if what we did for clojure still makes since in java, all of the design stuff I see as very minor and we can address it later.

storm-core/src/jvm/org/apache/storm/daemon/Grouper.java
abhishekagarwal87: Honestly I failed to notice that LoadAwareCustomStreamGrouping has the same interface needed by Task class. It will fit in perfectly. The only caveat is the naming as many implementations of `LoadAwareCustomStreamGrouping` wouldn't exactly be load aware.

storm-core/src/jvm/org/apache/storm/daemon/GrouperFactory.java
abhishekagarwal87: will do

storm-core/src/jvm/org/apache/storm/daemon/GrouperFactory.java
abhishekagarwal87: Good catch. This problem will go away once I refactor the classes.

storm-core/src/jvm/org/apache/storm/daemon/Task.java
abhishekagarwal87: This makes sense. I will change it

storm-core/src/jvm/org/apache/storm/daemon/Task.java
abhishekagarwal87: That's right.

storm-core/src/jvm/org/apache/storm/utils/ConfigUtils.java
abhishekagarwal87: will do. Out of Curiosity, are the calls to emitSampler object done in single thread?

storm-core/src/jvm/org/apache/storm/daemon/Task.java
hustfxj: componentObject  is ok for me

abhishekagarwal87:
@revans2 - Addressed your comments. Please review.

revans2:
+1 looks good to me

revans2:
The changes look good but STORM-1663 and STORM-1665 are currently making master unusable so I don't want to merge new things in until they are fixed.

PR #1251, opened 2016-03-24
STORM-1654 HBaseBolt creates tick tuples with no interval when we don't set flushIntervalSecs

HeartSaVioR:
Please also refer #893 to see why this change is necessary.

Set 'default' flush interval seconds to HBaseBolt, since taking half of message timeout secs doesn't work, as #893 showed.
Since we should hardcode default flush interval seconds for now, I think we can change default flush interval to 1s since I guess it's enough for HBase, and anyone never set up tuple timeout seconds to 1. (When if this is happening, all bolts with batching manner using tick tuple should also fail.) 

I'll craft separate PR to be applied to 1.x-branch.

### DIFF
--- a/external/storm-hbase/src/main/java/org/apache/storm/hbase/bolt/AbstractHBaseBolt.java
+++ b/external/storm-hbase/src/main/java/org/apache/storm/hbase/bolt/AbstractHBaseBolt.java
@@ -43,7 +43,6 @@ public abstract class AbstractHBaseBolt extends BaseRichBolt {
     protected HBaseMapper mapper;
     protected String configKey;
     protected int batchSize = 15000;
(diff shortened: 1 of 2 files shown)


harshach:
+1

priyank5485:
+1 (NB). @HeartSaVioR I had created https://issues.apache.org/jira/browse/STORM-1626 
Will close that one since its a duplicate of STORM-1654. Please let me know if its not true.

HeartSaVioR:
@priyank5485 Right. STORM-1626 is symptom of STORM-1654. Sorry I didn't recognize STORM-1654 exists.

harshach:
Thanks @HeartSaVioR pushed into 1.x-branch and master.

PR #1254, opened 2016-03-24
STORM-1655 (1.x) Flux doesn't set return code to non-zero when there's any exception while deploying topology to remote cluster

HeartSaVioR:
* Modify Flux.runCli() to not catching exceptions so that exception can be propagated to main()


unsleepy22:
+1

knusbaum:
+1

hustfxj:
+1

ptgoetz:
+1

PR #1260, opened 2016-03-25
Fix wait-for-desired-code-replication always choosing distributed mod…

srdo:
…e regardless of actual cluster mode

Since total-storm-conf is the merge of daemon conf and topology conf, it more than likely doesn't reflect which STORM-CLUSTER-MODE Nimbus was started with.


srdo:
Nevermind this, the right mode is set when using the local cluster from testing4j, so this is probably fine.

PR #1274, opened 2016-03-29
Shade Objenesis in storm-core

srdo:
Versions of Kryo past 2.23 have Objenesis as a regular dependency instead of included in the Kryo jar. https://github.com/EsotericSoftware/kryo/commit/f21208643e883fde952ad883fd81e5d7709e87eb Since Kryo was recently upgraded, Objenesis is missing from storm-core. Objenesis is included through Carbonite, so we can still exclude it from Kryo to avoid getting it in /lib when making a distribution.

### DIFF
--- a/storm-core/pom.xml
+++ b/storm-core/pom.xml
@@ -503,6 +503,7 @@
                             <include>org.apache.curator:*</include>
                             <include>com.twitter:carbonite</include>
                             <include>com.twitter:chill-java</include>
+                            <include>org.objenesis:objenesis</include>
                             <include>org.tukaani:xz</include>
                             <include>org.yaml:snakeyaml</include>
                             <include>org.jgrapht:jgrapht-core</include>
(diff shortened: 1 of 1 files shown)


srdo:
Test failure seems unrelated.

srdo:
This doesn't really work. Shading Objenesis means that code with a dependency on storm-core won't get a functional Kryo, since the transitive dependency on Kryo refers to the original Objenesis package.

As far as I can tell, if we remove the objenesis exclusion from the Kryo dependency and remove the shading, this should work like it did pre-upgrade. An Objenesis jar will go in /lib, but since Kryo previously contained Objenesis unshaded in the Kryo jar, it should work out to be the same as before.

Alternatively, we could also shade Kryo in order to get it off the dependency reduced POM, so projects depending on storm-core won't pull Kryo as a dependency of storm-core.

hustfxj:
Yes, you are right. Objenesis is missing from storm-core. I suggest we should Integrate the Objenesis's code in order to avoid to rely on the Objenesis jar.

srdo:
@hustfxj Do you mean we should shade it? I don't understand why copying the Objenesis code into Storm would improve things. If we put it in the original package it's equivalent to just depending on the Objenesis jar. If we move it, Kryo can't find it unless we use shading. If we shade it we also need to shade Kryo in order for storm-core to work as a dependency during tests (otherwise we declare a dependency on Kryo which depends on an unshaded Objenesis).

srdo:
Seems like Objenesis is being reincluded here https://github.com/apache/storm/pull/1279. Closing this.

PR #1277, opened 2016-03-30
STORM-1129: Update ui to use topology name

priyank5485:


abhishekagarwal87:
@priyank5485 I see that you have the changed the logviewer urls in the topology page to use topology name but there are no changes in logviewer.clj. I think logviewer links should just point to topology id since log links are anyway tied to a specific instance of topology run.

storm-core/src/ui/public/component.html
```diff
@@ -176,7 +176,7 @@
           var loc = $(row[3])[0]; // logviewer URL
           return '<input type="checkbox" class="workerActionCheckbox"'+
               'id="'+checkboxId+'" value="'+host_port+'"'+checkedString+'/> '+
-              '<a href="'+loc.protocol+'//'+loc.host+'/dumps/'+topologyId+'/'+
+              '<a href="'+loc.protocol+'//'+loc.host+'/dumps/'+topologyName+'/'+
```
abhishekagarwal87: should be left as it is.

priyank5485:
@abhishekagarwal87 Thanks for the catch. I have updated the PR. Can you please review?

storm-core/src/clj/org/apache/storm/ui/core.clj
```diff
@@ -466,6 +466,21 @@
...
+        matching-topologies-by-name (filter filter-fun-name summary)
+        matching-topologies-by-id (filter filter-fun-id summary)
+        matching-topologies
+        (cond
+          (not-empty matching-topologies-by-name) matching-topologies-by-name
+          (not-empty matching-topologies-by-id) matching-topologies-by-id
+          :else nil)
+        _ (when (empty? matching-topologies) (throw (NotAliveException. (str topology-name-or-id " is not alive"))))
+        id ((first matching-topologies) "id")]
+    id))
```
knusbaum: We should probably prioritize matching ids here instead of matching names. As it is now, If I give my topology a name that matches another topology's id, the topology whose id I stole will become inaccessible. Since we're prioritizing names, it will match that topology's id to my topology's name, and display my topology instead.

priyank5485: @knusbaum Thats true. But if we prioritize ids would not your topology become inaccessible? I think going forward we should use topology names in ui and elsewhere so that we can have permanent links and since thats the identifier given by the user. Support for lookup with ids was added for backwards compatibility as suggested by @revans2 and others.

knusbaum: That's fine with me.

abhishekagarwal87:
Thanks @priyank5485 - Though I am still not sure how logviewer links will work. topologyId variable is not assigned any value. Original code -
https://github.com/apache/storm/blob/master/storm-core/src/ui/public/component.html#L136

priyank5485:
@abhishekagarwal87 Please look at the $.getJson method. Line 190 in component.html. Now, instead of topologyId being assigned on onReady i do it in $.getJson. Its a side effect of squashing commits that makes it hard to look at the commits that addressed review comments. Sorry about that. Do you guys usually keep separate commits and squash in the end? I will do that next time.

abhishekagarwal87:
I think either way is fine. My only concern was logviewer links which I think you have addressed.

harshach:
@priyank5485 sorry for the delay on this. Can you up merge this patch.

PR #1278, opened 2016-03-30
remove unnecessary registerSerialization about TransactionAttempt because we will register the class in the SerializationFactory

hustfxj:
very minor.


satishd:
+1

hustfxj:
we register the class "org.apache.storm.transactional.TransactionAttempt" in SerializationFactory , But we register the class "org.apache.storm.trident.topology.TransactionAttempt" in MasterBatchCoordinator. So it will occur serialization errors when we run trident topology. Do you Think ? @revans2 @satishd

satishd:
@hustfxj That was a good catch. We may need to keep `org.apache.storm.trident.topology.TransactionAttempt` registered  as it is serialized/deserialized over the wire in trident topologies.

PR #1257, opened 2016-03-25
[STORM-1279] port backtype.storm.daemon.supervisor to java

hustfxj:
1 port supervisor.clj to java;
2 Update all the callings to supervisor;
3 Supervisor's class hierarchy as follows:
SupervisorManger is supervisor' manger which can clean and shutdown supervisor;
SyncSupervisorEvent is responsible for downloading/removing assignments and topologys' files;
SyncProcessEvent is responsible for starting/killing workers;
SupervisorUtils have common methods;
4 create local-supervisor.clj for local mode.
5. fix the fails on supervisor test in Windows

PR #1280, opened 2016-03-30
[STORM-1667] Log the IO exception when deleting worker pid dir

zhuoliu:
We find a race condition in supervisor which can sometimes crash supervisor.
This race condition exists because: shutdown-worker is called by both sync-processes and synchronize-supervisor. One pid directory might have already been deleted when rmr-as-user tries to delete it. 
To avoid crashing the supervisor, we should log the exception rather than throw it.

### DIFF
--- a/storm-core/src/clj/org/apache/storm/daemon/supervisor.clj
+++ b/storm-core/src/clj/org/apache/storm/daemon/supervisor.clj
@@ -297,9 +297,7 @@
           (rmr-as-user conf id (worker-pid-path conf id pid))
           (rmpath (worker-pid-path conf id pid))
           (rmpath (worker-tmp-root conf id pid))
-          (catch IOException e
-            (log-warn-error e "Failed to cleanup pid dir: " pid " for worker " id". Will retry later"))
-          (catch RuntimeException e
+          (catch Exception e
             (log-warn-error e "Failed to cleanup pid dir: " pid " for worker " id". Will retry later")))))
           ;; on windows, the supervisor may still holds the lock on the worker directory
     (try-cleanup-worker conf id))
--- a/storm-core/src/clj/org/apache/storm/daemon/supervisor.clj
+++ b/storm-core/src/clj/org/apache/storm/daemon/supervisor.clj
@@ -269,10 +269,7 @@
   (catch IOException e
     (log-warn-error e "Failed to cleanup worker " id ". Will retry later"))
   (catch RuntimeException e
-    (log-warn-error e "Failed to cleanup worker " id ". Will retry later")
-    )
-  (catch java.io.FileNotFoundException e (log-message (.getMessage e)))
-    ))
+    (log-warn-error e "Failed to cleanup worker " id ". Will retry later"))))
 
 (defn shutdown-worker [supervisor id]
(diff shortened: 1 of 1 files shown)


storm-core/src/clj/org/apache/storm/daemon/supervisor.clj
```diff
@@ -269,10 +269,7 @@
   (catch IOException e
     (log-warn-error e "Failed to cleanup worker " id ". Will retry later"))
   (catch RuntimeException e
-    (log-warn-error e "Failed to cleanup worker " id ". Will retry later")
-    )
-  (catch java.io.FileNotFoundException e (log-message (.getMessage e)))
```
zhuoliu: Delete redundant catch of FileNotFoundException since it has already been caught as IOException

hustfxj:
Thank you. I will merge it [1257] (https://github.com/apache/storm/pull/1257)

storm-core/src/clj/org/apache/storm/daemon/supervisor.clj
```diff
@@ -296,11 +293,15 @@
...
         (try
+          (rmr-as-user conf id (worker-pid-path conf id pid))
           (rmpath (worker-pid-path conf id pid))
           (rmpath (worker-tmp-root conf id pid))
-          (catch Exception e)))) ;; on windows, the supervisor may still holds the lock on the worker directory
+          (catch IOException e
+            (log-warn-error e "Failed to cleanup pid dir: " pid " for worker " id". Will retry later"))
+          (catch RuntimeException e
+            (log-warn-error e "Failed to cleanup pid dir: " pid " for worker " id". Will retry later")))))
+          ;; on windows, the supervisor may still holds the lock on the worker directory
```
satishd: Above code catches `IOException` and `RuntimeException` only. You may still want to catch any `Exception` and ignore it like earlier.

zhuoliu: Hi @satishd , your concern makes good sense, however, I do not find any other type of exception that might be thrown by the try block (the worker-launcher-and-wait in rmr-as-user throws InterruptedException and catches it already). It this is case, catching specific exceptions should be preferable to catching Exception.

satishd: @zhuoliu IMHO,  `Exception` should be caught and logged and we should not really constrain with catching only `IOException` and `RuntimeException`. When any of those methods is changed to throw some other `Exception` then this code should also be changed to handle that, otherwise this code would throw that Exception and get into unexpected scenarios. That may be the reason why `Exception` is caught in earlier code.

zhuoliu: @satishd, got your point. It would be the responsibility of the one who added new code to catch other possible exception, or they may prefer to not catch and let it fail fast upon specific scenario. 
But for now, since there is any non-deterministic exception might be thrown in Windows environment, we may still ignore all types of exception, which might also be why the previous code did this way ("catch Exception").

satishd:
+1 LGTM

HeartSaVioR:
+1 Great finding.

knusbaum:
+1. Will merge tomorrow after 24hr period has passed.

PR #1285, opened 2016-03-31
[STORM-1671] Enable logviewer to delete a dir without yaml

zhuoliu:
For those old and dead worker directories, in some weird case, there is no yaml file in it. We should enable logviewer to delete them. 
Manually tested. It is able to delete an empty workers-artifacts/topo/port dir.


zhuoliu:
Will also cherry-pick this patch to 1.x.

HeartSaVioR:
+1 with assuming manual test has passed.
@zhuoliu Would you like to apply this to 1.0.0? If then adding epic to issue would be better to classify.

zhuoliu:
@HeartSaVioR  Sure. Added this issue to epic Storm-1491. Will apply it to 1.0 once gets merged into master. 
Upmerged.

knusbaum:
+1

ptgoetz:
+1

PR #1287, opened 2016-03-31
STORM-1674: Addresses network flood from KafkaSpout to kafka server.

moesol:
* Allows minBytes in fetch request to be configured
* Defaults new configuration KafkaConfig.fetchMinBytes to 1.
* Defaults fetchMaxWait to 100ms instead of 10000ms.

1) minBytes was 1
2) maxWait was 100

Hopefully the risk of this change is low because
the old behavior can be restored using the API by setting
KafkaConfig.fetchMaxWait to 10000
KafkaConfig.fetchMinBytes to 0

### DIFF
--- a/external/storm-kafka/src/jvm/storm/kafka/KafkaConfig.java
+++ b/external/storm-kafka/src/jvm/storm/kafka/KafkaConfig.java
@@ -30,7 +30,8 @@ public class KafkaConfig implements Serializable {
 
     public int fetchSizeBytes = 1024 * 1024;
     public int socketTimeoutMs = 10000;
-    public int fetchMaxWait = 10000;
+    public int fetchMaxWait = 100;
+    public int fetchMinBytes = 1;
     public int bufferSizeBytes = 1024 * 1024;
     public MultiScheme scheme = new RawMultiScheme();
     public boolean forceFromStart = false;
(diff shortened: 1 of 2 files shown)


knusbaum:
Please file a [Jira](http://issues.apache.org/jira/) for this and update the title of the PR. I think it should be considered a bug, as this could potentially be detrimental to the network bandwidth.

sachingsachin:
Very useful observation @moesol 
Comparing the [KafkaConfig.java](https://github.com/apache/storm/blob/a4f9f8bc5b4ca85de487a0a868e519ddcb94e852/external/storm-kafka/src/jvm/org/apache/storm/kafka/KafkaConfig.java) and [Kafka Configuration](https://kafka.apache.org/08/configuration.html), it seems several other useful configurations like socket.timeout.ms, rebalance.max.retries etc. are also absent from KafkaConfig.java

We should probably add most (if not all) of these options to our spout-configuration.

If we agree that it's useful, then we can add the remaining configuration options to this pull request OR I can submit a new one for the same.

abhishekagarwal87:
This should also go to 1.x-branch given that it is bug and can improve the network bandwidth for lower load topologies

moesol:
@knusbaum I added the JIRA ticket and updated the title a while ago. Are you the person who will actually perform the merge?

HeartSaVioR:
@moesol Sorry to participate too lately.
Could you craft the pull request based on master, 1.x-branch, and 0.10.x-branch? If you mind to craft several pull requests for the same, you can craft only one PR which is against the master branch.

FYI, IMO, since we release Apache Storm 1.0.0, we can't handle 4 main version lines concurrently so fixing this at 0.9.x-branch doesn't make sense for now.

harshach:
@moesol can you please close this PR . Its already merged here https://github.com/apache/storm/commit/f069cfb08099a54448bba3f96f20c5b929ea6af3

moesol:
Merged via different PR. Closing this PR.

PR #1296, opened 2016-04-01
STORM-1675 - Allow submitting multiple jars from the client

abhishekagarwal87:
The PR is not complete yet and needs some refinements. I wanted to get some feedback first. I tried out multiple approaches -
1. Submitting individual jar one by one - Right now the stormjar has a hardcoded key so that both supervisor and nimbus can download/upload the jar. This becomes complicated if there is more than one jar. The keys for all the jars have to be passed from nimbus to supervisor. Then these keys have to be updated, removed, waited--upon-for-code-replication etc. 
2. Build an app-assembly jar programmatically - It is not desired as it changes the internal structure of jar. which can result in unknown issues e.g. manifests have to be handled correctly. 
3. Packing jars into a zip, upload it and unzip it and supervisor - This is the current approach. It seemed to me the cleanest approach. 

I tested out word-count-topology with two jars and they were added to classpath correctly - 
`/usr/local/storm/apache-storm-2.0.0-SNAPSHOT/bin/storm jar /usr/local/storm/apache-storm-2.0.0-SNAPSHOT/lib/kryo-3.0.3.jar:../../examples/storm-starter/target/storm-starter-2.0.0-SNAPSHOT.jar:../../examples/storm-starter/multilang/resources:../../storm-multilang/python/src/main/resources/resources org.apache.storm.starter.WordCountTopology remote`

I can also add an option to turn zip/unzip off for single jar topologies.

### DIFF
--- a/storm-core/src/clj/org/apache/storm/daemon/supervisor.clj
+++ b/storm-core/src/clj/org/apache/storm/daemon/supervisor.clj
@@ -27,7 +27,7 @@
            [org.apache.commons.io FileUtils])
   (:use [org.apache.storm config util log converter local-state-converter])
   (:import [org.apache.storm.generated AuthorizationException KeyNotFoundException WorkerResources])
-  (:import [org.apache.storm.utils NimbusLeaderNotFoundException VersionInfo])
+  (:import [org.apache.storm.utils NimbusLeaderNotFoundException VersionInfo Zipper])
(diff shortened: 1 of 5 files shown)


abhishekagarwal87:
Any comments/suggestions?

HeartSaVioR:
I think STORM-2016 covers this, so we can close this.
@abhishekagarwal87 What do you think?

PR #1297, opened 2016-04-01
STORM-1670 LocalState#get(String) can throw FileNotFoundException which may result supervisor.clj#sync-processes stop assigning new workers/assignments 

satishd:
LocalState#get(String) can throw FileNotFoundException which may result supervisor.clj#sync-processes stop assigning new workers/assignments etc and supervisor goes down later.

### DIFF
--- a/storm-core/src/jvm/org/apache/storm/utils/VersionedStore.java
+++ b/storm-core/src/jvm/org/apache/storm/utils/VersionedStore.java
@@ -18,9 +18,13 @@
 package org.apache.storm.utils;
 
 import java.io.IOException;
+import java.nio.file.DirectoryStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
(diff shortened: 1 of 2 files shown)


harshach:
+1

ptgoetz:
+1

hustfxj:
+1

ptgoetz:
@satishd For some reason this is causing the storm-core clojure tests to fail consistently. Could you take a look?

satishd:
@ptgoetz Tests in `org.apache.storm.versioned-store-test` does not really create a file with the created version but invoked `succeedVersion` which internally creates a `<version-path>.VERSION`. Those tests should have created a file with whatever data that wants to be saved.

ptgoetz:
+1 (again). Confirmed this fixes the test failures on 1.x-branch.

PR #1300, opened 2016-04-02
STORM-1677 Test resource files are excluded from source distribution, which makes logviewer-test failing

HeartSaVioR:
* rename '.log' files to '.log.test' which can avoid exclusion when building source distribution

I can create pull request for 1.x-branch too, but just cherry-picking `d5e1112` would be easier if we cancel voting 1.0.0 RC1 and revert some commits.


vesense:
@HeartSaVioR I'm not sure whether it is caused by `*.log` rule in gitignore.
If *true*, we can apply your patch(`d5e1112 `), or create another patch just change the rule from `*.log` to `derby.log`, or revert the `e58c67a` commit(btw, although I don't recommend using the last way).

HeartSaVioR:
@vesense 
I explained why this happens from JIRA issue, please refer to https://issues.apache.org/jira/browse/STORM-1677. 
'log' files are already stored to git repo but not included to source distribution.

ptgoetz:
+1

PR #1303, opened 2016-04-03
STORM-1678: abstract batch processing to common api `BatchHelper`

vesense:
Some external projects(eg: `storm-hbase`, `storm-hive`, `storm-mongodb`) have the same batch processing logic.
Abstracting for that could make code simpler and cleaner.


HeartSaVioR:
+1 Nice work.

PR #1307, opened 2016-04-04
[STORM-1681] - Bug in scheduling cyclic topologies when scheduling with RAS

jerrypeng:
There is a bug in the bfs algorithm in RAS that does not correctly account for components already visited during the breadth first traveral


jerrypeng:
Non-related failure

revans2:
+1

PR #1314, opened 2016-04-06
[STORM-1690] fix backpressure flag initialization

zhuoliu:
Found that in different DisruptorQueues in a worker, the _enableBackpressure flags have both true and false, which cause topology to be blocked.


redsanket:
+1

zhuoliu:
Sorry, this is not a bug. No need to fix.

PR #1316, opened 2016-04-07
[STORM-1642] Catch Exception when deserialization failed. 

liurenjie1024:
This patch aims to fix bug STORM-1642.  When we failed to deserialize a tuple, we should skip this tuple and print a log rather than just shutdown the work.


abhishekagarwal87:
How was deserialization failure resulting in NPE? Other than that, deserialization failure is the problem with an application itself which application should correct by registering correct serializer/deserializer. storm framework should not skip such tuples on its own.

liurenjie1024:
As I have mentioned in bug report, the reason why NPE happens is that storm is vulnerable to fraud message from  processes outside the cluster. To reproduce the NPE, you just need to send a message [taskid 0] to [host]:[port] from anywhere, where taskid is id of one of the tasks running on [host]:[port]. In this case, storm will return a TaskMessage with payload set to null to the deserializer. 
Storm does not check whether the task message is from processes within the cluster, so deserialization may fail. I  think storm should skip the fruad task message rather than shutdowning the worker when deserialization failed.

abhishekagarwal87:
I can't think of a reason of why would such a message be sent to storm from outside the cluster. Ideally only cluster machines and daemons should have access to the worker ports. Or if it indeeds needs to be solved, then a better method would be to ignore the zero length payload and not add `new TaskMessage(task, null)`.  Before you make any change, let's wait for comments from someone who is more familiar with this part of code.

liurenjie1024:
The reason why such a message is sent is that our security team is scanning all the ports listened on each server and tries to detect potential weakness.
At first thought I just wanted to ignore messages with zero payload, but I think ignoring messages failed to be deserialized would be more robust. I think this also matches storm's design since it throws away messages whose target task id is not in the worker and this patch is just a complement.

storm-core/src/clj/backtype/storm/daemon/executor.clj
```diff
@@ -431,13 +431,14 @@
...
-              (fast-list-iter [task-id task-ids]
+            (if tuple
+              ((when debug? (log-message "Processing received message FOR " task-id " TUPLE: " tuple))
+              (if task-id
                 (tuple-action-fn task-id tuple)
-                ))
+                ;; null task ids are broadcast tuples
+                (fast-list-iter [task-id task-ids]
+                  (tuple-action-fn task-id tuple)
+                  ))))
```
knusbaum: Lots of parenthetical and formatting problems here.

knusbaum:
Why not no both? I don't see any reason to propagate `TaskMessage`s with null payloads, but I agree we should be dropping tuples that failed to deserialize in the iterator. 

This is very much on the critical path, so someone else might want to weigh in.

liurenjie1024:
NPE is just a special case of deserialization failure, right?

knusbaum:
What do you mean?

If you mean null payloads are a special case, I disagree. In fact, it's not a failure at all. We received a null message, so we can perform a NOP as soon as possible.

liurenjie1024:
Yes, I agree with you that we should drop a task message with null payload as soon as possible, and I'll push another commit to drop task messages with null payload.

storm-core/src/jvm/backtype/storm/messaging/netty/MessageBatch.java
```diff
@@ -112,6 +115,9 @@ private void writeTaskMessage(ChannelBufferOutputStream bout, TaskMessage messag
             throw new RuntimeException("Task ID should not exceed "+Short.MAX_VALUE);
         
         bout.writeShort((short)task_id);
+        if (payload_len == 0) {
+            LOG.warn("Zero length payload to task {}.", (short)task_id);
```
gzm55: is this a debugging log? should remove them?

knusbaum:
If you want to clean up the log messages, this still looks good. We'd want this instead to go into master and 1.x, though.

PR #1320, opened 2016-04-07
[STORM-1696]-1.x-branch status not sync if zk fails in backpressure

zhuoliu:
When there is a zk exception happens during worker-backpressure!,
there is a bad state which can block the topology from running normally any more.

The root cause: in worker/mk-backpressure-handler
if the worker-backpressure! fails once due to zk connection exception,
next time when this method gets called by WordBackpressureThread, because (when (not= prev-backpressure-flag curr-backpressure-flag) will never be true, the remote zk node can not be synced with local state.
This problem can cause a topology to be blocked!

This also explains why we will not see any problem when testing in a stable (zk never fail) environment.

Solution is quite straightforward: first change the zk status, if succeeds, change local status.
This change has been manually tested for a topology that causes congestion.

### DIFF
--- a/storm-core/src/clj/org/apache/storm/daemon/worker.clj
+++ b/storm-core/src/clj/org/apache/storm/daemon/worker.clj
@@ -138,16 +138,19 @@
             assignment-id (:assignment-id worker)
             port (:port worker)
             storm-cluster-state (:storm-cluster-state worker)
-            prev-backpressure-flag @(:backpressure worker)]
-        (when executors
-          (reset! (:backpressure worker)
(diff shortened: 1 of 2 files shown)


zhuoliu:
I can either put the refactoring part as a separate pull request or within this.

HeartSaVioR:
@zhuoliu 
+1 Nice finding. 
Could you address this to master branch, too?

zhuoliu:
@HeartSaVioR sure. Fix for master branch: https://github.com/apache/storm/pull/1338

PR #1322, opened 2016-04-08
STORM-1698 Asynchronous MetricsConsumerBolt

HeartSaVioR:
* change MetricsConsumerBolt's behavior to asynchronus manner
  * to avoid bad side effect of topology
* for details please refer JIRA issue: https://issues.apache.org/jira/browse/STORM-1698

### DIFF
--- a/conf/storm.yaml.example
+++ b/conf/storm.yaml.example
@@ -41,8 +41,10 @@
 ## Metrics Consumers
 # topology.metrics.consumer.register:
 #   - class: "org.apache.storm.metric.LoggingMetricsConsumer"
+#     max.retain.metric.tuples: 100
 #     parallelism.hint: 1
 #   - class: "org.mycompany.MyMetricsConsumer"
+#     max.retain.metric.tuples: 100
 #     parallelism.hint: 1
 #     argument:
 #       - endpoint: "metrics-collector.mycompany.org"
--- a/storm-core/src/jvm/org/apache/storm/daemon/StormCommon.java
+++ b/storm-core/src/jvm/org/apache/storm/daemon/StormCommon.java
@@ -383,10 +383,12 @@ public class StormCommon {
             for (Map<String, Object> info : registerInfo) {
                 String className = (String) info.get("class");
                 Object argument = info.get("argument");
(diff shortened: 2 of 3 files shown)


abhishekagarwal87:
The _taskQueue is unbounded queue and if metric consumer bolt is indeed slow, this queue would keep growing, leading to another set of problems. Though it is probably ok as a short term fix.

HeartSaVioR:
@abhishekagarwal87 
Yes, I also have two ideas to avoid or minimize those issues.

1) discard policy
2) filter metrics

About 1) _taskQueue will become bounded queue. And when it's full, we discard some enqueued tasks via policy. Policy could be drop oldest, drop latest, drop randomly, or even provide interface for plug-in.

About 2), actually there're many metrics provided by Storm by default. Furthermore, some modules like storm-kafka provides their metrics, too. If users set pattern what they want to subscribe or filter out that would be great.

I'm going to create an umbrella issue regarding improvement of metrics and add STORM-1698 as subtask.

abhishekagarwal87:
Filtering would be a good feature to add. I have also been trying to gather some opinion on in-process aggregation at component level, without much success :) Codahale metric registry is a great alternative but that would translating many metrics in the storm-core to codahale compliant implementations e.g. CountMetric to Counter. I am willing to spend that effort if approach sounds good.

HeartSaVioR:
You got me. :)
I was also finding aggregation metrics on worker side, but after struggling I gave it. 
Why I gave is that IMetric.getValueAndReset() is not multi-threads friendly. Currently metrics are updated at receive queue handler thread of executors, and that thread also handles metrics tick tuple so there's no race condition. 
Maybe most way of in process aggregation should resolve this issue.

HeartSaVioR:
gave it -> gave up

HeartSaVioR:
Storm already use codahale metric registery but it's used for exposing internal metrics via JXM.

HeartSaVioR:
I'm addressing the (2), and being wondering that we have to address (1) since it drops metrics data anyway so it should be avoided.

unsleepy22:
I suppose this PR has been merged into https://github.com/apache/storm/pull/1324 ? I'll skip this one.

HeartSaVioR:
Will be also addressed in #1324, closing.

PR #1323, opened 2016-04-08
STORM-1698 Asynchronous MetricsConsumerBolt (1.x)

HeartSaVioR:
* change MetricsConsumerBolt's behavior to asynchronus manner
  * to avoid bad side effect of topology
* for details please refer JIRA issue: https://issues.apache.org/jira/browse/STORM-1698

PR #1325, opened 2016-04-09
STORM-1700 Introduce 'whitelist' / 'blacklist' option to MetricsConsumer (1.x)

HeartSaVioR:
> NOTE: This is based on STORM-1698, PR #1323  

* Users can set whitelist or blacklist to filter out metrics by name
  * if none of them specified (by default), no metrics are filtered out
* how to match: substring match with regular expression
  * use both ^ and $ when you want to match strictly (full string match)
* added unit test

### DIFF
--- a/conf/storm.yaml.example
+++ b/conf/storm.yaml.example
@@ -41,8 +41,10 @@
 ## Metrics Consumers
 # topology.metrics.consumer.register:
 #   - class: "org.apache.storm.metric.LoggingMetricsConsumer"
+#     max.retain.metric.tuples: 100
 #     parallelism.hint: 1
 #   - class: "org.mycompany.MyMetricsConsumer"
+#     max.retain.metric.tuples: 100
 #     parallelism.hint: 1
 #     argument:
 #       - endpoint: "metrics-collector.mycompany.org"
--- a/conf/storm.yaml.example
+++ b/conf/storm.yaml.example
@@ -39,12 +39,21 @@
 #     - "server2"
 
 ## Metrics Consumers
+## max.retain.metric.tuples
+## - task queue will be unbounded when max.retain.metric.tuples is equal or less than 0.
+## whitelist / blacklist
+## - when none of configuration for metric filter are specified, it'll be treated as 'pass all'.
+## - you need to specify either whitelist or blacklist, or none of them. You can't specify both of them.
+## - you can specify multiple whitelist / blacklist with regular expression
 # topology.metrics.consumer.register:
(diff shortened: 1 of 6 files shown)


HeartSaVioR:
I'll rebase before merging.

storm-core/src/clj/org/apache/storm/daemon/common.clj
```diff
@@ -298,18 +299,21 @@
                       {[comp-id METRICS-STREAM-ID] :shuffle})
                     (into {}))
         
-        mk-bolt-spec (fn [class arg p]
+        mk-bolt-spec (fn [class arg p max-retain-metric-tuples whitelist blacklist]
                        (thrift/mk-bolt-spec*
                         inputs
-                        (org.apache.storm.metric.MetricsConsumerBolt. class arg)
+                        (org.apache.storm.metric.MetricsConsumerBolt. class arg max-retain-metric-tuples (FilterByMetricName. whitelist blacklist))
```
ptgoetz: Are MetricsFilters intended to be pluggable, or do we only want to allow filtering by name?

conf/storm.yaml.example
```diff
@@ -39,10 +39,13 @@
 #     - "server2"
 
 ## Metrics Consumers
+## NOTE: task queue will be unbounded when max.retain.metric.tuples is equal or less than 0.
 # topology.metrics.consumer.register:
 #   - class: "org.apache.storm.metric.LoggingMetricsConsumer"
+#     max.retain.metric.tuples: 100
 #     parallelism.hint: 1
 #   - class: "org.mycompany.MyMetricsConsumer"
+#     max.retain.metric.tuples: 100
```
ptgoetz: It would be helpful to have an example of how to specify include/exclude filters (whitelist/blacklist).

storm-core/src/clj/org/apache/storm/daemon/common.clj
HeartSaVioR: @ptgoetz 
I intended to make it pluggable, but if we just want to allow filtering by name, pluggable incurs unneeded specifying class name. So flexible vs easy to use.
I'd like to get opinions on this side, and happy to change when we want to make it flexible.

conf/storm.yaml.example
HeartSaVioR: @ptgoetz Good. I'll address according to how we decide to make it pluggable vs flexible.

harshach:
+1

ptgoetz:
+1

HeartSaVioR:
Github fails to track this branch is merged into master. Already merged. Closing.

PR #1326, opened 2016-04-10
STORM-1693: Do not cancel the task after stats are rendered

abhishekagarwal87:
Fresher task is cancelled right after 3 seconds when executor heartbeat is sent. getValueAndReset can rotate the buckets but it is not time bound. Metric tick is in same queue as the input tuples from spout. If processing input tuples is slow, it may take much more to generate metrics and call getValueAndReset and thus `System.currentTimeMillis() - _bucketStart` is more than 10 minutes.

### DIFF
--- a/storm-core/src/clj/org/apache/storm/stats.clj
+++ b/storm-core/src/clj/org/apache/storm/stats.clj
@@ -153,25 +153,25 @@
   [^SpoutExecutorStats stats stream latency-ms]
   (.incBy ^MultiCountStatAndMetric (stats-failed stats) stream (stats-rate stats)))
 
-(defn- cleanup-stat! [stat]
+(defn- close-stat! [stat]
   (.close stat))
 
 (defn- cleanup-common-stats!
   [^CommonStats stats]
   (doseq [f COMMON-FIELDS]
-    (cleanup-stat! (f stats))))
+    (close-stat! (f stats))))
 
 (defn cleanup-bolt-stats!
   [^BoltExecutorStats stats]
   (cleanup-common-stats! (:common stats))
   (doseq [f BOLT-FIELDS]
-    (cleanup-stat! (f stats))))
+    (close-stat! (f stats))))
 
 (defn cleanup-spout-stats!
   [^SpoutExecutorStats stats]
   (cleanup-common-stats! (:common stats))
   (doseq [f SPOUT-FIELDS]
-    (cleanup-stat! (f stats))))
+    (close-stat! (f stats))))
 
 (defn- value-stats
   [stats fields]
@@ -188,14 +188,12 @@
 
 (defn value-bolt-stats!
   [^BoltExecutorStats stats]
-  (cleanup-bolt-stats! stats)
   (merge (value-common-stats (:common stats))
          (value-stats stats BOLT-FIELDS)
          {:type :bolt}))
 
 (defn value-spout-stats!
   [^SpoutExecutorStats stats]
-  (cleanup-spout-stats! stats)
   (merge (value-common-stats (:common stats))
          (value-stats stats SPOUT-FIELDS)
          {:type :spout}))
@@ -209,6 +207,16 @@
 (defmethod render-stats! BoltExecutorStats
   [stats]
   (value-bolt-stats! stats))
+
+(defmulti cleanup-stats! class-selector)
+
+(defmethod cleanup-stats! SpoutExecutorStats
+  [stats]
+  (cleanup-spout-stats! stats))
+
+(defmethod cleanup-stats! BoltExecutorStats
+  [stats]
+  (cleanup-bolt-stats! stats))
 
 (defmulti thriftify-specific-stats :type)
 (defmulti clojurify-specific-stats class-selector)
--- a/storm-core/src/jvm/org/apache/storm/metric/internal/RateTracker.java
+++ b/storm-core/src/jvm/org/apache/storm/metric/internal/RateTracker.java
@@ -17,7 +17,6 @@
  */
 package org.apache.storm.metric.internal;
 
-import java.util.Timer;
 import java.util.TimerTask;
 import java.util.concurrent.atomic.AtomicLong;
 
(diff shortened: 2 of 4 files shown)


abhishekagarwal87:
Remove the _task.cancel from other places and redundant cleanup calls. Build failure doesn't seem to be related.

storm-core/src/jvm/org/apache/storm/metric/internal/CountStatAndMetric.java
```diff
@@ -197,12 +196,6 @@ long readApproximateTime(long value, long timeSpent, long[] bucketTime, long[] b
         return total;
     }
 
-    public void close() {
-        if (_task != null) {
-            _task.cancel();
```
revans2: If we don't cancel it we leak memory and other things in the tests.  We at least need to close the stat when we are all done, not after calling render-stats.

storm-core/src/jvm/org/apache/storm/metric/internal/LatencyStatAndMetric.java
```diff
@@ -248,12 +248,6 @@ private synchronized void rotate(long lat, long count, long timeSpent, long targ
         return Utils.zeroIfNaNOrInf(((double) totalLat) / totalCount);
     }
 
-    public void close() {
-        if (_task != null) {
-            _task.cancel();
```
revans2: Here too we do need to cancel the task eventually.

storm-core/src/jvm/org/apache/storm/metric/internal/MultiCountStatAndMetric.java
```diff
@@ -103,10 +103,4 @@ public Object getValueAndReset() {
         }
         return ret;
     }
-
-    public void close() {
-        for (CountStatAndMetric cc: _counts.values()) {
-            cc.close();
```
revans2: We still need this too.

storm-core/src/jvm/org/apache/storm/metric/internal/MultiLatencyStatAndMetric.java
```diff
@@ -100,10 +100,4 @@ public Object getValueAndReset() {
         }
         return ret;
     }
-
-    public void close() {
-        for (LatencyStatAndMetric l: _lat.values()) {
-            l.close();
```
revans2: And here too.

storm-core/src/jvm/org/apache/storm/metric/internal/RateTracker.java
```diff
@@ -100,12 +99,6 @@ synchronized double reportRate(long currentTime) {
         return events * 1000.0 / duration;
     }
 
-    public void close() {
-        if (_task != null) {
-            _task.cancel();
```
revans2: And here we need it.

storm-core/src/jvm/org/apache/storm/metric/internal/CountStatAndMetric.java
abhishekagarwal87: @revans2 - As far as I could see, close is being called only in the render-stats. Tests initialize with negative number of buckets so _task is never initialized at all.

storm-core/src/clj/org/apache/storm/stats.clj
```diff
@@ -188,14 +168,12 @@
...
 (defn value-bolt-stats!
   [^BoltExecutorStats stats]
-  (cleanup-bolt-stats! stats)
   (merge (value-common-stats (:common stats))
          (value-stats stats BOLT-FIELDS)
          {:type :bolt}))
 
 (defn value-spout-stats!
   [^SpoutExecutorStats stats]
-  (cleanup-spout-stats! stats)
```
revans2: Really the issue is just that cleanup-bolt-stats! and cleanup-spout-stats! needs to move from value-*-stats! to something that is called from the executor shutdown.  

https://github.com/abhishekagarwal87/storm/blob/STORM-1693/storm-core/src/clj/org/apache/storm/daemon/executor.clj#L409-L417

I thought that is where they were called from when I saw them, but I obviously was wrong.

storm-core/src/jvm/org/apache/storm/metric/internal/RateTracker.java
revans2: It should have been called when the disruptor queue shuts down, but it wasn't.  Another bug.

https://github.com/abhishekagarwal87/storm/blob/STORM-1693/storm-core/src/jvm/org/apache/storm/utils/DisruptorQueue.java#L392-L400

storm-core/src/clj/org/apache/storm/stats.clj
abhishekagarwal87: Thanks @revans2 . Cancelling the task during the executor shutdown makes sense. I will modify the PR accordingly and restore the close functions.

revans2:
@abhishekagarwal87 Thanks for finding/fixing these issues.  I don't think we want to just rip out the shutdown logic, it can cause leaks.  I would really prefer for us to fix the shutdown logic so that we do it properly.

abhishekagarwal87:
I will add the shutdown for both Multi*Metrics and RateTracker in executor shutdown flow. I suppose that is the only time, these fresher tasks have to be cancelled?

revans2:
The only reason to cancel them is to avoid a memory leak, and mostly in the tests, because for the worker when we do it when the worker is shutting down.

abhishekagarwal87:
@revans2 - Addressed your comments.

revans2:
Great +1.  But we need to do the same fixes for master too.  If this is blocking the 1.x release then I am OK with doing it on a follow on JIRA, but if it is not blocking then I would prefer to have them go in together so we don't violate process.

abhishekagarwal87:
No problem. I will open a PR for master branch as well.

abhishekagarwal87:
PR for master: https://github.com/apache/storm/pull/1333

HeartSaVioR:
+1

PR #1329, opened 2016-04-11
STORM-1704 When logviewer_search.html opens daemon file, next search always show no result

HeartSaVioR:
https://issues.apache.org/jira/browse/STORM-1704

* ensures that is-daemon parameter is passed around multiple searches
* set logviewerUrl to '/daemonlog' when search is done with is-daemon=yes


harshach:
+1. @HeartSaVioR should we merge this for 1.x-branch as well.

HeartSaVioR:
@harshach Yes sure, #1330 is ready for reviewing. Could you review that as well?

PR #1330, opened 2016-04-11
STORM-1704 (1.x) When logviewer_search.html opens daemon file, next search always show no result

HeartSaVioR:
https://issues.apache.org/jira/browse/STORM-1704

For master please refer #1329.

* ensures that is-daemon parameter is passed around multiple searches
* set logviewerUrl to '/daemonlog' when search is done with is-daemon=yes


harshach:
+1

PR #1331, opened 2016-04-11
STORM-1705: Cap number of retries for a failed message

abhishekagarwal87:
This PR also makes retry strategy pluggable.

### DIFF
--- a/external/storm-kafka/src/jvm/org/apache/storm/kafka/ExponentialBackoffMsgRetryManager.java
+++ b/external/storm-kafka/src/jvm/org/apache/storm/kafka/ExponentialBackoffMsgRetryManager.java
@@ -27,17 +27,25 @@ import java.util.concurrent.ConcurrentHashMap;
 
 public class ExponentialBackoffMsgRetryManager implements FailedMsgRetryManager {
 
-    private final long retryInitialDelayMs;
-    private final double retryDelayMultiplier;
-    private final long retryDelayMaxMs;
+    private long retryInitialDelayMs;
+    private double retryDelayMultiplier;
+    private long retryDelayMaxMs;
+    private int retryLimit;
 
-    private Queue<MessageRetryRecord> waiting = new PriorityQueue<MessageRetryRecord>(11, new RetryTimeComparator());
-    private Map<Long,MessageRetryRecord> records = new ConcurrentHashMap<Long,MessageRetryRecord>();
+    private Queue<MessageRetryRecord> waiting;
+    private Map<Long,MessageRetryRecord> records;
 
-    public ExponentialBackoffMsgRetryManager(long retryInitialDelayMs, double retryDelayMultiplier, long retryDelayMaxMs) {
-        this.retryInitialDelayMs = retryInitialDelayMs;
-        this.retryDelayMultiplier = retryDelayMultiplier;
-        this.retryDelayMaxMs = retryDelayMaxMs;
+    public ExponentialBackoffMsgRetryManager() {
+
+    }
+
+    public void prepare(SpoutConfig spoutConfig) {
+        this.retryInitialDelayMs = spoutConfig.retryInitialDelayMs;
+        this.retryDelayMultiplier = spoutConfig.retryDelayMultiplier;
+        this.retryDelayMaxMs = spoutConfig.retryDelayMaxMs;
+        this.retryLimit = spoutConfig.retryLimit;
+        this.waiting = new PriorityQueue<MessageRetryRecord>(11, new RetryTimeComparator());
+        this.records = new ConcurrentHashMap<Long,MessageRetryRecord>();
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


knusbaum:
+1 LGTM

external/storm-kafka/pom.xml
```diff
@@ -129,7 +129,7 @@
             <groupId>org.apache.storm</groupId>
             <artifactId>storm-core</artifactId>
             <version>${project.version}</version>
-            <scope>provided</scope>
+            <!--scope>provided</scope-->
```
HeartSaVioR: Is it intended?

abhishekagarwal87: good catch @HeartSaVioR . I had to make this change for in-IDE testing. I will revert this.

abhishekagarwal87: Reverted the changes in pom.

external/storm-kafka/src/jvm/org/apache/storm/kafka/PartitionManager.java
```diff
@@ -46,6 +46,9 @@
     private final CountMetric _fetchAPIMessageCount;
```
harshach: since we are refactoring this code. Can we simplify partitionManager constructor instead of the current state we can have few private methods for all the code in the constructor.

abhishekagarwal87: Agree that constructor is somewhat bloated. The whole class can use some refactoring. can we do it in a separate PR?

abhishekagarwal87: @harshach - is it fine with you to do the refactoring separately? Apologies for a long pause.

abhishekagarwal87:
@harshach - ping

harshach:
+1

HeartSaVioR:
+1
Though it modifies public interface FailedMsgRetryManager, but it was only used to storm-kafka internally.
Btw, when we push this and release, we should respect backward compatibility for FailedMsgRetryManager since it will be used to pluggable implementation.

abhishekagarwal87:
I agree @HeartSaVioR. Happy to incorporate feedback on the interface improve it further. One limitation I see is that interface takes offsets. It is not possible to add an implementation which can store failed messages in memory and return them (like new Kafka spout). But the interface still provides the control over actions on message retry e.g. one can log the offset in a log file if it exhausts the number of retries etc.

abhishekagarwal87:
can this be merged? Please let me know if I need to raise a separate PR for master.

HeartSaVioR:
@abhishekagarwal87 
Could you please add pluggable FailedMsgRetryManager to README of storm-kafka to let users know how to use? Thanks in advance.

external/storm-kafka/src/jvm/org/apache/storm/kafka/SpoutConfig.java
```diff
@@ -37,6 +37,8 @@
     public long retryInitialDelayMs = 0;
     public double retryDelayMultiplier = 1.0;
     public long retryDelayMaxMs = 60 * 1000;
+    public int retryLimit = Integer.MAX_VALUE;
```
ptgoetz: This is probably an unlikely edge case, but theoretically possible: If a tuple is retried `Integer.MAX_VALUE` times, it will be dropped.

To truly turn off retry limits, I would suggest setting this to `-1` and adding logic to check for that in logic that checks to see if the retry limit has been reached.

external/storm-kafka/src/jvm/org/apache/storm/kafka/ExponentialBackoffMsgRetryManager.java
```diff
@@ -86,15 +94,23 @@ public Long nextFailedMessageToRetry() {
     }
 
     @Override
-    public boolean shouldRetryMsg(Long offset) {
+    public boolean shouldReEmitMsg(Long offset) {
         MessageRetryRecord record = this.records.get(offset);
         return record != null &&
                 this.waiting.contains(record) &&
                 System.currentTimeMillis() >= record.retryTimeUTC;
```
ptgoetz: I would be nice to log a WARN message when we are dropping messages so users have a way of knowing when/why it is happening.

ptgoetz:
Two minor nits. I'm +1 once they are addressed.

abhishekagarwal87:
Thanks @ptgoetz and @HeartSaVioR . One thing I missed in my implementation is how to pass any additional configuration to failedMsgRetryManager. I can either add an extra HashMap or pass stormConf in the prepare method of failedMsgRetryManager. It's minor but still any suggestions?

abhishekagarwal87:
I have taken care of comments and added documentation in storm-kafka readme. Please review the documentation as well.

ptgoetz:
+1

abhishekagarwal87:
@ptgoetz - unfortunately there was a bug in this PR. I have raised #1481 to fix that bug. Sorry about that.

PR #1335, opened 2016-04-12
[STORM-1646] Fix Kafka unit tests

ppoulosk:
This fixes two unit test failures in external/kafka I saw when running tests repeatedly, especially on low resource VMs.

There was a race in the Exponential Backoff Manager that was fixed by switching the test to use simulated time.

The Kafka Utils Test was seeing failures on underpowered systems due to a timeout value that was too small. Added log4j2 properties xml to be able to change the log level printed in the output. Also changed metadata.fetch.timeout.ms to max.block.ms, as kafka 0.9 had deprecated the config. The test output had warnings like

39890 [main] WARN o.a.k.c.p.KafkaProducer - metadata.fetch.timeout.ms config is deprecated and will be removed soon. Please use max.block.ms

### DIFF
--- a/external/storm-kafka/src/jvm/org/apache/storm/kafka/ExponentialBackoffMsgRetryManager.java
+++ b/external/storm-kafka/src/jvm/org/apache/storm/kafka/ExponentialBackoffMsgRetryManager.java
@@ -17,6 +17,7 @@
  */
 package org.apache.storm.kafka;
 
+import org.apache.storm.utils.Time;
 import java.util.Comparator;
 import java.util.HashSet;
 import java.util.Map;
@@ -72,7 +73,7 @@ public class ExponentialBackoffMsgRetryManager implements FailedMsgRetryManager 
     public Long nextFailedMessageToRetry() {
         if (this.waiting.size() > 0) {
             MessageRetryRecord first = this.waiting.peek();
-            if (System.currentTimeMillis() >= first.retryTimeUTC) {
+            if (Time.currentTimeMillis() >= first.retryTimeUTC) {
                 if (this.records.containsKey(first.offset)) {
                     return first.offset;
                 } else {
@@ -90,7 +91,7 @@ public class ExponentialBackoffMsgRetryManager implements FailedMsgRetryManager 
         MessageRetryRecord record = this.records.get(offset);
         return record != null &&
                 this.waiting.contains(record) &&
-                System.currentTimeMillis() >= record.retryTimeUTC;
+                Time.currentTimeMillis() >= record.retryTimeUTC;
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


kishorvpatil:
LGTM. +1

revans2:
+1

srdo:
There's also a metadata.fetch.timeout.ms here 
https://github.com/apache/storm/blob/master/external/storm-kafka/src/test/org/apache/storm/kafka/bolt/KafkaBoltTest.java#L211

vesense:
LGTM

ppoulosk:
Changed usage of metadata.fetch.timeout.ms in KafkaBoltTest as well.

jerrypeng:
+1

HeartSaVioR:
`max.block.ms` is not available for 0.8.x so IMO it would be not good to change storm-kafka module to strictly support 0.9 or higher. Users who use Kafka 0.9 or higher could use storm-kafka-client even though storm-kafka is not compatible with 0.9, but not opposite.

ppoulosk:
@HeartSaVioR,

I assumed that because storm-kafka already builds against kafka 0.9 and that the changes to max.block.ms were only in unit test code it would not affect any user compatibility.  Is that wrong?

HeartSaVioR:
Issue is that some users report that storm-kafka 1.0.0 with 0.9 is not working properly.

Please refer https://issues.apache.org/jira/browse/STORM-1755 and #1386.
When we adjust the version into 0.8.x, that option becomes useless for test codes, which could incur failures which seems hard to find.

But using `metadata.fetch.timeout.ms` wouldn't make any issues for Kafka 0.8 and 0.9. And I don't think storm-kafka can support Kafka same or upper than 0.10.0 clearly so no issue for using deprecated property. But this is just my 2 cents and I'd like to hear about other opinions, too.

srdo:
storm-kafka will probably support 0.10.0, I believe the Kafka developers version their APIs so new clusters can talk to old clients. That shouldn't affect the choice of configuration parameters though, since any 0.8.x support in 0.10.0 would expect 0.8.x configuration parameters.

See https://github.com/apache/kafka/blob/trunk/core/src/main/scala/kafka/api/ApiVersion.scala and how it's used in https://github.com/apache/kafka/blob/trunk/core/src/main/scala/kafka/server/KafkaApis.scala if you're interested. The way I understand it is the client indicates which version of the API it wants to use, and then the broker uses that version of the API for requests/responses.

ppoulosk:
Thanks for the explanation.  Will revert that change.

srdo:
I might have been a bit quick on the trigger there. The Kafka coding guidelines target backward compatibility for one release only. So storm-kafka might not support 0.10.0.

(http://kafka.apache.org/coding-guide.html under backwards compatibility)

HeartSaVioR:
Yeah, I guess Kafka doesn't strictly respect semantic versioning, or their major version seems to 2nd number of the version at least before reaching 1.0.0.

ppoulosk:
Replaced by https://github.com/apache/storm/pull/1658

PR #1336, opened 2016-04-13
[STORM-1709] Added group by support in storm sql standalone mode

arunmahadevan:
Added group by support for storm-sql in standalone mode and handle user defined aggregation operations as well.


arunmahadevan:
@harshach @haohui can you take a look?

harshach:
+1

PR #1337, opened 2016-04-13
STORM-1475: Add storm-elasticsearch2 module

dossett:
This adds external/storm-elasticsearch2.  es2 is basically a direct port of the original module, but updated for ES 2.0.1 API changes as needed.

The EsIndexBolt also adds options around:
- tick tuples
- batch updates
  -synchronous vs. asynchronous modes

Open to better ways to support ES1 and ES2 side by side.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -223,7 +223,7 @@
         <clojure.tools.cli.version>0.2.4</clojure.tools.cli.version>
         <disruptor.version>3.3.2</disruptor.version>
         <jgrapht.version>0.9.0</jgrapht.version>
-        <guava.version>16.0.1</guava.version>
+        <guava.version>18.0</guava.version>
         <netty.version>3.9.0.Final</netty.version>
         <sysout-over-slf4j.version>1.0.2</sysout-over-slf4j.version>
         <log4j-over-slf4j.version>1.6.6</log4j-over-slf4j.version>


dossett:
The build failure reminded of something I had somehow suppressed. Elastic depends on guava 18, while storm is currently at 16.  Now that we're building toward 2.0, is this an appropriate place to upgrade?  I upgraded locally and all tests passed.  Committing that for the travis build and further discussion here.

harshach:
@dossett is there any way to merge the existing module with new one? instead of having another module

dossett:
@harshach I could not find a way to keep it in one module and include both ES 1.X and 2.X dependencies.  I will be out of the office for a couple of weeks, but will try again when I return. I'll also look at the rat errors from the travis failure.

mazhechao:
Yes, ES 2.x has released for a long peroid of time. It's a good news for us if storm support ES 2.x.

dossett:
@HeartSaVioR You commented on ES 2.X in another PR (#1396).  This PR adds an ES2.X bolt but essentially duplicates the code in the regular ES bolt (with very minor differences).  Do you have thoughts on this approach?  I would love to avoid duplicate code bases, but don't have a clear idea of how to do that.

HeartSaVioR:
@dossett 
ES2 doesn't shade their Guava dependency? I guess ES1 is doing that.
I'm not sure we're OK to upgrade Guava version only for this, but there's another PR which gets rid of Guava from storm-core, so it might be OK.

knusbaum:
#1676 is a no-go. We need guava in storm-core.

PR #1340, opened 2016-04-14
STORM-1682: Refactor DynamicBrokersReader to lookup leader informatio…

srdo:
…n via Kafka metadata API. Put Zookeeper metadata lookup into new class.

See https://issues.apache.org/jira/browse/STORM-1682.

This solution splits DynamicBrokersReader into two. Instead of looking up leaders and brokers in Zookeeper, ZkBrokersReader will use the brokers in Zookeeper to construct one or more SimpleConsumers, which DynamicBrokersReader can then use to lookup partition leaders.

DynamicBrokersReader should now cause the spout to restart in the following scenarios:
If it can't find any brokers in Zookeeper
If it can't contact any of the brokers it found in Zookeeper
If wildcard topics are listed in Zookeeper, and no broker has metadata for all of them

In order to test behavior when brokers shut down, it was necessary to separate the Kafka test broker from Zookeeper so multiple brokers can run in a cluster, which is the reason for modifications in many tests.


srdo:
This runs reliably for me locally, Travis seemed to run into a heap space limit.

srdo:
Closing this temporarily. I need to check that this works on 0.8.x

srdo:
The tests pass on 0.8.2.1, which is the version targeted by https://github.com/apache/storm/pull/1386. Reopening.

abhishekagarwal87:
Given that partitions information is refreshed every 30 seconds, and this particular exception is rare, is this change necessary? Also when the exception occurs, does it kill the worker jvm?

srdo:
It's not strictly necessary, no. It's just an alternate way of looking up leaders that avoids the concurrency issue when looking up the leader in Zookeeper. As a side effect, it may give more informative error logs when the metadata API returns errors, than being told that a node in Zookeeper is missing.

Yes, the RuntimeExceptions will crash the worker. I put them in before realizing that refresh was called periodically. I could replace them with error logs?

srdo:
Closing this, we aren't using this spout anymore and it's a big change to fix some very minor pausing

PR #1343, opened 2016-04-15
[STORM-1714] StatefulBolts ends up as normal bolts while using TopologyBuilder.setBolt without parallelism

arunmahadevan:
Refactored IstatefulBolt to be independent of IRichBolt so that setBolt overloads in TopologyBuilder can work correctly.

### DIFF
--- a/storm-core/src/jvm/org/apache/storm/topology/CheckpointTupleForwarder.java
+++ b/storm-core/src/jvm/org/apache/storm/topology/CheckpointTupleForwarder.java
@@ -53,6 +53,10 @@ public class CheckpointTupleForwarder implements IRichBolt {
     private long lastTxid = Long.MIN_VALUE;
     private AnchoringOutputCollector collector;
 
+    public CheckpointTupleForwarder() {
+        this(null);
+    }
+
     public CheckpointTupleForwarder(IRichBolt bolt) {
         this.bolt = bolt;
         transactionRequestCount = new HashMap<>();
@@ -86,6 +90,10 @@ public class CheckpointTupleForwarder implements IRichBolt {
     @Override
     public void declareOutputFields(OutputFieldsDeclarer declarer) {
         bolt.declareOutputFields(declarer);
+        declareCheckpointStream(declarer);
+    }
+
+    protected void declareCheckpointStream(OutputFieldsDeclarer declarer) {
         declarer.declareStream(CHECKPOINT_STREAM_ID, new Fields(CHECKPOINT_FIELD_TXID, CHECKPOINT_FIELD_ACTION));
     }
 
--- a/storm-core/src/jvm/org/apache/storm/topology/CheckpointTupleForwarder.java
+++ b/storm-core/src/jvm/org/apache/storm/topology/CheckpointTupleForwarder.java
@@ -45,41 +45,18 @@ import static org.apache.storm.spout.CheckpointSpout.*;
  * can flow through the entire topology DAG.
  * </p>
  */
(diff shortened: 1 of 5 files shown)


harshach:
+1. @arunmahadevan we want this in 1.x-branch as well right?

arunmahadevan:
@harshach yes both master and 1.x branch

storm-core/src/jvm/org/apache/storm/topology/CheckpointTupleForwarder.java
```diff
@@ -53,6 +53,10 @@
     private long lastTxid = Long.MIN_VALUE;
     private AnchoringOutputCollector collector;
 
+    public CheckpointTupleForwarder() {
```
HeartSaVioR: Do we need this constructor? Especially it will incur NPE since this.bolt is accessed from various methods.

HeartSaVioR: Oh I'm seeing StatefulBoltExecutor extends this class.

IMO StatefulBoltExecutor extending CheckpointTupleForwarder could make us confusing, since StatefulBoltExecutor always call default constructor of CheckpointTupleForwarder, which means that StatefulBoltExecutor should hide all methods which access (directly or indirectly) this.bolt to avoid NPE.

Could we introduce base class extracting common logic from CheckpointTupleForwarder?

storm-core/src/jvm/org/apache/storm/topology/IStatefulBolt.java
```diff
@@ -27,5 +32,17 @@
  * state updates. The stateful bolts are expected to anchor the tuples while emitting
  * and ack the input tuples once its processed.</p>
  */
-public interface IStatefulBolt<T extends State> extends IStatefulComponent<T>, IRichBolt {
+public interface IStatefulBolt<T extends State> extends IStatefulComponent<T> {
```
HeartSaVioR: Good fix. :)

storm-core/src/jvm/org/apache/storm/topology/CheckpointTupleForwarder.java
arunmahadevan: Since the IStatefulBolt is no longer an IRichBolt it cannot be passed to the base class. Makes sense to move the common functionality to a base class which both StatefulBoltExecutor and CheckpointTupleForwarder inherit to avoid the confusion. Will do the refactoring.

arunmahadevan:
@HeartSaVioR refactored and extracted common logic to a common parent of StatefulBoltExecutor and CheckpointTupleForwarder. Please review.

storm-core/src/jvm/org/apache/storm/topology/StatefulBoltExecutor.java
```diff
@@ -74,6 +71,23 @@ void prepare(Map stormConf, TopologyContext context, OutputCollector collector,
...
 
     @Override
+    public void cleanup() {
+        bolt.cleanup();
+    }
+
+    @Override
+    public void declareOutputFields(OutputFieldsDeclarer declarer) {
+        bolt.declareOutputFields(declarer);
+        super.declareCheckpointStream(declarer);
```
HeartSaVioR: Nit: It doesn't need to be called with `super.`.

HeartSaVioR:
+1 with nit comment which could be skipped.

HeartSaVioR:
+1

PR #1346, opened 2016-04-17
[Storm 1717] Support Redis INRC , INRCBY and INRCBYFLOAT

darionyaphet:
this pull request is add to support redis incr incrby and incrbyfloat . :)


HeartSaVioR:
Thanks for the update.
Btw, I intended to not add incr* cause Storm provides 'at least once' so they makes flawed result when tuples are replayed.

How about adding GEO commands? It's planned to being introduced as Redis 3.2.0 (currently RC3).
Jedis already supports GEO commands on 2.8.0 and latest releases for Jedis is 2.8.1.

darionyaphet:
Yes `GEO` is a new feature in `Redis 3.2` and I open a new ticket [Support GEO in storm-redis](https://issues.apache.org/jira/browse/STORM-1720)

darionyaphet:
@HeartSaVioR  When streaming replay , the `INCR` may work duplicate . But I think user should use checkpoint to save streaming processing status and restart from the checkpoint after the topology shutdown .

HeartSaVioR:
@darionyaphet 
Unfortunately, Storm doesn't guarantee exactly-once unless you use Trident with appropriate Spout and State. Checkpointing doesn't guarantee exactly-once.
http://storm.apache.org/releases/1.0.0/State-checkpointing.html
http://storm.apache.org/releases/1.0.0/Trident-state.html

darionyaphet:
@HeartSaVioR understood ! So `Redis Bolt` should be idempotency .

HeartSaVioR:
@darionyaphet 
Non-idempotent implementation of bolt should be used with caution, so instead of making any confusion or error-prone, I'd love for users to create their own bolt via extending AbstractRedisBolt when they want to use it with risk.

PR #1352, opened 2016-04-21
STORM-1723 (1.x) Introduce ClusterMetricsConsumer

HeartSaVioR:
* ClusterMetricsConsumer publishes cluster-side related metrics into consumers
  * like MetricsConsumer for topology metrics
* Users can implement IClusterMetricsConsumer and configure to cluster conf. file to take effect
  * Please refer conf/storm.yaml.example for more details on configuring
  * Nimbus should be launched with additional jars which are needed for IClusterMetricsConsumer
* Also did some refactor to nimbus.clj

I don't address this for master yet since ui/core.clj is not converted to Java but stats.clj is converted to Java. 
If we want to move some methods in this patch to stats.clj, I'll follow up reviews, and start working on crafting patch against master when patch is ready to merge.


HeartSaVioR:
Github denies to show the changes of nimbus.clj. ;(
Attached diff file. (Github also denies to upload file with 'diff' extension.)
[STORM-1723-1.x-v1-diff.txt](https://github.com/apache/storm/files/229830/STORM-1723-1.x-v1-diff.txt)

storm-core/src/jvm/org/apache/storm/metric/ClusterConsumerExecutor.java
```diff
@@ -0,0 +1,87 @@
...
+    public static final Logger LOG = LoggerFactory.getLogger(ClusterConsumerExecutor.class);
+
+    private IClusterMetricsConsumer metricsConsumer;
+    private String consumerClassName;
+    private Object registerationArgument;
+    private ExecutorService consumerExecutorService;
+
+    public ClusterConsumerExecutor(String consumerClassName, Object registerationArgument) {
+        this.consumerClassName = consumerClassName;
+        this.registerationArgument = registerationArgument;
```
satishd: typo: `registerationArgument ` should be `registrationArgument `

storm-core/src/jvm/org/apache/storm/metric/ClusterConsumerExecutor.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    public void prepare() {
+        try {
+            metricsConsumer = (IClusterMetricsConsumer)Class.forName(consumerClassName).newInstance();
```
satishd: Registered `consumerClassName` should be available as part of the same classloader as `ClusterConsumerExecutor `. Is this a valid assumption? So, user should package them accordingly.

storm-core/src/jvm/org/apache/storm/metric/ClusterConsumerExecutor.java
HeartSaVioR: Nice finding. I'll address it.

storm-core/src/jvm/org/apache/storm/metric/ClusterConsumerExecutor.java
HeartSaVioR: Yes right.
Users are encouraged to package their consumer as a fat jar and place to system environment STORM_EXT_CLASSPATH_DAEMON or /extlib-daemon.
Placing jar to system environment STORM_EXT_CLASSPATH or /extlib also works but it will affect non-daemon (including worker) so former would be better way.

storm-core/src/jvm/org/apache/storm/metric/ClusterConsumerExecutor.java
HeartSaVioR: Thanks, addressed.

unsleepy22:
I couldn't comment on nimbus.clj, how about rename "cluster-consumer-executors" to "cluster-metrics-consumers" since we have "executors" in tasks already.

unsleepy22:
for the ClusterInfo class, do we need to add a "cluster.name" property? In a real metrics consumer case, the external storage may want to consumer a bunch of cluster metrics and display them, thus a "cluster.name" may be necessary to distinguish different clusters.

HeartSaVioR:
@unsleepy22 
Oh there's a missing spot, ClusterConsumerExecutor should be renamed to ClusterMetricsConsumerExecutor. 
I would like to replace 'executor' from that name but I couldn't find better name due to short English. If we're OK to only rename 'ClusterConsumerExecutor' to 'ClusterMetricsConsumerExecutor', I think renaming 'cluster-consumer-executors' to 'cluster-metrics-consumer-executors' makes sense, too.

HeartSaVioR:
@unsleepy22 
About 'cluster.id', I agree what you considered, but also we need to consider that Storm itself is not aware of multiple clusters. So IMO it would be not better to name cluster by Storm side.

But we can open possibility to provide option to name cluster id (`storm.cluster.id`) from yaml configuration. In fact setting cluster id to UUID automatically doesn't help much since users can't recognize its id.

What do you think?

HeartSaVioR:
Also addressed some missed spots.

unsleepy22:
Maybe it's better to put multi-cluster support in 2.0, in which I think multi-cluster can be supported at least in web ui. So we can leave it be for now if we don't see the need from end users.

HeartSaVioR:
Updating diff files again.

[STORM-1723-1.x-v2-diff.txt](https://github.com/apache/storm/files/233825/STORM-1723-1.x-v2-diff.txt)

unsleepy22:
+1

conf/defaults.yaml
```diff
@@ -285,3 +285,6 @@ pacemaker.kerberos.users: []
 #default storm daemon metrics reporter plugins
 storm.daemon.metrics.reporter.plugins:
      - "org.apache.storm.daemon.metrics.reporters.JmxPreparableReporter"
```
abhishekagarwal87: can't we piggyback on codahale to report these metrics? We can have one more config
```storm.cluster.metrics.reporter.plugins:
   - "org.apache.storm.daemon.metrics.reporters.JmxPreparableReporter"
```
I haven't gone through diff in nimbus completely but I suppose they can be modeled as gauages. It would be lot more simplified in my opinion.

HeartSaVioR: Please refer this file:
https://github.com/apache/storm/blob/master/docs/storm-metrics-profiling-internal-actions.md
(btw, link for http://storm.apache.org/releases/1.0.0/storm-metrics-profiling-internal-actions.html is broken, and so 2.0.0-SNAPSHOT is.)

Codahale metrics is included as shaded so we don't guarantee other plugins are working properly.

We want to expose those metrics to consumer to store external services/storages, and where to store should be flexible by make it pluggable.

abhishekagarwal87: Ahh I see.

storm-core/src/clj/org/apache/storm/daemon/nimbus.clj
```diff
@@ -1355,48 +1366,843 @@
 (defmethod blob-sync :local [conf nimbus]
   nil)
 
-(defserverfn service-handler [conf inimbus]
-  (.prepare inimbus conf (master-inimbus-dir conf))
-  (log-message "Starting Nimbus with conf " conf)
-  (let [nimbus (nimbus-data conf inimbus)
-        blob-store (:blob-store nimbus)
-        principal-to-local (AuthUtils/GetPrincipalToLocalPlugin conf)
+(defn extract-cluster-metrics [^ClusterSummary summ]
```
harshach: since we are doing this in ui isn't it better move it to helpers or utils and import that here in nimbus

storm-core/src/clj/org/apache/storm/daemon/nimbus.clj
```diff
@@ -1355,48 +1366,843 @@
...
-              {:storm-name storm-name
-               :storm-cluster-state storm-cluster-state
-               :all-components all-components
-               :launch-time-secs launch-time-secs
-               :assignment assignment
-               :beats beats
-               :topology topology
-               :task->component task->component
-               :base base}))
+        (fn [^String storm-id operation]
```
harshach: it will be great if we can preserve previous spaces looks like this diff showing up unnecessarily

harshach:
+1 . If possible we should move extracting cluster summary to a common file and use that in ui and nimbus

hmcl:
+1. I am also in favor of making cluster summary modular.

storm-core/src/clj/org/apache/storm/daemon/nimbus.clj
HeartSaVioR: Oh that's fix for whitespaces but I agree with you. I'll address it.

storm-core/src/clj/org/apache/storm/daemon/nimbus.clj
HeartSaVioR: good point. I'll address it.

HeartSaVioR: Thinking it once more, we're in progress of porting so just referring unported thing is better than refactor.
I'll make this method to refer ui function to get rid of duplication.

HeartSaVioR:
@harshach @hmcl Addressed review comments, and rebased with commits squashed.
I feel applying this to 1.0.x seems to be against semantic versioning, so I would not address it.
Do we want to apply this to master?

HeartSaVioR:
Never mind. Pull request for master is here: #1460.

PR #1353, opened 2016-04-22
storm-1726: use Put#addColumn to replace the deprecated Put#add

lujinhong:
Besides that, any plan to use Table to replace HTable, and use BufferMutation to cache ?

### DIFF
--- a/external/storm-hbase/src/main/java/org/apache/storm/hbase/common/HBaseClient.java
+++ b/external/storm-hbase/src/main/java/org/apache/storm/hbase/common/HBaseClient.java
@@ -58,14 +58,14 @@ public class HBaseClient {
             put.setDurability(durability);
             for (ColumnList.Column col : cols.getColumns()) {
                 if (col.getTs() > 0) {
-                    put.add(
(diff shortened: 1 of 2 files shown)


Parth-Brahmbhatt:
@lujinhong Can you file a jira to make the move. I think I have a patch lying around in my repo which brings some perf improvements in addition to moving the code to newer APIs.

lujinhong:
https://issues.apache.org/jira/browse/STORM-1726?filter=12335179

HeartSaVioR:
+1
@Parth-Brahmbhatt Would you have a plan to submit your patch for the improvement?

HeartSaVioR:
@lujinhong Would you also craft pull request for 1.x-branch? I can also cherry-pick so please submit if you really don't mind.

HeartSaVioR:
@lujinhong Never mind. It would be better to cherry-pick since patch is really tiny.

HeartSaVioR:
cc. @ptgoetz @harshach Could you review this as well? Just want to check we're OK to drop support HBase 0.9x.

PR #1354, opened 2016-04-22
STORM-1724 Fill up lacking contents to Metrics documentation

HeartSaVioR:
Please keep in mind while reviewing that this is written by not native and good writing skill for English.
So you may want to correct many places which are wrong or weird use of vocabulary, grammar, nuance, and so on.

### DIFF
--- a/docs/Metrics.md
+++ b/docs/Metrics.md
@@ -8,7 +8,7 @@ It's used internally to track the numbers you see in the Nimbus UI console: coun
 
 ### Metric Types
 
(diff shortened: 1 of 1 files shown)


HeartSaVioR:
And please note that this doesn't represent current porting state (so 1.x and master could be same.)  I guess we should clear up documentation when porting is done.

HeartSaVioR:
Could anyone take a look?

harshach:
+1. This looks great.

HeartSaVioR:
@harshach Thanks for reviewing! I'll merge and reflect the website.

PR #1357, opened 2016-04-22
STORM-1725: Kafka Spout New Consumer API - KafkaSpoutRetryExponential Backoff method should use HashMap instead of TreeMap not to throw Exception

hmcl:
- Replace TreeMap with HashMap


hmcl:
This issue was found by @jianbzhou and credit should be given to him (for discussion details see https://github.com/apache/storm/pull/1131#issuecomment-213211034)

harshach:
+1

PR #1358, opened 2016-04-23
STORM-1727: add storm 1.0.0 rename note

vesense:
I think it is **important** to add the NOTE in the 1.0.0 release docs to let more storm users know this change.
The content is from Abhishek. Thanks @abhishekagarwal87 .
Preview:
preview

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -3,6 +3,19 @@ title: Documentation
 layout: documentation
 documentation: true
 ---
+
+
+> #### NOTE
+
+> In the latest version, the class packages have been changed from "backtype.storm" to "org.apache.storm" so the topology code compiled with older version won't run on the Storm 1.0.0 just like that. Backward compatibility is available through following configuration 
+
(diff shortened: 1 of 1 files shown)


HeartSaVioR:
+1

Btw, I think we would be better to consistently add contents I stated from dev@ to 'release news' for some releases.
We should notify users before installing Apache Storm 1.x, so that users don't encounter unknown issue in point of their view after installing. And I expect users to read the release note before install new version.

vesense:
@HeartSaVioR Totally agree with you. But I don't know how to modify the docs(like release news) in the asf site SVN.

HeartSaVioR:
@vesense We can just initiate discussion with dev@ list and one of committer can modify it when decision is made.

abhishekagarwal87:
Agree with @HeartSaVioR - This should be added to the 1.0 release announcement page.

ptgoetz:
+1

vesense:
@ptgoetz Seems like this change should be published to the storm web site.

PR #1361, opened 2016-04-26
STORM-1729 (1.x) Get rid of reflections while recording stats

HeartSaVioR:
* define MultiCount/LatencyStatAndMetric in let statement and set type hint to there

Before applying I got around 860,000 emitted/sec from BasicTopology, and after applying I got around 940,000~960,000 emitted/sec.
`topology.disruptor.batch.size=10000`

And it removes reflection warnings from lines which are recording stats.

Btw, though after applying this, `invokeMatchingMethod` is still on top 10. I'll track where it occurs.

### DIFF
--- a/storm-core/src/clj/org/apache/storm/stats.clj
+++ b/storm-core/src/clj/org/apache/storm/stats.clj
@@ -116,11 +116,13 @@
 
 (defn emitted-tuple!
   [stats stream]
-  (.incBy ^MultiCountStatAndMetric (stats-emitted stats) ^Object stream ^long (stats-rate stats)))
+  (let [^MultiCountStatAndMetric emitted (stats-emitted stats)]
+    (.incBy emitted ^Object stream ^long (stats-rate stats))))
 
 (defn transferred-tuples!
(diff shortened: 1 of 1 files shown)


unsleepy22:
+1
BTW. can we have more performance tests in more cases so that we can see the overall performance change?

HeartSaVioR:
@unsleepy22 
Yeah, I really want to, but in order to do that we need a stable machines and benchmark tests.
We can give it a try with https://github.com/yahoo/streaming-benchmarks first.

HeartSaVioR:
And also ThroughputVsLatency topology in storm-starter.

HeartSaVioR:
@unsleepy22 Let's move to #1362. I'll address the performance tests there.

HeartSaVioR:
Please refer #816. I'm confused completely why this happens cause those lines are introduced at 16c33efac39edea7e2d77df319d0ee3fd407383a which fixes the reflection issue. Why it is broken again?

ptgoetz:
+1 (forgot to do so before merging)

PR #1362, opened 2016-04-26
STORM-1731 (1.x) Avoid looking up debug / backpressure enable flags within critical path

HeartSaVioR:
This pull request is on top of STORM-1729 (#1361) to avoid merge conflicts.

Even though we set topology.debug to false, we actually hit performance decrement at lookup flag.
Just emitting one tuple eventually hits multiple places on lookup which are all critical paths.

Let me summary the performance improvements with BasicTopology:

Patch | emitted/sec (around)
-------- | ----------------------------
STORM 1.0.1-SNAPSHOT | 860,000
STORM-1729 | 940,000~960,000
| STORM-1731 (including STORM-1729) | 1,500,000 ~ 1,600,000 |

Nearly 2x improvement observed from the start.

I measured the number with my dev. machine so numbers could be a bit unstable.
I really appreciated if someone takes a benchmark with stable environment and stable way and posts the number.

PR #1364, opened 2016-04-27
STORM-1731 (0.10.x) Avoid looking up debug / backpressure enable flags within critical path

HeartSaVioR:
* preload the value of flags and use that value
  * topology.debug

Pull request for 1.x version is here: #1362 

Please note that this pull request is NOT on top of STORM-1729 since STORM-1729 is not valid for 0.10.x.
I don't take performance tests on this pull request, since #1362 shows the benefit.


harshach:
+1

vesense:
+1 Great Finding.

PR #1367, opened 2016-04-27
STORM-1739 (1.x) update the mini JAVA version dependency in 0.10.0 and above.

ziyunhx:
Storm drops supporting JDK 1.6 in Apache Storm 0.10.0 and above.


ptgoetz:
+1

PR #1369, opened 2016-04-27
STORM-1735: Nimbus should log that replication succeeded when min rep…

srdo:
…licas was reached exactly


HeartSaVioR:
+1 Nice finding.
@srdo I could cherry-pick into 1.x-branch since changeset is really small but then merging should wait to release 1.0.1. Could you craft pull request against 1.x-branch if you really don't mind?

srdo:
@HeartSaVioR Sure, please do :)

PR #1360, opened 2016-04-25
STORM-1733 Flush stdout and stderr before calling "os.execvp" to prevent log loss.

dsKarthick:
bin/storm.py emits the following crucial information that is lost because we don't flush the stdout before exec. 
```
2016-04-25T08:23:43.17141 Running: java -server -Dstorm.options= -Dstorm.home= -Xmx1024m -Dlogfile.name=nimbus.log -Dlogback.configurationFile=logback/cluster.xml  backtype.storm.ui.core.nimbus
```

Observed Environment:
```
OS: CentOS release 6.5
Kernel: 2.6.32-431.el6.x86_64
Python version: Python 2.7.2
```

PR #1371, opened 2016-04-27
STORM-1733 (0.9.x) Flush stdout and stderr before calling "os.execvp" to prevent log loss.

dsKarthick:
Please see https://github.com/apache/storm/pull/1360

@arunmahadevan @vesense


dsKarthick:
Invalid. Please ignore.

PR #1376, opened 2016-04-28
STORM-1736. Change KafkaTestBroker.buildKafkaConfig to new KafkaConfig api.

harshach:


ptgoetz:
The patch looks okay to me, but I would like to know why it's necessary. I can't reproduce the issuue from the branch or rc2 build.

HeartSaVioR:
I also don't encounter this through 1.0.1 RC1 to 1.0.1 RC3. Did you set the custom Kafka version while building?

abhishekagarwal87:
This would also break the compatibility of storm 1.0.0 with 0.8.x kafka client API.

harshach:
@HeartSaVioR @abhishekagarwal87 this issues happens with Kafka 0.10 branch. 
Yes it does break test compatability with 0.8.x . Keep in mind that KafkaConfig is an internal kafka class that we use to start a in-process kafka broker. We shouldn't be calling it as breaking comptability with client api. Also this change will work with Kafka 0.9.x on wards

abhishekagarwal87:
@harshach - with this change, test compilation will fail once the #1386 is merged. Since this is the internal class used only in test, do we really need to make it compatible with 0.10 branch?

harshach:
@abhishekagarwal87 commented on your PR https://github.com/apache/storm/pull/1386 .

PR #1377, opened 2016-04-28
STORM-1741: remove unconditional setting of JAVA_HOME from storm-env.sh

ptgoetz:


conf/storm-env.sh
```diff
@@ -19,6 +19,6 @@
 # Set Storm specific environment variables here.
 
 # The java implementation to use.
-export JAVA_HOME=${JAVA_HOME}
+#export JAVA_HOME=${JAVA_HOME}
```
erikdw: Maybe leave a comment about why dead code is being left here?

ptgoetz:
As this is a very minor change that fixes a blocker, if there are no objections I will merge it in order to proceed with the 1.0.1 rc3 release.

If there are any objections after the fact, we can cancel the release and revert the change.

erikdw:
Merge away!  Thanks for the update.

abhishekagarwal87:
+1

PR #1378, opened 2016-04-29
add convenience methods for checking tuple type in node.js library

daluu:
Convenience methods for checking if tuple is tick or heartbeat tuple to match python version, also fix some comment typos.


HeartSaVioR:
+1
Could you craft pull requests for 1.x-branch and 0.10.x-branch, too?

PR #1380, opened 2016-04-29
STORM-1745: Add partition to log output in PartitionManager

srdo:
Adding topic name and partition id to the log output makes it easier to debug when an issue arises.


HeartSaVioR:
@srdo 
Could you give an example of actual changes in logs?

srdo:
@HeartSaVioR Sure. The current logs look like this "Starting Kafka 10.4.5.1:0 from offset 15520", which doesn't tell you which topic it's affecting. The logs in L259 and L269 don't tell you which partition/topic they're affecting at all. This PR changes the log to use the toString on Partition for logging, which looks like this: "Starting Kafka 10.4.5.3 Partition{host=10.4.5.3:9092, topic=dce-data, partition=26} from offset 1612424"

HeartSaVioR:
@srdo Okay, thanks for explaining. +1

PR #1381, opened 2016-04-29
STORM-1746: Downgrade NOT_LEADER_FOR_PARTITION log to warn from error

srdo:
This error is transient most of the time and will occur during normal operation. Downgrading it to warning makes the error log much less noisy.

### DIFF
--- a/external/storm-kafka/src/jvm/org/apache/storm/kafka/KafkaUtils.java
+++ b/external/storm-kafka/src/jvm/org/apache/storm/kafka/KafkaUtils.java
@@ -213,11 +213,6 @@ public class KafkaUtils {
                 throw new TopicOffsetOutOfRangeException(msg);
             } else {
                 String message = "Error fetching data from [" + partition + "] for topic [" + topic + "]: [" + error + "]";
-                if (error.equals(KafkaError.NOT_LEADER_FOR_PARTITION)) {
-                    LOG.warn(message);
-                } else {
-                    LOG.error(message);
-                }
                 throw new FailedFetchException(message);
             }
         } else {
--- a/external/storm-kafka/src/jvm/org/apache/storm/kafka/KafkaUtils.java
+++ b/external/storm-kafka/src/jvm/org/apache/storm/kafka/KafkaUtils.java
@@ -183,7 +183,7 @@ public class KafkaUtils {
     }
 
     public static ByteBufferMessageSet fetchMessages(KafkaConfig config, SimpleConsumer consumer, Partition partition, long offset)
-            throws TopicOffsetOutOfRangeException, FailedFetchException,RuntimeException {
+            throws TopicOffsetOutOfRangeException, FailedFetchException, RuntimeException {
         ByteBufferMessageSet msgs = null;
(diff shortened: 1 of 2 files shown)


external/storm-kafka/src/jvm/org/apache/storm/kafka/KafkaUtils.java
```diff
@@ -213,7 +213,11 @@ public static ByteBufferMessageSet fetchMessages(KafkaConfig config, SimpleConsu
                 throw new TopicOffsetOutOfRangeException(msg);
             } else {
                 String message = "Error fetching data from [" + partition + "] for topic [" + topic + "]: [" + error + "]";
-                LOG.error(message);
+                if (error.equals(KafkaError.NOT_LEADER_FOR_PARTITION)) {
```
abhishekagarwal87: Worker gets killed due to this error. It should be logged as error only.

srdo: It should get caught here https://github.com/apache/storm/blob/master/external/storm-kafka/src/jvm/org/apache/storm/kafka/KafkaSpout.java#L142 or here https://github.com/apache/storm/blob/master/external/storm-kafka/src/jvm/org/apache/storm/kafka/trident/TridentKafkaEmitter.java#L78

abhishekagarwal87: I see. Then we can remove this log itself since the same message is being logged again at https://github.com/apache/storm/blob/master/external/storm-kafka/src/jvm/org/apache/storm/kafka/KafkaSpout.java#L143. 
TridentKafkaSpout is swallowing up the exception. You can add the exception in the log there.

srdo:
This has the effect of downgrading all errors from fetch to warning level. Maybe the error log in KafkaUtils should be added back with a check for whether it's NOT_LEADER_FOR_PARTITION?

srdo:
Looks like a flaky test. Going to try again.

external/storm-kafka/src/jvm/org/apache/storm/kafka/KafkaUtils.java
```diff
@@ -213,7 +213,9 @@ public static ByteBufferMessageSet fetchMessages(KafkaConfig config, SimpleConsu
                 throw new TopicOffsetOutOfRangeException(msg);
             } else {
                 String message = "Error fetching data from [" + partition + "] for topic [" + topic + "]: [" + error + "]";
-                LOG.error(message);
+                if (!error.equals(KafkaError.NOT_LEADER_FOR_PARTITION)) {
```
abhishekagarwal87: I would still suggest to remove this log completely since we are refreshing for all such errors and a warning with the exception should suffice.

srdo: Okay, will do.

srdo:
Closing since it seems like this is a fairly irrelevant change to what is now legacy code.

PR #1384, opened 2016-04-30
STORM-1750: Ensure worker dies when report-error-and-die is called. M…

srdo:
…ake ZkStateStorage set_data try setting data if node creation fails because the node exists

Similar changes probably need to be made to 0.10.x and 1.x to prevent executors from disappearing until the worker is manually rebooted. The change to ZkStateStorage may not be strictly necessary to fix this issue, but it should reduce the number of exceptions thrown out of set_data due to some component creating a node after another has passed the exists check.

### DIFF
--- a/storm-core/src/jvm/org/apache/storm/cluster/ZKStateStorage.java
+++ b/storm-core/src/jvm/org/apache/storm/cluster/ZKStateStorage.java
@@ -28,6 +28,7 @@ import org.apache.storm.utils.Utils;
 import org.apache.storm.zookeeper.Zookeeper;
 import org.apache.zookeeper.CreateMode;
 import org.apache.zookeeper.KeeperException;
+import org.apache.zookeeper.KeeperException.NodeExistsException;
 import org.apache.zookeeper.Watcher;
 import org.apache.zookeeper.data.ACL;
 import org.slf4j.Logger;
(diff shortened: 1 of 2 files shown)


HeartSaVioR:
+1 
@srdo Nice finding. Could you address this to 1.x-branch and 0.10.x-branch, too? Thanks in advance!

srdo:
Hi @HeartSaVioR,
Sure, I'll make another pair of PRs soon.

storm-core/src/jvm/org/apache/storm/cluster/ZKStateStorage.java
```diff
@@ -189,7 +190,15 @@ public void set_data(String path, byte[] data, List<ACL> acls) {
             Zookeeper.setData(zkWriter, path, data);
         } else {
             Zookeeper.mkdirs(zkWriter, Zookeeper.parentPath(path), acls);
-            Zookeeper.createNode(zkWriter, path, data, CreateMode.PERSISTENT, acls);
+            try {
+                Zookeeper.createNode(zkWriter, path, data, CreateMode.PERSISTENT, acls);
+            } catch (RuntimeException e) {
```
Jackyele: Why not catch NodeExistsException directly

srdo: Not sure how? createNode() wraps it in RuntimeException, so it has to be unwrapped first, right? https://github.com/apache/storm/blob/master/storm-core/src/jvm/org/apache/storm/zookeeper/Zookeeper.java#L129

Jackyele: yeah, we have to catch RuntimeError first; and how about using **Utils. exceptionCauseIsInstanceOf** instead for consistency, just like https://github.com/apache/storm/blob/master/storm-core/src/jvm/org/apache/storm/zookeeper/Zookeeper.java#L165 does

srdo: Good idea, I hadn't noticed that function :)

HeartSaVioR:
+1 again

PR #1387, opened 2016-05-02
STORM-1756: Explicitly null KafkaServer reference in KafkaTestBroker …

srdo:
…to prevent out of memory on large test classes.

I'm not clear on whether JUnit keeps the reference for the lifetime of the module test, or whether they're released after all tests in a class have run. Either way, a test class with a KafkaTestBroker field and many tests would likely still run out of memory.

### DIFF
--- a/external/storm-kafka/src/test/org/apache/storm/kafka/KafkaTestBroker.java
+++ b/external/storm-kafka/src/test/org/apache/storm/kafka/KafkaTestBroker.java
@@ -78,7 +78,12 @@ public class KafkaTestBroker {
         return port;
     }
     public void shutdown() {
-        kafka.shutdown();
+        if (kafka != null) {
+            kafka.shutdown();
+            kafka.awaitShutdown();
+        }
+        //Ensure kafka server is eligible for garbage collection immediately
+        kafka = null;
         if (zookeeper.getState().equals(CuratorFrameworkState.STARTED)) {
             zookeeper.close();
         }


HeartSaVioR:
Hi, @srdo 
Since I didn't see any OOMEs, I'm curious that it is based on assumption or you could provide scenario to reproduce.

srdo:
@HeartSaVioR I saw OOME on one of my branches, because each KafkaServerStartable allocates a 130MB byte array for the log cleaner thread. A heap dump showed that there were 8 of these in memory at the time of the exception. I was going off of http://junit.org/junit4/faq.html#atests_18 to explain it, but I can't reproduce it now. I'll close this.

srdo:
If this turns out to be a problem later, disabling the log cleaner thread during tests is probably a better fix. The tests don't make logs long enough to be compacted anyway.

HeartSaVioR:
Please reopen this issue if you saw intermittent test failure. What I asked is that it's a theory or it occurred some issues.

srdo:
I did see OOME due to KafkaServer resources. 8x 130MB byte arrays caused the VM to run out of memory. As I said, I can't reproduce it. It's probably not an issue on master when storm-kafka gets downgraded to 0.8.2.1 (here https://github.com/apache/storm/pull/1386), since one of the changes in Kafka 0.9.0.1 was to enable the log cleaner thread by default (and thus allocate some very large byte arrays when starting a server).

HeartSaVioR:
+1. It doesn't hurt though we downgrade kafka version for storm-kafka to 0.8.x.

PR #1389, opened 2016-05-03
STORM-1750 (1.x): Ensure worker dies when report-error-and-die is called. M…

srdo:
…ake zookeeper_state_factory set-data try setting data if node creation fails because the node exists.

Backport of STORM-1750


HeartSaVioR:
+1

PR #1309, opened 2016-04-04
STORM-1680, Added Kafka Spout Config FetchByte to Storm-Kafka

supermonk:
Added Kafka Spout Config FetchByte to Storm-Kafka  https://issues.apache.org/jira/browse/STORM-1680

PR #1391, opened 2016-05-03
(STORM-1674) Idle KafkaSpout consumes more bandwidth than needed

moesol:
- Allows minBytes in fetch request to be configured
- Defaults new configuration KafkaConfig.fetchMinBytes to 1.
- Defaults fetchMaxWait to 100ms instead of 10000ms.

1) minBytes was 1
2) maxWait was 100

Hopefully the risk of this change is low because
the old behavior can be restored using the API by setting
KafkaConfig.fetchMaxWait to 10000
KafkaConfig.fetchMinBytes to 0

### DIFF
--- a/external/storm-kafka/src/jvm/storm/kafka/KafkaConfig.java
+++ b/external/storm-kafka/src/jvm/storm/kafka/KafkaConfig.java
@@ -30,7 +30,8 @@ public class KafkaConfig implements Serializable {
 
     public int fetchSizeBytes = 1024 * 1024;
     public int socketTimeoutMs = 10000;
-    public int fetchMaxWait = 10000;
+    public int fetchMaxWait = 100;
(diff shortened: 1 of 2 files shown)


harshach:
@moesol there is patch that merged into 1.0 release https://github.com/apache/storm/pull/1309/files . Does that changes work for you. Any reason for the MaxWait to be at 100

moesol:
Without the MaxWait being set to 100, it is set to 10000, which means that the bolt will waiting for 10 seconds when no messages are available. On the 1.x branch, I was going to use:

+    public int fetchMaxWait = FetchRequest.DefaultMaxWait();
+    public int fetchMinBytes = FetchRequest.DefaultMinBytes();

to get Kafka default behavior of 100/1.

PR #1393, opened 2016-05-03
(STORM-1674) Idle KafkaSpout consumes more bandwidth than needed

moesol:
* Allows minBytes in fetch request to be configured
    * Defaults new configuration KafkaConfig.fetchMinBytes to 1.
    * Defaults fetchMaxWait to 100ms instead of 10000ms.

    1) minBytes was 1
    2) maxWait was 100


harshach:
+1

ptgoetz:
+1 Nice catch!

ptgoetz:
I'm in the process of merging this. Other committers please hold off. More info to follow...

ptgoetz:
Thanks @moesol I merged this into master, 1.x-branch and 1.0.x-branch.

Note that with 1.0.x-branch this created a conflict, as `KafkaConfig.minFetchByte` had been introduced to do the same thing as your `KafkaConfig.fetchMinBytes` variable. These are public variables, so they make up part of the API.

Since 1.0.1 had already been released with the `minFetchByte` varibable, to ensure backward compatibility I used the `minFetchByte` variable name across branches.

While I prefer the variable name you chose, I think it's more important to ensure backward compatibility. :)

Thanks again!

moesol:
I think it's awesome that this change is getting in! I'll be looking out for the next release that contains this PR.<br> 

Robert Hastings, Lab SDL: 619-553-0826, Cell: 858-395-4436, Moebius: 858-455-5431x210 
SIPR: hasting@spawar.navy.smil.mil 
JWICS: hasting@spawar.navy.ic.gov

PR #1394, opened 2016-05-03
STORM-1761: Storm-Solr Example Throws ArrayIndexOutOfBoundsException in Remote Cluster Mode

hmcl:
Please cherry-pick this patch into 0.10.x-branch and 1.x-branch as well.


HeartSaVioR:
+1 Nice finding.
Btw, I guess it's due to inconsistencies between sample / test topologies, one argument for only specifying topology name, or two arguments which one is "remote" and next one is topology name.
I don't have strong opinion but it would be better to have unified way to run.

hustfxj:
+1

PR #1396, opened 2016-05-04
[storm-elasticsearch]Upgrade elasticsearch version from 1.6.0 to 2.3.2

kojiisd:


sweetest:
I was also doing similar work, code change is almost the same. Looks good to me!

dossett:
I have a similar PR (https://github.com/apache/storm/pull/1337) that keeps Elastic 1.x and 2.x side by side to support both.  Elastic Search 1.7 was released just a few months ago so I think Storm should continue to support 1.X.  That said, there are big drawbacks to duplicated code, etc. required to support both.

HeartSaVioR:
I don't think we can drop support for 1.x. If ES is not backward compatible between 1.x and 2.x, we have to bear with supporting multiple versions though it requires duplicated code, etc, for some times.

knusbaum:
I'm going to close this. We're not going to drop 1.x support and there's another PR up adding support for 2.x without dropping 1.x.

PR #1419, opened 2016-05-16
STORM-1838 update OffsetEntry to support EARLIEST/LATEST FirstPollOffsetStrategy

flisky:
I have no idea why the comment "leave the acked offsets as they were to resume where it left off" failed, but this patch simply fixes my problem.

### DIFF
--- a/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
+++ b/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
@@ -147,7 +147,9 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
             for (TopicPartition tp : partitions) {
                 final OffsetAndMetadata committedOffset = kafkaConsumer.committed(tp);
                 final long fetchOffset = doSeek(tp, committedOffset);
-                setAcked(tp, fetchOffset);
+                if (!consumerAutoCommitMode) {
+                  acked.put(tp, new OffsetEntry(tp, fetchOffset));
+                }
             }
             initialized = true;
             LOG.debug("Initialization complete");
@@ -179,13 +181,6 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
                 fetchOffset = kafkaConsumer.position(tp);
             }
             return fetchOffset;
-        }
-    }
-
-    private void setAcked(TopicPartition tp, long fetchOffset) {
-        // If this partition was previously assigned to this spout, leave the acked offsets as they were to resume where it left off
-        if (!consumerAutoCommitMode && !acked.containsKey(tp)) {
-            acked.put(tp, new OffsetEntry(tp, fetchOffset));
         }
     }
 


ptgoetz:
testing JIRA integration, please ignore

ptgoetz:
another JIRA integration test. Ignore.

hmcl:
@flisky can you please share the code, or at least the steps to reproduce the scenario where this code does not work? Another alternative is that you enable all the logging levels for us to understand what is happening. That could should be necessary, and I am not very comfortable just removing it without understanding what is going on. 

Without that code, I believe that the scenario where a consumer rebalance occurs, and offsets 1,2 have been hacked, but 3 is still pending, and 4, 5 have also been acked, won't work. That code is there to cover for this scenario.

flisky:
It turns out that I used `FirstPollOffsetStrategy#LATEST`, and `UNCOMMITTED` version would be okay.

hmcl:
@flisky so what does your last comment mean? That this patch no longer applies?

flisky:
Sorry, I didn't say it clearly, but this patch still applies.

Ｗhen a consumer re-balance occurs and **FirstPollOffsetStrategy#EARLIEST/LATEST**,  the retained partition wouldn't use the committed offset to resume but use the -2 / -1 offset, as I observed, so this is reason why the comment (`leave the acked offsets as they were to resume where it left off`) doesn't apply.

flisky:
By the way, it seems that we didn't handle the situation `offset out of range` when I test the `UNCOMMITTED_LATEST` strategy.

hmcl:
@flisky can you please provide specific examples or a code snippet that I can use to test what you are saying. Can you also post the logs (perhaps on JIRA) that show the cases that are failing? What do you mean by `offset out of range`?

As for your first comment, please keep in mind that the `FirstPollOffsetStrategy`, as the name indicates, it's only for the first poll, i.e., the first time a `KafkaSpout` polls from Kafka. It does not apply to consumer rebalance. Quoting the [javadoc](https://goo.gl/PyR7Cm) of "The offset used by the Kafka spout in the first poll to Kafka broker." Therefore, the `LATEST/EARLIEST` strategies only apply in the first poll, not on every consumer rebalance. Furthermore, if a consumer rebalance occurs, but the same partition is assigned to this spout, and some offsets have already been emitted and acked, we wan to keep those, and only process the ones that haven't yet been acked because they have have failed and waiting to be retried (according to exponential backoff) or may be on the "wire". 

What do you mean by -2 , -1? Can you please give specific examples. Thanks.

flisky:
> as the name indicates, it's only for the first poll

As I observed, it isn't. `KafkaSpoutConsumerRebalanceListener#onPartitionsAssigned` called `initialize` method, so FirstPollOffsetStrategy works on every consumer rebalance.

Can you please confirm this? Thanks!

hmcl:
@flisky I am looking at this now

flisky:
friendly ping, @hmcl

hmcl:
@flisky some stuff came up. Apologies for the delay in my response. I will get back to you today or tomorrow as I am working on this now.

PR #1420, opened 2016-05-16
STORM-1841: Address a few minor issues in windowing and doc

arunmahadevan:
1. Do not accept negative values for window length or sliding interval in BaseWindowedBolt
2. Added static factories for Count and Duration for ease of use
3. Explicitly call out the sliding behavior and when the first window is evaluated in the windowing doc.


arunmahadevan:
@harshach can you take a look ? This needs to be applied to 1.x branch as well.

HeartSaVioR:
+1. Nice addressing.

harshach:
+1

PR #1422, opened 2016-05-17
STORM-1845: use UTF-8 instead of default encoding

vesense:


darionyaphet:
Hi  @vesense If I don't specified charset which one will be used ?

vesense:
default: `lookupCharset(System.getProperty("file.encoding")) == null ? "UTF-8" : System.getProperty("file.encoding")`

hmcl:
@vesense what is the motivation behind this change? Prior to your proposed change, the charset used is the one that is defined by the java system property `file.encoding`. If this property is absent  the file encoding used will be `UTF-8` in JDK8. 

My understanding is that your proposed change will eliminate the chance to customize an arbitrary charset using `file.encoding`. Is this the goal?

If the goal is to set `UTF-8` by default, perhaps once should simply set in a configuration file the property `file.encoding=UTF-8`

PR #1428, opened 2016-05-18
STORM-1848: Make KafkaMessageId and Partition serializable to support…

srdo:
… eventlogging

### DIFF
--- a/external/storm-kafka/src/jvm/org/apache/storm/kafka/Partition.java
+++ b/external/storm-kafka/src/jvm/org/apache/storm/kafka/Partition.java
@@ -18,10 +18,11 @@
 package org.apache.storm.kafka;
 
 import com.google.common.base.Objects;
+import java.io.Serializable;
 import org.apache.storm.trident.spout.ISpoutPartition;
 
 
-public class Partition implements ISpoutPartition {
+public class Partition implements ISpoutPartition, Serializable {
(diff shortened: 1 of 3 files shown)


abhishekagarwal87:
+1

HeartSaVioR:
+1
Build failures are due to nexus issue.

srdo:
Does it make sense to restrict the emit functions on SpoutOutputCollector to take only serializable messageIds?

revans2:
@srdo that would not be a backwards compatible change for something that never had to have this before.  And we don't require it to be serializable, if you have it properly registered with kryo then it will handle it instead.

srdo:
You're right, I forgot about backwards compatibility, I mainly wanted to ensure this doesn't pop up in other spouts :) Is there any way for the Kafka spout to register with Kryo instead?

How about a comment on the SpoutOutputCollector interface about this?

hmcl:
+1

revans2:
+1 for the comment. I think that would be great, and having the KafkaSpout register things with kryo would be good, but I don't know if the API is setup to support that, as the KafkaSpout would need to have access to the topology config before submitting the topology.

srdo:
I'll leave it at adding the comments then.

ptgoetz:
+1

PR #1432, opened 2016-05-19
Bump version to 1.1.0-SNAPSHOT

HeartSaVioR:
Made a mistake to fix version only pom.xml on root directory.

For preventing same mistake again, I created this pull request instead of commit and push directly.

@ptgoetz @harshach Could you take a look? I did find & replace "1.0.1-SNAPSHOT" to "1.1.0-SNAPSHOT" against whole directory so I guess I don't miss a spot, but would like to confirm.

### DIFF
--- a/examples/storm-starter/pom.xml
+++ b/examples/storm-starter/pom.xml
@@ -20,7 +20,7 @@
   <parent>
       <artifactId>storm</artifactId>
       <groupId>org.apache.storm</groupId>
-      <version>1.0.2-SNAPSHOT</version>
+      <version>1.1.0-SNAPSHOT</version>
       <relativePath>../../pom.xml</relativePath>
   </parent>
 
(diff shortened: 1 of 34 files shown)


harshach:
+1

dossett:
In other projects I use `mvn versions:set` and `mvn versions:commit` to make changes like this to every pom file.  Simple local experiment looks for the Storm repo.  I'll confirm that and maybe add a note to the developer documentation if that would be helpful.

HeartSaVioR:
Oh thanks @dossett that is great tip for me! If it works like a charm, I believe adding a note is helpful for many people.

kishorvpatil:
+1

ptgoetz:
+1

ptgoetz:
Thanks @HeartSaVioR. Merged.

PR #1423, opened 2016-05-17
STORM-1842 Forward references in storm.thrift cause tooling issues

HeartSaVioR:
Please refer [STORM-1842](https://issues.apache.org/jira/browse/STORM-1842) for more details.

* python doesn't support forward reference so ttypes.py has a problem regarding this issue
* reorder HBPulse, HBRecords, HBNodes to resolve forward reference issue

PR #1434, opened 2016-05-19
Reorder to satisfy types lookup

et1975:
C# thrift compiler fails to process existing order of types


HeartSaVioR:
Hi @et1975 thanks for the contribution. It's filed to STORM-1842 and #1423 is also ready to review.
Could you check #1423 and see it also works for C# thrift compiler?

et1975:
#1423 works, closing.

PR #1435, opened 2016-05-19
[STORM-1851] Fix default nimbus impersonation authorizer config

arunmahadevan:
"nimbus.impersonation.authorizer" is set to "ImpersonationAuthorizer" by default and this causes issues when a user tries to submit topology as a different user in secure mode since the "nimbus.impersonation.acl" configuration is not set by default. Users need to set nimbus.impersonation.acl first before they can submit topology as any user other than "storm" in secure mode.

Removing this config allows users to submit topologies in secure mode as any user by default. Users can set up impersonation by providing both authorizer and the acls via storm.yaml.

### DIFF
--- a/conf/defaults.yaml
+++ b/conf/defaults.yaml
@@ -76,7 +76,6 @@ nimbus.topology.validator: "org.apache.storm.nimbus.DefaultTopologyValidator"
 topology.min.replication.count: 1
 topology.max.replication.wait.time.sec: 60
 nimbus.credential.renewers.freq.secs: 600
-nimbus.impersonation.authorizer: "org.apache.storm.security.auth.authorizer.ImpersonationAuthorizer"
 nimbus.queue.size: 100000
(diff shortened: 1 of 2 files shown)


arunmahadevan:
@harshach can you take a look ?

harshach:
+1

hmcl:
+1
@arunmahadevan perhaps you can update the docs stating what you say in your comment as it will bring awareness to the users on what is going on by default, as well as setup impersonation.

arunmahadevan:
@hmcl the details of setting up impersonation is documented in https://github.com/apache/storm/blob/master/docs/SECURITY.md
Maybe I can explicitly mention the default behavior.

arunmahadevan:
@hmcl updated the docs.

arunmahadevan:
@ptgoetz can you review and merge?

ptgoetz:
+1

PR #1429, opened 2016-05-18
[Storm-1728] TransactionalTridentKafkaSpout error

whoahbot:

PR #1438, opened 2016-05-20
fix for PartitionedTridentSpoutExecutor STORM-1728

lerouxrgd:
UT passed locally for storm-core and storm-kafka.
Also tested in a dev setup (Storm and Kafka running locally).


ptgoetz:
@lerouxrgd Thanks for the patch, but this looks like a duplicate of #1429 which is eligible for merge. Would you mind closing this PR?

The earlier PR was raised when we had Github-JIRA integration issues, which is why it didn't show up as a JIRA comment.

PR #1440, opened 2016-05-20
STORM-1853: Replace ClassLoaderObjectInputStream with ObjectInputStream

abhishekagarwal87:


redsanket:
@abhishekagarwal87 There seems to be a KafkaBoltTest failure, can you take a look at it

abhishekagarwal87:
@redsanket Test failure is unrelated. Also storm-kafka tests pass on my local box.

HeartSaVioR:
+1

I just compared ClassLoaderObjectInputStream and ObjectInputStream, and found ObjectInputStream uses latestUserDefinedLoader() which calls sun.misc.VM.latestUserDefinedLoader(). While I'm not expert on this, but it's not same as ClassLoader.getSystemClassLoader().

If anyone is an expert or having knowledge on classloader please explain this to avoid same issue. Thanks in advance.

rama-nallamilli:
Not sure what the difference is but the impact of using the System Class Loader rather than userDefined means that any locally built sources that are added to the classpath by a build tool cannot be picked up (causing class not found defs).  Currently attempting to build with this patch locally and will see if it resolves the issue, will update.

rama-nallamilli:
Confirmed this patch fixed the issue for us, class loading now behaves correctly for locally built sources.  Out of curiosity - this PR is for master, does that mean the fix will be part of storm 2.0.0 or is it also being reapplied to the storm 1.0.x branch?  Also does anybody have any ideas when the next release cycle is for storm.

abhishekagarwal87:
This will go in 1.0.2 which is planned for a release soon. 
cc @ptgoetz

PR #1441, opened 2016-05-23
STORM-1861 Fix storm script bug to not fork java process

csivaguru:


bin/storm.py
```diff
@@ -241,7 +240,6 @@ def jar(jarfile, klass, *args):
             extrajars=[tmpjar, USER_CONF_DIR, STORM_BIN_DIR],
             args=args,
             daemon=False,
```
hmcl: @revans2 can you please take a look and let us know if there is any special reason for this process to be forked? This process being forked affects the return code. If there is no reason for this process to be forked, this fix will correct the return code. Thanks.

ptgoetz: @hmcl I don't see a reason to fork the submission process.

I'm +1 for the patch.

hmcl: +1

revans2: The goal was so that the code after this call would execute.
```
os.remove(tmpjar)
```

Without that we leak a file.

ptgoetz: Ah... that makes sense now. So how concerned are we about leaking a tmp file vs. allowing users to build tools around topology submission (and get the right exit code)?

Would it make sense to have a separate command to do the topology transform instead of in-lining it with topology submission? I.e. two steps: 1. transform topology, 2. submit topology jar output from step 1.

revans2: This was supposed to make it transparent to the end user.

Why can't we return the exit code save it and then return it after the delete?

csivaguru: @revans2 I was planning to update this patch with the same approach (save the exit code and return it after deleting the tmpjar)

csivaguru:
@harshach, @revans2, @ptgoetz, @hmcl  I am abandoning the PR in favor of https://github.com/apache/storm/pull/1448 which I made against the right branch, 1.x-branch

PR #1444, opened 2016-05-25
[STORM-1277] port backtype.storm.daemon.executor to java

unsleepy22:
Sorry for the long delay on this.

This PR is not 100% completed yet, things remaining:
1. a few tests will fail and I'm still working on this (basically with method replacements like local-executor/local-transfer-executor-tuple, which was originally executor/mk-executor-transfer-fn)
2. my code base is behind master, so after all tests pass I will make a full merge from master
Hopefully I can finish the problems by this week.

Still, the base structure/design of the executor won't change, so I think I can get reviews from you first on the design.


unsleepy22:
sorry close and re-submit

PR #1449, opened 2016-05-27
[STORM-1868] Modify TridentKafkaWordCount to run in distributed mode

arunmahadevan:
@harshach @ptgoetz can you take a look ?


ptgoetz:
+1 This should probably also be applied to the 1.* branches.

harshach:
merged into master, 1.x, 1.0.x

PR #1451, opened 2016-05-31
STORM-1136: Command line module to return kafka spout offsets lag and…

priyank5485:
… display in storm ui.

### DIFF
--- a/external/storm-kafka-monitor/src/main/java/org/apache/storm/kafka/monitor/KafkaOffsetLagResult.java
+++ b/external/storm-kafka-monitor/src/main/java/org/apache/storm/kafka/monitor/KafkaOffsetLagResult.java
@@ -1,0 +1,102 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ *   or more contributor license agreements.  See the NOTICE file
+ *   distributed with this work for additional information
+ *   regarding copyright ownership.  The ASF licenses this file
+ *   to you under the Apache License, Version 2.0 (the
+ *   "License"); you may not use this file except in compliance
+ *   with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+
+package org.apache.storm.kafka.monitor;
+
+import org.json.simple.JSONAware;
+
+/**
+ * Class representing the log head offsets, spout offsets and the lag for a topic
+ */
+public class KafkaOffsetLagResult implements JSONAware {
+    private String topic;
+    private int partition;
+    private long consumerCommittedOffset;
+    private long logHeadOffset;
+    private long lag;
+
+    public KafkaOffsetLagResult(String topic, int parition, long consumerCommittedOffset, long logHeadOffset) {
+        this.topic = topic;
+        this.partition = parition;
+        this.consumerCommittedOffset = consumerCommittedOffset;
+        this.logHeadOffset = logHeadOffset;
+        this.lag = this.logHeadOffset - this.consumerCommittedOffset;
+    }
+
+    public String getTopic() {
+        return topic;
+    }
+
+    public int getPartition() {
+        return partition;
+    }
+
+    public long getConsumerCommittedOffset() {
+        return consumerCommittedOffset;
+    }
+
+    public long getLogHeadOffset() {
+        return logHeadOffset;
+    }
+
+    public long getLag() {
+        return lag;
+    }
+
+    @Override
+    public String toString() {
+        return "KafkaOffsetLagResult{" +
+                "topic='" + topic + '\'' +
+                ", partition=" + partition +
+                ", consumerCommittedOffset=" + consumerCommittedOffset +
+                ", logHeadOffset=" + logHeadOffset +
+                ", lag=" + lag +
+                '}';
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+
+        KafkaOffsetLagResult that = (KafkaOffsetLagResult) o;
+
+        if (partition != that.partition) return false;
+        if (consumerCommittedOffset != that.consumerCommittedOffset) return false;
+        if (logHeadOffset != that.logHeadOffset) return false;
+        if (lag != that.lag) return false;
+        return !(topic != null ? !topic.equals(that.topic) : that.topic != null);
+
+    }
+
+    @Override
+    public int hashCode() {
+        int result = topic != null ? topic.hashCode() : 0;
+        result = 31 * result + partition;
+        result = 31 * result + (int) (consumerCommittedOffset ^ (consumerCommittedOffset >>> 32));
+        result = 31 * result + (int) (logHeadOffset ^ (logHeadOffset >>> 32));
+        result = 31 * result + (int) (lag ^ (lag >>> 32));
+        return result;
+    }
+
+    @Override
+    public String toJSONString() {
+        return "{\"topic\":\"" + topic + "\",\"partition\":" + partition + ",\"consumerCommittedOffset\":" + consumerCommittedOffset + "," +
+                "\"logHeadOffset\":" + logHeadOffset + ",\"lag\":" + lag + "}";
+    }
+}
--- a/external/storm-kafka-monitor/src/main/java/org/apache/storm/kafka/monitor/KafkaOffsetLagUtil.java
+++ b/external/storm-kafka-monitor/src/main/java/org/apache/storm/kafka/monitor/KafkaOffsetLagUtil.java
@@ -1,0 +1,376 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ *   or more contributor license agreements.  See the NOTICE file
+ *   distributed with this work for additional information
+ *   regarding copyright ownership.  The ASF licenses this file
+ *   to you under the Apache License, Version 2.0 (the
+ *   "License"); you may not use this file except in compliance
+ *   with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+
+package org.apache.storm.kafka.monitor;
+
+import kafka.api.OffsetRequest;
+import kafka.api.PartitionOffsetRequestInfo;
+import kafka.common.TopicAndPartition;
+import kafka.javaapi.OffsetResponse;
+import kafka.javaapi.consumer.SimpleConsumer;
+import org.apache.commons.cli.CommandLine;
+import org.apache.commons.cli.CommandLineParser;
+import org.apache.commons.cli.DefaultParser;
+import org.apache.commons.cli.HelpFormatter;
+import org.apache.commons.cli.Options;
+import org.apache.curator.framework.CuratorFramework;
+import org.apache.curator.framework.CuratorFrameworkFactory;
+import org.apache.curator.retry.RetryOneTime;
+import org.apache.kafka.clients.consumer.KafkaConsumer;
+import org.apache.kafka.clients.consumer.OffsetAndMetadata;
+import org.apache.kafka.common.PartitionInfo;
+import org.apache.kafka.common.TopicPartition;
+import org.json.simple.JSONValue;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Properties;
+
+/**
+ * Utility class for querying offset lag for kafka spout
+ */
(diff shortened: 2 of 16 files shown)


HeartSaVioR:
@priyank5485 Could you attach UI screenshot to see output easily? Thanks in advance.

priyank5485:
kafkaspoutlags_success
kafkaspoutlags_failure

priyank5485:
@HeartSaVioR I just attached the two screen shots. One when it successfully managed to get the lag result. Other one by stopping kafka in which case it will just show a message in the UI indicating the cause.

harshach:
@priyank5485 Nice!!. Can we rename the section as KafkaSpout Lag

priyank5485:
@harshach Will do that. Just to confirm, you mean the section header in UI under Topology Stats, right? Also, since the topology can technically have more than one kafka spout and we are showing lags for all spouts there I was thinking of keeping it as plural. May be Kafka Spouts Lag

Does that sound okay?

harshach:
@priyank5485 Yes. sounds good.

priyank5485:
@harshach Done.

external/storm-kafka-monitor/src/main/java/org/apache/storm/kafka/monitor/KafkaOffsetLagResult.java
```diff
@@ -0,0 +1,102 @@
+/*
```
harshach: can we put this in org.apache.storm.tools.KafkaOffsetLagTool.java

external/storm-kafka-monitor/src/main/java/org/apache/storm/kafka/monitor/KafkaOffsetQueryInfoNew.java
```diff
@@ -0,0 +1,76 @@
...
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+package org.apache.storm.kafka.monitor;
+
+/**
+ * Class representing information for querying kafka for log head offsets, consumer offsets and the difference for new kafka spout using new consumer api
+ */
+public class KafkaOffsetQueryInfoNew {
```
harshach: Should we call this as NewKafkaSpoutOffsetQuery and OldKafkaSpoutOffsetQery

external/storm-kafka/src/jvm/org/apache/storm/kafka/KafkaSpout.java
```diff
@@ -188,6 +189,47 @@ public void declareOutputFields(OutputFieldsDeclarer declarer) {
         }
     }
 
+    @Override
+    public Map<String, Object> getComponentConfiguration () {
```
harshach: looks to be quite a bit of logic in here. If it makes sense lets move most of it into KafkaUtils.java

harshach:
@priyank5485 this looks great. Left some minor nit-picks.

darionyaphet:
LGTM

priyank5485:
@harshach I have renamed Kafka Log Head Offset to Latest Offset, since Kafka is present in ui header and topic is a column already. I have also renamed the Query classes as per your suggestion.

Regarding moving it under package tool, I thought of this module specific to kafka and getting the offset lags. If you are okay with it we can keep it as is instead of creating a tools package. Or any other suggestion if you have.

Regarding moving getComponentConfiguration to KafkaUtils, KafkaUtils exist only in the old module. So i will keep it in the spout for now. I plan to add getComponentConfiguration for trident as well. If there is code duplication there, I will try to reuse it for old module and new module(when trident is ready).

Let me know if there is anything else you want me to address. Otherwise I will squash the commits

harshach:
@priyank5485 I overlooked that this code is in external/module we can keep the classes as it is. My comment regarding getComponentConfiguration as it looks to be doing bit of logic in KafkaSpout and my suggestion was to move that code to KafkaUtils and call buildKafkaSpoutComponentConfig and return the config map . This way spout code looks cleaner and may be buildKafkaSpoutComponentConfig can be reused for trident. 
Go ahead with you commit squashes and you can add me to sponsor list since its in external module

external/storm-kafka-monitor/src/main/java/org/apache/storm/kafka/monitor/KafkaOffsetLagUtil.java
```diff
@@ -0,0 +1,374 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Properties;
+
+/**
+ * Utility class for querying offset lag for kafka spout
+ */
+public class KafkaOffsetLagUtil {
+    private static final String OPTION_TOPIC_SHORT = "t";
```
hmcl: what is the difference between SHORT and LONG ? why t vs topics ?

external/storm-kafka-monitor/src/main/java/org/apache/storm/kafka/monitor/KafkaOffsetLagUtil.java
```diff
@@ -0,0 +1,374 @@
...
+        if (oldKafkaSpoutOffsetQuery.getPartitions() != null) {
+            String[] partitions = oldKafkaSpoutOffsetQuery.getPartitions().split(",");
+            String[] leaders = oldKafkaSpoutOffsetQuery.getLeaders().split(",");
+            for (int i = 0; i < leaders.length; ++i) {
+                if (!result.containsKey(leaders[i])) {
+                    result.put(leaders[i], new ArrayList<TopicPartition>());
+                }
+                result.get(leaders[i]).add(new TopicPartition(oldKafkaSpoutOffsetQuery.getTopic(), Integer.parseInt(partitions[i])));
+            }
+        } else {
```
hmcl: when can this scenario occur? There is a lot of code in some of these methods. It would be ideal to put some comments hinting on what is going on, and/or split into smaller methods.

external/storm-kafka-monitor/src/main/java/org/apache/storm/kafka/monitor/KafkaOffsetLagUtil.java
priyank5485: I think its just convenience. You can use -t or --topics.

external/storm-kafka-monitor/src/main/java/org/apache/storm/kafka/monitor/KafkaOffsetLagUtil.java
priyank5485: Old kafka spout has two ways for its configuration. 
1. StaticHosts
2. ZkHosts
StaticHosts is when user adds the leader for each partition statically at the time of building the topology. In that case we need to use that information to get latest offsets. For ZkHosts the KafkaSpout finds leaders in the prepare method using zk nodes and consumes data. Thats what the if else represents. I will add some documentation.

harshach:
Thanks @priyank5485 . This looks great and will be greatly helpful for all the users of KafkaSpout. Lets open another JIRA for trident support. +1.

priyank5485:
opened https://issues.apache.org/jira/browse/STORM-1891to track trident kafka spouts

storm-core/src/ui/public/topology.html
```diff
@@ -384,8 +389,51 @@ <h2 id="topology-resources-header">Topology resources</h2>
...
+                var isJson = function (input) {
+                  try {
+                    JSON.parse(input);
+                  } catch (e) {
+                    return false;
+                  }
+                  return true;
+                };
+                var kafkaSpoutsValidResults = kafkaSpoutLags.filter(function (ele) {return isJson(ele.spoutLagResult);});
+                var kafkaSpoutsErrorResults = kafkaSpoutLags.filter(function (ele) {return !isJson(ele.spoutLagResult);});
```
abellina: Could the response still be json in the error case? I think the response should be JSON at all times.

storm-core/src/ui/public/topology.html
```diff
@@ -384,8 +389,51 @@ <h2 id="topology-resources-header">Topology resources</h2>
...
+                var data = {};
+                if (kafkaSpoutsValidResults.length > 0) {
+                  data.kafkaSpoutsLagResults = [];
+                  kafkaSpoutsValidResults.forEach(function(ele) {
+                    var spoutLagResult = JSON.parse(ele.spoutLagResult);
+                    spoutLagResult.forEach(function(ele2) {
```
abellina: Could the rest endpoint return id instead of spoutId, and this loop can go away?

abellina: or it returns spoutId (as it does now), and the rest of the code is adapted to take spoutId.

abellina:
Could the new tables be made optional? That way when a topology doesn't use KafkaSpout, the tables don't show in the UI.

storm-core/src/ui/public/topology.html
priyank5485: @abellina If its an error we only want to show an appropriate message to the user as a string. Also, not that the top level response is still a json. I think string is fine here.

storm-core/src/ui/public/topology.html
priyank5485: We would still need the for loop as need to flatten the results out. I chose to use id in the json passed to template since the template uses id in other tables for ui. If you are fine we can leave it as is.

priyank5485:
@abellina The tables are optional already. If there are no kafka spouts then the response will be an empty array. Can you try running a non kafka spout topology and check? I did.

priyank5485:
@harshach Rebased 1.x branch since there were some merges since the PR was raised.

storm-core/src/ui/public/topology.html
```diff
@@ -384,8 +389,51 @@ <h2 id="topology-resources-header">Topology resources</h2>
...
             $('#topology-actions [data-toggle="tooltip"]').tooltip();
             $('#topology-visualization [data-toggle="tooltip"]').tooltip();
+
+            var lagUrl = "/api/v1/topology/"+topologyId+"/lag";
+            $.getJSON(lagUrl,function(lagResponse,status,jqXHR) {
+              if (lagResponse !== null && lagResponse !== undefined && lagResponse instanceof Array && lagResponse.length > 0) {
+                var kafkaSpoutsLagTemplate = $(template).filter("#topology-kafka-spouts-lag-template").html();
+                var spoutsErrorTemplate = $(template).filter("#topology-spouts-lag-error-template").html();
+
+                var kafkaSpoutLags = lagResponse.filter(function(ele) {return ele.spoutType === "KAFKA";});
```
abellina: is the /lag endpoint returning spouts that aren't KafkaSpout?

storm-core/src/ui/public/topology.html
abellina: @priyank5485 The result currently looks like (if I understand correctly):

[{spoutId: "spout-1", spoutLagResult: "{topic: '..', partition: '..', logHeadOffset: '..', ...}",
 {spoutId: "spout-2", spoutLagResult: "error message"]

I think this should be split into two arrays. Sample proposed response:

// notice no strings, it is strictly JSON.
 {valid: [{spoutId: "spout-1", spoutLagResult: [topic: '..', partition: '..', logHeadOffset: '..', ...]}, ... ],
  invalid: [{spoutId: "spout-2", error: ""}, ...]}

This tells you at a glance if something had an error: resp.invalid.length > 0, tells you if there are non-errors spouts: resp.valid.length > 0, and allows you to loop through resp.valid and resp.invalid to populate your tables.

abellina:
@priyank5485 Yes I see the tables don't get populated if lag responses are empty, but we still end up with the div markup on the page. It seems that this may be an area where other spouts could put their stats in the future. Could you call the div topology-spout-stats, and put a comment in the html that says it gets populated if stats are available?

storm-core/src/ui/public/topology.html
harshach: @abellina Yes thats the idea to extend it for other spouts. For now only kafkaspout supported

storm-core/src/ui/public/topology.html
priyank5485: @abellina The spoutLagResult property is actually an array. Hence the flattening. 

Let me elaborate the limitations of valid and invalid approach. The TopologySpoutLag class in storm-core called by ui server does not know anything about if the spoutLagResult is valid or not. Reason is,  the way to get lag information for kafka and other spouts is handled by making a shell call handled by an external module since we did not want any direct dependency on storm core. For now we have kafka, but we can add other types of spouts as well. Plus the fields in ui or the template for a different type of spout could be different. The only commonality I found was if something went wrong for getting lag info for underlying spout(kafka or other) then an error message is sent in response. This will work for all types of spouts. To do what you are saying we will have to inspect the response in TopologySpoutLag class. I felt it did not matter if it were ui doing that or server. I preferred ui. Let me know if it makes sense or not and if you still think we need to change something.

priyank5485:
@abellina I have added a comment for that div. As far as renaming it to topology-spout-stats is concerned I prefer the current name since right below that div there is a div showing stats for the spout like complete latency, etc. I dont want the two to be confused.

harshach:
sitll +1. @abellina can you address @priyank5485 response.

storm-core/src/ui/public/topology.html
abellina: @priyank5485 yes, sorry that was my intention but I didn't write it correctly in my comment, spoutLagResult should be a JSON array. Thanks for the background!

That said, I think that TopologySpoutLag should do this (my two cents). That way the output from the rest interface doesn't need further inspection and some other script can call on your new lag endpoint and get a JSON object they use without having to do the extra testing. I am not sure that valid/invalid are the best labels here by the way, just an example I came up with.

abellina:
@harshach I added a comment on the format for the response from the lag endpoint. I still believe the backend should do some of the work the UI is currently doing. 

@priyank5485 div name makes sense now. Thanks!

harshach:
@abellina lets handle the concern you've in https://issues.apache.org/jira/browse/STORM-1891.

PR #1452, opened 2016-05-31
Python 3 print statement error

Darkless012:


macheins:
Maybe should add a `from __future__ import print_function` to ensure backward compatibility.

HeartSaVioR:
In fact, no need to add __future__ `for now`, but it's more safer to add this.
+1

amitripshtos:
+1

HeartSaVioR:
Thanks @Darkless012, I merged into master, 1.x, 1.0.x branch so it'll be reflected to upcoming 1.0.2.

PR #1453, opened 2016-05-31
STORM-1873 Implement alternative behaviour for late tuples

kosii:


storm-core/src/jvm/org/apache/storm/Config.java
```diff
@@ -1872,6 +1872,13 @@
     @isString
     public static final String TOPOLOGY_BOLTS_TUPLE_TIMESTAMP_FIELD_NAME = "topology.bolts.tuple.timestamp.field.name";
 
+    /**
+     * Bolt-specific configuration for windowed bolts to specify whether late tuples should be emitted on a stream
+     * called _late, instead of being logged with INFO level.
+     */
+    @isBoolean
+    public static final String TOPOLOGY_BOLTS_EMIT_LATE_TUPLE = "topology.bolts.emit.late.tuple";
```
arunmahadevan: Any reason for using a boolean and hardcoding the late tuple stream instead of letting users specify the late tuple stream (like in your earlier [patch](https://github.com/kosii/storm/commit/216c991da3c5b6c6cac1b25182b86507c3fb5e9e#diff-bf65d6d71c61b8fe2d118c4f7433de14R1880)) ? It may be better to let users specify the late tuple stream since it gives them more flexibility.

storm-core/src/jvm/org/apache/storm/topology/WindowedBoltExecutor.java
```diff
@@ -268,8 +277,12 @@ public void execute(Tuple input) {
             if (waterMarkEventGenerator.track(input.getSourceGlobalStreamId(), ts)) {
                 windowManager.add(input, ts);
             } else {
+                if (emitLateTuples()) {
+                    windowedOutputCollector.emit(LATE_TUPLE_STREAM, input, new Values(input));
+                } else {
+                    LOG.info("Received a late tuple {} with ts {}. This will not processed.", input, ts);
```
arunmahadevan: nit: (existing) `will not` -> `will not be`

storm-core/src/jvm/org/apache/storm/topology/base/BaseWindowedBolt.java
```diff
@@ -255,6 +255,17 @@ public BaseWindowedBolt withTimestampField(String fieldName) {
     }
 
     /**
+     * Turns on or off, on a per component basis, the emission of late tuples on a stream
+     * called _late.
+     *
+     * @param flag the flag to indicate whether to emit late tuples
+     */
+    public BaseWindowedBolt withLateTupleEmission(Boolean flag) {
+        windowConfiguration.put(Config.TOPOLOGY_BOLTS_EMIT_LATE_TUPLE, flag);
```
arunmahadevan: good to add validation to ensure this is used with timestamp field (like you had in your earlier patch).

arunmahadevan:
I have gone through the details and have left some comments. Can you update the [documentation](https://github.com/apache/storm/blob/master/docs/Windowing.md) where it references late tuples and may be explain how to use this feature? 

It would also be good to add some unit tests.

storm-core/src/jvm/org/apache/storm/Config.java
kosii: I realized that it was hard to achieve a consistent behavior in the earlier implementation, because at the moment of calling `declareOutputFields`, we only have access to the component specific parameters, but if the user puts the key in the global config but forgets to specify at component level, then the stream won't be declared, but in the `execute` method it will still going to try to emit messages to the non-existing stream.
On the other hand in the `initWindowManager` method, I can check whether the stream exist, and raise an exception if not. Does it seem okay to you? Because in this case I'd also prefer the per component based approach.

arunmahadevan: Yes it sounds reasonable. An exception can be thrown if the `lateTupleStream` is set in the config but the same is not defined in the bolt's output streams.

kosii: Cool, I'm going to update this pull request with the first version

storm-core/src/jvm/org/apache/storm/topology/WindowedBoltExecutor.java
kosii: fixed

kosii:
I updated the docs, and I'm going to work on the unit tests while the patch is being reviewed

storm-core/src/jvm/org/apache/storm/topology/base/BaseWindowedBolt.java
```diff
@@ -255,6 +255,22 @@ public BaseWindowedBolt withTimestampField(String fieldName) {
...
 
     /**
+     * Specify a stream id on which late tuples are going to be emitted. When using this method,
+     * {@link Config#TOPOLOGY_BOLTS_TUPLE_TIMESTAMP_FIELD_NAME} should be defined on a per component basis before
+     * calling this method, otherwise {@link IllegalArgumentException} will be thrown.
+     *
+     * @param streamId the name of the stream used to emit late tuples on
+     */
+    public BaseWindowedBolt withLateTupleStream(String streamId) {
+        if (!windowConfiguration.containsKey(Config.TOPOLOGY_BOLTS_TUPLE_TIMESTAMP_FIELD_NAME)) {
```
arunmahadevan: On re-thinking, can this validation be moved to `initWindowManager` or `validate` so that we don't impose an ordering in the builder?

storm-core/src/jvm/org/apache/storm/topology/base/BaseWindowedBolt.java
```diff
@@ -255,6 +255,22 @@ public BaseWindowedBolt withTimestampField(String fieldName) {
     }
 
     /**
+     * Specify a stream id on which late tuples are going to be emitted. When using this method,
+     * {@link Config#TOPOLOGY_BOLTS_TUPLE_TIMESTAMP_FIELD_NAME} should be defined on a per component basis before
```
arunmahadevan: may be refer to `withTimeStampField` method instead of the config constant.

storm-core/src/jvm/org/apache/storm/topology/WindowedBoltExecutor.java
```diff
@@ -284,6 +298,10 @@ public void cleanup() {
 
     @Override
     public void declareOutputFields(OutputFieldsDeclarer declarer) {
+        String lateTupleStream = (String) getComponentConfiguration().get(Config.TOPOLOGY_BOLTS_LATE_TUPLE_STREAM);
+        if (lateTupleStream != null) {
+            declarer.declareStream(lateTupleStream, new Fields("late_tuple"));
```
arunmahadevan: Can you make "late_tuple" a constant and put a note in the javadoc of `withLateTupleStream` and Windowing.md that tuple can be accessed via the field "late_tuple" ?

kosii: should it be public, so that users could refer to it by the variable name?

arunmahadevan: @kosii yes it can be public static final.

storm-core/src/jvm/org/apache/storm/topology/WindowedBoltExecutor.java
```diff
@@ -188,6 +189,10 @@ private void validate(Map stormConf, Count windowLengthCount, Duration windowLen
             }
             waterMarkEventGenerator = new WaterMarkEventGenerator<>(manager, watermarkInterval,
                                                                     maxLagMs, getComponentStreams(context));
+        } else {
+            if (stormConf.containsKey(Config.TOPOLOGY_BOLTS_TUPLE_TIMESTAMP_FIELD_NAME)) {
```
arunmahadevan: Shouldn't this be Config.TOPOLOGY_BOLTS_LATE_TUPLE_STREAM ?

kosii: Yes, you are right. I'm gonna add more unit tests

arunmahadevan:
+1 Looks good

storm-core/test/jvm/org/apache/storm/topology/WindowedBoltExecutorTest.java
```diff
@@ -28,19 +28,20 @@
...
 import org.junit.Test;
+import org.mockito.Matchers;
 import org.mockito.Mockito;
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
satishd: minor nit: Better not to have wildcard imports.

satishd:
+1 LGTM except a minor nit.

storm-core/test/jvm/org/apache/storm/topology/WindowedBoltExecutorTest.java
kosii: should I squash after fixing it?

satishd: Sure, it is suggested to do that.

kosii:
@arunmahadevan @satishd nit is fixed and commits are squashed. thanks for the review!

ptgoetz:
+1

arunmahadevan:
Thanks @kosii merged to master and 1.x-branch. I did not merge to 1.0.x-branch since this adds a new api and as @ptgoetz suggested it can be targeted for 1.1.0 release.

PR #1457, opened 2016-06-01
Fix jira-github-join tools: Apache Jira doesn't allow request with de…

HeartSaVioR:
…fault python user-agent


HeartSaVioR:
I'd be happy if someone take a look.

abellina:
@HeartSaVioR, do you have an error that JIRA returns? Should there be a ticket against Atlassian?

HeartSaVioR:
@abellina It seems to be intermittent. I can't reproduce that and it is fine now.

PR #1458, opened 2016-06-02
STORM-1878: Flux can now handle IStatefulBolts

sakanaou:
We noticed that Flux could not yet handle IStatefulBolts in a topology definition. With this small addition we can successfully build and deploy our topology.


arunmahadevan:
@sakanaou Thanks for the patch. 

Can you also add an example flux topology with an IStatefulBolt similar to to existing ones under https://github.com/apache/storm/tree/master/external/flux/flux-examples ?

sakanaou:
@arunmahadevan I have added an example as requested. It is basically an extension of the word count example with state support.

arunmahadevan:
+1

ptgoetz:
+1

arunmahadevan:
Thanks @sakanaou merged to master, 1.x and 1.0.x branches.

PR #1459, opened 2016-06-02
[STORM-1875]Separate Jedis/JedisCluster Config

darionyaphet:
[STORM-1875](https://issues.apache.org/jira/browse/STORM-1875)

Separate Jedis / JedisCluster to provide full operations for each environment to users .

### DIFF
--- a/external/storm-redis/src/main/java/org/apache/storm/redis/bolt/AbstractRedisBolt.java
+++ b/external/storm-redis/src/main/java/org/apache/storm/redis/bolt/AbstractRedisBolt.java
@@ -17,6 +17,7 @@
  */
 package org.apache.storm.redis.bolt;
 
+import org.apache.storm.redis.common.config.JedisConfig;
 import org.apache.storm.task.OutputCollector;
 import org.apache.storm.task.TopologyContext;
 import org.apache.storm.topology.base.BaseRichBolt;
@@ -24,6 +25,8 @@ import org.apache.storm.redis.common.config.JedisClusterConfig;
 import org.apache.storm.redis.common.config.JedisPoolConfig;
 import org.apache.storm.redis.common.container.JedisCommandsContainerBuilder;
 import org.apache.storm.redis.common.container.JedisCommandsInstanceContainer;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import redis.clients.jedis.JedisCommands;
 
 import java.util.Map;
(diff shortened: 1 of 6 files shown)


HeartSaVioR:
The problem is actually not simple. The blocking point is JedisCommands interface, not config.
JedisCommands exposes operations which use only single key. Since some multiple keys operations of Redis is not suitable for Redis Cluster (operation is available but not making sense), Jedis and JedisCluster takes different interfaces, which blocks us to provide unified interface from storm-redis side.

One simple way is separating existing classes to RedisLookupBolt / RedisClusterLookupBolt, RedisStoreBolt / RedisClusterStoreBol, as State/MapState did, but it incurs backward incompatibility if users use RedisLookupBolt / RedisStoreBolt to access Redis Cluster.

HeartSaVioR:
One other way without breaking compatibility but not beauty is adding some utility methods to AbstractRedisBolt like, 

```
public JedisCluster asJedisCluster(JedisCommands instance) throws IllegalArgumentException
public JedisCluster asJedis(JedisCommands instance) throws IllegalArgumentException
```

and let users convert their JedisCommands instance to actual type before using full features.

darionyaphet:
Both `Jedis` and `JedisCluster ` have implemented `JedisCommands` interface . So I think we can use `JedisCommands` calling the same method on single redis instance and redis cluster . 

Just use `JedisConfig` to control instance `JedisContainer` or `JedisCommandsInstanceContainer` maybe a good idea and support two difference interface is useless :P

HeartSaVioR:
@darionyaphet 
This is Jedis class diagram when Jedis was 2.5.0. At that time JedisCluster doesn't support multiple keys operation so now it becomes more complicated.
https://lh5.googleusercontent.com/-kMPsNHe6t2I/U0Dn01HxIoI/AAAAAAAAt4U/U8I0j570rI4/s3200/Jedis+Class+Diagram+without+fields_methods.jpg

DISCLAIMER: I'm also a committer of Jedis. :)

darionyaphet:
@HeartSaVioR Good !  I will try to split into two class . BTW Do you have any idea to unified `Jedis` and `JedisCluster` interface ?

darionyaphet:
`Storm-Redis` have support multiple keys operation ?

HeartSaVioR:
No for predefined operations (StoreBolt, LookupBolt) but we can still support full features for AbstractRedisBolt and let users extend this class.

darionyaphet:
OK , I will try to separating classes :D

PR #1460, opened 2016-06-03
STORM-1723 Introduce ClusterMetricsConsumer

HeartSaVioR:
Pull request for 1.x is here: #1352 

* ClusterMetricsConsumer publishes cluster-side related metrics into consumers
  * like MetricsConsumer for topology metrics
* Users can implement IClusterMetricsConsumer and configure to cluster conf. file to take effect
  * Please refer conf/storm.yaml.example for more details on configuring
  * Nimbus should be launched with additional jars which are needed for IClusterMetricsConsumer
* Also did some refactor to nimbus.clj

### DIFF
--- a/conf/storm.yaml.example
+++ b/conf/storm.yaml.example
@@ -46,3 +46,12 @@
 #     parallelism.hint: 1
 #     argument:
 #       - endpoint: "metrics-collector.mycompany.org"
+
+## Cluster Metrics Consumers
+# storm.cluster.metrics.consumer.register:
+#   - class: "org.apache.storm.metric.LoggingClusterMetricsConsumer"
+#   - class: "org.mycompany.MyMetricsConsumer"
+#     argument:
+#       - endpoint: "metrics-collector.mycompany.org"
+#
+# storm.cluster.metrics.consumer.publish.interval.secs: 60
--- a/storm-core/src/jvm/org/apache/storm/metric/ClusterMetricsConsumerExecutor.java
+++ b/storm-core/src/jvm/org/apache/storm/metric/ClusterMetricsConsumerExecutor.java
@@ -1,0 +1,70 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.storm.metric;
+
+import org.apache.storm.Config;
+import org.apache.storm.metric.api.DataPoint;
+import org.apache.storm.metric.api.IClusterMetricsConsumer;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Collection;
+
+public class ClusterMetricsConsumerExecutor {
+    public static final Logger LOG = LoggerFactory.getLogger(ClusterMetricsConsumerExecutor.class);
+
+    private IClusterMetricsConsumer metricsConsumer;
+    private String consumerClassName;
+    private Object registrationArgument;
+
+    public ClusterMetricsConsumerExecutor(String consumerClassName, Object registrationArgument) {
+        this.consumerClassName = consumerClassName;
+        this.registrationArgument = registrationArgument;
+    }
+
+    public void prepare() {
+        try {
+            metricsConsumer = (IClusterMetricsConsumer)Class.forName(consumerClassName).newInstance();
+        } catch (Exception e) {
+            throw new RuntimeException("Could not instantiate a class listed in config under section " +
+                    Config.STORM_CLUSTER_METRICS_CONSUMER_REGISTER + " with fully qualified name " + consumerClassName, e);
+        }
+
+        metricsConsumer.prepare(registrationArgument);
+    }
+
+    public void handleDataPoints(final IClusterMetricsConsumer.ClusterInfo clusterInfo, final Collection<DataPoint> dataPoints) {
+        try {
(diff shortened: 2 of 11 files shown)


HeartSaVioR:
whitespace again. I'll take care of this.

HeartSaVioR:
Fixed whitespaces, rebased and squashed.

conf/storm.yaml.example
```diff
@@ -46,3 +46,12 @@
 #     parallelism.hint: 1
 #     argument:
 #       - endpoint: "metrics-collector.mycompany.org"
+
+## Cluster Metrics Consumers
+# storm.cluster.metrics.consumer.register:
+#   - class: "org.apache.storm.metric.LoggingClusterMetricsConsumer"
+#   - class: "org.mycompany.MyMetricsConsumer"
+#     argument:
+#       - endpoint: "metrics-collector.mycompany.org"
```
harshach: is this REST api end point and I assume these arguments are passed as constructor argument to consumers

storm-core/src/jvm/org/apache/storm/metric/ClusterMetricsConsumerExecutor.java
```diff
@@ -0,0 +1,87 @@
...
+        metricsConsumer.prepare(registrationArgument);
+
+        ThreadPoolExecutor singleThreadExecutor = new ThreadPoolExecutor(1, 1, 0L, TimeUnit.MILLISECONDS,
+                new LinkedBlockingQueue<Runnable>());
+        consumerExecutorService = MoreExecutors.getExitingExecutorService(singleThreadExecutor);
+    }
+
+    public void handleDataPoints(final IClusterMetricsConsumer.ClusterInfo clusterInfo, final Collection<DataPoint> dataPoints) {
+        final IClusterMetricsConsumer metricsConsumerForArg = this.metricsConsumer;
+        consumerExecutorService.execute(new Runnable() {
```
harshach: given that we are using scheduleRecurring to run the this do we need the executor service here.

conf/storm.yaml.example
HeartSaVioR: @harshach 
This is just an example which is same as above, `Metrics Consumer`. Class name could be changed to represent Cluster Metrics but it's just an example so I guess that would be fine.

storm-core/src/jvm/org/apache/storm/metric/LoggingClusterMetricsConsumer.java
```diff
@@ -0,0 +1,86 @@
...
+ * ```
+ *
+ */
+public class LoggingClusterMetricsConsumer implements IClusterMetricsConsumer {
+    public static final Logger LOG = LoggerFactory.getLogger(LoggingClusterMetricsConsumer.class);
+
+    static private String padding = "                       ";
+
+    @Override
+    public void prepare(Object registrationArgument) {
```
harshach: wouldn't be better to get this Map<String, ?> config to have more flexibility in passing the arguments.

storm-core/src/jvm/org/apache/storm/metric/LoggingClusterMetricsConsumer.java
```diff
@@ -0,0 +1,86 @@
...
+    public void cleanup() {
+    }
+
+    private void logDataPoints(Collection<DataPoint> dataPoints, StringBuilder sb, String header) {
+        for (DataPoint p : dataPoints) {
+            sb.delete(header.length(), sb.length());
+            sb.append(p.getName())
+                    .append(padding).delete(header.length()+23,sb.length()).append("\t")
+                    .append(p.getValue());
+            LOG.info(sb.toString());
```
harshach: can we make this at DEBUG level other wise this might get bit noisy.

harshach:
@HeartSaVioR Overall looks good to me. I have few minor nits.

storm-core/src/jvm/org/apache/storm/metric/ClusterMetricsConsumerExecutor.java
HeartSaVioR: I don't want to let Nimbus die when MetricsConsumer has an issue so exception is being thrown.
With enclosing this with ExecutorService, Nimbus itself is not affected though metricsConsumer.handleDataPoints() throws some exceptions.

But we can still enclose call with try-catch with getting rid of ExecutorService. What do you think?

storm-core/src/jvm/org/apache/storm/metric/LoggingClusterMetricsConsumer.java
HeartSaVioR: This configuration is from yaml, same as MetricsConsumer, so it can be String, List of String, Map<String, Object>, and so on.

I don't like this way but having similar interface with IMetricsConsumer would be easier to use.

storm-core/src/jvm/org/apache/storm/metric/LoggingClusterMetricsConsumer.java
HeartSaVioR: LoggingClusterMetricsConsumer will not be used as default, and it logs other place as same as topology metrics consumer. Please refer to change of log4j2.cluster.xml.

HeartSaVioR: It's same as https://github.com/apache/storm/blob/master/storm-core/src/jvm/org/apache/storm/metric/LoggingMetricsConsumer.java

storm-core/src/jvm/org/apache/storm/metric/ClusterMetricsConsumerExecutor.java
harshach: @HeartSaVioR  I feel like enclosing try , catch is better than using a executor service unless there is any specific need other than catching exceptions in metrics consumer. If thats the case I prefer using try, catch than executorService.

storm-core/src/jvm/org/apache/storm/metric/LoggingClusterMetricsConsumer.java
harshach: ok makes sense.

storm-core/src/jvm/org/apache/storm/metric/LoggingClusterMetricsConsumer.java
harshach: LGTM.

storm-core/src/jvm/org/apache/storm/metric/ClusterMetricsConsumerExecutor.java
HeartSaVioR: @harshach Addressed. Thanks!

harshach:
+1

PR #1461, opened 2016-06-05
[STORM-1882] Expose TextFileReader public

darionyaphet:
[STORM-1882](https://issues.apache.org/jira/browse/STORM-1882)
TextFileReader is package-private .Should make TextFileReader a public class to expose it to user and hdfs spout.


harshach:
+1

ptgoetz:
+1

PR #1462, opened 2016-06-05
[STORM-1883] FileReader extends Closeable Interface

darionyaphet:
[STORM-1883](https://issues.apache.org/jira/browse/STORM-1883)

use Closeable Interface to decorate FileReader to support close()


dossett:
Are there any functional changes here?  I'm all in favor of using standard java interfaces, but I don't see any benefits of `closeable` being used here.  Possible changes could include:

- Using the try-with-resources idiom now that FileReader is `closeable` [1]
- Concrete classes no longer have to catch their own `IOException`s since close is declared to throw that [2]

[1] - https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html
[2] - https://github.com/darionyaphet/storm/blob/ad74ad10012df6999c94652993bcd66e90db53ff/external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/TextFileReader.java#L104

darionyaphet:
I think `FileReader` should same with 'java.io. FileReader' . Implement `Closeable` interface also have a better extension structure .

HeartSaVioR:
I'm also in favor of using standard java interface. Just it seems not have any valuable benefits. I'm fine either change or keep it as is.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/FileReader.java
```diff
@@ -20,10 +20,11 @@
 
 import org.apache.hadoop.fs.Path;
 
+import java.io.Closeable;
 import java.io.IOException;
 import java.util.List;
 
-interface FileReader {
+interface FileReader extends Closeable{
```
harshach: please add space after the Closeable

harshach:
+1 after addressing the minor comment . @roshannaik can you review as well.

PR #1470, opened 2016-06-07
STORM-1886 Extend KeyValueState iface with delete

kosii:
The patch also provides implementation for delete in
RedisKeyValueState and InMemoryKeyValueState.

### DIFF
--- a/external/storm-redis/src/main/java/org/apache/storm/redis/state/RedisKeyValueState.java
+++ b/external/storm-redis/src/main/java/org/apache/storm/redis/state/RedisKeyValueState.java
@@ -17,6 +17,7 @@
  */
 package org.apache.storm.redis.state;
 
+import com.google.common.base.Optional;
 import org.apache.storm.state.DefaultStateSerializer;
 import org.apache.storm.state.KeyValueState;
 import org.apache.storm.state.Serializer;
@@ -31,8 +32,8 @@ import org.apache.commons.codec.binary.Base64;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
-import java.util.Set;
-import java.util.HashSet;
+import java.util.List;
+import java.util.ArrayList;
 import java.util.concurrent.ConcurrentHashMap;
 
 /**
@@ -46,13 +47,12 @@ public class RedisKeyValueState<K, V> implements KeyValueState<K, V> {
     private final String namespace;
     private final String prepareNamespace;
     private final String txidNamespace;
+    private final String tombstone;
     private final Serializer<K> keySerializer;
-    private final Serializer<V> valueSerializer;
+    private final Serializer<Optional<V>> valueSerializer;
     private final JedisCommandsInstanceContainer jedisContainer;
     private Map<String, String> pendingPrepare;
-    private Set<String> pendingDeletePrepare;
     private Map<String, String> pendingCommit;
-    private Set<String> pendingDeleteCommit;
     private Map<String, String> txIds;
 
     public RedisKeyValueState(String namespace) {
@@ -60,23 +60,23 @@ public class RedisKeyValueState<K, V> implements KeyValueState<K, V> {
     }
 
     public RedisKeyValueState(String namespace, JedisPoolConfig poolConfig) {
-        this(namespace, poolConfig, new DefaultStateSerializer<K>(), new DefaultStateSerializer<V>());
+        this(namespace, poolConfig, new DefaultStateSerializer<K>(), new DefaultStateSerializer<Optional<V>>());
     }
 
-    public RedisKeyValueState(String namespace, JedisPoolConfig poolConfig, Serializer<K> keySerializer, Serializer<V> valueSerializer) {
+    public RedisKeyValueState(String namespace, JedisPoolConfig poolConfig, Serializer<K> keySerializer, Serializer<Optional<V>> valueSerializer) {
         this(namespace, JedisCommandsContainerBuilder.build(poolConfig), keySerializer, valueSerializer);
     }
 
     public RedisKeyValueState(String namespace, JedisCommandsInstanceContainer jedisContainer,
-                              Serializer<K> keySerializer, Serializer<V> valueSerializer) {
+                              Serializer<K> keySerializer, Serializer<Optional<V>> valueSerializer) {
         this.namespace = namespace;
         this.prepareNamespace = namespace + "$prepare";
         this.txidNamespace = namespace + "$txid";
         this.keySerializer = keySerializer;
         this.valueSerializer = valueSerializer;
+        this.tombstone = encode(valueSerializer.serialize(Optional.<V>absent()));
         this.jedisContainer = jedisContainer;
-        this.pendingPrepare = new HashMap<>();
-        this.pendingDeletePrepare = new HashSet<>();
+        this.pendingPrepare = new ConcurrentHashMap<>();
         initTxids();
         initPendingCommit();
     }
@@ -103,11 +103,9 @@ public class RedisKeyValueState<K, V> implements KeyValueState<K, V> {
             if (commands.exists(prepareNamespace)) {
                 LOG.debug("Loading previously prepared commit from {}", prepareNamespace);
                 pendingCommit = Collections.unmodifiableMap(commands.hgetAll(prepareNamespace));
-                pendingDeleteCommit = Collections.emptySet();
             } else {
                 LOG.debug("No previously prepared commits.");
                 pendingCommit = Collections.emptyMap();
-                pendingDeleteCommit = Collections.emptySet();
             }
         } finally {
             jedisContainer.returnInstance(commands);
@@ -118,11 +116,8 @@ public class RedisKeyValueState<K, V> implements KeyValueState<K, V> {
     public void put(K key, V value) {
         LOG.debug("put key '{}', value '{}'", key, value);
         String redisKey = encode(keySerializer.serialize(key));
-        synchronized (this) {
-            pendingPrepare.put(redisKey,
-                    encode(valueSerializer.serialize(value)));
-            pendingDeletePrepare.remove(redisKey);
-        }
+        pendingPrepare.put(redisKey,
+                encode(valueSerializer.serialize(Optional.of(value))));
     }
 
     @Override
@@ -130,17 +125,11 @@ public class RedisKeyValueState<K, V> implements KeyValueState<K, V> {
         LOG.debug("get key '{}'", key);
         String redisKey = encode(keySerializer.serialize(key));
         String redisValue = null;
-        boolean found = false;
-        synchronized (this) {
-            if (pendingPrepare.containsKey(redisKey) || pendingDeletePrepare.contains(redisKey)) {
-                redisValue = pendingPrepare.get(redisKey);
-                found = true;
-            } else if (pendingCommit.containsKey(redisKey) || pendingDeleteCommit.contains(redisKey)) {
-                redisValue = pendingCommit.get(redisKey);
-                found = true;
-            }
-        }
-        if (!found) {
+        if (pendingPrepare.containsKey(redisKey)) {
+            redisValue = pendingPrepare.get(redisKey);
+        } else if (pendingCommit.containsKey(redisKey)) {
+            redisValue = pendingCommit.get(redisKey);
+        } else {
             JedisCommands commands = null;
             try {
                 commands = jedisContainer.getInstance();
@@ -151,7 +140,7 @@ public class RedisKeyValueState<K, V> implements KeyValueState<K, V> {
         }
         V value = null;
         if (redisValue != null) {
-            value = valueSerializer.deserialize(decode(redisValue));
+            value = valueSerializer.deserialize(decode(redisValue)).orNull();
         }
         LOG.debug("Value for key '{}' is '{}'", key, value);
         return value;
@@ -164,17 +153,16 @@ public class RedisKeyValueState<K, V> implements KeyValueState<K, V> {
     }
 
     @Override
-    public void delete(K key) {
+    public V delete(K key) {
         LOG.debug("delete key '{}'", key);
         String redisKey = encode(keySerializer.serialize(key));
-        synchronized (this) {
-            pendingDeletePrepare.add(redisKey);
-            pendingPrepare.remove(redisKey);
-        }
+        V curr = get(key);
+        pendingPrepare.put(redisKey, tombstone);
+        return curr;
     }
 
     @Override
-    public void prepareCommit(long txid){
+    public void prepareCommit(long txid) {
         LOG.debug("prepareCommit txid {}", txid);
         validatePrepareTxid(txid);
         JedisCommands commands = null;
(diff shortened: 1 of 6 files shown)


external/storm-redis/src/main/java/org/apache/storm/redis/state/RedisKeyValueState.java
```diff
@@ -77,7 +81,8 @@ public RedisKeyValueState(String namespace, JedisCommandsInstanceContainer jedis
         this.keySerializer = keySerializer;
         this.valueSerializer = valueSerializer;
         this.jedisContainer = jedisContainer;
-        this.pendingPrepare = new ConcurrentHashMap<>();
```
arunmahadevan: Could we somehow handle the deletes by maintaining an entry (with some extra info) within the pendingPrepare itself and avoid the extra hash sets and synchronized ? Something along the lines of maintaining a map of
`key -> {ADD or DELETE, value}`. The values to be deleted could be written to a separate Redis Set during prepare.

During `delete`, insert a DELETE entry into pendingPrepare.
During `prepare`, snapshot pendingPrepare to pendingCommit and also save the DELETE entries to a redis set.
During `commit`, delete the DELETE entries from redis map.
During `get`, if key is in pendingPrepare, pendingCommit with DELETE or the key is not found return null.

external/storm-redis/src/main/java/org/apache/storm/redis/state/RedisKeyValueState.java
```diff
@@ -104,9 +109,11 @@ private void initPendingCommit() {
             if (commands.exists(prepareNamespace)) {
                 LOG.debug("Loading previously prepared commit from {}", prepareNamespace);
                 pendingCommit = Collections.unmodifiableMap(commands.hgetAll(prepareNamespace));
+                pendingDeleteCommit = Collections.emptySet();
```
arunmahadevan: during init, shouldn't the pendingDeleteCommit also be restored from previously saved state ?

external/storm-redis/src/main/java/org/apache/storm/redis/state/RedisKeyValueState.java
```diff
@@ -104,9 +109,11 @@ private void initPendingCommit() {
...
             } else {
                 LOG.debug("No previously prepared commits.");
                 pendingCommit = Collections.emptyMap();
+                pendingDeleteCommit = Collections.emptySet();
```
arunmahadevan: pendingDeleteCommits should probably be saved to a separate namespace say a redis set?

external/storm-redis/src/main/java/org/apache/storm/redis/state/RedisKeyValueState.java
```diff
@@ -116,20 +123,30 @@ private void initPendingCommit() {
...
         String redisKey = encode(keySerializer.serialize(key));
         String redisValue = null;
-        if (pendingPrepare.containsKey(redisKey)) {
-            redisValue = pendingPrepare.get(redisKey);
-        } else if (pendingCommit.containsKey(redisKey)) {
-            redisValue = pendingCommit.get(redisKey);
-        } else {
+        boolean found = false;
+        synchronized (this) {
+            if (pendingPrepare.containsKey(redisKey) || pendingDeletePrepare.contains(redisKey)) {
```
arunmahadevan: if pendingDeletePrepare contains the key the value should not be returned.

external/storm-redis/src/main/java/org/apache/storm/redis/state/RedisKeyValueState.java
```diff
@@ -163,15 +190,20 @@ public void prepareCommit(long txid) {
                 LOG.debug("Prepared txn already exists, will merge", txid);
                 pendingPrepare.putAll(pendingCommit);
             }
-            if (!pendingPrepare.isEmpty()) {
+            if (!pendingPrepare.isEmpty() || !pendingDeletePrepare.isEmpty()) {
                 commands.hmset(prepareNamespace, pendingPrepare);
+                commands.hdel(prepareNamespace, pendingDeletePrepare.toArray(new String[pendingDeletePrepare.size()]));
```
arunmahadevan: Shouldn't you save the pendingDeletePrepare keys separately into a redis hashset. If a previously saved key (only in redis) is deleted, this may not work.

external/storm-redis/src/main/java/org/apache/storm/redis/state/RedisKeyValueState.java
```diff
@@ -184,15 +216,17 @@ public void commit(long txid) {
         JedisCommands commands = null;
         try {
             commands = jedisContainer.getInstance();
-            if (!pendingCommit.isEmpty()) {
+            if (!pendingCommit.isEmpty() || !pendingDeleteCommit.isEmpty()) {
                 commands.hmset(namespace, pendingCommit);
```
arunmahadevan: Redis may throw error if pendingCommit is empty. you may want to confirm.

external/storm-redis/src/main/java/org/apache/storm/redis/state/RedisKeyValueState.java
```diff
@@ -203,12 +237,14 @@ public void commit() {
         JedisCommands commands = null;
         try {
             commands = jedisContainer.getInstance();
-            if (!pendingPrepare.isEmpty()) {
+            if (!pendingPrepare.isEmpty() || !pendingDeletePrepare.isEmpty()) {
                 commands.hmset(namespace, pendingPrepare);
```
arunmahadevan: Redis may throw error if pendingPrepare is empty. you may want to confirm.

storm-core/src/jvm/org/apache/storm/state/KeyValueState.java
```diff
@@ -45,4 +45,11 @@
      * @return the value or defaultValue if no mapping is found
      */
     V get(K key, V defaultValue);
+
+    /**
+     * Deletes the value mapped to the key, if there is any
+     *
+     * @param key   the key
+     */
+    void delete(K key);
```
arunmahadevan: Could probably be changed to `V remove(K key)` along the lines of java.util.Map.

arunmahadevan:
@kosii the delete api is useful. Lets also try to see if we could handle the deletes by maintaining an entry within the pendingPrepare itself and avoid the extra hashsets and synchronized as mentioned in https://github.com/apache/storm/pull/1470/files#r66784961

kosii:
Thanks @arunmahadevan for the thorough review, I'm gonna address each comments asap

external/storm-redis/src/main/java/org/apache/storm/redis/state/RedisKeyValueState.java
kosii: Yes, you are right, I forget easily that the `pendingPrepare`/`pendingCommit` doesn't actually contain the snapshot, just a delta

storm-core/src/jvm/org/apache/storm/state/KeyValueState.java
kosii: It was my initial idea, but in case of the Redis implementation it needs and extra database request , as the key might not currently in one of the `pending*` structures. Does the overhead of the extra query looks negligible to you?

external/storm-redis/src/main/java/org/apache/storm/redis/state/RedisKeyValueState.java
kosii: Using guava's `Optional` seems like a good idea to you?

external/storm-redis/src/main/java/org/apache/storm/redis/state/RedisKeyValueState.java
kosii: if `pendingDeletePrepare` contains the key, then it was already removed from `pendingPrepare`, so it will return `null`

external/storm-redis/src/main/java/org/apache/storm/redis/state/RedisKeyValueState.java
kosii: Yes, you're right!

external/storm-redis/src/main/java/org/apache/storm/redis/state/RedisKeyValueState.java
kosii: Yes, you're right!

kosii:
A short list of modifications

- I went for the `Optional` class. Using your idea largely simplified the code.
- delete now returns the current value
- I didn't use a different set to store `pendingPrepare`, the deleted keys are represented with tombstones in the redis hash

external/storm-redis/src/main/java/org/apache/storm/redis/state/RedisKeyValueState.java
```diff
@@ -60,22 +64,23 @@ public RedisKeyValueState(String namespace) {
...
     }
 
-    public RedisKeyValueState(String namespace, JedisPoolConfig poolConfig, Serializer<K> keySerializer, Serializer<V> valueSerializer) {
+    public RedisKeyValueState(String namespace, JedisPoolConfig poolConfig, Serializer<K> keySerializer, Serializer<Optional<V>> valueSerializer) {
         this(namespace, JedisCommandsContainerBuilder.build(poolConfig), keySerializer, valueSerializer);
     }
 
     public RedisKeyValueState(String namespace, JedisCommandsInstanceContainer jedisContainer,
-                              Serializer<K> keySerializer, Serializer<V> valueSerializer) {
+                              Serializer<K> keySerializer, Serializer<Optional<V>> valueSerializer) {
```
arunmahadevan: It might be better to keep the Optional internal to the implementation.

external/storm-redis/src/main/java/org/apache/storm/redis/state/RedisKeyValueState.java
```diff
@@ -60,22 +64,23 @@ public RedisKeyValueState(String namespace) {
...
         base64Encoder = new BASE64Encoder();
         base64Decoder = new BASE64Decoder();
         this.namespace = namespace;
         this.prepareNamespace = namespace + "$prepare";
         this.txidNamespace = namespace + "$txid";
         this.keySerializer = keySerializer;
         this.valueSerializer = valueSerializer;
+        this.tombstone = encode(valueSerializer.serialize(Optional.<V>absent()));
```
arunmahadevan: Can `tombstone` be a static final constant ?

external/storm-redis/src/main/java/org/apache/storm/redis/state/RedisKeyValueState.java
```diff
@@ -161,7 +176,11 @@ public void prepareCommit(long txid) {
...
                 LOG.debug("Prepared txn already exists, will merge", txid);
-                pendingPrepare.putAll(pendingCommit);
+                for (Map.Entry<String, String> e: pendingCommit.entrySet()) {
+                    if (!pendingPrepare.containsKey(e.getKey())) {
+                        pendingPrepare.put(e.getKey(), e.getValue());
+                    }
+                }
             }
             if (!pendingPrepare.isEmpty()) {
                 commands.hmset(prepareNamespace, pendingPrepare);
```
arunmahadevan: A potential issue (already existing) which could be fixed here or in a separate JIRA. If the pendingPrepare is updated after the hmset but before its assigned to pendingCommit, the pendingCommit will not reflect the entities in prepareNamespace. I think if the pendingPrepare is snapshotted/copied before hmset and the copy is used for both hmset and assigned to pendingCommit it would avoid this race condition.

arunmahadevan:
@kosii I have gone through the latest changes and left a few comments. The idea of using tombstone looks nice and once the comments are addressed its good to merge.

external/storm-redis/src/main/java/org/apache/storm/redis/state/RedisKeyValueState.java
```diff
@@ -60,22 +64,23 @@ public RedisKeyValueState(String namespace) {
     }
 
     public RedisKeyValueState(String namespace, JedisPoolConfig poolConfig) {
-        this(namespace, poolConfig, new DefaultStateSerializer<K>(), new DefaultStateSerializer<V>());
+        this(namespace, poolConfig, new DefaultStateSerializer<K>(), new DefaultStateSerializer<Optional<V>>());
     }
 
-    public RedisKeyValueState(String namespace, JedisPoolConfig poolConfig, Serializer<K> keySerializer, Serializer<V> valueSerializer) {
+    public RedisKeyValueState(String namespace, JedisPoolConfig poolConfig, Serializer<K> keySerializer, Serializer<Optional<V>> valueSerializer) {
```
kosii: by having a `private final Serializer<Optional<V>> internalValueSerializer`?

external/storm-redis/src/main/java/org/apache/storm/redis/state/RedisKeyValueState.java
kosii: I'm not sure, but I don't think so, as the `internalValueSerializer` would depend on the `V` type parameter

external/storm-redis/src/main/java/org/apache/storm/redis/state/RedisKeyValueState.java
kosii: I'd fix it in a new JIRA, that way it might be released sooner

external/storm-redis/src/main/java/org/apache/storm/redis/state/RedisKeyValueState.java
arunmahadevan: Sounds good.  If I understood correctly, the user provided serializer would be used to serialize the value to bytes which is then wrapped in an Optional<byte[]> and serialized via the DefaultStateSerializer.

external/storm-redis/src/main/java/org/apache/storm/redis/state/RedisKeyValueState.java
arunmahadevan: Internally absent seems to return a static final Absent<Object> cast to Optional<T>, so the type does not really matter. Its not major, but the tombstone could be a constant reference to Optional.<Object>absent().The current one is also fine.

external/storm-redis/src/main/java/org/apache/storm/redis/state/RedisKeyValueState.java
arunmahadevan: Sure

HeartSaVioR:
@kosii Any update on this?

kosii:
@HeartSaVioR I addressed most of the comments in my local repo, but not pushed yet. If I have some time this week (probably Friday) I'll clean it up, and commit. Sorry for the delay

arunmahadevan:
@kosii can you up-merge and push your local repo changes so that we can merge this?

kosii:
@arunmahadevan I'm very sorry for disappearing for such a long time. I addressed all the comments, I also fixed the race condition in my last commit

arunmahadevan:
@kosii thanks for working on this. Will take a look one final time and merge it if it looks fine.

arunmahadevan:
+1

arunmahadevan:
@kosii thanks for the PR. Merged it to master. 

Can you also raise a PR for 1.x-branch ? I am getting some merge conflicts while trying to cherry-pick.

kosii:
Yes, I'll look into that!

kosii:
@arunmahadevan actually my patch depends on #2020, which isn't yet in 1.x-branch. how should we proceed?

arunmahadevan:
@kosii can you cherry pick the commits from 1606 plus your commits and raise a pr for 1.x branch?

kosii:
@arunmahadevan yes, check #1811 please

arunmahadevan:
Thanks @kosii Merged to master and 1.x-branch. You can close the PR.

PR #1474, opened 2016-06-08
support for wildcard topics ( storm-kafka-client )

nabhanelrahman:
these changes allow you to define a KafkaSpout that subscribes to a wildcard of topics.  For example clickstreams.*.log.  This is already supported in external/storm-kafka ( 0.8.x ); adding this feature to external/storm-kafka-client ( 0.9.x ).

+ Added a new public constructer to KafkaSpoutStream that takes in a java.util.regex.Pattern.  So instead of creating a KakfaSpoutStreams.  You can create a single KafkaSpoutStream that takes in a wildcard.  This is because a kafkaConsumer can only subscribe to a single regex pattern.

+ Added a test called KafkaSpoutWildcardTopologyMain.java that demonstrates how to setup a KakfaSpout that subscribes to a wildcard of topics.

### DIFF
--- a/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
+++ b/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
@@ -72,8 +72,10 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
     private transient boolean initialized;                              // Flag indicating that the spout is still undergoing initialization process.
     // Initialization is only complete after the first call to  KafkaSpoutConsumerRebalanceListener.onPartitionsAssigned()
 
+    private KafkaSpoutStream kafkaSpoutStream;
     private KafkaSpoutStreams kafkaSpoutStreams;                        // Object that wraps all the logic to declare output fields and emit tuples
     private transient KafkaSpoutTuplesBuilder<K, V> tuplesBuilder;      // Object that contains the logic to build tuples for each ConsumerRecord
+    private transient KafkaSpoutTupleBuilder<K, V> tupleBuilder; 
 
     private transient Map<TopicPartition, OffsetEntry> acked;           // Tuples that were successfully acked. These tuples will be committed periodically when the commit timer expires, after consumer rebalance, or on close/deactivate
     private transient Set<KafkaSpoutMessageId> emitted;                 // Tuples that have been emitted but that are "on the wire", i.e. pending being acked or failed
@@ -84,6 +86,7 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
     public KafkaSpout(KafkaSpoutConfig<K, V> kafkaSpoutConfig) {
         this.kafkaSpoutConfig = kafkaSpoutConfig;                 // Pass in configuration
         this.kafkaSpoutStreams = kafkaSpoutConfig.getKafkaSpoutStreams();
+        this.kafkaSpoutStream = kafkaSpoutConfig.getKafkaSpoutStream();
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -72,8 +72,10 @@
     private transient boolean initialized;                              // Flag indicating that the spout is still undergoing initialization process.
     // Initialization is only complete after the first call to  KafkaSpoutConsumerRebalanceListener.onPartitionsAssigned()
 
+    private KafkaSpoutStream kafkaSpoutStream;
     private KafkaSpoutStreams kafkaSpoutStreams;                        // Object that wraps all the logic to declare output fields and emit tuples
     private transient KafkaSpoutTuplesBuilder<K, V> tuplesBuilder;      // Object that contains the logic to build tuples for each ConsumerRecord
+    private transient KafkaSpoutTupleBuilder<K, V> tupleBuilder;
```
harshach: can't we just tuplesBuilder for this . Any reason to have another tupleBuilder.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -72,8 +72,10 @@
     private transient boolean initialized;                              // Flag indicating that the spout is still undergoing initialization process.
     // Initialization is only complete after the first call to  KafkaSpoutConsumerRebalanceListener.onPartitionsAssigned()
 
+    private KafkaSpoutStream kafkaSpoutStream;
```
harshach: any reason to not handle this using kafkaSpoutStreams

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
nabhanelrahman: the subscribe method() https://kafka.apache.org/090/javadoc/org/apache/kafka/clients/consumer/KafkaConsumer.html#subscribe(java.util.regex.Pattern,%20org.apache.kafka.clients.consumer.ConsumerRebalanceListener)

only supports one regex pattern.  TuplesBuilder is used to define multiple topics, which is not applicable for this case.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
nabhanelrahman: KafkaSpoutStreams is a collection of KafkaSpoutStream because you can have multiple topics for the kafkaconsumer.subscribe(List<String topics... )

That is not the case of kafkaconsumer.subscribe(Pattern ).  You can only have one pattern per Spout, so it doesn't make sense to put in Streams instead of Stream

nabhanelrahman:
@harshach, just reaching out on the status of this pull request?  Does it look good?

harshach:
@nabhanelrahman I still think naming of those variables very confusing and doesn't tell the intended purpose.
@hmcl  can you take a look at the pr.

hmcl:
@harshach @nabhanelrahman looking at this now. Somehow I missed this PR earlier when going throw the PRs for reviewing.

nabhanelrahman:
@harshach I can rename the variables:

private KafkaSpoutStream kafkaWildcardSpoutStream or kafkaPatternSpoutStream

and

private transient KafkaSpoutTupleBuilder<K, V> tupleWildcardBuilder or tuplePatternBuilder

what do you think?

hmcl:
@nabhanelrahman your patch enlightened that it would be better to add the `emit` and `getOutputFields` methods to `KafkaSpoutStream`. Furthermore, there were a couple of design decisions that would be ideal to do at this point in order to reuse code, and keep it clean and modular. Since it would be too verbose to go over those discussions in here, I just created a pull request with the changes that I believe are adequate to address the wildcard topic pattern requirement. This is the pull request: https://github.com/apache/storm/pull/1516

Although I had this in my todo list, since you already worked on it, it's totally fine for you to get the credit for it. Hence feel free to assign the JIRA to you, and recommit the PR with your username. Please mention that we worked on this together, and it's all good. Thanks for your understanding.

nabhanelrahman:
@hmcl, I integrated pull #1516 and tested it, and it looks good.  The class naming looks a lot more intuitive.  I resubmitted pull #1516 as #1518.  I only have read-only permission in JIRA ( https://issues.apache.org/jira/browse/STORM-1930 ).  So I can't do any changes their.  My JIRA handle is "nabhan".  I am closing out the pull request since #1518 is now the one we want.

PR #1479, opened 2016-06-10
[STORM-1880] Support EXISTS Command Storm-Redis

darionyaphet:
[STORM-1880](https://issues.apache.org/jira/browse/STORM-1880)

add exists command in storm-redis LookupBolt

### DIFF
--- a/external/storm-redis/src/main/java/org/apache/storm/redis/common/mapper/RedisDataTypeDescription.java
+++ b/external/storm-redis/src/main/java/org/apache/storm/redis/common/mapper/RedisDataTypeDescription.java
@@ -23,7 +23,7 @@ import java.io.Serializable;
  * RedisDataTypeDescription defines data type and additional key if needed for lookup / store tuples.
  */
 public class RedisDataTypeDescription implements Serializable {
-    public enum RedisDataType { STRING, HASH, LIST, SET, SORTED_SET, HYPER_LOG_LOG, GEO }
+    public enum RedisDataType { STRING, HASH, LIST, SET, SORTED_SET, HYPER_LOG_LOG, GEO, EXISTS }
 
     private RedisDataType dataType;
     private String additionalKey;
--- a/external/storm-redis/src/main/java/org/apache/storm/redis/bolt/RedisLookupBolt.java
+++ b/external/storm-redis/src/main/java/org/apache/storm/redis/bolt/RedisLookupBolt.java
@@ -109,6 +109,10 @@ public class RedisLookupBolt extends AbstractRedisBolt {
                     lookupValue = jedisCommand.geopos(additionalKey, key);
                     break;
 
+                case EXISTS:
+                    lookupValue = jedisCommand.exists(key);
+                    break;
+
                 default:
                     throw new IllegalArgumentException("Cannot process such data type: " + dataType);
             }


external/storm-redis/src/main/java/org/apache/storm/redis/common/mapper/RedisDataTypeDescription.java
```diff
@@ -23,7 +23,7 @@
  * RedisDataTypeDescription defines data type and additional key if needed for lookup / store tuples.
  */
 public class RedisDataTypeDescription implements Serializable {
-    public enum RedisDataType { STRING, HASH, LIST, SET, SORTED_SET, HYPER_LOG_LOG, GEO }
+    public enum RedisDataType { STRING, HASH, LIST, SET, SORTED_SET, HYPER_LOG_LOG, GEO, EXISTS }
```
abhishekagarwal87: I am not fully aware of Redis terminology but EXISTS doesn't look like a data type. How do these data types relate to the redis operation. e.g. SORTED_SET and zscore?

darionyaphet: Yes . Maybe we should rename `RedisDataType` to `Command` it's more suitable .

abhishekagarwal87: would recommend that if RedisDataType is internal to storm code.

darionyaphet: `zscore` is a function about `Sorted Set`

darionyaphet: `RedisMapper` point out `RedisDataType` is hard to internal it into storm code ? I'm not sure ...

HeartSaVioR:
Hi @darionyaphet.
Thanks for the contribution.
Btw, RedisLookupBolt and RedisStoreBolt are intended to map tuple to Redis data or opposite (You could imagine object mapper or ORM), not intended to support operations. `EXISTS` seems not fit to intention of them.

darionyaphet:
@HeartSaVioR  

Yes  I have realized that `RedisDataType` is a series of data format .  But `EXISTS` is really heavily used in Redis usage . 

Should we split data and operation into difference bolt ?

HeartSaVioR:
@darionyaphet I'm curious that `exists` fits for use cases of Storm. Could you share your use cases?

darionyaphet:
@HeartSaVioR check `exists` is very useful .  Sometimes we will check if a key is in a key set to decide if the value should be process .

HeartSaVioR:
@darionyaphet 
Instead of having `exists` as datatype, we can have a flag that indicates checking only availability. It can be optimized for STRING and HASH type with exists and hexists, and also available anyway for other types (lookup and simply check that whether it's not null or null). 
Or we even have new filter Bolt that filters tuples by availability. Might be more clearer.
@darionyaphet @abhishekagarwal87 What do you think?

abhishekagarwal87:
+1 for a filter bolt.

darionyaphet:
filter bolt seems better

HeartSaVioR:
OK. Filed [STORM-1919](https://issues.apache.org/jira/browse/STORM-1919) for FilterBolt.
We can close this and mark STORM-1880 as "Won't fix".

@darionyaphet Since I have an idea in mind how to implement this, so if you don't mind I'll work on STORM-1919. Are you OK with it?

darionyaphet:
Sure . I will close it :)

PR #1480, opened 2016-06-10
STORM-1881: Create fat-jar of storm-redis which contains needed dependencies

sakanaou:
`storm-redis` was packaged in a way that it was not immediately usable when put into `extlib` folder. `ClassNotFoundException`s would occur.

Via `maven-shade-plugin` a "fat"-jar is created so all depending code is packaged in.

### DIFF
--- a/external/storm-redis/pom.xml
+++ b/external/storm-redis/pom.xml
@@ -59,17 +59,20 @@
         <dependency>
             <groupId>com.google.guava</groupId>
             <artifactId>guava</artifactId>
-            <version>18.0</version>
+            <version>${guava.version}</version>
+            <scope>provided</scope>
         </dependency>
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-core</artifactId>
             <version>${jackson.version}</version>
+            <scope>provided</scope>
         </dependency>
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
             <version>${jackson.version}</version>
+            <scope>provided</scope>
         </dependency>
         <!--test dependencies -->
         <dependency>
@@ -83,4 +86,22 @@
             <scope>test</scope>
         </dependency>
     </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-shade-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <phase>package</phase>
+                        <goals>
+                            <goal>shade</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
+        </plugins>
+    </build>
+
 </project>


abhishekagarwal87:
just curious, how much is the size of fat jar?

sakanaou:
@abhishekagarwal87: Original jar is 65K, shaded jar is 709K. guava and jackson are already contained in storm-core so those can be ignored, luckily.

abhishekagarwal87:
Thanks @sakanaou

HeartSaVioR:
We're shading the libraries which you're changing scope to `provided` in storm-core, so it will not be provided at runtime.
Since storm-redis depends on Guava, placing uber storm-redis jar to extlib incurs dependency issue in whole topologies in cluster. If we still really want to place this to extlib, we may need to consider shading guava and others for storm-redis module.

sakanaou:
@HeartSaVioR: `storm-core/pom.xml` includes a `maven-shade-plugin` step that lists both guava and jackson as artifactSet-includes without relocations. Why should they then not be provided via storm-core? I have successfully used storm-redis without placing guava or jackson explicitly in the classpath so they must be present already. From my understanding they are provided by the "fat" storm-core jar.

abhishekagarwal87:
guava dependency is relocated in storm-core.
https://github.com/apache/storm/blob/2a435b0289d43494eb2f834e94ed61c175f9eea6/storm-core/pom.xml#L593

sakanaou:
@abhishekagarwal87: Ok, didn't spot that. Was only looking for "guava" relocations.

So, the dependency could be additionally included relocated into storm-redis. Or is it possible to use the already relocated package which is provided via storm-core? This would save some resources. I'm not very experienced with the `maven-shade-plugin` yet, unfortunately.

What is preferred?

abhishekagarwal87:
If you use already relocated classes, you make storm-redis module dependent on a particular version of storm-core. relocation pattern is subject to changes in storm-core (since it is internal) and if that happens, your module stops being compatible with storm-core. So you should avoid using already relocated classes in storm-core.

HeartSaVioR:
Let me have a look. I guess storm-redis doesn't need to depend on other than storm-core and Jedis.

HeartSaVioR:
Getting rid of Guava seems easy since storm-redis uses Precondition, ImmutableMap, Maps, Lists.
Getting rid of Jackson seems not easy since we take StateConfig to JSON format so we need JSON parser anyway.

It might need to explain current status and discuss how to resolve it from dev@.

HeartSaVioR:
Sorry for long delay, say, 1 month.
I've met similar issue on Storm SQL, so I've initiate discussion to dev@ list.
Along with discussion, I'm trying to see how Spark handles dependencies and adding them to executor. I guess I understand how it works, but I didn't play with how to resolve transitive dependencies and download them.

HeartSaVioR:
STORM-2016 is on reviewing, and it should help this get resolved.

HeartSaVioR:
@sakanaou 
Since STORM-2016 is merged into master and 1.x-branch, so I think we can close this. 
Could you please doing it? Thanks in advance!

sakanaou:
Closed as requested due to STORM-2016.

PR #1481, opened 2016-06-10
STORM-1705: Fix cap-retry bug

abhishekagarwal87:


ptgoetz:
+1 Thanks @abhishekagarwal87.

Since this is a one-line change blocking the release, I'm going to merge it in the interest of proceeding with the release. If there are any objections we can cancel the next vote and revert the change.

abhishekagarwal87:
Adding @harshach - do you want #1482 (STORM-1876) in 1.0.2?

abhishekagarwal87:
Thinking further on it, STORM-1876 is needed when building from the source. So it should be fine if not included in the release candidate. 
@ptgoetz  - +1 for merging this now and going ahead with release.

PR #1483, opened 2016-06-11
STORM-1875 : Separate Jedis/JedisCluster Config

darionyaphet:
[STORM-1875](https://issues.apache.org/jira/browse/STORM-1875)  Separate Jedis / JedisCluster to provide full operations for each environment to users .

Reference to : #1459 

Separated existing classes to `RedisLookupBolt`  `RedisClusterLookupBolt` `RedisStoreBolt` `RedisClusterStoreBolt`


HeartSaVioR:
It breaks backward compatibility so we can't go on this. 
And in fact, separation of Jedis and JedisCluster is only needed when users want to use multiple keys operation, which is not applied to RedisLookupBolt and RedisStoreBolt.
So it might be better to go with dirty way like I said later.

I'm open to better ideas without breaking backward compatibility.

PR #1486, opened 2016-06-14
STORM-1890: ensure static UI resources are re-requested with differen…

abellina:
…t Storm package builds

Note it doesn't perform cache busting for logviewer. That is creating its own HTML, where I think it can be refactored to use some static file in public/, that is a separate task.

Thanks! I'll backport to 1.x if this looks good.


abellina:
Build errors are unrelated to change.

kishorvpatil:
LGTM. +1

abellina:
I am going to close this and will re-open with some changes to add special not-caching headers for files like index.html or topology.html. While testing this for the 1.x branch I realized those files have the same issue, but you can't use cache busting there. So we need to ensure ring adds the no-cache headers for .html files that we serve.

PR #1487, opened 2016-06-14
off-by-one error for UNCOMMITTED_LATEST and UNCOMMITTED_EARLIEST

nabhanelrahman:
In the source code documentation, it says for both UNCOMMITTED_LATEST
and UNCOMMITTED_EARLIEST it will fetch the last committed offset ( if it
exists ).  But the code fetches last committed + 1;

If a topology is killed/crashes and then is redeployed, this off-by-one
error causes one message per partition to be dropped.

### DIFF
--- a/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
+++ b/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
@@ -167,7 +167,7 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
                     fetchOffset = kafkaConsumer.position(tp);
                 } else {
                     // By default polling starts at the last committed offset. +1 to point fetch to the first uncommitted offset.
-                    fetchOffset = committedOffset.offset() + 1;
+                    fetchOffset = committedOffset.offset();
                     kafkaConsumer.seek(tp, fetchOffset);
                 }
             } else {    // no commits have ever been done, so start at the beginning or end depending on the strategy


nabhanelrahman:
@srdo, 

i have this property set:

 props.put(KafkaSpoutConfig.Consumer.ENABLE_AUTO_COMMIT, "true");

Here are my steps:

1) create 10 partition topic
2) deploy topology
3) send 10 msgs
4) confirmed 10 msgs where received
5) kill topology
6) send 100 msgs
7) re-deploy topology
8) confirmed that 90 msgs were received ( was expecting 100 msgs).  that means the first message in each partition was skipped when the initialized was called.

srdo:
You probably shouldn't enable auto commit if you can't accept some message loss. As I understand it, it makes the Kafka consumer commit the messages it has received from poll() every so often. That means that messages can be lost if they've been pulled out of the consumer, but the topology hasn't acked them yet. If you disable auto commit, the messages will instead be committed when the topology acks them.

I can't reproduce this issue. Here are the steps I took:

1) Do a fresh install of Kafka 0.9.0.1
2) Do a fresh install of Storm 1.0.1, start Nimbus and 1 supervisor on localhost
3) Create a 1 partition topic on a newly installed 0.9.0.1 Kafka broker
4) Deploy a topology containing only the Kafka spout and a bolt printing the messages it receives to system.out.
5) Send 5 messages to Kafka using the kafka-console-producer.sh script in kafka/bin
6) Verify that the topology prints those messages
7) Kill the topology and wait for 30 seconds for it to be completely shut down
8) Send 5 more messages
9) Redeploy
10) Confirm that all 5 new messages are received

Here's the code I used https://github.com/srdo/TestKafkaSpout

nabhanelrahman:
I am about to reproduce the issue with auto-commit=true (i lose messages).  How do we reconcile auto-commit not working?  Do we hardcode auto-commit=false when creating the KafkaConsumer object for KafkaSpout?

srdo:
Could you check also if you see the issue when auto commit is set to false?

Are you running your topology in local mode or deploying it with storm jar?

hmcl:
@nabhanelrahman I am addressing this and other edge cases in a patch that I am about to submit. I am afraid that just removing the +1 won't work for all the cases.

nabhanelrahman:
@hmcl,  i am closing out this pull request because I am assuming that you will be addressing this issue in your patch.  

In the meantime, do you think I will not have data loss if auto-commit is set to false?

PR #1490, opened 2016-06-16
STORM-1902 (1.x): add a simple & flexible FileNameFormat for storm-hdfs

vesense:
1. flexible. not fixed as `{prefix}{componentId}-{taskId}-{rotationNum}-{timestamp}{extension}`
2. support parameters. such as `$TIME` `$NUM` `$HOST` etc.

### DIFF
--- a/external/storm-hdfs/README.md
+++ b/external/storm-hdfs/README.md
@@ -158,8 +158,20 @@ For example:
 
 By default, prefix is empty and extenstion is ".txt".
 
+**New FileNameFormat:**
 
-The provided `org.apache.storm.hdfs.format.SimpleFileNameFormat` is flexible, supports parameters such as `$TIME` `$NUM` `$HOST` etc.
+The new provided `org.apache.storm.hdfs.format.SimpleFileNameFormat` and `org.apache.storm.hdfs.trident.format.SimpleFileNameFormat` are more flexible, and the `withName` method support parameters as following:
+
+* $TIME - current time. use `withTimeFormat` to format.
+* $NUM - rotation number
+* $HOST - local host name
+* $PARTITION - partition index (`org.apache.storm.hdfs.trident.format.SimpleFileNameFormat` only)
+* $COMPONENT - component id (`org.apache.storm.hdfs.format.SimpleFileNameFormat` only)
+* $TASK - task id (`org.apache.storm.hdfs.format.SimpleFileNameFormat` only)
+
+eg: `seq.$TIME.$HOST.$COMPONENT.$NUM.dat`
+
+The default file `name` is `$TIME.$NUM.txt`, and the default `timeFormat` is `yyyyMMddHHmmss`.
 
 
 ### Sync Policies
--- a/external/storm-hdfs/README.md
+++ b/external/storm-hdfs/README.md
@@ -159,6 +159,8 @@ For example:
 By default, prefix is empty and extenstion is ".txt".
 
 
+The provided `org.apache.storm.hdfs.format.SimpleFileNameFormat` is flexible, supports parameters such as `$TIME` `$NUM` `$HOST` etc.
+
 
 ### Sync Policies
 Sync policies allow you to control when buffered data is flushed to the underlying filesystem (thus making it available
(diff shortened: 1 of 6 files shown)


dossett:
Some unit tests would be very helpful, especially to verify the substitution logic.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/format/SimpleFileNameFormat.java
```diff
@@ -0,0 +1,96 @@
...
+     *            file name
+     * @return
+     */
+    public SimpleFileNameFormat withName(String name) {
+        this.name = name;
+        return this;
+    }
+
+    public SimpleFileNameFormat withTimeFormat(String timeFormat) {
+        this.timeFormat = timeFormat;
```
dossett: How about constructing a SimpleDateFormat here to validate the timeFormat string?  If it's an invalid string, better to catch it here at setup time.

vesense: Agree with you. Done.

vesense:
Thanks @dossett for your review. Updated.

external/storm-hdfs/src/test/java/org/apache/storm/hdfs/bolt/format/TestSimpleFileNameFormat.java
```diff
@@ -0,0 +1,77 @@
...
+    @Test
+    public void testDefaults() {
+        SimpleFileNameFormat format = new SimpleFileNameFormat();
+		format.prepare(null, createTopologyContext());
+        String path = format.getPath();
+        String name = format.getName(1, System.currentTimeMillis());
+
+        Assert.assertEquals("/storm", path);
+        String now = new SimpleDateFormat("yyyyMMddHHmmss").format(System.currentTimeMillis());
+        Assert.assertEquals(now+".1.txt", name);
```
dossett: This is a nit, but this test might occasionally fail since the time is taken at two different points which might resolve to different seconds.

external/storm-hdfs/src/test/java/org/apache/storm/hdfs/bolt/format/TestSimpleFileNameFormat.java
```diff
@@ -0,0 +1,77 @@
...
+        SimpleFileNameFormat format = new SimpleFileNameFormat()
+            .withName("$TIME.$HOST.$COMPONENT.$TASK.$NUM.txt")
+            .withPath("/mypath")
+            .withTimeFormat("yyyy-MM-dd HH:mm:ss");
+        format.prepare(null, createTopologyContext());
+        String path = format.getPath();
+        String name = format.getName(1, System.currentTimeMillis());
+
+        Assert.assertEquals("/mypath", path);
+        String now = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss").format(System.currentTimeMillis());
```
dossett: Same comment as above.

dossett:
Two minor nits noted inline, but I am otherwise +1.  Test failures look unrelated.  

This is a nice addition.

external/storm-hdfs/src/test/java/org/apache/storm/hdfs/bolt/format/TestSimpleFileNameFormat.java
vesense: nice catch. fixed.

external/storm-hdfs/src/test/java/org/apache/storm/hdfs/bolt/format/TestSimpleFileNameFormat.java
vesense: Done.

vesense:
@dossett  Updated and rebased.

dossett:
+1

external/storm-hdfs/README.md
```diff
@@ -159,6 +159,8 @@ For example:
 By default, prefix is empty and extenstion is ".txt".
 
 
+The provided `org.apache.storm.hdfs.format.SimpleFileNameFormat` is flexible, supports parameters such as `$TIME` `$NUM` `$HOST` etc.
```
HeartSaVioR: It may be better to elaborate how to use so that users don't need to look at javadoc.

HeartSaVioR:
The change looks good. I just would like to add kind explanation to letting users know and use new feature.

HeartSaVioR:
+1 Thanks for following up so quickly.

HeartSaVioR:
Close commit message not against master seems not work.
@vesense PR is merged to 1.x-branch. Could you please close this? Thanks in advance!

PR #1491, opened 2016-06-16
STORM-1907: PartitionedTridentSpoutExecutor has incompatible types that cause ClassCastException

hmcl:
- Generify on Object rather than Integer because example is broken
 - Fix TridentWordCount example to print more meaningful output


hmcl:
This PR is related to https://github.com/apache/storm/pull/1429 but I found this issue independently, in my local branch, which was a bit outdated. This patch still makes sense since it gets rid of five warnings that are caused by having ITridentSpout.BatchCoordinator as raw type. Furthermore, it also prints more the actual word count output in the example in file `TridentWordCount`

harshach:
+1. Build failure is not related to the patch.

PR #1493, opened 2016-06-16
[STORM-1908] Support Storm HBase ZooKeeper Config

darionyaphet:
[STORM-1908](https://issues.apache.org/jira/browse/STORM-1908)

Support Storm HBase ZooKeeper Config . About ZooKeeper host , port and parent .

### DIFF
--- a/external/storm-hbase/src/main/java/org/apache/storm/hbase/bolt/AbstractHBaseBolt.java
+++ b/external/storm-hbase/src/main/java/org/apache/storm/hbase/bolt/AbstractHBaseBolt.java
@@ -32,7 +32,6 @@ import org.slf4j.LoggerFactory;
 import java.util.HashMap;
 import java.util.Map;
 
-// TODO support more configuration options, for now we're defaulting to the hbase-*.xml files found on the classpath
 public abstract class AbstractHBaseBolt extends BaseRichBolt {
     private static final Logger LOG = LoggerFactory.getLogger(AbstractHBaseBolt.class);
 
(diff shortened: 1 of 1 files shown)


HeartSaVioR:
Since we override all options being provided from HBase config, these three parameters can be fed as well. I think just removing TODO would be fine.

darionyaphet:
@HariSekhon what you mean is should remove `TODO` when all HBase options have provided ?

HariSekhon:
@darionyaphet wrong @ person?

darionyaphet:
@HariSekhon Yep  So sorry ~

HeartSaVioR:
@darionyaphet 
I mean we already support the thing what TODO describes. 
Please refer https://github.com/apache/storm/blob/master/external/storm-hbase/src/main/java/org/apache/storm/hbase/bolt/AbstractHBaseBolt.java#L66-L68

So this issue is invalid, and TODO can be removed.

darionyaphet:
@HeartSaVioR  I got we can set config using `Config` and don't have to support new API ?

HeartSaVioR:
@darionyaphet 
You can set properties which we can pass to hbase-*.xml into Storm configuration Map (under specific key for hbase), and storm-hbase merges (override) that configuration with default HBase configuration which reads hbase-*.xml to initialize. So passing them to Storm configuration Map before submitting topology would be sufficient.

darionyaphet:
@HeartSaVioR Yes I know . Because Zookeeper properties is very useful , so I wish to open a new API to support it :) . If you think it's unnecessarily , I will close this PR .

HeartSaVioR:
@darionyaphet Let's close for now. We can enrich description to README how to pass ZK properties when we think it would be better to guide passing ZK properties.

darionyaphet:
@HeartSaVioR   OK  Close it :)

PR #1498, opened 2016-06-17
STORM-1911 IClusterMetricsConsumer should use seconds to timestamp unit

HeartSaVioR:
* to have consistency with IMetricsConsumer

PR for 1.x: #1499


harshach:
+1

PR #1500, opened 2016-06-17
STORM-1913: Additions and Improvements for Trident RAS API

knusbaum:
New behavior: 
- Added a way to specify default values for operations in Trident
- New behavior requires all components in group have same types of resources set (or defaults)
- Added a way to specify resources for master-coord Spouts.

### DIFF
--- a/storm-core/src/jvm/org/apache/storm/topology/BaseConfigurationDeclarer.java
+++ b/storm-core/src/jvm/org/apache/storm/topology/BaseConfigurationDeclarer.java
@@ -47,7 +47,7 @@ public abstract class BaseConfigurationDeclarer<T extends ComponentConfiguration
         if(val!=null) val = val.intValue();
         return addConfiguration(Config.TOPOLOGY_MAX_SPOUT_PENDING, val);
     }
-    
+
     @Override
     public T setNumTasks(Number val) {
         if (val != null) val = val.intValue();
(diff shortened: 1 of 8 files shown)


storm-core/src/jvm/org/apache/storm/topology/BaseConfigurationDeclarer.java
```diff
@@ -56,7 +56,11 @@ public T setNumTasks(Number val) {
 
     @Override
     public T setMemoryLoad(Number onHeap) {
-        return setMemoryLoad(onHeap, Utils.getDouble(conf.get(Config.TOPOLOGY_COMPONENT_RESOURCES_OFFHEAP_MEMORY_MB)));
+        if (onHeap != null) {
+            onHeap = onHeap.doubleValue();
+            return addConfiguration(Config.TOPOLOGY_COMPONENT_RESOURCES_ONHEAP_MEMORY_MB, onHeap);
+        }
+        return null;
```
HeartSaVioR: Minor: This can be replaced with `setMemoryLoad(onHeap, null)` but I'm OK to split two methods completely.

knusbaum: Maybe it would make more sense for `setMemoryLoad(Number onHeap, Number offHeap)` to call into this one instead.

HeartSaVioR: Yes that might be better.

knusbaum: @HeartSaVioR fixed.

d2r:
Overall it looks OK, but it seems there are some issues with the build.

d2r:
OK, looks like one of the tests is failing a check for sane resource assignments per-stream.

d2r:
+1 looks good to me.

@HeartSaVioR any other comments?

HeartSaVioR:
@d2r Nope, since I don't have background on RAS implementation yet. Please go ahead.

HeartSaVioR:
@knusbaum Please rebase and squash the commits before merging. It would be better to have meaningful log message.

PR #1502, opened 2016-06-20
STORM-1633: Document blobstore to command-line-client.md

vesense:


HeartSaVioR:
+1
You already resolved what I pointed out, and this is even better. :)

PR #1513, opened 2016-06-23
STORM-1925 Remove Nimbus thrift call from Nimbus itself

HeartSaVioR:
PR for 1.x: #1514

* Nimbus fails to start in secure mode since authrization is not passed

Extract the logic of getClusterInfo to new function 'get-cluster-info' and let both of callers call this function instead.


HeartSaVioR:
Confirmed that before patch Nimbus has crashed in kerberized environment, and this patch fixes that issue.

harshach:
+1

PR #1514, opened 2016-06-23
STORM-1925 Remove Nimbus thrift call from Nimbus itself (1.x)

HeartSaVioR:
PR for master: #1513 

* Nimbus fails to start in secure mode since authrization is not passed
* Extract the logic of getClusterInfo to new function 'get-cluster-info'
  * and let both of callers call this function instead


harshach:
+1

PR #1515, opened 2016-06-23
[STORM-1929] Check when create topology

darionyaphet:
[STORM-1929](https://issues.apache.org/jira/browse/STORM-1929)

Add some check when create topology :

1. Spout and Bolt id shouldn't conflict

2. createTopology's spout and bolt set shouldn't empty .


storm-core/src/jvm/org/apache/storm/topology/TopologyBuilder.java
```diff
@@ -113,6 +113,15 @@ public StormTopology createTopology() {
         Map<String, Bolt> boltSpecs = new HashMap<>();
         Map<String, SpoutSpec> spoutSpecs = new HashMap<>();
         maybeAddCheckpointSpout();
+
+        if (_spouts.size() == 0) {
+            throw new IllegalArgumentException("Spouts is not set");
+        }
+
+        if (_bolts.size() == 0) {
```
revans2: We have topologies that only have spouts (no bolts).  I know it is kind of an odd corner case where someone is just reading from an input and writing to an output in one shot, but it is something that we have.

storm-core/src/jvm/org/apache/storm/topology/TopologyBuilder.java
```diff
@@ -179,8 +188,13 @@ public BoltDeclarer setBolt(String id, IRichBolt bolt) throws IllegalArgumentExc
     public BoltDeclarer setBolt(String id, IRichBolt bolt, Number parallelism_hint) throws IllegalArgumentException {
         validateUnusedId(id);
         initCommon(id, bolt, parallelism_hint);
-        _bolts.put(id, bolt);
-        return new BoltGetter(id);
+
+        if (_bolts.containsKey(id)) {
+            throw new IllegalArgumentException("Bolt " + id + " have existed");
```
revans2: Can we say `"duplicate bolt id found " + id` instead?  I think it sounds a bit cleaner.

storm-core/src/jvm/org/apache/storm/topology/TopologyBuilder.java
```diff
@@ -339,8 +353,13 @@ public SpoutDeclarer setSpout(String id, IRichSpout spout) throws IllegalArgumen
     public SpoutDeclarer setSpout(String id, IRichSpout spout, Number parallelism_hint) throws IllegalArgumentException {
         validateUnusedId(id);
         initCommon(id, spout, parallelism_hint);
-        _spouts.put(id, spout);
-        return new SpoutGetter(id);
+
+        if (_spouts.containsKey(id)) {
+            throw new IllegalArgumentException("Spout " + id + " have existed");
```
revans2: Similar here `"duplicate spout id found " + id`

storm-core/src/jvm/org/apache/storm/topology/TopologyBuilder.java
darionyaphet: only have spout and don't have bolt is a strange use case . In my mine spout should just read message into topology , any other operation should move into bolts . 

I have remove the check of `_bolts.size()` .

PR #1517, opened 2016-06-24
STORM-1919 Introduce FilterBolt on storm-redis

HeartSaVioR:
* introduce RedisFilterBolt, and relevant class (RedisFilterMapper)
* add example topology: WhitelistWordCount
* update how to use to README.md
* also correct some javadocs

It should be easy to backport to 1.x-branch as well so I didn't create another pull request against 1.x-branch.


HeartSaVioR:
@darionyaphet Could you check if this bolt covers your origin intention? Thanks!

vesense:
LGTM

external/storm-redis/README.md
```diff
@@ -21,13 +21,16 @@ use it as a maven dependency:
...
+One tuple will be matched to one key / value pair, and you can define match pattern to ```TupleMapper```.
 
-These interfaces are combined with ```RedisLookupMapper``` and ```RedisStoreMapper``` which fit ```RedisLookupBolt``` and ```RedisStoreBolt``` respectively.
+You can also choose data type from ```RedisDataTypeDescription``` to use. Please refer ```RedisDataTypeDescription.RedisDataType``` to see what data types are supported. In some data types (hash and sorted set, and set if only RedisFilterBolt), it requires additional key and converted key from tuple becomes element.
+
+These interfaces are combined with ```RedisLookupMapper``` and ```RedisStoreMapper``` and ```RedisFilterMapper``` which fit ```RedisLookupBolt``` and ```RedisStoreBolt```, and ```RedisFilterBolt``` respectively.
+(When you want to implement RedisFilterMapper, be sure to set declareOutputFields() to declare same fields to input stream, since FilterBolt forwards input tuples when they exist on Redis.)
```
darionyaphet: Actually  `RedisFilterMapper` and `RedisLookupMapper` are the same . It get a key from tuple then use to query from `Redis` . The difference is `RedisFilterBolt` use `RedisFilterMapper` and `RedisLookupBolt` use `RedisLookupMapper` . Should the tow mappers could extends from one basic abstract class ? 

`RedisFilterMapper` and `RedisLookupBolt` will not use getValueFromTuple() , this method will always return `null`. So maybe I will open another ticket to discuss Mapper  Structure .

darionyaphet:
@HeartSaVioR Thank you for you Great Work . I have a small question about `Mapper` extention structure .

external/storm-redis/README.md
HeartSaVioR: @darionyaphet 
Yeah right. It can be arranged but we would want to resolve this with backward-compatible way. Maybe adding @Deprecated for 1.x / 1.0.x and removing for master.

external/storm-redis/src/main/java/org/apache/storm/redis/bolt/RedisFilterBolt.java
```diff
@@ -0,0 +1,146 @@
...
+        try {
+            jedisCommand = getInstance();
+
+            switch (dataType) {
+                case STRING:
+                    found = jedisCommand.exists(key);
+                    break;
+
+                case SET:
+                    if (additionalKey == null) {
```
abhishekagarwal87: since data type is known at the time of construction, can this check be moved to constructor itself? Failing fast is preferred.

external/storm-redis/src/main/java/org/apache/storm/redis/bolt/RedisFilterBolt.java
```diff
@@ -0,0 +1,146 @@
...
+                    found = jedisCommand.zrank(additionalKey, key) != null;
+                    break;
+
+                case HYPER_LOG_LOG:
+                    found = jedisCommand.pfcount(key) > 0;
+                    break;
+
+                case GEO:
+                    List<GeoCoordinate> geopos = jedisCommand.geopos(additionalKey, key);
+                    found = geopos != null && geopos.size() > 0;
```
abhishekagarwal87: minor. Better to wrap in parentheses for clarity
` found = (geopos != null && geopos.size() > 0)`

external/storm-redis/src/main/java/org/apache/storm/redis/bolt/RedisFilterBolt.java
HeartSaVioR: @abhishekagarwal87 Great point! Will fix.

external/storm-redis/src/main/java/org/apache/storm/redis/bolt/RedisFilterBolt.java
HeartSaVioR: Good point. Will fix.

external/storm-redis/src/test/java/org/apache/storm/redis/topology/WhitelistWordCount.java
```diff
@@ -0,0 +1,155 @@
...
+import org.apache.storm.tuple.Fields;
+import org.apache.storm.tuple.ITuple;
+import org.apache.storm.tuple.Tuple;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Map;
+import java.util.Random;
+
+public class WhitelistWordCount {
```
abhishekagarwal87: Let's have this in examples instead since there is no test in the class.

HeartSaVioR: Not sure where to move. Do you have an idea?
Maybe we would like to have a rule for this case, and apply all external modules since many modules already have example topologies in test package.

external/storm-redis/src/main/java/org/apache/storm/redis/bolt/RedisFilterBolt.java
HeartSaVioR: Addressed.

external/storm-redis/src/main/java/org/apache/storm/redis/bolt/RedisFilterBolt.java
HeartSaVioR: Addressed.

external/storm-redis/src/test/java/org/apache/storm/redis/topology/WhitelistWordCount.java
abhishekagarwal87: They do. One reason I know of is the dependency collision. But test doesn't seem the right folder, because a user is not going to look in the test folder. We can do the following - 
1. Module level sub-directories within examples e.g. examples/storm-redis, examples/storm-opentsdb etc.
2. examples file within the module which are not shipped with jar. The link to such examples can be put up in the documentation so that users can refer to the example code and copy/run it accordingly. 
what do you say?

HeartSaVioR: @abhishekagarwal87 
I'm OK for that, but I'd prefer to get this out of scope for pull request.
This needs discussion to have a generic rule across all external modules rather than fix for only this module.
Could you initiate the discussion to dev@ or file an issue? Thanks in advance!

abhishekagarwal87: @HeartSaVioR That's fine with me. I have initiated the discussion. 
+1 for the PR.

PR #1520, opened 2016-06-26
[STORM-1927] Upgrade Jetty and Ring

darionyaphet:
[STORM-1927](https://issues.apache.org/jira/browse/STORM-1927)

Jetty 7 is EOL , upgrade to Jetty 9 & Ring could also support it.


harshach:
+1

PR #1521, opened 2016-06-26
STORM-1901: Avro Integration for Storm-Kafka

vesense:
init version for code review, changes are following:
- separate Avro from storm-hdfs as a new subproject. (can used by storm-hdfs & storm-kafka, or others modules)
- add new `Config.TOPOLOGY_AVRO_SERIALIZER` in Config
- add new `Config.TOPOLOGY_AVRO_CONFLUENT_SCHEMA_REGISTRY_URL` in Config
- add `DirectAvroSerializer` and `DefaultDirectAvroSerializer` in storm-avro
- add new `AvroSchema` in storm-kafka
- add new `DirectAvroTupleToKafkaMapper` in storm-kafka
- new unit tests
- document improvements


vesense:
@dossett I add you as the Committer Sponsors because you have do a lot in storm avro. Hope you have chance to take a look.

vesense:
@ptgoetz @HeartSaVioR Could you also take a look?

dossett:
Thanks @vesense I will review this week.

harshach:
@vesense is this intended to be a schema registry, i.e users publishing their schemas IMO having schema registry again in storm is not a good idea.
By looking at the code seems to be a module containing avro related modules and mostly related to storm-kafka module. From what I see this is a AvroSerializer and AvroDeserializer if so shouldn't these part of storm-kafka module instead of being its own external module. Trying to understand where its applicable except the storm-kafka module at this point.

external/storm-kafka/README.md
```diff
@@ -360,6 +361,44 @@ For Trident:
...
 
+## Avro Integration for Storm-Kafka
+
+To integrate with Avro you **must** register the appropriate Kryo serializers with your topology configuration.  A convenience method is provided for this:
+
+`AvroUtils.addAvroKryoSerializations(conf);`
+
+###AvroSchema
+
+`AvroSchema` is an implementation of `Schema`. You can use AvroSchema to read Avro GenericRecord from Kafka:
```
dossett: Should be `AvroScheme` and `Scheme`

external/storm-kafka/src/jvm/org/apache/storm/kafka/AvroScheme.java
```diff
@@ -0,0 +1,59 @@
...
+import org.apache.storm.avro.DefaultDirectAvroSerializer;
+import org.apache.storm.avro.DirectAvroSerializer;
+import org.apache.storm.spout.Scheme;
+import org.apache.storm.tuple.Fields;
+import org.apache.storm.tuple.Values;
+import org.apache.storm.utils.Utils;
+
+public class AvroScheme implements Scheme {
+    public static final String AVRO_SCHEME_KEY = "avro";
+    DirectAvroSerializer serializer = new DefaultDirectAvroSerializer();
```
dossett: Could this be more flexible, the way AbstractAvroSerializer is above?  I would see creating an AvroSchema and supplying a serializer for it to use based on how the avro data is being put onto the Kafka topic.

My original use case was to read avro data that had been serialized onto a kafka topic with a Confluent serializer.  My solution was to create a bolt that that read the raw data with rawscheme, deserialized the raw bytes into a GenericRecord, and then emitted the generic record.

Something like this would have been very nice if I could give it a ConfluentDeserializer via dependency injection or something similar.

dossett:
@harshach My interpretation of the PR is that it moves some Avro speficif code I added to storm-hdfs into a more sharable module.

@vesense I like the idea of making it easier to read/write avro data in a Storm topology, thanks for looking at this.  I don't know that it needs to be its own module though, maybe it could be somewhere in `org.apache.storm.utils` ?

external/storm-kafka/README.md
vesense: Thx. Will fix.

external/storm-kafka/src/jvm/org/apache/storm/kafka/AvroScheme.java
vesense: >Could this be more flexible, the way AbstractAvroSerializer is above? I would see creating an AvroSchema and supplying a serializer for it to use based on how the avro data is being put onto the Kafka topic.

According to my knowledge, `AbstractAvroSerializer` is for the scene serializing data between workers(due to network communications). We should register Avro Record to kryo because Kryo dose not support it primitively. This is why I define a new interface `DirectAvroSerializer` to serialize/deserialize between avro Record and raw bytes. Also, this is the reason that I separate Avro from storm-hdfs. Without the avro record registry to kyro, storm-kafka integration with Avro will get an error like [STORM-1504](https://issues.apache.org/jira/browse/STORM-1504).

>My original use case was to read avro data that had been serialized onto a kafka topic with a Confluent serializer. My solution was to create a bolt that that read the raw data with rawscheme, deserialized the raw bytes into a GenericRecord, and then emitted the generic record.

I think it is even better we deserialized the raw bytes into GenericRecord in Spout intead of a bolt.

>Something like this would have been very nice if I could give it a ConfluentDeserializer via dependency injection or something similar.

Yes, I will update the code to support `ConfluentDeserializer`.

Thanks @dossett for your quick reply. Anything above is wrong please fix me. Thanks again.

vesense:
The reason why I define the new interface `DirectAvroSerializer` and move avro to a sharable module is explained as above.

@harshach This PR is based on STORM-1504, you can take a look at it.

@dossett Your reviews have been addressed. Please have a look.

vesense:
@dossett BTW, I have a question about `ConfluentAvroSerializer`:
I can't see any unit tests for `ConfluentAvroSerializer`. I wonder how kryo create the object by using construction `ConfluentAvroSerializer(Kryo k, Map stormConf)` with parameters (ConfluentAvroSerializer is specified by avro.serializer).

PR #1379, opened 2016-04-29
STORM-1742 (1.x) More accurate 'complete latency'

HeartSaVioR:
PR for master: #1523 

* Acker computes 'complete latency' and sends back to Spout
  * start time: Acker receiving ACK_INIT from Spout
  * end time: Acker receiving ACK message which make the tuple tree completed
* When Acker provides complete latency, Spout just uses this value to record that latency
  * exception case: tuple timed out - Spout doesn't receive failed information from Acker

Btw, in fact Spout measures latency only acked tuple, not failed tuple. If performance matters, we can get rid of using time delta with fail / reset timeout cases.

PR #1523, opened 2016-06-27
STORM-1742 More accurate 'complete latency'

HeartSaVioR:
PR for 1.x: #1379 

* Acker computes 'complete latency' and sends back to Spout
  * start time: Acker receiving ACK_INIT from Spout
  * end time: Acker receiving ACK message which make the tuple tree completed
* When Acker provides complete latency, Spout just uses this value to record that latency
  * exception case: tuple timed out - Spout doesn't receive failed information from Acker

#1379 contains the result of functional test and performance test. Please refer that pull request.


harshach:
+1

PR #1526, opened 2016-06-27
STORM-1928 ShellSpout should check heartbeat while ShellSpout is waiting for subprocess to sync

HeartSaVioR:
When nextTuple is not calling at any chances (max spout pending, backpressure introduced at 1.0.0, and so on) multi-lang spout don't receive any message from ShellSpout, thus no messages from multi-lang spout side. (because multilang spout and ShellSpout communicates like ping-pong, unlike multilang bolt and ShellBolt)
In result, no heartbeat is being marked, and finally raises subprocess heartbeat timeout.

This PR changes ShellSpout to only check heartbeat while ShellSpout is waiting for subprocess to sync.
And this PR also changes ShellSpout.die logic similar to ShellBolt, since I found ShellSpout's die message was logged but worker itself was not died after logging. Only that task was died.

### DIFF
--- a/storm-core/src/jvm/org/apache/storm/spout/ShellSpout.java
+++ b/storm-core/src/jvm/org/apache/storm/spout/ShellSpout.java
@@ -25,6 +25,8 @@ import org.apache.storm.multilang.ShellMsg;
 import org.apache.storm.multilang.SpoutMsg;
 import org.apache.storm.task.TopologyContext;
 import org.apache.storm.utils.ShellProcess;
+
+import java.util.Arrays;
 import java.util.Map;
 import java.util.HashMap;
 import java.util.List;
@@ -32,6 +34,7 @@ import java.util.TimerTask;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
 
 import clojure.lang.RT;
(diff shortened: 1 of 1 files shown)


HeartSaVioR:
Could anyone review this? This is critical issue on multilang spout.

storm-core/src/jvm/org/apache/storm/spout/ShellSpout.java
```diff
@@ -47,6 +50,8 @@
     private String[] _command;
     private Map<String, String> env = new HashMap<>();
     private ShellProcess _process;
+    private volatile boolean _running = true;
+    private volatile Throwable _exception;
```
revans2: nit: this is only ever set to a RuntimeException already, and then it is wrapped in a RuntimeException.  It might be nice to just have it be a RuntimeException to begin with and then we can skip the wrapping.

revans2:
one minor nit, but overall I am +1 with or without the change.

storm-core/src/jvm/org/apache/storm/spout/ShellSpout.java
HeartSaVioR: Yes it would be better to fix this. Thanks for pointing out. Fixed.

revans2:
+1 but all 4 CI jobs failed.  They look unrelated to this, as some of them are failing because of a RAT problem.  Lets fix the build ASAP before we merge in much more.

revans2:
The file without a license is external/storm-kafka-monitor/src/main/java/org/apache/storm/kafka/monitor/KafkaPartitionOffsetLag.java

I'll file a separate JIRA for that.

HeartSaVioR:
Just rebased with master which fixes RAT issue. I'll merge after build shows green or at least issues are all unrelated.

HeartSaVioR:
OK. Only one of build which is run by `!storm-core` failed. Merging.

PR #1527, opened 2016-06-28
[STORM-1932] Change Duration's value data type

darionyaphet:
[STORM-1932](https://issues.apache.org/jira/browse/STORM-1932)

BaseWindowedBolt Duration value is integer now . 

Use long to express time interval maybe better.

### DIFF
--- a/storm-core/src/jvm/org/apache/storm/topology/WindowedBoltExecutor.java
+++ b/storm-core/src/jvm/org/apache/storm/topology/WindowedBoltExecutor.java
@@ -76,7 +76,7 @@ public class WindowedBoltExecutor implements IRichBolt {
         this.bolt = bolt;
     }
 
-    private int getTopologyTimeoutMillis(Map stormConf) {
+    private long getTopologyTimeoutMillis(Map stormConf) {
         if (stormConf.get(Config.TOPOLOGY_ENABLE_MESSAGE_TIMEOUTS) != null) {
             boolean timeOutsEnabled = (boolean) stormConf.get(Config.TOPOLOGY_ENABLE_MESSAGE_TIMEOUTS);
             if (!timeOutsEnabled) {
@@ -98,7 +98,7 @@ public class WindowedBoltExecutor implements IRichBolt {
         return maxPending;
     }
 
-    private void ensureDurationLessThanTimeout(int duration, int timeout) {
+    private void ensureDurationLessThanTimeout(long duration, long timeout) {
         if (duration > timeout) {
             throw new IllegalArgumentException("Window duration (length + sliding interval) value " + duration +
                                                        " is more than " + Config.TOPOLOGY_MESSAGE_TIMEOUT_SECS +
@@ -117,7 +117,7 @@ public class WindowedBoltExecutor implements IRichBolt {
     private void validate(Map stormConf, Count windowLengthCount, Duration windowLengthDuration,
                           Count slidingIntervalCount, Duration slidingIntervalDuration) {
 
-        int topologyTimeout = getTopologyTimeoutMillis(stormConf);
+        long topologyTimeout = getTopologyTimeoutMillis(stormConf);
         int maxSpoutPending = getMaxSpoutPending(stormConf);
         if (windowLengthCount == null && windowLengthDuration == null) {
             throw new IllegalArgumentException("Window length is not specified");
--- a/storm-core/src/jvm/org/apache/storm/topology/base/BaseWindowedBolt.java
+++ b/storm-core/src/jvm/org/apache/storm/topology/base/BaseWindowedBolt.java
@@ -66,10 +66,14 @@ public abstract class BaseWindowedBolt implements IWindowedBolt {
      * Holds a Time duration for time based windows and sliding intervals.
      */
     public static class Duration {
-        public final int value;
+        public final long value;
 
         public Duration(int value, TimeUnit timeUnit) {
-            this.value = (int) timeUnit.toMillis(value);
+            this.value = timeUnit.toMillis(value);
+        }
+
+        public Duration(long value) {
+            this.value = value;
         }
 
         /**
--- a/storm-core/src/jvm/org/apache/storm/trident/windowing/config/WindowConfig.java
+++ b/storm-core/src/jvm/org/apache/storm/trident/windowing/config/WindowConfig.java
@@ -31,13 +31,13 @@ public interface WindowConfig extends Serializable {
      * Returns the length of the window.
      * @return
      */
-    public int getWindowLength();
+    public long getWindowLength();
 
     /**
      * Returns the sliding length of the moving window.
      * @return
      */
-    public int getSlidingLength();
+    public long getSlidingLength();
 
     /**
      * Gives the type of windowing. It can be any of {@code Type} values.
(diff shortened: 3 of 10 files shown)


harshach:
+1

HeartSaVioR:
@darionyaphet 
Build is failing with your changes. Could you take care of this?
I'd recommend to build the entire code by `mvn clean install` and see success at least once, and publish your change.

darionyaphet:
@HeartSaVioR It seems fixed .

HeartSaVioR:
@darionyaphet Great. Thanks for the work. +1

storm-core/src/jvm/org/apache/storm/topology/WindowedBoltExecutor.java
```diff
@@ -76,7 +76,7 @@ public WindowedBoltExecutor(IWindowedBolt bolt) {
         this.bolt = bolt;
     }
 
-    private int getTopologyTimeoutMillis(Map stormConf) {
+    private long getTopologyTimeoutMillis(Map stormConf) {
```
arunmahadevan: TOPOLOGY_MESSAGE_TIMEOUT_SECS is integer. I don't think this method should return a long.

storm-core/src/jvm/org/apache/storm/topology/base/BaseWindowedBolt.java
```diff
@@ -66,10 +66,14 @@ public String toString() {
...
     public static class Duration {
-        public final int value;
+        public final long value;
 
         public Duration(int value, TimeUnit timeUnit) {
-            this.value = (int) timeUnit.toMillis(value);
+            this.value = timeUnit.toMillis(value);
+        }
+
+        public Duration(long value) {
```
arunmahadevan: Please add TimeUnit as a second arg instead of automatically assuming the time unit. Also have an overloaded `public static Duration of(long milliseconds)`. For other static factories long does not make much sense.

storm-core/src/jvm/org/apache/storm/windowing/CountEvictionPolicy.java
```diff
@@ -26,10 +26,10 @@
  * @param <T> the type of event tracked by this policy.
  */
 public class CountEvictionPolicy<T> implements EvictionPolicy<T> {
-    private final int threshold;
+    private final long threshold;
```
arunmahadevan: For counts int should be sufficient. 

If you are changing it to long, the `currentCount` below should be changed to AtomicLong. You should also change the definition of `Count` class and add overloaded methods to take a long value.

arunmahadevan:
@darionyaphet You need to make more changes,

1. you should use `longValue()` instead of the `intValue()` in the `initWindowManager` where the values are read from storm conf.
2. The corresponding constants defined in the Config classes like TOPOLOGY_BOLTS_WINDOW_LENGTH_DURATION_MS are validated for integer range since its annotated with ` @isInteger`. This needs to be removed and appropriate annotation should be specified.

Its also good to add some unit tests to validate the changes.

storm-core/src/jvm/org/apache/storm/topology/WindowedBoltExecutor.java
```diff
@@ -117,7 +117,7 @@ private void ensureCountLessThanMaxPending(int count, int maxPending) {
     private void validate(Map stormConf, Count windowLengthCount, Duration windowLengthDuration,
                           Count slidingIntervalCount, Duration slidingIntervalDuration) {
 
-        int topologyTimeout = getTopologyTimeoutMillis(stormConf);
+        long topologyTimeout = getTopologyTimeoutMillis(stormConf);
```
satishd: This should be `int`, toplogy-timeout-millis type is int.

storm-core/src/jvm/org/apache/storm/trident/windowing/config/WindowConfig.java
```diff
@@ -31,13 +31,13 @@
      * Returns the length of the window.
      * @return
      */
-    public int getWindowLength();
+    public long getWindowLength();
```
satishd: If we want to change counts also to `Long` then `SlidingCountWindow` and `TumblingCountWindow` should be changed respectively. But there is no real need to have `Long` for count based windowing.

storm-core/src/jvm/org/apache/storm/windowing/CountEvictionPolicy.java
satishd: Agree with @arunmahadevan

darionyaphet:
@arunmahadevan why don't have `@isLong` ? millisecond using it to decorate is better

storm-core/src/jvm/org/apache/storm/topology/WindowedBoltExecutor.java
darionyaphet: finally , the return value is `timeout * 1000` , it have convert to millisecond . Long is a better choose .

storm-core/src/jvm/org/apache/storm/topology/base/BaseWindowedBolt.java
darionyaphet: actually `Duration(long value)` is represent the time interval in milliseconds, so it doesn't time unit to covert .

storm-core/src/jvm/org/apache/storm/windowing/CountEvictionPolicy.java
darionyaphet: the `CountWindow` and `DurationWindow` both extend from WindowConfig and `windowLength` and `slideLength` are Long type.

storm-core/src/jvm/org/apache/storm/trident/windowing/config/WindowConfig.java
darionyaphet: using generics to decide return value's date type is a good idea ?

storm-core/src/jvm/org/apache/storm/topology/WindowedBoltExecutor.java
arunmahadevan: Had missed that the value is getting converted to ms. Ok to have a long.

storm-core/src/jvm/org/apache/storm/topology/WindowedBoltExecutor.java
```diff
@@ -76,7 +76,7 @@ public WindowedBoltExecutor(IWindowedBolt bolt) {
         this.bolt = bolt;
     }
 
-    private int getTopologyTimeoutMillis(Map stormConf) {
+    private long getTopologyTimeoutMillis(Map stormConf) {
         if (stormConf.get(Config.TOPOLOGY_ENABLE_MESSAGE_TIMEOUTS) != null) {
             boolean timeOutsEnabled = (boolean) stormConf.get(Config.TOPOLOGY_ENABLE_MESSAGE_TIMEOUTS);
             if (!timeOutsEnabled) {
```
arunmahadevan: return Long.MAX_VALUE below.

storm-core/src/jvm/org/apache/storm/topology/base/BaseWindowedBolt.java
arunmahadevan: To have less confusion, its better to have only one ctor with explicit TimeUnit and add a `public static Duration of(long milliseconds)`.

arunmahadevan:
Either you can use `@isNumber` or add a `@isLong` validator to decorate the millisecond values, its not already available.

PR #1535, opened 2016-07-01
STORM-1941 Nimbus discovery can fail when zookeeper reconnect happens.

HeartSaVioR:
* delete ephemeral node first when reconnected handler is called

PR for 1.x: #1534

Please read [STORM-1941](https://issues.apache.org/jira/browse/STORM-1941) before reviewing. I left full description from there.

### DIFF
--- a/storm-core/src/jvm/org/apache/storm/cluster/StormClusterStateImpl.java
+++ b/storm-core/src/jvm/org/apache/storm/cluster/StormClusterStateImpl.java
@@ -219,6 +219,8 @@ public class StormClusterStateImpl implements IStormClusterState {
                 LOG.info("Connection state listener invoked, zookeeper connection state has changed to {}", connectionState);
                 if (connectionState.equals(ConnectionState.RECONNECTED)) {
                     LOG.info("Connection state has changed to reconnected so setting nimbuses entry one more time");
+                    // explicit delete for ephmeral node to ensure this session creates the entry.
+                    stateStorage.delete_node(ClusterUtils.nimbusPath(nimbusId));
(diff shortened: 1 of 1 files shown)


storm-core/src/jvm/org/apache/storm/cluster/StormClusterStateImpl.java
```diff
@@ -219,6 +219,8 @@ public void stateChanged(CuratorFramework curatorFramework, ConnectionState conn
                 LOG.info("Connection state listener invoked, zookeeper connection state has changed to {}", connectionState);
                 if (connectionState.equals(ConnectionState.RECONNECTED)) {
                     LOG.info("Connection state has changed to reconnected so setting nimbuses entry one more time");
+                    // explicit delete for ephmeral node to ensure this session creates the entry.
+                    stateStorage.delete_node(ClusterUtils.nimbusPath(nimbusId));
```
harshach: Aren't we using ephemeral node for this? If not I think we should do that no? Instead of manually deleting the node.

HeartSaVioR: @harshach 
This is for edge case for ephemeral node not deleted while session is closed.
Please refer http://issues.apache.org/jira/browse/STORM-1941.

I know it's odd but it is happening, and other article states this behavior.
https://www.box.com/blog/a-gotcha-when-using-zookeeper-ephemeral-nodes/

harshach:
@HeartSaVioR makes sense. +1

satishd:
+1

PR #1537, opened 2016-07-01
STORM-1942: remove extra div that was causing weird margins in topolo…

abellina:
…gy.html


harshach:
+1. No need to file jiras for small changes. Just open a PR.

abellina:
@priyank5485 mind taking a look at this one?

PR #1538, opened 2016-07-02
Sending the Data to Kafka as a batch, instead of sending one tuple ea…

supermonk:
Sending the Data to Kafka as a batch, instead of sending one tuple each. 0.10 storm-kafka version uses Producer.scala (kafka-client) class which has two methods, we are using now the second method which sends in batch.

Ref : https://github.com/apache/kafka/blob/623ab1e7c6497c000bc9c9978637f20542a3191c/core/src/main/scala/kafka/javaapi/producer/Producer.scala

### DIFF
--- a/external/storm-kafka/src/jvm/storm/kafka/trident/TridentKafkaState.java
+++ b/external/storm-kafka/src/jvm/storm/kafka/trident/TridentKafkaState.java
@@ -77,7 +77,7 @@ public class TridentKafkaState implements State {
         producer = new Producer(config);
     }
 
-    @SuppressWarnings({"rawtypes", "unchecked", "unused"})
+    @SuppressWarnings({"rawtypes", "unchecked"})
     public void updateState(List<TridentTuple> tuples, TridentCollector collector) {
         String topic = null;
         List<KeyedMessage> batchList = new ArrayList<KeyedMessage>(tuples.size());
@@ -101,12 +101,8 @@ public class TridentKafkaState implements State {
         }
         // Sending Batch
         try {
-            if (batchList != null) {
-                producer.send(batchList);
-                LOG.debug("Sending the Batch " + batchList.hashCode());
-            } else {
-                LOG.warn("BatchList is null " + batchList);
-            }
+            producer.send(batchList);
+            LOG.debug("Sending the Batch " + batchList);
         } catch (Exception ex) {
             String errorMsg = "Could not send messages = " + tuples + " to topic = " + topic;
             LOG.warn(errorMsg, ex);
--- a/external/storm-kafka/src/jvm/storm/kafka/trident/TridentKafkaState.java
+++ b/external/storm-kafka/src/jvm/storm/kafka/trident/TridentKafkaState.java
@@ -85,10 +85,13 @@ public class TridentKafkaState implements State {
 		for (TridentTuple tuple : tuples) {
 			try {
 				topic = topicSelector.getTopic(tuple);
-				batchList
-						.add(new KeyedMessage(topic, mapper.getKeyFromTuple(tuple), 
-						mapper.getMessageFromTuple(tuple)));
-				LOG.debug("Updated Batch");
+				if (topic != null) {
+					batchList.add(
+							new KeyedMessage(topic, mapper.getKeyFromTuple(tuple), mapper.getMessageFromTuple(tuple)));
+					LOG.debug("Updated Batch");
+				} else {
+					LOG.warn("skipping key = " + mapper.getKeyFromTuple(tuple) + ", topic selector returned null.");
+				}
 			} catch (Exception ex) {
 				String errorMsg = "Error while filling up List for Batching";
 				LOG.warn(errorMsg, ex);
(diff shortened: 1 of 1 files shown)


external/storm-kafka/src/jvm/storm/kafka/trident/TridentKafkaState.java
```diff
@@ -76,24 +77,41 @@ public void prepare(Map stormConf) {
         producer = new Producer(config);
     }
 
-    public void updateState(List<TridentTuple> tuples, TridentCollector collector) {
-        String topic = null;
-        for (TridentTuple tuple : tuples) {
-            try {
-                topic = topicSelector.getTopic(tuple);
+    @SuppressWarnings("rawtypes")
```
abellina: @supermonk please use spaces, not tabs.

HeartSaVioR:
@supermonk 
Thanks for the work. Since this is not trivial patch (though diff is small) could you file an issue for Apache JIRA and set title to have JIRA issue ID as prefix?
I'd also recommend you to craft pull request against master, and leave a message that you would love to have this to 0.10.x.
Thanks in advance!

external/storm-kafka/src/jvm/storm/kafka/trident/TridentKafkaState.java
supermonk: @abellina, Thanks for the comment. I have fixed it now. let me know if any queries.

supermonk:
@Parth-Brahmbhatt , @harshach  Could you review my change and let me know if it is fine to push to Kafka as batch rather than tuple by tuple

external/storm-kafka/src/jvm/storm/kafka/trident/TridentKafkaState.java
```diff
@@ -76,24 +77,40 @@ public void prepare(Map stormConf) {
...
-                String errorMsg = "Could not send message with key = " + mapper.getKeyFromTuple(tuple)
-                        + " to topic = " + topic;
+                String errorMsg = "Error while filling up List for Batching";
                 LOG.warn(errorMsg, ex);
                 throw new FailedException(errorMsg, ex);
             }
         }
+        // Sending Batch
+        try {
+            if (batchList != null) {
```
abellina: will we ever hit this if? Seems like batchList will always be not null.

external/storm-kafka/src/jvm/storm/kafka/trident/TridentKafkaState.java
abellina: Thanks @supermonk

supermonk:
@abellina : Thanks for the comment. I have now fixed it. Let me know if any queries.

external/storm-kafka/src/jvm/storm/kafka/trident/TridentKafkaState.java
```diff
@@ -76,24 +77,36 @@ public void prepare(Map stormConf) {
...
         for (TridentTuple tuple : tuples) {
             try {
                 topic = topicSelector.getTopic(tuple);
-
-                if(topic != null) {
-                    producer.send(new KeyedMessage(topic, mapper.getKeyFromTuple(tuple),
+                if (topic != null) {
+                    batchList.add(new KeyedMessage(topic, mapper.getKeyFromTuple(tuple),
                             mapper.getMessageFromTuple(tuple)));
+                    LOG.debug("Updated Batch");
```
harshach: lets remove this log dont' see much use for it.

external/storm-kafka/src/jvm/storm/kafka/trident/TridentKafkaState.java
```diff
@@ -76,24 +77,36 @@ public void prepare(Map stormConf) {
...
-                        + " to topic = " + topic;
+                String errorMsg = "Error while filling up List for Batching";
                 LOG.warn(errorMsg, ex);
                 throw new FailedException(errorMsg, ex);
             }
         }
+        // Sending Batch
+        try {
+            producer.send(batchList);
+            LOG.debug("Sending the Batch " + batchList);
```
harshach: lets not log the entire batchList . Probably useful to add the batchList.size

harshach:
@supermonk 
I am not quite sure if this patch helps a lot. We do have internal batching in kafka producer api. It can be controlled by batch.size so even if you call producer.send for every tuple it won't send that tuple immediately . It will only call send once the batch.size is reached.

supermonk:
@harshach : Thanks for the update.   I know that batch.size is useful in Java Client for Latest Kafka Build, or do you think in the older version it should hold good.

The storm 0.10 uses scala version and in trident (storm-kafka) I have not seen anywhere the batch.size begin used. As per the below blog, there is no specification done in the config.

**So I believe neither we are passing in the config, as per readme of Storm-kafka nor we are doing the below calculation and pushing the config internally.**

Kindly correct me if I am wrong!


https://gist.github.com/mrflip/5958028 
spout_batch_kb ~= max_fetch_size * kafka_machines * kpartitions_per_broker / 1024

supermonk:
@harshach : I have removed the extra logs. Thanks for the comment!

knusbaum:
I'm going to close this PR. It hasn't been touched in a few months, it is still missing a Jira, and we have decided to halt any non-bug work on 10.* branches.

In addition, it looks to me like `batch.size` is available for Kafka at least as far back as 0.8.0, so utilizing that sounds like the correct way to address this issue. 

However, I only spent a brief time looking at the docs, so If you think this is still a valuable change for the 1.x and master branches, please file a Jira and open PRs to those branches.

PR #1540, opened 2016-07-06
STORM-1945 Fix NPE bugs on topology spout lag for storm-kafka-monitor

HeartSaVioR:
PR for 1.x-branch: #1541

* Fix several spots which are throwing NPE on TopologySpoutLag
  * Use a trick to get classname of Spout without requiring storm-kafka and storm-kafka-client
* Move storm-kafka-monitor jar to toollib directory
  * since jars in extlib are added to worker classpath which we don't want for this jar

@priyank5485 @harshach Please take a look and comment. Thanks!

### DIFF
--- a/bin/storm-kafka-monitor
+++ b/bin/storm-kafka-monitor
@@ -40,4 +40,4 @@ else
   JAVA="$JAVA_HOME/bin/java"
 fi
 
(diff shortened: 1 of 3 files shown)


HeartSaVioR:
Already got +1 from 1.x branch PR and diff is same. Merging.

PR #1541, opened 2016-07-06
STORM-1945 Fix NPE bugs on topology spout lag for storm-kafka-monitor (1.x)

HeartSaVioR:
PR for master: #1540 

* Fix several spots which are throwing NPE on TopologySpoutLag
  * Use a trick to get classname of Spout without requiring storm-kafka and storm-kafka-client
* Move storm-kafka-monitor jar to toollib directory
  * since jars in extlib are added to worker classpath which we don't want for this jar

@priyank5485 @harshach Please take a look and comment. Thanks!

### DIFF
--- a/storm-core/src/jvm/org/apache/storm/utils/TopologySpoutLag.java
+++ b/storm-core/src/jvm/org/apache/storm/utils/TopologySpoutLag.java
@@ -43,26 +43,44 @@ public class TopologySpoutLag {
     public static List<Map<String, Object>> lag (StormTopology stormTopology, Map topologyConf) {
         List<Map<String, Object>> result = new ArrayList<>();
         Map<String, SpoutSpec> spouts = stormTopology.get_spouts();
-        Object object = null;
+        String className = null;
         for (Map.Entry<String, SpoutSpec> spout: spouts.entrySet()) {
             try {
                 SpoutSpec spoutSpec = spout.getValue();
                 ComponentObject componentObject = spoutSpec.get_spout_object();
-                object = Utils.getSetComponentObject(componentObject);
-                if (object.getClass().getCanonicalName().endsWith("storm.kafka.spout.KafkaSpout")) {
+                // FIXME: yes it's a trick so we might be better to find alternative way...
+                className = getClassNameFromComponentObject(componentObject);
+                logger.debug("spout classname: {}", className);
+                if (className.endsWith("storm.kafka.spout.KafkaSpout")) {
                     result.add(getLagResultForNewKafkaSpout(spout.getKey(), spoutSpec, topologyConf));
-                } else if (object.getClass().getCanonicalName().endsWith("storm.kafka.KafkaSpout")) {
+                } else if (className.endsWith("storm.kafka.KafkaSpout")) {
                     result.add(getLagResultForOldKafkaSpout(spout.getKey(), spoutSpec, topologyConf));
                 }
             } catch (Exception e) {
-                logger.warn("Exception thrown while getting lag for spout id: " + spout.getKey() + " and spout class: " + object.getClass().getCanonicalName());
-                logger.warn("Exception message:" + e.getMessage());
+                logger.warn("Exception thrown while getting lag for spout id: " + spout.getKey() + " and spout class: " + className);
+                logger.warn("Exception message:" + e.getMessage(), e);
             }
         }
         return result;
     }
 
+    private static String getClassNameFromComponentObject(ComponentObject componentObject) {
+        try {
+            Object object = Utils.getSetComponentObject(componentObject);
+            return object.getClass().getCanonicalName();
+        } catch (RuntimeException e) {
+
+            if (e.getCause() instanceof ClassNotFoundException) {
+                return e.getCause().getMessage().trim();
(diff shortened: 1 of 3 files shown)


harshach:
+1.

storm-core/src/jvm/org/apache/storm/utils/TopologySpoutLag.java
```diff
@@ -43,26 +43,44 @@
     public static List<Map<String, Object>> lag (StormTopology stormTopology, Map topologyConf) {
         List<Map<String, Object>> result = new ArrayList<>();
         Map<String, SpoutSpec> spouts = stormTopology.get_spouts();
-        Object object = null;
+        String className = null;
         for (Map.Entry<String, SpoutSpec> spout: spouts.entrySet()) {
             try {
                 SpoutSpec spoutSpec = spout.getValue();
                 ComponentObject componentObject = spoutSpec.get_spout_object();
-                object = Utils.getSetComponentObject(componentObject);
```
HeartSaVioR: The problem starts here. For many situation we can't deserialize Spout object since UI doesn't have dependency for that Spout, for example, storm-kafka or storm-kafka-client.

harshach: @HeartSaVioR how do we deserialize for supervisors or others they don't have any dependency on  spouts or any other component.

harshach:
@HeartSaVioR while we are here can we fix the JSON response as well.
currently what we have is this
[{"spoutLagResult":"[{\"topic\":\"topic\",\"partition\":0,\"consumerCommittedOffset\":1175610,\"logHeadOffset\":5634192,\"lag\":4458582},{\"topic\":\"topic\",\"partition\":1,\"consumerCommittedOffset\":1251325,\"logHeadOffset\":5634192,\"lag\":4382867}]","spoutId":"spout","spoutType":"KAFKA"}]

we should be returning an array instead lets have something like this

{ "spoutId" : "spout"
  "spoutType" :  "KAFKA"
  "spoutLagResult" : [
      "topic": // key is topic name
            ["partition0": {\"consumerCommittedOffset\":1175610,\"logHeadOffset\":5634192,\"lag\":4458582},
             "partition2: {\"consumerCommittedOffset\":1175610,\"logHeadOffset\":5634192,\"lag\":4458582},
            ],
      "topic2" : 
                 ["partition0": {\"consumerCommittedOffset\":1175610,\"logHeadOffset\":5634192,\"lag\":4458582},
             "partition2: {\"consumerCommittedOffset\":1175610,\"logHeadOffset\":5634192,\"lag\":4458582},
            ]
  
}

storm-core/src/jvm/org/apache/storm/utils/TopologySpoutLag.java
HeartSaVioR: @harshach Only task calls this method, no need to deserialize spout / bolt objects on daemons. Now UI is an exception case.

storm-dist/binary/src/main/assembly/binary.xml
```diff
@@ -340,10 +340,10 @@
         </fileSet>
         <!-- $STORM_HOME/extlib -->
         <fileSet>
-            <directory>${project.basedir}/../../external/storm-kafka-monitor/target</directory>
+            <directory></directory>
```
abhishekagarwal87: This change is not clear. Directory change from extlib to toollib is happening in other section. Shouldn't this be removed entirely

HeartSaVioR: It's needed to create empty directory "extlib". 
https://github.com/apache/storm/blob/1.0.x-branch/storm-dist/binary/src/main/assembly/binary.xml#L335

abhishekagarwal87: I see. Since storm-kafka-monitor was being copied always, this section was not needed anymore. Thanks for clarifying.

HeartSaVioR:
@harshach 
I guess you mean this: 
```
[
   {
      "spoutId":"spout",
      "spoutType":"KAFKA",
      "spoutLagResult":{
         "topic":{
            "partition0":{
               "consumerCommittedOffset":1175610,
               "logHeadOffset":5634192,
               "lag":4458582
            },
            "partition2":{
               "consumerCommittedOffset":1175610,
               "logHeadOffset":5634192,
               "lag":4458582
            }
         },
         "topic2":{
            "partition0":{
               "consumerCommittedOffset":1175610,
               "logHeadOffset":5634192,
               "lag":4458582
            },
            "partition2":{
               "consumerCommittedOffset":1175610,
               "logHeadOffset":5634192,
               "lag":4458582
            }
         }
      }
   }
]
```

I agree this is better for removing duplicated thing and also easy to find by path. I'll update it.

HeartSaVioR:
It would be better for changing JSON response to file a new issue and make it out of scope for this PR, since [STORM-1945](https://issues.apache.org/jira/browse/STORM-1945) is a bugfix which is marked as critical.
Filed new issue for changing JSON response format: [STORM-1950](https://issues.apache.org/jira/browse/STORM-1950)

PR #1543, opened 2016-07-07
STORM-1939 Ignore InterruptedException on ShellBoltMessageQueue

HeartSaVioR:
InterruptedException makes false alarm for ShellBolt shutting down.
Since it's likely throwing InterruptedException when ShellBolt is shutting down, so we're ok to just consume that and continue.

Please see https://issues.apache.org/jira/browse/STORM-1939 for more details.

It should be applied to master, 1.x, 1.0.x branches.

### DIFF
--- a/storm-core/src/jvm/org/apache/storm/task/ShellBolt.java
+++ b/storm-core/src/jvm/org/apache/storm/task/ShellBolt.java
@@ -158,8 +158,8 @@ public class ShellBolt implements IBolt {
 
             _pendingWrites.putBoltMsg(boltMsg);
         } catch(InterruptedException e) {
-            String processInfo = _process.getProcessInfoString() + _process.getProcessTerminationInfoString();
-            throw new RuntimeException("Error during multilang processing " + processInfo, e);
+            // It's likely that Bolt is shutting down so no need to throw RuntimeException
+            // just ignore
         }
     }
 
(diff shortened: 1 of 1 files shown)


storm-core/src/jvm/org/apache/storm/task/ShellBolt.java
```diff
@@ -158,8 +158,8 @@ public void execute(Tuple input) {
 
             _pendingWrites.putBoltMsg(boltMsg);
         } catch(InterruptedException e) {
-            String processInfo = _process.getProcessInfoString() + _process.getProcessTerminationInfoString();
-            throw new RuntimeException("Error during multilang processing " + processInfo, e);
+            // It's likely that Bolt is shutting down so no need to throw RuntimeException
```
abhishekagarwal87: It will be good to have a jvm wide flag to indicate whether jvm shutdown is in progress. There is also a java hack for the same. I can add that flag in a separate PR. Do you mind holding onto this PR till I come up with that?

HeartSaVioR: Yes sure. Please ping me once it's done.

HeartSaVioR: @abhishekagarwal87 Do you have update on this? If your work is expected to take longer, please review this first and we could handle your improvement by separate issue.

abhishekagarwal87: My bad @HeartSaVioR. You can merge this for now. I have captured the improvement in 
https://issues.apache.org/jira/browse/STORM-1600

abhishekagarwal87:
+1

PR #1547, opened 2016-07-08
STORM-1954: Large Trident topologies can cause memory issues due to DefaultResourceDeclarer object reading config

knusbaum:


harshach:
+1

PR #1551, opened 2016-07-11
STORM-1959 Add missing license header to KafkaPartitionOffsetLag

HeartSaVioR:
This is just an addition of missing license header (not code change), so I think it can be merged without waiting one day rule. I feel Travis CI pass and another one +1 by other committer are sufficient to check in.
It should be applied to also 1.x branch.


revans2:
+1 pending travis

revans2:
Travis passed this is just a comment/documentation change, not a code change so I will push this in without waiting 24 hours.

PR #1552, opened 2016-07-11
STORM-1960. Add CORS support to STORM UI Rest api.

harshach:


harshach:
@HeartSaVioR @arunmahadevan pinging for a review

arunmahadevan:
@harshach the changes look good. May be you could also add a note about this feature in the storm rest api docs. Since CORS is going to be on by default, if it makes sense we could take a storm config option where by users can optionally turn it off.

HeartSaVioR:
Same as @arunmahadevan. I'm wondering that we need to open detailed configurations for these, but we can add it later when someone claims he/she needs that.

harshach:
@HeartSaVioR @arunmahadevan don't want to add one more config for this. Let me know if you see any issues of keeping CORS. Will document in REST API doc.

revans2:
I thought we already did a lot of this for the search side of things.

https://github.com/apache/storm/blob/master/storm-core/src/jvm/org/apache/storm/ui/UIHelpers.java#L243-L245

How is this different and can we remove the other code if we are doing this?

harshach:
@revans2 when we tried rebalance, debug and changing the logger that particular code wasn't working. Yes that can be removed.

arunmahadevan:
@harshach May be for now we can document in the REST api doc and expose configs in future if needed.

harshach:
@revans2  for master I think this part and the one you linked https://github.com/apache/storm/blob/master/storm-core/src/jvm/org/apache/storm/ui/UIHelpers.java#L151 is taking care of it . Where as in 1.x-branch we don't have that. I added patch here for 1.x-branch https://github.com/apache/storm/pull/1557

harshach:
closing this PR as I don't a see a necessity to add this in master for above said reasons.

PR #1556, opened 2016-07-12
STORM-1737: storm-kafka-client has compilation errors with Apache Kafka 0.10

hmcl:


harshach:
+1

external/storm-kafka/src/test/org/apache/storm/kafka/KafkaTestBroker.java
```diff
@@ -17,21 +17,21 @@
...
-import org.apache.commons.io.FileUtils;
-
 import java.io.File;
 import java.io.IOException;
 import java.util.Properties;
 
+import kafka.server.KafkaConfig;
+import kafka.server.KafkaServerStartable;
+
 /**
```
HeartSaVioR: Out of scope, but we need to get rid of this comment.

HeartSaVioR:
What's our plan to support version lines of Kafka? I think we need to guide storm-kafka to users that for 0.8 you need to blabla, and for 0.9, and for 0.10.
And storm-kafka-client is no longer compatible with 0.9, it would be better to go with [STORM-1876](https://issues.apache.org/jira/browse/STORM-1876) (#1482). If this patch merged in without STORM-1876, 0.9 users should download the code, and **change directory to external/storm-kafka**, and build module with proper `kafka.version` and `kafka.artifact.id` parameters.

harshach:
@HeartSaVioR 0.9 users won't be able to use the new API anyway. So why are we worried about that part. ). 0.9 users can still use storm-kafka but the new kafka consumer in storm-kafka-client

harshach:
@HeartSaVioR I don't think this patch need to wait on storm-kafka changes in STORM-1876. Irrespective both these patches storm-kafka will work against 0.8, 0.9 and 0.10

HeartSaVioR:
@harshach 
Sorry I've confused with compatibility between client and broker version. Seems like kafka 0.9 broker is compatible with kafka 0.8 client. I guess you're based on this. Is it correct?

harshach:
@HeartSaVioR Yes. Users can still use the storm-kafka connector (old api) to connect to any version of kafka. Given that this is new consumer api we better move to compatible API soon. Hence the reason to push this as part of 1.0.2 release.

HeartSaVioR:
@hmcl Please update README.md to describe this module is compatible with 0.10.0.

@harshach We're having potential users which already uses storm-kafka-client with Kafka 0.9. It might be better for them to guide not updating their storm-kafka-client version as same as storm-core due to client version change, or dropping usage of storm-kafka-client and back to storm-kafka.

HeartSaVioR:
And let's clarify: what I'm thinking about is not only API level compatibility but Kafka client - broker version compatibility. We modified storm-kafka version to 0.8 due to this issue.

harshach:
@HeartSaVioR we released 1.0 like few weeks back and even they are using the new consumer api in 0.10 will be api compatible with 0.9 as well it just the parameter types that are changed.

abhishekagarwal87:
@harshach just wanted to make sure that compatibility of kafka 0.10  with 0.9 broker has been tested. 

Also given that there are three version lines of kafka client and two different kafka modules within storm, we should clearly document which combinations of kafka-client are supported. 

e.g.
storm-kafka supports kafka client 0.8.x and 0.9.x
storm-kafka-client supports kafka client 0.10.x

Apart from that kafka dependencies should be marked provided, so that users are forced to chose a kafka client version explicitly. Whether the kafka client version chosen by user works with broker, should be left for user to decide.

harshach:
@abhishekagarwal87 We can put that in the README on which version we are going to support for storm-kafka-client.
agree on making the dependencies provided.

abhishekagarwal87:
I will make the scope changes in #1482

harshach:
@HeartSaVioR is this ok to be merged in.

HeartSaVioR:
Where we would want to update support versions on storm-kafka-client? 
I'd be +1 when we're clear to where we handle that thing.

harshach:
@HeartSaVioR As stated before we can add this to README of storm-kafka-client

harshach:
@hmcl can you update the README with a Note: supports from kaka-client 0.10 version on wards

HeartSaVioR:
Sorry to not clear on `where`: I mean which PR (or new issue) will handle updating the doc. Now I'm +1 once @hmcl updates the doc.

harshach:
+1.

PR #1557, opened 2016-07-12
STORM-1960. Add CORS support to STORM UI Rest api

harshach:


storm-core/src/clj/org/apache/storm/ui/helpers.clj
```diff
@@ -221,10 +221,7 @@
 (defnk json-response
   [data callback :serialize-fn to-json :status 200 :headers {}]
   {:status status
-   :headers (merge {"Cache-Control" "no-cache, no-store"
-                    "Access-Control-Allow-Origin" "*"
-                    "Access-Control-Allow-Headers" "Content-Type, Access-Control-Allow-Headers, Access-Controler-Allow-Origin, X-Requested-By, X-Csrf-Token, Authorization, X-Requested-With"}
-              (if (not-nil? callback) {"Content-Type" "application/javascript;charset=utf-8"}
+   :headers ((if (not-nil? callback) {"Content-Type" "application/javascript;charset=utf-8"}
```
abellina: The Cache-Control no-cache, no-store is something we want to keep for json responses.

harshach: @abellina cors support added in this patch returns those as part of the headers

arunmahadevan: I think you still need to put `merge` before the `if` since it has to be merged with the headers argument.

harshach: @abellina you are right. Addressed in the PR.
@arunmahadevan the change will address your comment as well.

arunmahadevan:
+1

PR #1559, opened 2016-07-13
STORM-1965 Show warning message when dot is used for topology name or component name

HeartSaVioR:
* since having dot to its name will bother most of metrics consumer

It can be port back easily, and I'd like to get this merged to 1.x, and 1.0.x too.


abhishekagarwal87:
These warning messages can be confusing for someone who is not familiar with time series db and the special effect of dot. If at all, these warning messages should be in the MetricConsumer Implementation for the time series DB and not in the storm submitter class.

HeartSaVioR:
@abhishekagarwal87 Yeah I think you're right. It's time-series DB specific restriction. Thanks for reminding me.

PR #1560, opened 2016-07-13
STORM-1966 Expand metric having Map type as value into multiple metrics based on entries

HeartSaVioR:
PR for 1.x: #1561 

* populate metric before applying filter so that filter can play with populated metrics
* add relevant configs to metrics consumer registration and describe them to storm.config.yaml
* add unit test

Please refer https://issues.apache.org/jira/browse/STORM-1966 for more details.

### DIFF
--- a/conf/storm.yaml.example
+++ b/conf/storm.yaml.example
@@ -45,6 +45,12 @@
 ## - when none of configuration for metric filter are specified, it'll be treated as 'pass all'.
 ## - you need to specify either whitelist or blacklist, or none of them. You can't specify both of them.
 ## - you can specify multiple whitelist / blacklist with regular expression
+## expandMapType: expand metric with map type as value to multiple metrics
+## - set to true when you would like to apply filter to expanded metrics
+## - default value is false which is backward compatible value
+## metricNameSeparator: separator between origin metric name and key of entry from map
+## - only effective when expandMapType is set to true
+## - default value is "."
 # topology.metrics.consumer.register:
 #   - class: "org.apache.storm.metric.LoggingMetricsConsumer"
 #     max.retain.metric.tuples: 100
@@ -57,6 +63,8 @@
 #     parallelism.hint: 1
 #     argument:
 #       - endpoint: "metrics-collector.mycompany.org"
+#     expandMapType: true
+#     metricNameSeparator: "."
 
 ## Cluster Metrics Consumers
 # storm.cluster.metrics.consumer.register:
--- a/storm-core/src/jvm/org/apache/storm/daemon/StormCommon.java
+++ b/storm-core/src/jvm/org/apache/storm/daemon/StormCommon.java
@@ -17,13 +17,10 @@
  */
 package org.apache.storm.daemon;
 
-import com.codahale.metrics.MetricRegistry;
 import org.apache.storm.Config;
 import org.apache.storm.Constants;
 import org.apache.storm.Thrift;
 import org.apache.storm.cluster.IStormClusterState;
-import org.apache.storm.daemon.metrics.MetricsUtils;
-import org.apache.storm.daemon.metrics.reporters.PreparableReporter;
 import org.apache.storm.generated.Bolt;
 import org.apache.storm.generated.ComponentCommon;
 import org.apache.storm.generated.GlobalStreamId;
@@ -39,9 +36,9 @@ import org.apache.storm.metric.EventLoggerBolt;
 import org.apache.storm.metric.MetricsConsumerBolt;
 import org.apache.storm.metric.SystemBolt;
 import org.apache.storm.metric.filter.FilterByMetricName;
+import org.apache.storm.metric.util.DataPointExpander;
 import org.apache.storm.security.auth.IAuthorizer;
 import org.apache.storm.task.IBolt;
-import org.apache.storm.testing.NonRichBoltTracker;
 import org.apache.storm.utils.ConfigUtils;
 import org.apache.storm.utils.IPredicate;
 import org.apache.storm.utils.ThriftTopologyUtils;
@@ -83,6 +80,15 @@ public class StormCommon {
 
     public static final String EVENTLOGGER_COMPONENT_ID = "__eventlogger";
     public static final String EVENTLOGGER_STREAM_ID = "__eventlog";
+
+    public static final String TOPOLOGY_METRICS_CONSUMER_CLASS = "class";
+    public static final String TOPOLOGY_METRICS_CONSUMER_ARGUMENT = "argument";
+    public static final String TOPOLOGY_METRICS_CONSUMER_MAX_RETAIN_METRIC_TUPLES = "max.retain.metric.tuples";
+    public static final String TOPOLOGY_METRICS_CONSUMER_PARALLELISM_HINT = "parallelism.hint";
+    public static final String TOPOLOGY_METRICS_CONSUMER_WHITELIST = "whitelist";
+    public static final String TOPOLOGY_METRICS_CONSUMER_BLACKLIST = "blacklist";
+    public static final String TOPOLOGY_METRICS_CONSUMER_EXPAND_MAP_TYPE = "expandMapType";
+    public static final String TOPOLOGY_METRICS_CONSUMER_METRIC_NAME_SEPARATOR = "metricNameSeparator";
 
     public static String getStormId(final IStormClusterState stormClusterState, final String topologyName) {
         List<String> activeTopologys = stormClusterState.activeStorms();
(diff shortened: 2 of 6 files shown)


storm-core/src/jvm/org/apache/storm/metric/api/IMetricsConsumer.java
```diff
@@ -54,6 +54,25 @@ public String toString() {
...
         public Object value;
+
+        @Override
+        public boolean equals(Object o) {
+            if (this == o) return true;
+            if (!(o instanceof DataPoint)) return false;
+
+            DataPoint dataPoint = (DataPoint) o;
+
+            if (name != null ? !name.equals(dataPoint.name) : dataPoint.name != null) return false;
```
abhishekagarwal87: it will be good to have parantheses around individual comparisons e.g. (name != null) or it can be simplified to StringUtils.equals(name, dataPoint.name)

conf/storm.yaml.example
```diff
@@ -45,6 +45,10 @@
 ## - when none of configuration for metric filter are specified, it'll be treated as 'pass all'.
 ## - you need to specify either whitelist or blacklist, or none of them. You can't specify both of them.
 ## - you can specify multiple whitelist / blacklist with regular expression
+## populateMapType: expand metric with map type as value to multiple metrics
```
abhishekagarwal87: can this be renamed to expandMapType?

conf/storm.yaml.example
```diff
@@ -45,6 +45,10 @@
...
+## - set to true when you would like to apply filter to populated metrics
```
abhishekagarwal87: Please document the default value as well

storm-core/src/jvm/org/apache/storm/metric/util/DataPointPopulator.java
```diff
@@ -0,0 +1,61 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+public class DataPointPopulator implements Serializable {
```
abhishekagarwal87: I would prefer the name DataPointExpander. What do you think?

storm-core/src/jvm/org/apache/storm/metric/util/DataPointPopulator.java
```diff
@@ -0,0 +1,61 @@
...
+
+  public Collection<IMetricsConsumer.DataPoint> populateDataPoint(IMetricsConsumer.DataPoint dataPoint) {
+    if (!populateMapType) {
+      return Collections.singletonList(dataPoint);
+    }
+
+    List<IMetricsConsumer.DataPoint> dataPoints = new ArrayList<>();
+
+    if (dataPoint.value == null) {
+      LOG.warn("Data point with name " + dataPoint.name + " is null. Discarding." + dataPoint.name);
```
abhishekagarwal87: let's use {} syntax for consistency

storm-core/src/jvm/org/apache/storm/metric/util/DataPointPopulator.java
```diff
@@ -0,0 +1,61 @@
...
+    } else if (dataPoint.value instanceof Map) {
+      Map<String, Object> dataMap = (Map<String, Object>) dataPoint.value;
```
abhishekagarwal87: There is no guarantee that key type will be string. Might be better to assume it to of Object type and wrap entry.getKey() in String.valueOf

storm-core/src/jvm/org/apache/storm/metric/util/DataPointPopulator.java
```diff
@@ -0,0 +1,61 @@
...
+    List<IMetricsConsumer.DataPoint> populatedDataPoints = new ArrayList<>();
+
+    for (IMetricsConsumer.DataPoint dataPoint : dataPoints) {
+      populatedDataPoints.addAll(populateDataPoint(dataPoint));
+    }
+
+    return populatedDataPoints;
+  }
+
+  public Collection<IMetricsConsumer.DataPoint> populateDataPoint(IMetricsConsumer.DataPoint dataPoint) {
```
abhishekagarwal87: this currently does not handle nested maps. Is that intended?

HeartSaVioR: At least built-in metrics have at most one depth Map, but no more (I mean no nested Map). The reason we would want to have such kind of logic is to handle built-in metrics. In normal, having Map as value is not recommended (which is what Bobby said too flexible).

For example, some metrics consumers take care of nested map, some of others only take care one depth map, the others don't take care of map. Users shouldn't rely on them.

storm-core/src/jvm/org/apache/storm/metric/util/DataPointPopulator.java
HeartSaVioR: Yeah good point. Will address.

storm-core/src/jvm/org/apache/storm/metric/util/DataPointPopulator.java
HeartSaVioR: Thanks for finding. I copied the code from Ambari metrics Storm sink which uses Log4j so missed this. I'll address it.

storm-core/src/jvm/org/apache/storm/metric/util/DataPointPopulator.java
HeartSaVioR: Yes `expand` would be better. I was not sure which word is good to represent the behavior. We can replace all `populate` to `expand`.

conf/storm.yaml.example
HeartSaVioR: Yes I was thinking about that but missed it. I'll address it.

conf/storm.yaml.example
HeartSaVioR: OK Will do.

storm-core/src/jvm/org/apache/storm/metric/api/IMetricsConsumer.java
HeartSaVioR: This is IntelliJ generated code so might not beauty and doesn't rely on other library. StringUtils.equals() would be better. I'll address it.

HeartSaVioR: Since we're on JDK7, I can just use Objects.equals and Objects.deepEquals.

HeartSaVioR:
Thanks for thoughtful review, @abhishekagarwal87 ! Addressed your review comments.

storm-core/src/jvm/org/apache/storm/metric/MetricsConsumerBolt.java
```diff
@@ -43,19 +44,21 @@
...
+    private final Predicate<IMetricsConsumer.DataPoint> _filterPredicate;
+    private final DataPointExpander _populator;
 
     private final BlockingQueue<MetricsTask> _taskQueue;
     private Thread _taskExecuteThread;
     private volatile boolean _running = true;
 
     public MetricsConsumerBolt(String consumerClassName, Object registrationArgument, int maxRetainMetricTuples,
-                               Predicate<IMetricsConsumer.DataPoint> filterPredicate) {
+                               Predicate<IMetricsConsumer.DataPoint> filterPredicate, DataPointExpander populator) {
```
abhishekagarwal87: Minor nit. populator variable name can be changed as well.

abhishekagarwal87:
Minor nit regarding variable name. Other than that +1. 
Please make sure to squash commits before you merge :)

storm-core/src/jvm/org/apache/storm/metric/MetricsConsumerBolt.java
HeartSaVioR: Oh I missed that. Thanks!

HeartSaVioR:
Addressed and squashed.

abhishekagarwal87:
Thanks @HeartSaVioR . Looks good to me

HeartSaVioR:
@harshach Could you take a look?

ptgoetz:
+1

storm-core/src/jvm/org/apache/storm/daemon/StormCommon.java
```diff
@@ -391,9 +388,13 @@ public static void addEventLogger(Map conf, StormTopology topology) {
                 List<String> whitelist = (List<String>) info.get("whitelist");
                 List<String> blacklist = (List<String>) info.get("blacklist");
                 FilterByMetricName filterPredicate = new FilterByMetricName(whitelist, blacklist);
+                Boolean expandMapType = Utils.getBoolean(info.get("expandMapType"), false);
```
harshach: can we make these strings into some final String constants

harshach:
@HeartSaVioR looks good to me. Just a minor nitpick +1 after that.

HeartSaVioR:
@harshach Addressed.

I found RAT issue again (my bad) and fixed it.

Btw, I didn't address review comments to PR for 1.x-branch. Once I'm done I'll mention reviewers  to review from PR for 1.x-branch as well.

harshach:
Thanks @HeartSaVioR . +1.

HeartSaVioR:
I also addressed review comments and also RAT issue to 1.x PR. 
Please take a look at #1561 as well. Thanks!

PR #1562, opened 2016-07-13
TridentHiveTopology fails to run due to incompatibility between thrift dependencies

hmcl:
- hive-hcatalog-streaming:jar depends on thrift 0.9.3
 - storm-hive:jar depends on thrift 0.9.0
 - Update storm-hive thrift version to 0.9.3


hmcl:
This fix likely has to be backported to 1.x-branch

harshach:
+1

PR #1563, opened 2016-07-14
[STORM-1968] Storm logviewer does not work for nimbus.log in secure cluster

arunmahadevan:
logviewer invokes "get-log-user-group-whitelist" which tries to get the worker metadata file by invoking "get-log-metadata-file". In the case of nimbus.log clojure-from-yaml-file  returns nil and the authorization fails.

Modify clojure-from-yaml-file to return an empty map in case of failures so that the authorization can continue.


abhishekagarwal87:
+1

harshach:
+1

HeartSaVioR:
+1

ptgoetz:
+1

PR #1564, opened 2016-07-14
STORM-1969 Modify HiveTopology to show usage of non-partition table

HeartSaVioR:
There're some kinds of topology in storm-hive, but all of them uses partition so we don't have example for accessing non-partition table. It would be better to have one for automated tests.

Instead of creating a new topology, we can modify HiveTopology to not use partition at all. BucketTestHiveTopology and TridentHiveTopology also use DelimitedRecordHiveMapper.withTimeAsPartitionField() so I think usage of withTimeAsPartitionField() is covered.

It can be ported back easily so I don't create pull requests for other branches.
We would be better to have this to 1.x and maybe 1.0.x, too.


harshach:
+1

harshach:

PR #1566, opened 2016-07-14
[STORM-1971] HDFS Timed Synchronous Policy

darionyaphet:
[STORM-1971 HDFS Timed Synchronous Policy](https://issues.apache.org/jira/browse/STORM-1971)

When the data need to be wrote to HDFS is not very large in quantity . 

We need a timed synchronous policy to flush cached date into HDFS periodically.

### DIFF
--- a/external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/sync/TimedSyncPolicy.java
+++ b/external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/sync/TimedSyncPolicy.java
@@ -15,32 +15,34 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.apache.storm.hdfs.trident.sync;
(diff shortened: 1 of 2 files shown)


external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/sync/CountSyncPolicy.java
```diff
@@ -26,7 +26,7 @@
  * have been processed.
  */
 public class CountSyncPolicy implements SyncPolicy {
-    private int count;
+    private final int count;
```
dossett: Great catch to make that final. I would be +1 on that change in a separate PR.

darionyaphet: @dossett OK fixed :)

PR #1568, opened 2016-07-15
[STORM-1964] Unexpected behavior when using count window together with timestamp extraction

arunmahadevan:
Do not use timestamp to determine the event count in WatermarkCountEvictionPolicy
when count based trigger is used.


ptgoetz:
+1

PR #1569, opened 2016-07-17
[STORM-1974] Using System.lineSeparator to replacement write a new line

darionyaphet:
[STORM-1974 : Using System.lineSeparator to replacement write a new line](https://issues.apache.org/jira/browse/STORM-1974)

Using System.lineSeparator to replacement write a new line . 

It will write message in once and reduce writer synchroniz .


ptgoetz:
@darionyaphet This change has the same effect as the code it replaces and seems unnecessary as there is no synchronization going on here.

darionyaphet:
@ptgoetz This's just graceful than write twice .

PR #1570, opened 2016-07-17
[STORM-1975]  Support default value for KafkaBolt

darionyaphet:
[STORM-1975 Support default value for KafkaBolt](https://issues.apache.org/jira/browse/STORM-1975)

Support some default values (acks , key and value serializer)  for kafka bolt

### DIFF
--- a/external/storm-kafka/src/jvm/org/apache/storm/kafka/bolt/KafkaBolt.java
+++ b/external/storm-kafka/src/jvm/org/apache/storm/kafka/bolt/KafkaBolt.java
@@ -56,23 +56,36 @@ public class KafkaBolt<K, V> extends BaseRichBolt {
 
     private static final Logger LOG = LoggerFactory.getLogger(KafkaBolt.class);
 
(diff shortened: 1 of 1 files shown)


harshach:
@darionyaphet given that old consumer api is deprecated and we've new storm-kafka-client lets get our development efforts with storm-kafka-client. Unless there is a bug I am against adding any new features to storm-kafka.

darionyaphet:
@harshach  `Storm-kafka-client` is a new spout to consume data from Kafka .  It seems not include `KafkaBolt`. Is it good to port `Storm-kafka`'s `KafkaBolt` into `Storm-kafka-client` ?

ptgoetz:
@harshach @darionyaphet I don't think we're in a position to deprecate storm-kafka yet for two reasons:

1. storm-kafka-client does not have feature parity with storm-kafka yet (e.g. bolt, Trident support)
2. storm-kafka-client requires kafka 0.10.x. It will take some time for all users to upgrade to that version.

PR #1571, opened 2016-07-18
STORM-1976 Remove cleanup-corrupt-topologies!

HeartSaVioR:
For 1.x branch: #1572

* cleanup-corrupt-topologies! was born from non-H/A Nimbus age
* Keeping this with Nimbus H/A makes various issues which in result topology going away

Please refer [STORM-1976](https://issues.apache.org/jira/browse/STORM-1976) for more details.

PR #1572, opened 2016-07-18
STORM-1976 Remove cleanup-corrupt-topologies! (1.x)

HeartSaVioR:
For master branch: #1571 

* cleanup-corrupt-topologies! was born from non-H/A Nimbus age
* Keeping this with Nimbus H/A makes various issues which in result topology going away

Please refer [STORM-1976](https://issues.apache.org/jira/browse/STORM-1976) for more details.

### DIFF
--- a/storm-core/src/clj/org/apache/storm/daemon/nimbus.clj
+++ b/storm-core/src/clj/org/apache/storm/daemon/nimbus.clj
@@ -1180,19 +1180,6 @@
                                      (> (line :timestamp) cutoff-age)) curr-history))]
       (ls-topo-hist! topo-history-state new-history))))
 
-(defn cleanup-corrupt-topologies! [nimbus]
-  (let [storm-cluster-state (:storm-cluster-state nimbus)
-        blob-store (:blob-store nimbus)
-        code-ids (set (code-ids blob-store))
-        active-topologies (set (.active-storms storm-cluster-state))
-        corrupt-topologies (set/difference active-topologies code-ids)]
-    (doseq [corrupt corrupt-topologies]
-      (log-message "Corrupt topology " corrupt " has state on zookeeper but doesn't have a local dir on Nimbus. Cleaning up...")
-      (.remove-storm! storm-cluster-state corrupt)
-      (if (instance? LocalFsBlobStore blob-store)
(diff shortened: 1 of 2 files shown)


harshach:
+1 . build failures are not related to the patch.

revans2:
+1, but I would like to see a tool that can kill a corrupted topology without nimbus being involved, or instructions somewhere on how to do it by logging directly into ZooKeeper.

HeartSaVioR:
@revans2 
Yes agreed. After removing this code we can't remove corrupted topology in normal way since no nimbus can gain leadership. It would be nice to have a tool to show replication count of topologies *without leader nimbus involved* to determine corrupted topology, and kill one if users want to.

harshach:
@revans2 @HeartSaVioR why can't we enforce the min.replication.count by looking nimbus.seeds if there are more than one host in that list we should set the min.replication.count to 2. User can override the setting by specifying it in storm.yaml. My understanding is they are not doing it and we are relying on min.replication.count to be 1.

revans2:
@HeartSaVioR actually I have thought about it more and I am probably overreacting.  If you want to call these blockers  it is not that big of a deal.  I would just like the community have a follow on discussion about how we really want HA to behave in these situations now that we have had some experience with people running with HA.

revans2:
@harshach I don't think that would fix the issue, although it is an improvement.  If I think about it more I think what happened was that they were not running with HA/expecting HA to work, but ran into a bad situation where they really would have wanted to be running with HA, and then when they tried to recover, by bringing up a new nimbus instance, all of the topologies were killed.

harshach:
@HeartSaVioR @revans2 are we ok with merging this in and file a follow-up JIRA to have a discussion around the HA and expectations from user.

revans2:
Yes I am good with merging this in.

HeartSaVioR:
Yes I think discussion of Nimbus H/A and BlobStore would be longer, so we need to apply short-term fix first. There're lots of critical bugfixes in 1.0.2 so we shouldn't drag the release because of struggling long-term fix.

HeartSaVioR:
Regarding tool, I filed an issue: https://issues.apache.org/jira/browse/STORM-1985

PR #1575, opened 2016-07-19
STORM-1600: Do not report exceptions after jvm shutdown

abhishekagarwal87:
@HeartSaVioR  - Here is the proposed change. Can you suggest some way to reproduce a scenario which I could also add as a test case?

### DIFF
--- a/storm-core/src/clj/org/apache/storm/daemon/executor.clj
+++ b/storm-core/src/clj/org/apache/storm/daemon/executor.clj
@@ -205,15 +205,17 @@
      :report-error-and-die (reify
                              Thread$UncaughtExceptionHandler
                              (uncaughtException [this _ error]
-                               (try
-                                 ((:report-error <>) error)
-                                 (catch Exception e
-                                   (log-error e "Error while reporting error to cluster, proceeding with shutdown")))
-                               (if (or
-                                    (Utils/exceptionCauseIsInstanceOf InterruptedException error)
-                                    (Utils/exceptionCauseIsInstanceOf java.io.InterruptedIOException error))
-                                 (log-message "Got interrupted excpetion shutting thread down...")
-                                 ((:suicide-fn <>)))))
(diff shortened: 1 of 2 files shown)


storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1772,6 +1783,7 @@ public static String uuid() {
     public static void exitProcess (int val, String msg) {
         String combinedErrorMessage = "Halting process: " + msg;
         LOG.error(combinedErrorMessage, new RuntimeException(combinedErrorMessage));
+        isJVMShutdownInitiated.set(Boolean.TRUE);
```
HeartSaVioR: ShellSpout and ShellBolt just call System.exit() directly from their die() method. You may want to change those too.

HeartSaVioR:
Regarding test case, we're trying to shutdown JVM and throwing exception concurrently, so I wonder we can make it consistent. I also don't have an idea regarding the test which tries to shutdown JVM, and wonder it can be passed without affecting other tests.

storm-core/src/clj/org/apache/storm/daemon/executor.clj
```diff
@@ -205,15 +205,17 @@
...
+                               (if (Utils/isShutdownUnderProgress)
+                                 (log-warn error "Uncaught exception in thread after jvm shutdown")
+                                 ((try
+                                    ((:report-error <>) error)
+                                    (catch Exception e
+                                      (log-error e "Error while reporting error to cluster, proceeding with shutdown")))
+                                   (if (or
+                                         (Utils/exceptionCauseIsInstanceOf InterruptedException error)
+                                         (Utils/exceptionCauseIsInstanceOf java.io.InterruptedIOException error))
+                                     (log-message "Got interrupted excpetion shutting thread down...")
```
abellina: nitpick: fix spelling of exception.

PR #1578, opened 2016-07-20
STORM-1988 Kafka Offset not showing due to bad classpath

HeartSaVioR:
Adding prefix and/or postfix between * just doesn't work on classpath, which I didn't recognize that.

Confirmed from CentOS6 with OpenJDK 1.8.0_91 / Oracle JDK 1.8.0_77 and OSX 10.11.5 with Oracle JDK 1.8.0_66.

Patch can be ported back to 1.x-branch easily.

### DIFF
--- a/bin/storm-kafka-monitor
+++ b/bin/storm-kafka-monitor
@@ -40,4 +40,4 @@ else
   JAVA="$JAVA_HOME/bin/java"
 fi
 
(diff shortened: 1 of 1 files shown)


harshach:
+1

ptgoetz:
+1

PR #1579, opened 2016-07-20
STORM-1989: X-Frame-Options support for Storm UI

tibkiss:


ptgoetz:
+1

tibkiss:
Thanks @ptgoetz for the review!

The Travis test failure seems unrelated to me: 
The base version of the patch also fails with test errors.

ptgoetz:
@tibkiss Yes. I tested locally and the build succeeded.

harshach:
@tibkiss can you post a patch against 1.x-branch as well. Also how does this works with CORS enabled.

tibkiss:
@harshach : The patch is not interfering with CORS filter. It is an additional filter addressing the ClickJacking vulnerability. 

@ptgoetz : Super thanks for merging the PR to both branches. Next time I'll address the delivery to additional branches quicker.

PR #1580, opened 2016-07-20
[STORM-1990] Make some constant unvisible

darionyaphet:
[STORM-1990 Make some constant unvisible](https://issues.apache.org/jira/browse/STORM-1990)

Some constant used as default config , make them private and unvisible .


abhishekagarwal87:
@darionyaphet - are these changes necessary?

PR #1583, opened 2016-07-21
STORM-1979: Storm Druid Connector implementation.

omkreddy:


omkreddy:
This implementation uses Druid's Tranquility library (https://github.com/druid-io/tranquility)
to send to druid store. Some of the concepts are borrowed from Tranquility storm connector implementation (https://github.com/druid-io/tranquility/blob/master/docs/storm.md)

@harshach pinging for review

external/storm-druid/README.md
```diff
@@ -0,0 +1,125 @@
...
+
+```java
+
+public class SampleDruidBeamFactoryImpl implements DruidBeamFactory<Map<String, Object>> {
+
+    @Override
+    public Beam<Map<String, Object>> makeBeam(Map<?, ?> conf, IMetricsContext metrics) {
+
+
+        final String indexService = "druid/overlord"; // Your overlord's druid.service
```
harshach: can you add some comments around these two.

external/storm-druid/README.md
```diff
@@ -0,0 +1,125 @@
...
+        final String discoveryPath = "/druid/discovery"; // Your overlord's druid.discovery.curator.path
+        final String dataSource = "test";
```
harshach: same here why are we calling dataSource test?

external/storm-druid/pom.xml
```diff
@@ -0,0 +1,88 @@
...
+        </dependency>
+        <dependency>
+            <groupId>io.druid</groupId>
+            <artifactId>tranquility-core_2.11</artifactId>
+            <version>0.8.2</version>
+        </dependency>
+        <dependency>
+            <groupId>org.scala-lang</groupId>
+            <artifactId>scala-library</artifactId>
+            <version>2.11.8</version>
```
harshach: we should be making these provided

external/storm-druid/src/main/java/org/apache/storm/druid/bolt/DruidBeamBolt.java
```diff
@@ -0,0 +1,115 @@
...
+public class DruidBeamBolt<E> extends BaseRichBolt {
+    private static final Logger LOG = LoggerFactory.getLogger(DruidBeamBolt.class);
+
+    private volatile  OutputCollector collector;
+    private DruidBeamFactory<E> beamFactory = null;
+    private int batchSize;
+    private Tranquilizer<E> tranquilizer = null;
+
+    public DruidBeamBolt(DruidBeamFactory<E> beamFactory) {
+        this(beamFactory, 2000);
```
harshach: lets remove this constructor as users might not notice that we are setting the batchSize to 2000.

ptgoetz:
@omkreddy I haven't had a chance to review it fully yet, but you can add me as a committer sponsor.

external/storm-druid/src/main/java/org/apache/storm/druid/bolt/DruidBeamBolt.java
```diff
@@ -0,0 +1,115 @@
...
+
+    }
+
+    @Override
+    public void execute(final Tuple tuple) {
+      Future future = tranquilizer.send((E)tuple.getValue(0));
+      future.addEventListener(new FutureEventListener() {
+          @Override
+          public void onFailure(Throwable cause) {
+              if(cause instanceof MessageDroppedException) {
```
harshach: should we not retry here? instead of calling ack

satishd: > MessageDroppedException: Exception indicating that a message was dropped "on purpose" by the beam. This is not a recoverable exception and so the message must be discarded.

These tuples can be acked and send all these discarded tuples to a given stream so that user can know which tuples are discarded and take an appropriate action.

external/storm-druid/src/main/java/org/apache/storm/druid/bolt/DruidBeamBolt.java
```diff
@@ -0,0 +1,115 @@
...
+                  synchronized (collector) {
```
satishd: Does `IOutputCollector#ack/fail` need synchronization?

satishd: We may not need any synchronization here and at other places as well.

[STORM-841](https://issues.apache.org/jira/browse/STORM-841) confirms it is threadsafe.

external/storm-druid/src/main/java/org/apache/storm/druid/bolt/DruidBeamBolt.java
omkreddy: currently druid logs discarded messages.  Also I need to find way to handle this in Trident bolt. I will do this when need arises.

omkreddy:
@ptgoetz Thanks. I added you as committer sponsor.

external/storm-druid/src/main/java/org/apache/storm/druid/bolt/DruidBeamBolt.java
omkreddy: removed the constructor.  Now users has to pass the batchSize. as of now we have only one config value. will add config class when need arises.

external/storm-druid/src/main/java/org/apache/storm/druid/bolt/DruidBeamBolt.java
harshach: I am +1 on @satishd recommendation. Atleast we can drive through configuration i.e druid can discard messages and we can provide discard stream of sorts via config if users configures enableDiscardStream and discardStream="druid-discard-stream" than you write the tuples there. Logging on druid is great but this will allow users to act on them immediately.

external/storm-druid/pom.xml
harshach: I take back the earlier comment. sorry about that,  the dependencies shouldn't be provided as we don't want to package them into the druid jar.
also can we drive these versions from parent pom.xml so that we can configure better. Also make the scala version 2.10.5 unless druid recommends 2.11.8.

external/storm-druid/src/main/java/org/apache/storm/druid/trident/DruidBeamState.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    public void send(List<E> events ) {
+        try {
+            LOG.info("Sending %d events", events.size());
+            Await.result(beam.sendBatch(JavaConversions.collectionAsScalaIterable(events).toList()));
+        } catch (Exception e) {
+            final String errorMsg = "Failed in writing messages to Druid";
+            LOG.error(errorMsg, e);
+            throw new RuntimeException(errorMsg);
```
harshach: do we want to throw the run-time exception. What if this is a temporary failure and on a re-try we might succeed in those cases throwing run time exception will kill the worker JVM itself.

external/storm-druid/src/main/java/org/apache/storm/druid/trident/DruidBeamState.java
```diff
@@ -0,0 +1,73 @@
...
+        }
+    }
+
+    public void close() {
+        try {
+            Await.result(beam.close());
+        } catch (Exception e) {
+            final String errorMsg = "Error while closing Druid beam client";
+            LOG.error(errorMsg, e);
+            throw new RuntimeException(errorMsg);
```
harshach: same as above. we shouldn't be throwing run-time exceptions when close is called.

external/storm-druid/src/main/java/org/apache/storm/druid/trident/DruidBeamState.java
```diff
@@ -0,0 +1,73 @@
...
+        }
+    }
+
+    @Override
+    public void beginCommit(Long txid) {
```
harshach: @arunmahadevan I remember you recommended that we should make all trident states to implement beginCommit. Can you add details if we can do this here.

external/storm-druid/src/test/java/org/apache/storm/druid/SampleDruidBeamFactoryImpl.java
```diff
@@ -0,0 +1,116 @@
...
+
+    public SampleDruidBeamFactoryImpl(Map<String, Object> factoryConf) {
+        this.factoryConf = factoryConf; // This can be used to pass config values
+    }
+
+    @Override
+    public Beam<Map<String, Object>> makeBeam(Map<?, ?> conf, IMetricsContext metrics) {
+
+
+        final String indexService = "druid/overlord"; // Your overlord's druid.service
```
harshach: should we make these as configurable options ? and alos there is no leading / here

external/storm-druid/src/main/java/org/apache/storm/druid/trident/DruidBeamState.java
```diff
@@ -0,0 +1,73 @@
...
+    Beam<E> beam = null;
+
+    public DruidBeamState(Beam<E> beam) {
+        this.beam = beam;
+    }
+
+    public void send(List<E> events ) {
+        try {
+            LOG.info("Sending %d events", events.size());
+            Await.result(beam.sendBatch(JavaConversions.collectionAsScalaIterable(events).toList()));
```
harshach: looks like indentation missed here

external/storm-druid/src/main/java/org/apache/storm/druid/bolt/DruidBeamBolt.java
```diff
@@ -0,0 +1,103 @@
...
+                beamFactory.makeBeam(stormConf, context),
+                batchSize,
+                Tranquilizer.DefaultMaxPendingBatches(),
+                Tranquilizer.DefaultLingerMillis());
+        this.tranquilizer.start();
+    }
+
+    @Override
+    public void execute(final Tuple tuple) {
+      Future future = tranquilizer.send((E)tuple.getValue(0));
```
harshach: we are looking at only the first value in tuple. What is the data format that we are expecting for the user to send. It might be confusing for the user if they send two values in the tuple but we only take the first one. May be we should mapper interface like hbase or hive bolts.

external/storm-druid/src/main/java/org/apache/storm/druid/trident/DruidBeamStateUpdater.java
```diff
@@ -0,0 +1,41 @@
...
+import java.util.List;
+
+public class DruidBeamStateUpdater<E> extends BaseStateUpdater<DruidBeamState<E>> {
+
+    @Override
+    public void updateState(DruidBeamState<E> state, List<TridentTuple> tuples, TridentCollector collector) {
+        List<E> list = new LinkedList<>();
+
+        for (TridentTuple tuple: tuples) {
+            list.add((E)tuple.getValue(0));
```
harshach: same as above its not clear why are we only picking the first value in the tuple.

external/storm-druid/src/main/java/org/apache/storm/druid/trident/DruidBeamState.java
arunmahadevan: @harshach most of the trident states ignore the `beginCommit` and `commit`, because it may be very difficult to handle idempotent updates in a generic state. beginCommit and commit is currently implemented only in HdfsState. The downside is that such states would provide only at-least once semantics.

However key value based stores can provide an `IBackingMap` implementation whereby exactly once semantics can be supported. For e.g. the Redis trident implementation provides a RedisMapState. 

If druid supports key/value based abstraction, the connector should probably provide a MapState implementation. If not the beginCommit and commit should be handled within the DruidBeamState if we are looking to provide exactly once.

external/storm-druid/src/main/java/org/apache/storm/druid/trident/DruidBeamState.java
omkreddy: I Agree.  Druid library has internal retry mechanism. Now i not throwing  runtime exception.

external/storm-druid/src/main/java/org/apache/storm/druid/trident/DruidBeamState.java
omkreddy: Druid Tranquility library itself  does not support exactly-once semantics and it is not key-value store.

external/storm-druid/src/main/java/org/apache/storm/druid/bolt/DruidBeamBolt.java
omkreddy: implemented above discussed mechanism. now uses have option to process dropped/discarded messages.

external/storm-druid/src/main/java/org/apache/storm/druid/bolt/DruidBeamBolt.java
omkreddy: Druid tranquility client library take a generic type for the events to send. 
The bolt need to know the type of the event. By default the bolt expects to receive tuples in which the zeroth element is event type. Now this event position in tuple list is made configurable. Users can specify the event position using DruidConfig.

omkreddy:
@harshach @satishd I have addressed review comments. Pl review.

harshach:
@omkreddy you need to upmerge the PR.

external/storm-druid/src/main/java/org/apache/storm/druid/bolt/DruidBeamBolt.java
harshach: @omkreddy I need to look into this more to understand. But we shouldn't be saying that the first tuple is the one that goes into druid. Take a look other connectors in hbase, hive see if mapper is something we can use here.

omkreddy: @harshach I discussed this approach with @satishd  and I felt this way users can send whatever event type they want.  

upmerged the PR.

satishd: @omkreddy I guess we have discussed to add a default mapper implementation to return a field for the give name.  But user can always implement Mapper interface and return respective object to be sent.

external/storm-druid/src/main/java/org/apache/storm/druid/trident/DruidBeamStateUpdater.java
```diff
@@ -0,0 +1,48 @@
...
+    public void updateState(DruidBeamState<E> state, List<TridentTuple> tuples, TridentCollector collector) {
+        List<E> discardedTuples = state.update(tuples, collector);
+        processDiscardedTuples(discardedTuples);
+    }
+
+    /**
+     * Users can override this method to  process the discarded Tuples
+     * @param discardedTuples
+     */
+    private void processDiscardedTuples(List<E> discardedTuples) {
```
satishd: This should be `protected` as users can extend this and add their own functionality to handle discard tuples.

external/storm-druid/src/main/java/org/apache/storm/druid/trident/DruidBeamStateUpdater.java
```diff
@@ -0,0 +1,48 @@
...
+        List<E> discardedTuples = state.update(tuples, collector);
+        processDiscardedTuples(discardedTuples);
+    }
+
+    /**
+     * Users can override this method to  process the discarded Tuples
+     * @param discardedTuples
+     */
+    protected void processDiscardedTuples(List<E> discardedTuples) {
+        LOG.debug("discarded messages : %s" , discardedTuples);
```
satishd: Logger format convention is {} instead of %s. You may want to have 
`LOG.debug("discarded messages : [{}]" , discardedTuples)`

external/storm-druid/src/main/java/org/apache/storm/druid/bolt/DruidBeamBolt.java
omkreddy: @satishd Thanks. Earlier I misunderstood your suggestion.
@harshach I added a mapper interface and default implementation.
Thanks for the  review. Updated the PR.

external/storm-druid/src/main/java/org/apache/storm/druid/bolt/DruidConfig.java
```diff
@@ -0,0 +1,114 @@
...
+        return new Builder();
+    }
+
+    public static class Builder {
+        private int maxBatchSize = Tranquilizer.DefaultMaxBatchSize();
+        private int maxPendingBatches = Tranquilizer.DefaultMaxPendingBatches();
+        private long lingerMillis = Tranquilizer.DefaultLingerMillis();
+        private boolean blockOnFull =  Tranquilizer.DefaultBlockOnFull();
+        private boolean enableDiscardStream = false;
+        private String discardStreamId ="druid-discard-stream";
```
satishd: "druid-discard-stream" can be declared as a public constant like DEFAULT_DISCARD_STREAM_ID in `DruidConfig` class, so that users can use thos stream id while building the topology. 

I prefer to remove `enableDiscardStream` and have only `discardStreamId` which should be given by the user if user wants discarded events to be forwarded to that stream. This API makes it simpler to set only the discardStreamId.

satishd: @harshachlet me know your opinion on the above comment.

satishd:
@omkreddy You can add me as committer sponsor for this module.

omkreddy:
@satishd added you as committer sponsor.  also addressed review comments

harshach:
+1. Nice work @omkreddy .

external/storm-druid/src/main/java/org/apache/storm/druid/bolt/DruidConfig.java
harshach: @satishd sorry missed it. That makes sense. @omkreddy can you address the above as well.

harshach:
missed @satishd suggestions above. Once those addressed I am +1.

external/storm-druid/src/main/java/org/apache/storm/druid/bolt/TupleDruidEventMapper.java
```diff
@@ -0,0 +1,42 @@
...
+
+import org.apache.storm.tuple.ITuple;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Converts {@link ITuple} to Event
+ */
+public final class TupleDruidEventMapper<E> implements ITupleDruidEventMapper {
+    private static final Logger LOG = LoggerFactory.getLogger(TupleDruidEventMapper.class);
```
satishd: @omkreddy You may want to remove this as it is never used.

external/storm-druid/src/main/java/org/apache/storm/druid/bolt/TupleDruidEventMapper.java
```diff
@@ -0,0 +1,42 @@
...
+
+    public static final String DEFAULT_FIELD_NAME = "event";
+
+    public TupleDruidEventMapper() {
+    }
+
+    @Override
+    public E getEvent(ITuple tuple) {
+        return (E) tuple.getValueByField(DEFAULT_FIELD_NAME);
```
satishd: You may want to take the field name in constructor and use it like `tuple.getValueByField(eventFieldName);`. Otherwise, user should always need to extend this class and override `getEvent(ITuple)` method.

external/storm-druid/src/main/java/org/apache/storm/druid/bolt/DruidBeamBolt.java
```diff
@@ -0,0 +1,111 @@
...
+ * Some of the concepts are borrowed from Tranquility storm connector implementation.
+ * (https://github.com/druid-io/tranquility/blob/master/docs/storm.md)
+ *
+ * By default this Bolt expects to receive tuples in which "event" field gives your event type.
+ * This logic can be changed by implementing ITupleDruidEventMapper interface.
+ * <p/>
+ *
+ */
+public class DruidBeamBolt<E> extends BaseRichBolt {
+    private static final Logger LOG = LoggerFactory.getLogger(DruidBeamBolt.class);
```
satishd: `LOG` is never used. You can keep it if you want to use some logging in this class.

satishd:
@omkreddy Let me know once the comments are addressed.

omkreddy:
@satishd Thanks for the review. addressed review comments.

external/storm-druid/src/main/java/org/apache/storm/druid/bolt/DruidConfig.java
```diff
@@ -0,0 +1,112 @@
...
+public class DruidConfig implements Serializable {
+
+    public static final String DEFAULT_DISCARD_STREAM_ID = "druid-discard-stream";
+
+    //Tranquilizer configs for DruidBeamBolt
+    private int maxBatchSize;
+    private int maxPendingBatches;
+    private long lingerMillis;
+    private boolean blockOnFull;
+    private boolean enableDiscardStream;
```
satishd: @omkreddy We do not need this as it was suggested in earlier comments. User would give only discardStreamId if discarded events need to be handled.

external/storm-druid/src/main/java/org/apache/storm/druid/bolt/DruidBeamBolt.java
```diff
@@ -0,0 +1,110 @@
...
+
+    @Override
+    public void execute(final Tuple tuple) {
+      Future future = tranquilizer.send((druidEventMapper.getEvent(tuple)));
+      future.addEventListener(new FutureEventListener() {
+          @Override
+          public void onFailure(Throwable cause) {
+              if (cause instanceof MessageDroppedException) {
+                  collector.ack(tuple);
+                  if (druidConfig.isEnableDiscardStream())
```
satishd: You can check `druidConfig.getDiscardStreamId() != null` and emit to that stream, no need to have `isEnableDiscardStream`.

omkreddy: @satishd  Since this is internal implementation, I decided to use existing boolean variable.  I felt boolean check is better than null check.  Updated the PR.

satishd: @omkreddy It is not really an internal implementation. `druidConfig.isEnableDiscardStream()` was still using `enableDiscardStream`, which should be set by user. Idea is to avoid that and I can see latest changes reflect on that.

external/storm-druid/README.md
```diff
@@ -0,0 +1,143 @@
+# Storm Druid Bolt and TridentState
+
+This module provides core Storm and Trident bolt implementations for writing data to [Druid] (http://druid.io/) data store.
```
satishd: @omkreddy Remove space between [] and () to make MD generate a hyperlink. Currently it shows as text. Same with other places in this doc.

satishd:
+1 after addressing minor comment in README doc.

omkreddy:
@satishd  addressed review comments and squashed the commits.

satishd:
@omkreddy I will merge it in tomorrow w.r.t 24hrs wait policy after last commit. Thanks for your work.

satishd:
Thanks @omkreddy , pushed this change to master and 1.x-branch, added you as a contributor.

PR #1584, opened 2016-07-21
STORM-1992 Added artifacts to make storm.js into an npm package

mzbyszynski:
Posting for feedback! Please let me know if I am on the right track. This PR relates to [STORM-1992](https://issues.apache.org/jira/browse/STORM-1992)

``` bash
$ mvn deploy -P npmjs
```

Would publish the package to npmjs.com as **storm-multilang-js** (I could not think of a better name that was not taken) along with the other deployments the build performs. It would have be manually executed right now because you need to have run `npm login` before hand to enter your credentials. I have not found a clean way to make this automated, but if that is a requirement I can probably get it working in an unclean way.

I published the package with a bogus version number here as an example: https://www.npmjs.com/package/storm-multilang-js

I added a `stormjs` utility to make it easier to call a js spout or bolt from a topology by taking care of some of the boilerplate. I also cleaned up `storm.js` a little bit by adding missing semi-colons and making string quotation consistent.

I'm unsure about the README.md file for the npm package. Should I add more to it or should it just point to documentation living somewhere else?

Thank you!


ptgoetz:
My apologies for letting this go without review for so long.

+1

PR #1586, opened 2016-07-23
STORM-1839: Storm spout implementation for Amazon Kinesis Streams.

priyank5485:


external/storm-kinesis/src/main/java/org/apache/storm/kinesis/spout/Config.java
```diff
@@ -0,0 +1,166 @@
...
+package org.apache.storm.kinesis.spout;
+
+import com.amazonaws.services.kinesis.model.ShardIteratorType;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.Serializable;
+import java.util.Date;
+
+public class Config implements Serializable {
```
harshach: given that this is a public class going to be used by devs. Can we rename this as KinesisConfig

external/storm-kinesis/src/main/java/org/apache/storm/kinesis/spout/Config.java
```diff
@@ -0,0 +1,166 @@
...
+    public KinesisConnectionInfo getKinesisConnectionInfo () {
+        return kinesisConnectionInfo;
+    }
+
+    public Long getMaxUncommittedRecords () {
+        return maxUncommittedRecords;
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
harshach: Do we need equals and hash code . Not sure where do we do comparisons of config object.

external/storm-kinesis/src/main/java/org/apache/storm/kinesis/spout/ExponentialBackoffRetrier.java
```diff
@@ -0,0 +1,164 @@
...
+        // if first failure add it to the count map
+        if (!failCounts.containsKey(messageId)) {
+            failCounts.put(messageId, 0L);
+        }
+        // increment the fail count as we started with 0
+        Long failCount = failCounts.get(messageId);
+        failCounts.put(messageId, ++failCount);
+        // if fail count is greater than maxRetries, discard or ack. for e.g. for maxRetries 3, 4 failures are allowed at maximum
+        if (failCount > maxRetries) {
+            LOG.debug("maxRetries reached so dropping " + messageId);
```
harshach: can you make this as warn as it can be useful to see this in the logs by default

external/storm-kinesis/src/main/java/org/apache/storm/kinesis/spout/ExponentialBackoffRetrier.java
```diff
@@ -0,0 +1,164 @@
...
+            failCounts.remove(messageId);
+            return false;
+        }
+        // if reached so far, add it to the set of messages waiting to be retried with next retry time based on how many times it failed
+        retryTimes.put(messageId, getRetryTime(failCount));
+        retryMessageSet.add(messageId);
+        LOG.debug("Scheduled " + messageId + " for retry at " + retryTimes.get(messageId) + " and retry attempt " + failCount);
```
harshach: I recommend this to be at warn level as well.

external/storm-kinesis/src/main/java/org/apache/storm/kinesis/spout/ExponentialBackoffRetrier.java
```diff
@@ -0,0 +1,164 @@
...
+        if (maxRetries < 0) {
+            throw new IllegalArgumentException("maxRetries cannot be negative.");
+        }
+    }
+    @Override
+    public boolean failed(KinesisMessageId messageId) {
+        LOG.debug("Handling failed message " + messageId);
+        // if maxRetries is 0, dont retry and return false as per interface contract
+        if (maxRetries == 0) {
+            LOG.debug("maxRetries set to 0. Hence not queueing " + messageId);
```
harshach: warn here as well

external/storm-kinesis/src/main/java/org/apache/storm/kinesis/spout/ExponentialBackoffRetrier.java
```diff
@@ -0,0 +1,164 @@
...
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.SortedSet;
+import java.util.TreeSet;
+
+public class ExponentialBackoffRetrier implements FailedMessageRetryHandler, Serializable {
+    private static final Logger LOG = LoggerFactory.getLogger(ExponentialBackoffRetrier.class);
+    // Wait interfal for retrying after first failure
+    private final Long initialDelayMillis;
```
harshach: should we have any defaults for these variables.

external/storm-kinesis/src/main/java/org/apache/storm/kinesis/spout/ExponentialBackoffRetrier.java
```diff
@@ -0,0 +1,164 @@
...
+    // Base for exponential function in seconds for retrying for second, third and so on failures
+    private final Long baseSeconds;
+    // Maximum number of retries
+    private final Long maxRetries;
+    // map to track number of failures for each kinesis message that failed
+    private Map<KinesisMessageId, Long> failCounts = new HashMap<>();
+    // map to track next retry time for each kinesis message that failed
+    private Map<KinesisMessageId, Long> retryTimes = new HashMap<>();
+    // sorted set of records to be retrued based on retry time. earliest retryTime record comes first
+    private SortedSet<KinesisMessageId> retryMessageSet = new TreeSet<>(new RetryTimeComparator());
```
harshach: are we going to store the same message in retryMessageSet and retryTimes?

external/storm-kinesis/src/main/java/org/apache/storm/kinesis/spout/Config.java
priyank5485: Will change it.

external/storm-kinesis/src/main/java/org/apache/storm/kinesis/spout/Config.java
priyank5485: We dont need equals and hashcode. I will remove it.

external/storm-kinesis/src/main/java/org/apache/storm/kinesis/spout/ExponentialBackoffRetrier.java
priyank5485: I will change it

external/storm-kinesis/src/main/java/org/apache/storm/kinesis/spout/ExponentialBackoffRetrier.java
priyank5485: I will change it

external/storm-kinesis/src/main/java/org/apache/storm/kinesis/spout/ExponentialBackoffRetrier.java
priyank5485: I will change it.

external/storm-kinesis/src/main/java/org/apache/storm/kinesis/spout/ExponentialBackoffRetrier.java
priyank5485: ExponentialBackoffRetrier has two constructors. The no args constructor will set the default values for these variables.

external/storm-kinesis/src/main/java/org/apache/storm/kinesis/spout/ExponentialBackoffRetrier.java
priyank5485: No. Messages are not stored anywhere in ExponentialBackoffRetrier. retryTimes is just used to implement the comparator interface that will be used by retryMessageSet TreeSet to order the messages on their next retry time. You can check out the RetryTimeComparator in this class

external/storm-kinesis/src/main/java/org/apache/storm/kinesis/spout/KinesisConnectionInfo.java
```diff
@@ -0,0 +1,137 @@
...
+        }
+        if (region == null) {
+            throw new IllegalArgumentException("region cannot be null");
+        }
+        serializedKinesisCredsProvider = getKryoSerializedBytes(credentialsProvider);
+        serializedkinesisClientConfig = getKryoSerializedBytes(clientConfiguration);
+        this.recordsLimit = recordsLimit;
+        this.region = region;
+
+        this.credentialsProvider = null;
```
harshach: we don't need to explicitly assign a null to these.

external/storm-kinesis/src/main/java/org/apache/storm/kinesis/spout/KinesisConnectionInfo.java
```diff
@@ -0,0 +1,137 @@
...
+
+    private Object getKryoDeserializedObject (final byte[] ser) {
+        final Kryo kryo = new Kryo();
+        final Input input = new Input(new ByteArrayInputStream(ser));
+        kryo.setInstantiatorStrategy(new StdInstantiatorStrategy());
+        return kryo.readClassAndObject(input);
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
harshach: we probably can remove these equals, hash here as well.

external/storm-kinesis/src/main/java/org/apache/storm/kinesis/spout/KinesisRecordsManager.java
```diff
@@ -0,0 +1,566 @@
...
+                            "infinitely");
+                }
+            } else {
+                LOG.debug("failedPerShard does not contain " + failedMessageId + ". This should never happen.");
+            }
+        }
+        LOG.debug("No failed record to emit for now. Hence will try to emit new records");
+        // if maximum uncommitted records count has reached, so dont emit any new records and return
+        if (!(getUncommittedRecordsCount() < config.getMaxUncommittedRecords())) {
+            LOG.debug("maximum uncommitted records count has reached. so not emitting any new records and returning");
```
harshach: this should be warn

external/storm-kinesis/src/main/java/org/apache/storm/kinesis/spout/KinesisRecordsManager.java
```diff
@@ -0,0 +1,566 @@
...
+            return;
+        }
+        // early return as no shard is assigned - probably because number of executors > number of shards
+        if (toEmitPerShard.isEmpty()) {
+            LOG.debug("No shard is assigned to this task. Hence not emitting any tuple.");
```
harshach: same here

external/storm-kinesis/src/main/java/org/apache/storm/kinesis/spout/KinesisRecordsManager.java
```diff
@@ -0,0 +1,566 @@
...
+            // if the retry service returns a message that is not in failed set then ignore it. should never happen
+            BigInteger failedSequenceNumber = new BigInteger(failedMessageId.getSequenceNumber());
+            if (failedPerShard.containsKey(failedMessageId.getShardId()) && failedPerShard.get(failedMessageId.getShardId()).contains(failedSequenceNumber)) {
+                if (!failedandFetchedRecords.containsKey(failedMessageId)) {
+                    fetchFailedRecords(failedMessageId);
+                }
+                if (emitFailedRecord(collector, failedMessageId)) {
+                    failedPerShard.get(failedMessageId.getShardId()).remove(failedSequenceNumber);
+                    config.getFailedMessageRetryHandler().failedMessageEmitted(failedMessageId);
+                    return;
```
harshach: if we can make multiple returns into one or two we should try.

external/storm-kinesis/src/main/java/org/apache/storm/kinesis/spout/KinesisRecordsManager.java
```diff
@@ -0,0 +1,566 @@
...
+            return;
+        }
+
+        if (shouldFetchNewRecords()) {
+            fetchNewRecords();
+        }
+        emitNewRecord(collector);
+    }
+
+    void ack (KinesisMessageId kinesisMessageId) {
```
harshach: if we can document via comments on this ack logic that would be great.

external/storm-kinesis/README.md
```diff
@@ -0,0 +1,139 @@
...
+As you can see above the spout takes an object of Config in its constructor. The constructor of Config takes 8 objects as explained below.
+
+#### `String` streamName
+name of kinesis stream to consume data from
+
+#### `ShardIteratorType` shardIteratorType
+3 types are supported - TRIM_HORIZON(beginning of shard), LATEST and AT_TIMESTAMP. By default this argument is ignored if state for shards 
+is found in zookeeper. Hence they will apply the first time a topology is started. If you want to use any of these in subsequent runs of the topology, you 
+will need to clear the state of zookeeper node used for storing sequence numbers
+#### `RecordToTupleMapper` recordToTupleMapper
```
harshach: line break

external/storm-kinesis/src/main/java/org/apache/storm/kinesis/spout/KinesisRecordsManager.java
```diff
@@ -0,0 +1,566 @@
...
+        try {
+            GetRecordsResult getRecordsResult = fetchRecords(shardIterator);
+            if (getRecordsResult != null) {
+                List<Record> records = getRecordsResult.getRecords();
+                LOG.debug("Records size from fetchFailedRecords is " + records.size());
+                // update the shard iterator to next one in case this fetch does not give the message.
+                shardIteratorPerFailedMessage.put(kinesisMessageId, getRecordsResult.getNextShardIterator());
+                if (records.size() == 0) {
+                    LOG.debug("No records returned from kinesis. Hence sleeping for 1 second");
+                    Thread.sleep(1000);
```
harshach: should this be configurable ? or probably a candidate for exponential back off . we can address this in follow-up jira

external/storm-kinesis/src/main/java/org/apache/storm/kinesis/spout/KinesisRecordsManager.java
```diff
@@ -0,0 +1,566 @@
...
+                        KinesisMessageId current = new KinesisMessageId(kinesisMessageId.getStreamName(), kinesisMessageId.getShardId(), record.getSequenceNumber());
+                        if (failedPerShard.get(kinesisMessageId.getShardId()).contains(new BigInteger(current.getSequenceNumber()))) {
+                            failedandFetchedRecords.put(current, record);
+                            shardIteratorPerFailedMessage.remove(current);
+                        }
+                    }
+                }
+            }
+        } catch (InterruptedException ie) {
+            LOG.debug("Thread interrupted while sleeping", ie);
```
harshach: lets make these warns for all the exceptions unless you think this might result lot of noise. Even than it will be helpful to print them given that these are critical errors.

external/storm-kinesis/src/main/java/org/apache/storm/kinesis/spout/KinesisRecordsManager.java
```diff
@@ -0,0 +1,566 @@
...
+        } catch (ExpiredIteratorException ex) {
+            LOG.debug("shardIterator for failedRecord " + kinesisMessageId + " has expired. Refreshing shardIterator");
+            refreshShardIteratorForFailedRecord(kinesisMessageId);
+        } catch (ProvisionedThroughputExceededException pe) {
+            try {
+                LOG.debug("ProvisionedThroughputExceededException occured. Check your limits. Sleeping for 1 second.", pe);
+                Thread.sleep(1000);
```
harshach: same as before

external/storm-kinesis/src/main/java/org/apache/storm/kinesis/spout/KinesisRecordsManager.java
```diff
@@ -0,0 +1,566 @@
...
+
+    private void shutdownCurator () {
+        curatorFramework.close();
+    }
+
+    private void shutdownKinesisClient () {
+        kinesisClient.shutdown();
+    }
+
+}
```
harshach: This is big file thats doing lot of things. We should some of these methods into Utils class or something  like zk related methods into its own class.

external/storm-kinesis/src/main/java/org/apache/storm/kinesis/spout/ZkInfo.java
```diff
@@ -0,0 +1,153 @@
...
+    // number of retry attempts for zk
+    private final Integer retryAttempts;
+    // time to sleep between retries in milliseconds
+    private final Integer retryIntervalMs;
+
+    /**
+     * Default constructor that uses defaults for a local setup
+     */
+    public ZkInfo () {
+        this("localhost:2181", "/kinesisOffsets", 20000, 15000, 10000L, 3, 2000);
```
harshach: I think this is useful for testing but I can see users tripping over this :) so can we remove this.

external/storm-kinesis/src/main/java/org/apache/storm/kinesis/spout/ZkInfo.java
```diff
@@ -0,0 +1,153 @@
...
+                ", zkNode='" + zkNode + '\'' +
+                ", sessionTimeoutMs=" + sessionTimeoutMs +
+                ", connectionTimeoutMs=" + connectionTimeoutMs +
+                ", commitIntervalMs=" + commitIntervalMs +
+                ", retryAttempts=" + retryAttempts +
+                ", retryIntervalMs=" + retryIntervalMs +
+                '}';
+    }
+
+    @Override
```
harshach: remove these as well.

harshach:
@priyank5485 looks great!. Mostly nit-picks. I'll go over the retry handle one more time and add any comments if needed.

external/storm-kinesis/src/main/java/org/apache/storm/kinesis/spout/KinesisConnectionInfo.java
priyank5485: Will change it

external/storm-kinesis/src/main/java/org/apache/storm/kinesis/spout/KinesisConnectionInfo.java
priyank5485: Will change it

external/storm-kinesis/src/main/java/org/apache/storm/kinesis/spout/KinesisRecordsManager.java
priyank5485: I thought about it but if i try to reduce the return statements it will create more if else statements. If its not a big issue we can keep as is.

external/storm-kinesis/src/main/java/org/apache/storm/kinesis/spout/KinesisRecordsManager.java
priyank5485: I documented the logic in the commit method since thats where it writes to zookeeper. Let me know if that is not enough or if i should move it to somewhere else.

external/storm-kinesis/README.md
priyank5485: Will change it

external/storm-kinesis/src/main/java/org/apache/storm/kinesis/spout/KinesisRecordsManager.java
priyank5485: I would prefer this to be not configurable. Reason is kinesis api somewhere mentioned in their documentation that if you dont receive any records it is good to sleep for a second. Reason is that they some limits for GetRecords request per second and it will throw an exception if that limit is reached or violated. Although I am catching that exception I wanted to avoid that. Making it configurable and go beyond 1 second can slow down the spout in some cases as well. For example, sometimes it takes multiple requests to get to the portion of the shard where the message at a sequence number you want to read is present. In that case it does not make sense to backoff in an exponential fashion. You just want to make sure you get there sooner without violating the limit.

external/storm-kinesis/src/main/java/org/apache/storm/kinesis/spout/ZkInfo.java
priyank5485: Will remove it

priyank5485:
@harshach Thanks for the review. I have addressed your comments and pushed a commit.

external/storm-kinesis/src/main/java/org/apache/storm/kinesis/spout/KinesisRecordsManager.java
harshach: @priyank5485 looks like this method is going through a switch statement of sort, i.e its checking bunch if conditions and doing actions on that part. It will be great if you can add why you are doing it. For the future reference it will be important as this part is most important on when to ack a message.

external/storm-kinesis/src/main/java/org/apache/storm/kinesis/spout/KinesisRecordsManager.java
harshach: @priyank5485 do you think its possible to move some of these methods into its Utils file or some sorts.

harshach:
@priyank5485 left couple of comments

priyank5485:
@harshach I added some more comments in ack method and commit method for acking logic. I also moved code interacting with zk and kinesis in to its own class to make KinesisRecordsManager smaller. Please review when you get a chance

external/storm-kinesis/src/main/java/org/apache/storm/kinesis/spout/KinesisRecordsManager.java
```diff
@@ -0,0 +1,449 @@
...
+        try {
+            GetRecordsResult getRecordsResult = kinesisConnection.fetchRecords(shardIterator);
+            if (getRecordsResult != null) {
+                List<Record> records = getRecordsResult.getRecords();
+                LOG.debug("Records size from fetchFailedRecords is " + records.size());
+                // update the shard iterator to next one in case this fetch does not give the message.
+                shardIteratorPerFailedMessage.put(kinesisMessageId, getRecordsResult.getNextShardIterator());
+                if (records.size() == 0) {
+                    LOG.warn("No records returned from kinesis. Hence sleeping for 1 second");
+                    Thread.sleep(1000);
```
harshach: Sorry missed these earlier. Any reason for it to sleep exactly one sec or should it be configurable.

harshach:
+1. Build failures not related to the patch.

external/storm-kinesis/src/main/java/org/apache/storm/kinesis/spout/KinesisRecordsManager.java
priyank5485: I would prefer this to be not configurable. Reason is kinesis api somewhere mentioned in their documentation that if you dont receive any records it is good to sleep for a second. Reason is that they some limits for GetRecords request per second and it will throw an exception if that limit is reached or violated. Although I am catching that exception I wanted to avoid that. Making it configurable and go beyond 1 second can slow down the spout in some cases as well. For example, sometimes it takes multiple requests to get to the portion of the shard where the message at a sequence number you want to read is present. In that case it does not make sense to backoff in an exponential fashion. You just want to make sure you get there sooner without violating the limit.

revans2:
@harshach actually some of the build errors are related see https://issues.apache.org/jira/browse/STORM-2021

PR #1587, opened 2016-07-24
[STORM-1991] Support auto.commit.interval in Kafka Client

darionyaphet:
[STORM-1991 Support auto.commit.interval in Kafka Client](https://issues.apache.org/jira/browse/STORM-1991)

Support auto.commit.interval in new Kafka spout

### DIFF
--- a/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
+++ b/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
@@ -107,7 +107,9 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
         tuplesBuilder = kafkaSpoutConfig.getTuplesBuilder();
 
         if (!consumerAutoCommitMode) {     // If it is auto commit, no need to commit offsets manually
-            commitTimer = new Timer(500, kafkaSpoutConfig.getOffsetsCommitPeriodMs(), TimeUnit.MILLISECONDS);
+            long autoCommitInterval = kafkaSpoutConfig.getAutoCommitIntervalMs();
+            long offsetsCommitPeriod = kafkaSpoutConfig.getOffsetsCommitPeriodMs();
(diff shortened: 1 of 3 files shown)


harshach:
@hmcl can you review this.

hmcl:
@darionyaphet can you please clarify why this patch is needed? In my understanding it shouldn't be necessary to have any logic for auto commit interval. If one wishes to set the auto commit interval, one can do so using the kafka properties `autocommit.interval.ms` and `autocommit.enable`. These properties will then be passed the `KafkaConsumer` [here](https://github.com/apache/storm/blob/master/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java#L350)

The [commit timer](https://github.com/apache/storm/blob/master/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java#L111) is only used when in non auto commit mode. The 500 is only the initial delay. One can make this initial delay configurable, but it may not be of extreme importance to do so.

darionyaphet:
Hi @hmcl  `auto.commit.interval.ms` is the interval to committed offset into zookeeper and it default value is `60 * 1000 (1S)`.  I found there is a constant in `KafkaSpoutConfig .Consumer` , so I try to expose a method to config it. The commit time is `500ms` in `KafkaSpoutConfig` ,actually they are the same .

hmcl:
@darionyaphet this kafka spout does not use zookeeper as the new Kafka API writes the offsets to a topic rather than to zookeeper. As I mentioned in the initial comment, this change is not necessary. Furthermore, all the kafka properties are supposed to go in a map, as illustrated in  [this example](https://github.com/apache/storm/blob/master/external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/test/KafkaSpoutTopologyMainNamedTopics.java#L115-L123)

At last, the commitTimer is not used in auto commit mode. 

I would like to suggest that you close this PR.

PR #1593, opened 2016-07-26
STORM-1995: close input stream and other tweaks to downloadChunk

abellina:
When the input stream is being removed from its parent collection (after reading fully) in downloadChunk it isn't being closed, causing a leak.

Thanks to @tgravescs for making this change.

### DIFF
--- a/storm-core/src/clj/org/apache/storm/daemon/nimbus.clj
+++ b/storm-core/src/clj/org/apache/storm/daemon/nimbus.clj
@@ -1793,16 +1793,16 @@
         (.mark nimbus:num-downloadChunk-calls)
         (check-authorization! nimbus nil nil "fileDownload")
         (let [downloaders (:downloaders nimbus)
-              ^BufferFileInputStream is (.get downloaders id)]
+              ^BufferInputStream is (.get downloaders id)]
           (when-not is
             (throw (RuntimeException.
-                    "Could not find input stream for that id")))
+                    "Could not find input stream for id " id)))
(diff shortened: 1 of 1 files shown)


storm-core/src/clj/org/apache/storm/daemon/nimbus.clj
```diff
@@ -1793,16 +1793,16 @@
         (.mark nimbus:num-downloadChunk-calls)
         (check-authorization! nimbus nil nil "fileDownload")
         (let [downloaders (:downloaders nimbus)
-              ^BufferFileInputStream is (.get downloaders id)]
+              ^BufferInputStream is (.get downloaders id)]
```
HeartSaVioR: Is there a reason to change type hint class name?
http://storm.apache.org/releases/1.0.0/javadocs/org/apache/storm/utils/BufferFileInputStream.html
BufferFileInputStream is not a "is-a" BufferInputStream, and it also has close() method itself.

abellina: @HeartSaVioR, in beginFileDownload (just above the impl for downloadChunk) the type used for the input stream and put into :downloaders is a BufferInputStream, hence the change in downloadChunk.

HeartSaVioR: @abellina OK I didn't know about the context. Great change.

HeartSaVioR:
+1

PR #1594, opened 2016-07-26
STORM-2006 Storm metrics feature improvement: support per-worker level metrics aggregation (1.x)

HeartSaVioR:
PR for master: #1595
- SystemBolt handles task level metrics via two mode
  - non-aggregate: same to previous, just applying expansion and pass to MetricConsumerBolts
  - aggregate: apply expansion, do aggregation, pass aggregated metrics to MetricConsumerBolts
- all task level metrics should pass by SystemBolt within its worker
  - it drops all connections between tasks and MetricsConsumerBolts
  - only SystemBolts and MetricConsumerBolts will be connected
- move configurations: expandMapType and metricNameSeparator to global
  - since SystemBolt needs to handle expansion when both worker level aggregation and expandMapType are turned on

Manually tested via matrix of options : aggregate x expand

I'm not sure how to create unit tests playing with SystemBolt, but will try to come up unit tests.

### DIFF
--- a/conf/defaults.yaml
+++ b/conf/defaults.yaml
@@ -260,6 +260,10 @@ topology.disruptor.batch.size: 100
 topology.disruptor.batch.timeout.millis: 1
 topology.disable.loadaware: false
 topology.state.checkpoint.interval.ms: 1000
+topology.metrics.aggregate.per.worker: false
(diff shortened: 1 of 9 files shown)


HeartSaVioR:
Closing in favor of STORM-2153, we're going to build new metrics.

PR #1595, opened 2016-07-26
STORM-2006 Storm metrics feature improvement: support per-worker level metrics aggregation

HeartSaVioR:
PR for 1.x : #1594 
- SystemBolt handles task level metrics via two mode
  - non-aggregate: same to previous, just applying expansion and pass to MetricConsumerBolts
  - aggregate: apply expansion, do aggregation, pass aggregated metrics to MetricConsumerBolts
- all task level metrics should pass by SystemBolt within its worker
  - it drops all connections between tasks and MetricsConsumerBolts
  - only SystemBolts and MetricConsumerBolts will be connected
- move configurations: expandMapType and metricNameSeparator to global
  - since SystemBolt needs to handle expansion when both worker level aggregation and expandMapType are turned on

This could break #1445 and it's already broken. We ideally want to stop adding new feature on 1.x and concentrate on 2.x but life is not easy.
I can volunteer to fix the conflict if this PR breaks #1445 in many places.

### DIFF
--- a/conf/defaults.yaml
+++ b/conf/defaults.yaml
@@ -259,6 +259,10 @@ topology.disruptor.batch.size: 100
 topology.disruptor.batch.timeout.millis: 1
 topology.disable.loadaware: false
 topology.state.checkpoint.interval.ms: 1000
+topology.metrics.aggregate.per.worker: false
+topology.metrics.aggregate.metric.evict.secs: 5
+topology.metrics.expand.map.type: false
+topology.metrics.metric.name.separator: "."
 
 # Configs for Resource Aware Scheduler
 # topology priority describing the importance of the topology in decreasing importance starting from 0 (i.e. 0 is the highest priority and the priority importance decreases as the priority number increases).
--- a/storm-core/src/jvm/org/apache/storm/daemon/StormCommon.java
+++ b/storm-core/src/jvm/org/apache/storm/daemon/StormCommon.java
@@ -36,7 +36,6 @@ import org.apache.storm.metric.EventLoggerBolt;
 import org.apache.storm.metric.MetricsConsumerBolt;
 import org.apache.storm.metric.SystemBolt;
 import org.apache.storm.metric.filter.FilterByMetricName;
-import org.apache.storm.metric.util.DataPointExpander;
 import org.apache.storm.security.auth.IAuthorizer;
 import org.apache.storm.task.IBolt;
 import org.apache.storm.task.WorkerTopologyContext;
@@ -91,8 +90,6 @@ public class StormCommon {
     public static final String TOPOLOGY_METRICS_CONSUMER_PARALLELISM_HINT = "parallelism.hint";
     public static final String TOPOLOGY_METRICS_CONSUMER_WHITELIST = "whitelist";
     public static final String TOPOLOGY_METRICS_CONSUMER_BLACKLIST = "blacklist";
-    public static final String TOPOLOGY_METRICS_CONSUMER_EXPAND_MAP_TYPE = "expandMapType";
-    public static final String TOPOLOGY_METRICS_CONSUMER_METRIC_NAME_SEPARATOR = "metricNameSeparator";
 
     @SuppressWarnings("unchecked")
     public static String getStormId(final IStormClusterState stormClusterState, final String topologyName) {
@@ -347,14 +344,6 @@ public class StormCommon {
         return common;
     }
 
-    public static void addMetricStreams(StormTopology topology) {
-        for (Object component : allComponents(topology).values()) {
-            ComponentCommon common = getComponentCommon(component);
-            StreamInfo streamInfo = Thrift.outputFields(Arrays.asList("task-info", "data-points"));
-            common.put_to_streams(Constants.METRICS_STREAM_ID, streamInfo);
-        }
-    }
-
     public static void addSystemStreams(StormTopology topology) {
         for (Object component : allComponents(topology).values()) {
             ComponentCommon common = getComponentCommon(component);
@@ -397,18 +386,11 @@ public class StormCommon {
         topology.put_to_bolts(EVENTLOGGER_COMPONENT_ID, eventLoggerBolt);
     }
 
-    @SuppressWarnings("unchecked")
-    public static Map<String, Bolt> metricsConsumerBoltSpecs(Map conf, StormTopology topology) {
-        Map<String, Bolt> metricsConsumerBolts = new HashMap<>();
+    public static Map<String, Bolt> metricsConsumerBoltSpecs(Map conf) {
+        Map<String, Bolt> metricsConsumerBolts = new HashMap<String, Bolt>();
 
-        Set<String> componentIdsEmitMetrics = new HashSet<>();
-        componentIdsEmitMetrics.addAll(allComponents(topology).keySet());
-        componentIdsEmitMetrics.add(Constants.SYSTEM_COMPONENT_ID);
-
-        Map<GlobalStreamId, Grouping> inputs = new HashMap<>();
-        for (String componentId : componentIdsEmitMetrics) {
-            inputs.put(Utils.getGlobalStreamId(componentId, Constants.METRICS_STREAM_ID), Thrift.prepareShuffleGrouping());
-        }
+        Map<GlobalStreamId, Grouping> inputs = new HashMap<GlobalStreamId, Grouping>();
+        inputs.put(Utils.getGlobalStreamId(Constants.SYSTEM_COMPONENT_ID, Constants.METRICS_AGGREGATE_STREAM_ID), Thrift.prepareShuffleGrouping());
 
         List<Map<String, Object>> registerInfo = (List<Map<String, Object>>) conf.get(Config.TOPOLOGY_METRICS_CONSUMER_REGISTER);
         if (registerInfo != null) {
@@ -426,13 +408,8 @@ public class StormCommon {
                 List<String> blacklist = (List<String>) info.get(
                     TOPOLOGY_METRICS_CONSUMER_BLACKLIST);
                 FilterByMetricName filterPredicate = new FilterByMetricName(whitelist, blacklist);
-                Boolean expandMapType = Utils.getBoolean(info.get(
-                    TOPOLOGY_METRICS_CONSUMER_EXPAND_MAP_TYPE), false);
-                String metricNameSeparator = Utils.getString(info.get(
-                    TOPOLOGY_METRICS_CONSUMER_METRIC_NAME_SEPARATOR), ".");
-                DataPointExpander expander = new DataPointExpander(expandMapType, metricNameSeparator);
                 MetricsConsumerBolt boltInstance = new MetricsConsumerBolt(className, argument,
-                    maxRetainMetricTuples, filterPredicate, expander);
+                    maxRetainMetricTuples, filterPredicate);
                 Bolt metricsConsumerBolt = Thrift.prepareSerializedBoltDetails(inputs,
                     boltInstance, null, phintNum, metricsConsumerConf);
 
@@ -453,7 +430,7 @@ public class StormCommon {
     }
 
     public static void addMetricComponents(Map conf, StormTopology topology) {
-        Map<String, Bolt> metricsConsumerBolts = metricsConsumerBoltSpecs(conf, topology);
+        Map<String, Bolt> metricsConsumerBolts = metricsConsumerBoltSpecs(conf);
         for (Map.Entry<String, Bolt> entry : metricsConsumerBolts.entrySet()) {
             topology.put_to_bolts(entry.getKey(), entry.getValue());
         }
@@ -463,7 +440,9 @@ public class StormCommon {
     public static void addSystemComponents(Map conf, StormTopology topology) {
         Map<String, StreamInfo> outputStreams = new HashMap<>();
         outputStreams.put(Constants.SYSTEM_TICK_STREAM_ID, Thrift.outputFields(Arrays.asList("rate_secs")));
+        outputStreams.put(Constants.METRICS_STREAM_ID, Thrift.directOutputFields(Arrays.asList("task-info", "data-points")));
         outputStreams.put(Constants.METRICS_TICK_STREAM_ID, Thrift.outputFields(Arrays.asList("interval")));
+        outputStreams.put(Constants.METRICS_AGGREGATE_STREAM_ID, Thrift.outputFields(Arrays.asList("task-info", "data-points")));
         outputStreams.put(Constants.CREDENTIALS_CHANGED_STREAM_ID, Thrift.outputFields(Arrays.asList("creds")));
 
         Map<String, Object> boltConf = new HashMap<>();
@@ -485,7 +464,6 @@ public class StormCommon {
         addEventLogger(stormConf, ret);
         addMetricComponents(stormConf, ret);
         addSystemComponents(stormConf, ret);
-        addMetricStreams(ret);
         addSystemStreams(ret);
 
         validateStructure(ret);
--- a/storm-core/src/jvm/org/apache/storm/metric/SystemBolt.java
+++ b/storm-core/src/jvm/org/apache/storm/metric/SystemBolt.java
@@ -18,18 +18,33 @@
 package org.apache.storm.metric;
 
 import org.apache.storm.Config;
+import org.apache.storm.Constants;
 import org.apache.storm.metric.api.IMetric;
+import org.apache.storm.metric.api.IMetricsConsumer;
+import org.apache.storm.metric.util.DataPointExpander;
 import org.apache.storm.task.IBolt;
 import org.apache.storm.task.OutputCollector;
 import org.apache.storm.task.TopologyContext;
 import org.apache.storm.tuple.Tuple;
+import org.apache.storm.tuple.Values;
+import org.apache.storm.utils.Time;
 import org.apache.storm.utils.Utils;
 import clojure.lang.AFn;
 import clojure.lang.IFn;
 import clojure.lang.RT;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
-import java.lang.management.*;
+import java.lang.management.GarbageCollectorMXBean;
+import java.lang.management.ManagementFactory;
+import java.lang.management.MemoryMXBean;
+import java.lang.management.MemoryUsage;
+import java.lang.management.RuntimeMXBean;
+import java.util.ArrayList;
+import java.util.Collection;
 import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
 import java.util.Map;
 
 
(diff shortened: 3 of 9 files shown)


conf/defaults.yaml
```diff
@@ -259,6 +259,10 @@ topology.disruptor.batch.size: 100
 topology.disruptor.batch.timeout.millis: 1
 topology.disable.loadaware: false
 topology.state.checkpoint.interval.ms: 1000
+topology.metrics.aggregate.per.worker: false
```
harshach: we should make this as true and document it. As keeping this false can easily put lot of metrics into users metric storage. If they are not aware of it can get into issues.

harshach:
+1 . Minor nit.

conf/defaults.yaml
HeartSaVioR: @harshach 
Set this to 'true' breaks current metrics consumer implementations, since it delegates aggregation to metrics consumer by passing list of values. We need to guide the way how to aggregate these values to well-known metrics consumer implementations.

Btw, we should try avoiding to address current metrics feature and start re-designing new metrics feature. To tell the truth, this feature is actually closer to a hot-fix instead of improvement. 
Even I found the design issue while implementing this. Current metrics feature doesn't allow us to aggregate the values, since we can't add the metadata (which function we need to apply to aggregate) to metrics. Note that we expand metrics from value of metric, so adding metadata to metric doesn't make sense. Metric shouldn't use Object as value, especially some types are intended to be expanded.

HeartSaVioR: I also think current pluggable metrics consumer design is too flexible. Users rely on metrics consumers which are open source from individual or specific vendor, which is easy to be not maintained. I guess most popular implementation of metrics consumer is verisign/storm-graphite, and it's not maintained for months. We should provide stable consumer implementation. If we don't want to handle various time-series DBs, we can open at most pluggable connector interface.

harshach:
+1

conf/defaults.yaml
ptgoetz: @HeartSaVioR 
If I'm following the code/conversation correctly, this flag essentially changes the IMetricConsumer interface (specifically the value set in the DataPoint class - which is of type Object).  That makes me very uneasy.

Is there a way to make the interface more strongly typed?

Thinking out loud: What about splitting the interface in two: One that does aggregation and one that doesn't, then switching calls based on which interface is implemented?

I'm not opposed to breaking backwards compatibility and bumping up the major version if necessary. In other words I'd rather see a clean design than preserve backwards compatibility with respect to the metrics features.

harshach: @ptgoetz I don't see any interface changes. Its just additional config in absence of it or the default it will continue to work with older implementation of IMetricsConsumer.

ptgoetz: @harshach you are correct, there are no *explicit* interface changes. But there is an *implicit* contract with existing implementations that expect `DataPoint.value` to be a specific type, and that contract is broken when this flag is set to `true`. So this is *effectively* an API change that's just hidden by the fact that it is configurable. 

> Btw, we should try avoiding to address current metrics feature and start re-designing new metrics feature. To tell the truth, this feature is actually closer to a hot-fix instead of improvement.

I agree with @HeartSaVioR that this approach is less than optimal and very brittle. If we're going to break backward compatibility, we should go ahead and improve the design of the API.

I really don't like the idea that a configuration change can effectively change an API. I'd rather change the API to be more explicit in terms of data types sent to the implementation.

HeartSaVioR: @ptgoetz 
As you said this is just due to a design flaw of current metrics feature. 
Return type of getValueAndReset (Object) just makes whole things opaque, which is really bad.
Also we even didn't document a *implicit contract* that Map type of value of metric should be expanded to multiple metrics. How to process metrics is completely up to metrics consumer developer, and some metrics consumers even don't expand Map type of value. I'm introducing similar thing from this change, which I think it is still bad, but if you read the conversation, you may notice that there's no other choice with current metrics.

@harshach @ptgoetz 
This must be a last call of new feature/improvement of current metrics feature. To tell the truth, I should never improve current metrics feature and concentrate on new metrics feature, but I didn't and couldn't, because of several reasons. My bad.

Actually my bad decision started from our progress of Storm 2.0.0 plan. For the first time I'm just waiting for JStorm merger phase 2 so that JStorm metrics feature to be applied. But we even didn't finish phase 1 in 6 months, which is being said that it is expected to be finished in 1Q. So I was end up with going on my own work, but have been considered about the backward compatibility, and once I started considering that, I couldn't get out of it.

I'm still in doubt that we can finish JStorm merger by end of this year, since phase 1 seems to be no progress on more than 1 month. If we would want to concentrate port & merger works and make this finally happen, I can wait for that. (Sure I'll participate the works.) But if it's out of our mind, we should have the time to design and implement new metrics feature.

ptgoetz: @HeartSaVioR Just to be clear, please don't take my comments to be a criticism of your work. You've done a fantastic job with metrics improvements.

Through that work it's become apparent that the existing API for metrics consumers is lacking in a number of ways. I view this as an opportunity to correct that.

We're incrementally moving toward semantic versioning. I'm not opposed to breaking backward compatibility with the existing metrics API and moving this branch to 2.x. The fact that master is "2.0" is just a technicality that can easily be changed.

I'd like to fix the API to better reflect the work you've done, rather than to implement a one-off fix.

HeartSaVioR: @ptgoetz No worries. I didn't take your comments in that way. It was just my criticism of **current metrics API**.

Storm community discussed about metrics several times, and the only one that I can get is the needs of new metrics feature. 

This is a last discussion which also contains requirements of metrics feature. I don't think we should address all of them, but addressing much of them would be great.
http://mail-archives.us.apache.org/mod_mbox/storm-dev/201605.mbox/%3CCAF5108jB=4aL0Z1qMMnEe7U4Yx_9TZjD11BJYhKtkTtVJoTdhQ@mail.gmail.com%3E

Adding fixes to current metrics seems not competitive to recent metrics from others stream frameworks, and even JStorm, which is the thing we would eventually evaluate (I already did a first pass and looks promising) and port if it's better. This patch only has a value with current metrics, and I don't want to break backward compatibility because of will-be-deprecated-and-dropped feature.

HeartSaVioR: FYI, I tried to port JStorm metrics to 1.x some months ago, but it touches many paths of codebase so we would end up having hard time to do the work if port is not finished.

harshach:
@HeartSaVioR  You improved the metrics collection and quality of those metrics as well. This is immensely useful in debugging any topologies and operationalizing the Storm it self. I am not sure if there any users (atleast I haven't seen one) using the existing topology.metrics.consumer as its performance is bad and can get the topology to a halt.
 In light of these issues what you did here is will be extremely helpful and learnings from here will help in the 2.0 changes as well. 
Having said that this fix critical as we end up flooding the storage of metrics with exponential number of metrics.
@ptgoetz "that contract is broken when this flag is set to true".  Yes and we are shipping this as false. So Unless user is consciously makes this as true its not breaking anything. Sufficient documentation around this will help on what to expect.
  We also need to make a judgement call on what it will impact . With this fix in users can go ahead store the metrics much more optimal fashion than what was happening before.
IMHO this is not a backward-incompatible change and it will help users in storing the metrics better.

I am still +1 on merging this patch.

"Btw, we should try avoiding to address current metrics feature and start re-designing new metrics feature. To tell the truth, this feature is actually closer to a hot-fix instead of improvement."

@HeartSaVioR  agree. But this will get the users to start using the metrics feature in storm. Before any of these changes all users depends on are UI metrics and with these changes in they can more accurately look at whats going on the cluster. Also they can add metrics into their topology. we definitely should improve this feature for 2.0.

HeartSaVioR:
@harshach @ptgoetz 
I believe the best approach is letting users don't care about them.

While I still think we need to renew metrics feature, I also think one of valid approach for current metrics is spending some more time to provide some official implementations of metrics consumers which covers popular time-series DBs like Grafana, and OpenTSDB, InfluxDB, and some others if we have ideas.

This makes us feeling similar to dog-fooding, and will show how good or how bad is the metrics feature transparently.
(To tell the truth, I realized the limitations of current metrics feature clearly while playing with Ambari Metric Service Storm sink.)
Contracts will be mostly on our own, so no delegation for the hard thing on users or dev. of metric consumers.

Just I'm not 100% sure open source metrics consumer projects will reflect our changed metrics API or contract in time. So while we're improving the feature, it could unlikely get to the end-users. This is my another concern.

HeartSaVioR:
cc. @abhishekagarwal87 
I guess you seems to operate Storm metrics in production, so would like to hear your thought about this. (Not meant to review the code. I don't have operation experience for metrics on production, so need some help on this area.)

ptgoetz:
@HeartSaVioR I don't think I was very clear in articulating my proposal. What I'm saying is rather than making this a configuration switch, make it dynamic based on what interface the consumer implements. So it would involve:

1. Leave `IMetricsConsumer` as-is, but deprecated.
2. Introduce a new interface for aggregated metrics (e.g. `IAggregateMetricsConsumer`)
3. At runtime, look at which interface the registered consumer implements, and alter the logic based on that (rather than a configuration switch)

With that we would have both backward compatibility and a new presumably cleaner API for aggregated metrics. The internals my not be pretty, but that would be invisible to the user.

From the discussion thread you pointed out, there seems to be consensus around such an approach:

Quoting @revans2:

> On 1.x we add in a new API for both reporting metrics and collecting in parallel to the old API.  We leave IMetric and IMetricsConsumer in place but deprecated.  As we move internal metrics over from the old interface to the new one, we either keep versions of the old ones in place or we provide a translation shim going from the new to the old. 

Quoting @abhishekagarwal87:

> Sounds good. Having two separate metric reporters may be confusing but it
is better than breaking the client code.

Quoting you:

> Personally I'm also in favor of maintaining old API (but deprecated) and
adding new API. It's ideal way, and that's what many projects are trying to do, and so on
the other project I'm also maintaining.

HeartSaVioR:
@ptgoetz 
That consensus was made with context `breaking change for 1.x`. Sure if we introduce breaking change in 1.x we need to support the old one. But if we go on introducing breaking change only for next major version, I don't think we must follow our consensus.

We even could introduce completely different way like JStorm. In JStorm metrics are sent to Nimbus, and plugin to external storage is also there, so it might not be topology specific. In this case we could not provide the translation shim, but I don't think it matters if we introduce the change for next major version.

When we talk about new API I'd like to focus the new one, not considering how to support the old one. For me, this change is more suited to the old one. There's another level to aggregate, component level, and it might be more preferred thing. Why I take worker level aggregation instead of component level aggregation is that it's only possible way to achieve with not breaking current design. That's why we should not consider about old one while designing new one.

storm-core/src/jvm/org/apache/storm/daemon/StormCommon.java
```diff
@@ -436,7 +414,9 @@ public static void addMetricComponents(Map conf, StormTopology topology) {
     public static void addSystemComponents(Map conf, StormTopology topology) {
         Map<String, StreamInfo> outputStreams = new HashMap<String, StreamInfo>();
         outputStreams.put(Constants.SYSTEM_TICK_STREAM_ID, Thrift.outputFields(Arrays.asList("rate_secs")));
+        outputStreams.put(Constants.METRICS_STREAM_ID, Thrift.directOutputFields(Arrays.asList("task-info", "data-points")));
```
abhishekagarwal87: is it needed?

storm-core/src/jvm/org/apache/storm/metric/SystemBolt.java
```diff
@@ -156,7 +219,103 @@ private void registerMetrics(TopologyContext context, Map<String, String> metric
 
     @Override
     public void execute(Tuple input) {
-        throw new RuntimeException("Non-system tuples should never be sent to __system bolt.");
```
abhishekagarwal87: can we check that the incoming stream is METRICS_STREAM and throw same exception if its not?

storm-core/src/jvm/org/apache/storm/metric/SystemBolt.java
```diff
@@ -156,7 +219,103 @@ private void registerMetrics(TopologyContext context, Map<String, String> metric
...
+    }
+
+    private void handleMetricTupleInAggregateMode(IMetricsConsumer.TaskInfo taskInfo, Collection<IMetricsConsumer.DataPoint> expandedDataPoints) {
+        Map<Integer, TaskInfoToDataPointsPair> taskToMetricTupleMap = intervalToTaskToMetricTupleMap.get(taskInfo.updateIntervalSecs);
+        if (taskToMetricTupleMap == null) {
+            taskToMetricTupleMap = new HashMap<>();
+            intervalToTaskToMetricTupleMap.put(taskInfo.updateIntervalSecs, taskToMetricTupleMap);
+        }
+
+        taskToMetricTupleMap.put(taskInfo.srcTaskId, new TaskInfoToDataPointsPair(taskInfo, expandedDataPoints));
```
abhishekagarwal87: I didn't get this. It overrides the older information with newer one. what if the data point for same task is received before 'isOKtoAggregateMetricsTuples` is true.

abhishekagarwal87:
@HeartSaVioR - Thanks for putting in efforts on improving metrics.  The approach of passing metrics via system bolt looks agreeable to me. Though I would still prefer "Two types of metric consumers approach". 
Secondly regarding aggregation, component level aggregation is much more preferred. That is what we do in external systems. Simply for the reason that  `total number of messages processed by component A` is more useful and actionable information than `total number of messages processed on this machine and port`. If the component information is being let go during aggregation, I don't think it will be used much. At least I wouldn't in our production system. 

Load among tasks is usually homogenous and uniform (law of large numbers) but same is not true for different components which execute different code altogether. I think just aggregating per worker+component combination would also reduce the amount of metrics being emitted and still keep the metrics useful.

Thirdly, the assumption that a metric being declared with 60 second update interval will actually come in 60 second intervals is not correct. They can come with 120 seconds apart or they can also come 5 seconds apart (depending on how fast or slow the task is).

storm-core/src/jvm/org/apache/storm/metric/SystemBolt.java
HeartSaVioR: It is intended behavior. If the data point for same task on same interval is received and it overwrites the old value, it means old value is not aggregated in time. If we want to keep also the old value then it's more confusing that what values we want to aggregate together. 

This is due to design issue on current metrics feature (late metrics) and if it occurs, metric value would be flawed anyway (I mean value is not measured for given interval) so we can't take good care of them. Fixing this issue should be a one of the requirements for new metrics feature.

storm-core/src/jvm/org/apache/storm/metric/SystemBolt.java
HeartSaVioR: Makes sense. I'll address this.

storm-core/src/jvm/org/apache/storm/daemon/StormCommon.java
HeartSaVioR: Yes we send the metrics tick tuple to each task with source stream to that, so it's needed.

HeartSaVioR: Oh sorry I've confused. We send the metrics tuple from each task with fake source stream which is `SYSTEM_TASK_ID METRICS_STREAM_ID`, so it's needed. Please see the change of `metric-tick`.

HeartSaVioR:
@abhishekagarwal87 
1. I think new type of metrics consumer should refer the new metrics feature which addresses our concerns on current metrics. Aggregation is just a one of them so I don't want to introduce new type of metrics consumer without consideration of other things. It doesn't mean I started working for the new metrics feature, just like a hotfix for current metrics feature.

2. Worker+component combination is what this PR is trying to achieve. `convertMetricsTupleMapKeyedByTaskInfo` is for that. You can see that I only change the task id and timestamp in TaskInfo, and use TaskInfo to aggregate the values, which means metrics from different components are not aggregated to same list.
I also don't think aggregated metric value is useful if they are aggregated from various components.

3. Yes I'm aware of it, and that's why I introduce `eviction` of the metric points.
Btw, in that care, metric value would be flawed anyway. It is already out of control so we should focus how to not making this situation instead of how to correct the value.
And triggering (sending metrics tick tuple to tasks) can't guarantee this.

As we were talking about metrics from user@ and dev@ list, there're a lot of things to correct from current metrics feature. Things are coupled each other so this is also not easy to fix them one by one. Let's move out to the new metrics feature rather than saying them with current metrics.

HeartSaVioR:
@ptgoetz @harshach @abhishekagarwal87 
I just had the limitations of current metrics feature only in my mind, but it would be better to write it down and share to persuade the need for new metrics feature.

I didn't grant the wiki access so I just use google doc. Please let me know if we want to have this in Apache infra. I can move this out to wiki, or just post to dev@ mailing list.

https://docs.google.com/document/d/11H3STXr2QTnfOqisgOnvNt8TjXS_cSqb86VM6vdAo8s/edit?usp=sharing

@unsleepy22 AFAIK you worked on the change of JStorm metrics. Could you check the doc and confirm that JStorm metrics feature addressed the limitations of Storm metrics feature?

harshach:
@HeartSaVioR It will be good to have confluent wiki setup like other projects in apache do. We can have proposal setup there and go through the process of discussion in mailing list. I would start on user list so that we can hear from any existing or future user opinions as well.

HeartSaVioR:
@harshach 
The wish-list contains ideas/opinions in discussions on user@ and also dev@. 

http://mail-archives.apache.org/mod_mbox/storm-user/201604.mbox/%3CCAF5108iwhGgae7Z2AKgMVNP4PWwV52GVc7Ex5h2oQ7fAAPX2PA%40mail.gmail.com%3E

I've initiated discussions several times for this, so I'm sure what @abhishekagarwal87 sorted out is enough set of wish-list for new metrics feature.

Btw, we didn't use wiki for such purpose, so it might be better to talk about the format, the process, and so on. If we just want to have a page without specific format, I'll move this to wiki page once I granted write access to wiki.

HeartSaVioR:
OK. I break down my doc to two pieces and posted to wiki page.

- https://cwiki.apache.org/confluence/display/STORM/Limitations+of+current+metrics+feature
- https://cwiki.apache.org/confluence/display/STORM/Wishlist+for+new+metrics+feature

Since I'm not working on new metrics feature, I didn't put them into proposal / design doc category. I think we will adopt or address many places including metrics while evaluating JStorm features from Phase 2, so would like to wait for that.

harshach:
@HeartSaVioR I would like to see this merged into 1.x-branch. Given suggestions here can you make the changes , introducing IAggregatedMetrics interface and leave the existing the interface as it is.

HeartSaVioR:
We shouldn't make the change only available for 1.x, especially this is only effective when ecosystem (I mean metrics consumer implementators) should follow up. If we renew metrics feature to JStorm thing we might even don't have a chance to deprecate interfaces before removing them for 2.0.
As I stated to wiki page, what we need to change is metrics feature itself.
Let's wait for phase 2 to have a chance to renew, or someone claims this should be merged in 1.x branch.

HeartSaVioR:
@ptgoetz @harshach 
I just changed my mind to introduce new interface as Taylor suggested as it doesn't need much work effort, and it anyway helps some users reducing the traffic.
I'll mention again once it's done.

HeartSaVioR:
Struggled with this but no luck. This is not that easy as just checking the interface type.

1. Entire logic is strongly tied to IMetricsConsumer and IMetricConsumer.DataPoint. Given that we would want to introduce IAggregatedMetricsConsumer and IAggregatedMetricsConsumer.DataPoint (or AggregatedDataPoint), we need to fix all of the usages including Filter and MetricsConsumerBolt and so on. We can't fix IMetricsConsumer and its DataPoint as non-backward compatible way.

2. Aggregating metrics should be handled from SystemBolt, not MetricsConsumerBolt which we initialize the metrics consumer. We're allowing users to set multiple metrics consumers for the topology, which means that user can set both IMetricsConsumer and IAggregatedMetricsConsumer. Do we want to push the metrics to both unaggregated and aggregated things via separate stream?

Please share the idea with details if anyone has better and clean one. Thanks.

HeartSaVioR:
Closing in favor of STORM-2153, we're going to build new metrics.

PR #1597, opened 2016-07-27
STORM-1285 port backtype.storm.command.get-errors to java

HeartSaVioR:
[STORM-1285](https://issues.apache.org/jira/browse/STORM-1285)

Please look at the implementation of `getComponentErrors()` since it simplifies the origin logic.


storm-core/src/jvm/org/apache/storm/command/GetErrors.java
```diff
@@ -0,0 +1,59 @@
...
+import org.apache.storm.utils.Utils;
+import org.json.simple.JSONValue;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class GetErrors {
+    public static void main(String[] args) throws Exception {
+        final String name = args[0];
```
abhishekagarwal87: It would be nice to have a check on args length here and throw "Topology name not passed" exception in case args is empty.

abhishekagarwal87:
Minor comment. Looks good otherwise.

storm-core/src/jvm/org/apache/storm/command/GetErrors.java
abhishekagarwal87: Missed it earlier. This class doesn't have licence headers. Build is failing because of that.

HeartSaVioR: That would be great. Will address.

HeartSaVioR:
Thanks for reviewing. I address your comments.

abhishekagarwal87:
+1. Build failure is in storm-hive, unrelated to this change.

storm-core/src/jvm/org/apache/storm/command/GetErrors.java
```diff
@@ -0,0 +1,80 @@
...
+import org.json.simple.JSONValue;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class GetErrors {
+    public static void main(String[] args) throws Exception {
+        if (args.length == 0) {
+            throw new IllegalArgumentException("You should provide topology name.");
```
satishd: Can we have a better error message? It can be something like `Topology name must be provided.`

satishd:
Minor comment, +1 once addressed.

storm-core/src/jvm/org/apache/storm/command/GetErrors.java
HeartSaVioR: Yes it seems better. I'll address it.

HeartSaVioR:
@abhishekagarwal87 @satishd 
Thanks for the review. I addressed all of comments.

PR #1598, opened 2016-07-27
STORM-2010 port org.apache.storm.command.heartbeats.clj to Java

HeartSaVioR:
[STORM-2010](https://issues.apache.org/jira/browse/STORM-2010)


storm-core/src/jvm/org/apache/storm/command/Heartbeats.java
```diff
@@ -0,0 +1,65 @@
...
+import org.apache.storm.stats.StatsUtil;
+import org.apache.storm.utils.ConfigUtils;
+import org.apache.storm.utils.Utils;
+import org.json.simple.JSONValue;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Map;
+
+public class Heartbeats {
```
abhishekagarwal87: Licence headers have to be added here.

storm-core/src/jvm/org/apache/storm/command/Heartbeats.java
```diff
@@ -0,0 +1,65 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger(Heartbeats.class);
+
+    public static void main(String[] args) throws Exception {
+        String command = args[0];
```
abhishekagarwal87: should we have a check for length here? ArrayOutOfBoundsException may be confusing for cli user.

storm-core/src/clj/org/apache/storm/command/heartbeats.clj
```diff
@@ -1,54 +0,0 @@
...
-           ClusterWorkerHeartbeat))
-         "Nothing"))
-      
-      (log-message "Usage: heartbeats [list|get] path"))
-    
-    (try
-      (.close cluster)
-      (catch Exception e
-        (log-message "Caught exception: " e " on close."))))
-  (System/exit 0))
```
abhishekagarwal87: This is missing in java equivalent. Only reason for placing it could be that some foreground threads are started by the above code.

HeartSaVioR: Great. I missed the meaning of this line. It forces process to be terminated. I'll address it. Thanks!

storm-core/src/jvm/org/apache/storm/command/Heartbeats.java
HeartSaVioR: Yes will address. I was focused on porting origin code as it is, but it would be better to address here.

storm-core/src/jvm/org/apache/storm/command/Heartbeats.java
HeartSaVioR: Thanks, addressed.

HeartSaVioR:
Thanks @abhishekagarwal87 for reviewing. I addressed your comments.

abhishekagarwal87:
+1

storm-core/src/jvm/org/apache/storm/command/Heartbeats.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    private static void handleGetCommand(IStateStorage cluster, String path) {
+        String message;
+        byte[] hb = cluster.get_worker_hb(path, false);
+        if (hb != null) {
+            Map<String, Object> heartbeatMap = StatsUtil.convertZkWorkerHb(Utils.deserialize(hb, ClusterWorkerHeartbeat.class));
+            message = JSONValue.toJSONString(heartbeatMap);
+        } else {
+            message = "Nothing";
```
harshach: can we changes this to "No Heartbeats found"

storm-core/src/jvm/org/apache/storm/command/Heartbeats.java
```diff
@@ -0,0 +1,89 @@
...
+            throw new IllegalArgumentException("You should provide both command and path.");
+        }
+
+        String command = args[0];
+        String path = args[1];
+
+        Map<String, Object> conf = ConfigUtils.readStormConfig();
+        IStateStorage cluster = ClusterUtils.mkStateStorage(conf, conf, null, new ClusterStateContext());
+
+        System.out.println("Command: [" + command + "]");
```
satishd: Can we use `LOG.info()` here instead of `System.out.println`?

storm-core/src/jvm/org/apache/storm/command/Heartbeats.java
```diff
@@ -0,0 +1,89 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.util.Map;
+
+public class Heartbeats {
+    private static final Logger LOG = LoggerFactory.getLogger(Heartbeats.class);
+
+    public static void main(String[] args) throws Exception {
+        if (args.length < 2) {
+            throw new IllegalArgumentException("You should provide both command and path.");
```
satishd: Exception message can be something like `Command and path arguments must be provided`

storm-core/src/jvm/org/apache/storm/command/Heartbeats.java
```diff
@@ -0,0 +1,89 @@
...
+            case "list":
+                handleListCommand(cluster, path);
+                break;
+
+            case "get":
+                handleGetCommand(cluster, path);
+                break;
+
+            default:
+                LOG.info("Usage: heartbeats [list|get] path");
```
satishd: It should be something like `Usage: java Heartbeats [list|get] path` instead of using `heartbeats`

HeartSaVioR: Execution detail depends on outside of this class, `storm` and `storm.py` and it would be better for this class to not care about that.

storm-core/src/jvm/org/apache/storm/command/Heartbeats.java
HeartSaVioR: Yes it's more clear. I'll address it.

storm-core/src/jvm/org/apache/storm/command/Heartbeats.java
HeartSaVioR: This is from origin code but it seems that we can change it. I'll address it.

storm-core/src/jvm/org/apache/storm/command/Heartbeats.java
HeartSaVioR: OK that would be better. I'll address it.

HeartSaVioR:
Thanks @satishd @harshach for reviewing. Addressed or commented review comments.

satishd:
+1 LGTM

PR #1602, opened 2016-08-01
Fix command to run RollingTopWords example

Zintinio:
Appended target directory to command. Previously resulted in ClassNotFound exception.


HeartSaVioR:
Example 1 and example 2 seems inconsistent. Good catch. +1

HeartSaVioR:
Thanks @Zintinio I merged into master, 1.x, 1.0.x branches.

PR #1591, opened 2016-07-26
STORM-1038: Upgrade netty to 4.x in 1.x-branch

hsun-cnnxty:
This is to add the feature to 1.x-branch.  The original PR for master branch is #728.

PR #728, opened 2015-09-09
[STORM-1038] Upgraded netty to 4.x

hsun-cnnxty:
Upgraded the netty transportation layer to 4.x to take advantage of its memory management efficiency.

PR #1604, opened 2016-08-02
[STORM-2013] Upgrade Netty to 3.10.6

darionyaphet:
[STORM-2013 Upgrade Netty to 3.10.6](https://issues.apache.org/jira/browse/STORM-2013)

Since Netty 3.9 , it have fix some bugs and improve performance .

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -233,7 +233,7 @@
         <disruptor.version>3.3.2</disruptor.version>
         <jgrapht.version>0.9.0</jgrapht.version>
         <guava.version>16.0.1</guava.version>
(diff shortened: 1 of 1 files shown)


HeartSaVioR:
@darionyaphet 
Since 3.10.6 is EOL of 3.x, I'd rather move on Netty 4.
Patches are already here #728 (master) #1591 (1.x), and when we're OK with the performance test I think we can check them in.

darionyaphet:
@HeartSaVioR  OK maybe move to netty 4 is better thanks :)

PR #1605, opened 2016-08-02
STORM-2014: Put logic around dropping messages into RetryService, rem…

srdo:
…ove maxRetry setting from new KafkaSpout

https://issues.apache.org/jira/browse/STORM-2014

This PR removes maxRetry from the KafkaSpout and changes the RetryService interface slightly so the schedule method can communicate back to the spout that the message should be dropped. Retry logic belongs to the RetryService interface, and it's nice for users if they can easily plug in their own handling of messages that will be dropped (custom logging for example).


priyank5485:
+1 (non-binding). I had talked to @hmcl briefly about this. We should have his review as well

srdo:
@hmcl ping. Any input on this?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutRetryExponentialBackoff.java
```diff
@@ -144,7 +145,12 @@ public String toString() {
     /**
      * The time stamp of the next retry is scheduled according to the exponential backoff formula ( geometric progression):
      * nextRetry = failCount == 1 ? currentTime + initialDelay : currentTime + delayPeriod^(failCount-1) where failCount = 1, 2, 3, ...
-     * nextRetry = Min(nextRetry, currentTime + maxDelay)
+     * nextRetry = Min(nextRetry, currentTime + maxDelay).
+     * 
+     * While retrying a record, no new records are committed until the previous polled records have been acked. This guarantees at once delivery of
```
hmcl: This comment should probably not go here, but rather somewhere in the Spout code. (retry service knows nothing about exactly once or at most once semantics)

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -330,11 +328,9 @@ public void ack(Object messageId) {
...
         final KafkaSpoutMessageId msgId = (KafkaSpoutMessageId) messageId;
         emitted.remove(msgId);
-        if (msgId.numFails() < maxRetries) {
-            msgId.incrementNumFails();
-            retryService.schedule(msgId);
-        } else { // limit to max number of retries
-            LOG.debug("Reached maximum number of retries. Message [{}] being marked as acked.", msgId);
+        msgId.incrementNumFails();
+        if (!retryService.schedule(msgId)) {
+            LOG.debug("Retry service indicated message should not be retried. Message [{}] being marked as acked.", msgId);
```
hmcl: I would leave the message as it was before. "Reached maximum number of retries. Message [{}] being marked as acked."

It adds overhead and it is irrelevant to be mentioning the retry service here.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutRetryService.java
```diff
@@ -29,14 +29,18 @@
  */
 public interface KafkaSpoutRetryService extends Serializable {
     /**
-     * Schedules this {@link KafkaSpoutMessageId} if not yet scheduled, or updates retry time if it has already been scheduled.
+     * Schedules this {@link KafkaSpoutMessageId} if not yet scheduled, or
+     * updates retry time if it has already been scheduled. May also indicate
```
hmcl: It may also indicate...

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutRetryExponentialBackoff.java
srdo: Sure. I'll move it to findNextCommitOffset in the spout.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
srdo: I changed it because retryService can be any implementation here. It may decide not to schedule a message for other reasons than reaching the retry cap. I'll change it back if you still prefer the previous log message.

hmcl: @srdo I am not sure I completely follow your reasoning. The most meaningful piece of information that we want to show to the user here is that if the max number of retries has been reached, no more retries happen. If the user reads a message saying that the retry service marked it not to be retried, he still does not know the cause.

Looking at the code, the `boolean schedule(msgId)` only returns false only if the max number of retries has been reached.

Although the retry service log messages hints ant the max retries cap being reached, the user will have to have both logs enabled to have both spouts. I favor the message as it was before, as it is straight to the point and provides good insight on what is happening.

hmcl:
I am +1 overall. Please format the commit message to be easy to read and squash the two commits. This is a simple change that should have only one commit. We can merge after that.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
srdo: @hmcl I was referring to RetryService being an interface users can override, which doesn't mention max retries in the schedule javadoc, and this part of the code technically not "knowing about" max retries, because users may supply a retry service that doesn't schedule tuples for some other reason. I think you're right though, it was probably clearer before.

srdo:
Squashed. I hope this commit message is clearer.

harshach:
+1

HeartSaVioR:
+1

PR #1608, opened 2016-08-05
STORM-2016 Topology submission improvement: support adding local jars and maven artifacts on submission (1.x)

HeartSaVioR:
* JIRA issue: http://issues.apache.org/jira/browse/STORM-2016
* design doc: https://cwiki.apache.org/confluence/display/STORM/A.+Design+doc%3A+adding+jars+and+maven+artifacts+at+submission
* discussion thread: http://mail-archives.apache.org/mod_mbox/storm-dev/201608.mbox/%3CCAF5108i9+tJaNZ0LgRkTMkVQEL7F+53k9uyzxcT6zhSU6OHx9Q@mail.gmail.com%3E

----

* bin/storm now supports "--jars" and "--packages" options
  * it's only effective with "storm jar" and "storm sql"
* introduce new module: storm-submit to help resolving dependencies with handling transitive dependencies
* StormSubmitter will upload dependencies to BlobStore when submitting topology
* Supervisor will download dependencies from BlobStore when such topology is assigned
* Supervisor will launch workers with adding downloaded dependencies to worker classpath

TODO

- [x] documentation
- [x] craft pull request against master branch

Btw, it might be better to place some modules out of external since 'external' has most of non storm-core modules and flux, sql, storm-submit, storm-kafka-monitor are not kind of connectors.

### DIFF
--- a/bin/storm.py
+++ b/bin/storm.py
@@ -24,6 +24,7 @@ import shlex
 import tempfile
 import uuid
 import subprocess as sub
+import json
 
 import sys
 
@@ -96,6 +97,8 @@ if JAVA_HOME and not os.path.exists(JAVA_CMD):
     sys.exit(1)
 STORM_EXT_CLASSPATH = os.getenv('STORM_EXT_CLASSPATH', None)
 STORM_EXT_CLASSPATH_DAEMON = os.getenv('STORM_EXT_CLASSPATH_DAEMON', None)
+DEP_JARS_OPTS = []
+DEP_ARTIFACTS_OPTS = []
 
 def get_config_opts():
     global CONFIG_OPTS
@@ -153,6 +156,42 @@ def confvalue(name, extrapaths, daemon=True):
         if tokens[0] == "VALUE:":
             return " ".join(tokens[1:])
     return ""
+
+def resolve_dependencies(artifacts):
+    if len(artifacts) == 0:
+        return {}
+
+    print("Resolving dependencies on demand: artifacts (%s)" % artifacts)
+    sys.stdout.flush()
+
+    # TODO: should we move some external modules to outer place?
+
+    # storm-submit module doesn't rely on storm-core and relevant libs
+    extrajars = get_jars_full(STORM_DIR + "/external/storm-submit-tools")
+    classpath = normclasspath(os.pathsep.join(extrajars))
+
+    command = [
+        JAVA_CMD, "-client", "-cp", classpath, "org.apache.storm.submit.command.DependencyResolverMain",
+        ",".join(artifacts)
+    ]
+
+    p = sub.Popen(command, stdout=sub.PIPE)
+    output, errors = p.communicate()
+    if p.returncode != 0:
+        raise RuntimeError("dependency handler returns non-zero code: code<%s> syserr<%s>" % (p.returncode, errors))
+
+    # python 3
+    if not isinstance(output, str):
+        output = output.decode('utf-8')
+
+    # For debug purpose, uncomment when you need to debug DependencyResolver
+    #print("Resolved dependencies: %s" % output)
+
+    try:
+        out_dict = json.loads(output)
+        return out_dict
+    except:
+        raise RuntimeError("dependency handler returns non-json response: sysout<%s>", output)
 
 def print_localconfvalue(name):
     """Syntax: [storm localconfvalue conf-name]
@@ -232,44 +271,91 @@ def jar(jarfile, klass, *args):
     The process is configured so that StormSubmitter
     (http://storm.apache.org/releases/current/javadocs/org/apache/storm/StormSubmitter.html)
     will upload the jar at topology-jar-path when the topology is submitted.
+
+    When you want to ship other jars which is not included to application jar, you can pass them to --jars option with comma-separated string.
+    For example, --jars "your-local-jar.jar,your-local-jar2.jar" will load your-local-jar.jar and your-local-jar2.jar.
+    And when you want to ship maven artifacts and its transitive dependencies, you can pass them to --artifacts with comma-separated string.
+    You can also exclude some dependencies like what you're doing in maven pom.
+    Please add exclusion artifacts with '^' separated string after the artifact.
+    For example, --artifacts "redis.clients:jedis:2.9.0,org.apache.kafka:kafka_2.10:0.8.2.2^org.slf4j:slf4j-log4j12" will load jedis and kafka artifact and all of transitive dependencies but exclude slf4j-log4j12 from kafka.
+
+    Complete example of both options is here: `./bin/storm jar example/storm-starter/storm-starter-topologies-*.jar org.apache.storm.starter.RollingTopWords blobstore-remote2 remote --jars "./external/storm-redis/storm-redis-1.1.0.jar,./external/storm-kafka/storm-kafka-1.1.0.jar" --artifacts "redis.clients:jedis:2.9.0,org.apache.kafka:kafka_2.10:0.8.2.2^org.slf4j:slf4j-log4j12"`
+
+    When you pass jars and/or artifacts options, StormSubmitter will upload them when the topology is submitted, and they will be included to classpath of both the process which runs the class, and also workers for that topology.
     """
+    global DEP_JARS_OPTS, DEP_ARTIFACTS_OPTS
+
+    local_jars = DEP_JARS_OPTS
+    artifact_to_file_jars = resolve_dependencies(DEP_ARTIFACTS_OPTS)
+
     transform_class = confvalue("client.jartransformer.class", [CLUSTER_CONF_DIR])
     if (transform_class != None and transform_class != "nil"):
         tmpjar = os.path.join(tempfile.gettempdir(), uuid.uuid1().hex+".jar")
         exec_storm_class("org.apache.storm.daemon.ClientJarTransformerRunner", args=[transform_class, jarfile, tmpjar], fork=True, daemon=False)
+        extra_jars = [tmpjar, USER_CONF_DIR, STORM_BIN_DIR]
+        extra_jars.extend(local_jars)
+        extra_jars.extend(artifact_to_file_jars.values())
         topology_runner_exit_code = exec_storm_class(
                 klass,
                 jvmtype="-client",
-                extrajars=[tmpjar, USER_CONF_DIR, STORM_BIN_DIR],
+                extrajars=extra_jars,
                 args=args,
                 daemon=False,
                 fork=True,
-                jvmopts=JAR_JVM_OPTS + ["-Dstorm.jar=" + tmpjar])
+                jvmopts=JAR_JVM_OPTS + ["-Dstorm.jar=" + tmpjar] +
+                        ["-Dstorm.dependency.jars=" + ",".join(local_jars)] +
+                        ["-Dstorm.dependency.artifacts=" + json.dumps(artifact_to_file_jars)])
         os.remove(tmpjar)
         sys.exit(topology_runner_exit_code)
     else:
+        extra_jars=[jarfile, USER_CONF_DIR, STORM_BIN_DIR]
+        extra_jars.extend(local_jars)
+        extra_jars.extend(artifact_to_file_jars.values())
         exec_storm_class(
             klass,
             jvmtype="-client",
-            extrajars=[jarfile, USER_CONF_DIR, STORM_BIN_DIR],
+            extrajars=extra_jars,
             args=args,
             daemon=False,
-            jvmopts=JAR_JVM_OPTS + ["-Dstorm.jar=" + jarfile])
+            jvmopts=JAR_JVM_OPTS + ["-Dstorm.jar=" + jarfile] +
+                    ["-Dstorm.dependency.jars=" + ",".join(local_jars)] +
+                    ["-Dstorm.dependency.artifacts=" + json.dumps(artifact_to_file_jars)])
 
 def sql(sql_file, topology_name):
     """Syntax: [storm sql sql-file topology-name]
 
     Compiles the SQL statements into a Trident topology and submits it to Storm.
+
+    --jars and --artifacts options available for jar are also applied to sql command.
+    Please refer "help jar" to see how to use --jars and --artifacts options.
+    You normally want to pass these options since you need to set data source to your sql which is an external storage in many cases.
     """
+    global DEP_JARS_OPTS, DEP_ARTIFACTS_OPTS
+
+    local_jars = DEP_JARS_OPTS
+    artifact_to_file_jars = resolve_dependencies(DEP_ARTIFACTS_OPTS)
+
+    sql_core_jars = get_jars_full(STORM_DIR + "/external/sql/storm-sql-core")
+    sql_runtime_jars = get_jars_full(STORM_DIR + "/external/sql/storm-sql-runtime")
+
+    # include storm-sql-runtime jar(s) to local jar list
+    local_jars.extend(sql_runtime_jars)
+
     extrajars=[USER_CONF_DIR, STORM_BIN_DIR]
-    extrajars.extend(get_jars_full(STORM_DIR + "/external/sql/storm-sql-core"))
-    extrajars.extend(get_jars_full(STORM_DIR + "/external/sql/storm-sql-runtime"))
+    extrajars.extend(local_jars)
+    extrajars.extend(artifact_to_file_jars.values())
+
+    # include this for running StormSqlRunner, but not for generated topology
+    extrajars.extend(sql_core_jars)
+
     exec_storm_class(
         "org.apache.storm.sql.StormSqlRunner",
         jvmtype="-client",
         extrajars=extrajars,
         args=[sql_file, topology_name],
-        daemon=False)
+        daemon=False,
+        jvmopts=["-Dstorm.dependency.jars=" + ",".join(local_jars)] +
+                ["-Dstorm.dependency.artifacts=" + json.dumps(artifact_to_file_jars)])
 
 def kill(*args):
     """Syntax: [storm kill topology-name [-w wait-time-secs]]
@@ -746,6 +832,8 @@ def parse_config_opts(args):
   curr.reverse()
   config_list = []
   args_list = []
+  jars_list = []
+  artifacts_list = []
 
   while len(curr) > 0:
     token = curr.pop()
@@ -754,18 +842,24 @@ def parse_config_opts(args):
     elif token == "--config":
       global CONFFILE
       CONFFILE = curr.pop()
+    elif token == "--jars":
+      jars_list.extend(curr.pop().split(','))
+    elif token == "--artifacts":
+      artifacts_list.extend(curr.pop().split(','))
     else:
       args_list.append(token)
 
-  return config_list, args_list
+  return config_list, jars_list, artifacts_list, args_list
 
 def main():
     if len(sys.argv) <= 1:
         print_usage()
         sys.exit(-1)
-    global CONFIG_OPTS
-    config_list, args = parse_config_opts(sys.argv[1:])
+    global CONFIG_OPTS, DEP_JARS_OPTS, DEP_ARTIFACTS_OPTS
+    config_list, jars_list, artifacts_list, args = parse_config_opts(sys.argv[1:])
     parse_config(config_list)
+    DEP_JARS_OPTS = jars_list
+    DEP_ARTIFACTS_OPTS = artifacts_list
     COMMAND = args[0]
     ARGS = args[1:]
     (COMMANDS.get(COMMAND, unknown_command))(*ARGS)
--- a/docs/Command-line-client.md
+++ b/docs/Command-line-client.md
@@ -8,6 +8,7 @@ This page describes all the commands that are possible with the "storm" command 
 These commands are:
 
 1. jar
+1. sql
 1. kill
 1. activate
 1. deactivate
@@ -32,7 +33,6 @@ These commands are:
 1. pacemaker
 1. set_log_level
 1. shell
-1. sql
 1. upload-credentials
 1. version
 1. help
@@ -42,6 +42,22 @@ These commands are:
 Syntax: `storm jar topology-jar-path class ...`
 
 Runs the main method of `class` with the specified arguments. The storm jars and configs in `~/.storm` are put on the classpath. The process is configured so that [StormSubmitter](javadocs/org/apache/storm/StormSubmitter.html) will upload the jar at `topology-jar-path` when the topology is submitted.
+
+When you want to ship other jars which is not included to application jar, you can pass them to `--jars` option with comma-separated string.
+For example, --jars "your-local-jar.jar,your-local-jar2.jar" will load your-local-jar.jar and your-local-jar2.jar.
+And when you want to ship maven artifacts and its transitive dependencies, you can pass them to `--artifacts` with comma-separated string. You can also exclude some dependencies like what you're doing in maven pom. Please add exclusion artifacts with '^' separated string after the artifact. For example, `--artifacts "redis.clients:jedis:2.9.0,org.apache.kafka:kafka_2.10:0.8.2.2^org.slf4j:slf4j-log4j12"` will load jedis and kafka artifact and all of transitive dependencies but exclude slf4j-log4j12 from kafka.
+
+Complete example of both options is here: `./bin/storm jar example/storm-starter/storm-starter-topologies-*.jar org.apache.storm.starter.RollingTopWords blobstore-remote2 remote --jars "./external/storm-redis/storm-redis-1.1.0.jar,./external/storm-kafka/storm-kafka-1.1.0.jar" --artifacts "redis.clients:jedis:2.9.0,org.apache.kafka:kafka_2.10:0.8.2.2^org.slf4j:slf4j-log4j12"`
+
+When you pass jars and/or artifacts options, StormSubmitter will upload them when the topology is submitted, and they will be included to classpath of both the process which runs the class, and also workers for that topology.
+
+### sql
+
+Syntax: `storm sql sql-file topology-name`
+
+Compiles the SQL statements into a Trident topology and submits it to Storm.
+
+`--jars` and `--artifacts` options are also applied to `sql` command. You normally want to pass these options since you need to set data source to your sql which is an external storage in many cases.
 
 ### kill
 
@@ -65,7 +81,7 @@ Deactivates the specified topology's spouts.
 
 Syntax: `storm rebalance topology-name [-w wait-time-secs] [-n new-num-workers] [-e component=parallelism]*`
 
-Sometimes you may wish to spread out where the workers for a topology are running. For example, let's say you have a 10 node cluster running 4 workers per node, and then let's say you add another 10 nodes to the cluster. You may wish to have Storm spread out the workers for the running topology so that each node runs 2 workers. One way to do this is to kill the topology and resubmit it, but Storm provides a "rebalance" command that provides an easier way to do this. 
+Sometimes you may wish to spread out where the workers for a topology are running. For example, let's say you have a 10 node cluster running 4 workers per node, and then let's say you add another 10 nodes to the cluster. You may wish to have Storm spread out the workers for the running topology so that each node runs 2 workers. One way to do this is to kill the topology and resubmit it, but Storm provides a "rebalance" command that provides an easier way to do this.
 
 Rebalance will first deactivate the topology for the duration of the message timeout (overridable with the -w flag) and then redistribute the workers evenly around the cluster. The topology will then return to its previous state of activation (so a deactivated topology will still be deactivated and an activated topology will go back to being activated).
 
@@ -216,7 +232,7 @@ See Setting up a Storm cluster for more information.(http://storm.apache.org/doc
 Syntax: `storm set_log_level -l [logger name]=[log level][:optional timeout] -r [logger name] topology-name`
 
 Dynamically change topology log levels
-    
+
 where log level is one of: ALL, TRACE, DEBUG, INFO, WARN, ERROR, FATAL, OFF
 and timeout is integer seconds.
 
@@ -244,12 +260,6 @@ Syntax: `storm shell resourcesdir command args`
 Makes constructing jar and uploading to nimbus for using non JVM languages
 
 eg: `storm shell resources/ python topology.py arg1 arg2`
-
-### sql
-
-Syntax: `storm sql sql-file topology-name`
-
-Compiles the SQL statements into a Trident topology and submits it to Storm.
 
 ### upload-credentials
 
--- a/storm-core/src/clj/org/apache/storm/config.clj
+++ b/storm-core/src/clj/org/apache/storm/config.clj
@@ -141,6 +141,10 @@
   [conf stormconf-path]
   (merge conf (clojurify-structure (Utils/fromCompressedJsonConf (FileUtils/readFileToByteArray (File. stormconf-path))))))
 
+(defn read-supervisor-storm-code-given-path
+  [stormcode-path]
+  (Utils/deserialize (FileUtils/readFileToByteArray (File. stormcode-path)) StormTopology))
+
 (defn master-storm-metafile-path [stormroot ]
   (str stormroot file-path-separator "storm-code-distributor.meta"))
 
@@ -224,8 +228,7 @@
   [conf storm-id]
   (let [stormroot (supervisor-stormdist-root conf storm-id)
         topology-path (supervisor-stormcode-path stormroot)]
-    (Utils/deserialize (FileUtils/readFileToByteArray (File. topology-path)) StormTopology)
-    ))
+    (read-supervisor-storm-code-given-path topology-path)))
 
 (defn worker-user-root [conf]
   (str (absolute-storm-local-dir conf) "/workers-users"))
--- a/storm-core/src/clj/org/apache/storm/daemon/supervisor.clj
+++ b/storm-core/src/clj/org/apache/storm/daemon/supervisor.clj
@@ -28,7 +28,7 @@
   (:import [org.apache.storm.utils NimbusLeaderNotFoundException VersionInfo])
   (:import [java.nio.file Files StandardCopyOption])
   (:import [org.apache.storm Config])
-  (:import [org.apache.storm.generated WorkerResources ProfileAction])
+  (:import [org.apache.storm.generated WorkerResources ProfileAction StormTopology])
   (:import [org.apache.storm.localizer LocalResource])
   (:use [org.apache.storm.daemon common])
   (:require [org.apache.storm.command [healthcheck :as healthcheck]])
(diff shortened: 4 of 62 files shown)


HeartSaVioR:
Changed to \[WIP] since there're failures of unit tests. I'll take care of that so please continue reviewing pull request.

HeartSaVioR:
Fixed build failure on supervisor_test.

HeartSaVioR:
Builds for PR against 1.x branch is failing and I fixed that from #1609. I'll rebase once #1609 is merged to 1.x-branch.

HeartSaVioR:
[STORM-2016](http://issues.apache.org/jira/browse/STORM-2016) can also resolve [STORM-1435](http://issues.apache.org/jira/browse/STORM-1435) via applying different strategy, specifying dependencies to --jars and --packages instead of creating uber jar.
(Yes in fact STORM-2016 is started to resolve STORM-1435.)

Note that `storm sql` adds storm-sql-core and its transitive dependencies to classpath to run Runner, and also adds storm-sql-runtime and its transitive dependencies to classpath to run topology by adding them to --jars option. 

With [STORM-2023](https://issues.apache.org/jira/browse/STORM-2023) (#1610), we can run the example of storm-sql by below command.

> ./bin/storm sql order_filtering.sql order_filtering --packages "org.apache.storm:storm-sql-kafka:1.1.1-SNAPSHOT,org.apache.storm:storm-kafka:1.1.1-SNAPSHOT,org.apache.kafka:kafka_2.10:0.8.2.2\!org.slf4j:slf4j-log4j12,org.apache.kafka:kafka-clients:0.8.2.2"

Without STORM-2023, we still have a chance to run the example of storm-sql by below command 

> ./bin/storm sql order_filtering.sql order_filtering --packages "org.apache.storm:storm-sql-kafka:1.1.1-SNAPSHOT,org.apache.storm:storm-kafka:1.1.1-SNAPSHOT,org.apache.kafka:kafka_2.10:0.8.2.2\!org.slf4j:slf4j-log4j12,org.apache.kafka:kafka-clients:0.8.2.2,org.apache.calcite:calcite-core:1.4.0-incubating\!commons-dbcp:commons-dbcp\!com.google.code.findbugs:jsr305\!org.codehaus.janino:janino\!org.codehaus.janino:commons-compiler\!org.pentaho:pentaho-aggdesigner-algorithm\!com.fasterxml.jackson.core:jackson-annotations"

bin/storm.py
```diff
@@ -154,6 +157,39 @@ def confvalue(name, extrapaths, daemon=True):
             return " ".join(tokens[1:])
     return ""
 
+def resolve_dependencies(packages):
```
revans2: nit: It would be nice to have a short circuit here so if there are no dependencies to resolve we don't have to call into another java application.

revans2:
Conceptually I like the idea, and overall the code looks good.  My biggest concern is around cleanup of the bobs after the topology finishes.  I don't see anywhere in the code that it is doing that, but I didn't dig too deeply into it.

satishd:
@revans2 This issue was raised in the proposal discussion [here](https://mail-archives.apache.org/mod_mbox/storm-dev/201608.mbox/%3CCAF5108hbsHL3MO6HU47ncayV%2BgcMZ1s-B2izvRFmL7ot-TcqJg%40mail.gmail.com%3E). I do not think that proposal handles removing the blobs.
@HeartSaVioR Do you want to address this issue in this PR or you have plans to address it later? 

I have not yet started reviewing this PR, I will try to review it soon.

HeartSaVioR:
@revans2 @satishd 
Yes I've excluded that for simplification. Since this is a new feature I'd love to show that this concept works.
Moreover, in order to do that Nimbus should be modified but I also concerned with Nimbus port work. If we (including @longdafeng) are OK to address this before porting I'll do that work in this PR.

Btw, since jars are not shared but packages and transitive dependencies are meant to be shared, we should handle them different. We should remove blobs which are passed to --jars.

bin/storm.py
HeartSaVioR: Good point. Will address.

HeartSaVioR: Addressed.

abhishekagarwal87:
@HeartSaVioR - not sure if it will be of help, but you can also checkout https://github.com/apache/storm/pull/1296/files. I was trying out something similar.

HeartSaVioR:
@abhishekagarwal87 
Thanks, I took a look and your modification is simpler.
Btw, I was thinking same approach, but there're some cases to handle while assembling jars (for example, the content of schema files for Spring modules should be all added into one file) so it doesn't work that cases. That's why I choose this approach. It's already used widely by Spark users indeed.

HeartSaVioR:
@revans2 @satishd 
Addressed removing jars when topology is killed. Please note that artifacts will be not removed since it's shared among whole topologies.
Since it modifies Nimbus, we can still merge this in without this feature, but I don't want to be blocked by the thing even pull request is not submitted.

HeartSaVioR:
Added documentation, and replace packages to artifacts since it's clearer for its meaning.

I think it's done with 1.x branch, so I'll go on porting this to master branch.

HeartSaVioR:
Porting this to master forces me to resolve huge conflict or modify the code manually. We shouldn't leave progress of Java port as it is.

bin/storm.py
```diff
@@ -754,18 +827,24 @@ def parse_config_opts(args):
     elif token == "--config":
       global CONFFILE
       CONFFILE = curr.pop()
+    elif token == "--jars":
+      jars_list.extend(curr.pop().split(','))
+    elif token == "--artifacts":
```
manuzhang: Is "artifacts" a maven specific word ? How about "dependencies" ?

bin/storm.py
```diff
@@ -754,18 +827,24 @@ def parse_config_opts(args):
     elif token == "--config":
       global CONFFILE
       CONFFILE = curr.pop()
+    elif token == "--jars":
```
manuzhang: Are these two options common for commands ? If so, could we have some help info in the `print_command` function ?

bin/storm.py
HeartSaVioR: No, the term `artifact` is also used from Ivy, too.
http://ant.apache.org/ivy/history/latest-milestone/terminology.html#artifact

I think term `dependencies` is closer to jars + artifacts since jars will be also added to classpath.

bin/storm.py
HeartSaVioR: It's only used for `storm jar` and `storm sql`, but showing them to `print_command` would be better. I'll address this.
Thanks for the suggestion.

HeartSaVioR: It would be better to add its usage to comment doc of `jar` and `sql` since print_usage(jar) will print out the usage of jar, and so on sql.

docs/Command-line-client.md
```diff
@@ -43,6 +43,19 @@ Syntax: `storm jar topology-jar-path class ...`
 
 Runs the main method of `class` with the specified arguments. The storm jars and configs in `~/.storm` are put on the classpath. The process is configured so that [StormSubmitter](javadocs/org/apache/storm/StormSubmitter.html) will upload the jar at `topology-jar-path` when the topology is submitted.
 
+When you want to ship other jars which is not included to application jar, you can pass them to `--jars` option with comma-separated string.
```
manuzhang: neither system provided nor included to application jar ?

manuzhang: Could we have some examples here ?

HeartSaVioR: That is not mandatory, but it only makes sense when that jar is not presented neither system provided nor included to application jar. If system or application jar provides that, that will be used. This is similar to precedence of classloading between system and application jar.

HeartSaVioR: And yes we can have example on this. Will address.

HeartSaVioR:
@manuzhang Thanks for checking the details of documentation. 
Please leave a comment when you're done reviewing of this. Thanks!

HeartSaVioR:
@revans2 @satishd @abhishekagarwal87 
Please review again when you get some time. I'll squash the commits and go on porting this to master branch once some of us give +1 for this.
Thanks in advance!

docs/Command-line-client.md
manuzhang: It's not clear whether system jar or application jar will get loaded and may create some surprise if the uploaded jar doesn't work.  Maybe some warnings somewhere.

HeartSaVioR: It would be ideal if we can show same artifacts with different versions as warning, but I'm not sure it would be possible, especially application jar is an uber jar so we can't get artifacts within that jar.

If you mean we just would be better to show some general warnings, yes that would be good to have. Btw, this is closer to general behavior of determining worker classpath (even without these options), so it would be better to explain this in general manner.
I'm not sure where is the best place to address this. Let me think about it.

bin/storm.py
manuzhang: It's a bit strange to have `storm jar --jars`. Maybe `--dep-jars` ?

storm-core/src/jvm/org/apache/storm/generated/Assignment.java
```diff
@@ -787,15 +787,15 @@ public void read(org.apache.thrift.protocol.TProtocol iprot, Assignment struct)
           case 2: // NODE_HOST
             if (schemeField.type == org.apache.thrift.protocol.TType.MAP) {
               {
-                org.apache.thrift.protocol.TMap _map548 = iprot.readMapBegin();
-                struct.node_host = new HashMap<String,String>(2*_map548.size);
-                String _key549;
-                String _val550;
-                for (int _i551 = 0; _i551 < _map548.size; ++_i551)
+                org.apache.thrift.protocol.TMap _map564 = iprot.readMapBegin();
```
manuzhang: Forgive me for a dumb and irrelevant question. Why do we keep generated files in source ?

storm-core/src/clj/org/apache/storm/daemon/supervisor.clj
```diff
@@ -893,29 +891,61 @@
...
     (for [[k, data] blobstore-map]
       (get-blob-localname data k))))
 
 (defn download-blobs-for-topology-succeed?
+  [target-dir file-names]
+  (if-not (empty? file-names)
+    (every? #(Utils/checkFileExists target-dir %) file-names)
+    true))
+
+(defn download-blobs-in-blobstore-map-for-topology-succeed?
```
manuzhang: It seems these methods could be private ?

storm-core/src/clj/org/apache/storm/daemon/supervisor.clj
```diff
@@ -875,14 +875,12 @@
     (worker-launcher-and-wait conf (storm-conf TOPOLOGY-SUBMITTER-USER) ["blob" path] :log-prefix (str "setup blob permissions for " path))))
 
 (defn download-blobs-for-topology!
-  "Download all blobs listed in the topology configuration for a given topology."
-  [conf stormconf-path localizer tmproot]
-  (let [storm-conf (read-supervisor-storm-conf-given-path conf stormconf-path)
-        blobstore-map (storm-conf TOPOLOGY-BLOBSTORE-MAP)
+  [conf storm-conf localizer tmproot blobs fn-local-resources fn-symlink-name]
```
manuzhang: update the comments instead of removing it ?

storm-core/src/jvm/org/apache/storm/generated/Assignment.java
HeartSaVioR: There's no dumb question. This is just a one of practices on Storm project. As always there're some pros and cons regarding this:

pros: Thrift compiler is only needed for contributors who modifies storm.thrift, and the case is actually not often.
cons: Thrift generates codes which makes huge diff even though just adding only one field.

IMO, installing thrift compiler because of just build project is bad, especially OSPs are using different versions of thrift. While disadvantage is also not a trivial, I think advantage is valuable enough.

storm-core/src/clj/org/apache/storm/daemon/supervisor.clj
HeartSaVioR: I don't mind if it's private or not, since we already ported this to Java in master branch.
There're other functions which could be private.

storm-core/src/clj/org/apache/storm/daemon/supervisor.clj
HeartSaVioR: OK I'll do.

storm-core/src/clj/org/apache/storm/config.clj
```diff
@@ -141,6 +141,10 @@
   [conf stormconf-path]
   (merge conf (clojurify-structure (Utils/fromCompressedJsonConf (FileUtils/readFileToByteArray (File. stormconf-path))))))
 
+(defn read-supervisor-storm-code-given-path
```
manuzhang: maybe update the `read-supervisor-topology` function below to use this

HeartSaVioR: Great point! Will address.

external/storm-submit/src/main/java/org/apache/storm/submit/command/DependencyResolverMain.java
```diff
@@ -0,0 +1,105 @@
...
+import java.util.Arrays;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
+public class DependencyResolverMain {
+
+    public static void main(String[] args) {
+        if (args.length < 1) {
+            throw new IllegalArgumentException("packages must be presented.");
```
manuzhang: artifacts now ? Same for error message below

external/storm-submit/src/main/java/org/apache/storm/submit/dependency/AetherUtils.java
```diff
@@ -0,0 +1,81 @@
...
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.List;
+
+public class AetherUtils {
+    private AetherUtils() {
+    }
+
+    public static Dependency parseDependency(String dependency) {
+        List<String> dependencyAndExclusions = Arrays.asList(dependency.split("!"));
```
manuzhang: "!" is a special bach character that refers to previous command. Although we can escape it with "\", is there a better choice ?

external/storm-submit/src/main/java/org/apache/storm/submit/command/DependencyResolverMain.java
HeartSaVioR: Great finding. Will address.

external/storm-submit/src/main/java/org/apache/storm/submit/dependency/AetherUtils.java
```diff
@@ -0,0 +1,81 @@
...
+
+        return new Exclusion(groupId, artifactId, classifier, extension);
+    }
+
+    public static String artifactToString(Artifact artifact) {
+        StringBuilder buffer = new StringBuilder(128);
+        buffer.append(artifact.getGroupId());
+        buffer.append(':').append(artifact.getArtifactId());
+        buffer.append(':').append(artifact.getExtension());
+        if(artifact.getClassifier().length() > 0) {
```
manuzhang: nit: space here

external/storm-submit/src/main/java/org/apache/storm/submit/dependency/AetherUtils.java
HeartSaVioR: '!' occasionally represents 'not', so I think it would be easier to be recognized as excluded. Do we feel similar from '~'?

Refer to http://www.gnu.org/software/bash/manual/bashref.html#Quoting we should avoid to use '$', '`', '\', '"', '!', '@', '*' in double quotes if we want to avoid escape.

storm-core/src/jvm/org/apache/storm/dependency/DependencyUploader.java
```diff
@@ -0,0 +1,166 @@
...
+    private boolean uploadDependencyToBlobStore(String key, File dependency)
+            throws KeyAlreadyExistsException, AuthorizationException, IOException {
+
+        boolean uploadNew = false;
+        try {
+            // FIXME: we can filter by listKeys() with local blobstore when STORM-1986 is going to be resolved
+            // as a workaround, we call getBlobMeta() for all keys
+            blobStore.getBlobMeta(key);
+        } catch (KeyNotFoundException e) {
+            // TODO: do we want to add ACL here?
```
satishd: Can we file JIRA for this? This is a low priority to have any acls user may want to provide.

storm-core/src/jvm/org/apache/storm/StormSubmitter.java
```diff
@@ -233,6 +236,23 @@ public static void submitTopologyAs(String name, Map stormConf, StormTopology to
...
+
+                // Dependency uploading only makes sense for distributed mode
+                List<String> jarsBlobKeys;
+                List<String> artifactsBlobKeys;
+
+                DependencyUploader uploader = new DependencyUploader();
+                try {
+                    uploader.init();
+                    jarsBlobKeys = uploadDependencyJarsToBlobStore(uploader);
+                    artifactsBlobKeys = uploadDependencyArtifactsToBlobStore(uploader);
```
satishd: Both `jarsBlobKeys` and `artifactsBlobKeys` should be removed from blob store when topology submission is failed.

storm-core/src/jvm/org/apache/storm/dependency/DependencyUploader.java
HeartSaVioR: Before that, we need to check that nimbus and supervisor can access these blobs. I didn't look deeply with blobstore so I would like someone to confirm this.
@revans2 Could you guide this thing?

storm-core/src/jvm/org/apache/storm/StormSubmitter.java
HeartSaVioR: We can do that for jarsBlobKeys but can't do that for artifactsBlobKeys since we share artifact blobs and we skip uploads if there're same artifact uploaded before.

I'll address this for jarsBlobKeys. Thanks!

external/storm-submit/src/main/java/org/apache/storm/submit/dependency/AetherUtils.java
HeartSaVioR: OK will address.

storm-core/src/jvm/org/apache/storm/dependency/DependencyUploader.java
satishd: I do not think we should have acls for artifacts. If it is possible to support acls in this scenario, it should be done only for jars option.

satishd:
@HeartSaVioR Overall LGTM. I will revisit once the comments are addressed/resolved.

HeartSaVioR:
@manuzhang Thanks, I addressed your latest comments from https://github.com/apache/storm/pull/1608/commits/a9f4886fd2f093c7a4450b55ef809fc9826c820e

Not sure we have better representation than '!' without conflicting bash special char and also not used char for maven artifact.

@satishd Thanks, I addressed your last comment from https://github.com/apache/storm/pull/1608/commits/17cb2dd333205578bed8f8af8a51072b18daf32d

For custom ACL for jars blobs I'd rather see possible issues first when we set ACL to blobs. If it's no issue at all and we agree that there could be a case of this I'll file an issue.
Please take a look again. Thanks!

external/storm-submit/src/main/java/org/apache/storm/submit/dependency/AetherUtils.java
manuzhang: another common character for **not** is `^`

storm-core/src/jvm/org/apache/storm/StormSubmitter.java
```diff
@@ -238,39 +239,24 @@ public static void submitTopologyAs(String name, Map stormConf, StormTopology to
...
                 DependencyUploader uploader = new DependencyUploader();
                 try {
                     uploader.init();
                     jarsBlobKeys = uploadDependencyJarsToBlobStore(uploader);
                     artifactsBlobKeys = uploadDependencyArtifactsToBlobStore(uploader);
+                    setDependencyBlobsToTopology(topology, jarsBlobKeys, artifactsBlobKeys);
+                    submitTopologyInDistributeMode(name, topology, opts, progressListener, asUser, conf, serConf);
+                } catch (Throwable e) {
+                    // remove uploaded jars blobs, not artifacts since they're shared across the cluster
+                    uploader.deleteBlobs(jarsBlobKeys);
```
satishd: InvalidTopologyException and AlreadyAliveException definitely indicates that the topology is not successfully deployed and respective jars can be removed from blob store.
 
@HeartSaVioR 
But this code assumes that any Exception thrown here indicates that the submitted topolgy is not deployed successfully.  I am not sure whether that assumption is always right.

HeartSaVioR: @satishd 
In fact, submitTopologyWithOpts() can also throw AuthorizationException and org.apache.thrift.TException, too.

Let's back to flow of logic.
Please note that we didn't catch Throwable (some known Exceptions and also RuntimeException) and just pass to caller. And there's also no return for Nimbus.Client.submitTopology(), which indicates that unless we get TException or other Exceptions it should be treated as succeed. (If not we can't guarantee topology deploy is succeed.)

I think this assumption is right for now, and assumption should be right unless we add some operations after calling submitTopologyInDistributeMode() or adding some operations after logging "Finished submitting topology".

bin/storm.py
HeartSaVioR: I think --jars is representing its meaning properly so I'd like to keep it as it is.
If there's other one which would want to change option to non-duplicated name, I'll do that.

storm-core/src/jvm/org/apache/storm/StormSubmitter.java
satishd: @HeartSaVioR Is not there a possibility that the topology is submitted successfully on nimbus but client receives thrift error because of some timeout or connection error? 
My preference is to remove those topologies only when we know definitely that the topology submission is failed.

HeartSaVioR: @satishd 
Yes I think the scenarios you stated can be occurred. But also Nimbus throws TException when Nimbus is unable to handle the submission, say, runtime exception occurred.

So it's more like loosely catching with less risk vs hardly catching with more risk.
(Risk could be reduced via checking topology name once again indeed.)

After thinking it again, leaving blobs seems less harmful than whole contents of topology couldn't be downloaded at all. I'll follow your suggestion.

external/storm-submit/src/main/java/org/apache/storm/submit/dependency/AetherUtils.java
HeartSaVioR: Personally I prefer `!`, but if escaping matters I think `^` is also fine. I'll address this.

storm-core/src/jvm/org/apache/storm/StormSubmitter.java
HeartSaVioR: Addressed.

external/storm-submit/src/main/java/org/apache/storm/submit/dependency/AetherUtils.java
HeartSaVioR: Addressed.

satishd:
+1
@HeartSaVioR Nice work.

HeartSaVioR:
@satishd 
Thanks for the thoughtful review and vote. I rebased the commits into one.

@revans2 @manuzhang @abhishekagarwal87 
Please continue reviewing. I'll merge this by tomorrow if there's no more discussion. Thanks.

harshach:
@HeartSaVioR can you give it 2 days given that its a big patch. Thanks.

HeartSaVioR:
@harshach Sure, I'll wait for the review during this week. Btw, Actual modification is around (or less than) 1000 lines, not that much.

manuzhang:
@HeartSaVioR The code looks good (non-binding). I tried running `storm sql` but ended up with timeout exceptions 

```
Exception in thread "main" java.lang.RuntimeException: org.sonatype.aether.resolution.DependencyResolutionException: Failed to collect dependencies for org.apache.storm:storm-sql-kafka:jar:1.1.1-SNAPSHOT (compile)
        at org.apache.storm.submit.command.DependencyResolverMain.main(DependencyResolverMain.java:63)
Caused by: org.sonatype.aether.resolution.DependencyResolutionException: Failed to collect dependencies for org.apache.storm:storm-sql-kafka:jar:1.1.1-SNAPSHOT (compile)
        at org.sonatype.aether.impl.internal.DefaultRepositorySystem.resolveDependencies(DefaultRepositorySystem.java:371)
        at org.apache.storm.submit.dependency.DependencyResolver.resolve(DependencyResolver.java:67)
        at org.apache.storm.submit.command.DependencyResolverMain.main(DependencyResolverMain.java:52)
Caused by: org.sonatype.aether.collection.DependencyCollectionException: Failed to collect dependencies for org.apache.storm:storm-sql-kafka:jar:1.1.1-SNAPSHOT (compile)
        at org.sonatype.aether.impl.internal.DefaultDependencyCollector.collectDependencies(DefaultDependencyCollector.java:258)
        at org.sonatype.aether.impl.internal.DefaultRepositorySystem.resolveDependencies(DefaultRepositorySystem.java:333)
        ... 2 more
Caused by: org.sonatype.aether.resolution.ArtifactDescriptorException: Failed to read artifact descriptor for org.apache.kafka:kafka_2.10:jar:0.8.2.2\
        at org.apache.maven.repository.internal.DefaultArtifactDescriptorReader.loadPom(DefaultArtifactDescriptorReader.java:282)
        at org.apache.maven.repository.internal.DefaultArtifactDescriptorReader.readArtifactDescriptor(DefaultArtifactDescriptorReader.java:172)
        at org.sonatype.aether.impl.internal.DefaultDependencyCollector.process(DefaultDependencyCollector.java:412)
        at org.sonatype.aether.impl.internal.DefaultDependencyCollector.collectDependencies(DefaultDependencyCollector.java:240)
        ... 3 more
Caused by: org.sonatype.aether.resolution.ArtifactResolutionException: Could not transfer artifact org.apache.kafka:kafka_2.10:pom:0.8.2.2\ from/to central (http://repo1.maven.org/maven2/): Error transferring file: Connection timed out
        at org.sonatype.aether.impl.internal.DefaultArtifactResolver.resolve(DefaultArtifactResolver.java:537)
        at org.sonatype.aether.impl.internal.DefaultArtifactResolver.resolveArtifacts(DefaultArtifactResolver.java:216)
        at org.sonatype.aether.impl.internal.DefaultArtifactResolver.resolveArtifact(DefaultArtifactResolver.java:193)
        at org.apache.maven.repository.internal.DefaultArtifactDescriptorReader.loadPom(DefaultArtifactDescriptorReader.java:267)
        ... 6 more
Caused by: org.sonatype.aether.transfer.ArtifactTransferException: Could not transfer artifact org.apache.kafka:kafka_2.10:pom:0.8.2.2\ from/to central (http://repo1.maven.org/maven2/): Error transferring file: Connection timed out
        at org.sonatype.aether.connector.wagon.WagonRepositoryConnector$4.wrap(WagonRepositoryConnector.java:975)
        at org.sonatype.aether.connector.wagon.WagonRepositoryConnector$4.wrap(WagonRepositoryConnector.java:966)
        at org.sonatype.aether.connector.wagon.WagonRepositoryConnector$GetTask.flush(WagonRepositoryConnector.java:707)
        at org.sonatype.aether.connector.wagon.WagonRepositoryConnector$GetTask.flush(WagonRepositoryConnector.java:701)
        at org.sonatype.aether.connector.wagon.WagonRepositoryConnector.get(WagonRepositoryConnector.java:452)
        at org.sonatype.aether.impl.internal.DefaultArtifactResolver.resolve(DefaultArtifactResolver.java:456)
        ... 9 more
Caused by: org.apache.maven.wagon.TransferFailedException: Error transferring file: Connection timed out
        at org.apache.maven.wagon.providers.http.LightweightHttpWagon.fillInputData(LightweightHttpWagon.java:143)
        at org.apache.maven.wagon.StreamWagon.getInputStream(StreamWagon.java:116)
        at org.apache.maven.wagon.StreamWagon.getIfNewer(StreamWagon.java:88)
        at org.apache.maven.wagon.StreamWagon.get(StreamWagon.java:61)
        at org.sonatype.aether.connector.wagon.WagonRepositoryConnector$GetTask.run(WagonRepositoryConnector.java:615)
        at org.sonatype.aether.util.concurrency.RunnableErrorForwarder$1.run(RunnableErrorForwarder.java:60)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
        at java.lang.Thread.run(Thread.java:745)
Caused by: java.net.ConnectException: Connection timed out
        at java.net.PlainSocketImpl.socketConnect(Native Method)
        at java.net.AbstractPlainSocketImpl.doConnect(AbstractPlainSocketImpl.java:350)
        at java.net.AbstractPlainSocketImpl.connectToAddress(AbstractPlainSocketImpl.java:206)
        at java.net.AbstractPlainSocketImpl.connect(AbstractPlainSocketImpl.java:188)
        at java.net.SocksSocketImpl.connect(SocksSocketImpl.java:392)
        at java.net.Socket.connect(Socket.java:589)
        at java.net.Socket.connect(Socket.java:538)
        at sun.net.NetworkClient.doConnect(NetworkClient.java:180)
        at sun.net.www.http.HttpClient.openServer(HttpClient.java:432)
        at sun.net.www.http.HttpClient.openServer(HttpClient.java:527)
        at sun.net.www.http.HttpClient.<init>(HttpClient.java:211)
        at sun.net.www.http.HttpClient.New(HttpClient.java:308)
        at sun.net.www.http.HttpClient.New(HttpClient.java:326)
        at sun.net.www.protocol.http.HttpURLConnection.getNewHttpClient(HttpURLConnection.java:1169)
        at sun.net.www.protocol.http.HttpURLConnection.plainConnect0(HttpURLConnection.java:1105)
        at sun.net.www.protocol.http.HttpURLConnection.plainConnect(HttpURLConnection.java:999)
        at sun.net.www.protocol.http.HttpURLConnection.connect(HttpURLConnection.java:933)
        at sun.net.www.protocol.http.HttpURLConnection.getInputStream0(HttpURLConnection.java:1513)
        at sun.net.www.protocol.http.HttpURLConnection.getInputStream(HttpURLConnection.java:1441)
        at java.net.HttpURLConnection.getResponseCode(HttpURLConnection.java:480)
        at org.apache.maven.wagon.providers.http.LightweightHttpWagon.fillInputData(LightweightHttpWagon.java:115)
        ... 8 more
```
I may have some problems connecting to maven central but why the resolver didn't look at my local repo which did contain the required dependencies.

HeartSaVioR:
@manuzhang 1.1.1-SNAPSHOT is not officially released (we don't release SNAPSHOT officially) so you need to build entire Storm code with `mvn clean install` before doing that. We can change the example to 1.1.0 or so after the feature is released to official.

manuzhang:
@HeartSaVioR yes, I did so and verified I have all the dependencies locally. I also modified the `DependencyResolverTest` to resolve `1.1.1-SNAPSHOT` and the UT passed.

HeartSaVioR:
@manuzhang OK great. Thanks for the update.

HeartSaVioR:
@manuzhang A bit confusing so would like to double check on this. Do you succeed to resolve 1.1.1-SNAPSHOT after `mvn clean install`? Or you're still failing?

manuzhang:
The exception is caused by the escaping "\" character previously. Things are working now after I updated to the latest version.

bin/storm.py
```diff
@@ -232,44 +271,91 @@ def jar(jarfile, klass, *args):
...
+    --jars and --artifacts options available for jar are also applied to sql command.
+    Please refer "help jar" to see how to use --jars and --artifacts options.
+    You normally want to pass these options since you need to set data source to your sql which is an external storage in many cases.
     """
+    global DEP_JARS_OPTS, DEP_ARTIFACTS_OPTS
+
+    local_jars = DEP_JARS_OPTS
+    artifact_to_file_jars = resolve_dependencies(DEP_ARTIFACTS_OPTS)
+
+    sql_core_jars = get_jars_full(STORM_DIR + "/external/sql/storm-sql-core")
```
harshach: making this hardcoded path to external code doesn't sound right.  we should make sql as top-level directory

bin/storm.py
harshach: whats the intention behind artifacts?

HeartSaVioR: This is already documented for `jar` so you can refer `help jar` to see explanation. I've also documented to doc page.

The intention behind picking `artifacts` is that, I picked `packages` first since this is what Spark already used, but realized why they use `packages` as term is that they maintain spark-packages. So `artifacts` is the alternatives for me.
Btw, the term `artifact` is used by Maven and Ivy, and Aether also uses `Artifact` class to represent this.

bin/storm.py
HeartSaVioR: Definitely. I've also stated that thing to somewhere other pull requests.
`external` has various kinds of modules, and we need to arrange that. 

Maybe we can leave kind of connectors as it is (we can even rename on this), and get flux, sql, kafka-monitor (this module is referred from script in bin, too), and more to out of external.

Since this issue is not only regarding to storm-sql I would like to initiate the discussion before making the change, and handle this to another issue. What do you think?

harshach: @HeartSaVioR sounds good to me. We can address this in another JIRA.

harshach:
+1

HeartSaVioR:
Please hold off merging. I'll create pull request against master. We can merge both of them together.

HeartSaVioR:
OK, I created pull request against master: #1626

HeartSaVioR:
@harshach @satishd @manuzhang 
I just fixed one thing from nimbus.clj: just log exception instead of crashing nimbus when `blob-rm-dependency-jars-in-topology` throws Exception. Before that nimbus test is failing intermittent.

You can just review nimbus.clj once again. Thanks in advance!

HeartSaVioR:
I found a bug while testing other stuff with applied cluster. 
StormSubmitter doesn't handle empty --jars option properly. Will address this shortly both this and PR for master branch.

HeartSaVioR:
Done. Fixed things after receiving +1 are here:

1. nimbus.clj: just log exception instead of crashing nimbus when blob-rm-dependency-jars-in-topology
2. DependencyPropertiesParser.java: check --jars parameter string is empty, and treat it as empty list (also add new unit test testing this change)

HeartSaVioR:
Rebased with current 1.x-branch.

HeartSaVioR:
Rebased with current 1.x-branch again.

HeartSaVioR:
Changed module name as suggestion from pull request on master branch.

HeartSaVioR:
Merged via df8346ce1a61ccad2a0c3f715b6c3a53bdaf5670

PR #1609, opened 2016-08-05
fix parent version of storm-kinesis module (1.x)

HeartSaVioR:
Builds on PR for 1.x-branch are failing because of this.


HeartSaVioR:
Since all builds are failing, and it's not about code itself, I'll just merge this right now. Please revert if you think it should get +1.

PR #1536, opened 2016-07-01
Storm 1890 ensure we refetch static resources after package build

abellina:
Original pull request: https://github.com/apache/storm/pull/1486

I added in this PR the no-cache header for html pages in the / directory. To recap, this is to add cache-busting to Storm UI.

I also upgraded ring-core to 1.3.1 and compojure to 1.1.9 to match ring-core versions without changing too much. Versions above 1.1.9 seem to have issues with Clout not compiling for clojure 1.7. We could potentially go to a much newer version which fixes the issues (1.4.0), but that should be a different PR IMHO, and I am not sure how much we care since this is going away on conversion to Java.

PR #1612, opened 2016-08-05
STORM-1890: looks like too many calls were moved to getStatic

abellina:
Reverting some of the changes I made in: https://github.com/apache/storm/pull/1536

What is on master shouldn't break requests, but it is sloppy, so this should clean up. I didn't create a new JIRA but can if appropriate. Will PR for 1.x and 1.0.x


HeartSaVioR:
+1

PR #1613, opened 2016-08-05
Fix emitting of tuples from bolts with Python 3

ThijsMergaert:
In Python 3, the map() function no longer returns a list, but instead an iterable object, which is not JSON serializable.  To fix this, convert the result of the map() call to a list with the list() function.  This is still compatible with Python 2 (although will create a copy of the list).


HeartSaVioR:
@ThijsMergaert 
Sorry to comment too late.

I'm not sure how much python users use Storm python multi-lang directly, and what's more preferred version, 2 vs 3. If many users still rely on 2, it might be better to avoid creating list and find other workaround. (checking version and if-else?)

PR #1614, opened 2016-08-05
STORM-1890: looks like too many calls were moved to getStatic. It doe…

abellina:
…snt break things, but it is sloppy, so reverting some to $.get


harshach:
+1

HeartSaVioR:
+1

PR #1616, opened 2016-08-08
STORM-2026 Inconsistency between (SpoutExecutor, BoltExecutor) and (spout-transfer-fn, bolt-transfer-fn)

HeartSaVioR:
* fix Executor, SpoutExecutor, BoltExecutor to not calling init() while creating Executor itself
* call init() for the first time in async loop

Please refer https://github.com/apache/storm/pull/1445#discussion_r72933087 and comment thread for this.

### DIFF
--- a/storm-core/src/jvm/org/apache/storm/executor/Executor.java
+++ b/storm-core/src/jvm/org/apache/storm/executor/Executor.java
@@ -204,8 +204,8 @@ public abstract class Executor implements Callable, EventHandler<Object> {
                 throw Utils.wrapInRuntime(ex);
             }
         }
-        executor.init(idToTask);
 
+        executor.idToTask = idToTask;
         return executor;
     }
 
@@ -242,8 +242,6 @@ public abstract class Executor implements Callable, EventHandler<Object> {
     }
 
     public abstract void tupleActionFn(int taskId, TupleImpl tuple) throws Exception;
-
-    public abstract void init(Map<Integer, Task> idToTask);
 
     @SuppressWarnings("unchecked")
     @Override
--- a/storm-core/src/jvm/org/apache/storm/executor/spout/SpoutExecutor.java
+++ b/storm-core/src/jvm/org/apache/storm/executor/spout/SpoutExecutor.java
@@ -79,8 +79,11 @@ public class SpoutExecutor extends Executor {
         this.spoutThrottlingMetrics = new SpoutThrottlingMetrics();
     }
 
-    @Override
     public void init(final Map<Integer, Task> idToTask) {
+        while (!stormActive.get()) {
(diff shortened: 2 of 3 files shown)


storm-core/src/jvm/org/apache/storm/executor/Executor.java
```diff
@@ -204,8 +204,8 @@ public static Executor mkExecutor(Map workerData, List<Long> executorId, Map<Str
                 throw Utils.wrapInRuntime(ex);
             }
         }
-        executor.init(idToTask);
 
+        executor.idToTask = idToTask;
```
satishd: @HeartSaVioR It is better to pass  `idToTask` object as part of constructor instead of setting here. idToTask can be created before calling constructor.

storm-core/src/jvm/org/apache/storm/executor/spout/SpoutExecutor.java
```diff
@@ -79,8 +79,11 @@ public SpoutExecutor(final Map workerData, final List<Long> executorId, Map<Stri
         this.spoutThrottlingMetrics = new SpoutThrottlingMetrics();
     }
 
-    @Override
     public void init(final Map<Integer, Task> idToTask) {
+        while (!stormActive.get()) {
```
satishd: The assumption here is that this method is invoked only once. Do no we want to handle for multiple invocations?

HeartSaVioR: I just follow previous executor implementation, and invoking init() to multiple times is not intended not only for while statement, but also rest of statements.

storm-core/src/jvm/org/apache/storm/executor/Executor.java
HeartSaVioR: @satishd 
Constructor of Task needs executor so Executor should be created prior to construct idToTask.

satishd: @HeartSaVioR Sorry, missed that. This code looks convoluted because it is ported from clojure. I am fine with that for now as these PRs are more about poring from clojure to Java. We can address these later.

satishd:
+1, LGTM

storm-core/src/jvm/org/apache/storm/executor/Executor.java
HeartSaVioR: Yes, actually the code style is more likely to clojure itself indeed. It's needed for ease of port work, and we can address later.

satishd:
Thanks @HeartSaVioR

HeartSaVioR:
Thanks @satishd for reviewing and merging!

PR #1618, opened 2016-08-10
[STORM-2032] removes warning in case more than one metrics tuple is received

d2r:


storm-core/src/jvm/org/apache/storm/messaging/netty/StormClientHandler.java
```diff
@@ -60,7 +60,6 @@ public void messageReceived(ChannelHandlerContext ctx, MessageEvent event) {
                 //This should be the metrics, and there should only be one of them
```
HeartSaVioR: If we want to remove warning message because this is not a part of design, how about modifying this comment as well?
Maybe like this : `This should be the metrics, and it always picks the last one when more than one metrics messages are available.`

storm-core/src/jvm/org/apache/storm/messaging/netty/StormClientHandler.java
```diff
@@ -60,7 +60,6 @@ public void messageReceived(ChannelHandlerContext ctx, MessageEvent event) {
...
                 List<TaskMessage> list = (List<TaskMessage>)message;
                 if (list.size() < 1) throw new RuntimeException("Didn't see enough load metrics ("+client.getDstAddress()+") "+list);
-                if (list.size() != 1) LOG.warn("Messages are not being delivered fast enough, got "+list.size()+" metrics messages at once("+client.getDstAddress()+")");
```
satishd: nit: Can we have a debug statement saying more than one task message received instead of the removed warning message?

satishd:
+1, would like minor nit to be addressed.

storm-core/src/jvm/org/apache/storm/messaging/netty/StormClientHandler.java
HeartSaVioR: I'm also OK for this approach. If we pick this, comment doesn't need to be modified.

d2r:
OK, no problem. Changed it to debug.

HeartSaVioR:
+1 Great finding.

HeartSaVioR:
Note: I'll cherry pick 4f0f79b since it's only effective commit.

PR #1619, opened 2016-08-11
Fix parent version of storm-druid and storm-kinesis

HeartSaVioR:
* during the rollback they still keep versions since it's manually fixed after the release process


knusbaum:
+1

jerrypeng:
+1 just faced this issue.  Glad you already have a fix up

satishd:
+1

ptgoetz:
+1 nice catch. Those components were added after the last RC, which is why they weren't included in the revert.

PR #1398, opened 2016-05-04
[STORM-1766] - A better algorithm server rack selection for RAS

jerrypeng:

PR #1621, opened 2016-08-11
Backport to 1.x [STORM-1766] - A better algorithm server rack selection for RAS

jerrypeng:
Backport of #1398 to 1.x branch.  I'm not sure this actually needs a PR, but since it's been a while since #1398 was merged, I'll put one anyways since the code went into 2.x a while ago

### DIFF
--- a/storm-core/src/jvm/org/apache/storm/scheduler/Cluster.java
+++ b/storm-core/src/jvm/org/apache/storm/scheduler/Cluster.java
@@ -103,6 +103,9 @@ public class Cluster {
         this.status.putAll(src.status);
         this.topologyResources.putAll(src.topologyResources);
         this.blackListedHosts.addAll(src.blackListedHosts);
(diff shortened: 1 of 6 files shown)


knusbaum:
+1

storm-core/src/jvm/org/apache/storm/scheduler/Cluster.java
```diff
@@ -103,6 +103,9 @@ public Cluster(Cluster src) {
         this.status.putAll(src.status);
         this.topologyResources.putAll(src.topologyResources);
         this.blackListedHosts.addAll(src.blackListedHosts);
+        if (src.networkTopography != null) {
```
harshach: is this supposed to be == null. why are we creating new Map if there is one already.

HeartSaVioR:
+1

PR #1622, opened 2016-08-11
STORM-1999 Update docs with OutputCollector's threadsafety

HeartSaVioR:
* From Storm 1.0.0, OutputCollector is thread safe

This should be also downmerged to 1.x and 1.0.x branches, and applied to docs of 1.0.0, 1.0.1, 1.0.2, 2.0.0-SNAPSHOT.

NOTE: This is non-code change so lazy consensus is applied. (don't need to wait for +1 and also 24 hrs)


satishd:
+1

PR #1623, opened 2016-08-11
STORM-841 Thread-safeness of OutputCollector has documented contrary to two official doc (0.10.x)

HeartSaVioR:
* Fix for 0.10.x

This should be also downmerged to 0.9.x branch, and applied to docs of 0.10.0, 0.10.1, 0.9.6.

NOTE: This is non-code change so lazy consensus is applied. (don't need to wait for +1 and also 24 hrs)


satishd:
+1

PR #1624, opened 2016-08-12
[STORM-2037] debug operation should be whitelisted in SimpleAclAuthorizer

arunmahadevan:
While enabling/disabling topology event logging, theres a call to assert-authorized-user with "debug" as the operation in ui/core.clj.
This needs to be whitelisted in SimpleAclAuthorizer to get it working in secure mode.


harshach:
+1. We should be introducing a Enum class to capture these ACLs in one place and let the pluggable authorizers use that class instead of free-form strings.

harshach:

PR #1627, opened 2016-08-15
STORM-2039: backpressure refactoring in worker and executor

abellina:
For 1.x branch, will submit a PR for 2.x branch.

@zhuoliu worked on the initial refactor. I added some due code review comments.

### DIFF
--- a/storm-core/src/clj/org/apache/storm/daemon/worker.clj
+++ b/storm-core/src/clj/org/apache/storm/daemon/worker.clj
@@ -160,10 +160,10 @@
   check highWaterMark and lowWaterMark for backpressure"
   (disruptor/disruptor-backpressure-handler
     (fn []
-      (reset! (:transfer-backpressure worker) true)
+      (log-debug "worker " (:worker-id worker) " transfer-queue is congested, set backpressure flag true")
       (WorkerBackpressureThread/notifyBackpressureChecker (:backpressure-trigger worker)))
     (fn []
-      (reset! (:transfer-backpressure worker) false)
+      (log-debug "worker " (:worker-id worker) " transfer-queue is not congested, set backpressure flag false")
       (WorkerBackpressureThread/notifyBackpressureChecker (:backpressure-trigger worker)))))
 
 (defn mk-transfer-fn [worker]
@@ -316,7 +316,6 @@
       :load-mapping (LoadMapping.)
       :assignment-versions assignment-versions
       :backpressure (atom false) ;; whether this worker is going slow
-      :transfer-backpressure (atom false) ;; if the transfer queue is backed-up
       :backpressure-trigger (atom false) ;; a trigger for synchronization with executors
       :throttle-on (atom false) ;; whether throttle is activated for spouts
       )))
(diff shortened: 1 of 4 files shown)


storm-core/src/clj/org/apache/storm/daemon/worker.clj
```diff
@@ -743,7 +736,7 @@
                         (fn [& args]
                           (check-credentials-changed)
                           (if ((:storm-conf worker) TOPOLOGY-BACKPRESSURE-ENABLE)
-                            (check-throttle-changed))))
+                            (topology-backpressure-callback))))
```
knusbaum: This feels like the wrong place for this. 
For some reason the backpressure is tacked onto the credentials timer?

I know this was already the way it worked, but maybe there's a better place for it now? I don't think it's good that the value of the `task.credentials.poll.secs` config affects the frequency of backpressure polling.

storm-core/src/clj/org/apache/storm/daemon/worker.clj
```diff
@@ -316,7 +316,6 @@
       :load-mapping (LoadMapping.)
       :assignment-versions assignment-versions
       :backpressure (atom false) ;; whether this worker is going slow
-      :transfer-backpressure (atom false) ;; if the transfer queue is backed-up
       :backpressure-trigger (atom false) ;; a trigger for synchronization with executors
```
knusbaum: Let's just make this an Object. The fact that it's an atom is misleading.

knusbaum:
+1 for the changes. My comments are suggestions for a couple more.
The travis failures look unrelated. Storm Core passes.

storm-core/src/clj/org/apache/storm/daemon/worker.clj
abellina: yeah, agreed. I'll add another timer for this.

storm-core/src/clj/org/apache/storm/daemon/worker.clj
abellina: Nice catch. Will do.

abellina:
@knusbaum I made the changes suggested. Also removed an extra fn wrapper on that function triggered by the timer.

knusbaum:
+1

zhuoliu:
@abellina looks good to me, just one extra suggestion:
In
https://github.com/apache/storm/blob/master/storm-core/src/clj/org/apache/storm/daemon/worker.clj#L156
To reduce the traffic to ZK, we checked the previous flag to update to Zookeeper only when it is necessary (the flag has changed), but still there may some weird case we have inconsistency between local "prev-backpressure-flag" and the worker backpressure node at Zookeeper (see STORM-1949).

So to strike a tradeoff and avoid the possible wrong stall of a topology, could you add do a random number generation here, e.g., generate a number between 0-9, if it is zero, we will call the try block no matter "prev-backpressure-flag" and "curr-backpressure-flag" is equal or not.

Thanks a lot!

abellina:
@zhuoliu thanks for the comment. I am taking a look at STORM-1949 to try and reproduce it, and will comment there. I think that is a separate issue from this PR so we should see a potential fix to STORM-1949 in a different PR. Are you ok with this going in as is?

zhuoliu:
sure, Alex. Really appreciate if you can submit a separate PR for 1949.
+1 to this one!

zhuoliu:
Since with the refactoring we directly access the disruptor queue to get the backpressure status rather than having an additional flag in executor data which is error-prone when flipping (which is very possibly the root cause of the halt in STORM-1949), we may NOT need to add the ADDITIONAL ZK pressure which I mentioned in the above.

HeartSaVioR:
+1

PR #1628, opened 2016-08-16
STORM-2041. Make Java 8 as minimum requirement for 2.0 release.

harshach:


HeartSaVioR:
Please modify .travis.yml as well so that build doesn't occur on Oracle JDK 7.

harshach:
@HeartSaVioR done.

HeartSaVioR:
+1 for the change but would like to hold this until we finish a formal process.

For me changing minimum requirement is an important decision for the project so handling this with only general consensus seems not enough.
I don't remember how we change minimum requirement as Java 7, but ideally we need to VOTE to confirm/record the decision of the PMC. Some PMC members already gave positive feedbacks so it tends to be passed, but it would be better to hear more voices from other PMC members, especially representing OK (+1) or NO (-1).

harshach:
@HeartSaVioR whats the formal process? did we ever define that. I already started discussion thread.  There are 3 pmc's who said +1 and so far there is no -1. Its past 72 hrs as well.

HeartSaVioR:
@harshach 
Yes, bylaws states that.

http://storm.apache.org/contribute/BYLAWS.html
```
Voting
Decisions regarding the project are made by votes on the primary project development mailing list (dev@storm.apache.org). 
```

IMO all the decision without non-formal vote is just a general consensus. We were implicitly allowing general consensus for relatively not important things (like infra), but changing Java requirement is not the case for me.

kishorvpatil:
+1

satishd:
+1

harshach:
@HeartSaVioR started a thread or we can use the PR votes for consensus among PMC.

HeartSaVioR:
@harshach Thanks for the quick follow up. I just voted to VOTE thread. Since we've seen some PMCs' positive feedback I guess it should be reflected to vote result.

arunmahadevan:
+1

revans2:
@HeartSaVioR Technically I don't think the bylaws cover changing minimum requirements beyond a code change.  If you want to modify the bylaws in that way I would be happy to support you in this. 

+1

HeartSaVioR:
I will just do +1 and go merging since we already see our consensus from multiple places.
I don't want to become a lawyer :) so that's enough.

PR #1629, opened 2016-08-17
Fix code in KafkaSpoutRetryExponentialBackoff.java

leedohyun:
Additionally, Is there some reasons for use as below in README.md.
`
new KafkaSpoutRetryExponentialBackoff.TimeInterval(500, TimeUnit.MICROSECONDS)
`

I think this is better.
`
KafkaSpoutRetryExponentialBackoff.TimeInterval.microSeconds(500)
`


HeartSaVioR:
@leedohyun 
+1 for the change, and yes I also think it would be better to change like you stated since other parameters already use that method.
Could you also change that thing? Thanks in advance.

leedohyun:
@HeartSaVioR 
Thank you for your review!!

HeartSaVioR:
+1 Thanks for the work @leedohyun

HeartSaVioR:
Merged into master, 1.x, 1.0.x branches respectively.

PR #1635, opened 2016-08-19
STORM-1434 Support the GROUP BY clause in StormSQL

HeartSaVioR:
* Support GROUP BY for Trident
* Implement basic functions for aggregation
* Change the way of converting Calcite logical plan to Trident logical plan
  * before: creating codes and compile them
  * after: use Trident features, only creating code block if evaluation is needed
    * Janino comes in to help evaluating code block in runtime

This change can handle the below sql statements in runtime:

```
CREATE EXTERNAL TABLE ORDERS (ID INT PRIMARY KEY, UNIT_PRICE INT, QUANTITY INT) LOCATION 'kafka://localhost:2181/brokers?topic=orders' TBLPROPERTIES '{"producer":{"bootstrap.servers":"localhost:9092","acks":"1","key.serializer":"org.apache.storm.kafka.IntSerializer","value.serializer":"org.apache.storm.kafka.ByteBufferSerializer"}}'
CREATE EXTERNAL TABLE LARGE_ORDERS (ID INT PRIMARY KEY, TOTAL INT) LOCATION 'kafka://localhost:2181/brokers?topic=large_orders' TBLPROPERTIES '{"producer":{"bootstrap.servers":"localhost:9092","acks":"1","key.serializer":"org.apache.storm.kafka.IntSerializer","value.serializer":"org.apache.storm.kafka.ByteBufferSerializer"}}'
INSERT INTO LARGE_ORDERS SELECT ID, UNIT_PRICE * QUANTITY AS TOTAL FROM ORDERS WHERE UNIT_PRICE * QUANTITY > 50
```

```
CREATE EXTERNAL TABLE ORDERS (ID INT PRIMARY KEY, UNIT_PRICE INT, QUANTITY INT) LOCATION 'kafka://localhost:2181/brokers?topic=orders' TBLPROPERTIES '{"producer":{"bootstrap.servers":"localhost:9092","acks":"1","key.serializer":"org.apache.storm.kafka.IntSerializer","value.serializer":"org.apache.storm.kafka.ByteBufferSerializer"}}'
CREATE EXTERNAL TABLE SUMMARY_ORDERS (ID INT PRIMARY KEY, UNIT_PRICE_AVG DOUBLE, SUM_OF_QUANTITY INT) LOCATION 'kafka://localhost:2181/brokers?topic=large_orders' TBLPROPERTIES '{"producer":{"bootstrap.servers":"localhost:9092","acks":"1","key.serializer":"org.apache.storm.kafka.IntSerializer","value.serializer":"org.apache.storm.kafka.ByteBufferSerializer"}}'
INSERT INTO SUMMARY_ORDERS SELECT ID, AVG(UNIT_PRICE) AS UNIT_PRICE_AVG, SUM(QUANTITY) AS SUM_OF_QUANTITY FROM ORDERS GROUP BY ID
```

This can be easily backported to 1.x branch since it only touches storm-sql module.

When reviewing, please share your idea if you find something to optimize further.
I didn't play with Trident hardly so I'm not sure I'm following the best practice.

I didn't touch UDF yet since coverage of STORM-1434 would be too broad if STORM-1434 contains UDF. Will file an issue to track and plan to address that.

Also STORM-2016 should be addressed soon in order to run the topology.

TODO

- [x] documentation update
- [x] fix AVG to not changing origin type


external/sql/storm-sql-core/src/jvm/org/apache/storm/sql/compiler/ExprCompiler.java
```diff
@@ -67,8 +68,9 @@ public ExprCompiler(PrintWriter pw, JavaTypeFactory typeFactory) {
   public String visitInputRef(RexInputRef rexInputRef) {
     String name = reserveName();
     String typeName = javaTypeName(rexInputRef);
+    String boxedTypeName = boxedJavaTypeName(rexInputRef);
     pw.print(String.format("%s %s = (%s)(_data.get(%d));\n", typeName, name,
-                           typeName, rexInputRef.getIndex()));
+                           boxedTypeName, rexInputRef.getIndex()));
```
HeartSaVioR: Converting Object to primitive type is allowed for Java 7 or upper, but Janino throws error while compiling script so I just converted to boxed type.

HeartSaVioR:
Found a bug: AVG() changes column type to double whatever column type is, which shouldn't.
Will add [WIP] to title and fixing this.

HeartSaVioR:
storm-sql relies on clojure.lang.Numbers class which will not be available. Should fix that too.

HeartSaVioR:
I just removed `replace clojure.lang.Numbers to other implementations` from TODO because I can't find the easy way to do that. 
I'd like to address this as separate issue, and any ideas or suggestions are more than welcomed.

HeartSaVioR:
Fixed handling return type of AVG(<field>). 
If the type of field is a fixed number, AVG() does floor division and returns quotient, and if the type of field is a floating number, AVG() does division and returns floating point.

HeartSaVioR:
Ready to review now.

HeartSaVioR:
@harshach @haohui @arunmahadevan 
Could you take a look?

external/sql/storm-sql-core/src/jvm/org/apache/storm/sql/StormSqlImpl.java
```diff
@@ -137,18 +137,12 @@ public void submit(
     }
   }
 
-  private void packageTopology(Path jar, CompilingClassLoader cl, AbstractTridentProcessor processor) throws IOException {
+  private void packageEmptyTopology(Path jar) throws IOException {
```
arunmahadevan: Can you put some comments on why are we packaging an empty jar and where the code for executing the topology is going to be ?

HeartSaVioR: Yes that would be better. I'll address this.

external/sql/storm-sql-core/src/jvm/org/apache/storm/sql/compiler/backends/trident/TridentPostOrderRelNodeVisitor.java
```diff
@@ -0,0 +1,155 @@
...
+import org.apache.calcite.rel.core.Window;
+import org.apache.calcite.rel.stream.Delta;
+import org.apache.storm.trident.TridentTopology;
+import org.apache.storm.trident.fluent.IAggregatableStream;
+import org.apache.storm.trident.topology.TridentTopologyBuilder;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public abstract class TridentPostOrderRelNodeVisitor {
```
arunmahadevan: There is an abstract class PostOrderRelNodeVisitor which implements the same logic. Any reason for duplicating it here? Maybe enhance PostOrderRelNodeVisitor to accept input streams (generic type) and have TridentPostOrderRelNodeVisitor extend it?

HeartSaVioR: I just wanted to avoid any changes on standalone mode since we're constructing topology (or components) differently for two modes now.
And also I'm curious that Trident stream building pattern is common for others.

What do you think if we change PostOrderRelNodeVisitor to below?

```
public abstract class PostOrderRelNodeVisitor<T> {
  public final T traverse(RelNode n) throws Exception {
    List<T> inputStreams = new ArrayList<>();
    for (RelNode input : n.getInputs()) {
      inputStreams.add(traverse(input));
    }

    if (n instanceof Aggregate) {
      return visitAggregate((Aggregate) n, inputStreams);
    ...
  }

  public T visitAggregate(Aggregate aggregate, List<T> inputStreams) throws Exception {
    return defaultValue(aggregate, inputStreams);
  }
...
```

Is it OK for you to apply this for standalone mode?

external/sql/storm-sql-core/src/jvm/org/apache/storm/sql/compiler/backends/trident/TridentLogicalPlanCompiler.java
```diff
@@ -0,0 +1,344 @@
...
+            Fields outputFields = aggregateStream.getOutputFields();
+            // assuming handleAggregateCall does project so that new output field is always placed to end of fields
+            String newFieldName = outputFields.get(outputFields.size() - 1);
+
+            joiningStreams.add(aggregateStream);
+            joiningFields.add(groupByFields);
+            newOutputFields.add(newFieldName);
+        }
+
+        return topology.join(joiningStreams, joiningFields, new Fields(newOutputFields), JoinType.INNER).name(stageName);
```
arunmahadevan: here we create a separate stream per aggregate function and thats why the join here ? May be we can just execute the different aggregate functions in a single Function and emit the results to avoid the separate streams and join. It may produce optimized topology, but I am not sure.

external/sql/storm-sql-runtime/src/jvm/org/apache/storm/sql/runtime/trident/operations/CountBy.java
```diff
@@ -0,0 +1,52 @@
...
+    private final String inputFieldName;
+
+    public CountBy(String inputFieldName) {
+        this.inputFieldName = inputFieldName;
+    }
+
+    @Override
+    public Long init(TridentTuple tuple) {
+        if (TridentUtils.valueFromTuple(tuple, inputFieldName) == null) {
+            return 0L;
```
arunmahadevan: Is it supposed to count only the tuples with non-null values for inputFieldName ? If so would it return the right count since combine cant check this?

external/sql/storm-sql-core/src/jvm/org/apache/storm/sql/compiler/backends/trident/TridentLogicalPlanCompiler.java
HeartSaVioR: http://storm.apache.org/releases/1.0.1/Trident-API-Overview.html

> The benefits of CombinerAggregators are seen when you use them with the aggregate method instead of partitionAggregate. In that case, Trident automatically optimizes the computation by doing partial aggregations before transferring tuples over the network.

While reading same tuples for multiple time feels redundant, handling all aggregations in one base function will drop the chance for optimization like this.

I'd like to try out Trident optimizations for all of cases, and address optimization for Trident itself when needed.

Btw, I also would like to replace this to chainedAgg(). I'll address this.

external/sql/storm-sql-runtime/src/jvm/org/apache/storm/sql/runtime/trident/operations/CountBy.java
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        return 1L;
+    }
+
+    @Override
+    public Long combine(Long val1, Long val2) {
+        return val1 + val2;
```
arunmahadevan: Should it return `val1 + 1` ?

external/sql/storm-sql-runtime/src/jvm/org/apache/storm/sql/runtime/trident/operations/CountBy.java
HeartSaVioR: > A CombinerAggregator returns a single tuple with a single field as output. CombinerAggregators run the init function on each input tuple and use the combine function to combine values until there's only one value left. If there's no tuples in the partition, the CombinerAggregator emits the output of the zero function.

Based on the doc, the return of init() represents the value for the tuple. It's like `map` before calling `reduce`.

external/sql/storm-sql-runtime/src/jvm/org/apache/storm/sql/runtime/trident/operations/CountBy.java
HeartSaVioR: According to the doc, val2 should be the return value of init(). This is why it can only count non-null fields with init() implementation.

arunmahadevan:
@HeartSaVioR I have gone through the changes and left a few comments. Like the idea of using Janino for compiling which results in cleaner code.

external/sql/storm-sql-core/src/jvm/org/apache/storm/sql/compiler/backends/trident/TridentPostOrderRelNodeVisitor.java
arunmahadevan: yes, something along this lines looks good.

external/sql/storm-sql-runtime/src/jvm/org/apache/storm/sql/runtime/trident/operations/CountBy.java
arunmahadevan: OK its clear now.

HeartSaVioR:
@arunmahadevan 
Thanks for the review. I addressed your comments except handling aggregate functions. Will push and comment once I'm done with that.

HeartSaVioR:
@arunmahadevan 
Also addressed handling aggregate functions. Please take a look again when you have some time. Thanks!

HeartSaVioR:
> SELECT GRPID, COUNT(*) AS CNT, MAX(AGE) AS MAX_AGE, MIN(AGE) AS MIN_AGE, AVG(AGE) AS AVG_AGE, MAX(AGE) - MIN(AGE) AS DIFF FROM FOO GROUP BY GRPID	

Below link is a Trident logical/physical plan for above query.
https://docs.google.com/spreadsheets/d/1tTyIxbNafvVw0tQjWjiBQCKDCPbbnfbT5JZ7F5JEBNE/edit?usp=sharing

arunmahadevan:
@HeartSaVioR I have gone through your recent changes and it looks good.

In the logical plan you have posted, there are multiple processors just for the project (before the aggregation stage). Shouldn't there be just one processor to project all the required fields ? Ideally something like `Project -> aggregate -> Project`. Maybe I am missing something, anyways its more of an optimization which you can take up  later.

Overall I am +1

HeartSaVioR:
@arunmahadevan Yes I've seen some places to optimize. Trident groups nodes so there're many nodes but there will be only three components - one spout and two bolts. So for that situation Trident itself might not need to be optimized further, but there seems to be a room for optimizing logical plan.

HeartSaVioR:
Optimizing logical / physical plan is the long term works so I would like to revisit again.
(Will file an issue for analyzing logical plan for several queries. Might file additional - maybe umbrella issues for the optimizations.)
I'll go on merging.

PR #1640, opened 2016-08-21
Fix version command in storm.cmd

zxtwonder:
Move version command from shellcommands to corecommands. VersionInfo.main() needs to be executed to print out the version information.


HeartSaVioR:
+1

PR #1646, opened 2016-08-24
STORM-1870 Allow FluxShellBolt/Spout set custom "componentConfig" via yaml

HeartSaVioR:
* add addComponentConfig methods to FluxShellSpout and FluxShellBolt
* address unit tests

This also addresses Parsely/streamparse#263 and Parsely/streamparse#266 

cc. @Darkless012

This can be easily backported to 1.x-branch, which means it can be included to upcoming 1.1.0 release.


HeartSaVioR:
@ptgoetz Could you have a look?

ptgoetz:
+1

PR #1647, opened 2016-08-24
Get rid of compiler plugin setting for external modules

HeartSaVioR:
That setting makes some modules to apply language level for lower version of JDK.

It can be applied to 1.x-branch as well since flux refers JDK 1.6.


revans2:
+1

revans2:
@HeartSaVioR Thanks I pulled this into master, 1.x and 1.0.x

PR #1648, opened 2016-08-24
STORM-2054 DependencyResolver should be aware of relative path and absolute path

HeartSaVioR:
* Fix Booter to handle the path with awareness of relative vs absolute
* Modify DependencyResolverTest to remove temporary directory after tests for that class are done

This breaks whole builds for master and 1.x branch due to RAT failing.

It should be also applied to 1.x-branch.

### DIFF
--- a/external/storm-submit-tools/src/main/java/org/apache/storm/submit/dependency/Booter.java
+++ b/external/storm-submit-tools/src/main/java/org/apache/storm/submit/dependency/Booter.java
@@ -1,12 +1,13 @@
-/*
- * Licensed to the Apache Software Foundation (ASF) under one or more
- * contributor license agreements.  See the NOTICE file distributed with
- * this work for additional information regarding copyright ownership.
- * The ASF licenses this file to You under the Apache License, Version 2.0
- * (the "License"); you may not use this file except in compliance with
- * the License.  You may obtain a copy of the License at
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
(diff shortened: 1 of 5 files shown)


HeartSaVioR:
@revans2 This is also a fix for STORM-2059. please review this.

external/storm-submit-tools/src/main/java/org/apache/storm/submit/dependency/Booter.java
```diff
@@ -37,16 +37,19 @@ public static RepositorySystemSession newRepositorySystemSession(
             RepositorySystem system, String localRepoPath) {
         MavenRepositorySystemSession session = new MavenRepositorySystemSession();
 
-        // find homedir
-        String home = System.getProperty("storm.home");
-        if (home == null) {
-            home = ".";
-        }
+        File repoDir = new File(localRepoPath);
```
manuzhang: Why not leave the handling of path to the caller of `Booter` ?

HeartSaVioR: This code is from Zeppelin and I just fixed the path handling bug, but I think your suggestion would be better. I'll reflect it.

HeartSaVioR: Addressed.

HeartSaVioR:
Also fixed license form as @manuzhang found.

revans2:
+1 fixes the rat issues on storm-submit-tools.

PR #1649, opened 2016-08-24
STORM-2052: Kafka Spout - New Client API - Performance Improvements

hmcl:
This patch should be back-ported to 1.0.x branch and 1.x-branch. Thank you.

### DIFF
--- a/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
+++ b/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
@@ -123,7 +123,7 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
     private class KafkaSpoutConsumerRebalanceListener implements ConsumerRebalanceListener {
         @Override
         public void onPartitionsRevoked(Collection<TopicPartition> partitions) {
-            LOG.debug("Partitions revoked. [consumer-group={}, consumer={}, topic-partitions={}]",
+            LOG.info("Partitions revoked. [consumer-group={}, consumer={}, topic-partitions={}]",
                     kafkaSpoutConfig.getConsumerGroupId(), kafkaConsumer, partitions);
             if (!consumerAutoCommitMode && initialized) {
                 initialized = false;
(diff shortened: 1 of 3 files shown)


external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -221,7 +224,22 @@ private boolean commit() {
     }
 
     private boolean poll() {
-        return !waitingToEmit() && numUncommittedOffsets < kafkaSpoutConfig.getMaxUncommittedOffsets();
+        final int maxUncommittedOffsets = kafkaSpoutConfig.getMaxUncommittedOffsets();
+        final boolean poll = !waitingToEmit() && numUncommittedOffsets < maxUncommittedOffsets;
+
+        if (!poll) {
+            final boolean isLoggable = logTimer.isExpiredResetOnTrue();
```
harshach: do we need this check since these are debug logs.

harshach:
overall looks good. Can you  squash the commits.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
hmcl: This check is here in order to throttle the number of messages that get printed. Although logs are with debug level, a lot of messages will get printed (one each time nextTuple is called), which will spam the logs, and limit the ability to debug. What this code does is to print one message per second, rather than one per call to nextTuple. By doing this, no info is lost, and things are much cleaner.

hmcl:
@harshach I can squash the commits, but I left these three commits on purpose. Each commit is a logical groping. One is docs, one is the default values, and one is the logging changes. They all have the same JIRA ticket, but the commit log messages suit the changes. The reason behind this choice is that it is easier to cherry pick only part of the changes to a given branch - if needed. I also believe that it makes it easier for an external person to follow what each patch addresses.

I am OK with either approach. This squash is pretty simple. Let me know if you prefer that this is indeed squashed, and I will do so. Thanks!

harshach:
@hmcl all the commit title as the same text "Kafka Spout - New Client API - Performance Improvements"

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
harshach: Given that this is debug it won't print in normal op. If I turn on the debug logging my expectations is to see the logs immediately. If I don't see the messages than I will think that the debug log . 
Either way I am ok with it.

harshach:
can you also fix the JIRA title . Given that these are default config changes and not really performance fixes in the code.

harshach:
@hmcl we cherry-pick the entire JIRA into other branches. It doesn't make sense to cherry-pick partial commits of a JIRA for releases. So squash the commits into one as they are all related to this JIRA.

hmcl:
@harshach Done!

harshach:
+1

harshach:

PR #1652, opened 2016-08-29
Storm 2063 thread name in worker log improvements

abellina:
@d2r did the original work for this.


HeartSaVioR:
+1
Also nice catch for storm-submit-tools!

PR #1653, opened 2016-08-29
STORM-2063: turn on thread name in worker.xml

abellina:
Bring the change in:
https://github.com/apache/storm/pull/1652

To 2.x. Only the worker.xml, since we were already checking for empty thread names in 2.x


HeartSaVioR:
+1

PR #1655, opened 2016-08-30
STORM-2067 Fix "array element type mismatch" from compute-executors in nimbus.clj

HeartSaVioR:
* force set type to Map when calling into-array with seq which has Maps
  * Clojure optimization occurs type mismatch: PersistentArrayMap vs PersistentHashMap

Please refer http://issues.apache.org/jira/browse/STORM-2067 for more details.

This code only resides on master branch. No need to backport.


HeartSaVioR:
+1 from STORM-2057 #1656 (by @arunmahadevan )

PR #1656, opened 2016-08-30
STORM-2057 Support JOIN statement in Storm SQL

HeartSaVioR:
> * Note for merger
>
>This is on top of STORM-2067 which is not applicable for 1.x-branch. So in order to make porting back easier, STORM-2067 should be handled first and this pull request should be rebased to updated master.

This pull request accomplishes supporting JOIN statement in Storm SQL for Trident mode.
INNER, LEFT OUTER, RIGHT OUTER, FULL OUTER JOIN are supported.

Since this feature is based on Trident join, some limitations follow:

- Only equi-join is supported.
- Joining will be occurred for each batch, not across batches.
  - http://storm.apache.org/releases/1.0.2/Trident-API-Overview.html
  - Please refer `Merges & Joins` section
- Trident join provides only one value for joining fields (two)
  - For inner join, that would be OK
  - For outer joins, one of values should be null, which might be not OK
    - applied dirty workaround: if whole values of non-joining fields are null, value of joining field is also null
    - will try to handle it properly (TODO)

TODO
- [x] Update documentation (with limitation)
- [x] Try to fork Trident join feature to fit to SQL join column output

### DIFF
--- a/storm-core/src/jvm/org/apache/storm/trident/TridentTopology.java
+++ b/storm-core/src/jvm/org/apache/storm/trident/TridentTopology.java
@@ -27,6 +27,7 @@ import org.apache.storm.grouping.CustomStreamGrouping;
 import org.apache.storm.topology.BoltDeclarer;
 import org.apache.storm.topology.IRichSpout;
 import org.apache.storm.topology.SpoutDeclarer;
+import org.apache.storm.trident.operation.impl.PreservingFieldsOrderJoinerMultiReducer;
 import org.apache.storm.tuple.Fields;
 import org.apache.storm.utils.Utils;
 
@@ -46,7 +47,6 @@ import org.apache.storm.trident.graph.GraphGrouper;
 import org.apache.storm.trident.graph.Group;
 import org.apache.storm.trident.operation.DefaultResourceDeclarer;
 import org.apache.storm.trident.operation.GroupedMultiReducer;
-import org.apache.storm.trident.operation.ITridentResource;
 import org.apache.storm.trident.operation.MultiReducer;
 import org.apache.storm.trident.operation.impl.FilterExecutor;
 import org.apache.storm.trident.operation.impl.GroupedMultiReducerExecutor;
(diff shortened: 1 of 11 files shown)


HeartSaVioR:
Fixed classification issue for outer join.
It still needs to apply JoinResultFromTridentToSQLFunction to rearrange and project to get rid of old fields. 
If SeparatedJoiningValuesJoinerMultiReducer would emit tuple with preserved fields order we can get rid of JoinResultFromTridentToSQLFunction and additional project. (Name might need to be changed.)
I'll try to work of it.

HeartSaVioR:
I just pushed final version of JoinerMultiReducer for Storm SQL which preserves fields order for each stream, which will be directly matched to Calcite output fields.

I'll remove WIP once I finish updating document.

HeartSaVioR:
Just addressed documentation. Ready for the review.

storm-core/src/jvm/org/apache/storm/trident/operation/impl/JoinerMultiReducer.java
```diff
@@ -139,21 +138,5 @@ private boolean increment(List[] lengths, int[] indices, int j, int overrideInde
...
-            sides = new List[numSides];
-            indices = new int[numSides];
-            this.group = group;
-            for(int i=0; i<numSides; i++) {
-                sides[i] = new ArrayList<List>();
-            }            
-        }
-    }
-    
+
```
HeartSaVioR: Moved JoinState to its own Java file since two classes refer JoinState now.
It's a kind of public class, but only JoinerMultiReducer uses this and it resides on `impl` directory so I think it's OK to move.

HeartSaVioR:
Rebased only STORM_2057 related commits into one.

storm-core/src/jvm/org/apache/storm/trident/operation/impl/PreservingFieldsOrderJoinerMultiReducer.java
```diff
@@ -0,0 +1,175 @@
...
+            }
+            return ret;
+        }
+    }
+
+
+    //return false if can't increment anymore
+    //TODO: DRY this code up with what's in ChainedAggregatorImpl
+    private boolean increment(List[] lengths, int[] indices, int j, int overrideIndex) {
+        if(j==-1) return false;
```
harshach: space after if

storm-core/src/jvm/org/apache/storm/trident/operation/impl/PreservingFieldsOrderJoinerMultiReducer.java
```diff
@@ -0,0 +1,175 @@
...
+        if(j==overrideIndex) {
```
harshach: same here

storm-core/src/jvm/org/apache/storm/trident/operation/impl/PreservingFieldsOrderJoinerMultiReducer.java
```diff
@@ -0,0 +1,175 @@
...
+            return increment(lengths, indices, j-1, overrideIndex);
+        }
+        indices[j]++;
+        if(indices[j] >= lengths[j].size()) {
```
harshach: there are few other places. Lets keep this consistent.

ptgoetz:
Before we get too out of control over white space, we should think about establishing a style guide.

I don't see any style violations here that we haven't seen before (and accepted).

storm-core/src/jvm/org/apache/storm/trident/TridentTopology.java
```diff
@@ -277,6 +277,35 @@ public Stream join(List<Stream> streams, List<Fields> joinFields, Fields outFiel
               outFields);
     }
 
+    public Stream joinSQL(Stream s1, Fields joinFields1, Stream s2, Fields joinFields2, Fields outFields) {
```
arunmahadevan: Why are we exposing new apis in TridentTopology and what extra functionality does joinSQL provide ? It may confuse users if they see both Join and JoinSQL. IMO, we should have only `join` in the trident apis. The SQL specific parts should be private and may be defined outside of TridentTopology if its specific to SQL implementation.

HeartSaVioR: @arunmahadevan 
This provides `join` with preserving parameters. I agree that `joinSQL` is bad. I just haven't found any good naming, since `joinWithPreservingParameters` is bad, too. ;) I would like to have good alternative name, maybe `join` with `mode` parameter would be also fine for me.

Btw, implementing this to outside of storm-core is not easy since it relies on existing implementations which is not open to public. Lots of duplications or exposing unneeded things will be required if we want to move this to SQL runtime.

HeartSaVioR:
@harshach 
Style violations(?) are originally from JoinerMultiReducer, and PreservingFieldsOrderJoinerMultiReducer just copies the code from there. I like your style guide, indeed.

@ptgoetz 
I agree that we need style guide and/or code convention for Java, since we're moving to Java codebase. I'd also love to see automatic formatting, like `mvn java-formatter:format`.

https://github.com/xetorthio/jedis/blob/master/hbase-formatter.xml
https://github.com/xetorthio/jedis/blob/master/Makefile#L301

storm-core/src/jvm/org/apache/storm/trident/TridentTopology.java
arunmahadevan: May be you can have `join` with extra mode parameter as you mentioned. Also good to have some javadocs to explain the difference. I don't get what preserving parameters means without going through the code.

HeartSaVioR: Current `join` emits the tuples reordering via list of join fields -> list of non-joined fields from all streams.
http://storm.apache.org/releases/1.0.2/Trident-API-Overview.html

Preserving parameters mean the tuples emitted from the join preserve the fields order. It is needed for creating result of SQL join. This is in fact more than reordering fields, meaning separating joined fields from streams.

Let's assume that stream SA has 'a', 'b', 'c', and stream SB has 'c', 'd', 'e'. If we join two streams with SA.c = SB.c with left outer join, 'd' and 'e' will be null if there's no matched SB row for SA row R1. In this case SB.c should be null, but we can't determine this from current implementation of `join`, because current `join` only returns SA.c as joined fields.
We can't even determine SB is not matched via checking non-joined fields as null, since 'd', 'e' could be null even though there's matched SB row.

harshach:
@ptgoetz we should have a style guide or at least we should've consistency. my comments are because I've seen multiple combinations where some code as space and others don't.

HeartSaVioR:
@arunmahadevan @harshach 
I just addressed your review comments. Please take a look again.

Adding Javadoc for Trident is not easy task since there's no method which has javadoc on Trident. We might need to add javadoc for all of them, but I feel this is out of scope.

arunmahadevan:
+1

HeartSaVioR:
Thanks @arunmahadevan for reviewing. Since I've also addressed style review comments I'll just go on merging this.

PR #1660, opened 2016-08-30
STORM-2070 Fix sigar native binary download link

HeartSaVioR:
* previous link goes 404, googlecode seems change archive url

This should be also merged back to 1.x, 1.0.x branches.


revans2:
+1

hackerwin7:
the fix is worked to build

PR #1662, opened 2016-08-31
[STORM-1344]  Remove sql command from storm-jdbc build.

ppoulosk:
Works around STORM-1344 by ensuring table is not double-created.


abellina:
could we remove the file all together? Seems like unit tests are also creating that table...

HeartSaVioR:
We need to ensure that created table from unit tests will be dropped when cleaning up tests. Removing data directory would be also OK.

HeartSaVioR:
@abellina I got what you are saying.
@ppoulosk Let's remove test.sql and relevant setting from pom.xml since JdbcClientTest is already in charge to create table.

ppoulosk:
@HeartSaVioR, @abellina, will fix as you suggest.

ppoulosk:
Fixed.  Re-ran tests.  Thanks.

HeartSaVioR:
OK. Checked changeset doesn't affect the build result. Nice work. +1

HeartSaVioR:
@ppoulosk 
Could you rebase this into master and squash commits into one? I guess discarding first commit would be good. 
Also please add STORM-1344 as prefix to commit message since we don't get any information from the commit message.

ppoulosk:
@HeartSaVioR Thanks for the review!!  I've squashed commits and added JIRA to commit comment.

abellina:
@ppoulosk were you able to test without the plugin all together?

ppoulosk:
@abellina Yes.  I've now pushed the change without the plugin.

abellina:
+1 Thanks

HeartSaVioR:
still +1

PR #1663, opened 2016-08-31
STORM-2072 Add map, flatMap with different outputs (T->V) in Trident

HeartSaVioR:
* Introduce overloading methods of map and flatMap
  * Now users can specify new output fields which replace origin fields
* Introduce OperationMapFunction and OperationFlatMapFunction
  * which also extends Operation
  * Trident will automatically determine this and handle properly

Also updated example code and docs page.

This change is needed for optimizing Storm SQL on Trident mode to have minimized steps.
I'll work on optimization on top of this and submit that as well.

PR #1664, opened 2016-08-31
STORM-2073 [Storm SQL] Reduce multi-steps on visitProject into one

HeartSaVioR:
This change is on top of STORM-2072 (#1663).

In STORM-1434 we revamped the way to build Trident topology for Storm SQL.

While revamping visitProject(), it should handle multiple things (calculate expression, projection) and moreover Trident doesn't allow duplicated field name. So we end up having multiple steps - each -> project -> each -> project - which doesn't look good.

STORM-2072 is introducing a way to map / flatMap with specifying different output fields. With this, we can reduce visitProject() to have just 1 step instead of 4 steps.


HeartSaVioR:
Before
https://docs.google.com/spreadsheets/d/1tTyIxbNafvVw0tQjWjiBQCKDCPbbnfbT5JZ7F5JEBNE/edit#gid=0

After
https://docs.google.com/spreadsheets/d/1160ycB16aiUXuge8-bFItxKMw3xfh8Adz0356NlQa_8/edit#gid=0

19 streams are reduced to 13 streams.

Still finding out how to handle AVG without current trick (calculating SUM and COUNT temporary and dividing) with keeping benefit with combiner aggregator (partition aggregation -> global aggregation).

HeartSaVioR:
Closing this in favor of STORM-1446 and also STORM-2148. This is applied to STORM-1446.

PR #1665, opened 2016-08-31
STORM-2074: fix storm-kafka-monitor NPE bug

vesense:
https://issues.apache.org/jira/browse/STORM-2074

### DIFF
--- a/external/storm-kafka-monitor/src/main/java/org/apache/storm/kafka/monitor/KafkaOffsetLagUtil.java
+++ b/external/storm-kafka-monitor/src/main/java/org/apache/storm/kafka/monitor/KafkaOffsetLagUtil.java
@@ -89,10 +89,6 @@ public class KafkaOffsetLagUtil {
                     printUsageAndExit(options, OPTION_ZK_SERVERS_LONG + " and " + OPTION_ZK_COMMITTED_NODE_LONG + " are required  with " +
                             OPTION_OLD_CONSUMER_LONG);
                 }
-                String zkNode = commandLine.getOptionValue(OPTION_ZK_COMMITTED_NODE_LONG);
-                if (zkNode == null || zkNode.length() <= 1) {
-                	 printUsageAndExit(options, OPTION_ZK_COMMITTED_NODE_LONG+" '"+zkNode+"' is invalid.");
-                }
                 String[] topics = commandLine.getOptionValue(OPTION_TOPIC_LONG).split(",");
                 if (topics != null && topics.length > 1) {
                     printUsageAndExit(options, "Multiple topics not supported with option " + OPTION_OLD_CONSUMER_LONG + ". Either a single topic or a " +
@@ -377,12 +373,11 @@ public class KafkaOffsetLagUtil {
         curatorFramework.start();
         String partitionPrefix = "partition_";
         String zkPath = oldKafkaSpoutOffsetQuery.getZkPath();
-         if (zkPath.endsWith("/")) {
+        if (zkPath.endsWith("/")) {
             zkPath = zkPath.substring(0, zkPath.length()-1);
         }
         if (curatorFramework.checkExists().forPath(zkPath) == null) {
-        	System.out.printf("--zk-node '%s' is not exists.%n", zkPath);
-        	System.exit(1);
+            throw new IllegalArgumentException(OPTION_ZK_COMMITTED_NODE_LONG+" '"+zkPath+"' dose not exists.");
         }
         byte[] zkData;
         try {
--- a/external/storm-kafka-monitor/src/main/java/org/apache/storm/kafka/monitor/KafkaOffsetLagUtil.java
+++ b/external/storm-kafka-monitor/src/main/java/org/apache/storm/kafka/monitor/KafkaOffsetLagUtil.java
@@ -89,6 +89,10 @@ public class KafkaOffsetLagUtil {
                     printUsageAndExit(options, OPTION_ZK_SERVERS_LONG + " and " + OPTION_ZK_COMMITTED_NODE_LONG + " are required  with " +
                             OPTION_OLD_CONSUMER_LONG);
                 }
+                String zkNode = commandLine.getOptionValue(OPTION_ZK_COMMITTED_NODE_LONG);
+                if (zkNode == null || zkNode.length() <= 1) {
+                	 printUsageAndExit(options, OPTION_ZK_COMMITTED_NODE_LONG+" '"+zkNode+"' is invalid.");
+                }
                 String[] topics = commandLine.getOptionValue(OPTION_TOPIC_LONG).split(",");
                 if (topics != null && topics.length > 1) {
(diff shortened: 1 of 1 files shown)


vesense:
this should be merged into 1.x-branch.

external/storm-kafka-monitor/src/main/java/org/apache/storm/kafka/monitor/KafkaOffsetLagUtil.java
```diff
@@ -89,6 +89,10 @@ public static void main (String args[]) {
                     printUsageAndExit(options, OPTION_ZK_SERVERS_LONG + " and " + OPTION_ZK_COMMITTED_NODE_LONG + " are required  with " +
                             OPTION_OLD_CONSUMER_LONG);
                 }
+                String zkNode = commandLine.getOptionValue(OPTION_ZK_COMMITTED_NODE_LONG);
```
priyank5485: While adding OPTION_ZK_NODE_COMMITTED_NODE_LONG to options it is added using the method options.addOption(String opt, String longOpt, boolean hasArg, String description) The third argument is set to true in the call which specifies that the argument must have a value. Does not that take care of null or empty value? Do we need to have this explicit check here?

external/storm-kafka-monitor/src/main/java/org/apache/storm/kafka/monitor/KafkaOffsetLagUtil.java
```diff
@@ -373,16 +377,20 @@ private static Options buildOptions () {
         curatorFramework.start();
         String partitionPrefix = "partition_";
         String zkPath = oldKafkaSpoutOffsetQuery.getZkPath();
-        if (!zkPath.endsWith("/")) {
-            zkPath += "/";
+         if (zkPath.endsWith("/")) {
```
priyank5485: Is there a reason for removing trailing / here and then appending in the loop below?  The loop already has a check for curatorFramework.checkExists().forPath(path). May be we can just print and exit therE? Also, can you call the printUsageAndExit method here to be consistent? Or is there a reason you could not use it?

external/storm-kafka-monitor/src/main/java/org/apache/storm/kafka/monitor/KafkaOffsetLagUtil.java
vesense: `zkNode == null` is not unnecessary since `hasArg` is set to true. and I will remove.
but `zkNode.length() <= 1` is necessary because zkPath should start with `/`, so the valid path length should `> 1`.

external/storm-kafka-monitor/src/main/java/org/apache/storm/kafka/monitor/KafkaOffsetLagUtil.java
```diff
@@ -373,16 +377,20 @@ private static Options buildOptions () {
...
+            zkPath = zkPath.substring(0, zkPath.length()-1);
+        }
+        if (curatorFramework.checkExists().forPath(zkPath) == null) {
+        	System.out.printf("--zk-node '%s' is not exists.%n", zkPath);
```
abellina: %n? You mean.. println? Also small nit: "is not exists" => "does not exist"

This is also using tabs so the indentation is off. Why not use spaces (4 spaces)?

external/storm-kafka-monitor/src/main/java/org/apache/storm/kafka/monitor/KafkaOffsetLagUtil.java
vesense: I think it is better to check the variable before use it.  and the zkPath must not ends with `/`.
the `printUsageAndExit` method needs argument  `Options` which is not available in `getOldConsumerOffsetsFromZk`. so I use `exit` directly.

external/storm-kafka-monitor/src/main/java/org/apache/storm/kafka/monitor/KafkaOffsetLagUtil.java
vesense: Will fix. Thanks

vesense:
Thanks @priyank5485 @abellina I think I have addressed your comments.

external/storm-kafka-monitor/src/main/java/org/apache/storm/kafka/monitor/KafkaOffsetLagUtil.java
priyank5485: I am not sure if just checking the length <= 1 is really that useful. As you mentioned that it should start with / may be we add check satisfying all the conditions for a valid zookeeper node name? Or just remove any check? We already have a check to see if node exists later.

external/storm-kafka-monitor/src/main/java/org/apache/storm/kafka/monitor/KafkaOffsetLagUtil.java
priyank5485: This is called from TopologySpoutLag using ShellUtils to display the results in UI. It is expected that KafkaOffsetLagUtil will always exit with code 0, unless there is a command line options error. TopologySpout guarantees that it will never call with wrong command line arguments. For any other runtime errors it will still exit with code 0, printing out the error condition on System.out. Reason is we want the error message to be displayed in UI. That is why there is a catch all block in the main method. If exit code is not 0, ShellUtils will throw an exception and there will be no meaningful message printed on storm UI. Instead of exiting here, can you throw an exception with a message like "zk node does not exist" and let the main method catch it. If you want to test out what I mentioned you can build a quick topology with kafka spout and look at storm ui.

external/storm-kafka-monitor/src/main/java/org/apache/storm/kafka/monitor/KafkaOffsetLagUtil.java
vesense: we can just remove the check since `curatorFramework.checkExists().forPath(zkPath)` will do the same check for zkpath. Thanks.

vesense:
@HeartSaVioR Can we merge this in? and, it's very easy to cherry-pick to 1.x-branch.

HeartSaVioR:
I'd like to wait for @priyank5485 and @abellina to finish reviewing. I'm now reviewing and testing Supervisor V2 and it could take some times.
FYI: I'm on vacation on next week, and will be traveling in several days (not too long but not available for that moment).

vesense:
@HeartSaVioR OK, have a good time.

priyank5485:
@vesense Not sure if you saw my comment about exiting with a code of 1 in getOldConsumerOffsetsFromZk. Can you check that comment and address it?

vesense:
Oh, I missed it. Will update. Thanks @priyank5485

vesense:
@priyank5485 Updated. Can you take a look again?

priyank5485:
+1 (non-binding)

HeartSaVioR:
+1

PR #1642, opened 2016-08-22
STORM-2018: Supervisor V2.  

revans2:
Still needs run as user and CGroup work, but the rest is working

Any feedback on this would be welcome.  I am particularly interested in the readability of the code and how easy is it to understand vs the original supervisor.

All the unit tests pass and I have done a lot of manual testing switching back and forth between the original supervisor and supervisor V2 to ensure that this can be a rolling upgrade.

There are still a number of TODOs in the code.  Most of them are for removing the original supervisor code and cleaning up the result of that.

PR #1667, opened 2016-09-01
STORM-2076: Add new atom to prevent sync-processes from deleting new …

srdo:
…topology code

@HeartSaVioR Could you take a look at this? I'll make a similar change on master if you think this is the right solution.

### DIFF
--- a/storm-core/src/clj/org/apache/storm/daemon/supervisor.clj
+++ b/storm-core/src/clj/org/apache/storm/daemon/supervisor.clj
@@ -328,6 +328,7 @@
    :assignment-id (.getAssignmentId isupervisor)
    :my-hostname (hostname conf)
    :curr-assignment (atom nil) ;; used for reporting used ports when heartbeating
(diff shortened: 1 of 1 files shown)


HeartSaVioR:
Hi @srdo,

Great catch and thanks for the contribution.
Btw, as we're continuously suffering race conditions in supervisor, @revans2 proposed new supervisor design and submitted pull request - [STORM-2018](issues.apache.org/jira/browse/STORM-2018) #1642 

I'm not sure when we can finish reviewing and deciding to merge, but review comments are positive so I guess we can switch supervisor to V2 in near future. For now #1642 is for master, but since we found race even 1.0.2 again, I think we should port back to 1.x and 1.0.x and release.

Could you help to review #1642 if you don't mind? Thanks!

srdo:
@HeartSaVioR Sure, the new design looks much nicer. I agree that backporting the new design if it works well makes more sense than to keep patching these race conditions.

srdo:
Closing this since SupervisorV2 was merged

PR #1668, opened 2016-09-01
STORM-2040 Fix bug on assert-can-serialize

HeartSaVioR:
* type of element of tuple-batch is changed to AddressedTuple but not reflected to assert-can-serialize
* it only raises issue when topology.testing.always.try.serialize is true

This should also be ported back to 1.x and 1.0.x branches.


harshach:
+1

PR #1669, opened 2016-09-01
STORM-2078: enable paging in worker datatable

abellina:
This can get cherry-picked to 1.x


revans2:
+1

abellina:
@revans2 I moved the call to my function that checks for whether the worker components should show or not to a callback that triggers on each render of the datatable. I noticed a bug where filtering would cause odd behavior... this fixes it.

ppoulosk:
+1, NB

HeartSaVioR:
+1

PR #1671, opened 2016-09-02
STORM-2080: storm-submit-tools license check failure

manuzhang:


HeartSaVioR:
Duplicated #1648 (STORM-2054)
Actually there's also a bug which doesn't handle relative path vs absolute path.

manuzhang:
Oh, I missed that. I will comment there and, FYI, some license headers are not in the right form

PR #1674, opened 2016-09-06
STORM-2083: Blacklist scheduler

nilday:
A resolve of [STORM-2083](https://issues.apache.org/jira/browse/STORM-2083)


storm-core/src/jvm/org/apache/storm/scheduler/blacklist/BlacklistScheduler.java
```diff
@@ -0,0 +1,212 @@
+package org.apache.storm.scheduler.blacklist;
```
vesense: missing apache license, same as the following class

storm-core/src/jvm/org/apache/storm/scheduler/blacklist/BlacklistScheduler.java
```diff
@@ -0,0 +1,212 @@
...
+import org.apache.storm.scheduler.*;
+import org.apache.storm.scheduler.blacklist.reporters.IReporter;
+import org.apache.storm.scheduler.blacklist.strategies.IBlacklistStrategy;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.*;
+
+/**
+ * Created by howard.li on 2016/6/29.
```
vesense: remove the author info in javadoc

storm-core/src/jvm/org/apache/storm/scheduler/blacklist/BlacklistScheduler.java
```diff
@@ -0,0 +1,212 @@
...
+            reporter=(IReporter)Class.forName(reporterClassName).newInstance();
+        } catch (ClassNotFoundException e) {
+            LOG.error("Can't find blacklist reporter for name {}",reporterClassName);
+        } catch (InstantiationException e) {
+            LOG.error("Throw InstantiationException blacklist reporter for name {}", reporterClassName);
+        } catch (IllegalAccessException e) {
+            LOG.error("Throw illegalAccessException blacklist reporter for name {}", reporterClassName);
+        }
+
+        String strategyClassName=_conf.containsKey(Config.BLACKLIST_SCHEDULER_STRATEGY)?(String)_conf.get(Config.BLACKLIST_SCHEDULER_STRATEGY):"";
```
vesense: space between `=`, `?` and `:`

storm-core/src/jvm/org/apache/storm/scheduler/blacklist/BlacklistScheduler.java
```diff
@@ -0,0 +1,212 @@
...
+        } catch (IllegalAccessException e) {
+            LOG.error("Throw illegalAccessException blacklist strategy for name {}", reporterClassName);
+        }
+
+        nimbusMonitorFreqSecs=(Integer)_conf.get(Config.NIMBUS_MONITOR_FREQ_SECS);
+        blacklistStrategy.prepare(reporter,toleranceTime,toleranceCount,resumeTime,nimbusMonitorFreqSecs);
+
+        toleranceBuffer=new CircularBuffer<HashMap<String,Set<Integer>>>(toleranceTime/nimbusMonitorFreqSecs);
+        cachedSupervisors=new HashMap<>();
+
```
vesense: unnecessary new line, same as follwing

storm-core/src/jvm/org/apache/storm/scheduler/blacklist/CircularBuffer.java
```diff
@@ -0,0 +1,182 @@
...
+    private int consumerIndex = 0;
+
+    // capacity must be a power of 2 at all times
+    private int capacity;
+    //private int maxCapacity;
+
+    // we mask with capacity -1.  This variable caches that values
+    //private int bitmask;
+
+    private Serializable[] q;
```
vesense: I think we can this `q` more readable

storm-core/src/jvm/org/apache/storm/scheduler/blacklist/CircularBuffer.java
```diff
@@ -0,0 +1,182 @@
...
+            }
+
+            private void checkForModification() {
+                if (ci != consumerIndex) throw new ConcurrentModificationException();
+                if (pi != producerIndex) throw new ConcurrentModificationException();
+            }
+        };
+    }
+
+    public static void main(String []args){
```
vesense: move this to a unit test?

storm-core/src/jvm/org/apache/storm/scheduler/blacklist/strategies/DefaultBlacklistStrategy.java
```diff
@@ -0,0 +1,142 @@
...
+        _toleranceTime = toleranceTime;
+        _toleranceCount = toleranceCount;
+        _resumeTime = resumeTime;
+        _nimbusMonitorFreqSecs = nimbusMonitorFreqSecs;
+        blacklist = new TreeMap<>();
+    }
+
+    @Override
+    public Set<String> getBlacklist(CircularBuffer<HashMap<String, Set<Integer>>> toleranceBuffer, Cluster cluster, Topologies topologies) {
+        //Set<String> blacklist=new HashSet<String>();
```
vesense: remove

revans2:
Overall I like the concept here, and if a supervisor is appearing/disappearing a lot we probably do want to blacklist that supervisor.  That said I have a few system concerns.

1) I would like to see this feature made a part of nimbus, and not so much be a scheduler.  The algorithm is generic enough that we could easily wrap all of the schedulers.  If you let nimbus hand it the scheduler that it is supposed to wrap through the constructor, then you can make the internals agnostic to the scheduler underneath.  This would also fix some of the build dependency issues you where having with needing to build blacklising after building clojure.
2) I like having the reporting plugin, but I really want to see blacklisted nodes show up on the storm UI.  We have the supervisor pages now, and the supervisor table on the main page.  If I am an administrator I would much rather look at a UI to see what is happening with a supervisor instead of parsing a lot of logs.
3) Cluster wide failures.  Blacklisting is a good feature until something odd happens and the entire cluster is blacklisted.  (completely theoretical) Lets say that we have nimbus HA and it is the primary nimbus nodes that gets lots of network loss.  After a while it blacklists the entire cluster, when it is just nimbus that is bad.  I want to be sure that we have something in place that can detect and handle appropriately a situation where the majority of the nodes appear to be bad.
4) master.  This patch is just for the 1.x branch.  That is fine, but before we can merge it in we need a patch for master as well.

vesense:
Agree with @revans2 's opinion.  Especially, make this feature as a part of nimbus and show the blacklist nodes on the UI.
For the third point, I feel we can add a  task(eg: `ping` the standby nimbus or one node we think it has good network and less pressure) for nimbus to ensure itself is healthy before checking other supervisor nodes.

revans2:
My example was theoretical, I honestly don't know if in practice nimbus would see the supervisor appear then disappear if nimbus's nic was bad.  The more common case would be if a ZK nic was bad, that could cause this.  But that is beside the point.  There are lots of different things that could make most of the cluster look bad or actually go bad for real.  I am fine with detecting/handling some cases that we know about and can reproduce, but we should have some sort of default catch all.

For example if HDFS loses too many nodes it goes into read only mode, but for YARN it ignores it but uses metrics to alert the cluster owners.  If we feel that because we are primarily compute, like YARN and want to have a metric about how many nodes are blacklisted that seems like a perfectly fine default.  If we run into other situations and can detect/auto-correct them even better.

knusbaum:
We could just have some "relative badness" heuristic. I would expect the supervisors would be in a bell curve of "goodness/badness" and we can just blacklist the ones in the bad part of the tail, that way if they're all *kind of* bad, they all stay un-blacklisted.

revans2:
For me either the supervisor works or it does not.  If it doesn't work don't use it.  If it is slightly better then another supervisor does not tell me much.  If too many don't work let someone know so we can do something about it.

knusbaum:
@revans2 +1 for that. That's simpler to understand and implement.

vesense:
Thanks @revans2 for the details. get it.

nilday:
Thanks for all the advises you give. For the suggestions given by @revans2 :
1) We concerned implement blacklist in nimbus before. As a newbie in storm contribution and clojure, I choose to implement it as a scheduler so I can write the code in Java and has the minimum affect to the storm core so we can control the risk. The BlacklistScheduler now uses the DefaultScheduler underlying, and we can easily edit so code to let it support configuration to any scheduler. I would like to have a try to add the blacklist to nimbus, as I can't wait someone else implement it for us.
2)Showing blacklist on UI is good idea.
3)We have the same worry as you do. In the PR I submit this time, we have some code dealing with it. If the cluster have too many blacklist leading to lack of slots, the *DefaultBlacklistStrategy* will use *releaseBlacklistWhenNeeded* method to temporarily resume some supervisors from blacklist so we can try to assign some job to it. It's not good enough but at least it's a try. This is definitely a problem, I think there must be a config which can switch on or off the blacklist feature before it's finally stable enougth. @knusbaum talked about some heuristic algorithm, which we also had in our mind before. We think we may use the number of bad slots on one machine and the number of topoplogies they belong to  to calculate the healthiness of a machine.   The idea is not matrue enough so we haven't implement it, and we can write another IBlacklistStrategy to do that. 

As I am a  newbie, there may be  a lot barriers in front of me. I will be grateful if I can have your assistance when I face them. Thanks a lot.

revans2:
@nilday I was thinking that you didn't really need to change your code much at all, and still leave it in java.  I would be perfectly happy to have your scheduler take a scheduler as input to the constructor and wrap it.  Then you could modify one small place in nimbus.clj.

https://github.com/apache/storm/blob/master/storm-core/src/clj/org/apache/storm/daemon/nimbus.clj#L116-L132

to wrap the returned scheduler in your implementation.

The UI work can be a follow on JIRA, but I would like to see at least a metric put in place that shows the number of blacklisted nodes so we can at least set up alerts on it.

knusbaum:
In addition to the ones @vesense noted, there are many other operators without spaces around them. Adding space around `=`, `+`, `:`, `<`, etc, will make the code more readable.

storm-core/src/jvm/org/apache/storm/scheduler/blacklist/BlacklistScheduler.java
```diff
@@ -0,0 +1,212 @@
+package org.apache.storm.scheduler.blacklist;
+
+import org.apache.storm.Config;
+import org.apache.storm.scheduler.*;
+import org.apache.storm.scheduler.blacklist.reporters.IReporter;
+import org.apache.storm.scheduler.blacklist.strategies.IBlacklistStrategy;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.*;
```
knusbaum: I think we want to avoid such large wildcard imports.

storm-core/src/jvm/org/apache/storm/scheduler/blacklist/BlacklistScheduler.java
```diff
@@ -0,0 +1,212 @@
...
+            LOG.error("Throw InstantiationException blacklist reporter for name {}", reporterClassName);
+        } catch (IllegalAccessException e) {
+            LOG.error("Throw illegalAccessException blacklist reporter for name {}", reporterClassName);
+        }
+
+        String strategyClassName=_conf.containsKey(Config.BLACKLIST_SCHEDULER_STRATEGY)?(String)_conf.get(Config.BLACKLIST_SCHEDULER_STRATEGY):"";
+        try {
+            blacklistStrategy=(IBlacklistStrategy)Class.forName(strategyClassName).newInstance();
+        } catch (ClassNotFoundException e) {
+            LOG.error("Can't find blacklist strategy for name {}",reporterClassName);
```
knusbaum: I think you mean `strategyClassName` in these error messages?

storm-core/src/jvm/org/apache/storm/scheduler/blacklist/BlacklistScheduler.java
```diff
@@ -0,0 +1,212 @@
...
+        } catch (InstantiationException e) {
+            LOG.error("Throw InstantiationException blacklist strategy for name {}", reporterClassName);
+        } catch (IllegalAccessException e) {
+            LOG.error("Throw illegalAccessException blacklist strategy for name {}", reporterClassName);
+        }
+
+        nimbusMonitorFreqSecs=(Integer)_conf.get(Config.NIMBUS_MONITOR_FREQ_SECS);
+        blacklistStrategy.prepare(reporter,toleranceTime,toleranceCount,resumeTime,nimbusMonitorFreqSecs);
```
knusbaum: You catch errors above without handling the error, which would leave `blacklistStrategy` null. This will throw an NPE here. Probably better to just throw the exceptions when `blacklistStrategy` can't be found/instantiated.

knusbaum: Same thing for reporter, actually. You catch the exceptions and continue, leaving reporter null. This will blow up in the `DefaultBlacklistStrategy`

storm-core/src/jvm/org/apache/storm/scheduler/blacklist/strategies/DefaultBlacklistStrategy.java
```diff
@@ -0,0 +1,142 @@
...
+
+        _toleranceTime = toleranceTime;
+        _toleranceCount = toleranceCount;
+        _resumeTime = resumeTime;
+        _nimbusMonitorFreqSecs = nimbusMonitorFreqSecs;
+        blacklist = new TreeMap<>();
+    }
+
+    @Override
+    public Set<String> getBlacklist(CircularBuffer<HashMap<String, Set<Integer>>> toleranceBuffer, Cluster cluster, Topologies topologies) {
```
knusbaum: The only thing that's ever actually used in `toleranceBuffer` here (except for logging it) is the key of the hashmap. It looks like `toleranceBuffer` should really be a `List<String>` called `supervisorsWithFailures` or something like that. Currently, both the type and the name are misleading.

storm-core/src/jvm/org/apache/storm/scheduler/blacklist/strategies/IBlacklistStrategy.java
```diff
@@ -0,0 +1,37 @@
...
+
+public interface IBlacklistStrategy {
+
+    public void prepare(IReporter reporter, int toleranceTime, int toleranceCount, int resumeTime, int nimbusMonitorFreqSecs);
+
+    public Set<String> getBlacklist(CircularBuffer<HashMap<String, Set<Integer>>> toleranceBuffer, Cluster cluster, Topologies topologies);
+
+    public void resumeFromBlacklist();
+
+    //public void releaseBlacklistWhenNeeded(Cluster cluster,Topologies topologies);
```
knusbaum: Remove?

storm-core/src/jvm/org/apache/storm/scheduler/blacklist/BlacklistScheduler.java
```diff
@@ -0,0 +1,212 @@
...
+
+    @Override
+    public void prepare(Map conf) {
+        LOG.info("prepare black list scheduler");
+        LOG.info(conf.toString());
+        defaultScheduler=new DefaultScheduler();
+        defaultScheduler.prepare(conf);
+        _conf=conf;
+        if(_conf.containsKey(Config.BLACKLIST_SCHEDULER_TOLERANCE_TIME)){
+            toleranceTime=(Integer)_conf.get(Config.BLACKLIST_SCHEDULER_TOLERANCE_TIME);
```
kishorvpatil: Should we use [Utils.getInt](https://github.com/apache/storm/blob/master/storm-core/src/jvm/org/apache/storm/utils/Utils.java#L753) or [Utils.getInt](https://github.com/apache/storm/blob/master/storm-core/src/jvm/org/apache/storm/utils/Utils.java#L792)

storm-core/src/jvm/org/apache/storm/scheduler/blacklist/BlacklistScheduler.java
```diff
@@ -0,0 +1,212 @@
...
+        }
+        if(_conf.containsKey(Config.BLACKLIST_SCHEDULER_TOLERANCE_COUNT)){
```
kishorvpatil: Should we use [Utils.getInt](https://github.com/apache/storm/blob/master/storm-core/src/jvm/org/apache/storm/utils/Utils.java#L753) or [Utils.getInt](https://github.com/apache/storm/blob/master/storm-core/src/jvm/org/apache/storm/utils/Utils.java#L792) to avoid `Null` being assigned.

storm-core/src/jvm/org/apache/storm/scheduler/blacklist/BlacklistScheduler.java
```diff
@@ -0,0 +1,212 @@
...
+            toleranceCount=(Integer)_conf.get(Config.BLACKLIST_SCHEDULER_TOLERANCE_COUNT);
+        }
+        if(_conf.containsKey(Config.BLACKLIST_SCHEDULER_RESUME_TIME)){
+            resumeTime=(Integer)_conf.get(Config.BLACKLIST_SCHEDULER_RESUME_TIME);
+        }
```
kishorvpatil: Should we use [Utils.getInt](https://github.com/apache/storm/blob/master/storm-core/src/jvm/org/apache/storm/utils/Utils.java#L753) or [Utils.getInt](https://github.com/apache/storm/blob/master/storm-core/src/jvm/org/apache/storm/utils/Utils.java#L792) to avoid `Null` being assigned.

storm-core/src/jvm/org/apache/storm/scheduler/blacklist/BlacklistScheduler.java
```diff
@@ -0,0 +1,212 @@
...
+
+    @Override
+    public void schedule(Topologies topologies, Cluster cluster) {
+        LOG.info("running Black List scheduler");
+        Map<String, SupervisorDetails> supervisors = cluster.getSupervisors();
+        for(Map.Entry<String,SupervisorDetails> entry:supervisors.entrySet()){
+            SupervisorDetails supervisorDetails=entry.getValue();
+            String supervisorName=entry.getKey();
+            Set<Integer> ports=supervisorDetails.getAllPorts();
+            LOG.info("supervisor: "+supervisorDetails.getHost()+" ports"+ports);
```
kishorvpatil: Would it help to override `toString` method to `SupervisorDetails` class?

HeartSaVioR:
@nilday Do you have any updates?

nilday:
@HeartSaVioR not yet, busy with something else. Will do it as soon as I have some time.

storm-core/src/jvm/org/apache/storm/scheduler/blacklist/BlacklistScheduler.java
nilday: Maybe it's not necessary? [BLACKLIST_SCHEDULER_TOLERANCE_TIME  ] (https://github.com/nilday/storm/blob/blacklist-scheduler/storm-core/src/jvm/org/apache/storm/Config.java#L2164) has Annotation of 
 [ isInteger] (https://github.com/apache/storm/blob/master/storm-core/src/jvm/org/apache/storm/validation/ConfigValidationAnnotations.java#L106)

storm-core/src/jvm/org/apache/storm/scheduler/blacklist/BlacklistScheduler.java
nilday: same as above

storm-core/src/jvm/org/apache/storm/scheduler/blacklist/BlacklistScheduler.java
nilday: same as above

nilday:
Thanks for you all guys. I will consider seriously all your suggestions.
But currently I'll be still working on 1.0.x branch ,since my company has just update to 1.0 version  and master branch is very unstable with so many changes (Nimbus,which scheduler depend on,  will be changed into Java, and it's still in progress). Once we have stable blacklist scheduler in our production environment working for some while and master branch has all the dependencies ready, I will migrate this to master.
Thanks again.

storm-core/src/jvm/org/apache/storm/scheduler/blacklist/BlacklistScheduler.java
nilday: SupervisorDetails is a class with all the supervisors info in it. It's not necessary to show all the info here. But if I override toString method, I have to put everything in it.

nilday:
As soon as I find out that nimbus  has been transferred to Java, I start to work on this again. The blacklist scheduler has been working on our production environment(Storm v1.0.1) for several months and it worked fine!
 I read carefully for all your suggestions and try to fix our improve them in some way. Here them are listed:
[X] Add Apache Header
[X] Remove AuthorInfo
[X] space between =, ? and :
[X] unnecessary new line
[X] q more readable
[X] move CircularBuffer main to a unit test
[X] remove comment in DefaultBlacklistStrategy.java#L54
[X] work on master
[X] blacklist scheduler wrap others in nimbus
[ ] blacklist on UI
[X] blacklist to metric
[X] avoid wildcard imports
[X]  ...che/storm/scheduler/blacklist/BlacklistScheduler.java#L66 strategyClassName
[X] Adding space around =, +, :, <, etc
[X] NPE when blacklistStrategy or reporter can't be found
[X] toleranceBuffer type and name misleading
[X] Remove unused comments in IBlacklistStrategy

For the UI work, I may open another issue and work on it. The rest works are listed above.

storm-core/src/jvm/org/apache/storm/scheduler/blacklist/BlacklistScheduler.java
```diff
@@ -0,0 +1,252 @@
...
+        });
+    }
+
+    @Override
+    public void schedule(Topologies topologies, Cluster cluster) {
+        LOG.debug("running Black List scheduler");
+        Map<String, SupervisorDetails> supervisors = cluster.getSupervisors();
+        for (Map.Entry<String, SupervisorDetails> entry : supervisors.entrySet()) {
+            SupervisorDetails supervisorDetails = entry.getValue();
+            String hosts = supervisorDetails.getHost();
```
abellina: nit, hosts -> host

storm-core/src/jvm/org/apache/storm/scheduler/blacklist/BlacklistScheduler.java
```diff
@@ -0,0 +1,252 @@
...
+            throw new RuntimeException(e);
+        } catch (InstantiationException e) {
+            LOG.error("Throw InstantiationException blacklist strategy for name {}", strategyClassName);
+            throw new RuntimeException(e);
+        } catch (IllegalAccessException e) {
+            LOG.error("Throw illegalAccessException blacklist strategy for name {}", strategyClassName);
+            throw new RuntimeException(e);
+        }
+
+        nimbusMonitorFreqSecs = (Integer) _conf.get(Config.NIMBUS_MONITOR_FREQ_SECS);
```
abellina: I think this should be a separate setting for the blacklist scheduler. It is not intuitive that changing the nimbus setting, modifies the behavior of this strategy.

storm-core/src/jvm/org/apache/storm/scheduler/blacklist/BlacklistScheduler.java
abellina: honestly, this whole loop is just for debugging purposes. I'd argue we should probably remove it, the UI shows this already.

storm-core/src/jvm/org/apache/storm/scheduler/blacklist/BlacklistScheduler.java
```diff
@@ -0,0 +1,252 @@
...
+                }
+            } else {
+                cachedSupervisors.put(key, supervisorDetails.getAllPorts());//new supervisor to cache
+            }
+        }
+
+        badSupervisorsTolerance.add(badSupervisors);
+    }
+
+    private Set<Integer> badSlots(Map<String, SupervisorDetails> supervisors, String supervisorKey) {
```
abellina: this should take SupervisorDetails directly instead of the supervisors map.

storm-core/src/jvm/org/apache/storm/scheduler/blacklist/BlacklistScheduler.java
```diff
@@ -0,0 +1,252 @@
...
+        SupervisorDetails supervisor = supervisors.get(supervisorKey);
+        Set<Integer> cachedSupervisorPorts = cachedSupervisors.get(supervisorKey);
+        Set<Integer> supervisorPorts = supervisor.getAllPorts();
+
+        Set<Integer> newPorts = Sets.difference(supervisorPorts, cachedSupervisorPorts);
+        if (newPorts.size() > 0) {
+            cachedSupervisors.put(supervisorKey, Sets.union(newPorts, supervisor.getAllPorts()));
```
abellina: shouldn't the value here be just supervisor.getAllPorts()? The union doesn't hurt, just I think "newPorts" is a subset of supervisor.getAllPorts.

storm-core/src/jvm/org/apache/storm/scheduler/blacklist/BlacklistScheduler.java
```diff
@@ -0,0 +1,252 @@
...
+        }
+
+        Set<Integer> difference = Sets.difference(cachedSupervisorPorts, supervisorPorts);
```
abellina: isn't the set "difference" === "newPorts"?

storm-core/src/jvm/org/apache/storm/scheduler/blacklist/BlacklistScheduler.java
nilday: I think it's hard to avoid the influence of NIMBUS_MONITOR_FREQ_SECS, as the blacklist scheduler is a SCHEDULER at first. It is called every NIMBUS_MONITOR_FREQ_SECS.
I take this conf here only to calculate the length of sliding window. The window size is "toleranceTime / nimbusMonitorFreqSecs ", so every time the scheduler is called ,the window will slide for 1 step.
I think even if I create a new configure for this, it MUST equal to NIMBUS_MONITOR_FREQ_SECS so that the sliding window can work properly . So there is no need to do so.

storm-core/src/jvm/org/apache/storm/scheduler/blacklist/BlacklistScheduler.java
nilday: this loop will be removed

storm-core/src/jvm/org/apache/storm/scheduler/blacklist/BlacklistScheduler.java
nilday: will do so

storm-core/src/jvm/org/apache/storm/scheduler/blacklist/BlacklistScheduler.java
nilday: Well , I think it's a bug here. what I mean is Sets.union(newPorts, cachedSupervisorPorts) to add new port to cachedSupervisors.

storm-core/src/jvm/org/apache/storm/scheduler/blacklist/BlacklistScheduler.java
```diff
@@ -0,0 +1,252 @@
...
+    }
+
+    private void badSupervisors(Map<String, SupervisorDetails> supervisors) {
+        Set<String> cachedSupervisorsKeySet = cachedSupervisors.keySet();
+        Set<String> supervisorsKeySet = supervisors.keySet();
+
+        Set<String> badSupervisorKeys = Sets.difference(cachedSupervisorsKeySet, supervisorsKeySet);//cached supervisor doesn't show up
+        HashMap<String, Set<Integer>> badSupervisors = new HashMap<String, Set<Integer>>();
+        for (String key : badSupervisorKeys) {
+            badSupervisors.put(key, cachedSupervisors.get(key));
```
nilday: same as above.  this will only and bad supervisors but new ones. 
btw I think 'symmetricDifference' is what you mean.

storm-core/src/jvm/org/apache/storm/scheduler/blacklist/BlacklistScheduler.java
```diff
@@ -0,0 +1,252 @@
...
+        Set<Integer> badSlots = new HashSet<>();
+        for (int port : difference) {
+            badSlots.add(port);
+        }
+        return badSlots;
+    }
+
+    public Set<String> getBlacklistHosts(Cluster cluster, Topologies topologies) {
+        Set<String> blacklist = blacklistStrategy.getBlacklist(badSupervisorsTolerance.toList(), cluster, topologies);
+        Set<String> blacklistHost = new HashSet<>();
```
nilday: will do so.

storm-core/src/jvm/org/apache/storm/scheduler/blacklist/BlacklistScheduler.java
abellina: Ok. But isn't the result of Sets.union(newPorts, cachedSupervisorPorts) == supervisorPorts? I think you just need to: cachedSupervisors.put(supervisorKey, supervisorPorts);

nilday: Not exactly.
Image that:
In normal case the supervisor.slots.ports:[1,2,3,4],and the cachedSupervisors is also [1,2,3,4].Now port 1 break down ,then the input supervisorPorts now will be [2,3,4], but we'd like to keep cachedSupervisors to be [1,2,3,4] not [2,3,4].
The union operation is for this case : 
we reconfigure the supervisor.slots.ports form [1,2,3,4] to [2,3,4,5](may be port 1 is used for sth. else). The cachedSupervisors will now be [1,2,3,4,5] now. and port 1 will be added to blacklist soon because it will always be badSlot. But after the **BLACKLIST_SCHEDULER_TOLERANCE_TIME** ,it will be removed from cachedSupervisors  by **removeLongTimeDisappearFromCache()**, and the supervisor will be resumed from blacklist.
In this case, the supervisor which changed slot configure will be miss-added to blacklist for **BLACKLIST_SCHEDULER_RESUME_TIME**. But in the case that this will occasionally, it is acceptable.

storm-core/src/jvm/org/apache/storm/scheduler/blacklist/CircularBuffer.java
```diff
@@ -0,0 +1,174 @@
...
+        //bitmask = oldBuffer.bitmask;
+        underlying = new Serializable[oldBuffer.underlying.length];
+        System.arraycopy(oldBuffer.underlying, 0, underlying, 0, underlying.length);
+    }
+
+    private boolean isFull() {
+        return size == capacity;
+    }
+
+    public boolean add(Serializable obj) {
```
revans2: Shouldn't this be

```public boolean add(T obj)```

Also if the is overriding an existing method we should use the `@Override` annotation.  This is for all of the methods here.

File: storm-core/src/jvm/org/apache/storm/scheduler/blacklist/CircularBuffer.java [hunk 25]
```diff
@@ -0,0 +1,174 @@
...
+    // This is the largest capacity allowed by this implementation
+    private static final int MAX_CAPACITY = 1 << 30;
+
+    private int size = 0;
+    private int producerIndex = 0;
+    private int consumerIndex = 0;
+
+    private int capacity;
+
+    private Serializable[] underlying;
```
revans2: Shouldn't this be an array of type `T`

storm-core/src/jvm/org/apache/storm/scheduler/blacklist/CircularBuffer.java
```diff
@@ -0,0 +1,174 @@
...
+            throw new IllegalArgumentException("Capacity greater than " +
+                    "allowed");
+        }
+
+        this.capacity = capacity;
+        underlying = new Serializable[this.capacity];
+    }
+
+    // Constructor used by clone()
+    private CircularBuffer(CircularBuffer oldBuffer) {
```
revans2: This shouldn't oldBuffer be `CircularBuffer<T>` or something like that.

storm-core/src/jvm/org/apache/storm/scheduler/blacklist/CircularBuffer.java
```diff
@@ -0,0 +1,174 @@
...
+import java.io.Serializable;
+import java.util.AbstractCollection;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.ConcurrentModificationException;
+import java.util.Iterator;
+import java.util.List;
+import java.util.NoSuchElementException;
+
+public final class CircularBuffer<T extends Serializable> extends AbstractCollection<T> implements Serializable {
```
revans2: I am a little confused why we need all of this.  It feels like we could do all of this with an ArrayBlockingQueue.

```
public final class CircularBuffer<T extends Serializable> extends ArrayBlockingQueue<T> {
    @Override
    public boolean add(T obj) {
        while (!offer(obj)) {
            poll();
        }
    }

    public List<T> toList() {
        return new ArrayList<>(this);
    }
}
```

revans2: And some javadocs explaining what it does any why would be helpful, because even though this is currently implemented as a CircularBuffer it has slightly different behavior from a typical Collection.

storm-core/src/jvm/org/apache/storm/scheduler/blacklist/strategies/DefaultBlacklistStrategy.java
```diff
@@ -0,0 +1,149 @@
...
+
+    private int _toleranceTime;
+    private int _toleranceCount;
+    private int _resumeTime;
+    private int _nimbusMonitorFreqSecs;
+
+    private TreeMap<String, Integer> blacklist;
+
+    @Override
+    public void prepare(IReporter reporter, int toleranceTime, int toleranceCount, int resumeTime, int nimbusMonitorFreqSecs) {
```
revans2: Could we have conf passed in here too, or not make this a plugin?  I know we don't need/use it now, but if we want this to be user pluggable we should think about what other things people might want to do.

storm-core/src/jvm/org/apache/storm/scheduler/blacklist/strategies/IBlacklistStrategy.java
```diff
@@ -0,0 +1,37 @@
...
+
+public interface IBlacklistStrategy {
+
+    public void prepare(IReporter reporter, int toleranceTime, int toleranceCount, int resumeTime, int nimbusMonitorFreqSecs);
+
+    public Set<String> getBlacklist(List<HashMap<String, Set<Integer>>> toleranceBuffer, Cluster cluster, Topologies topologies);
+
+    public void resumeFromBlacklist();
+
+}
```
revans2: Could we have some javadocs explaining how this is used and the life cycle of the strategy.

storm-core/src/jvm/org/apache/storm/scheduler/blacklist/strategies/IBlacklistStrategy.java
```diff
@@ -0,0 +1,37 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.storm.scheduler.blacklist.strategies;
+
+import org.apache.storm.scheduler.Cluster;
+import org.apache.storm.scheduler.Topologies;
+import org.apache.storm.scheduler.blacklist.CircularBuffer;
```
revans2: I don't think this is used.

storm-core/src/jvm/org/apache/storm/scheduler/blacklist/BlacklistScheduler.java
```diff
@@ -0,0 +1,245 @@
...
+
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.Callable;
+
+public class BlacklistScheduler implements IScheduler {
+    private static final Logger LOG = LoggerFactory.getLogger(BlacklistScheduler.class);
+    IScheduler underlyingScheduler;
```
nilday: Should I change it to private final?

storm-core/src/jvm/org/apache/storm/scheduler/blacklist/BlacklistScheduler.java
```diff
@@ -0,0 +1,245 @@
...
+        if (_conf.containsKey(Config.BLACKLIST_SCHEDULER_TOLERANCE_TIME)) {
+            toleranceTime = (Integer) _conf.get(Config.BLACKLIST_SCHEDULER_TOLERANCE_TIME);
+        }
+        if (_conf.containsKey(Config.BLACKLIST_SCHEDULER_TOLERANCE_COUNT)) {
+            toleranceCount = (Integer) _conf.get(Config.BLACKLIST_SCHEDULER_TOLERANCE_COUNT);
+        }
+        if (_conf.containsKey(Config.BLACKLIST_SCHEDULER_RESUME_TIME)) {
+            resumeTime = (Integer) _conf.get(Config.BLACKLIST_SCHEDULER_RESUME_TIME);
+        }
+        String reporterClassName = _conf.containsKey(Config.BLACKLIST_SCHEDULER_REPORTER) ? (String) _conf.get(Config.BLACKLIST_SCHEDULER_REPORTER) : "";
```
nilday: done and so as the below

storm-core/src/jvm/org/apache/storm/scheduler/blacklist/CircularBuffer.java
nilday: Sub class of ArrayBlockingQueue passed all my unit test. I think it may work as well.
CircularBuffer is just one possible choice for this function, and it's just first option come into my mind. It seems like ArrayBlockingQueue  is a better choice.
Thus, the name CircularBuffer is not precise enough. what would you suggest it should be ?

storm-core/src/jvm/org/apache/storm/scheduler/blacklist/CircularBuffer.java
nilday: This and all below will be solved if I change it to ArrayBlockingQueue.

storm-core/src/jvm/org/apache/storm/scheduler/blacklist/BlacklistScheduler.java
```diff
@@ -0,0 +1,245 @@
...
+    }
+
+    @Override
+    public void prepare(Map conf) {
+        LOG.info("prepare black list scheduler");
+        LOG.info(conf.toString());
+        underlyingScheduler.prepare(conf);
+        _conf = conf;
+        if (_conf.containsKey(Config.BLACKLIST_SCHEDULER_TOLERANCE_TIME)) {
+            toleranceTime = (Integer) _conf.get(Config.BLACKLIST_SCHEDULER_TOLERANCE_TIME);
```
nilday: Utils.getInt is done.
but I don't think it's a good idea that I set nimbusMonitorFreqSecs to 10 by default here I think change a bit in nimbus_test.clj is better.

storm-core/src/jvm/org/apache/storm/scheduler/blacklist/strategies/DefaultBlacklistStrategy.java
nilday: May be pass conf here is a better idea, but if I do so, I will have to parse conf again in DefaultBlacklistStrategy, and I 'll copy the code from BlacklistScheduler. I think it's not a good idea for maintainance. Any good idea?

storm-core/src/jvm/org/apache/storm/scheduler/blacklist/strategies/IBlacklistStrategy.java
nilday: removed

storm-core/src/jvm/org/apache/storm/scheduler/blacklist/strategies/IBlacklistStrategy.java
nilday: will do so

HeartSaVioR:
This feature looks like good thing to adopt. If @nilday is still willing to adopt this to the Storm codebase, I'd like to spend some time to review this as well.
@nilday Could you rebase once more if you really don't mind? Please note that we split storm-core to several modules, and all of your changes need to be placed to `storm-server` module.

nilday:
@HeartSaVioR I'll rebase it as soon as I can. As the storm-core has been restructured, it may take a while.

HeartSaVioR:
@nilday Sure, please take your time. Thanks for keeping interest of contributing. :)

storm-core/src/jvm/org/apache/storm/scheduler/blacklist/strategies/DefaultBlacklistStrategy.java
```diff
@@ -0,0 +1,165 @@
...
+                }
+            }
+            int shortage = totalNeedNumWorkers - availableSlotsNotInBlacklistCount;
+
+            if (shortage > 0) {
+                LOG.info("total needed num of workers :{}, available num of slots not in blacklist :{},num blacklist :{}, will release some blacklist."
+                        , totalNeedNumWorkers, availableSlotsNotInBlacklistCount, blacklist.size());
+                //release earliest blacklist
+                Set<String> readyToRemove = new HashSet<String>();
+                for (String supervisor : blacklist.keySet()) {//blacklist is treeMap sorted by value, value minimum meas earliest
```
HeartSaVioR: nit: space between { and /, and `value minimum meas` -> `minimum value means`

storm-core/src/jvm/org/apache/storm/scheduler/blacklist/strategies/DefaultBlacklistStrategy.java
```diff
@@ -0,0 +1,165 @@
...
+                    if (availableSupervisors.containsKey(supervisor)) {
+                        Set<Integer> ports = cluster.getAvailablePorts(availableSupervisors.get(supervisor));
+                        readyToRemove.add(supervisor);
+                        shortage -= ports.size();
+                        if (shortage <= 0) {//released enough supervisor
```
HeartSaVioR: nit: space between { and /

storm-core/test/jvm/org/apache/storm/scheduler/blacklist/FaultGenerateUtils.java
```diff
@@ -0,0 +1,72 @@
...
+import org.apache.storm.scheduler.SupervisorDetails;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class FaultGenerateUtils {
+
+    public static List<Map<String, SupervisorDetails>> getSupervisorsList(int supervisorCount, int slotCount, int[][][] faults) {
```
HeartSaVioR: This looks like not used anywhere. If then let's remove, and if not let's replace 3-dimension int array with other data structure, given that it affects readability.

storm-core/test/jvm/org/apache/storm/scheduler/blacklist/TestBlacklistScheduler.java
```diff
@@ -0,0 +1,326 @@
...
+        bs.schedule(topologies, cluster);
+        cluster = new Cluster(iNimbus, TestUtilsForBlacklistScheduler.removeSupervisorFromSupervisors(supMap, "sup-0"), TestUtilsForBlacklistScheduler.assignmentMapToImpl(cluster.getAssignments()), config);
+        bs.schedule(topologies, cluster);
+        cluster = new Cluster(iNimbus, TestUtilsForBlacklistScheduler.removeSupervisorFromSupervisors(supMap, "sup-0"), TestUtilsForBlacklistScheduler.assignmentMapToImpl(cluster.getAssignments()), config);
+        bs.schedule(topologies, cluster);
+        cluster = new Cluster(iNimbus, supMap, TestUtilsForBlacklistScheduler.assignmentMapToImpl(cluster.getAssignments()), config);
+        bs.schedule(topologies, cluster);
+        Set<String> hosts = new HashSet<>();
+        hosts.add("host-0");
+        Assert.assertEquals("blacklist", hosts, cluster.getBlacklistedHosts());
```
HeartSaVioR: Inline & replace hosts with `Collections.singleton("host-0")`

storm-core/test/jvm/org/apache/storm/scheduler/blacklist/TestBlacklistScheduler.java
```diff
@@ -0,0 +1,326 @@
...
+
+        cluster = new Cluster(iNimbus, TestUtilsForBlacklistScheduler.removePortFromSupervisors(supMap, "sup-0", 0), TestUtilsForBlacklistScheduler.assignmentMapToImpl(cluster.getAssignments()), config);
+        bs.schedule(topologies, cluster);
+        cluster = new Cluster(iNimbus, TestUtilsForBlacklistScheduler.removePortFromSupervisors(supMap, "sup-0", 0), TestUtilsForBlacklistScheduler.assignmentMapToImpl(cluster.getAssignments()), config);
+        bs.schedule(topologies, cluster);
+        cluster = new Cluster(iNimbus, supMap, new HashMap<String, SchedulerAssignmentImpl>(), config);
+        bs.schedule(topologies, cluster);
+        Set<String> hosts = new HashSet<>();
+        hosts.add("host-0");
+        Assert.assertEquals("blacklist", hosts, cluster.getBlacklistedHosts());
```
HeartSaVioR: Same here: Inline & replace hosts with Collections.singleton("host-0")

storm-core/test/jvm/org/apache/storm/scheduler/blacklist/TestBlacklistScheduler.java
```diff
@@ -0,0 +1,326 @@
...
+        cluster = new Cluster(iNimbus, TestUtilsForBlacklistScheduler.removeSupervisorFromSupervisors(supMap, "sup-0"), TestUtilsForBlacklistScheduler.assignmentMapToImpl(cluster.getAssignments()), config);
+        bs.schedule(topologies, cluster);
+        cluster = new Cluster(iNimbus, TestUtilsForBlacklistScheduler.removeSupervisorFromSupervisors(supMap, "sup-0"), TestUtilsForBlacklistScheduler.assignmentMapToImpl(cluster.getAssignments()), config);
+        bs.schedule(topologies, cluster);
+        cluster = new Cluster(iNimbus, supMap, new HashMap<String, SchedulerAssignmentImpl>(), config);
+        bs.schedule(topologies, cluster);
+        Set<String> hosts = new HashSet<>();
+        hosts.add("host-0");
+        Assert.assertEquals("blacklist", hosts, cluster.getBlacklistedHosts());
+        for (int i = 0; i < 300 / 10 - 2; i++) {
```
HeartSaVioR: Let's put comment for explaining expression in for statement. (`300 / 10 - 2`)

storm-core/test/jvm/org/apache/storm/scheduler/blacklist/TestBlacklistScheduler.java
```diff
@@ -0,0 +1,326 @@
...
+            bs.schedule(topologies, cluster);
+        }
+        Assert.assertEquals("blacklist", hosts, cluster.getBlacklistedHosts());
+        bs.schedule(topologies, cluster);
+        hosts.clear();
+        Assert.assertEquals("blacklist", hosts, cluster.getBlacklistedHosts());
```
HeartSaVioR: Remove above line (`hosts.clear()`) and inline & replace hosts with `Collections.emptySet()`.

storm-core/test/jvm/org/apache/storm/scheduler/blacklist/TestBlacklistScheduler.java
```diff
@@ -0,0 +1,326 @@
...
+        BlacklistScheduler bs = new BlacklistScheduler(new DefaultScheduler());
+        bs.prepare(config);
+        bs.schedule(topologies, cluster);
+        cluster = new Cluster(iNimbus, TestUtilsForBlacklistScheduler.removeSupervisorFromSupervisors(supMap, "sup-0"), TestUtilsForBlacklistScheduler.assignmentMapToImpl(cluster.getAssignments()), config);
+        bs.schedule(topologies, cluster);
+        cluster = new Cluster(iNimbus, TestUtilsForBlacklistScheduler.removeSupervisorFromSupervisors(supMap, "sup-0"), TestUtilsForBlacklistScheduler.assignmentMapToImpl(cluster.getAssignments()), config);
+        bs.schedule(topologies, cluster);
+        cluster = new Cluster(iNimbus, supMap, new HashMap<String, SchedulerAssignmentImpl>(), config);
+        bs.schedule(topologies, cluster);
+        Set<String> hosts = new HashSet<>();
```
HeartSaVioR: Let's make it one-liner : `Set<String> hosts = Collections.singleton("host-0");`

storm-core/test/jvm/org/apache/storm/scheduler/blacklist/TestBlacklistScheduler.java
```diff
@@ -0,0 +1,326 @@
...
+        BlacklistScheduler bs = new BlacklistScheduler(new DefaultScheduler());
+        bs.prepare(config);
+        bs.schedule(topologies, cluster);
+        cluster = new Cluster(iNimbus, TestUtilsForBlacklistScheduler.removeSupervisorFromSupervisors(supMap, "sup-0"), TestUtilsForBlacklistScheduler.assignmentMapToImpl(cluster.getAssignments()), config);
+        bs.schedule(topologies, cluster);
+        cluster = new Cluster(iNimbus, TestUtilsForBlacklistScheduler.removeSupervisorFromSupervisors(supMap, "sup-0"), TestUtilsForBlacklistScheduler.assignmentMapToImpl(cluster.getAssignments()), config);
+        bs.schedule(topologies, cluster);
+        cluster = new Cluster(iNimbus, supMap, TestUtilsForBlacklistScheduler.assignmentMapToImpl(cluster.getAssignments()), config);
+        bs.schedule(topologies, cluster);
+        Set<String> hosts = new HashSet<>();
```
HeartSaVioR: Same here: Let's make it one-liner : `Set<String> hosts = Collections.singleton("host-0");`

storm-core/test/jvm/org/apache/storm/scheduler/blacklist/TestBlacklistScheduler.java
```diff
@@ -0,0 +1,326 @@
...
+        hosts.add("host-0");
+        Assert.assertEquals("blacklist", hosts, cluster.getBlacklistedHosts());
+        topoMap.put(topo2.getId(), topo2);
+        topoMap.put(topo3.getId(), topo3);
+        topoMap.put(topo4.getId(), topo4);
+        topologies = new Topologies(topoMap);
+        cluster = new Cluster(iNimbus, supMap, TestUtilsForBlacklistScheduler.assignmentMapToImpl(cluster.getAssignments()), config);
+        bs.schedule(topologies, cluster);
+        hosts.clear();
```
HeartSaVioR: Same here: Remove above line (hosts.clear()) and inline & replace hosts with Collections.emptySet().

storm-core/test/jvm/org/apache/storm/scheduler/blacklist/TestBlacklistScheduler.java
```diff
@@ -0,0 +1,326 @@
...
+        Assert.assertEquals("blacklist", hosts, cluster.getBlacklistedHosts());
+    }
+
+    @Test
+    public void TestList() {
```
HeartSaVioR: This method looks like not easy to read and understand. Could we make this method better? Or at least put more comments for all `if (count == )`.

storm-core/test/jvm/org/apache/storm/scheduler/blacklist/TestBlacklistScheduler.java
```diff
@@ -0,0 +1,326 @@
...
+                Assert.assertEquals("blacklist", hosts, cluster.getBlacklistedHosts());
+            }
+            count++;
+        }
+
+    }
+
+    @Test
+    public void removeLongTimeDisappearFromCache(){
+        INimbus iNimbus=new TestUtilsForBlacklistScheduler.INimbusTest();
```
HeartSaVioR: nit: space between s and =, and = and n.

storm-core/test/jvm/org/apache/storm/scheduler/blacklist/TestBlacklistScheduler.java
```diff
@@ -0,0 +1,326 @@
...
+
+        Map<String, SupervisorDetails> supMap=TestUtilsForBlacklistScheduler.genSupervisors(3,4);
```
HeartSaVioR: nit: space between p and =, and = and T.

storm-core/test/jvm/org/apache/storm/scheduler/blacklist/TestBlacklistScheduler.java
```diff
@@ -0,0 +1,326 @@
...
+        TopologyDetails topo1 = TestUtilsForBlacklistScheduler.getTopology("topo-1", config, 5, 15, 1, 1, currentTime - 2,true);
+        topoMap.put(topo1.getId(), topo1);
+
+        Cluster cluster = new Cluster(iNimbus, supMap, new HashMap<String, SchedulerAssignmentImpl>(), config);
+        Topologies topologies = new Topologies(topoMap);
+        BlacklistScheduler bs=new BlacklistScheduler(new DefaultScheduler());
+        bs.prepare(config);
+        bs.schedule(topologies,cluster);
+        cluster = new Cluster(iNimbus, TestUtilsForBlacklistScheduler.removeSupervisorFromSupervisors(supMap,"sup-0"),TestUtilsForBlacklistScheduler.assignmentMapToImpl(cluster.getAssignments()), config);
+        for(int i=0;i<20;i++){
```
HeartSaVioR: Let's put a comment why we call schedule() exactly 20 times here. (I mean what exactly
the `20` means.)

storm-core/test/jvm/org/apache/storm/scheduler/blacklist/TestBlacklistScheduler.java
```diff
@@ -0,0 +1,326 @@
...
+        config.put(Config.BLACKLIST_SCHEDULER_TOLERANCE_COUNT,2);
+        config.put(Config.BLACKLIST_SCHEDULER_RESUME_TIME,300);
+        Map<String, TopologyDetails> topoMap = new HashMap<String, TopologyDetails>();
+
+        TopologyDetails topo1 = TestUtilsForBlacklistScheduler.getTopology("topo-1", config, 5, 15, 1, 1, currentTime - 2,true);
+        topoMap.put(topo1.getId(), topo1);
+
+        Cluster cluster = new Cluster(iNimbus, supMap, new HashMap<String, SchedulerAssignmentImpl>(), config);
+        Topologies topologies = new Topologies(topoMap);
+        BlacklistScheduler bs=new BlacklistScheduler(new DefaultScheduler());
```
HeartSaVioR: nit: space between s and =, and = and n.

storm-core/test/jvm/org/apache/storm/scheduler/blacklist/TestBlacklistScheduler.java
```diff
@@ -0,0 +1,326 @@
...
+        bs.prepare(config);
+        bs.schedule(topologies,cluster);
+        cluster = new Cluster(iNimbus, TestUtilsForBlacklistScheduler.removeSupervisorFromSupervisors(supMap,"sup-0"),TestUtilsForBlacklistScheduler.assignmentMapToImpl(cluster.getAssignments()), config);
+        for(int i=0;i<20;i++){
+            bs.schedule(topologies,cluster);
+        }
+        Set<String> cached=new HashSet<>();
```
HeartSaVioR: nit: space between d and =, and = and n.

storm-core/test/jvm/org/apache/storm/scheduler/blacklist/TestBlacklistScheduler.java
```diff
@@ -0,0 +1,326 @@
...
+        cached.add("sup-1");
+        cached.add("sup-2");
+        Assert.assertEquals(cached,bs.cachedSupervisors.keySet());
+        cluster = new Cluster(iNimbus, supMap, new HashMap<String, SchedulerAssignmentImpl>(), config);
+        bs.schedule(topologies,cluster);
+        cluster = new Cluster(iNimbus, TestUtilsForBlacklistScheduler.removePortFromSupervisors(supMap,"sup-0",0),TestUtilsForBlacklistScheduler.assignmentMapToImpl(cluster.getAssignments()), config);
+        for(int i=0;i<20;i++){
+            bs.schedule(topologies,cluster);
+        }
+        Set<Integer> cachedPorts=new HashSet<>();
```
HeartSaVioR: nit: space between s and =, and = and n.

You could use `Sets.newHashSet(1, 2, 3);` given that we depend on Guava.

storm-core/src/jvm/org/apache/storm/scheduler/blacklist/BlacklistScheduler.java
```diff
@@ -0,0 +1,250 @@
...
+    public BlacklistScheduler(IScheduler underlyingScheduler) {
+        this.underlyingScheduler = underlyingScheduler;
+    }
+
+    @Override
+    public void prepare(Map conf) {
+        LOG.info("prepare black list scheduler");
+        underlyingScheduler.prepare(conf);
+        _conf = conf;
+        if (_conf.containsKey(Config.BLACKLIST_SCHEDULER_TOLERANCE_TIME)) {
```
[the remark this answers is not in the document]
revans2: `Utils.getInt` also works with a default value.

HeartSaVioR:
I've crafted rebased version of this patch to #2343. I've pinged to @nilday for some period but didn't receive feedback so I thought it's OK to take it over with preserving the credit.

PR #1676, opened 2016-09-06
STORM-2085: Remove guava from storm-core pom.

knusbaum:
guava isn't used by storm-core.

### DIFF
--- a/storm-core/pom.xml
+++ b/storm-core/pom.xml
@@ -252,11 +252,6 @@
             <scope>compile</scope>
         </dependency>
         <dependency>
-            <groupId>com.google.guava</groupId>
-            <artifactId>guava</artifactId>
-            <scope>compile</scope>
-        </dependency>
-        <dependency>
             <groupId>org.apache.logging.log4j</groupId>
             <artifactId>log4j-api</artifactId>
         </dependency>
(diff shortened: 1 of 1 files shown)


knusbaum:
Do we want this on any of the 1.* branches?

HeartSaVioR:
We shaded Guava which downstream shouldn't rely on, so I think we can apply this to 1.x if applicable.
(I'm not sure we also want to apply this to 1.0.x.)

ps. Since external modules use Guava so we still maintain the main version of Guava.

HeartSaVioR:
I'm +1 on this.
@revans2 @ptgoetz Could you review this as well? This is a change of dependency but removing shaded dependency. Which version line do you think we can apply?

revans2:
I'm +1 on this, but I am a little nervous I may have started using guava as part of the supervisor V2.  Let me try and run some tests.

revans2:
Nope I was wrong it compiles just fine and there is no guava except being pulled in by curator test.

Yes +1 for this, and +1 for pulling it into 1.x if possible.

revans2:
No wait, dependency:analyze indicates that guava is used?  Not sure how it compiled though. Investigating.

revans2:
Ya the supervisor does not come up with this patch.

knusbaum:
Further inspection reveals that guava *is* a dependency. I'm closing this.

PR #1677, opened 2016-09-07
STORM-1985: Admin Commands

kamleshbhatt:


revans2:
For reference the original code that this is replacing was
```
(defn cleanup-corrupt-topologies! [nimbus]
  (let [storm-cluster-state (:storm-cluster-state nimbus)
        blob-store (:blob-store nimbus)
        code-ids (set (code-ids blob-store))
        active-topologies (set (.active-storms storm-cluster-state))
        corrupt-topologies (set/difference active-topologies code-ids)]
    (doseq [corrupt corrupt-topologies]
      (log-message "Corrupt topology " corrupt " has state on zookeeper but doesn't have a local dir on Nimbus. Cleaning up...")
      (.remove-storm! storm-cluster-state corrupt)
      (if (instance? LocalFsBlobStore blob-store)
        (doseq [blob-key (get-key-list-from-id (:conf nimbus) corrupt)]
          (.remove-blobstore-key! storm-cluster-state blob-key))))))
```

storm-core/src/jvm/org/apache/storm/command/AdminCommands.java
```diff
@@ -0,0 +1,114 @@
...
+                break;
+            default:
+                throw new RuntimeException("" + command + " is not a supported admin command");
+        }
+
+    }
+
+    private static void initialize() {
+        Map conf = ConfigUtils.readStormConfig();
+        ClientBlobStore clientBlobStore = Utils.getClientBlobStore(conf);
```
revans2: The original code produced a nimbus blob store, not a client blob store.
```
blob-store (Utils/getNimbusBlobStore conf (NimbusInfo/fromConf conf))
```

The nimbus blob store gives you super user privlages and when nimbus is down it will still work.

storm-core/src/jvm/org/apache/storm/command/AdminCommands.java
```diff
@@ -0,0 +1,114 @@
...
+        List<String> servers = (List<String>) conf.get(Config.STORM_ZOOKEEPER_SERVERS);
+        Object port = conf.get(Config.STORM_ZOOKEEPER_PORT);
+        List<ACL> acls = null;
+        if (Utils.isZkAuthenticationConfiguredStormServer(conf)) {
+            acls = adminZkAcls();
+        }
+        try {
+            IStormClusterState stormClusterState = ClusterUtils.mkStormClusterState(conf, acls, new ClusterStateContext(DaemonType.UNKNOWN));
```
revans2: This should be `DaemonType.NIMBUS`  We are pretending to be nimbus.  For now it should not make much of a difference, but in the future it might.

storm-core/src/jvm/org/apache/storm/command/AdminCommands.java
```diff
@@ -0,0 +1,114 @@
...
+        final List<ACL> acls = new ArrayList<>();
+        acls.add(ZooDefs.Ids.CREATOR_ALL_ACL.get(0));
+        acls.add(new ACL((ZooDefs.Perms.READ ^ ZooDefs.Perms.CREATE), ZooDefs.Ids.ANYONE_ID_UNSAFE));
+        return acls;
+    }
+
+    private static void removeCorruptTopologies( ) {
+        Iterator<String> corruptTopologies = listCorruptTopologies();
+        while(corruptTopologies.hasNext()) {
+            stormClusterState.removeStorm(corruptTopologies.next());
```
revans2: In the original code we also deleted the blobs.

```
      (if (instance? LocalFsBlobStore blob-store)
        (doseq [blob-key (get-key-list-from-id (:conf nimbus) corrupt)]
          (.remove-blobstore-key! storm-cluster-state blob-key))))))
```

storm-core/src/jvm/org/apache/storm/command/AdminCommands.java
```diff
@@ -0,0 +1,114 @@
...
+        }
+    }
+
+    private static Iterator<String> listCorruptTopologies() {
+        Iterator<String> blobStoreTopologyIds = clientBlobStore.listKeys();
```
revans2: This is very different from the code-ids function
```
(defn code-ids [blob-store]
  (let [to-id (reify KeyFilter
                (filter [this key] (ConfigUtils/getIdFromBlobKey key)))]
(set (.filterAndListKeys blob-store to-id))))
```

This lists all of the keys in the blob store.  code-ids Converts the key into a topology id.

storm-core/src/jvm/org/apache/storm/command/AdminCommands.java
```diff
@@ -0,0 +1,114 @@
...
+        Set<String> activeTopologyIds = new HashSet<>(Zookeeper.getChildren(zk, conf.get(Config.STORM_ZOOKEEPER_ROOT) + ClusterUtils.STORMS_SUBTREE, false));
```
revans2: Lets not go to Zookeeper directly this is removing the IStormClusterState abstraction.

```
        Set<String> activeTopologyIds = new HashSet<>(stormClusterState.activeStorms());
```

storm-core/src/jvm/org/apache/storm/command/AdminCommands.java
```diff
@@ -0,0 +1,114 @@
...
+        HashSet<String> blobTopologyIds = Sets.newHashSet(blobStoreTopologyIds);
+        Sets.SetView<String> diffTopology = Sets.difference(activeTopologyIds, blobTopologyIds);
+        LOG.info("active-topology-ids [{}] blob-topology-ids [{}] diff-topology [{}]",
+                generateJoinedString(activeTopologyIds), generateJoinedString(blobTopologyIds),
+                generateJoinedString(diffTopology));
+        return diffTopology.iterator();
+    }
+
+    private static String generateJoinedString(Set<String> activeTopologyIds) {
```
revans2: I'm not sure we need this.  toString of most sets already return this information.

storm-core/src/jvm/org/apache/storm/command/AdminCommands.java
```diff
@@ -0,0 +1,114 @@
...
+
+import java.util.*;
+
+public class AdminCommands {
+
+    private static final Logger LOG = LoggerFactory.getLogger(Deactivate.class);
+    private static ClientBlobStore clientBlobStore;
+    private static IStormClusterState stormClusterState;
+    private static CuratorFramework zk;
+    private static Map conf;
```
revans2: I don't really like static global values.  For now this is fine, but if we add in more commands this could become a real mess, and I think it would be a lot cleaner to have an inner class that we instantiate to preform the operation for different sub commands.

revans2:
I am done going through the code.  Beyond this we need to update storm.py to add in this command, and documentation on how to use it.

kamleshbhatt:
Applied the review comments. Please take a look.

revans2:
Looks good to me.  I have a few tests that I want to run, but in general it looks good.

bin/storm.py
```diff
@@ -572,6 +572,22 @@ def kill_workers(*args):
         jvmtype="-client",
         extrajars=[USER_CONF_DIR, os.path.join(STORM_DIR, "bin")])
 
+def admin_command(*args):
```
revans2: can we rename this just admin, instead of admin_command?

bin/storm.py
```diff
@@ -572,6 +572,22 @@ def kill_workers(*args):
...
+    """Syntax: [storm admin_command cmd]
+
+    This is a proxy of nimbus and allow to execute admin commands. As of now it supports
+    command to remove corrupt topologies.
+    Nimbus doesn't clean up corrupted topologies automatically. This command should clean
+    up corrupt topologies i.e.topologies whose codes are not available on blobstore.
+    In future this command would support more admin commnds.
+    """
```
revans2: Can we list the sub commands here?

storm-core/src/jvm/org/apache/storm/command/AdminCommands.java
```diff
@@ -0,0 +1,133 @@
...
+            throw new IllegalArgumentException("Missing command.");
+        }
+        initialize();
+        String command = args[0];
+        String[] newArgs = Arrays.copyOfRange(args, 1, args.length);
+        switch (command) {
+            case "remove_corrupt_topologies":
+                removeCorruptTopologies();
+                break;
+            default:
```
revans2: Can we list the sub commands here before throwing the exception.

revans2: Also the Exception itself is surrounded by lots and lots of log messages that make it hard to see what is happening or why a command failed.  Might be nice to print some things out with white space around them to help distinguish what is happening.

revans2: Or not initialize until you know which command you are going to run.

storm-core/src/jvm/org/apache/storm/command/AdminCommands.java
```diff
@@ -0,0 +1,133 @@
...
+                Iterator<String> blobKeys = getKeyListFromId(corruptId).iterator();
+                while(blobKeys.hasNext()) {
+                    stormClusterState.removeBlobstoreKey(blobKeys.next());
+                }
+            }
+        }
+    }
+
+    private static Iterator<String> listCorruptTopologies() {
+        Set<String> blobStoreTopologyIds = nimbusBlobStore.filterAndListKeys(new KeyFilter<String>() {
```
revans2: I am getting an NPE here when I run this.  Not totally sure why.

bin/storm.py
kamleshbhatt: Done but admin might seems to be too generic.

bin/storm.py
kamleshbhatt: Done.

storm-core/src/jvm/org/apache/storm/command/AdminCommands.java
kamleshbhatt: Done. We might have to move initialize method again out of switch block because when we have more commands, repeating it again and again might not be a good idea.

storm-core/src/jvm/org/apache/storm/command/AdminCommands.java
kamleshbhatt: I am unable to test this out. Could you please provide some pointers on how can I setup a  test environment to investigate this issue.

revans2: What I did
```
mvn clean install -DskipTests
cd storm-dist/binary/
mvn clean package
tar -xvzf ./target/apache-storm-2.0.0-SNAPSHOT.tar.gz
cd apache-storm/
./bin/storm dev-zookeeper &
./bin/storm nimbus &
./bin/storm supervisor &
```
Wait for the processes to come up and then we can start testing
```
./bin/storm admin_command remove_corrupt_topologies
```

kamleshbhatt: Thanks for providing these steps. I fixed the issue and updated the pull request for review. Please take a look. 
However I had to change storm.cmd and storm-config.cmd as well to make it working. Please review these changes as well.

revans2:
@kamleshbhatt sorry I lost track of this, looking again now.

revans2:
+1 please upmerge though.  There is a very minor merge conflict.  Then ping me and I'll merge it in.

PR #1679, opened 2016-09-11
STORM-2087: storm-kafka-client - tuples not always being replayed

jfenc91:
I am having issues with topologies that fail tuples. The kafka offsets seemed to get stuck and the kafka spout eventually halted even though the last committed offset was no where near the end of the topic. 

Here are a few unit tests that I believe replicate my situation. The last 2 are currently failing which is reflective of what I am seeing in my topologies. Let me know if I missed anything! This seems like a pretty big oversight, so I am getting the feeling that something in the test is wrong. Thanks!

I added tests for the following cases:
- All tuples being acked.
- A tuple being failed in order it was emitted
- A tuple being failed out of the order it was emitted

Update: Also Provided a fix now as well.

### DIFF
--- a/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
+++ b/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
@@ -77,7 +77,7 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
     private KafkaSpoutStreams kafkaSpoutStreams;                        // Object that wraps all the logic to declare output fields and emit tuples
     private transient KafkaSpoutTuplesBuilder<K, V> tuplesBuilder;      // Object that contains the logic to build tuples for each ConsumerRecord
 
-    private transient Map<TopicPartition, OffsetEntry> acked;           // Tuples that were successfully acked. These tuples will be committed periodically when the commit timer expires, after consumer rebalance, or on close/deactivate
+    transient Map<TopicPartition, OffsetEntry> acked;           // Tuples that were successfully acked. These tuples will be committed periodically when the commit timer expires, after consumer rebalance, or on close/deactivate
     private transient Set<KafkaSpoutMessageId> emitted;                 // Tuples that have been emitted but that are "on the wire", i.e. pending being acked or failed
     private transient Iterator<ConsumerRecord<K, V>> waitingToEmit;         // Records that have been polled and are queued to be emitted in the nextTuple() call. One record is emitted per nextTuple()
     private transient long numUncommittedOffsets;                       // Number of offsets that have been polled and emitted but not yet been committed
@@ -266,19 +266,22 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
             if (offsetAndMeta != null) {
                 kafkaConsumer.seek(rtp, offsetAndMeta.offset() + 1);  // seek to the next offset that is ready to commit in next commit cycle
             } else {
-                kafkaConsumer.seekToEnd(toArrayList(rtp));    // Seek to last committed offset
+                kafkaConsumer.seek(rtp, acked.get(rtp).committedOffset + 1);    // Seek to last committed offset
             }
         }
     }
 
     // ======== emit  =========
     private void emit() {
-        emitTupleIfNotEmitted(waitingToEmit.next());
-        waitingToEmit.remove();
+        while(!emitTupleIfNotEmitted(waitingToEmit.next()) && waitingToEmit.hasNext()) {
+            waitingToEmit.remove();
+        }
     }
 
-    // emits one tuple per record
-    private void emitTupleIfNotEmitted(ConsumerRecord<K, V> record) {
+
+    //Emits one tuple per record
+    //@return true if tuple was emitted
+    private boolean emitTupleIfNotEmitted(ConsumerRecord<K, V> record) {
         final TopicPartition tp = new TopicPartition(record.topic(), record.partition());
         final KafkaSpoutMessageId msgId = new KafkaSpoutMessageId(record);
 
@@ -295,7 +298,9 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
                 retryService.remove(msgId);  // re-emitted hence remove from failed
             }
             LOG.trace("Emitted tuple [{}] for record [{}]", tuple, record);
+            return true;
         }
+        return false;
     }
 
     private void commitOffsetsForAckedTuples() {
@@ -451,7 +456,7 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
     /**
      * This class is not thread safe
      */
-    private class OffsetEntry {
+    class OffsetEntry {
         private final TopicPartition tp;
         private final long initialFetchOffset;  /* First offset to be fetched. It is either set to the beginning, end, or to the first uncommitted offset.
                                                  * Initial value depends on offset strategy. See KafkaSpoutConsumerRebalanceListener */
@@ -479,7 +484,7 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
             KafkaSpoutMessageId nextCommitMsg = null;     // this is a convenience variable to make it faster to create OffsetAndMetadata
 
             for (KafkaSpoutMessageId currAckedMsg : ackedMsgs) {  // complexity is that of a linear scan on a TreeMap
-                if ((currOffset = currAckedMsg.offset()) == initialFetchOffset || currOffset == nextCommitOffset + 1) {            // found the next offset to commit
+                if ((currOffset = currAckedMsg.offset()) == nextCommitOffset + 1) {            // found the next offset to commit
                     found = true;
                     nextCommitMsg = currAckedMsg;
                     nextCommitOffset = currOffset;
@@ -487,8 +492,9 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
                     LOG.debug("topic-partition [{}] has non-continuous offset [{}]. It will be processed in a subsequent batch.", tp, currOffset);
                     break;
                 } else {
-                    LOG.debug("topic-partition [{}] has unexpected offset [{}].", tp, currOffset);
-                    break;
+                    //Received a redundant ack. Ignore and continue processing.
+                    LOG.warn("topic-partition [{}] has unexpected offset [{}]. Current committed Offset [{}]",
+                            tp, currOffset,  committedOffset);
                 }
             }
 
@@ -530,6 +536,10 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
                         numCommittedOffsets, tp, numUncommittedOffsets);
             }
             LOG.trace("{}", this);
+        }
+
+        long getCommittedOffset() {
+            return committedOffset;
         }
 
         public boolean isEmpty() {
--- a/external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/SingleTopicKafkaSpoutTest.java
+++ b/external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/SingleTopicKafkaSpoutTest.java
@@ -1,0 +1,240 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ *   or more contributor license agreements.  See the NOTICE file
+ *   distributed with this work for additional information
+ *   regarding copyright ownership.  The ASF licenses this file
+ *   to you under the Apache License, Version 2.0 (the
+ *   "License"); you may not use this file except in compliance
+ *   with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+package org.apache.storm.kafka.spout;
+
+import info.batey.kafka.unit.KafkaUnitRule;
+import kafka.producer.KeyedMessage;
+import org.apache.kafka.clients.consumer.OffsetAndMetadata;
+import org.apache.storm.kafka.spout.builders.SingleTopicKafkaSpoutConfiguration;
+import org.apache.storm.spout.SpoutOutputCollector;
+import org.apache.storm.task.TopologyContext;
+import org.apache.storm.tuple.Values;
+import org.junit.Rule;
+import org.junit.Test;
+import org.mockito.ArgumentCaptor;
+
+import static org.junit.Assert.*;
+
+import java.util.Map;
+import java.util.stream.IntStream;
+import static org.mockito.Mockito.*;
+import static org.apache.storm.kafka.spout.builders.SingleTopicKafkaSpoutConfiguration.*;
+
+public class SingleTopicKafkaSpoutTest {
+
+    private class SpoutContext {
+        public KafkaSpout<String, String> spout;
+        public SpoutOutputCollector collector;
+
+        public SpoutContext(KafkaSpout<String, String> spout,
+                            SpoutOutputCollector collector) {
+            this.spout = spout;
+            this.collector = collector;
+        }
+    }
+
+    @Rule
+    public KafkaUnitRule kafkaUnitRule = new KafkaUnitRule();
+
+    void populateTopicData(String topicName, int msgCount) {
+        kafkaUnitRule.getKafkaUnit().createTopic(topicName);
+
+        IntStream.range(0, msgCount).forEach(value -> {
+            KeyedMessage<String, String> keyedMessage = new KeyedMessage<>(
+                    topicName, Integer.toString(value),
+                    Integer.toString(value));
+
+            kafkaUnitRule.getKafkaUnit().sendMessages(keyedMessage);
+        });
+    }
+
+    SpoutContext initializeSpout(int msgCount) {
+        populateTopicData(SingleTopicKafkaSpoutConfiguration.TOPIC, msgCount);
+        int kafkaPort = kafkaUnitRule.getKafkaPort();
+
+        TopologyContext topology = mock(TopologyContext.class);
+        SpoutOutputCollector collector = mock(SpoutOutputCollector.class);
+        Map conf = mock(Map.class);
+
+        KafkaSpout<String, String> spout = new KafkaSpout<>(getKafkaSpoutConfig(getKafkaSpoutStreams(), kafkaPort));
+        spout.open(conf, topology, collector);
+        spout.activate();
+        return new SpoutContext(spout, collector);
+    }
+    /*
+     * Asserts that the next possible offset to commit or the committed offset is the provided offset.
+     * An offset that is ready to be committed is not guarenteed to be already committed.
+     */
+    private void assertOffsetCommitted(int offset, KafkaSpout.OffsetEntry entry) {
+
+        boolean currentOffsetMatch = entry.getCommittedOffset() == offset;
+        OffsetAndMetadata nextOffset = entry.findNextCommitOffset();
+        boolean nextOffsetMatch =  nextOffset != null && nextOffset.offset() == offset;
+        assertTrue("Next offset: " +
+                        entry.findNextCommitOffset() +
+                        " OR current offset: " +
+                        entry.getCommittedOffset() +
+                        " must equal desired offset: " +
+                        offset,
+                currentOffsetMatch | nextOffsetMatch);
+    }
+
+    @Test
+    public void shouldContinueWithSlowDoubleAcks() throws Exception {
+        int messageCount = 20;
+        SpoutContext context = initializeSpout(messageCount);
+
+        //play 1st tuple
+        ArgumentCaptor<Object> messageIdToDoubleAck = ArgumentCaptor.forClass(Object.class);
+        context.spout.nextTuple();
+        verify(context.collector).emit(anyObject(), anyObject(), messageIdToDoubleAck.capture());
+        context.spout.ack(messageIdToDoubleAck.getValue());
+
+        IntStream.range(0, messageCount/2).forEach(value -> {
+            context.spout.nextTuple();
+        });
+
+        context.spout.ack(messageIdToDoubleAck.getValue());
+
+        IntStream.range(0, messageCount).forEach(value -> {
+            context.spout.nextTuple();
+        });
+
+        ArgumentCaptor<Object> remainingIds = ArgumentCaptor.forClass(Object.class);
+
+        verify(context.collector, times(messageCount)).emit(
+                eq(SingleTopicKafkaSpoutConfiguration.STREAM),
+                anyObject(),
+                remainingIds.capture());
+        remainingIds.getAllValues().iterator().forEachRemaining(context.spout::ack);
+
+        context.spout.acked.values().forEach(item -> {
+            assertOffsetCommitted(messageCount - 1, (KafkaSpout.OffsetEntry) item);
+        });
+    }
+
+    @Test
+    public void shouldEmitAllMessages() throws Exception {
+        int messageCount = 10;
+        SpoutContext context = initializeSpout(messageCount);
+
+
+        IntStream.range(0, messageCount).forEach(value -> {
+            context.spout.nextTuple();
+            ArgumentCaptor<Object> messageId = ArgumentCaptor.forClass(Object.class);
+            verify(context.collector).emit(
+                    eq(SingleTopicKafkaSpoutConfiguration.STREAM),
+                    eq(new Values(SingleTopicKafkaSpoutConfiguration.TOPIC,
+                            Integer.toString(value),
+                            Integer.toString(value))),
+            messageId.capture());
+            context.spout.ack(messageId.getValue());
+            reset(context.collector);
+        });
+
+        context.spout.acked.values().forEach(item -> {
+            assertOffsetCommitted(messageCount - 1, (KafkaSpout.OffsetEntry) item);
+        });
+    }
+
+    @Test
+    public void shouldReplayInOrderFailedMessages() throws Exception {
+        int messageCount = 10;
+        SpoutContext context = initializeSpout(messageCount);
+
+        //play and ack 1 tuple
+        ArgumentCaptor<Object> messageIdAcked = ArgumentCaptor.forClass(Object.class);
+        context.spout.nextTuple();
+        verify(context.collector).emit(anyObject(), anyObject(), messageIdAcked.capture());
+        context.spout.ack(messageIdAcked.getValue());
+        reset(context.collector);
+
+        //play and fail 1 tuple
+        ArgumentCaptor<Object> messageIdFailed = ArgumentCaptor.forClass(Object.class);
+        context.spout.nextTuple();
+        verify(context.collector).emit(anyObject(), anyObject(), messageIdFailed.capture());
+        context.spout.fail(messageIdFailed.getValue());
+        reset(context.collector);
+
+        //pause so that failed tuples will be retried
+        Thread.sleep(200);
+
+
+        //allow for some calls to nextTuple() to fail to emit a tuple
+        IntStream.range(0, messageCount + 5).forEach(value -> {
+            context.spout.nextTuple();
+        });
+
+        ArgumentCaptor<Object> remainingMessageIds = ArgumentCaptor.forClass(Object.class);
+
+        //1 message replayed, messageCount - 2 messages emitted for the first time
+        verify(context.collector, times(messageCount - 1)).emit(
+                eq(SingleTopicKafkaSpoutConfiguration.STREAM),
+                anyObject(),
+                remainingMessageIds.capture());
+        remainingMessageIds.getAllValues().iterator().forEachRemaining(context.spout::ack);
+
+        context.spout.acked.values().forEach(item -> {
+            assertOffsetCommitted(messageCount - 1, (KafkaSpout.OffsetEntry) item);
+        });
+    }
+
+    @Test
+    public void shouldReplayFirstTupleFailedOutOfOrder() throws Exception {
+        int messageCount = 10;
+        SpoutContext context = initializeSpout(messageCount);
+
+
+        //play 1st tuple
+        ArgumentCaptor<Object> messageIdToFail = ArgumentCaptor.forClass(Object.class);
+        context.spout.nextTuple();
+        verify(context.collector).emit(anyObject(), anyObject(), messageIdToFail.capture());
+        reset(context.collector);
+
+        //play 2nd tuple
+        ArgumentCaptor<Object> messageIdToAck = ArgumentCaptor.forClass(Object.class);
+        context.spout.nextTuple();
+        verify(context.collector).emit(anyObject(), anyObject(), messageIdToAck.capture());
+        reset(context.collector);
+
+        //ack 2nd tuple
+        context.spout.ack(messageIdToAck.getValue());
+        //fail 1st tuple
+        context.spout.fail(messageIdToFail.getValue());
+
+        //pause so that failed tuples will be retried
+        Thread.sleep(200);
+
+        //allow for some calls to nextTuple() to fail to emit a tuple
+        IntStream.range(0, messageCount + 5).forEach(value -> {
+            context.spout.nextTuple();
+        });
+
+        ArgumentCaptor<Object> remainingIds = ArgumentCaptor.forClass(Object.class);
+        //1 message replayed, messageCount - 2 messages emitted for the first time
+        verify(context.collector, times(messageCount - 1)).emit(
+                eq(SingleTopicKafkaSpoutConfiguration.STREAM),
+                anyObject(),
+                remainingIds.capture());
+        remainingIds.getAllValues().iterator().forEachRemaining(context.spout::ack);
+
+        context.spout.acked.values().forEach(item -> {
+            assertOffsetCommitted(messageCount - 1, (KafkaSpout.OffsetEntry) item);
+        });
+    }
+}
--- a/external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/builders/SingleTopicKafkaSpoutConfiguration.java
+++ b/external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/builders/SingleTopicKafkaSpoutConfiguration.java
@@ -1,0 +1,85 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ *   or more contributor license agreements.  See the NOTICE file
+ *   distributed with this work for additional information
+ *   regarding copyright ownership.  The ASF licenses this file
+ *   to you under the Apache License, Version 2.0 (the
+ *   "License"); you may not use this file except in compliance
+ *   with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+package org.apache.storm.kafka.spout.builders;
+
+import org.apache.storm.Config;
+import org.apache.storm.generated.StormTopology;
+import org.apache.storm.kafka.spout.*;
+import org.apache.storm.kafka.spout.test.KafkaSpoutTestBolt;
+import org.apache.storm.topology.TopologyBuilder;
+import org.apache.storm.tuple.Fields;
+
+import java.util.HashMap;
+import java.util.Map;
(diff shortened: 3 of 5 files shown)


jfenc91:
A larger refactor here is probably needed to make this more performant. These changes understandably seem to make the spout struggle to get the processing tuple count anywhere near the max spout pending.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -266,26 +266,32 @@ private void doSeekRetriableTopicPartitions() {
             if (offsetAndMeta != null) {
                 kafkaConsumer.seek(rtp, offsetAndMeta.offset() + 1);  // seek to the next offset that is ready to commit in next commit cycle
             } else {
-                kafkaConsumer.seekToEnd(toArrayList(rtp));    // Seek to last committed offset
+                kafkaConsumer.seek(rtp, acked.get(rtp).committedOffset + 1);    // Seek to last committed offset
```
jfenc91: Yup :) . shouldReplayInOrderFailedMessages and shouldReplayOutOfOrderFailedMessages Tests.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -266,26 +266,32 @@ private void doSeekRetriableTopicPartitions() {
...
             }
         }
     }
 
     // ======== emit  =========
     private void emit() {
-        emitTupleIfNotEmitted(waitingToEmit.next());
+        //Keep trying to send a tuple when requested
+        while(!emitTupleIfNotEmitted(waitingToEmit.next()) && waitingToEmit.hasNext())
```
jfenc91: I was having issues with the spout reaching the max spout pending limit. Although it is apparent now that there are other contributing factors. So I will go back to the simpler way here. No need to overcomplicate.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -479,16 +487,17 @@ public OffsetAndMetadata findNextCommitOffset() {
...
                     found = true;
                     nextCommitMsg = currAckedMsg;
                     nextCommitOffset = currOffset;
                 } else if (currAckedMsg.offset() > nextCommitOffset + 1) {    // offset found is not continuous to the offsets listed to go in the next commit, so stop search
                     LOG.debug("topic-partition [{}] has non-continuous offset [{}]. It will be processed in a subsequent batch.", tp, currOffset);
                     break;
                 } else {
-                    LOG.debug("topic-partition [{}] has unexpected offset [{}].", tp, currOffset);
-                    break;
+                    //Received a redundant ack. Ignore and continue processing.
```
jfenc91: So I added this because I was seeing acks on tuples that were behind the already committed offset. With that break statement in place, the result is a complete halt in processing. While this isn't pretty this is the only solution I could see.

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/SingleTopicKafkaSpoutTest.java
```diff
@@ -0,0 +1,248 @@
...
+        spout.nextTuple();
+        verify(collector).emit(anyObject(), anyObject(), messageIdFailed.capture());
+        spout.fail(messageIdFailed.getValue());
+        reset(collector);
+
+        //pause so that failed tuples will be retried
+        Thread.sleep(200);
+
+        //allow for some failures with +5 instead of -1
+        IntStream.range(0, messageCount + 5).forEach(value -> {
```
jfenc91: A call to the spout's emit() does not always mean a tuple is emitted. I will improve the comment.

jfenc91:
Thanks for the review here @srdo! I will have a few more PRs headed your way in a week or two for this kafka client to make it fully usable.

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/SingleTopicKafkaSpoutTest.java
```diff
@@ -0,0 +1,248 @@
...
+    @Rule
+    public KafkaUnitRule kafkaUnitRule = new KafkaUnitRule();
+
+
+    void populateTopicData(String topicName, int msgCount) {
+        kafkaUnitRule.getKafkaUnit().createTopic(topicName);
+
+        IntStream.range(0, msgCount).forEach(value -> {
+            KeyedMessage<String, String> keyedMessage = new KeyedMessage<>(
+                    topicName, ((Integer)value).toString(),
```
jfenc91: value is an int. So I changed the Integer casts to Integer.toString(value) .... that probably looks a bit nicer

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -277,7 +277,10 @@ private void emit() {
         waitingToEmit.remove();
     }
 
-    // emits one tuple per record
+    /**
+     * Emits one tuple per record
+     * @return True if tuple was emitted else false
```
hmcl: This method returns void. What does the @return True comment mean?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -451,11 +454,11 @@ public int compare(KafkaSpoutMessageId m1, KafkaSpoutMessageId m2) {
     /**
      * This class is not thread safe
      */
-    private class OffsetEntry {
+    class OffsetEntry {
         private final TopicPartition tp;
         private final long initialFetchOffset;  /* First offset to be fetched. It is either set to the beginning, end, or to the first uncommitted offset.
                                                  * Initial value depends on offset strategy. See KafkaSpoutConsumerRebalanceListener */
-        private long committedOffset;     // last offset committed to Kafka. Initially it is set to fetchOffset - 1
+        long committedOffset;     // last offset committed to Kafka. Initially it is set to fetchOffset - 1
```
hmcl: It violates encapsulation to make this field non private.

Is this package protected for test purposes? If so, wouldn't it be better to use the testing tool reflection based methods to test for internal state, rather than exposing the state to the outside.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -479,16 +482,17 @@ public OffsetAndMetadata findNextCommitOffset() {
             KafkaSpoutMessageId nextCommitMsg = null;     // this is a convenience variable to make it faster to create OffsetAndMetadata
 
             for (KafkaSpoutMessageId currAckedMsg : ackedMsgs) {  // complexity is that of a linear scan on a TreeMap
-                if ((currOffset = currAckedMsg.offset()) == initialFetchOffset || currOffset == nextCommitOffset + 1) {            // found the next offset to commit
+                if ((currOffset = currAckedMsg.offset()) == nextCommitOffset + 1) {            // found the next offset to commit
```
hmcl: With this change is the code going to work for a topic that does not yet have any records and/or commits in it?

I recall testing the condition using initialFetchOffset was necessary. Why isn't it necessary?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
hmcl: Do you have a test case where we can reproduce this consistently? This code is already  running in a large production environment, and according to the feedback I received, there are no issues of this nature. 

I am also a bit confused on what I mean by multiple acks. I am pretty sure storm guarantees that a tuple it's either acked (once), failed (once), or times out (which is equivalent to failing)

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
srdo: Guessing here, but if no tuples have been acked since partition reassignment and some have failed, this case should be hit. The right thing to do is then to start replaying from the last committed offset. If the consumer just skips to the end of the partition, it may skip some messages, and findNextCommitOffset will break on subsequent calls because the offsets that can be committed aren't continuous anymore.

@jfenc91 The tests you mention ack a message before a tuple is failed. Doesn't that prevent the spout from hitting this case, since there's now an offset in acked?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
srdo: It still seems strange to me if Storm acks a message multiple times. You're right though that the break statement should be removed, if the spout gets into a state where an offset <= committed is acked, the offset should probably just be logged and dropped since it's already committed. I'd still like to know how the spout gets into this state though.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
srdo: nextCommitOffset starts out as initialFetchOffset - 1 due to L467 and L481. Assuming a tuple isn't acked after being committed, the two checks should work out to be the same thing.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
srdo: I think this is left over from a revert.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
srdo: Hm. This probably isn't great if auto commit mode is enabled. It'll make the spout return to where it started processing the partition every time a tuple is failed. It might make sense to disable tuple retries when auto commit is enabled, I don't think we can support it properly in auto commit mode with the current RetryService interface, since we don't update committedOffset in auto commit mode. 

I think having fail() do nothing if auto commit is enabled makes sense. Otherwise, RetryService needs to return the list of offsets that should be retried so this code can seek directly to the smallest retry offset per partition instead.

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/builders/SingleTopicKafkaSpoutConfiguration.java
```diff
@@ -0,0 +1,85 @@
...
+    public static final String stream = "test_stream";
+    public static final String topic = "test";
+
+    static public Config getConfig() {
+        Config config = new Config();
+        config.setDebug(true);
+        return config;
+    }
+
+    static public StormTopology getTopolgyKafkaSpout(int port) {
```
srdo: Nitpick: Spelling error in method name

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/SingleTopicKafkaSpoutTest.java
```diff
@@ -0,0 +1,249 @@
...
+    public void shouldContinueWithSlowDoubleAcks() throws Exception {
+        int messageCount = 20;
+        populateTopicData(SingleTopicKafkaSpoutConfiguration.topic, messageCount);
+        int kafkaPort = kafkaUnitRule.getKafkaPort();
+
+        TopologyContext topology = mock(TopologyContext.class);
+        SpoutOutputCollector collector = mock(SpoutOutputCollector.class);
+        Map conf = mock(Map.class);
+
+        KafkaSpout spout = new KafkaSpout<>(getKafkaSpoutConfig(getKafkaSpoutStreams(), kafkaPort));
```
srdo: Nitpick: Remember to add generic types to the spout

srdo:
@jfenc91 I borrowed a few classes from this PR for another change to the spout here https://github.com/apache/storm/pull/1696. I hope you don't mind :)

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
jfenc91: I saw this a couple of times before I figured it out. I have not taken the time to reproduce this in a toy/test case, but given the error message this is clearly a storm or storm-kafka-client issue. I got to this state in about 30 minutes of running a topology processing 800k-300k tuples a minute with about 10s latency. The input to the topology was on the order of 2k-10k tuples per minute with a bolt that separated each input into multiple tuples. At startup there was a high amount of failures after the separation (I was making requests against an unwarmed ELB). I would guess that that is enough to reproduce with random data/failures.

jfenc91:
Thats fine @srdo . Thanks again for the reviews. I also created STORM-2106 to keep track of the consequences of this change.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
srdo: Another PR is fine, but I don't think there needs to be a big change. If people are running with auto commit mode on they've already kind of decided that they don't care about at least once delivery (similar to disabling tuple acking), so I think simply making the spout ignore failed tuples should be fine. Alternatively auto commit mode could be removed from KafkaSpoutConfig and just be an optimization the spout enables if tuple acking is disabled. I don't think it makes sense to run with tuple acking on and auto commit on, nor does it really make sense to disable tuple acking and disable auto commit. @hmcl do you have an opinion on this?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
srdo: I'm sorry to flip-flop on this, but after our discussion on STORM-2106, I think it makes sense to have this loop. Storm will by default pause for 1ms between each call to nextTuple if nothing is emitted. This could end up being a lot of pausing if there's a failed tuple far behind the latest acked offset. Having this loop prevents that pause from happening unless there really isn't anything to emit.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
hmcl: @srdo from what I recall, in scenarios where topic had no messages at all, this was necessary. I tuned some of these changes based on several integration tests I ran under several scenarios. As I mentioned in a prior comment, I have been in touch with people that are running this in production without any issues. You are proposing quite a few algorithmic changes. Are these changes caused by bugs ? Do you have test scenarios that guarantee that the proposed changes are not adding new bugs ? Or are these changes proposed by code analysis that leads you to believe that they are not necessary ?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -277,7 +277,9 @@ private void emit() {
         waitingToEmit.remove();
     }
 
-    // emits one tuple per record
+    /**
+     * Emits one tuple per record
+     */
```
hmcl: This is a private method so it should be a standard line comment rather than a javadoc.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
srdo: @jfenc91 About this being tested: If the two tests you mentioned run fine without this change, would you mind adding one that tests this? Maybe something like have a topic with a few messages in it and fail the first message, then check that the spout doesn't skip the messages following the failed message?

jfenc91: Not a problem. Performance/throughput is very important to me so I am happy to go back to this!

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
jfenc91: Not a problem. Performance/throughput is very important to me so I am happy to go back to this!

jfenc91:
@srdo Alright I think everything has been addressed here. I have actually been running this merged with your other PR for the last 12 hours processing 100M tuples and its looked pretty good. Only issue has been: "org.apache.kafka.clients.consumer.CommitFailedException: Commit cannot be completed since the group has already rebalanced and assigned the partitions to another member." which is unrelated to this PR.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
srdo: Looks good, thanks.

srdo:
This more or less seems done to me. The only thing that bugs me is Storm double acking tuples. The only case I could think of is if tuples time out and they're later acked by the acker bolt, but it seems like the executor actually handles that. As far as I can tell hitting the changed case in findNextCommitOffset should actually indicate a bug in the code, because we've received an ack for a tuple we've already committed to Kafka, which shouldn't be possible unless a tuple is double acked. I don't think that should be happening unless the spout is doing something wrong?

@jfenc91 are you still seeing that case (L498) getting hit even with these changes and the other PR?

hmcl:
@srdo @jfenc91 I am on vacation this week (with limited access to Internet) and I will be back on Monday. **Can we please holding on merging this until I can finish my review on Monday**. I implemented the original patch, and would like to review these changes.Thanks.

I am still a bit confused about the duplicated acking tuples. If that is the case, that could be a possible storm guarantees problem, which should be fixed at its root cause, and not worked around in spout code. Furthermore, from what I understand, as far as the findNextOffset method behavior, the duplicate acking should be irrelevant, because once all the tuples up to a certain offset have been acked, they will be committed in the next commit call. For example let's say `1,2,3,4,5, 8,9,10` have acked, on the next commit call, `1,2,3,4,5` should be committed, but `8,9,10` won't be until `6,7` get acked. If `6,7` never get acked, say (retry forever scenario), then the desired behavior is that `8,9,10` never get committed. @jfenc91 Can you help me understand how does the duplicate acking interfere in this scenario, and how?

srdo:
@hmcl I agree that if there really is a double acking problem somewhere else, it should be fixed there.

In your scenario say the spout commits 1...5 to Kafka and 4 is later acked. ackedMsgs now contains 4, and committedOffset is 5. With the previous code, that would permanently break findNextCommitOffset. 

findNextCommitOffset would start at 5, and the first message in ackedMsgs list is 4. The code breaks out of the loop over ackedMsgs, since the else block in L496 is hit, causing the function to return null. When null is returned to commitOffsetsForAckedTuples, that means that it will skip that topic partition when committing, which means that ackedMsgs doesn't get cleared. That means that on the next call to findNextCommitOffset the same thing happens again. The end result is that the spout never commits anything for that topic partition again.

I agree that we should never be in the case where 4 is acked after committedOffset has become larger than 4, which is why I think the else block in L496 should probably log at a higher level than debug.

jfenc91:
@srdo @hmcl I tracked down one source of the "double acking" 

It looks like I am being switched to reading from an out of sync replica. Looking at my offsets:
GROUP                          TOPIC                          PARTITION  CURRENT-OFFSET  LOG-END-OFFSET  LAG             OWNER
 bla_5_2           bla_5             102        237919          50979           -186940         consumer-7_/123.456.789.101

When examining with my other partition logs there is no way that this log-end-offset should be at 50k. It should be around 200k-300k. When the connector sends the request for offset 237920 it is out of range and the resets the offset to whatever the start offset is: https://github.com/apache/kafka/blob/0.10.0.1/clients/src/main/java/org/apache/kafka/clients/consumer/internals/Fetcher.java#L580-L589 . For me this happened to be uncomitted_earliest so I got messages starting at position 1. 

This is a really unfortunate situation. We should probably make an effort to handle it better so data loss / disruption is minimized. I kind of wonder if this ever happened with the kafka 8.

srdo:
@jfenc91 Ouch. I've never seen LEO drop like that unless unclean leader election was enabled. Out of curiosity which Kafka version are you running, and do you have unclean leader election enabled for this topic? If not, consider trying the kafka mailing list. I don't think the spout can or should handle the Kafka log suddenly getting truncated like that, since it's really a data loss scenario on the Kafka side. I can't really think of any way the spout could handle it other than resetting to earliest/latest, which it already does.

If the culprit is unclean leader election then it would have also happened in Kafka 0.8.

jfenc91:
I am on version 0.10.0.1. Interesting, didn't realize unclean leader election was a default... So ya I had that enabled. Turning that off now though! Thanks @srdo for the tip and hopefully that will make this problem go away!

srdo:
@jfenc91 Hope it works for you now. You're right though that the spout needs to handle unclean leader elections better. It's not great that the consumer group will remain "ahead" of the LEO and the spout will fail to commit offsets until it has caught up to where the unclean leader election happened. I think an option for a solution would be to have the spout check if the current LEO is behind committedOffset when it hits the else case in findNextCommitOffsets, and if so, it should reset the consumer group's committed offset to the either the LEO or the beginning of the partition, probably depending on a new configuration parameter. It's never going to be a nice situation to be in, but at least the spout shouldn't break when it happens, especially since it can happen with the default Kafka configuration.

srdo:
Thought about it a little more. The spout should probably just check when it polls messages that the messages are not behind the committedOffset. If they are, the committedOffset needs to move to the left.

It's a little irritating that the consumer doesn't expose an API to let the client know that an unclean leader election happened. It would be nice if for example ConsumerRebalanceListener had a method to tell the client to roll back the committed offset to the new LEO after an unclean leader was elected.

jfenc91:
I bumped the logging level for unexpected offsets to Warn..... Anyway can we get this and the other PR merged @srdo @hmcl? Admittedly there is still work that needs to be done, but this PR is pretty messy at this point and the changes here (and in the other PR) makes the kafka spout mostly usable for version 10.

srdo:
This PR LGTM at this point. I think we should raise a separate issue for better unclean leader election support, when that gets implemented the double acking test should probably be replaced with one that forces an unclean leader election between two test brokers.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -266,26 +266,31 @@ private void doSeekRetriableTopicPartitions() {
...
+
+    //Emits one tuple per record
+    //@return true if tuple was emitted
+    private boolean emitTupleIfNotEmitted(ConsumerRecord<K, V> record) {
         final TopicPartition tp = new TopicPartition(record.topic(), record.partition());
         final KafkaSpoutMessageId msgId = new KafkaSpoutMessageId(record);
 
         if (acked.containsKey(tp) && acked.get(tp).contains(msgId)) {   // has been acked
             LOG.trace("Tuple for record [{}] has already been acked. Skipping", record);
+            return false;
```
harshach: why are we adding returns here. Having one return makes the method look cleaner. This is an if else conditions and we are returning false at the end.

jfenc91: I removed the return statements @harshach

jfenc91:
Hey, trying not to let this thread die. Could we get some help here from a committer? @HeartSaVioR @revans2 Thanks!!

revans2:
@srdo @hmcl have all of your concerns been addressed?

srdo:
Yes. I think supporting unclean leader election in a nice way is a separate issue to this.

revans2:
I looked through the code and I didn't see anything that concerned me, and the travis test failure appears to be unrelated so I am +1 pending feedback form @hmcl 

@harshach you also reviewed this do you have any concerns still?

hmcl:
@revans2 I will finish my review by lunch time PST

revans2:
@hmcl take your time.  This is a critical piece of storm code and I really would like it right.

revans2:
@hmcl it has been a week, any hope of finishing the review?

revans2:
@knusbaum I noticed that you opened up #1753 and #1752.  Would you mind taking a look at this and see if there are going to be any conflicts?

hmcl:
@revans2 apologies for the delay... started it and then something came in the way... will finish by tomorrow for sure. Will do my best to get this merged in asap...

hmcl:
@revans2 reviewing this at this moment.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
hmcl: @jfenc91 @srdo This behavior still seems quite surprising to me. However, this fix probably won't have any bad side effect, so it should be OK. Perhaps you can add to the log message that we are ignoring it as it is out of order, duplicate, or something like that. 

@revans2 can you please comment on why, or if it is possible a tuple to be acked more than once?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
hmcl: @jfenc91 @srdo this is correct. @jfenc91 good finding!

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -266,19 +266,22 @@ private void doSeekRetriableTopicPartitions() {
...
+                kafkaConsumer.seek(rtp, acked.get(rtp).committedOffset + 1);    // Seek to last committed offset
             }
         }
     }
 
     // ======== emit  =========
     private void emit() {
-        emitTupleIfNotEmitted(waitingToEmit.next());
-        waitingToEmit.remove();
+        while(!emitTupleIfNotEmitted(waitingToEmit.next()) && waitingToEmit.hasNext()) {
```
hmcl: @jfenc91 I believe that there is no need for this while loop here. I ran all your test cases without it and they all pass. Unless there is a particular reason to have this loop in here, I would suggest that you remove it.

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/builders/SingleTopicKafkaSpoutConfiguration.java
```diff
@@ -0,0 +1,85 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.apache.storm.kafka.spout.KafkaSpoutConfig.FirstPollOffsetStrategy.EARLIEST;
+
+public class SingleTopicKafkaSpoutConfiguration {
+    public static final String stream = "test_stream";
+    public static final String topic = "test";
+
+    static public Config getConfig() {
```
hmcl: Nitpick: In all of the methods declared in this class, can you put the method modifiers in the order public static. That is the most standard way and recommended by the [JLS](https://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-8.4.3). I also believe it makes the code a bit easier to read.

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/SingleTopicKafkaSpoutTest.java
```diff
@@ -0,0 +1,250 @@
...
+import org.mockito.ArgumentCaptor;
+
+import static org.junit.Assert.*;
+
+import java.util.Map;
+import java.util.stream.IntStream;
+import static org.mockito.Mockito.*;
+import static org.apache.storm.kafka.spout.builders.SingleTopicKafkaSpoutConfiguration.*;
+
+public class SingleTopicKafkaSpoutTest {
```
hmcl: The blocks of code [B1](https://github.com/jfenc91/storm/blob/01f0c35a91251180266fe90b71ed4f34ca3339d6/external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/SingleTopicKafkaSpoutTest.java#L75-L85), [B2](https://github.com/jfenc91/storm/blob/01f0c35a91251180266fe90b71ed4f34ca3339d6/external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/SingleTopicKafkaSpoutTest.java#L118-L128), [B3](https://github.com/jfenc91/storm/blob/01f0c35a91251180266fe90b71ed4f34ca3339d6/external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/SingleTopicKafkaSpoutTest.java#L150-L160), [B4](https://github.com/jfenc91/storm/blob/01f0c35a91251180266fe90b71ed4f34ca3339d6/external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/SingleTopicKafkaSpoutTest.java#L201-L211) have pretty much the same initialization code, with exception of perhaps one or two values that can be parameterized. Can you please create an initialization or setup method with this code, and avoid all of the code duplication.

I didn't look in particular detail in other parts of this code, but it may also be easy to do the same thing for the multiple play and verify blocks of code in this class. If that's the case, that would be ideal.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
revans2: The issue is not with storm itself, but with the internal bookkeeping of the spout.  If Kafka rebalances partitions (or there are other recoverable errors) the spout resets all of it's internal bookkeeping, but there is no way to un-emit something.  So everything that was emitted, but not fully acked/committed when the recovery happened is still outstanding, and will come back for the spout to process.

1. emit A offset B
2. oh crap, recovering...
3. Recovered handling partition A again
4. emit A offset B again
5. ack A offset B (the first one)
6. ack A offset B (the second recovered one)

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
revans2: This is a performance improvement.  emitTupleIfNotEmitted does what it says.  It may or may not emit a tuple.  If it does not emit a tuple (because it has already been emitted) then the spout will exit nextTuple() having emitted nothing.  In that case storm will sleep for 1 ms, even though waitingToEmit is not necessarily empty.

hmcl:
+1 overall. It would be ideal to cleanup the test code according to my earlier suggestion. Thanks for the work @jfenc91

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
srdo: It can also happen if the Kafka cluster has unclean leader election enabled. In that case, the log end offset in Kafka may become smaller than the last offset the spout committed. It's a weird edge case, since most people probably don't need to have at least once in Storm while allowing message loss in Kafka. Bit of a newbie trap though, since unclean leader election is enabled by default in Kafka.

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/SingleTopicKafkaSpoutTest.java
jfenc91: I addressed the initialization part. I think that abstracting the different test cases away algorithmically would add too much complexity. It is nice to have test cases that are easily to understand.

jfenc91:
@hmcl thanks for the review. I believe I made all the requested changes here.

hmcl:
+1

@jfenc91 thanks for the extra cleanup.

@revans2 as far as I am concerned this patch is ready to merge. Thanks for your clarifying comments.

harshach:
Thanks for the patch @jfenc91 and for being patient with reviews.
Merged into master. I would like to merge this into 1.x-branch as well. There are few issues on SingleTopicKafkaSpoutTest.java as its using Java8 syntax. Can you please open another PR for 1.x-branch.

apiwoni:
We are on Apache Store 1.0.4 and I''m trying to convince people to upgrade because I suspect we are experiencing this bug. 
Here's what I see based on nextTuple cycle:

LOG: Topic partitions with entries ready to be retried [[TOPIC-0]]
LOG: topic-partition [TOPIC-0] has non-continuous offset [219355]. It will be processed in a subsequent batch.
org.apache.kafka.clients.consumer.KafkaConsumer#seekToEnd[ where current committed offset in Kafka = 219351]
Polled [0] records from Kafka. [6253] uncommitted offsets across all topic partitions

I would expect org.apache.storm.kafka.spout.KafkaSpout#pollKafkaBroker return few records after seekToEnd?

srdo:
@apiwoni You may be able to upgrade just the spout if you don't want to do a full cluster upgrade. I think the core APIs haven't changed since 1.0.0.

Are you using topic compaction in Kafka by any chance?

apiwoni:
@srdo We are not using topic compaction.

srdo:
@apiwoni Okay, just wanted to know because 1.0.x doesn't handle topic compaction very well, and I thought it might be related. 

Your log looks a lot like this issue. I'm not sure why you expect the poll to return records after seekToEnd? seekToEnd seeks to the end of the log in Kafka, and it doesn't actually do the seek until poll or position is called on the consumer. The result is that the consumer's position is at the last offset in the log, which means it won't emit anything until new messages are added to Kafka.

Due to this issue it will probably cause the spout to stop emitting entirely, since the failed offsets keep triggering the seekToEnd, which makes the spout skip past all new messages every time it polls.

I verified this behavior with a small example application with a single consumer that just calls seekToEnd and poll in a loop. It never emits any messages.

You should upgrade the spout. Whether you want to upgrade the rest of Storm is up to you, but be aware of https://issues.apache.org/jira/browse/STORM-2682 before deciding to jump to 1.1.1.

apiwoni:
@srdo  I'm not sure why you expect the poll to return records after seekToEnd? seekToEnd seeks to the end of the log in Kafka, and it doesn't actually do the seek until poll or position is called on the consumer. 

Based on Kafka spout loop and logs here's what I see and expect:

org.apache.storm.kafka.spout.KafkaSpout#pollKafkaBroker
	org.apache.storm.kafka.spout.KafkaSpout#doSeekRetriableTopicPartitions
		org.apache.storm.kafka.spout.KafkaSpoutRetryExponentialBackoff#retriableTopicPartitions
			// Log message indicates messageId is ready to be retried
			Topic partitions with entries ready to be retried [[TOPIC-0]]
		org.apache.storm.kafka.spout.KafkaSpout.OffsetEntry#findNextCommitOffset(TOPIC-0)
			// Returns null because of non-continuous offset
			topic-partition [TOPIC-0] has non-continuous offset [219355]. It will be processed in a subsequent batch.
		org.apache.kafka.clients.consumer.KafkaConsumer#seekToEnd(TOPIC-0)
			// This should seek to last committed offset for retriable partition, if exists, else to the last offset
			org.apache.kafka.clients.consumer.internals.SubscriptionState#needOffsetReset(OffsetResetStrategy.LATEST)
	org.apache.kafka.clients.consumer.KafkaConsumer#poll
		// I think poll should return records to be retried here from even though committed offset is 219355

Are you sure that after calling seekToEnd another call has been made to commit offset and subsequent call to seekToEnd did not got to last committed offset but to the last offset as comment in KafkaSpout states? 
seekToEnd calls org.apache.kafka.clients.consumer.internals.SubscriptionState#needOffsetReset(OffsetResetStrategy.LATEST) and my understanding of OffsetResetStrategy.LATEST is that if offset exists doesn't exist on the broker (consumer_offset topic or Zookeeper) it should indeed be set to last offset, but if current offset already exists that is where seekToEnd should go.

srdo:
@apiwoni Your expectation for 1.1.3 is wrong. seekToEnd doesn't seek to the last committed offset, it seeks to the last offset. One of the fixes made in this PR is to make doSeekRetriablePartitions seek to the last committed offset instead of the end of the partition when findNextCommitOffset returns null.

Here's the null case code from 1.0.4 https://github.com/apache/storm/blob/v1.0.4/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java#L283. Note that the comment in that line is wrong, and it's actually seeking to the end of the partition.

Here's the code for the same case after this PR https://github.com/apache/storm/pull/1679/files#diff-7d7cbc8f5444fa7ada7962033fc31c5eR269. Note that it now matches the comment.

Your understanding of OffsetResetStrategy.LATEST is also slightly off. The OffsetResetStrategy describes the behavior of the KafkaConsumer when the consumer requests an offset that is out of range. Offsets are out of range if they are smaller than the beginning of the log (e.g. if parts of the partition were deleted), or if they are larger than the offset of the largest message in the partition. For example if you have a partition with offsets 10-500 present and the consumer tries to seek to offset 0, you will trigger the OffsetResetStrategy. The strategy has no effect on seekToEnd. seekToEnd always seeks to the last offset on the partition, barring some additions made in 0.11.0.0 that have to do with transactional writes.

I think what happened in your case is what I described above, you hit one of the bugs solved by this PR, and as a result your consumer jumped to the end of the partition, which caused it to emit no messages.

PR #743, opened 2015-09-18
[STORM-1052] TridentKafkaState uses new Kafka Producer API

vesense:

PR #1682, opened 2016-09-13
STORM-2089 Replace Consumer of ISqlTridentDataSource with SqlTridentConsumer

HeartSaVioR:
* issue link: https://issues.apache.org/jira/browse/STORM-2089

This is for handling INSERT for tuples as batch if possible.

I also changed storm-sql-kafka to expose State directly, but this is effectively same as current since TridentKafkaState.updateState() does update via fully synchronous way.
(by looping each tuple and calling Future.get() immediately)

```
    public void updateState(List<TridentTuple> tuples, TridentCollector collector) {
        String topic = null;
        for (TridentTuple tuple : tuples) {
            try {

                if(topic != null) {
                    Future<RecordMetadata> result = producer.send(new ProducerRecord(topic,
                            mapper.getKeyFromTuple(tuple), mapper.getMessageFromTuple(tuple)));
                    try {
                        result.get();
                    } catch (ExecutionException e) {
                        String errorMsg = "Could not retrieve result for message with key = "
                                + mapper.getKeyFromTuple(tuple) + " from topic = " + topic;
                        LOG.error(errorMsg, e);
                        throw new FailedException(errorMsg, e);
                    }
                } else {
                    LOG.warn("skipping key = " + mapper.getKeyFromTuple(tuple) + ", topic selector returned null.");
                }
            } catch (Exception ex) {
                String errorMsg = "Could not send message with key = " + mapper.getKeyFromTuple(tuple)
                        + " to topic = " + topic;
                LOG.warn(errorMsg, ex);
                throw new FailedException(errorMsg, ex);
            }
        }
    }
```

We could improve it, but it would be better to handle it as another issue if needed.

### DIFF
--- a/external/sql/storm-sql-runtime/src/jvm/org/apache/storm/sql/runtime/ISqlTridentDataSource.java
+++ b/external/sql/storm-sql-runtime/src/jvm/org/apache/storm/sql/runtime/ISqlTridentDataSource.java
@@ -17,14 +17,49 @@
  */
 package org.apache.storm.sql.runtime;
 
-import org.apache.storm.trident.operation.Function;
-import org.apache.storm.trident.spout.IBatchSpout;
 import org.apache.storm.trident.spout.ITridentDataSource;
+import org.apache.storm.trident.state.StateFactory;
+import org.apache.storm.trident.state.StateUpdater;
 
 /**
  * A ISqlTridentDataSource specifies how an external data source produces and consumes data.
  */
 public interface ISqlTridentDataSource {
+  /**
+   * SqlTridentConsumer is a data structure containing StateFactory and StateUpdater for consuming tuples with State.
+   *
+   * Please note that StateFactory and StateUpdater should use same class which implements State.
+   *
+   * @see org.apache.storm.trident.state.StateFactory
+   * @see org.apache.storm.trident.state.StateUpdater
+   */
+  interface SqlTridentConsumer {
+    StateFactory getStateFactory();
+    StateUpdater getStateUpdater();
+  }
+
+  /**
+   * Provides instance of ITridentDataSource which can be used as producer in Trident.
+   *
+   * Since ITridentDataSource is a marker interface for Trident Spout interfaces, this method should effectively
+   * return an instance of one of these interfaces (can be changed if Trident API evolves) or descendants:
+   * - IBatchSpout
+   * - ITridentSpout
+   * - IPartitionedTridentSpout
+   * - IOpaquePartitionedTridentSpout
+   *
+   * @see org.apache.storm.trident.spout.ITridentDataSource
+   * @see org.apache.storm.trident.spout.IBatchSpout
+   * @see org.apache.storm.trident.spout.ITridentSpout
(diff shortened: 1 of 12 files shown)


external/sql/storm-sql-runtime/src/jvm/org/apache/storm/sql/runtime/ISqlTridentDataSource.java
```diff
@@ -17,14 +17,32 @@
...
+    public StateFactory getStateFactory() {
+      return stateFactory;
+    }
+
+    public StateUpdater getStateUpdater() {
+      return stateUpdater;
+    }
+  }
+
   ITridentDataSource getProducer();
```
vesense: how about add some javadoc for interface methods in `ISqlTridentDataSource` and `DataSource`?

vesense: how about add some javadoc for interface methods in `ISqlTridentDataSource` and `DataSource`?

HeartSaVioR: Makes sense. I'll add javadoc for methods and also SqlTridentConsumer.

vesense:
@HeartSaVioR 
one minor comment, others looks good to me.
about sync send `.get()`, please take a look at previous comments: https://github.com/apache/storm/pull/743#commitcomment-14195354

HeartSaVioR:
@vesense 
Yeah we need to call get() to react for the result. We could still send all requests first, and sync all to reduce waiting time. It's a kind of trade-off since it doesn't do fail-fast.

HeartSaVioR:
@vesense Addressed javadoc.

vesense:
+1
and I have filed a jira for improving the TridentKafkaState batch sending.

external/sql/storm-sql-runtime/src/jvm/org/apache/storm/sql/runtime/ISqlTridentDataSource.java
```diff
@@ -17,14 +17,62 @@
...
 public interface ISqlTridentDataSource {
+  /**
+   * SqlTridentConsumer is a data structure containing StateFactory and StateUpdater for consuming tuples with State.
+   *
+   * Please note that StateFactory and StateUpdater should use same class which implements State.
+   *
+   * @see org.apache.storm.trident.state.StateFactory
+   * @see org.apache.storm.trident.state.StateUpdater
+   */
+  class SqlTridentConsumer {
```
arunmahadevan: Its better to have this as an interface with getStateFactory and getStateUpdater.

HeartSaVioR: Good point. Will address.

HeartSaVioR:
@arunmahadevan Thanks for the review. I addressed your comment.

arunmahadevan:
+1

PR #1687, opened 2016-09-16
STORM-1694: Kafka Spout Trident Implementation Using New Kafka Consumer API

hmcl:
The Kafka Trident implementation is on top of the Trident logs improvement patch because they are related, and it makes it easier to merge the patch. There is already another PR for STORM-2097

### DIFF
--- a/examples/storm-starter/src/jvm/org/apache/storm/starter/trident/TridentKafkaClientWordCountNamedTopics.java
+++ b/examples/storm-starter/src/jvm/org/apache/storm/starter/trident/TridentKafkaClientWordCountNamedTopics.java
@@ -1,0 +1,122 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ *   or more contributor license agreements.  See the NOTICE file
+ *   distributed with this work for additional information
+ *   regarding copyright ownership.  The ASF licenses this file
+ *   to you under the Apache License, Version 2.0 (the
+ *   "License"); you may not use this file except in compliance
+ *   with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+
+package org.apache.storm.starter.trident;
+
+import org.apache.kafka.clients.consumer.ConsumerRecord;
+import org.apache.storm.kafka.spout.KafkaSpoutConfig;
+import org.apache.storm.kafka.spout.KafkaSpoutRetryExponentialBackoff;
+import org.apache.storm.kafka.spout.KafkaSpoutRetryService;
+import org.apache.storm.kafka.spout.KafkaSpoutStreams;
+import org.apache.storm.kafka.spout.KafkaSpoutStreamsNamedTopics;
+import org.apache.storm.kafka.spout.KafkaSpoutTupleBuilder;
+import org.apache.storm.kafka.spout.KafkaSpoutTuplesBuilder;
+import org.apache.storm.kafka.spout.KafkaSpoutTuplesBuilderNamedTopics;
+import org.apache.storm.kafka.spout.trident.KafkaTridentSpoutManager;
+import org.apache.storm.kafka.spout.trident.KafkaTridentSpoutOpaque;
+import org.apache.storm.trident.Stream;
+import org.apache.storm.trident.TridentState;
+import org.apache.storm.trident.TridentTopology;
+import org.apache.storm.trident.operation.builtin.Count;
+import org.apache.storm.trident.operation.builtin.Debug;
+import org.apache.storm.trident.testing.Split;
+import org.apache.storm.tuple.Fields;
+import org.apache.storm.tuple.Values;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+import static org.apache.storm.kafka.spout.KafkaSpoutConfig.FirstPollOffsetStrategy.EARLIEST;
+
+public class TridentKafkaClientWordCountNamedTopics extends TridentKafkaWordCount {
+    public TridentKafkaClientWordCountNamedTopics(String zkUrl, String brokerUrl) {
+        super(zkUrl, brokerUrl);
+    }
+
+    protected TridentState addTridentState(TridentTopology tridentTopology) {
+        final Stream spoutStream = tridentTopology.newStream("spout1", createOpaqueKafkaSpoutNew()).parallelismHint(1);
+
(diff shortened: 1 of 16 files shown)


HeartSaVioR:
Since it does load the two modules - storm-kafka and storm-kafka-client - but only load kafka-client with `${storm.kafka.client.version}`. Does storm-starter work with both storm-kafka and storm-kafka-client? If not, I guess we need to move out storm-kafka and storm-kafka-client to separate modules like recent examples modules.

examples/storm-starter/src/jvm/org/apache/storm/starter/trident/TridentKafkaClientWordCountWildcardTopics.java
```diff
@@ -0,0 +1,51 @@
...
+
+public class TridentKafkaClientWordCountWildcardTopics extends TridentKafkaClientWordCountNamedTopics {
+    private static final String TOPIC_WILDCARD_PATTERN = "test-trident(-1)?";
+
+    public TridentKafkaClientWordCountWildcardTopics(String zkUrl, String brokerUrl) {
+        super(zkUrl, brokerUrl);
+    }
+
+    public static void main(String[] args) throws Exception {
+        final String[] zkBrokerUrl = parseUrl(args);
```
harshach: do we still need zookeeper config. This topology using new KafkaSpout right?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/trident/KafkaTridentSpoutEmitter.java
```diff
@@ -0,0 +1,187 @@
...
+                topicPartitionsTrident.add(new KafkaTridentSpoutTopicPartition(topicPartition));
+            }
+        }
+        LOG.debug("OrderedPartitions = {}", topicPartitionsTrident);
+        return topicPartitionsTrident;
+    }
+
+    @Override
+    public void close() {
+        LOG.debug("Closed");
```
hmcl: Done

examples/storm-starter/src/jvm/org/apache/storm/starter/trident/TridentKafkaClientWordCountWildcardTopics.java
hmcl: Agree. Refactored examples in this [PR](https://github.com/apache/storm/pull/1757)

examples/storm-starter/src/jvm/org/apache/storm/starter/trident/TridentKafkaClientWordCountNamedTopics.java
```diff
@@ -0,0 +1,122 @@
...
+        final Stream spoutStream = tridentTopology.newStream("spout1", createOpaqueKafkaSpoutNew()).parallelismHint(1);
+
+        return spoutStream.each(spoutStream.getOutputFields(), new Debug(true))
+                .each(new Fields("str"), new Split(), new Fields("word"))
+                .groupBy(new Fields("word"))
+                .persistentAggregate(new DebugMemoryMapState.Factory(), new Count(), new Fields("count"));
+    }
+
+    private KafkaTridentSpoutOpaque<String, String> createOpaqueKafkaSpoutNew() {
+        return new KafkaTridentSpoutOpaque<String, String>(getKafkaTridentManager());
```
hmcl: Partially Done in refactored examples in this [PR](https://github.com/apache/storm/pull/1757).

There were some redundant "factory methods" that I removed. However, the code creating the "dependency" objects that need to be passed in is not 1 or two lines. I believe that a method with a meaningful name creating and initializing these "dependency" objects makes the code much more cohesive and easier to read. Furthermore, this class is extended for wildcard topics, and some of these methods overridden. 

I will be happy to write a more "copy" and "paste" like example in the docs if you feel it's appropriate. Please let me know.

examples/storm-starter/src/jvm/org/apache/storm/starter/trident/TridentKafkaClientWordCountNamedTopics.java
```diff
@@ -0,0 +1,122 @@
...
+    private static KafkaSpoutRetryExponentialBackoff.TimeInterval getTimeInterval(long delay, TimeUnit timeUnit) {
+        return new KafkaSpoutRetryExponentialBackoff.TimeInterval(delay, timeUnit);
+    }
+
+    protected KafkaSpoutStreams getKafkaSpoutStreams() {
+        final Fields outputFields = new Fields("str");
+        return new KafkaSpoutStreamsNamedTopics.Builder(outputFields, new String[]{"test-trident","test-trident-1"}).build();
+    }
+
+    protected static class TopicsTupleBuilder<K, V> extends KafkaSpoutTupleBuilder<K,V> {
```
hmcl: I am not sure I completely follow this observation. This same object exists for the Kafa Spout. This is only used for the user to be able to provide a custom implementation of how he wishes to create (build) a Tuple from the ConsumerRecord object. The users are free to, when creating their KafkaSpoutTupleBuilder implementation, specify which Deserializer they want to use. 

As far as default implementation, I am not sure I am following. I don't think we are providing any default implementation. The class KafkaSpoutTupleBuilder is abstract. This implementation here is for the sake of the example.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/trident/KafkaTridentSpoutBatchMetadata.java
```diff
@@ -0,0 +1,83 @@
...
+        if (records != null && !records.isEmpty()) {
+            firstOffset = records.get(0).offset();
+            lastOffset = records.get(records.size() - 1).offset();
+        } else {
+            if (lastBatch != null) {
+                firstOffset = lastBatch.firstOffset;
+                lastOffset = lastBatch.lastOffset;
+            }
+        }
+        LOG.debug("Created {}", this);
```
hmcl: logging "this" will call the overridden toString() method for this class, which prints the first and last offset.

hmcl:
@HeartSaVioR I believe I address your valid concerns about the kafka library versions in this [PR](https://github.com/apache/storm/pull/1757)

HeartSaVioR:
@hmcl OK thanks for following up. 
@harshach Please treat my comment as addressed, and also feel free to merge when you think it's OK to merge.

harshach:
@hmcl these PR titles are very confusing. Can you fix that in both the PRs. Not sure what it means by "Apache master storm 1694 top storm 2097". Just add the JIRA title as PR title.

harshach:
@hmcl also this PR is the same as this one with one more commit?
https://github.com/apache/storm/pull/1757/commits . If so can you close this one.

hmcl:
@harshach I have changed the titles of the PRs, however what is really important are the git commit messages, and those were correct. Once the patch is merged, no one will ever look at the PR titles again, only at the git commit logs.

I don't think we should close this PR in order to keep the history of the review comments. Also, the title of this PR reflects this patch. The title of the other PR reflects the other patch that is on top of this patch.

There is no extra work in merging the two PRs and the git commit history is the same.

hmcl:
@harshach done

PR #1688, opened 2016-09-19
STORM-2098: DruidBeamBolt: Pass DruidConfig.Builder as constructor argument

omkreddy:


HeartSaVioR:
AFAIK, 24hr policy and vote requirement is not in effect if it's for porting to other branches.
(If there're huge merge conflict you may want to treat PR as normal PR and having reviews.)

You don't need to wait for that, and you can also port back (and forth) while merging PR for master branch or so.

satishd:
Thanks @omkreddy

PR #1689, opened 2016-09-19
STORM-2099 Introduce new sql external module: storm-sql-redis

HeartSaVioR:
This patch is on top of [STORM-2089](https://issues.apache.org/jira/browse/STORM-2089) (#1682)

I moved JsonScheme and JsonSerializer to storm-sql-runtime since it's used from storm-sql-kafka and storm-sql-redis. In result, additional dependency (jackson-databind) for storm-sql-runtime is introduced.

Below is the example of defining Redis table:

```
CREATE EXTERNAL TABLE LARGE_ORDERS (ID INT PRIMARY KEY, TOTAL INT) LOCATION 'redis://localhost:7379' TBLPROPERTIES '{"data.type": "STRING", "data.additional.key": "addkey", "redis.timeout": "3000", "use.redis.cluster": "true"}'
```

LOCATION can be set to `redis://:[password]@[host]:[port]/[dbIdx]`.
Default mode is JedisPool (for single Redis). If you want to enable cluster mode you need to specify option "use.redis.cluster" to "true".

In cluster mode password and dbIdx is ignored. DB index is not valid for Cluster mode, but recently password for Cluster mode of Redis becomes valid. 
This will be addressed from storm-redis and addressed to storm-sql-redis later.

Please note that Redis(Cluster)State only supports STRING and HASH since State stores key and value.

### DIFF
--- a/external/sql/pom.xml
+++ b/external/sql/pom.xml
@@ -40,5 +40,6 @@
         <module>storm-sql-core</module>
         <module>storm-sql-runtime</module>
         <module>storm-sql-external/storm-sql-kafka</module>
+        <module>storm-sql-external/storm-sql-redis</module>
     </modules>
 </project>
--- a/external/sql/storm-sql-external/storm-sql-redis/pom.xml
+++ b/external/sql/storm-sql-external/storm-sql-redis/pom.xml
@@ -1,0 +1,80 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+ Licensed to the Apache Software Foundation (ASF) under one or more
+ contributor license agreements.  See the NOTICE file distributed with
+ this work for additional information regarding copyright ownership.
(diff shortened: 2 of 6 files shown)


HeartSaVioR:
In order to add Redis as data source for SQL, I added two artifacts to --artifacts parameter.

- org.apache.storm:storm-sql-redis:2.0.0-SNAPSHOT
- org.apache.storm:storm-redis:2.0.0-SNAPSHOT

Might be better to have a separate document for explaining the data sources for SQL.

HeartSaVioR:
There're some compile issues from test side. I'll handle it.

HeartSaVioR:
Build for '!storm-core' passed.

HeartSaVioR:
Rebased to reflect merged STORM-2089.

harshach:
+1

PR #1690, opened 2016-09-19
STORM-2100 Fix Trident SQL join tests to not rely on ordering

HeartSaVioR:
Issue link: [STORM-2100](https://issues.apache.org/jira/browse/STORM-2100)

The join doesn't guarantee preserving order (Trident and SQL itself) so we shouldn't rely on that.

Before this patch, tests are failing with JDK7 (with 1.x-branch). I confirmed that this patch makes tests succeed with 1.x-branch & JDK 1.7 (1.7.0_79) combination.

@satishd Could you please review this? Thanks.

### DIFF
--- a/external/sql/storm-sql-core/src/test/org/apache/storm/sql/compiler/backends/trident/TestPlanCompiler.java
+++ b/external/sql/storm-sql-core/src/test/org/apache/storm/sql/compiler/backends/trident/TestPlanCompiler.java
@@ -19,34 +19,34 @@
  */
 package org.apache.storm.sql.compiler.backends.trident;
 
+import com.google.common.collect.Lists;
+import org.apache.calcite.adapter.java.JavaTypeFactory;
+import org.apache.calcite.jdbc.JavaTypeFactoryImpl;
+import org.apache.calcite.rel.type.RelDataTypeSystem;
+import org.apache.commons.collections4.CollectionUtils;
 import org.apache.storm.Config;
 import org.apache.storm.ILocalCluster;
 import org.apache.storm.LocalCluster;
 import org.apache.storm.generated.StormTopology;
-import org.apache.storm.sql.runtime.AbstractValuesProcessor;
-import org.apache.storm.tuple.Fields;
-import org.apache.storm.tuple.Values;
-import org.apache.storm.utils.Utils;
-import org.apache.calcite.adapter.java.JavaTypeFactory;
-import org.apache.calcite.jdbc.JavaTypeFactoryImpl;
-import org.apache.calcite.rel.type.RelDataTypeSystem;
 import org.apache.storm.sql.TestUtils;
 import org.apache.storm.sql.TestUtils.MockSqlTridentDataSource.CollectDataFunction;
 import org.apache.storm.sql.compiler.TestCompilerUtils;
 import org.apache.storm.sql.runtime.ISqlTridentDataSource;
 import org.apache.storm.sql.runtime.trident.AbstractTridentProcessor;
+import org.apache.storm.trident.TridentTopology;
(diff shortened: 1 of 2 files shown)


external/sql/storm-sql-core/src/test/org/apache/storm/sql/compiler/backends/trident/TestPlanCompiler.java
```diff
@@ -274,4 +276,9 @@ private void waitForCompletion(long timeout, Callable<Boolean> cond) throws Exce
       Thread.sleep(100);
     }
   }
+
+  private void assertArrayWithoutOrder(List<Values> expected, List<List<Object>> actual) {
+    Assert.assertTrue("Two arrays are not same!\n"+ "Expected: " + expected + "\n" + "Actual: " + actual,
+            expected.containsAll(actual) && actual.containsAll(expected));
```
satishd: @HeartSaVioR You may want to check for size equality as there can be different number of equal/same elements in those collections.

HeartSaVioR: @satishd Yes I missed the duplicated items. 
Btw, if we assume the duplication, checking size equality seems not enough too, since `[a b b a a]` and `[b b a a b]` is same length but it's not actually same. 
Checking cardinality of each element seems to be the safest way, and if the type of element is comparable, we can copy & sort two arrays and compare with ordered way.

satishd: @HeartSaVioR Do you want to address that?

satishd: @HeartSaVioR Sorry, I did not complete my earlier comment. Pl handle the collection equality check appropriately.

HeartSaVioR: @satishd Addressed via using CollectionUtils.isEqualCollection() in commons collections 4.

http://commons.apache.org/proper/commons-collections/javadocs/api-release/org/apache/commons/collections4/CollectionUtils.html#isEqualCollection(java.util.Collection,%20java.util.Collection)

external/sql/storm-sql-core/src/test/org/apache/storm/sql/compiler/backends/trident/TestPlanCompiler.java
```diff
@@ -274,4 +274,9 @@ private void waitForCompletion(long timeout, Callable<Boolean> cond) throws Exce
       Thread.sleep(100);
     }
   }
+
+  private void assertArrayWithoutOrder(List<Values> expected, List<List<Object>> actual) {
+    Assert.assertTrue("Two arrays are not same!\n"+ "Expected: " + expected + "\n" + "Actual: " + actual,
```
satishd: minor nit: Can you say `two collections` instead of `two arrays`?

satishd:
+1, once the remaining nit-pick is done.

external/sql/storm-sql-core/src/test/org/apache/storm/sql/compiler/backends/trident/TestPlanCompiler.java
HeartSaVioR: @satishd Thanks for the detailed review. Changed to `Two lists` :) also rename the method name to represent list.

satishd:
Thanks @HeartSaVioR

PR #1692, opened 2016-09-20
Fix STORM-2017

kluoto:
Moving taskIdsQueue inside the lock should remove the race condition.


revans2:
+1 the change looks good.

@kluoto is this an issue for master too?

kluoto:
I think same issue is also on master too.

PR #1693, opened 2016-09-20
[STORM-1961] Stream api for storm core use cases

arunmahadevan:
The initial version of unified stream api for expressing streaming computation pipelines over the storm core spouts and bolts. Right now this provides at-least once guarantees and addresses only the storm core use cases.

For high level design see - https://issues.apache.org/jira/secure/attachment/12827547/UnifiedStreamapiforStorm.pdf

A few examples have been added which should give a basic idea on how to use the apis. More examples and detailed documentation will be added as followup tasks.

### DIFF
--- a/examples/storm-starter/src/jvm/org/apache/storm/starter/streams/BranchExample.java
+++ b/examples/storm-starter/src/jvm/org/apache/storm/starter/streams/BranchExample.java
@@ -1,0 +1,72 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.storm.starter.streams;
+
+import org.apache.storm.Config;
+import org.apache.storm.LocalCluster;
+import org.apache.storm.StormSubmitter;
+import org.apache.storm.starter.spout.RandomIntegerSpout;
+import org.apache.storm.streams.Stream;
+import org.apache.storm.streams.StreamBuilder;
+import org.apache.storm.streams.operations.Predicate;
+import org.apache.storm.streams.operations.mappers.ValueMapper;
+import org.apache.storm.utils.Utils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * An example that demonstrates the usage of {@link Stream#branch(Predicate[])} to split a stream
+ * into multiple branches based on predicates.
+ */
+public class BranchExample {
+    private static final Logger LOG = LoggerFactory.getLogger(BranchExample.class);
+
+    @SuppressWarnings("unchecked")
+    public static void main(String[] args) throws Exception {
+        StreamBuilder builder = new StreamBuilder();
+        Stream<Integer>[] evenAndOdd = builder
+                /*
+                 * Create a stream of random numbers from a spout that
+                 * emits random integers by extracting the tuple value at index 0.
+                 */
+                .newStream(new RandomIntegerSpout(), new ValueMapper<Integer>(0))
+                /*
+                 * Split the stream of numbers into streams of
+                 * even and odd numbers. The first stream contains even
+                 * and the second contains odd numbers.
+                 */
+                .branch(x -> (x % 2) == 0,
+                        x -> (x % 2) == 1);
+
+        evenAndOdd[0].forEach(x -> LOG.info("EVEN> " + x));
+        evenAndOdd[1].forEach(x -> LOG.info("ODD > " + x));
+
+        Config config = new Config();
+        if (args.length > 0) {
+            config.setNumWorkers(1);
+            StormSubmitter.submitTopologyWithProgressBar(args[0], config, builder.build());
+        } else {
+            LocalCluster cluster = new LocalCluster();
+            cluster.submitTopology("test", config, builder.build());
+            Utils.sleep(60000);
+            cluster.killTopology("test");
+            cluster.shutdown();
+        }
+    }
+
+}
--- a/examples/storm-starter/src/jvm/org/apache/storm/starter/streams/BranchExample.java
+++ b/examples/storm-starter/src/jvm/org/apache/storm/starter/streams/BranchExample.java
@@ -61,11 +61,10 @@ public class BranchExample {
             config.setNumWorkers(1);
             StormSubmitter.submitTopologyWithProgressBar(args[0], config, builder.build());
         } else {
-            LocalCluster cluster = new LocalCluster();
-            cluster.submitTopology("test", config, builder.build());
-            Utils.sleep(60000);
-            cluster.killTopology("test");
-            cluster.shutdown();
+            try (LocalCluster cluster = new LocalCluster();
+                 LocalCluster.LocalTopology topo = cluster.submitTopology("test", config, builder.build())) {
+                Utils.sleep(60_000);
+            }
         }
     }
 
--- a/examples/storm-starter/src/jvm/org/apache/storm/starter/streams/GroupByKeyAndWindowExample.java
+++ b/examples/storm-starter/src/jvm/org/apache/storm/starter/streams/GroupByKeyAndWindowExample.java
@@ -1,0 +1,109 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.storm.starter.streams;
+
+import org.apache.storm.Config;
+import org.apache.storm.LocalCluster;
+import org.apache.storm.StormSubmitter;
+import org.apache.storm.spout.SpoutOutputCollector;
+import org.apache.storm.streams.PairStream;
+import org.apache.storm.streams.StreamBuilder;
+import org.apache.storm.streams.operations.Reducer;
+import org.apache.storm.streams.operations.mappers.PairValueMapper;
+import org.apache.storm.streams.windowing.SlidingWindows;
+import org.apache.storm.streams.windowing.Window;
+import org.apache.storm.task.TopologyContext;
+import org.apache.storm.topology.OutputFieldsDeclarer;
+import org.apache.storm.topology.base.BaseRichSpout;
+import org.apache.storm.topology.base.BaseWindowedBolt.Count;
+import org.apache.storm.tuple.Fields;
+import org.apache.storm.tuple.Values;
+import org.apache.storm.utils.Utils;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * An example that shows the usage of {@link PairStream#groupByKeyAndWindow(Window)}
+ * and {@link PairStream#reduceByKeyAndWindow(Reducer, Window)}
+ */
+public class GroupByKeyAndWindowExample {
+    public static void main(String[] args) throws Exception {
+        StreamBuilder builder = new StreamBuilder();
+
+        // a stream of stock quotes
+        builder.newStream(new StockQuotes(), new PairValueMapper<String, Double>(0, 1))
+                /*
+                 * The elements having the same key within the window will be grouped
+                 * together and the corresponding values will be merged.
+                 */
+                .groupByKeyAndWindow(SlidingWindows.of(Count.of(6), Count.of(3)))
+                .print();
+
+        // a stream of stock quotes
+        builder.newStream(new StockQuotes(), new PairValueMapper<String, Double>(0, 1))
+                /*
+                 * The elements having the same key within the window will be grouped
+                 * together and their values will be reduced using the given reduce function.
+                 */
+                .reduceByKeyAndWindow((x, y) -> (x + y) / 2.0, SlidingWindows.of(Count.of(6), Count.of(3)))
+                .print();
+
+        Config config = new Config();
+        if (args.length > 0) {
+            config.setNumWorkers(1);
+            StormSubmitter.submitTopologyWithProgressBar(args[0], config, builder.build());
+        } else {
+            LocalCluster cluster = new LocalCluster();
+            config.put(Config.TOPOLOGY_STATE_PROVIDER, "org.apache.storm.redis.state.RedisKeyValueStateProvider");
+            cluster.submitTopology("test", config, builder.build());
+            Utils.sleep(60000);
+            cluster.killTopology("test");
+            cluster.shutdown();
+        }
+    }
+
+    private static class StockQuotes extends BaseRichSpout {
+        private final List<List<Values>> values = Arrays.asList(
+                Arrays.asList(new Values("AAPL", 100.0), new Values("GOOG", 780.0), new Values("FB", 125.0)),
+                Arrays.asList(new Values("AAPL", 105.0), new Values("GOOG", 790.0), new Values("FB", 130.0))
+        );
+        private SpoutOutputCollector collector;
+        private int index = 0;
+
+        @Override
+        public void open(Map conf, TopologyContext context, SpoutOutputCollector collector) {
+            this.collector = collector;
+        }
+
+        @Override
+        public void nextTuple() {
+            Utils.sleep(5000);
+            for (Values v : values.get(index)) {
+                collector.emit(v);
+            }
+            index = (index + 1) % values.size();
+        }
+
+        @Override
+        public void declareOutputFields(OutputFieldsDeclarer declarer) {
+            declarer.declare(new Fields("key", "val"));
+        }
+    }
+}
--- a/examples/storm-starter/src/jvm/org/apache/storm/starter/streams/GroupByKeyAndWindowExample.java
+++ b/examples/storm-starter/src/jvm/org/apache/storm/starter/streams/GroupByKeyAndWindowExample.java
@@ -52,6 +52,9 @@ public class GroupByKeyAndWindowExample {
                 /*
                  * The elements having the same key within the window will be grouped
                  * together and the corresponding values will be merged.
+                 *
+                 * The result is a PairStream<String, Iterable<Double>> with
+                 * 'stock symbol' as the key and 'stock prices' for that symbol within the window as the value.
                  */
                 .groupByKeyAndWindow(SlidingWindows.of(Count.of(6), Count.of(3)))
                 .print();
@@ -61,8 +64,11 @@ public class GroupByKeyAndWindowExample {
                 /*
                  * The elements having the same key within the window will be grouped
                  * together and their values will be reduced using the given reduce function.
+                 *
+                 * Here the result is a PairStream<String, Double> with
+                 * 'stock symbol' as the key and the maximum price for that symbol within the window as the value.
                  */
-                .reduceByKeyAndWindow((x, y) -> (x + y) / 2.0, SlidingWindows.of(Count.of(6), Count.of(3)))
+                .reduceByKeyAndWindow((x, y) -> x > y ? x : y, SlidingWindows.of(Count.of(6), Count.of(3)))
                 .print();
 
         Config config = new Config();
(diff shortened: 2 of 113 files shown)


arunmahadevan:
Updated doc -
https://docs.google.com/document/d/1Ew7uFF1UJ6e_zq0t4bM6A9auuEaArviAjjWYSpVFqPY/edit?usp=sharing

Pinging for reviews and feedback. Its a big patch mostly due the number of different apis, however the doc and the examples should be a good starting point.

@HeartSaVioR would like inputs from you esp. since it could be potentially used in storm-sql in future.

@ptgoetz would be great if you could take a look as well, since you were working on a prototype beam runner for storm. To validate the apis and identify gaps, I will try to do a prototype runner using the proposed apis.

@revans2 please take a look if you get a chance.

ptgoetz:
@arunmahadevan I will review. Before you start working on a prototype beam runner, let me push what I have to a feature branch. I have a fair amount of the basic scaffolding you would need.

HeartSaVioR:
@arunmahadevan 
I'm in the middle of reviewing updated design doc again. Since I've read it once so there will be no outstanding comments, but just would like to check again.

Btw, I merged this PR to my local and convert one of example on storm-starter (windowing example), and it seems to run fine. After reviewing design doc, I'll play with API sets and leave some feedbacks.

satishd:
It is better to discuss these APIs before working on Beam runner prototypes. Lets not jump on using these APIs to work on runner. Beam API is currently changing and we should discuss those API contracts and we should evaluate what kind of enhancements we need to do. We can create subtasks for the requirements and work on those later.

arunmahadevan:
@satishd right now its more of a prototype proof of concept for the proposed apis, not the final version. Once the beam api solidifies we can have a separate discussion to identify the gaps, but doesn't make sense to hold off the prototype implementation which is more for validating the proposed apis.

HeartSaVioR:
For me this stream API is not only for supporting Beam. This is what Storm needed to have even before Beam is open-sourced. We may want to label this API to experiment or unstable or something, but anyway I'd like to see it released not too late.

satishd:
@arunmahadevan I was saying we should not hold on this PR till Beam runner API is tried with these APIs. We should review these APIs soon and these may remain experimental for couple of minor releases. 

I was going through Beam features earlier and I plan to have a doc about Beam features and how we can start implementing some of those APIs. I will add details once Taylor's branch is open and the doc can be updated with that. I will share the doc once it is ready(in 2-3 weeks). We can create subtasks based on that doc and work on them later.

arunmahadevan:
@satishd proper beam support cannot happen until underlying storm primitives are enhanced and the windowing enhancements which has been pending is completed. This a big task which we can discuss separately. The prototype I was mentioning is just a proof of concept with the proposed api to see how it goes, when I get some cycles. This PR definitely need not wait on this and we can mark the apis "unstable" and release it sooner once the review is done.

satishd:
@arunmahadevan Ofcourse, BEAM support is not expected to be done in a large PR with all the functionality. The approach I mentioned is to do in stages starting with a doc to discuss BEAM features and involve community in all those stages. This is more about discussing/figuring out the requirements for state checkpointing, windowing and operators semantics.

revans2:
I have looked through the proposal, and I have a number of concerns about the API.  I mostly want to be sure that we can consistent terminology about what different operations are intended to do.  groupby is the big one for me, because it seems to be a routing function and not an actual groupby, but join also seemed a little off.

Why do we need so many lines of code to do a repartition/groupby in the word count example?

```
builder.newStream(new RandomSentenceSpout(), new ValueMapper<String>(0), 2)
      /*
       * split the sentences to words
       */
      .flatMap(s -> Arrays.asList(s.split(" ")))
      /*
       * create a stream of (word, 1) pairs
       */
      .mapToPair(w -> new Pair<>(w, 1))
      /*
       * group by word so that the same words end up in the same partition
       */
      .groupByKey()
      /*
       * streams parallelism changed to 3 and applies to downstream operations
       */
      .repartition(3) 
      /*
       * a ten seconds tumbling window
       */
      .window(TumblingWindows.of(Duration.seconds(10)))
      /*
       * compute the word counts in the last two second window
       */
      .aggregateByKey(new Count<>())
      /*
       * the words that occurred at-least five times in the last two seconds
       */
      .filter(x -> x.getSecond() >= 5)
      /*
       * print the results (word, count) to stdout
       */
      .print();
```

it feels like we could accomplish the entire thing much simpler (look at flink or spark streaming, even BEAM (although it is rather complex ultimetly the final code does the aggregation more cleanly).

```
builder.newStream(new RandomSentenceSpout(), new ValueMapper<String>(0), 2)
      .flatMap(s -> Arrays.asList(s.split(" ")))
      .window(TumblingWindows.of(Duration.seconds(10)))
      // short hand for .aggregateBy(1 /*index to use in grouping*/, new Count<>(), 3 /*new number of partitions although we probably want to internally actually have many more partitions but only 3 executors to start out with*/)
      .countBy(1, 3)
      .filter(x -> x.getSecond() >= 5)
      .print();
```

But I am most concerned about expanding this API in the future.  The stated plan is to start out with defining the API and implementing a back end that supports at most one and at least once processing and then move on to exactly once processing.  I really would like to have a plan in place on how we expect to do exactly once processing before we go too deeply into the API.  Even if we do mark the API as unstable.  I am really concerned about how state plays into this (inputs, outputs, windowing, and actual state).  We currently have two different types of spouts.  We have one type for at least once/at most once and another for exactly once.  The exactly once use case scares me a bit because if we want to support BEAM they have a very specific model on exactly once.  All operations are idempotent and replay includes a complete rollback to a known consistent checkpoint.  The trident spout does not do this (no ordering guarantees).  The other spouts also do not do this.  My concern is that the APIs for a lot of different user facing things may need to change for exactly once, and I would like to have a plan in place on how we want to support it and have thought out how that might impact the APIs before we release this to the public.

This also goes for the windows.  Having more robust windowing may impact some of these APIs.  Have we thought that through?

arunmahadevan:
@revans2  thanks for taking time to do detailed and thoughtful review.

> I have looked through the proposal, and I have a number of concerns about the API. I mostly want to be sure that we can consistent terminology about what different operations are intended to do. groupby is the big one for me, because it seems to be a routing function and not an actual groupby, but join also seemed a little off.

groupby is doing the fields grouping on key. I will rename it. May be it need NOT be exposed at all since reduce/aggregate etc can do implicit partitioning on the key. I didn't do implicit partitioning in aggregate since trident had a concept of partitionAggregate (keyed aggregate per partition) and thought it may be useful. May be its not needed.

Will add a groupByKey() api which returns an `Iterable<V>` similar to beam/spark.

> Why do we need so many lines of code to do a repartition/groupby in the word count example?

`repartition` is not really required, but was there to show the usage. I will remove it from the example. Groupby (partitionbykey) can also be removed and aggregate can do an implicit fields grouping.

>builder.newStream(new RandomSentenceSpout(), new ValueMapper<String>(0), 2)
      .flatMap(s -> Arrays.asList(s.split(" ")))
      .window(TumblingWindows.of(Duration.seconds(10)))
      // short hand for .aggregateBy(1 /*index to use in grouping*/, new Count<>(), 3 /*new number of partitions although we probably want to internally actually have many more partitions but only 3 executors to start out with*/)
      .countBy(1, 3)
      .filter(x -> x.getSecond() >= 5)
      .print();

Since count seems to be a common operation it can be added as an additional api. But aggregate is more general and users can provide different `Aggregators`.

>But I am most concerned about expanding this API in the future. The stated plan is to start out with defining the API and implementing a back end that supports at most one and at least once processing and then move on to exactly once processing. I really would like to have a plan in place on how we expect to do exactly once processing before we go too deeply into the API.

The state checkpointing mechanism we have can be extended to provide exactly once. (e.g with transactional updates and message de-duplication similar to the millwheel model), or saving source level spout offsets.

The initial ideas for phase 2 are listed here - https://docs.google.com/document/d/1Ew7uFF1UJ6e_zq0t4bM6A9auuEaArviAjjWYSpVFqPY/edit#heading=h.nmf14n

> My concern is that the APIs for a lot of different user facing things may need to change for exactly once, 

The changes would mostly be in the underlying framework and we may add some additional constraints like the emitted tuple should have a unique message Id field and re-emits with the same message id for exactly once. It can be non-intrusive without changes at the api level.

>This also goes for the windows. Having more robust windowing may impact some of these APIs. Have we thought that through?

The underlying windowing framework has to be enhanced before this can be supported. From the api point of view, the `window()` function right now takes a window config builder where we can add additional apis to support triggers and so on. Something like,
`stream.window(TumblingWindows.of(...).triggering(...).withAllowedLateness(Duration.standardDays(2))` and so on, and the window will trigger based on the window config.

revans2:
Ok I feel a lot better seeing how you have thought through most of these things.  I would like to see us write out some more examples, not just word count in what this new API would look like.  The we could have a better feel of what this might look like.

arunmahadevan:
@revans2 
A few basic examples are [here](https://github.com/arunmahadevan/storm/tree/9cc722194f30470eddba1b2257bab6ea5444ee94/examples/storm-starter/src/jvm/org/apache/storm/starter/streams).

Will add more examples, a Readme and api documentation, but I have very less free cycles to work on it right now, so will take some more time.

I also pushed a commit to change the groupBy semantics so that it returns a pair of `key, Iterable<V>`. The aggregation operations on key now does an automatic partitioning on the key (unless the stream is already partitioned by key). Examples are updated to reflect the same.

I will raise follow up tasks to add additional examples and apis like cogroup, transform and so on which I think would be useful. I plan to work on it incrementally when I get time, but if someone in the community wants to contribute, I will be happy to help.

@HeartSaVioR thanks for your inputs I also added Typed tuples (upto Tuple10 for now), see [TypedTupleExample](https://github.com/arunmahadevan/storm/blob/9cc722194f30470eddba1b2257bab6ea5444ee94/examples/storm-starter/src/jvm/org/apache/storm/starter/streams/TypedTupleExample.java) for usage.

arunmahadevan:
Have fixed a few issues, updated the examples and raised follow up [JIRAs](https://issues.apache.org/jira/issues/?jql=parent%20%3D%20STORM-1843%20ORDER%20BY%20key%20ASC) for tasks that can be taken up after the initial work. 

I think the PR is in a better shape now to do the code review. @revans2 @HeartSaVioR would appreciate if you could also take a look as time permits.

roshannaik:
This Streams API is built on top of the core APIs. The idea of building Beam APIs on top of this looks like the API stack is getting too tall... 3 levels!!   Like a tower of cards. 
In practice, not a great idea from the debugging/monitoring/other perspectives ... IMO.

HeartSaVioR:
Let's not worry about Beam API. Stream API is needed for Storm itself, not only for supporting Beam. Once we make shape a stream public API set, we could change internal state if we really need.

storm-core/src/jvm/org/apache/storm/streams/operations/aggregators/Sum.java
```diff
@@ -0,0 +1,45 @@
...
+ * limitations under the License.
+ */
+package org.apache.storm.streams.operations.aggregators;
+
+import org.apache.storm.streams.operations.CombinerAggregator;
+
+/**
+ * Computes the long sum of the input values
+ */
+public class Sum implements CombinerAggregator<Number, Long, Long> {
```
HeartSaVioR: Class name `Sum` with input type `Number` could make users confusing (even though we're describing the behavior to the javadoc). 

How about providing multiple Sum classes for types, or rename this class to LongSum, or even removing this class?

storm-core/src/jvm/org/apache/storm/streams/processors/Processor.java
```diff
@@ -0,0 +1,51 @@
...
+    /**
+     * Initializes the processor. This is typically invoked from the underlying
+     * storm bolt's prepare method.
+     *
+     * @param context the processor context
+     */
+    void init(ProcessorContext context);
+
+    /**
+     * Executes some operation on the input and possibly emits some result.
```
HeartSaVioR: nit: result -> results, operation -> operations

storm-core/src/jvm/org/apache/storm/streams/processors/ProcessorContext.java
```diff
@@ -0,0 +1,59 @@
...
+     * Forwards the input to downstream processors at specified stream.
+     *
+     * @param input  the input
+     * @param stream the stream to forward
+     * @param <T>    the type of the input
+     */
+    <T> void forward(T input, String stream);
+
+    /**
+     * Returns if the processing is in a windowed context and should wait for
```
HeartSaVioR: nit: Returns `true` if

storm-core/src/jvm/org/apache/storm/streams/processors/ProcessorContext.java
```diff
@@ -0,0 +1,59 @@
...
+     * punctuation before emitting results.
+     *
+     * @return if this is a windowed context or not
```
HeartSaVioR: nit: if -> whether

HeartSaVioR:
@arunmahadevan 
Btw, I'm seeing some failures with Redis state backend: StatefulWordCount and StateQueryExample. They're running fine with removing Redis state backend setting.

Could you take a look if you have time?

HeartSaVioR:
I just have a look at other streaming frameworks API, and feel that this is a good start. We can still add features after this patch.

I'm OK to add this for 2.0.0, but am also OK to add this for 1.x for experimental purpose. (No strong opinion since it involves additional works.)

Since this is a huge change and introduce a brand new API set, I'd like to wait for more people to gather more reviews and also opinions.

arunmahadevan:
@HeartSaVioR thanks for taking time to review and provide feedback. The Redis state was working for me earlier however after rebasing with master I am seeing exceptions (triggered from kryo) and I think it may be due to multiple threads using the same kryo instance. I will look into this and other comments.

roshannaik:
One  important thing to include in this API will be to allow users to be explicit about how to handle  errors.  I don't know if any API out there does it... so this would be unique to Storm. 

In short: 
Broadly speaking there are two kinds of tuple process errors that the users need to be concerned about:

1- **Retry worthy Errors** - For instance, failure to deliver to destination service due to connection issues.
2- **Not worth retrying** - For instance, Parsing errors due to bad data in the tuple. These problems can jam up the pipeline if they are retried repeatedly. Such tuples can be sent to a configurable Dead-Letter-Queue.

arunmahadevan:
@roshannaik 

>1- Retry worthy Errors

Right now theres no explicit `fail` api. When a stage in the stream completes processing (and possibly emits results), the underlying tuples are acked automatically. The only way spout will re-emit is after the message timeout. It may be good to have a fail fast api, but I am not sure how it would help since the replayed tuple could fail again in processing. Instead the processing logic itself can have some retry logic (say retry 3 times) and forward to an error stream  and ack the tuple.

> 2- Not worth retrying

This can be handled via branch logic. E.g. send valid values to stream1 and bad values to stream2.

I have opened https://issues.apache.org/jira/browse/STORM-2282 to discuss ideas around this and see what apis can be added.

arunmahadevan:
@HeartSaVioR have addressed the comments and fixed the issue with state handling. StatefulWordCount and StateQueryExample works fine for me locally.

HeartSaVioR:
@arunmahadevan Could we have separate patch for state handling? We should push the fix for 1.x and 1.0.x branches.

arunmahadevan:
@HeartSaVioR sure will raise as separate PR with the specific fix for state handling. 

If we plan to consider this PR for 1.x branch need to do some refactoring to make it work with Java 7 unless we plan to move 1.x to Java 1.8 soon.

roshannaik:
@arunmahadevan  
Its ok to keep retry-ing errors that are considered retry-worthy. The non-retry worthy are the ones we must fail fast and move them out as they will cause a jam and prevent good tuples from flowing. Also no good way to recover from that.

One approach to do this... 
For **non-retry worthy** error conditions (like bad data), any processing element in the pipeline can throw a specific exception. This can then be handled by the runtime to send to a configurable dead letter queue. Ideally this needs a new kind of Fail() notification in the spout to avoid re-emit. Alternatively, we can send the spout an ACK instead of FAIL to avoid retry. The DeadLetterQ bolt's metrics will capture these failure metrics. Best to no have each spout/bolt explicitly deal with dead letter queue ... that will complicate the topology definition...as every spout bolt will need to be configured and wired up to do this.

For **retry-worthy errors** (timeouts, destination unavailable, etc)... perhaps no special treatment is required as the existing retry mechanism will kick in.  
**However** in today's core API, there is one pain point for spout writers. Each spout needs to implement the logic to track inflight tuples and attempt retry on fail(). The implementation is moderately complicated as ACKs/Fails can come in any order. All the spouts have to do the same thing but end up doing slightly differently. Some how retry limits, some dont. This retry logic should ideally be lifted out of the Spout and handled in the API. This new API is a good opportunity to fix this issue.

arunmahadevan:
@roshannaik yes, it can be useful if we can come up with the right way to expose it. We can discuss this further in https://issues.apache.org/jira/browse/STORM-2282

arunmahadevan:
@revans2 @ptgoetz pinging again for review.

revans2:
@arunmahadevan I will try to find time to take a look at it soon.  Sorry for the delay.

examples/storm-starter/src/jvm/org/apache/storm/starter/streams/StateQueryExample.java
```diff
@@ -0,0 +1,110 @@
...
+ *  "jedisPoolConfig":{"host":"localhost", "port":6379,
+ *  "timeout":2000, "database":0, "password":"xyz"}}'
+ * </pre>
+ */
+public class StateQueryExample {
+    public static void main(String[] args) throws Exception {
+        StreamBuilder builder = new StreamBuilder();
+        StreamState<String, Long> ss = builder.newStream(new TestWordSpout(), new ValueMapper<String>(0), 2)
+                .mapToPair(w -> Pair.of(w, 1))
+                .updateStateByKey(0L, (count, val) -> count + 1);
```
revans2: OK It is the inital value.  Makes a lot of since but it might be good to point it out here in the example.

examples/storm-starter/src/jvm/org/apache/storm/starter/streams/BranchExample.java
```diff
@@ -0,0 +1,72 @@
...
+                 * emits random integers by extracting the tuple value at index 0.
+                 */
+                .newStream(new RandomIntegerSpout(), new ValueMapper<Integer>(0))
+                /*
+                 * Split the stream of numbers into streams of
+                 * even and odd numbers. The first stream contains even
+                 * and the second contains odd numbers.
+                 */
+                .branch(x -> (x % 2) == 0,
+                        x -> (x % 2) == 1);
```
arunmahadevan: Tuples that don't match any predicates are dropped. Its in the `branch` Javadoc, have updated the same and called it out as a note.

examples/storm-starter/src/jvm/org/apache/storm/starter/streams/GroupByKeyAndWindowExample.java
```diff
@@ -0,0 +1,109 @@
...
+    public static void main(String[] args) throws Exception {
+        StreamBuilder builder = new StreamBuilder();
+
+        // a stream of stock quotes
+        builder.newStream(new StockQuotes(), new PairValueMapper<String, Double>(0, 1))
+                /*
+                 * The elements having the same key within the window will be grouped
+                 * together and the corresponding values will be merged.
+                 */
+                .groupByKeyAndWindow(SlidingWindows.of(Count.of(6), Count.of(3)))
```
arunmahadevan: Its `PairStream<String, Iterable<Double>>`. Updated the comment and mentioned this.

examples/storm-starter/src/jvm/org/apache/storm/starter/streams/GroupByKeyAndWindowExample.java
```diff
@@ -0,0 +1,109 @@
...
+                .print();
+
+        // a stream of stock quotes
+        builder.newStream(new StockQuotes(), new PairValueMapper<String, Double>(0, 1))
+                /*
+                 * The elements having the same key within the window will be grouped
+                 * together and their values will be reduced using the given reduce function.
+                 */
+                .reduceByKeyAndWindow((x, y) -> (x + y) / 2.0, SlidingWindows.of(Count.of(6), Count.of(3)))
```
arunmahadevan: Updated the example to calculate the 'max' so that its unambiguous.

examples/storm-starter/src/jvm/org/apache/storm/starter/streams/TypedTupleExample.java
```diff
@@ -0,0 +1,64 @@
...
+    @SuppressWarnings("unchecked")
+    public static void main(String[] args) throws Exception {
+        StreamBuilder builder = new StreamBuilder();
+        /**
+         * The spout emits sequences of (Integer, Long, Long). TupleValueMapper can be used to extract fields
+         * from the values and produce a stream of typed tuple (Tuple3<Integer, Long, Long> in this case.
+         */
+        Stream<Tuple3<Integer, Long, Long>> stream = builder.newStream(new RandomIntegerSpout(), TupleValueMappers.of(0, 1, 2));
+
+        PairStream<Long, Integer> pairs = stream.mapToPair(t -> Pair.of(t._2 / 10000, t._1));
```
arunmahadevan: I followed similar pattern (_1, _2 .. ) as in Scala's Tuple classes. We could have methods like getThird, getFourth and so on but it could get cumbersome after some point (e.g. getFourteenth() etc). For Pair I thought it was better name it getFirst and getSecond than _1 and _2.

arunmahadevan:
@revans2 thanks for taking the time to review.

>I really would like to see how triggers and exactly once processing might fit into this. Not that we need it right now, but just that I would like to see how you would fit it in, because I can see how I might do it, but I would like to hear from others about it.

Right now the stateful operations (aggregations, join etc) uses an in-memory map to maintain state and relies on storm's acking mechanism to replay and rebuild the state. For exactly once we can build support at the storm framework level (e.g. enhance the stateful bolts) so that operations like aggregations and join can use the underlying framework state to maintain its state. Initial ideas are here - https://docs.google.com/document/d/1Ew7uFF1UJ6e_zq0t4bM6A9auuEaArviAjjWYSpVFqPY/edit#heading=h.nmf14n. 

Anyways this needs longer discussion. Would be great if you can share your ideas in the doc or in the JIRA https://issues.apache.org/jira/browse/STORM-2155

examples/storm-starter/src/jvm/org/apache/storm/starter/streams/TypedTupleExample.java
revans2: I agree get* can get too long.  Perhaps Pair could have public final _1 and _2 instead of private final first and second?  Again just for self consistency in the API.

arunmahadevan: Updated Pair to have public _1 and _2 and the usage in the example.

arunmahadevan:
@revans2 @HeartSaVioR any further comments on the patch? I think it would be nice to get it merged soon and get feedback from potential users. The apis are marked unstable right now. Will follow up this patch with a how to/primer for users to get started.

HeartSaVioR:
@arunmahadevan I'm just waiting for other committers' review, and IMO once @revans2 is OK to merge we can merge this.

revans2:
@HeartSaVioR and @arunmahadevan sorry I have been busy with other things.  I have not seen anything that I would consider to be blocking and since the API is marked as changing I am +1 to merge it in.

HeartSaVioR:
+1

PR #1694, opened 2016-09-20
STORM-2101: fixes npe in compute-executors in nimbus

abellina:
We observed compute-executors can throw NPE while shutting down a topology sporadically. 

joinMaps (join-maps) will NPE if one of the maps is nil. The only one that can be nil in this case is component->executors.


HeartSaVioR:
+1

PR #1701, opened 2016-09-21
STORM-2120 - Emit to _spoutConfig.outputStreamId

aichow:
Even though KafkaSpout.declareOutputFields declaresStream using outputStreamId (if present), the message gets emitted to a stream matching the Kafka topic it was read from. Looks like it may have been a merge conflict between the fix for STORM-1210 and STORM-1379.


HeartSaVioR:
+1 Nice finding.
`storm-kafka` CI build fails but looks unrelated.

Btw, IMO this bug is worth to file an issue, since tuple is sent to the wrong stream or throwing errors or even not sent.
@aichow Could you please file an issue and add issue name to prefix of PR's title? I can handle it if you mind.

aichow:
@HeartSaVioR Not sure if I tagged the JIRA issue with the right Affects Versions, so feel free to edit as appropriate. Thanks!

PR #1702, opened 2016-09-22
STORM-2127: Storm-eventhubs should use latest amqp and eventhubs-client versions

raviperi:
Eventhub client library has few key fixes that need to be included in storm-eventhub jar.
Fixes:
1. Fix for EventHubSender may throw NullPointerException on .close()
2. EventHubsSender keeps failing to send messages once it hits ConnectionClosedException

### DIFF
--- a/external/storm-eventhubs/pom.xml
+++ b/external/storm-eventhubs/pom.xml
@@ -15,26 +15,23 @@
  See the License for the specific language governing permissions and
  limitations under the License.
 -->
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
-    
+
     <parent>
         <artifactId>storm</artifactId>
         <groupId>org.apache.storm</groupId>
         <version>1.0.3-SNAPSHOT</version>
         <relativePath>../../pom.xml</relativePath>
     </parent>
-    
+
     <artifactId>storm-eventhubs</artifactId>
     <version>1.0.3-SNAPSHOT</version>
     <packaging>jar</packaging>
     <name>storm-eventhubs</name>
     <description>EventHubs Storm Spout</description>
-
-    <properties>
-        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <eventhubs.client.version>0.9.1</eventhubs.client.version>
-    </properties>
+    
     <build>
         <plugins>
             <plugin>
@@ -50,28 +47,29 @@
                 </executions>
                 <configuration>
                     <transformers>
-                        <transformer implementation="org.apache.maven.plugins.shade.resource.ApacheLicenseResourceTransformer">
+                        <transformer
+                            implementation="org.apache.maven.plugins.shade.resource.ApacheLicenseResourceTransformer">
                         </transformer>
                     </transformers>
                     <outputFile>target/${project.artifactId}-${project.version}-jar-with-dependencies.jar</outputFile>
                 </configuration>
-	        </plugin>
+            </plugin>
             <plugin>
-		        <artifactId>maven-antrun-plugin</artifactId>
-		        <executions>
-		          <execution>
-		            <phase>package</phase>
-		            <configuration>
-		              <tasks>
-		                <copy file="src/main/resources/config.properties" tofile="target/eventhubs-config.properties" />
-                    </tasks>
-		            </configuration>
-		            <goals>
-		              <goal>run</goal>
-		            </goals>
-		          </execution>
-		        </executions>
-	        </plugin>
+                <artifactId>maven-antrun-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <phase>package</phase>
+                        <configuration>
+                            <tasks>
+                                <copy file="src/main/resources/config.properties" tofile="target/eventhubs-config.properties" />
+                            </tasks>
+                        </configuration>
+                        <goals>
+                            <goal>run</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
         </plugins>
     </build>
     <dependencies>
@@ -104,10 +102,25 @@
             </exclusions>
         </dependency>
         <dependency>
+            <groupId>org.apache.qpid</groupId>
+            <artifactId>qpid-amqp-1-0-client</artifactId>
+            <version>${qpid.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.qpid</groupId>
+            <artifactId>qpid-amqp-1-0-client-jms</artifactId>
+            <version>${qpid.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.qpid</groupId>
+            <artifactId>qpid-amqp-1-0-common</artifactId>
+            <version>${qpid.version}</version>
+        </dependency>
+        <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
             <version>4.11</version>
             <scope>test</scope>
         </dependency>
-    </dependencies> 
+    </dependencies>
 </project>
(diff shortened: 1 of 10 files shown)


external/storm-eventhubs/pom.xml
```diff
@@ -21,19 +21,19 @@
     <parent>
         <artifactId>storm</artifactId>
         <groupId>org.apache.storm</groupId>
-        <version>1.0.3-SNAPSHOT</version>
+        <version>1.0.2</version>
```
srdo: These versions probably shouldn't be changed

srdo:
This seems fine once the storm versions are reverted, though you should get this on the master branch first. I'd like to mention before you spend too much time on trying to make the spout run reliably, that we've had some reliability problems with the com.microsoft.eventhubs.client:eventhubs-client since the underlying AMQP library isn't being maintained. There's a new eventhub client here based on Apache Proton which is likely to provide a better experience. https://github.com/Azure/azure-event-hubs/tree/master/java

srdo:
Hm, nevermind the new client, it requires Java 8. It's still an option for Storm 2.0 though.

external/storm-eventhubs/pom.xml
```diff
@@ -21,19 +21,19 @@
...
         <relativePath>../../pom.xml</relativePath>
     </parent>
     
     <artifactId>storm-eventhubs</artifactId>
-    <version>1.0.3-SNAPSHOT</version>
+    <version>1.0.2.1</version>
```
HeartSaVioR: Here too. Every modules should follow the Storm version.
Would be better to remove this line as same as other modules.

external/storm-eventhubs/pom.xml
```diff
@@ -21,19 +21,19 @@
...
     <packaging>jar</packaging>
     <name>storm-eventhubs</name>
     <description>EventHubs Storm Spout</description>
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <eventhubs.client.version>0.9.1</eventhubs.client.version>
+        <eventhubs.client.version>1.0.1</eventhubs.client.version>
```
HeartSaVioR: IMHO it might be better to maintain the version of dependencies from the root pom.xml.
Some modules do, and some modules don't, so it's not a mandatory but just food for thought.

external/storm-eventhubs/pom.xml
HeartSaVioR: Yes @srdo is right.

HeartSaVioR:
Btw, IMO, upgrading the dependency is worth to file an issue. 
@raviperi Could you file it and copy description of PR to issue's description?

external/storm-eventhubs/pom.xml
revans2: I agree having them in the root gives everyone a consistent view of dependencies.  It would be nice to move the version there.

revans2:
Is there a similar pull request to master?

external/storm-eventhubs/pom.xml
raviperi: Reverted them. I had initially changed it to validate compatibility with 1.0.2 binaries.

external/storm-eventhubs/pom.xml
raviperi: Got it. Reverted it to match the version convention.

raviperi:
The latest PR addresses the above comments
. Move dependencies to parent pom
. Revert changes to versions to reflect naming convention.

I will be sending a different PR to master branch, that includes the above changes.

external/storm-eventhubs/src/main/java/org/apache/storm/eventhubs/spout/EventHubSpout.java
```diff
@@ -121,8 +121,8 @@ public void preparePartitions(Map config, int totalTasks, int taskIndex, SpoutOu
         zkEndpointAddress = sb.toString();
       }
       stateStore = new ZookeeperStateStore(zkEndpointAddress,
-          (Integer)config.get(Config.STORM_ZOOKEEPER_RETRY_TIMES),
-          (Integer)config.get(Config.STORM_ZOOKEEPER_RETRY_INTERVAL));
+          Integer.parseInt(config.get(Config.STORM_ZOOKEEPER_RETRY_TIMES).toString()),
```
srdo: Does this compile? ZookeeperStateStore takes ints in these parameters. Also it doesn't seem backwards compatible to go from having these properties be Integer in config to having them be String.

external/storm-eventhubs/pom.xml
```diff
@@ -1,113 +1,120 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!--
- Licensed to the Apache Software Foundation (ASF) under one or more
- contributor license agreements.  See the NOTICE file distributed with
- this work for additional information regarding copyright ownership.
- The ASF licenses this file to You under the Apache License, Version 2.0
- (the "License"); you may not use this file except in compliance with
- the License.  You may obtain a copy of the License at
+<!-- Licensed to the Apache Software Foundation (ASF) under one or more contributor 
+	license agreements. See the NOTICE file distributed with this work for additional
```
srdo: Nitpick: Your editor seems to have replaced spaces with tabs, which has made the diff much larger than it needs to be. Could you reformat these files back to something closer to the original formatting?

external/storm-eventhubs/src/main/java/org/apache/storm/eventhubs/spout/StringEventDataScheme.java
```diff
@@ -0,0 +1,69 @@
...
+  private static final long serialVersionUID = 1L;
+
+  @Override
+  public List<Object> deserialize(Message message) {
+    final List<Object> fieldContents = new ArrayList<Object>();
+
+    for (Section section : message.getPayload()) {
+      if (section instanceof Data) {
+        Data data = (Data) section;
+        fieldContents.add(new String(data.getValue().getArray()));
```
srdo: I don't remember the details of these message formats, but you should probably specify an encoding for this String

external/storm-eventhubs/src/main/java/org/apache/storm/eventhubs/spout/EventHubSpout.java
```diff
@@ -152,7 +152,7 @@ public void open(Map config, TopologyContext context, SpoutOutputCollector colle
     try {
       preparePartitions(config, totalTasks, taskIndex, collector);
     } catch (Exception e) {
-      logger.error(e.getMessage());
+	  collector.reportError(e);
```
srdo: Someone else can correct me here, but I don't think collector.reportError puts anything in the log, so it might be nice to have both this and logger.error

revans2: It does log it, and puts it on the UI.

https://github.com/apache/storm/blob/28563ece16274bcd61827f38e960ce2d27a57dea/storm-core/src/jvm/org/apache/storm/executor/error/ReportError.java#L61

There is throttling around it so if we get too many errors too quickly they will not all be logged and not all of them will be written to ZK.

srdo: Thanks @revans2

raviperi:
Created a new PR https://github.com/apache/storm/pull/1717 for the master branch.

Created a new JIRA that reflects the  need for the aboce work.
STORM-2127 - Storm-eventhubs should use latest amqp and eventhubs-client versions

external/storm-eventhubs/src/main/java/org/apache/storm/eventhubs/spout/EventHubSpout.java
raviperi: It compiles :) 

The issue was that the values were being parsed as Long, and not Integer.
Instead of doing an explicit cast, I wanted to be safe and parse the numeric value.

srdo: Sorry, misread this bit. Read it as parseInt(conf).toString instead of parseInt(conf.toString) :)

external/storm-eventhubs/src/main/java/org/apache/storm/eventhubs/spout/StringEventDataScheme.java
raviperi: Yes, It is an important change, and I have that change in the pipeline.
I will send a different PR for that change, as I want to make sure that it does not break any assumptions client code and our own SCP.Net framework make.

raviperi:
@revans2 Please let me know if you have further comments.

revans2:
I am not really familiar with EventHub all that much so I don't really feel all that confident in reviewing it.  I mostly jumped on to answer a few generic questions.

harshach:
+1. @srdo do you have any further comments pending.

srdo:
No, this seems fine. +1

harshach:
@raviperi do you have JIRA filed here for this https://issues.apache.org/jira/browse/STORM/

harshach:
@raviperi sorry for nitpick. we need the format of the title to be
" STORM-2127: Storm-eventhubs should use latest amqp and eventhubs-client versions"

raviperi:
@harshach  
Updated title as per your recommendation.

PR #1704, opened 2016-09-22
STORM-2111 [Storm SQL] support 'LIKE' and 'SIMILAR TO'

HeartSaVioR:
Note: This patch is on top of STORM-2089 in order to reduce any merge conflict and upmerging.

Reviewers may want to only take a look into 3e05056.

I also addressed some ExprCompiler's bug: doesn't wrap null as type in elvis operator which raises type mismatch.


HeartSaVioR:
Tests of storm-sql-core failed because of VM crashing. UT passes on my local dev.

HeartSaVioR:
Closing this since STORM-2125 (#1714) covers this.

PR #1697, opened 2016-09-20
STORM-2018: Supervisor V2

revans2:
Still need to do some more manual testing but the unit tests passed for me.

PR #1705, opened 2016-09-22
STORM-2117 Supervisor V2 with local mode extracts resources directory to the wrong directory

HeartSaVioR:
* it extracts the resources directory to topology root directory instead of temporary directory
* it should be extracted to temporary directory since we will move that directory to topology root directory

It didn't make unit tests failing, but I ran the tests from IntelliJ manually, IntelliJ adds some jars which contains resources directory, and this happens.

@revans2 Please take a look. Thanks!


revans2:
+1

I'll pull this into #1697 too

PR #1706, opened 2016-09-22
[STORM-2118] A few fixes for storm-sql standalone mode

arunmahadevan:
1. Cast the result, accumulator and value types correctly
2. Support aggregate functions with more than one argument

### DIFF
--- a/external/sql/storm-sql-runtime/src/test/org/apache/storm/sql/TestUtils.java
+++ b/external/sql/storm-sql-runtime/src/test/org/apache/storm/sql/TestUtils.java
@@ -67,10 +67,17 @@ public class TestUtils {
       return new PriorityQueue<>();
     }
     public static PriorityQueue<Integer> add(PriorityQueue<Integer> accumulator, Integer n, Integer val) {
-      if (accumulator.size() >= n) {
-        accumulator.remove();
+      if (n <= 0) {
+        return accumulator;
       }
-      accumulator.add(val);
+      if (accumulator.size() >= n) {
+        if (val > accumulator.peek()) {
+          accumulator.remove();
+          accumulator.add(val);
+        }
+      } else {
+        accumulator.add(val);
+      }
       return accumulator;
     }
     public static List<Integer> result(PriorityQueue<Integer> accumulator) {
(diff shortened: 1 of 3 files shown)


external/sql/storm-sql-runtime/src/test/org/apache/storm/sql/TestUtils.java
```diff
@@ -60,6 +62,25 @@ public static String result(String accumulator) {
   }
 
 
+  public static class TopN {
+    public static PriorityQueue<Integer> init() {
+      return new PriorityQueue<>();
+    }
+    public static PriorityQueue<Integer> add(PriorityQueue<Integer> accumulator, Integer n, Integer val) {
+      if (accumulator.size() >= n) {
```
HeartSaVioR: Logic need to be changed to add val first, and remove the lowest `if accumulator.size() > n` because val could be lower than lowest value of priority queue.
Suppose the input `3, 2, 1`, accumulator will be changed to 3 -> 2, 3 -> 1, 3 (should be 2, 3).

We could even do nothing when `(n > 0 && accumulator.size() >= n && accumulator.peek() >= val)`.

arunmahadevan: Thanks for catching. It should be 
```java
if (accumulator.size() >= n) {
  if (val > accumulator.peek()) {
    accumulator.remove();
    accumulator.add(val);
  }
} else {
  accumulator.add(n);
}
````

HeartSaVioR: One missed spot is that it should do nothing if n is 0. Now if n is 0, it will behave same as top 1. n should be greater than 0 anyway but we can do some defensive programming.
Other then that it looks good.

arunmahadevan: yes, it will throw NPE if n is 0, since it tries to peek from empty queue. will address this.

arunmahadevan:
@HeartSaVioR addressed comments.

external/sql/storm-sql-runtime/src/test/org/apache/storm/sql/TestUtils.java
HeartSaVioR: Yes you're right. It'll throw NPE.

HeartSaVioR:
@arunmahadevan Thanks for the quick fix. +1

PR #1710, opened 2016-09-23
STORM-1546: Adding Read and Write Aggregations for Pacemaker to make it HA compatible

knusbaum:


storm-core/src/jvm/org/apache/storm/pacemaker/PacemakerClientPool.java
```diff
@@ -0,0 +1,114 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentLinkedQueue;
+import org.apache.storm.Config;
+import org.apache.storm.generated.HBMessage;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class PacemakerClientPool {
+
+        private static final Logger LOG = LoggerFactory.getLogger(PacemakerClientPool.class);
```
d2r: check indentation

storm-core/src/jvm/org/apache/storm/pacemaker/PacemakerClientPool.java
```diff
@@ -0,0 +1,114 @@
...
+
+    private PacemakerClient getClientForServer(String server) {
+        PacemakerClient client = clientForServer.get(server);
+        if(client == null) {
+            client = new PacemakerClient(config, server);
+            clientForServer.put(server, client);
+        }
+        return client;
+    }
+
```
d2r: minor: extra line

storm-core/src/jvm/org/apache/storm/cluster/PaceMakerStateStorage.java
```diff
@@ -160,13 +179,23 @@ public void set_worker_hb(String path, byte[] data, List<ACL> acls) {
         int retry = maxRetries;
         while (true) {
             try {
+                HashSet<String> retSet = new HashSet<>();
+                int latest_time_secs = 0;
```
d2r: `latest_time_sec` not used?

storm-core/src/jvm/org/apache/storm/cluster/PaceMakerStateStorageFactory.java
```diff
@@ -18,48 +18,22 @@
 package org.apache.storm.cluster;
 
 import org.apache.storm.pacemaker.PacemakerClient;
+import org.apache.storm.pacemaker.PacemakerClientPool;
```
d2r: Remove PacemakerClient import, as it's not used?

storm-core/src/jvm/org/apache/storm/pacemaker/PacemakerClientHandler.java
```diff
@@ -69,7 +69,7 @@ else if(evm instanceof HBMessage) {
 
     @Override
     public void exceptionCaught(ChannelHandlerContext ctx, ExceptionEvent event) {
-        LOG.error("Connection to pacemaker failed", event.getCause());
+        LOG.error("Connection to pacemaker failed. Trying to reconnect {}", event.getCause().getMessage());
```
d2r: We don't want to log the stack here, only the message?

storm-core/test/jvm/org/apache/storm/PaceMakerStateStorageFactoryTest.java
```diff
@@ -20,13 +20,22 @@
...
-public class PaceMakerStateStorageFactoryTest {
+import java.util.HashMap;
+import java.util.List;
+import java.util.ArrayList;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
-    private PacemakerClient clientProxy;
+public class PaceMakerStateStorageFactoryTest {
+    private static Logger LOG = LoggerFactory.getLogger(PaceMakerStateStorageFactoryTest.class);
```
d2r: We don't seem to be logging anything in this test. Do we need the Logger?

storm-core/src/jvm/org/apache/storm/pacemaker/PacemakerClientHandler.java
knusbaum: I don't think we want the stack. It ends up absolutely filling Nimbus logs with relatively useless messages when a Pacemaker node goes down. I'm open to ideas, though, because a trace for an NPE or something would be nice.

knusbaum:
Addressed your comments. Not sure why they have not disappeared.

storm-core/src/jvm/org/apache/storm/Config.java
```diff
@@ -926,10 +926,10 @@
     public static final String UI_HTTPS_NEED_CLIENT_AUTH = "ui.https.need.client.auth";
 
     /**
-     * The host that Pacemaker is running on.
+     * The list of servers that Pacemaker is running on.
      */
-    @isString
-    public static final String PACEMAKER_HOST = "pacemaker.host";
+    @isStringList
+    public static final String PACEMAKER_SERVERS = "pacemaker.servers";
```
revans2: This is a breaking change that would be good to go back to 1.x as well.  I am OK with this, because I don't think anyone is really using pacemaker, but I would like to hear others opinions on it.

storm-core/src/jvm/org/apache/storm/pacemaker/PacemakerClientHandler.java
revans2: If we know which exceptions we want to ignore, perhaps just IOExceptions, then we can explicitly blacklist them, and even log them under warn, instead of error.

storm-core/src/jvm/org/apache/storm/pacemaker/PacemakerClientPool.java
```diff
@@ -0,0 +1,113 @@
...
+    
+    private ConcurrentHashMap<String, PacemakerClient> clientForServer = new ConcurrentHashMap<>();
+    private ConcurrentLinkedQueue<String> servers;
+    private Map config;
+
+    public PacemakerClientPool(Map config) {
+        this.config = config;
+        List<String> serverList = (List<String>)config.get(Config.PACEMAKER_SERVERS);
+        if(serverList == null) {
+            serverList = new ArrayList<String>();
```
revans2: nit: I don't think `<String>` is needed just `<>`

revans2: and in the following few lines too.

storm-core/src/jvm/org/apache/storm/pacemaker/PacemakerClientPool.java
```diff
@@ -0,0 +1,113 @@
...
+            return getWriteClient().send(m);
+        } catch (Exception e) {
+            rotateClients();
+            throw e;
+        }
+    }
+
+    public List<HBMessage> sendAll(HBMessage m) throws PacemakerConnectionException {
+        List<HBMessage> responses = new ArrayList<HBMessage>();
+        LOG.info("Using servers: {}", servers);
```
revans2: Perhaps this should be debug (especially if we are worried about filling the logs)

storm-core/src/jvm/org/apache/storm/cluster/PaceMakerStateStorage.java
```diff
@@ -159,10 +162,6 @@ public void set_worker_hb(String path, byte[] data, List<ACL> acls) {
                         ret = details;
                     }
                 }
-                if(ret == null) {
-                    throw new HBExecutionException("Failed to get a response.");
-                }
-                LOG.debug("Successful get_worker_hb");
                 return ret;
```
revans2: Do we need the ret == null check before this?  Is it possible to get more then one response and all of them are bogus?

knusbaum: Yes, I wasn't quite sure how we wanted to handle this, NULL responses are valid, and right now the code treats mismatched responses as if they were NULL. We can distinguish between them if we want the behavior to differ whether the pacemakers returned NULL details or returned an invalid message. Throwing on null is not what we wanted, though, since it will cause Nimbus to crash when nimbus tries to read a topology's heartbeats before it has started sending them.

Maybe we want to add some logic so that if ALL pacemakers return an invalid response, then we throw?

revans2: I think that is best, if all of them return something invalid or we get nothing back at all then we throw.

storm-core/src/jvm/org/apache/storm/Config.java
HeartSaVioR: I'm in favor of backward compatibility change for this, like nimbus.host and nimbus.seeds. For 2.0 we don't need to keep nimbus.host and pacemaker.host but 1.x still need to have it.

knusbaum: @HeartSaVioR Yep, the 1.x-branch PR will leave pacemaker.host in, but have it deprecated.

knusbaum:
@d2r @revans2 Any additional comments?

revans2:
+1

knusbaum:
Documentation updated.

d2r:
Thanks for doc update. +1

PR #1711, opened 2016-09-23
Added ByteKeyValueScheme class.

asmaier:
It is a common use case to send and receive byte key and values via Kafka. Unfortunately the storm-kafka package misses support for this. So I added a ByteKeyValueScheme in analogy to the StringKeyValueScheme class.

### DIFF
--- a/external/storm-kafka/src/jvm/org/apache/storm/kafka/ByteKeyValueScheme.java
+++ b/external/storm-kafka/src/jvm/org/apache/storm/kafka/ByteKeyValueScheme.java
@@ -1,0 +1,33 @@
+package org.apache.storm.kafka;
+
+
+import java.nio.ByteBuffer;
+import java.util.ArrayList;
+import java.util.List;
+import org.apache.storm.kafka.bolt.mapper.FieldNameBasedTupleToKafkaMapper;
+import org.apache.storm.spout.RawScheme;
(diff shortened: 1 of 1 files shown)


ptgoetz:
@asmaier Could you create a JIRA ticket for this and update the title so it's prefixed with the JIRA ID?

external/storm-kafka/src/jvm/org/apache/storm/kafka/ByteKeyValueScheme.java
```diff
@@ -0,0 +1,33 @@
+package org.apache.storm.kafka;
```
ptgoetz: Need Apache license header.

asmaier:
I created the JIRA ticket STORM-2123 and added an Apache license header.

HeartSaVioR:
@asmaier 
Could you make implementation similar to StringKeyValueScheme? It would help us to maintain twos easier.

asmaier:
Unfortunately there is a bug in the implementation of StringKeyValueScheme: https://issues.apache.org/jira/browse/STORM-2121
I believe this problem should be fixed first, before changing the implementation of ByteKeyValueScheme.

HeartSaVioR:
@asmaier Makes sense, and I missed that. Could you also provide a patch for also STORM-2121 if you don't mind? Thanks in advance.

PR #1712, opened 2016-09-23
STORM-2122: Cache dependency data, and serialize reading of the data.

revans2:


revans2:
The travis failure here is an odd one, it looks like it failed because some integration tests that didn't run before are now running and took so long travis timed them out.

HeartSaVioR:
+1

HeartSaVioR:
I've tested with --jars and --artifacts options. Works like a charm.

PR #1709, opened 2016-09-23
STORM-2116 [Storm SQL] Support 'CASE' statement

HeartSaVioR:
NOTE: This patch is on top of STORM-2089 and STORM-2111, and STORM-2113.

```
SELECT CASE WHEN NAME IN ('a', 'abc', 'abcde') THEN UPPER('a') 
WHEN UPPER(NAME) = 'AB' THEN 'b' ELSE {fn CONCAT(NAME, '#')} END FROM FOO
```
(`{fn CONCAT(a, b)}` is same to `a || b`. Please refer [SQL reference](http://calcite.apache.org/docs/reference.html) page on Calcite web site.)

`CASE` statement in above sql statement is translated to Java code as

```
import org.apache.storm.tuple.Values;
java.lang.String t1 = null;
// WHEN #0 THEN #1
java.lang.Boolean t2 = Boolean.FALSE;
java.lang.Boolean anyNull_t3 = Boolean.FALSE;
// operand #0
java.lang.String t5 = (java.lang.String)(_data.get(1));
final java.lang.String t6 = (java.lang.String) "a";
java.lang.Boolean t4 = org.apache.storm.sql.runtime.StormSqlFunctions.eq(t5,t6);
final java.lang.Boolean t7 = t4;
if ((t7 != null) && (t7)) { t2 = Boolean.TRUE; anyNull_t3 = Boolean.FALSE; }
else {
// updating null occurrence with operand #0
if ((t7 == null) && !(anyNull_t3)) { anyNull_t3 = Boolean.TRUE; }
// operand #1
java.lang.String t9 = (java.lang.String)(_data.get(1));
final java.lang.String t10 = (java.lang.String) "abc";
java.lang.Boolean t8 = org.apache.storm.sql.runtime.StormSqlFunctions.eq(t9,t10);
final java.lang.Boolean t11 = t8;
if ((t11 != null) && (t11)) { t2 = Boolean.TRUE; anyNull_t3 = Boolean.FALSE; }
else {
// updating null occurrence with operand #1
if ((t11 == null) && !(anyNull_t3)) { anyNull_t3 = Boolean.TRUE; }
// operand #2
java.lang.String t13 = (java.lang.String)(_data.get(1));
final java.lang.String t14 = (java.lang.String) "abcde";
java.lang.Boolean t12 = org.apache.storm.sql.runtime.StormSqlFunctions.eq(t13,t14);
final java.lang.Boolean t15 = t12;
[...]

PR #1707, opened 2016-09-22
STORM-2113 [Storm SQL] fix 'OR' and 'AND' operators handle more than 2 operands

HeartSaVioR:
NOTE: This is on top of STORM-2089 and STORM-2111 since this requires the ExprCompiler bugfix introduced on STORM-2111.

After the patch 'IN' and 'NOT IN' operator works properly. It also respects 'short circuit'.

PR #1714, opened 2016-09-26
STORM-2125 Use Calcite's implementation of Rex Compiler

HeartSaVioR:
Note: This patch is on top of STORM-2089.

This can cover STORM-2111(#1704), STORM-2113(#1707), STORM-2114, STORM-2116(#1709).

* Upgrade Calcite to latest 1.9.0
* borrow JaninoRexCompiler and modify to return Expression instead of Scalar (RexNodeToBlockStatementCompiler)
  * since we need to pass code block to Bolts, not executable method implementation on classloader
* modify codebase to use RexNodeToBlockStatementCompiler
* add some feature validation unit tests
* remove ExprCompiler and its test

Btw, Calcite Rex Compiler assigns the result of ITEM() to Object, so comparing result to other constant just fails. CAST() works but if any of the result of ITEM() is null or out of bound (array), it throws Exception at runtime. So only using them to select field would be more safe (but still a change when accessing array via index out of bounds).

I sent a mail to Calcite dev to ask this strange behavior.

http://mail-archives.apache.org/mod_mbox/calcite-dev/201609.mbox/%3CCAF5108ivAyETJNkRNBdEpuLzHWkUbQ5Rv7TDGPBPOp7pUNybGA%40mail.gmail.com%3E

Btw, I think it's not a big matter so even though it can't be fixed, instead of addressing it ASAP, I would like to document it as limitation. Other frameworks don't seem to support this, too.

@arunmahadevan Please review and comment, and also please comment what do you think about above behavior. Thanks in advance.


HeartSaVioR:
There's an another issue: when to initialize DataContext.

Calcite utilizes the variables in DataContext. and there're datetime related variables in DataContext.

The thing is, SQL standard says that functions such as CURRENT_TIMESTAMP return the same value throughout the query. Since we're querying in stream, when to initialize DataContext defines the boundary to show same datetime across rows. With joining stream, the issue is more complicated.

For now I just initialize DataContext at the query compilation phase and pass it to components, so that the value of current datetime is static (don't change) across workers in topology. If we would like to make it dynamic, we need to discuss.

external/sql/storm-sql-core/src/test/org/apache/storm/sql/compiler/backends/trident/TestPlanCompiler.java
```diff
@@ -255,6 +265,198 @@ public void testUdaf() throws Exception {
...
+    Map<String, Integer> map = ImmutableMap.of("b", 2, "c", 4);
+    Map<String, Map<String, Integer>> nestedMap = ImmutableMap.of("a", map);
+    Assert.assertArrayEquals(new Values[]{new Values(2, 4, nestedMap, Arrays.asList(100, 200, 300))}, getCollectedValues().toArray());
+  }
+
+  @Test
+  public void testDateKeywords() throws Exception {
+    int EXPECTED_VALUE_SIZE = 1;
+    String sql = "SELECT " +
+            "LOCALTIME, CURRENT_TIME, LOCALTIMESTAMP, CURRENT_TIMESTAMP, CURRENT_DATE " +
```
arunmahadevan: Is there some way to output formatted current_date like '2016-09-26' instead of an integer? May be we should also document what are the different date time and other functions supported by storm-sql.

pom.xml
```diff
@@ -264,7 +264,7 @@
         <junit.version>4.11</junit.version>
         <metrics-clojure.version>2.5.1</metrics-clojure.version>
         <hdrhistogram.version>2.1.7</hdrhistogram.version>
-        <calcite.version>1.4.0-incubating</calcite.version>
+        <calcite.version>1.9.0</calcite.version>
         <janino.version>2.7.8</janino.version>
```
arunmahadevan: If we dont use janino anymore we should remove this.

arunmahadevan:
Can you also document the list of functions that are now available since we are directly using calicite to generate the java code? Does it support all the standard SQL functions?

external/sql/storm-sql-core/src/test/org/apache/storm/sql/TestStormSql.java
```diff
@@ -174,26 +173,42 @@ public void testExternalDataSource() throws Exception {
   public void testExternalDataSourceNested() throws Exception {
     List<String> stmt = new ArrayList<>();
     stmt.add("CREATE EXTERNAL TABLE FOO (ID INT, MAPFIELD ANY, NESTEDMAPFIELD ANY, ARRAYFIELD ANY) LOCATION 'mocknested:///foo'");
-    stmt.add("SELECT STREAM ID, MAPFIELD, NESTEDMAPFIELD, ARRAYFIELD " +
+    stmt.add("SELECT STREAM ID, MAPFIELD['c'], NESTEDMAPFIELD, ARRAYFIELD " +
```
HeartSaVioR: MAPFIELD['a'] doesn't exist. NESTEDMAPFIELD['a'] exists.

external/sql/storm-sql-core/src/test/org/apache/storm/sql/TestStormSql.java
```diff
@@ -174,26 +173,42 @@ public void testExternalDataSource() throws Exception {
...
                      "FROM FOO " +
-                     "WHERE NESTEDMAPFIELD['a']['b'] = 2 AND ARRAYFIELD[1] = 200");
+                     "WHERE CAST(MAPFIELD['b'] AS INTEGER) = 2 AND CAST(ARRAYFIELD[1] AS INTEGER) = 200");
```
HeartSaVioR: 1. 
Yes CAST is required because the type of looked up value is Object. While we're saying based on only type on runtime, we might need to consider the type on schema.

2. 
Implicit cast is not possible because we can't determine the value type on compile time, especially we use ANY to represent its type.

external/sql/storm-sql-core/src/test/org/apache/storm/sql/TestStormSql.java
```diff
@@ -174,26 +173,42 @@ public void testExternalDataSource() throws Exception {
...
 
-  @Test
+  @Test(expected = RuntimeException.class)
   public void testExternalNestedInvalidAccess() throws Exception {
     List<String> stmt = new ArrayList<>();
+    // this triggers java.lang.RuntimeException: Cannot convert null to int
     stmt.add("CREATE EXTERNAL TABLE FOO (ID INT, MAPFIELD ANY, NESTEDMAPFIELD ANY, ARRAYFIELD ANY) LOCATION 'mocknested:///foo'");
     stmt.add("SELECT STREAM ID, MAPFIELD, NESTEDMAPFIELD, ARRAYFIELD " +
-                     "FROM FOO " +
-                     "WHERE NESTEDMAPFIELD['a']['b'] = 2 AND ARRAYFIELD['a'] = 200");
```
HeartSaVioR: MAPFIELD['a'] doesn't exist so this is a test for invalid access, map type.
Btw, I'll keep earlier test case but remove nested access from where since it just makes test failing before testing invalid array access.

external/sql/storm-sql-core/src/test/org/apache/storm/sql/compiler/backends/standalone/TestPlanCompiler.java
```diff
@@ -71,9 +71,9 @@ public void testLogicalExpr() throws Exception {
 
   @Test
   public void testNested() throws Exception {
-    String sql = "SELECT ID, MAPFIELD, NESTEDMAPFIELD, ARRAYFIELD " +
+    String sql = "SELECT ID, MAPFIELD['c'], NESTEDMAPFIELD, ARRAYFIELD " +
             "FROM FOO " +
-            "WHERE NESTEDMAPFIELD['a']['b'] = 2 AND ARRAYFIELD[1] = 200";
```
HeartSaVioR: Like I said before (PR desc and comment), with Calcite Janino compiler it is not possible. (even with explicit type information)
Let's wait for the response from Calcite dev@ list.

external/sql/storm-sql-core/src/test/org/apache/storm/sql/compiler/backends/trident/TestPlanCompiler.java
HeartSaVioR: There should be other function available, and we can provide some if it doesn't. Agreed on documenting supporting functions / features / behaviors.

pom.xml
HeartSaVioR: We will follow Janino from the Calcite dependency. I'll check it again and delete if it's not used anymore.

external/sql/storm-sql-runtime/src/jvm/org/apache/calcite/interpreter/StormDataContext.java
```diff
@@ -0,0 +1,79 @@
...
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.calcite.interpreter;
```
HeartSaVioR: I applied a trick because of accessor issue, and this seems to be improperly moved. I'll check again.

external/sql/storm-sql-core/src/test/org/apache/storm/sql/compiler/TestCompilerUtils.java
```diff
@@ -153,6 +158,29 @@ public static CalciteState sqlOverNestedTable(String sql)
                 .field("MAPFIELD", SqlTypeName.ANY)
                 .field("NESTEDMAPFIELD", SqlTypeName.ANY)
                 .field("ARRAYFIELD", SqlTypeName.ANY)
+
+                // TODO: replace field definitions with belows while Calcite can handle properly
```
HeartSaVioR: For me, type safely is more important than convenience. That's why we use static typing language, and also trying to restrict more by generic and so. IMO, we shouldn't rely Calcite to handle operators with 'ANY' type properly. If it can handle operators properly, it's more than essential feature.

As I commented earlier we can't keep the current syntax until we modify Calcite, or back to ExprCompiler which I really don't want to.
For now I'll try to CAST to 'ANY' explicitly and see it makes nested multi-level access possible.

HeartSaVioR:
I don't want to get blocked by non-SQL standard support unless we have clear use case or clear way to implement it. If my understanding is right, both Spark and Flink seems not support this as well, and also Samza SQL too.

When this gets merged into master, I expect Storm SQL will support most of SQL reference page on Calcite. We can add tests for checking features availability and document supported features, and it could take some time.

external/sql/storm-sql-core/src/test/org/apache/storm/sql/compiler/TestCompilerUtils.java
HeartSaVioR: From what I'm experimenting, nested map access seems supported, but its type should be ANY instead of MAP. I'll remove this TODO for now.

external/sql/storm-sql-core/src/test/org/apache/storm/sql/compiler/backends/standalone/TestPlanCompiler.java
HeartSaVioR: `CAST(NESTEDMAPFIELD['a']['b'] AS INTEGER) = 2 AND CAST(ARRAYFIELD[1] AS INTEGER) = 200` seems work. I think this is enough.

external/sql/storm-sql-runtime/src/jvm/org/apache/calcite/interpreter/StormDataContext.java
HeartSaVioR: Only StormContext needs to be placed under org.apache.calcite.interpreter. Will move this to org.apache.storm.sql.calcite.

pom.xml
HeartSaVioR: OK we can delete this.

HeartSaVioR:
@arunmahadevan 

Let me summary regarding nested map/array support.

* accessing nested element in nested map/array works without modifying Calcite.
  * it should be defined as ANY. Specifying MAP's value type to MAP doesn't seem to work. I don't understand this behavior but it works anyway.
* We need to wrap value with CAST if we need to use value in expression. 
  * For example, `MAPFIELD['a'] = 2` doesn't work but `CAST(MAPFIELD['a'] AS INTEGER) = 2` works.
  * Using value in projection (fields of SELECT) doesn't need to be wrapped with CAST.
* NOTE: this is NOT a safe operation, and errors are shown in runtime.
  * If value is null in any chance, comparing it to other will **throw RuntimeException** in runtime.
    * value itself could be null.
    * value could be null when we access array with wrong index or map with non-exist key. 
  * If we access array with index out of bound, it will **throw ArrayIndexOutOfBoundException** in runtime.

So there's a way to support current feature with adding some inconveniences (CAST), but it is no longer safe operation. Since user can't handle errors manually, this can make worker continuously crashed.

I asked Calcite dev. group about this and didn't get answer yet.

HeartSaVioR:
@arunmahadevan Other than that I addressed your review comments. Thanks for the thoughtful review.

HeartSaVioR:
I'm trying to address our concern of nested map / array since Calcite community (Julian) says it sounds like a bug.

Addressing it from CALCITE-1386 : https://github.com/apache/calcite/pull/283.
With applying CALCITE-1386, we could make tests back to origin. 
(It still need to wrap CAST when detailed type information is not available - I mean ANY)

Below is the commit which works with CALCITE-1386 patch.
https://github.com/HeartSaVioR/storm/commit/b2bcf9baebb8d9bd878c3d8a74c0d366ded7d399

@arunmahadevan Please take a look into above commit. Thanks!

HeartSaVioR:
Just a head up: my patch for Calcite (https://github.com/apache/calcite/pull/283) passes the review and in progress of merging. 
(handling some comment fixes.)
So I'm expecting that we can get runtime safety with next version of Calcite.

Do we want to hold on waiting next version of Calcite? Or just ship this as it is for now, and I will submit another pull request following up the change of Calcite?

HeartSaVioR:
Another head up: my patch got merged to Calcite master, and Calcite dev@ is discussing about releasing new minor version soon.

arunmahadevan:
@HeartSaVioR  +1 on this change. If the minor version of the calcite with your fixes can get released soon, it makes sense to wait for that so we can keep the current semantics for querying nested arrays/maps.

Can you also file a follow up JIRA to document the functions that are currently available within storm-sql and may be some examples for users to get started.

HeartSaVioR:
@arunmahadevan 
Regarding Calcite release, Calcite community is waiting for one issue to get resolved before starting RC. I guess RC will be available in this week or so. 
I'll watch Calcite community and do some change (Calcite version) and let you know again. 
(I didn't think about this but I need to modify this PR slightly because of Calcite version. ;) )

Regarding documenting available features, actually I didn't test all of SQL reference on Calcite. :) We can add test cases to see what features Storm SQL is able to provide, and adopt SQL reference page with some modifications. I'll file a follow up JIRA.

Regarding examples, yes, we don't have any useful documents or examples so it should be next work of Storm SQL. I'll file a follow up JIRA separately.

Thanks for reviewing!

HeartSaVioR:
Filed [STORM-2135](https://issues.apache.org/jira/browse/STORM-2135) and [STORM-2136](https://issues.apache.org/jira/browse/STORM-2136). I'd like to get some helps on these issues, but don't mind to work on these by myself.

HeartSaVioR:
Added full of scalar function tests which Calcite provides. Calcite 1.9.0 has some bugs so not all functions are working well, but it can be fixed from Calcite side.
(I left comments to test codes about bugs I found.)
Once we track and document the bugs to our doc. page, I think it would be OK.

HeartSaVioR:
Addressed documenting covered feature from other PR: #1725

HeartSaVioR:
FYI: 
The vote for 1.10.0 RC1 is just started so we can expect Calcite 1.10.0 to next week.
Submitted patches for recently found bugs to Calcite: CALCITE-1418 and CALCITE-1419.
Above issues are marked for 1.11.0, but it's not a thing to rush since at least we are documenting this for `known bug` from STORM-2135.

HeartSaVioR:
Just found and fixed bugs regarding aggregate calls and timezone issue in current_date.
Now aggregate also needs new Trident map API - STORM-2072 (#1663) to reduce the steps.
I'll address this after STORM-2072 is merged to master.

HeartSaVioR:
Calcite 1.10.0 RC1 vote passed and artifact is available to Maven.
Applied HeartSaVioR@b2bcf9b with changing Calcite version from 1.10.0-SNAPSHOT to 1.10.0.
Also rebased and squashed the commits.

@arunmahadevan I guess this would be the last review for this patch. Please take a look again. Thanks!

arunmahadevan:
@HeartSaVioR with this change we are going to be heavily dependent on the calcite's code generator to translate expressions correct?. Just thinking if we should have an option to make the code generation pluggable with an option to plug in our code generator as well. The calcite's code generator can be the default.

Just thinking if in future we want to do some optimizations around the generated code or have our own implementation for a built in function or want to fix some bugs in the generated code without waiting for a calcite release etc, it may not be possible with the current patch. Let me know what you think.

HeartSaVioR:
@arunmahadevan 
I'm not sure we can have our own expression compiler which supports features competitive to Calcite. It reminds me why this change is necessary.

We have been leaving expression compiler which has lots of unimplemented things and also bugs. It was pushed more than 6 months before and only few of contributors are taking care of. We have lots of milestones (not a single task) to do even without Storm SQL, like Unified API, extending Window feature (session, etc.), Beam integration, Java porting, Metrics, and so on. Even I have been focusing to work on Storm SQL, I give up keeping expression compiler because of having less eyes to watch, and error-prone way of building code block (string concatenation). We could save amount of time if we replace this faster, and it will save another amount of time.

Btw, normally we're often worry about dependency if dependency's activity is low or release interval is too long. But Calcite's release interval for minor release is even faster than Storm's bugfix release.
Calcite released 1.7.0 from Mar 22, 1.8.0 from Jun 13, and 1.9.0 from Sep 22, and 1.10.0 today. I have already fixed two bugs in unit tests and will be shipped to 1.10.1 or 1.11.0.
(Yes in this case we need to wait on that.)

I'm not on the fence of making it pluggable, but just wondering we have comparable plugs on that, and we can make an effort for that. Would you want to be a sponsor on current expression compiler?

arunmahadevan:
@HeartSaVioR I suggested to keep an option to plug our code generator. If you think calcites code generator (linq4j) addresses all the current cases then lets go with that. We can revisit having our own code generator within storm-sql once we have more contributors/focussed effort around sql.

arunmahadevan:
+1

HeartSaVioR:
@arunmahadevan Yes ideally we really want to have it. I just said that ExprCompiler seems not having a value to keep and maintain. Agreed with you. Thanks for reviewing!

PR #1717, opened 2016-09-27
STORM-2127: Storm-eventhubs should use latest amqp and eventhubs-client versions

raviperi:
Update eventhub client dependency. Move Storm-Eventhubs dependency vesion definitions to parent pom. Introduce new schemes for message handling


raviperi:
Below JIRA that reflects the need for the above work.
STORM-2127 - Storm-eventhubs should use latest amqp and eventhubs-client versions

harshach:
+1

harshach:
@raviperi  can you add the STORM jira number to the commit message and squash commits in both the PRs. Use the same title as the JIRA title https://issues.apache.org/jira/browse/STORM-2127

raviperi:
@harshach  
Updated title as per your recommendation.

PR #1720, opened 2016-09-29
STORM-1546: Adding Read and Write Aggregations for Pacemaker to make it HA compatible (1.x-branch)

knusbaum:
1.x-branch version of Pacemaker-HA


redsanket:
+1

HeartSaVioR:
Please add a documentation here too.

HeartSaVioR:
Seems like this PR has different diff compared to master (#1710) except Clojure things. This could be OK in short time, but will make us bothering if we need to fix something with Pacemaker.

I'd recommend to sync this with #1710 except pacemaker_state_factory.clj. I'm also OK to replace related clojure files to ported files in 1.x since it doesn't matter.

knusbaum:
@HeartSaVioR I've synced the java files between the branches so there are almost no differences.

knusbaum:
Also updated the docs.

HeartSaVioR:
Thanks for the change. +1

PR #1723, opened 2016-10-03
STORM-2131: Add blob command to worker-launcher, make stormdist directory not writeable by topo owner

knusbaum:


kishorvpatil:
+1

PR #1724, opened 2016-10-05
STORM-2131: Add blob command to worker-launcher, make stormdist directory not writeable by topo owner 

knusbaum:
Re-opening this, as I reverted the last one to fix a typo (#1723)

### DIFF
--- a/storm-core/src/jvm/org/apache/storm/daemon/supervisor/SupervisorUtils.java
+++ b/storm-core/src/jvm/org/apache/storm/daemon/supervisor/SupervisorUtils.java
@@ -104,6 +104,16 @@ public class SupervisorUtils {
         }
     }
 
+    public static void setupWorkerArtifactsDir(Map<String, Object> conf, Map<String, Object> stormConf, String dir) throws IOException {
+        if (Utils.getBoolean(conf.get(Config.SUPERVISOR_RUN_WORKER_AS_USER), false)) {
+            String logPrefix = "setup conf for " + dir;
+            List<String> commands = new ArrayList<>();
+            commands.add("artifacts-dir");
+            commands.add(dir);
+            processLauncherAndWait(conf, (String) (stormConf.get(Config.TOPOLOGY_SUBMITTER_USER)), commands, null, logPrefix);
+        }
+    }
+
     public static void rmrAsUser(Map<String, Object> conf, String id, String path) throws IOException {
         String user = Utils.getFileOwner(path);
         String logPreFix = "rmr " + id;
--- a/storm-core/src/jvm/org/apache/storm/daemon/supervisor/SupervisorUtils.java
+++ b/storm-core/src/jvm/org/apache/storm/daemon/supervisor/SupervisorUtils.java
@@ -96,7 +96,7 @@ public class SupervisorUtils {
 
     public static void setupStormCodeDir(Map<String, Object> conf, Map<String, Object> stormConf, String dir) throws IOException {
         if (Utils.getBoolean(conf.get(Config.SUPERVISOR_RUN_WORKER_AS_USER), false)) {
-            String logPrefix = "setup conf for " + dir;
+            String logPrefix = "Storm Code Dir Setup for " + dir;
             List<String> commands = new ArrayList<>();
             commands.add("code-dir");
             commands.add(dir);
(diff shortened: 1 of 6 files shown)


HeartSaVioR:
Leaving a note: #1697 needs to pull this to apply the change for 1.x.

knusbaum:
@kishorvpatil @revans2 @jerrypeng

kishorvpatil:
+1

storm-core/src/native/worker-launcher/impl/worker-launcher.c
```diff
@@ -485,14 +482,14 @@ int setup_stormdist_dir(const char* local_dir) {
 
       case FTS_DP:        // A directory being visited in post-order
       case FTS_DOT:       // A dot directory
+      case FTS_SL:        // A symbolic link
+      case FTS_SLNONE:    // A broken symbolic link
```
redsanket: just curious, why did we change the behaviour here

storm-core/src/native/worker-launcher/impl/worker-launcher.c
```diff
@@ -414,33 +414,30 @@ static int copy_file(int input, const char* in_filename,
...
+            strerror(errno), entry->fts_path);
      return -1;
   }
   mode_t mode = entry->fts_statp->st_mode;
-  mode_t new_mode = (mode & (S_IRWXU)) | S_IRGRP | S_IWGRP;
-  if ((mode & S_IXUSR) == S_IXUSR) {
-    new_mode = new_mode | S_IXGRP;
+  mode_t new_mode = (mode & (S_IRUSR | S_IXUSR)) | S_IRGRP | S_IWGRP;
+  if (user_write) {
+    new_mode = new_mode | S_IWUSR;
```
revans2: It was not completely obvious from reading the code what was happening here.  Could we put in some comments at a high level how the permissions are setup and why they are setup that way.

storm-core/src/native/worker-launcher/impl/worker-launcher.c
```diff
@@ -414,33 +414,30 @@ static int copy_file(int input, const char* in_filename,
...
+            strerror(errno), entry->fts_path);
     return -1;
   }
   return 0;
 }
 
-int setup_stormdist_dir(const char* local_dir) {
-  //This is the same as
-  //> chmod g+rwX -R $local_dir
-  //> chown -no-dereference -R $user:$supervisor-group $local_dir
```
revans2: Again a comment like this explaining what the program does would be very valuable.

storm-core/src/native/worker-launcher/impl/worker-launcher.c
revans2: I'm also not sure why we want the symlink/a broken symlink to be a noop?  We use lchown to not follow the symlink, so we will be sure to change ownership of the symlink itself (otherwise it will still be owned by the supervisor user).  Was chmod causing issues?  If so we probably want to split the chown and chmod portions and only call chmod on things that are not symlinks.

knusbaum: This is ported from a change @tgravescs made. We can ask him why.

storm-core/src/jvm/org/apache/storm/daemon/supervisor/SupervisorUtils.java
```diff
@@ -104,6 +104,16 @@ public static void setupStormCodeDir(Map<String, Object> conf, Map<String, Objec
         }
     }
 
+    public static void setupWorkerArtifactsDir(Map<String, Object> conf, Map<String, Object> stormConf, String dir) throws IOException {
+        if (Utils.getBoolean(conf.get(Config.SUPERVISOR_RUN_WORKER_AS_USER), false)) {
+            String logPrefix = "setup conf for " + dir;
```
knusbaum: No idea. I copied this over from setupStormCodeDir so they would look similar in the logs.
Can change both to actually make sense.

storm-core/src/native/worker-launcher/impl/main.c
```diff
@@ -155,7 +155,23 @@ int main(int argc, char **argv) {
       fflush(ERRORFILE);
       return INVALID_ARGUMENT_NUMBER;
     }
-    exit_code = setup_stormdist_dir(argv[optind]);
+    exit_code = setup_dir_permissions(argv[optind], 0);
+  } else if (strcasecmp("artifacts-dir", command) == 0) {
+      if (argc != 4) {
```
knusbaum: Fixed.

storm-core/src/native/worker-launcher/impl/worker-launcher.c
knusbaum: 👍

storm-core/src/native/worker-launcher/impl/worker-launcher.c
knusbaum: 👍

storm-core/src/native/worker-launcher/impl/worker-launcher.c
revans2: I talked with @tgravescs and ran some tests.  It works just fine, and it should not be a big deal.  I was a bit worried about the default permissions on the symlinks but they are open, and because they are all under a directory that should be locked down I am not concerned abut it.

revans2:
+1 looks good

PR #1726, opened 2016-10-07
STORM-2139: Let ShellBolts and ShellSpouts run with scripts from blobs

revans2:


HeartSaVioR:
+1 for the concept.
Btw, do you have use case for this, or does this come from idea? And how users can use this?
I just would like to see easiest way to use this concept, and also applied doc so that users can play with.

redsanket:
+1

revans2:
@HeartSaVioR we have someone who wants to ship their own version of python with a topology on a multi-tenant cluster.  They tried to package it as part of the jar, but were getting permission denied errors because jar loses all permissions.  This will unblock them, without requiring us to install a version of python for them to use.

You are right I should write up an example + some documentation on how to use it.  I'll see what I can do.

HeartSaVioR:
@revans2 
Interesting use case... I didn't even think about that. Thanks for clarifying.

revans2:
@HeartSaVioR @redsanket I added in some documentation for this.  If you have time to take another look that would be great.

HeartSaVioR:
+1 Thanks for adding the documentation @revans2 .

PR #1725, opened 2016-10-07
STORM-2135 document covered features of SQL reference in Calcite

HeartSaVioR:
NOTE: This should be merged after STORM-2125 (#1714) is merged to master

* add a new page "storm-sql-reference" based on Calcite SQL reference

PR #1681, opened 2016-09-12
STORM-1444 Support EXPLAIN statement in StormSQL

HeartSaVioR:
* when '--explain' is specified to topology name, activate explain mode
* when explain mode is activated, it analyzes each query statement, and shows query plan
  * it doesn't submit topologies
* Showing plan is provided by Calcite: it just uses that feature

As I also described to the doc, query plan for below query 
```
INSERT INTO LARGE_ORDERS SELECT ID, UNIT_PRICE * QUANTITY AS TOTAL FROM ORDERS 
WHERE UNIT_PRICE * QUANTITY > 50
```

is shown as: 

```
LogicalTableModify(table=[[LARGE_ORDERS]], operation=[INSERT], updateColumnList=[[]], flattened=[true]), id = 8
  LogicalProject(ID=[$0], TOTAL=[*($1, $2)]), id = 7
    LogicalFilter(condition=[>(*($1, $2), 50)]), id = 6
      EnumerableTableScan(table=[[ORDERS]]), id = 5
```

FYI, $0 is column index 0 (starting from 0), and so on. Expression is printed as prefix notation.
Calcite just provides this functionality, though I'm not clear whether this is friendly (acceptable by end-user) or not.
There's staled issue from Calcite ([CALCITE-155](https://issues.apache.org/jira/browse/CALCITE-155)) for displaying actual column names/aliases in explain plan.

Please have a look and share your opinion about this.
Thanks in advance!

PR #1727, opened 2016-10-09
STORM-2136 Write some examples to show the usage of Storm SQL

HeartSaVioR:
This is on top of STORM-2135 (#1725).

* This should be merged after STORM-2135 - #1725 (hence after STORM-2125 - #1714 , too)
* This should also be merged after STORM-1444 - #1681


HeartSaVioR:
I'll just merge it since preconditions are met and this change is limited to documentation.

PR #1728, opened 2016-10-10
STORM-2142 ReportErrorAndDie runs suicide function only when InterruptedException or InterruptedIOException is thrown

HeartSaVioR:
* There seems to be a small bug while porting


knusbaum:
+1
Looks like mistranslation. This restores the behavior of the old Clojure code.

knusbaum:
(Nice catch, BTW)

HeartSaVioR:
@knusbaum Thanks for the review. :)

PR #1729, opened 2016-10-10
[STORM-2144] Fix Storm-sql group-by behavior in standalone mode

arunmahadevan:
Fix group-by to not rely on monotonic group-by keys

### DIFF
--- a/external/sql/storm-sql-core/src/jvm/org/apache/storm/sql/compiler/backends/standalone/PlanCompiler.java
+++ b/external/sql/storm-sql-core/src/jvm/org/apache/storm/sql/compiler/backends/standalone/PlanCompiler.java
@@ -40,6 +40,7 @@ public class PlanCompiler {
       "// GENERATED CODE", "package " + PACKAGE_NAME + ";", "",
       "import java.util.Iterator;", "import java.util.Map;", "import java.util.HashMap;",
       "import java.util.List;", "import java.util.ArrayList;",
+      "import java.util.LinkedHashMap;",
       "import org.apache.storm.tuple.Values;",
       "import org.apache.storm.sql.runtime.AbstractChannelHandler;",
       "import org.apache.storm.sql.runtime.Channels;",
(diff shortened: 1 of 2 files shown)


satishd:
+1 LGTM

HeartSaVioR:
Found storm-jms version issue while checking Travis CI failing. Will submit a patch shortly.

HeartSaVioR:
Other than that I'm +1. IMHO, standalone mode also needs to have some abstraction or introduce something to not creating source code by manipulate string directly.

arunmahadevan:
@HeartSaVioR you are right, its very tedious to maintain. Right now its kind of a bridge to translate sql to java so that sql abstractions can be executed in storm core. IMO we should get out of standalone mode in future and add support in storm-sql to directly translate to storm-core topologies via the streams api proposed in STORM-1961 (https://github.com/apache/storm/pull/1693).

satishd:
+1 for replacing code generation with an abstract API as it is hard to maintain. But that can be a different task and it is not related to this PR.

harshach:
+1. Isn't standalone mode still helpful for devs instead of just moving to abstract API.

HeartSaVioR:
I guess Calcite itself can provide interpreter for handling query. If then standalone users can just move to Calcite interpreter, and standalone mode could be removed. I'm planning to find it from Calcite.

HeartSaVioR:
I guess `Interpreter` seems to be candidate for replacement of standalone more on storm-sql.

https://github.com/apache/calcite/blob/master/core/src/main/java/org/apache/calcite/interpreter/Interpreter.java

Unit tests are here.

https://github.com/apache/calcite/blob/master/core/src/test/java/org/apache/calcite/test/InterpreterTest.java

It still need to register tables and functions to root schema but storm-sql is same.

HeartSaVioR:
Spend some time to look into it but it looks like breaking changes are needed.
Interpreter expects static input table before initializing Interpreter, which is valid for most of case, but Storm SQL standalone mode works like streaming, emitting input tuples and receive the outputs.

satishd:
Thanks @arunmahadevan for the fix.

satishd:
Pushed to master branch also.

PR #1730, opened 2016-10-10
STORM-2145 Leave leader nimbus's hostname to log when trying to connect leader nimbus

HeartSaVioR:
* leave a log regarding leader Nimbus connection information when connecting to leader Nimbus
* also leave a log when Nimbus figures out its hostname without using storm.local.hostname

This should be also ported back to 1.x and 1.0.x.


harshach:
+1

PR #1734, opened 2016-10-10
STORM-2141 [Storm SQL] UDAFWrapper needs to provide parameter information when error invoking method

HeartSaVioR:
While developing UDF and UDAF, users should be aware of association between Calcite RelType and Java Type. 
Though Calcite validates SQL statement, there seems be the chance for UDAF to be called for unexpected parameter types, and "java.lang.IllegalArgumentException: argument type mismatch" is thrown.
We can grab it and provide expected parameters and actual parameters so that users can investigate why.


HeartSaVioR:
This is against STORM-2200 which is discussed on dev@. Closing.

PR #1735, opened 2016-10-13
STORM-203 Adding paths to default java library path

picadoh:
Adding /usr/lib64 and /usr/lib/x86_64-linux-gnu to default java library path


knusbaum:
/usr/lib64 is FHS standard. /usr/lib/x86_64-linux-gnu is Debian-specific. I think we should avoid setting distro-specific defaults.

picadoh:
yeah, makes total sense 👍 removed it from the path. thanks @knusbaum

revans2:
@knusbaum are you OK with the change?

I am +1

knusbaum:
+1

PR #1736, opened 2016-10-14
STORM-1446 [Storm SQL] Compile the Calcite logical plan to Storm Trident logical plan

HeartSaVioR:
First of all I should thank to @milinda for sharing great implementation, because this patch is merely just a port of SamzaSQL (though it was from @milinda repo, not Samza repo).

This is for converting Calcite logical plan to Storm's Trident logical plan (as my understanding of logical plan / physical plan is right). Trident itself does own physical planning.

After introducing this, much more things can be followed up, like
- setting up parallelism hint based on data source's hint (metadata)
- push down filter / project to data source
- go back generating source code of Trident topology, but avoiding raw String concatenation
- and whatever...

This is rather huge patch and it requires understanding of core concept of Calcite. If you haven't learn it, this slide can help you, as I did.
http://www.slideshare.net/JordanHalterman/introduction-to-apache-calcite

It doesn't touch standalone mode (but I will try later), and also doesn't touch feature itself so it doesn't require documentation, but adding javadoc comment might be needed. 
If reviewers want to see it please let me know so that I can work on.

Thanks in advance!

Commit log is below:
- Port SamzaSQL implementation to Storm
  - https://github.com/milinda/samza-sql
- Apply some rules to optimize
- optimize Calc
  - merge filter and projection scripts into one
  - also applying short circuit
- Modify Trident unit tests to use new query planner
- arrange some files
  - Move some files which are only used from standalone
  - Remove some files which are no longer used
- guard the possibility of stack overflow error on explaining
  - just leave error logs, and print out empty plan and continue
  - reported this behavior to Calcite community
- leave some comments to clarify what it means


HeartSaVioR:
Just fixed the RAT issue.

manuzhang:
@HeartSaVioR Could you please clarify what's the motive behind this big change ? What's the benefits ? Has this changed the [workflow of StormSQL](https://storm.apache.org/releases/2.0.0-SNAPSHOT/storm-sql-internal.html) ?

HeartSaVioR:
@manuzhang 
Thanks for showing interest on this. Let me rearrange your questions:

> Has this changed the workflow of StormSQL?

No, we still rely on Trident, and nothing changed in point of workflow.

> Motive behind this big change and benefits

This is started from [Julian's comment](https://issues.apache.org/jira/browse/STORM-1040?focusedCommentId=15034472&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-15034472) and also [Milinda's comment](https://issues.apache.org/jira/browse/STORM-1040?focusedCommentId=15035182&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-15035182).

For me having own relational algebras (rel) has several advantages, 

- We can push operator handling logic to rel itself. Before that we should traverse Calcite logical rel  tree with PostOrderRelNodeVisitor, and visitor needs to handle Calcite's rel directly. Now the logic how to configure Trident topology is all handled from separate rels.

- We sometimes want to have more derived rels compared to Calcite logical operators. One of example is `Join`.  There's only one logical rel regarding join in Calcite - LogicalJoin - but we're now converting LogicalJoin to EquiJoin if conditions are met. If we have various types of join it will make the difference. We're not prepared yet, but streaming scan vs table scan, and streaming insert vs table insert are the other cases.

```
TridentStormAggregateRel(group=[{0}], EXPR$1=[COUNT()])
  TridentStormCalcRel(expr#0..4=[{inputs}], expr#5=[0], expr#6=[>($t0, $t5)], DEPTID=[$t3], EMPID=[$t0], $condition=[$t6])
    TridentStormEquiJoinRel(condition=[=($2, $3)], joinType=[inner])
      TridentStormStreamScanRel(table=[[EMP]])
      TridentStormStreamScanRel(table=[[DEPT]])
```

We can even override the methods how to represent the rel in explain string if we think Calcite's explain is less informational. For example, showing initial parallelism (when we support) for Scan.

- This patch starts addressing query optimizations. One of example is Calc, which is for merging multiple calculations into one. No need to filter and projection separately. They're still not minimized (as projection is), but it can be addressed after STORM-2072. Defining derived rels helps further query optimizations, like filter pushdown. Calcite rels is not aware of data source characteristic, and we can include it to our own rels.

Please don't hesitate to ask questions. Thanks!

external/sql/storm-sql-core/src/jvm/org/apache/storm/sql/planner/StormRel.java
```diff
@@ -0,0 +1,32 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.storm.sql.planner;
+
+import org.apache.calcite.plan.Convention;
+import org.apache.storm.sql.planner.rel.StormRelNode;
+import org.apache.storm.sql.planner.trident.TridentPlanCreator;
+
+public interface StormRel extends StormRelNode {
```
manuzhang: I know the naming `StormRel` is inherent from Samza-SQL but it's quite confusing with `StormRelNode`.  What `StormRel` actually does is like a translator.

manuzhang:
@HeartSaVioR Thanks for the thorough explanation. I'd suggest record the motive and rationals somewhere.  It's hard to follow why something exists in the first place.

HeartSaVioR:
@manuzhang Yes issue description just links Julian's comment and I think it's not sufficient. I'll also write up motivation and rationale to issue description. Thanks.

HeartSaVioR:
Updated issue's desc.

HeartSaVioR:
Fixed some bugs, and renamed classes to reflect more specific.
Unit test passes and manually tested.

HeartSaVioR:
Reflect STORM-2072 : STORM-1446 now also addresses STORM-2073.

HeartSaVioR:
Also addressed STORM-2200: Drop Aggregate & Join support on Trident mode

external/sql/storm-sql-core/src/jvm/org/apache/storm/sql/StormSqlImpl.java
```diff
@@ -48,7 +45,10 @@
 import org.apache.storm.sql.parser.SqlCreateFunction;
 import org.apache.storm.sql.parser.SqlCreateTable;
 import org.apache.storm.sql.parser.StormParser;
+import org.apache.storm.sql.planner.trident.QueryPlanner;
+import org.apache.storm.sql.planner.StormRelUtils;
 import org.apache.storm.sql.runtime.*;
```
harshach: can we remove the wildcard import

external/sql/storm-sql-core/src/jvm/org/apache/storm/sql/StormSqlImpl.java
```diff
@@ -113,18 +113,13 @@ public void submit(
...
-        Planner planner = Frameworks.getPlanner(config);
-        SqlNode parse = planner.parse(sql);
-        SqlNode validate = planner.validate(parse);
-        RelNode tree = planner.convert(validate);
-        org.apache.storm.sql.compiler.backends.trident.PlanCompiler compiler =
-                new org.apache.storm.sql.compiler.backends.trident.PlanCompiler(dataSources, typeFactory, dataContext);
-        TridentTopology topo = compiler.compile(tree);
+        QueryPlanner planner = new QueryPlanner(schema);
+        AbstractTridentProcessor processor = planner.compile(dataSources, sql);
+        TridentTopology topo = processor.build(null);
```
harshach: any reason to pass null here?. Also a comment would be good on passing null

external/sql/storm-sql-core/src/jvm/org/apache/storm/sql/StormSqlImpl.java
```diff
@@ -166,8 +161,7 @@ public void explain(Iterable<String> statements) throws Exception {
         SqlNode validate = planner.validate(parse);
         RelNode tree = planner.convert(validate);
 
-        // TODO: change to all attributes when we change to cost-based planner
-        String plan = RelOptUtil.toString(tree, SqlExplainLevel.NON_COST_ATTRIBUTES);
+        String plan = StormRelUtils.explain(tree, SqlExplainLevel.ALL_ATTRIBUTES);
         System.out.println("plan>");
```
harshach: should we move all these println statements to Utils print statement and in future we can change this to file based approach?

HeartSaVioR: That's a good idea, but not related to this change. I'll file an issue regarding this.

external/sql/storm-sql-core/src/jvm/org/apache/storm/sql/StormSqlImpl.java
HeartSaVioR: Just removed that parameter since it is no longer needed.

external/sql/storm-sql-core/src/jvm/org/apache/storm/sql/StormSqlImpl.java
HeartSaVioR: Addressed.

HeartSaVioR:
@harshach Thanks for the review. I addressed your comments.

HeartSaVioR:
Addressed the change with master: update to Calcite 1.11.0

harshach:
+1

PR #1739, opened 2016-10-19
STORM-1443 [Storm SQL] Support customizing parallelism in StormSQL

HeartSaVioR:
NOTE: This is on top of STORM-1446 (#1736). I can make a patch without STORM-1446 but it makes me putting same effort twice so didn't do it. Please let me know if we want to have this patch based on current master.
- Add 'PARALLELISM' to table definition
  - default value is 1
- Set parallelism to new stream while creating stream with scan
  - downstream operators will also have same parallelism unless repartitioned
  - not apply parallelism to output table since it can trigger repartition

Below is the screenshot which runs SQL statement:

storm-1443-screenshot

```
CREATE EXTERNAL TABLE APACHE_LOGS (id INT PRIMARY KEY, remote_ip VARCHAR, request_url VARCHAR, request_method VARCHAR, status VARCHAR, request_header_user_agent VARCHAR, time_received_utc_isoformat VARCHAR, time_us DOUBLE) LOCATION 'kafka://localhost:2181/brokers?topic=apachelogs-v2' PARALLELISM 5
CREATE EXTERNAL TABLE APACHE_SLOW_LOGS (dummy_id INT PRIMARY KEY, request_url VARCHAR, request_method VARCHAR, cnt INT, time_elapsed_ms_min INT, time_elapsed_ms_max INT, time_elapsed_ms_avg INT) LOCATION 'kafka://localhost:2181/brokers?topic=apacheslowlogs-v2' TBLPROPERTIES '{"producer":{"bootstrap.servers":"localhost:9092","acks":"1","key.serializer":"org.apache.storm.kafka.IntSerializer","value.serializer":"org.apache.storm.kafka.ByteBufferSerializer"}}'
INSERT INTO APACHE_SLOW_LOGS SELECT MIN(ID), REQUEST_URL, REQUEST_METHOD, COUNT(*) AS CNT, MIN(TIME_US) / 1000 AS TIME_ELAPSED_MS_MIN, MAX(TIME_US) / 1000 AS TIME_ELAPSED_MS_MAX, AVG(TIME_US) / 1000 AS TIME_ELAPSED_MS_AVG FROM APACHE_LOGS GROUP BY REQUEST_URL, REQUEST_METHOD HAVING AVG(TIME_US) / 1000 >= 300
```

Please refer task count of each component. Task count of each component is 5 unless it's repartitioned due to aggregation.

You can narrow the change to only last commit. It's 120 lines of change.

### DIFF
--- a/docs/storm-sql.md
+++ b/docs/storm-sql.md
@@ -46,11 +46,19 @@ CREATE EXTERNAL TABLE table_name field_list
       OUTPUTFORMAT output_format_classname
     ]
     LOCATION location
+    [ PARALLELISM parallelism ]
     [ TBLPROPERTIES tbl_properties ]
     [ AS select_stmt ]
 ```
 
-You can find detailed explanations of the properties in [Hive Data Definition Language](https://cwiki.apache.org/confluence/display/Hive/LanguageManual+DDL). For example, the following statement specifies a Kafka spout and sink:
+You can find detailed explanations of the properties in [Hive Data Definition Language](https://cwiki.apache.org/confluence/display/Hive/LanguageManual+DDL). 
+
+`PARALLELISM` is StormSQL's own keyword which describes parallelism hint for input data source. This is same as providing parallelism hint to Trident Spout.
+As same as Trident, downstream operators are executed with same parallelism before repartition (Aggregation triggers repartition).
+
+Default value is 1, and this option is no effect on output data source. (We might change if needed. Normally repartition is the thing to avoid.)
+
+For example, the following statement specifies a Kafka spout and sink:
 
 ```
 CREATE EXTERNAL TABLE FOO (ID INT PRIMARY KEY) LOCATION 'kafka://localhost:2181/brokers?topic=test' TBLPROPERTIES '{"producer":{"bootstrap.servers":"localhost:9092","acks":"1","key.serializer":"org.apache.org.apache.storm.kafka.IntSerializer","value.serializer":"org.apache.org.apache.storm.kafka.ByteBufferSerializer"}}'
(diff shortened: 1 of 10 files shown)


vesense:
@HeartSaVioR Overall looks good to me.
And I have a question: Now I'm working on STORM-2147 which I think should be based on STORM-1443. Through this PR we can set the parallelism by specifying `PARALLELISM` in SQL,  I want to know how can I do this in `DataSourcesProvider`(i.e. how to set partition number in data sources, new APIs or anything else)?

HeartSaVioR:
@vesense 
Actually I had some time to think about STORM-2147.
There might be some ways to pass partition count to upstream, and easy way to do might be adding method to DataSourcesProvider. What I'm considering is that we're now only thinking about partition count, but Calcite supports table statistics which contains estimated row count (not available for streaming env., partitioning attributes, etc.). Is passing partition count exhaustive? I'm not sure.

As this is on top of STORM-1446 which needs understanding of Calcite, learning Calcite is more important for you to work on further works for Storm SQL, especially Storm SQL lacks reviewers to go forward. So if you haven't had time, let's take your time to get it.

vesense:
@HeartSaVioR 
>There might be some ways to pass partition count to upstream, and easy way to do might be adding method to DataSourcesProvider. What I'm considering is that we're now only thinking about partition count, but Calcite supports table statistics which contains estimated row count (not available for streaming env., partitioning attributes, etc.). Is passing partition count exhaustive? I'm not sure.

Perhaps I can give you some suggestions when I have a deeper understanding in Calcite. But now I think we can address STORM-2147 using a simple way before we get a better one.

>As this is on top of STORM-1446 which needs understanding of Calcite, learning Calcite is more important for you to work on further works for Storm SQL, especially Storm SQL lacks reviewers to go forward. So if you haven't had time, please take your time to get it.

Thanks for your advice, yes, learning Calcite is in my plan.

HeartSaVioR:
Rebased onto master since STORM-1446 is merged into master.

ptgoetz:
+1 Being able to control parallelism will make Storm SQL a lot more usable.

harshach:
+1. LGTM

PR #1743, opened 2016-10-24
STORM-2148 [Storm SQL] Trident mode: back to code generate and compile Trident topology

HeartSaVioR:
This is on top of STORM-1446 (#1736). 

Instead of fully generating topology code, I just generate classes for each expression, and let EvaluateXXX calls instance's method. This is much simpler and nearly no overhead added, and also much faster than evaluating code block.

I did a [simple benchmark](https://issues.apache.org/jira/browse/STORM-2148?focusedCommentId=15601152&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-15601152) between evaluating code block and calling instance's method. Please refer the link if you are interested.

Also manually tested.

You can narrow the change to only last commit. It's 700 lines of change.


HeartSaVioR:
Rebased onto master since STORM-1446 is merged into master.

ptgoetz:
+1

PR #1745, opened 2016-10-25
STORM-2174: Initial Base for Storm Beam Runner

ptgoetz:
This is an (very) initial pass at laying the foundation for a Beam runner for Storm. It currently only supports local mode, and is incomplete (and likely wrong in places) in terms of supporting the basic word count pipeline example.

The goal of getting this committed into a feature branch is to enable others to collaborate and contribute to the effort.

### DIFF
--- a/external/storm-beam/pom.xml
+++ b/external/storm-beam/pom.xml
@@ -31,7 +31,7 @@
 
     <name>Apache Storm Beam Runner</name>
     <properties>
-        <beam.version>0.2.0-incubating-SNAPSHOT</beam.version>
+        <beam.version>0.2.0-incubating</beam.version>
     </properties>
 
     <dependencies>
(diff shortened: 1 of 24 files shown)


revans2:
Looks good overall.  Didn't spend a lot of time digging in to all the details but it looks like a really nice start.

arunmahadevan:
@ptgoetz I had tried to run the word count with the storm runner when the patch was posted, but there were some issues with the way grouping is implemented. I didn't get time to dig in further. Overall its a good start and may be we should merge it in as is and refactor it to use the proposed streams api once that gets merged which would simplify a lot of things.

ptgoetz:
@revans2 @arunmahadevan Thanks for the review. I will merge this to the beam runner feature branch so others can submit pull requests against it.

@arunmahadevan Yes, the GBK implementation is wrong. It's a WIP to get data moving between components.

PR #1748, opened 2016-10-28
[STORM-2103][SQL] Introduce new sql external module: storm-sql-mongodb

vesense:
major changes as following:
- new KafkaDataSourcesProvider & unit tests
- change DataSourcesProvider method `constructTrident()`'s arg `properties` data type from `String` to `Properties`
- abstract common logic to Utils
- some minor improvements

### DIFF
--- a/external/sql/pom.xml
+++ b/external/sql/pom.xml
@@ -41,5 +41,6 @@
         <module>storm-sql-runtime</module>
         <module>storm-sql-external/storm-sql-kafka</module>
         <module>storm-sql-external/storm-sql-redis</module>
+        <module>storm-sql-external/storm-sql-mongodb</module>
     </modules>
 </project>
--- a/external/sql/storm-sql-core/src/jvm/org/apache/storm/sql/parser/SqlCreateTable.java
+++ b/external/sql/storm-sql-core/src/jvm/org/apache/storm/sql/parser/SqlCreateTable.java
@@ -17,6 +17,7 @@
  */
 package org.apache.storm.sql.parser;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
 import org.apache.calcite.sql.SqlCall;
 import org.apache.calcite.sql.SqlIdentifier;
 import org.apache.calcite.sql.SqlKind;
@@ -29,8 +30,11 @@ import org.apache.calcite.sql.SqlWriter;
 import org.apache.calcite.sql.parser.SqlParserPos;
 import org.apache.calcite.util.ImmutableNullableList;
 
+import java.io.IOException;
 import java.net.URI;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Properties;
 
 public class SqlCreateTable extends SqlCall {
   public static final SqlSpecialOperator OPERATOR = new SqlSpecialOperator(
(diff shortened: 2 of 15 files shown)


vesense:
since this PR has some common logics, hope it will be merged ASAP.

HeartSaVioR:
How it's related to STORM-1607? Is this on top of STORM-1607? Or they are independent?

vesense:
@HeartSaVioR they are independent.

vesense:
I want to make it clear that PRs getting merged order is https://github.com/apache/storm/pull/1748 https://github.com/apache/storm/pull/1751 https://github.com/apache/storm/pull/1754
Hope these PRs can be merged ASAP since many following up PRs are on the road.

HeartSaVioR:
@vesense Thanks for addressing numerous works. I hadn't have time recently and still have less spare time, but I'll try to find some time to review. Thanks in advance for your patience.

HeartSaVioR:
@vesense I just assigned some time to myself. I don't have MongoDB installed but code change looks good. One question though: it seems to assign serialization field: do we use only one field for storing all the things, or does Mongo (or BSON) take care of it automatically? 

I'd like to see 'how to use' documented for storm-sql-mongo but others are also not documented, so we would want to create a document for describing how to configure / use external data source in Storm SQL.

Btw, do you run the SQL statement with storm-sql-mongo? I don't have mongodb installed so it might take another hours or even days to do manual test. If you don't really mind, I'd recommend you to paste your SQL statements (including datasource definitions) which runs well with the patch.

HeartSaVioR:
@vesense 
I manually tested that it works. Thanks for the great work!
Btw, I'm seeing that only one fields are used for storing the value, but I guess some users occasionally would want to store the document as it is (not serialized to one field). I'm OK if there's specific reason to do it (for example, supporting schema matters) but if it's not, I feel "storing as it is" is more natural and what users expected without any configurations.

vesense:
@HeartSaVioR Thanks for your quick reply.
The reason tuple data is serialized to one field is that the method
  `write(List<Object> data, ByteBuffer buffer)` in interface `IOutputSerializer` serialize a tuple's fields and values to a `ByteBuffer` object. For this matters, I use one Mongo document field to store it. The sqe project also do like this.
About 'how to use' document, do you mind if I do this job in a future Jira? Now I'm working on some other tasks, such as storm-sql-hdfs and etc.

vesense:
@HeartSaVioR Thanks for your quick reply.
The reason tuple data is serialized to one field is that the method
  `write(List<Object> data, ByteBuffer buffer)` in interface `IOutputSerializer` serialize a tuple's fields and values to a `ByteBuffer` object. For this matters, I use one Mongo document field to store it. The sqe project also do like this.
About 'how to use' document, do you mind if I do this job in a future Jira? Now I'm working on some other tasks, such as storm-sql-hdfs and etc.

HeartSaVioR:
@vesense 
Regarding field, yeah I forgot it. We should have a chance to revamp the Storm SQL related APIs, but we could also file an issue and postpone it.
Regarding document, yes others are not documented as well, so should address it, but with others as well. We can file an issue to handle.
Other things look good to me. +1

HeartSaVioR:
Filed to https://issues.apache.org/jira/browse/STORM-2186 and https://issues.apache.org/jira/browse/STORM-2187.

vesense:
Thanks @HeartSaVioR for your response and field Jiras. After my current tasks completed, I will pick them up.
I will update #1751 accordingly, and then you can take a look if you have time.

PR #1749, opened 2016-10-28
STORM-2175: fix double close of workers

revans2:
Will put up a pull request for 1.x shortly.

PR #1750, opened 2016-10-28
STORM-2175: fix double close of workers

revans2:


revans2:
Will reopen once I have a better fix that gives more time and throws on a timeout.

srdo:
Tested this with our test set, the issues we were seeing seem fixed. Thank you for looking into this. +1

revans2:
@HeartSaVioR could you take a look at this and #1749 so I can merge this in?

HeartSaVioR:
@revans2 +1 and I'll take a look at #1749 very soon. Thanks for the great work.

PR #1755, opened 2016-11-01
error log for blobstore was missing a space between string 'key' and …

erikdw:
…the actual key value

Example:

2016-11-01 06:09:58.001 o.a.s.b.BlobStoreUtils [ERROR] Could not download blob with keysmoketest-l-1-1460514403-stormconf.ser

Notice how 'key' and 'smoketest...' run together?

Solution: make the error log similar to the one in the following function for when a blob couldn't be updated.


revans2:
+1

revans2:
@erikdw 

Thanks, I merged this into master.

PR #1756, opened 2016-11-01
STORM-1278: Port org.apache.storm.daemon.worker to java

abhishekagarwal87:
@revans2 - I may need some help in repairing worker unit tests.


revans2:
@abhishekagarwal87 happy to take a look.  I'll see what I can do on the tests.

storm-core/src/jvm/org/apache/storm/daemon/worker/LogConfigManager.java
```diff
@@ -0,0 +1,154 @@
...
+        this.latestLogConfig = latestLogConfig;
+        this.originalLogLevels = getLoggerLevels();
+        LOG.info("Started with log levels: {}", originalLogLevels);
+    }
+
+    public void processLogConfigChange(LogConfig logConfig) {
+        if (null != logConfig) {
+            LOG.debug("Processing received log config: {}", logConfig);
+            TreeMap<String, LogLevel> loggers = new TreeMap<>(logConfig.get_named_logger_level());
+            LoggerContext logContext = (LoggerContext) LogManager.getContext();
```
revans2: In the previous code it passed in a false.  Not sure if it makes much difference in our case though.

https://logging.apache.org/log4j/2.x/log4j-api/apidocs/org/apache/logging/log4j/LogManager.html#getContext(boolean)

revans2:
@abhishekagarwal87 I created a pull request to your repo for the failing tests 

https://github.com/abhishekagarwal87/storm/pull/7

I will keep looking at the pull request

storm-core/src/clj/org/apache/storm/daemon/local_executor.clj
```diff
@@ -25,18 +25,3 @@
     (let [val (AddressedTuple. task tuple)]
```
revans2: This entire file should really be removed and with-tracked-cluster should be updated to not use it.  It is obviously not needed because this function is never called, only over ridden.

storm-core/src/jvm/org/apache/storm/daemon/Task.java
```diff
@@ -179,22 +178,21 @@ public BuiltinMetrics getBuiltInMetrics() {
...
+                        ConfigUtils.supervisorStormDistRoot(conf, workerData.getTopologyId())),
+                ConfigUtils.workerPidsRoot(conf, workerData.getWorkerId()),
                 taskId,
-                (Integer) workerData.get(Constants.PORT),
-                (List<Integer>) workerData.get(Constants.TASK_IDS),
-                (Map<String, Object>) workerData.get(Constants.DEFAULT_SHARED_RESOURCES),
-                (Map<String, Object>) workerData.get(Constants.USER_SHARED_RESOURCES),
+                workerData.getPort(), workerData.getTaskIds(),
+                workerData.getDefaultSharedResources(),
+               workerData.getUserSharedResources(),
```
revans2: nit: indentation is off

revans2:
Perf numbers look good compared to the 1.x line (very non scientific though).  I didn't dig into it a lot.  Running throughput vs latency on my mac I saw the CPU utilization with this is about half of what it is on the 1.x line at 20k sentences per second.  Also the latencies all across the board are lower, not by a lot but a few ms.  Also the JITTER seems to be doing a lot better job on the java code and it was able to hit a steady state at 20k/sec in abut 30 seconds where as the clojure code was taking almost 3 mins to hit a steady state.

abhishekagarwal87:
That's good to hear Robert. I will resolve the conflicts and incorporate your suggestions. Thank you for your help.

abhishekagarwal87:
The PR is ready now.

storm-core/src/clj/org/apache/storm/testing.clj
```diff
@@ -667,23 +665,18 @@
           (.put "spout-emitted" (AtomicInteger. 0))
           (.put "transferred" (AtomicInteger. 0))
           (.put "processed" (AtomicInteger. 0))))
+      (LocalExecutor/setTrackId id#)
       (with-var-roots
```
revans2: We don't need with-var-roots any more, we could just call with simulated-time-local-cluster without it

revans2:
Looks great I am +1 even without the last comment I made (That code can go away when it is translated to java).

But since I also contributed some of the test code changes I really would like someone else to have a look too.

revans2:
@abhishekagarwal87 I know you are busy. If you don't have time to do the rework for this I would be happy to do it for you.

HeartSaVioR:
@abhishekagarwal87 @revans2 
I'm even OK that if @revans2 crafts a new PR on top of this, or file an issue regarding my comments and merge first @abhishekagarwal87 doesn't mind.

abhishekagarwal87:
I dont mind. Sorry about this. But it will take me sometime before I get back to the PR.

revans2:
OK I'll get started on this.  Should hopefully have something up soon.

storm-core/src/jvm/org/apache/storm/daemon/worker/Worker.java
```diff
@@ -0,0 +1,426 @@
...
+
+
+    /**
+     * make a handler for the worker's send disruptor queue to
+     * check highWaterMark and lowWaterMark for backpressure
+     */
+    private DisruptorBackpressureCallback mkDisruptorBackpressureHandler(WorkerState workerState) {
+        return new DisruptorBackpressureCallback() {
+            @Override public void highWaterMark() throws Exception {
+                workerState.transferBackpressure.set(true);
```
revans2: You are correct that was a bug that was fixed, but the logs were not fixed.  I'll update them to be correct.

revans2:
FYI I put up a separate pull request based off of this one at #1775

PR #1758, opened 2016-11-02
STORM-2185: Storm Supervisor doesn't delete directories properly sometimes

knusbaum:


revans2:
+1 I see that the race seems to be that if a worker is shut down after creating the directory, but before running chown on it, then we hit this situation.

HeartSaVioR:
+1. Commit message should be modified but can be handled while merging.

PR #1761, opened 2016-11-03
STORM-2184: Don't wakeup KafkaConsumer on shutdown, spout methods are…

srdo:
... not called by multiple threads at a time

See https://issues.apache.org/jira/browse/STORM-2184


revans2:
+1 The discussion on STORM-2184 makes since that wakeup seems to not be needed.

HeartSaVioR:
+1 KafkaConsumer seems to be initialized for each task (so used from only one thread), and so as ConsumerNetworkClient in KafkaConsumer.

PR #1762, opened 2016-11-05
STORM-2188:Interrupt all executor threads before joining in executor shutdown, l…

srdo:
…og when joining threads

This makes it easier to sift through the thread dump ReadClusterState outputs if a Slot fails to shut down in time.


HeartSaVioR:
I think we need to do this. +1

PR #1764, opened 2016-11-07
STORM-2190: reduce contention between submission and scheduling

revans2:


ptgoetz:
One question about the for loop in WordCountTopology. I'm +1 once that's answered/addressed.

storm-core/src/clj/org/apache/storm/daemon/nimbus.clj
```diff
@@ -1008,23 +1008,24 @@
...
-          )))
-    (->> new-assignments
-          (map (fn [[topology-id assignment]]
-            (let [existing-assignment (get existing-assignments topology-id)]
-              [topology-id (map to-worker-slot (newly-added-slots existing-assignment assignment))]
-              )))
-          (into {})
-          (.assignSlots inimbus topologies)))
-    (log-message "not a leader, skipping assignments")))
+    (locking (:sched-lock nimbus)
```
jerrypeng: Couldn't a wrong ordering of events happen  since we are locking when calculating a scheduling then unlocking and then locking and uploading the new scheduling and unlocking
for example:
T0: submit
T1: rebalance
T2: rebalance - calculate new scheduling
T3: submit - calculate new scheduling
T4: rebalance - upload new scheduling to zk
T5: submit - upload new scheduling to zk

even though we should end up with the scheduling calculated by the rebalance but we end up with scheduling calculated from the original submit.

revans2: @jerrypeng You are correct that this could happen.  I don't really think it will be that likely to happen in practice but I'll think about it and see if we can fix it.

jerrypeng: maybe its also time to start thinking about decentralized scheduling mechanism if certain scheduling strategies may take a while to compute a schedule, but that would require a major overhaul in nimbus.

revans2: I agree and now that nimbus is in java we can look at doing some refactoring along those lines.  If you feel that we need to do it now and that this is a blocker I can spend some time looking into how to do that better.

revans2:
Just so we don't miss the comment from @jerrypeng 

> Couldn't a wrong ordering of events happen since we are locking when calculating a scheduling then unlocking and then locking and uploading the new scheduling and unlocking
> for example:
> T0: submit
> T1: rebalance
> T2: rebalance - calculate new scheduling
> T3: submit - calculate new scheduling
> T4: rebalance - upload new scheduling to zk
> T5: submit - upload new scheduling to zk
> 
> even though we should end up with the scheduling calculated by the rebalance but we end up with scheduling calculated from the original submit.

Yes, that is correct.  We should do something here, and he suggested that perhaps as part of a refactor of Nimbus we should look at support for long running scheduling.

In the short term I think I might make scheduling and writing to ZK atomic, but long term I think I will file a JIRA to look at better scheduling.

revans2:
@ptgoetz @jerrypeng 

I made a few changes to thing.  I fixed the race condition and I addressed the review comments, but I also put in some optimizations to storm submitter.  We were literally calling getClusterInfo 3+ times for each topology submission, and because the ultimate goal of STORM-2190 is to make it more scalable this helps a lot.  There is still some lock contention, but it is much better then it was before.

If things look good here I will backport the changes to my other pull request.

ptgoetz:
+1

HeartSaVioR:
+1. Please apply the optimization to #1765 as well.

jerrypeng:
LGTM +1 @revans2  thanks for making the optimizations

revans2:
@HeartSaVioR I plan on doing that.

HeartSaVioR:
+1 again.

PR #1765, opened 2016-11-07
STORM-2190: reduce contention between submission and scheduling

revans2:


examples/storm-starter/src/jvm/org/apache/storm/starter/WordCountTopology.java
```diff
@@ -90,7 +90,9 @@ public static void main(String[] args) throws Exception {
     if (args != null && args.length > 0) {
       conf.setNumWorkers(3);
 
-      StormSubmitter.submitTopologyWithProgressBar(args[0], conf, builder.createTopology());
+      for (String name: args) {
```
ptgoetz: I'm not sure I understand the for loop here. Isn't this change just submitting the same topology multiple times with different names? Is this an artifact from manual testing?

ptgoetz:
One question about the for loop in WordCountTopology. I'm +1 once that's answered/addressed.

examples/storm-starter/src/jvm/org/apache/storm/starter/WordCountTopology.java
revans2: This was code that I put in to be able to reproduce the problem consistently.  If you want me to remove it I am happy to.  It does not really provide any real value beyond that.

ptgoetz:
+1

PR #1766, opened 2016-11-08
[STORM-2192] Add a new IAutoCredentials plugin to support SSL files

tgravescs:
Allows you to send your SSL files with your topology and update them in a secure fashion such that they aren't stored in an intermediate on disk location. You enable it using the topology.auto-credentials config and specify org.apache.storm.security.auth.AutoSSL. The user can specify the keystore/truststore files when submitting the topology with  -c "ssl.credential.files=directory/YourFiles". The files will be serialized along with your credentials and written out into the storm workers current directory as the base filename specified (directory part is not included). You should reference them as ./filename in your topology config. 

For example:
storm jar myjar hadooptest.topologies.KafkaSpoutTopology testtopo -c "ssl.credential.files=/home/user1/kafka.server.keystore.jks,/home/user1/kafka.server.truststore.jks"
You can also update the files if they are going to expire using the storm upload-credentials command
For example:
storm upload-credentials testtopo -c "ssl.credential.files=/home/user1/kafka.server.keystore.jks,/home/user1/kafka.server.truststore.jks"

### DIFF
--- a/storm-core/src/jvm/org/apache/storm/security/auth/AutoSSL.java
+++ b/storm-core/src/jvm/org/apache/storm/security/auth/AutoSSL.java
@@ -1,0 +1,161 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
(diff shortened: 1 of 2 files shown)


knusbaum:
+1

PR #1768, opened 2016-11-09
STORM-2194: Report error and die, not report error or die

chawco:
This appears to have just ported an existing bug from executor.clj -- this is what I believe the expected behaviour is/was.


revans2:
-1.

InterruptedException we often use to indicate that the process is shutting down, and we want to not blow up when we see them.  Perhaps what we want to do is to better document it.

chawco:
I've left some follow up comments in [1767](https://github.com/apache/storm/pull/1767).

PR #1770, opened 2016-11-10
STORM 2197: NimbusClient connections leak due to ThriftClient connection's leakage in case of errors.

satishd:
- NimbusClient connections leak due to ThriftClient connection's leakage in case of errors.


HeartSaVioR:
+1 Nice finding.

vesense:
+1

PR #1776, opened 2016-11-15
STORM-2200 [Storm SQL] Drop Aggregate & Join support on Trident mode

HeartSaVioR:
* Dropping aggregate & join support based on discussion
* These implementation will be reused when we introduce them again

Please note that this will only affect Trident mode: Standalone mode still supports aggregation & join.


HeartSaVioR:
STORM-1446 also addresses STORM-2200, so this patch doesn't need to be merged when STORM-1446 is merged earlier.

HeartSaVioR:
Just upmerged.

HeartSaVioR:
I'd like to get this merged before releasing 1.1.0 so that users don't experience existing features are dropped. If someone can take a look I'd really appreciate it.

harshach:
+1

PR #1779, opened 2016-11-15
STORM-1694: Kafka Spout Trident Implementation Using New Kafka Consumer API

hmcl:
@harshach I refactored the patch merged onto master to make it compile. The changes are to remove the use of lambdas and to fix some generics compile-time type resolution.


harshach:
+1

PR #1781, opened 2016-11-16
STORM-1369: Add MapState implementation to storm-cassandra.

mkoch1:
STORM-1369: Add MapState implementation to storm-cassandra.

https://issues.apache.org/jira/browse/STORM-1369

### DIFF
--- a/external/storm-cassandra/src/main/java/org/apache/storm/cassandra/executor/AsyncExecutor.java
+++ b/external/storm-cassandra/src/main/java/org/apache/storm/cassandra/executor/AsyncExecutor.java
@@ -22,17 +22,22 @@ import com.datastax.driver.core.ResultSet;
 import com.datastax.driver.core.ResultSetFuture;
 import com.datastax.driver.core.Session;
 import com.datastax.driver.core.Statement;
-import com.google.common.util.concurrent.*;
+import com.google.common.util.concurrent.FutureCallback;
+import com.google.common.util.concurrent.Futures;
+import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.SettableFuture;
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
+import org.apache.storm.topology.FailedException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.Serializable;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
-import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.Semaphore;
 import java.util.concurrent.atomic.AtomicInteger;
 
 /**
@@ -106,6 +111,7 @@ public class AsyncExecutor<T> implements Serializable {
     public SettableFuture<T> execAsync(final Statement statement, final T inputs) {
         return execAsync(statement, inputs, handler);
     }
+
     /**
      * Asynchronously executes the specified batch statement. Inputs will be passed to
      * the {@link #handler} once query succeed or failed.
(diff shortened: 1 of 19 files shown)


external/storm-cassandra/src/main/java/org/apache/storm/cassandra/executor/AsyncExecutor.java
```diff
@@ -23,16 +23,16 @@
...
+import org.apache.storm.topology.FailedException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.Serializable;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
+import java.util.*;
```
revans2: In general we prefer to not use the wildcard on includes.

external/storm-cassandra/src/main/java/org/apache/storm/cassandra/executor/AsyncExecutor.java
```diff
@@ -138,6 +139,109 @@ public void onFailure(Throwable t) {
...
+
+        final SettableFuture<List<T>> settableFuture = SettableFuture.create();
+        final AsyncContext<T> asyncContext = new AsyncContext<>(inputs, maxParallelRequests, settableFuture);
+
+        for (int i = 0; i < statements.size(); i++) {
+
+            // Acquire a slot
+            if (asyncContext.acquire()) {
+
+
```
revans2: nit: don't need the extra line

external/storm-cassandra/src/main/java/org/apache/storm/cassandra/executor/AsyncExecutor.java
```diff
@@ -138,6 +139,109 @@ public void onFailure(Throwable t) {
...
+                            }
+                        }
+
+                        @Override
+                        public void onFailure(Throwable throwable) {
+                            handler.failure(throwable, input);
+                            asyncContext
+                                    .exception(throwable)
+                                    .release();
+                            LOG.error(String.format("Failed to execute statement '%s' ", statement), throwable);
```
revans2: do we need a `pending.decrementAndGet();` here too?

external/storm-cassandra/src/main/java/org/apache/storm/cassandra/trident/state/CassandraBackingMap.java
```diff
@@ -0,0 +1,232 @@
...
+        this.conf = conf;
+        this.options = options;
+        List<String> allFields = options.keyFields.toList();
+        allFields.addAll(options.stateMapper.getStateFields().toList());
+        this.allFields = new Fields(allFields);
+    }
+
+    @Override
+    public List<T> multiGet(List<List<Object>> keys) {
+        LOG.info("multiGet fetching {} values.", keys.size());
```
revans2: This should probably be debug.

external/storm-cassandra/src/main/java/org/apache/storm/cassandra/trident/state/CassandraBackingMap.java
```diff
@@ -0,0 +1,232 @@
...
+            states.add(state);
+        }
+
+        return states;
+
+    }
+
+    @Override
+    public void multiPut(List<List<Object>> keys, List<T> values) {
+        LOG.info("multiPut writing {} values.", keys.size());
```
revans2: debug

external/storm-cassandra/src/main/java/org/apache/storm/cassandra/executor/AsyncExecutor.java
mkoch1: Good catch, it should also handle exceptions like the onSuccess method.

mkoch1:
Pls note that mutliPut needs an update - will push commit today or tomorrow.

mkoch1:
That should do it - the batching option has been removed from MapState, in favor of parallel processing with existing opaque and transactional logic to handle consistency. Cassandra batch statements are more trouble than they're worth.

ptgoetz:
Unfortunately this patch uses Cassandra-unit, which is LGPL licensed, so in it's current state we can't accept it. For that reason I am -1. We've run into this before, and I've lobbied the Cassandra-unit author to switch to an Apache compatible license (https://github.com/jsevellec/cassandra-unit/issues/163) to no avail.

We can, however, accept it if the Cassandra-unit dependency and tests that depend on it are removed.

Again, it's unfortunate. If more people encourage the Cassandra-unit author to choose a different license, perhaps he'll consider it.

mkoch1:
Did not notice that, sorry. Replaced now with an external resource implementation. The code is not all that complex so it's not a big deal. 

One thing I should point out: I borrowed the base implementation from a different project, also under Apache 2.0 license. I've included their copyright line in the file: https://github.com/apache/storm/pull/1781/commits/e46496337af84a610ef742c8cf5988d8dab19e47#diff-a8ce0c2f5e3ce0193d3ebc49f54301fa

Original code: https://github.com/spring-projects/spring-insight-plugins/blob/master/collection-plugins/cassandra12/src/test/java/com/springsource/insight/plugin/cassandra/embeded/EmbeddedCassandraService.java

Please let me know if there are any issues with this.

ptgoetz:
@mkoch1 Thanks for addressing the license issue. The way you handled the borrowed code looks fine, the important part is to leave the original copyright statement, which you did. Thanks for pointing that out.

+1

HeartSaVioR:
@mkoch1 Could you rebase this? After rebasing I'll merge this.

mkoch1:
Rebase is done, sorry for the delay.

HeartSaVioR:
@mkoch1 
Could you squash commits into one if you don't mind? I know it's more than 2000 lines change but looks like separating commits are not needed.
And I would like to wait for 1.1.0 RC vote to pass to avoid any conflicts with the release. That says, this is not going to be included to 1.1.0.

mkoch1:
Squashed. I don't really see how this should cause any conflicts though. The only changes in existing code are a couple of namespace corrections (to avoid wildcard imports, as suggested earlier) and documentation. Other than that, it's all new code in new classes.

HeartSaVioR:
+1

PR #1785, opened 2016-11-17
[STORM-2201] Add dynamic scheduler configuration loading.

ppoulosk:
This adds an interface and two implementations, one that will load from a local file, and another
that will load a config from artifactory.

It also modifies the ResourceAwareScheduler and Multitenant schedulers to have a plugin interface and configuration entries for the plugin.

Unit tests are also added for the implementations.


storm-core/src/jvm/org/apache/storm/Config.java
```diff
@@ -2177,13 +2177,37 @@
...
      * per user resource guarantees.
      */
     @isMapEntryCustom(keyValidatorClasses = {StringValidator.class}, valueValidatorClasses = {UserResourcePoolEntryValidator.class})
     public static final String RESOURCE_AWARE_SCHEDULER_USER_POOLS = "resource.aware.scheduler.user.pools";
 
+     /**
+     * A plugin that should load the user pools for the resource aware scheduler
+     */
+    @isString
+    public static final String RESOURCE_AWARE_SCHEDULER_USER_POOLS_LOADER = "resource.aware.scheduler.user.pools.loader";
```
revans2: If this is intended to be a subclass of something, it might be good to use `@ isImplementationOfClass` instead of `@isString`

revans2: Specifically `@isImplementationOfClass(implementsClass = org.apache.storm.scheduler.utils.IConfigLoader.class)`

storm-core/src/jvm/org/apache/storm/scheduler/multitenant/MultitenantScheduler.java
```diff
@@ -29,19 +29,35 @@
...
   }
  
   private Map<String, Number> getUserConf() {
+    // Try the loader plugin, if configured
+    if (_configLoader != null) {
+        Map<String, Number> ret = (Map<String, Number>)_configLoader.load();
+        if (ret != null) {
+            return ret;
+        } else {
+            LOG.debug("Config loader returned null");
```
revans2: Should this be a warn instead of a debug?

storm-core/src/jvm/org/apache/storm/scheduler/utils/ArtifactoryConfigLoader.java
```diff
@@ -0,0 +1,388 @@
...
+import org.apache.http.client.utils.URIBuilder;
+import org.apache.http.impl.client.HttpClientBuilder; 
+import org.apache.http.client.HttpClient; 
+import org.apache.http.util.EntityUtils;
+import org.yaml.snakeyaml.Yaml;
+import org.yaml.snakeyaml.constructor.SafeConstructor;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ArtifactoryConfigLoader implements IConfigLoader {
```
revans2: Could you add some javadocs here?  and add something to the docs to describe this new feature, how it works and how to use it?

storm-core/src/jvm/org/apache/storm/scheduler/utils/ArtifactoryConfigLoader.java
```diff
@@ -0,0 +1,388 @@
...
+    private boolean _cacheInitialized = false;
+    // Location of the file in the artifactory archive.  Also used to name file in cache.
+    private String _localCacheDir;
+    private String _artifactoryScheme = "http";
+    private String _baseDirectory = "/artifactory";
+    private int _lastReturnedTime = 0;
+    private int _timeoutSeconds = 10;
+    private Map _lastReturnedValue;
+
+    private static class OurResponseHandler implements ResponseHandler<String> {
```
revans2: Perhaps we can name this a bit better.  It appears to validate that the response code was in the 2xx range and then convert the body to a string.  Maybe `GETStringResponseHandler` would be better?

storm-core/src/jvm/org/apache/storm/scheduler/utils/ArtifactoryConfigLoader.java
```diff
@@ -0,0 +1,388 @@
...
+    };
+
+    // Protected so we can override this in testing
+    protected String doGet(String api, String artifact, String host, Integer port) {
+        URIBuilder builder = new URIBuilder().setScheme(_artifactoryScheme).setHost(host).setPort(port);
+
+        if (api != null) {
+            builder.setPath(_baseDirectory + api + artifact);
+        } else {
+            builder.setPath(_baseDirectory + artifact);
```
revans2: Do we want to filter out double back slashes? "//".  If I remember correctly I think I have had issues with artifactory and too many / characters, almost like it is running jetty behind the scenes.

revans2: If we filter out double backslashes we can also add in extra ones to be sure we have a slash between api and artifact instead of just assuming there is one.

storm-core/src/jvm/org/apache/storm/scheduler/utils/IConfigLoader.java
```diff
@@ -0,0 +1,33 @@
...
+public interface IConfigLoader {
+
+    /**
+     * Pass in configurable parameters for the location of the data, (filename / artifactory url).
+     * 
+     * See implementing classes for configuration details.
+     */
+    void prepare(Map conf);
+
+    Map load();
```
revans2: Can we make this at least a little generic?  Something like `Map<?,?>` instead of just `Map`

storm-core/src/jvm/org/apache/storm/scheduler/utils/IConfigLoader.java
```diff
@@ -0,0 +1,33 @@
...
+
+
+public interface IConfigLoader {
+
+    /**
+     * Pass in configurable parameters for the location of the data, (filename / artifactory url).
+     * 
+     * See implementing classes for configuration details.
+     */
+    void prepare(Map conf);
```
revans2: For conf can be please add the generics? `Map<String, Object>`

storm-core/src/jvm/org/apache/storm/scheduler/utils/SchedulerUtils.java
```diff
@@ -0,0 +1,64 @@
...
+                if (loader != null) {
+                    loader.prepare(loaderConf);
+                    return loader;
+                }
+            }
+        }
+        return null;
+    }
+
+    public static Map loadYamlFromFile(File file) {
```
revans2: Can we please add is some generics to the Map.

revans2: Can we also add in some javadocs or possibly change the name?  It is not obvious from the name or really anything else that it will eat exceptions and return null.

storm-core/src/jvm/org/apache/storm/scheduler/utils/SchedulerUtils.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.Map;
+import java.io.File;
+import org.yaml.snakeyaml.Yaml;
+import org.yaml.snakeyaml.constructor.SafeConstructor;
+import java.io.FileInputStream;
+import java.io.InputStreamReader;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class SchedulerUtils {
```
revans2: Now that we are on java 8 it might be good to move these static methods into the IConfigLoader interface.  in java 8 static methods can be part of an interface.

storm-core/src/jvm/org/apache/storm/Config.java
jerrypeng: Yup that is what @isImplementationOfClass is for.  We should use it

ppoulosk:
@revans2, Thanks for the review. I've merged in a commit to fix the issues you have pointed out.

abellina:
@ppoulosk can we add a document as @revans2 suggested? A markdown file should do it.

ppoulosk:
Markdown added for interface and describing the configuration of implementations.

abellina:
thanks @ppoulosk, will do another pass today.

ppoulosk:
@abellina @revans2 Could you look at the re-work on this pull?

docs/IConfigLoader.md
```diff
@@ -0,0 +1,58 @@
+---
+title: IConfigLoader
+layout: documentation
+documentation: true
+---
+
+
+### Introduction
+IConfigLoader is an interface designed to allow way to dynamically load scheduler resource constraints into scheduler implementations. Currently, the MultiTenant scheduler uses this interface to dynamically load the number of isolated nodes a given user has been guaranteed, and the ResoureAwareScheduler uses the interface to dynamically load per user resource guarantees.
```
revans2: Perhaps "designed to allow dynamic loading of scheduler resource constraints"

pom.xml
```diff
@@ -220,6 +220,7 @@
         <java_jmx.version>0.3.1</java_jmx.version>
         <compojure.version>1.1.9</compojure.version>
         <hiccup.version>0.3.6</hiccup.version>
+        <commons-logging.version>1.2</commons-logging.version>
```
abellina: why are we adding commons-logging? I think I am missing something.

storm-core/src/jvm/org/apache/storm/scheduler/resource/ResourceAwareScheduler.java
```diff
@@ -376,18 +388,12 @@ private void initialize(Topologies topologies, Cluster cluster) {
...
-     */
-    private Map<String, Map<String, Double>> getUserResourcePools() {
-        Object raw = this.conf.get(Config.RESOURCE_AWARE_SCHEDULER_USER_POOLS);
+    private Map<String, Map<String, Double>> convertToDouble(Map<String, Map<String, Number>> raw) {
+
         Map<String, Map<String, Double>> ret = new HashMap<String, Map<String, Double>>();
 
         if (raw != null) {
-            for (Map.Entry<String, Map<String, Number>> userPoolEntry : ((Map<String, Map<String, Number>>) raw).entrySet()) {
+            for (Map.Entry<String, Map<String, Number>> userPoolEntry : raw.entrySet()) {
```
abellina: nice catch

storm-core/src/jvm/org/apache/storm/scheduler/utils/ArtifactoryConfigLoader.java
```diff
@@ -0,0 +1,429 @@
...
+     * @return null on failure or the response string if return code is in 200 range
+     * <p>
+     * Protected so we can override this in unit tests
+     */
+    protected String doGet(String api, String artifact, String host, Integer port) {
+        URIBuilder builder = new URIBuilder().setScheme(_artifactoryScheme).setHost(host).setPort(port);
+
+        String path = null;
+        if (api != null) {
+            path = _baseDirectory + "/" + api + "/" + artifact;
```
abellina: instead of "/" here I'd use the .path method in URIBuilder to build up the path.

storm-core/src/jvm/org/apache/storm/scheduler/utils/ArtifactoryConfigLoader.java
```diff
@@ -0,0 +1,429 @@
...
+        } else {
+            path = _baseDirectory + "/" + artifact;
+        }
+
+        // Get rid of multiple '/' in url
+        path = path.replaceAll("/[/]+", "/");
```
abellina: if you do above, you may not need this one.

HeartSaVioR:
@ppoulosk Could you address @revans2 and @abellina review comments?

ppoulosk:
Will get to this today or tomorrow.  Sorry this slipped off of my radar.

pom.xml
ppoulosk: Fixing this.

docs/IConfigLoader.md
ppoulosk: Much better wording.  Thanks.

storm-core/src/jvm/org/apache/storm/scheduler/utils/ArtifactoryConfigLoader.java
ppoulosk: there is no .path method in URIbuilder.  It's in javax.ws.rs.core.UriBuilder, which isn't pulled into storm anywhere.  I'd prefer to leave it as it, but will change and pull in new UriBuilder class if there is a consensus to do so.  @revans2?

storm-core/src/jvm/org/apache/storm/scheduler/utils/ArtifactoryConfigLoader.java
ppoulosk: Same.

HeartSaVioR:
@ppoulosk Any updates?
@Ethanlm I guess you're taking over @ppoulosk work. Could you also consider taking over this work as well if possible?
This patch now addresses all comments from @revans2 and @abellina so seems like close to ready to merge. So if you and @ppoulosk don't want to continue this work, I'll just pick this and rebase.

Ethanlm:
@HeartSaVioR  I am willing to do it but I will need to ask @revans2 for his opinion. Thanks

Ethanlm:
@HeartSaVioR I asked @revans2 . I will start to do it.

Ethanlm:
I copied the code and submitted a new pull request. Also fixed some checkstyle issues. https://github.com/apache/storm/pull/2199

PR #1744, opened 2016-10-24
STORM-1276: line for line translation of nimbus to java

revans2:
There are some things I would like to refactor, but this is a function tested translation that I would like to get it, at least as a starting point.

PR #1786, opened 2016-11-18
STORM-1281: LocalCluster, testing4j and testing.clj to java

revans2:
This is based off of #1744, but I wanted to make sure that my changes were publicly available.

Now Testing.java replaces testing4j.clj and depends on an improved LocalCluster.java that replaces LocalCluster.clj.

testing.clj now depends on Testing.java and has been moved under storm-clojure.


storm-core/test/clj/org/apache/storm/clojure_test.clj
```diff
@@ -1,159 +0,0 @@
-;; Licensed to the Apache Software Foundation (ASF) under one
```
revans2: There is a copy of this in storm-clojure already so no need to two of them.

revans2:
The test failures look unrelated.  They seem to be the same windowing integration tests failing again.

revans2:
Rebased...

storm-core/src/jvm/org/apache/storm/testing/CompletableSpout.java
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.storm.testing;
+
+public interface CompletableSpout {
```
ptgoetz: Just a side note: The concept of a `CompleteableSpout` or `BoundedSpout` would be useful for batch-like flows (and would be necessary for implementing bounded flows in the beam runner). It may be worth considering moving this interface to a higher level (i.e. not in the "testing" package).

ptgoetz: Didn't mean to imply that the class should be moved now as part of this pull request, but rather that it's something we should consider doing later.

ptgoetz:
+1. I haven't done any manual testing of this patch, but I reviewed the code and it looks good.

I wouldn't mind seeing more javadoc comments in some of the new classes/interfaces, but it can always be added later, so I don't see that as a merge blocker or something that needs to be addressed right away.

storm-core/src/jvm/org/apache/storm/testing/CompletableSpout.java
revans2: I understood you.  Thanks for the clarification though.

revans2:
@ptgoetz I'll look at adding in some javadocs.  In general I tried to add javadocs where there were clojure doc strings, but there were not too many of them either.

revans2:
@ptgoetz I added in javadocs for most of the new classes

revans2:
Rebased now that java nimbus is in.

ptgoetz:
@revans2 Thanks. Still +1.

(One procedural nit: Rebasing/squashing on big PRs like this can make reviews a little harder. E.g. After you added the javadocs I wanted to review just that change, but the squash forced me to wade through all the other changes to find the ones I was looking for. Not a big deal, but something to keep in mind.)

PR #1787, opened 2016-11-19
STORM-2208: HDFS State Throws FileNotFoundException in Azure Data Lake Store file system (adl://)

hmcl:
- Close OutputStream before file gets deleted

### DIFF
--- a/external/storm-hdfs/src/main/java/org/apache/storm/hdfs/trident/HdfsState.java
+++ b/external/storm-hdfs/src/main/java/org/apache/storm/hdfs/trident/HdfsState.java
@@ -17,9 +17,6 @@
  */
 package org.apache.storm.hdfs.trident;
 
-import org.apache.storm.Config;
-import org.apache.storm.task.IMetricsContext;
-import org.apache.storm.topology.FailedException;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FSDataInputStream;
 import org.apache.hadoop.fs.FSDataOutputStream;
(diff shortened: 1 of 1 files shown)


hmcl:
Please backport to 1.x and 1.0.x branch. It should be a trivial merge.

harshach:
+1

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/trident/HdfsState.java
```diff
@@ -500,6 +500,7 @@ private void updateIndex(long txId) {
             bw.write(txnRecord.toString());
             bw.newLine();
             bw.flush();
+            out.close();       // Closing the output stream in finally block causes FileNotFoundException in Azure Data Lake Store file system (adl://)
```
revans2: Why not just wrap the output stream in a try block instead?

```
Path tmpPath = tmpFilePath(indexFilePath.toString());
TxnRecord txnRecord = new TxnRecord(txId, options.currentFile.toString(), this.options.getCurrentOffset());
try (FSDataOutputStream out = this.options.fs.create(tmpPath, true);
       BufferedWriter bw = new BufferedWriter(new OutputStreamWriter(out));) {
    bw.write(txnRecord.toString());
    bw.newLine();
    bw.flush();
}
```

Then you don't need to worry about the close happening before the delete, even in error cases.

hmcl:
@revans2 agree && Done! 

I followed your suggestion but most calls to fs will throw IOException. To keep only one try catch block the cleanest way to do it is to close the OutputStream inside the try block. Otherwise, we need 2 try / catch blocks.

revans2:
+1 looks god to me, although I am not an expert on azure, so I assume you have tested the new code and it works.

arunmahadevan:
+1, the changes look good. Not being able to close the stream after renaming / deleting the file is deviation from the hdfs/java fs standard which should also be supported by the azure implementation.

hmcl:
@arunmahadevan I agree! I have already filed a but to the Azure team to fix that case.

However, it does no harm that we account for this scenario.

hmcl:
@revans2 I just tested in the Azure cluster running AFS and the fix works. Please see attached screenshots.

broken_2016-11-22_6 43 08pm
fixed_2016-11-22_6 34 29pm

HeartSaVioR:
+1

PR #1791, opened 2016-11-22
STORM-2212: Remove Redundant Declarations in Maven POM Files

hmcl:
- Remove redundant declarations cause warnings and make the build files harder to extend and maintain


hmcl:
The variable ${storm.topology} is not declared anywhere, hence why it was removed

dossett:
+1

revans2:
+1

PR #1793, opened 2016-11-22
STORM-2214: add in cacheing of the Login

revans2:


storm-core/src/jvm/org/apache/storm/security/auth/kerberos/KerberosSaslTransportPlugin.java
```diff
@@ -50,6 +52,44 @@
 public class KerberosSaslTransportPlugin extends SaslTransportPlugin {
     public static final String KERBEROS = "GSSAPI"; 
     private static final Logger LOG = LoggerFactory.getLogger(KerberosSaslTransportPlugin.class);
+    private static Map <LoginCacheKey, Login> loginCache = new ConcurrentHashMap<>();
```
revans2: In case people are worried about the cache growing without bound the size of the cache will be proportional to the size of your jaas conf, as the key to the cache is essentially the jaas conf entry for the Login.

harshach:
+1. @revans2 so the issue without this patch is we are making a login everytime a client tries to connect.

revans2:
@harshach the issue is that when there are lots of supervisors the load placed on the KDC is a lot higher than it is for a similar number of Hadoop nodes.  Each new Login will contact the KDC and fetch a new TGT followed by a new service ticket.  If we can reuse the Subject then the TGT and the service tickets can be reused too reducing the load on the KDC.  The Login handles the life cycle of the Subject, creates it and destroys it, so we wanted to cache it at that level to avoid issues there.

revans2:
I also wanted to say we have been running with this in prod for a while now.

PR #1795, opened 2016-11-22
STORM-2215: validate blobs are present before submitting

revans2:


storm-core/src/jvm/org/apache/storm/StormSubmitter.java
```diff
@@ -587,4 +592,10 @@ private static double getMaxExecutorMemoryUsageForTopo(StormTopology topology, M
         }
         return largestMemoryOperator;
     }
+
+    private static Set<String> getListOfKeysFromBlobStore(Map<String, Object> stormConf) {
+        NimbusBlobStore client = new NimbusBlobStore();
```
HeartSaVioR: Need to guarantee calling client.shutdown().

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1338,6 +1339,26 @@ public static void handleUncaughtException(Throwable t) {
         }
     }
 
+    public static void validateTopologyBlobStoreMap(Map stormConf, Set<String> blobStoreKeys) throws InvalidTopologyException {
+        boolean containsAllBlobs = true;
```
HeartSaVioR: Minor: We're adding missing keys to Set so no need to have additional flag. `missingKeys.size() > 0` is same to `containsAllBlobs == false`.

revans2:
@HeartSaVioR I rebased and addressed the review comments, but I want to merge in #1744 first and then I will update this one again.

revans2:
#1744 is in and this is rebased

HeartSaVioR:
+1 Thanks for the work.

PR #1797, opened 2016-11-23
STORM-2210 remove array shuffle from ShuffleGrouping when counter res…

kevpeek:
This is a squashed version of https://github.com/apache/storm/pull/1790 made against master.


revans2:
+1, because this is from #1790 that was already approved I will not wait the full 24 hours.

PR #1798, opened 2016-11-25
Storm-2203 Add a getAll method to KeyValueState interface

aandis:
Adds a `getAll` to `InMemoryKeyValueState` and `RedisKeyValueState`. I'll add tests once the changes are approved.

### DIFF
--- a/external/storm-redis/src/main/java/org/apache/storm/redis/state/RedisKeyValueState.java
+++ b/external/storm-redis/src/main/java/org/apache/storm/redis/state/RedisKeyValueState.java
@@ -168,6 +168,27 @@ public class RedisKeyValueState<K, V> implements KeyValueState<K, V> {
     }
 
     @Override
+    public Map<K, V> getAll() {
+        Map<K, V> keyValues = new HashMap<K, V>();
+        Map<String, String> redisKeyValues = null;
+        JedisCommands commands = null;
+        try {
+            commands = jedisContainer.getInstance();
+            redisKeyValues = commands.hgetAll(namespace);
+        } finally {
+            jedisContainer.returnInstance(commands);
+        }
+        if (redisKeyValues != null) {
+            for (Map.Entry<String, String> entry: redisKeyValues.entrySet()) {
+                K key = keySerializer.deserialize(decode(entry.getKey()));
+                V value = valueSerializer.deserialize(decode(entry.getValue()));
+                keyValues.put(key, value);
+            }
+        }
+        return Collections.unmodifiableMap(keyValues);
+    }
+
+    @Override
     public void prepareCommit(long txid) {
         LOG.debug("prepareCommit txid {}", txid);
         validatePrepareTxid(txid);
--- a/external/storm-redis/src/main/java/org/apache/storm/redis/state/RedisKeyValueState.java
+++ b/external/storm-redis/src/main/java/org/apache/storm/redis/state/RedisKeyValueState.java
@@ -17,20 +17,20 @@
  */
 package org.apache.storm.redis.state;
 
-import com.google.common.base.Optional;
 import org.apache.storm.state.DefaultStateSerializer;
 import org.apache.storm.state.KeyValueState;
 import org.apache.storm.state.Serializer;
 import org.apache.storm.redis.common.config.JedisPoolConfig;
 import org.apache.storm.redis.common.container.JedisCommandsContainerBuilder;
 import org.apache.storm.redis.common.container.JedisCommandsInstanceContainer;
+import org.apache.storm.redis.utils.RedisEncoder;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import redis.clients.jedis.JedisCommands;
-import org.apache.commons.codec.binary.Base64;
 
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.Map;
 import java.util.List;
 import java.util.ArrayList;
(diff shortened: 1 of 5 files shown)


HeartSaVioR:
@aandis 
Looks great. Could you add tests if possible? You may want to rename the prefix of the title to `STORM-2203` so that it can be linked to JIRA issue automatically.
Thanks in advance!

storm-core/src/jvm/org/apache/storm/state/KeyValueState.java
```diff
@@ -45,4 +47,9 @@
      * @return the value or defaultValue if no mapping is found
      */
     V get(K key, V defaultValue);
+
+    /**
+     * @return all key value mappings as an unmodifiable map.
+     */
+    Map<K, V> getAll();
```
arunmahadevan: Return an Iterator of the entries in the map than the map itself. That way it will be more manageable even if the state is huge.

aandis: @arunmahadevan You mean an `iterator` over the original [state](https://github.com/apache/storm/pull/1798/files#diff-ad463cc68c1b4705bce8681f13dcd6d4R35)? `iterator` also has a `remove` method which shouldn't be a supported operation over `KeyValueState` since there's [delete](https://github.com/apache/storm/commit/285668742742da6316d38c1ef492c7c873b5a649#diff-4ac32a78649ca5bdd8e0ba38b7006a1eR2). That's the reason for creating and returning an unmodifiable map. :)

arunmahadevan: >You mean an iterator over the original state? 

yes it can be an `Iterator<Map.Entry<K, V>>`. 

the issue with map is that you need to load the entire state into the map before you can return and it will not scale if the state is huge.
 
>iterator also has a remove method which shouldn't be a supported operation 

remove may be good to have, but to keep it simple can throw an UnsupportedOperationException

arunmahadevan: At a high level I think you should be able to use redis hscan to implement the iterator and deserialize the key/values as you return them.

HeartSaVioR: You can implement custom iterator which stores cursor for `hscan` as internal state. Call hscan to get some elements and store to `queue` as internal state, pop and return it when next() is called. If queue is empty you can hscan again to determine there're more elements in the hash or it is end of elements in hash.

aandis: cool. I wasn't aware `hscan` gives you a `cursor`, redis is sort of new to me. I am gonna look into it and will update the pr. thanks guys. :)

aandis:
Ping @HeartSaVioR @arunmahadevan can you review this?

storm-core/src/jvm/org/apache/storm/state/KeyValueState.java
```diff
@@ -52,4 +55,9 @@
      * @param key   the key
      */
     V delete(K key);
+
+    /**
+     * @return an iterator over all key value mappings.
+     */
+    Iterator<Map.Entry<K, V>> iterator();
```
arunmahadevan: The KeyValueState can extend `Iterable<Entry<K, V>>` so that the state can be directly used in foreach constructs.

external/storm-redis/src/main/java/org/apache/storm/redis/state/RedisKeyValueState.java
```diff
@@ -161,13 +153,19 @@ public V get(K key, V defaultValue) {
...
+        String redisKey = encoder.encodeKey(key);
         V curr = get(key);
-        pendingPrepare.put(redisKey, TOMBSTONE);
+        pendingPrepare.put(redisKey, RedisEncoder.TOMBSTONE);
         return curr;
     }
 
     @Override
+    public Iterator<Map.Entry<K, V>> iterator() {
+        return new RedisKeyValueStateIterator<K, V>(namespace, jedisContainer,
```
arunmahadevan: You also need to consider the entries in `pendingPrepare` and `pendingCommit`. See the `get` api for example. `RedisKeyValueStateIterator` should iterate over pendingPrepare and pendingCommit before iterating over the entries in Redis.

HeartSaVioR:
@aandis Could you address review comments from @arunmahadevan?

aandis:
@HeartSaVioR yep. will do in next couple of days.

aandis:
@HeartSaVioR @arunmahadevan ready for review.

external/storm-redis/src/main/java/org/apache/storm/redis/state/RedisKeyValueStateIterator.java
```diff
@@ -0,0 +1,104 @@
...
+            return true;
+        } else {
+            return !cursor.equals("0");
+        }
+    }
+
+    @Override
+    public Map.Entry<K, V> next() {
+        Map.Entry<String, String> redisKeyValue = null;
+        if (pendingPrepareIterator != null && pendingPrepareIterator.hasNext()) {
```
aandis: > Instead of repeating the checks may be assign the current pending iterator to a variable in hasNext() and reuse it here.

This assumes `hasNext` is always called before calling `next` which may not be true isn't it?

arunmahadevan: see previous comment, call hasNext within next to validate and throw exception.

arunmahadevan:
+1

arunmahadevan:
@aandis thanks for the work, merged to master and 1.x-branch. You can close the PR.

aandis:
Thanks @arunmahadevan. :) Out of curiosity, why don't we use github's merge branch feature to merge branches directly from this page? Or was it just a temporary thing because of integration tests failing here?

PR #1807, opened 2016-12-01
fix NullPointException with acked.get(rtp)

cutd:
acked.get(rtp) maybe null when antocommit is true.

### DIFF
--- a/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
+++ b/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
@@ -262,11 +262,13 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
         final Set<TopicPartition> retriableTopicPartitions = retryService.retriableTopicPartitions();
 
         for (TopicPartition rtp : retriableTopicPartitions) {
-            final OffsetAndMetadata offsetAndMeta = acked.get(rtp).findNextCommitOffset();
-            if (offsetAndMeta != null) {
-                kafkaConsumer.seek(rtp, offsetAndMeta.offset() + 1);  // seek to the next offset that is ready to commit in next commit cycle
-            } else {
-                kafkaConsumer.seek(rtp, acked.get(rtp).committedOffset + 1);    // Seek to last committed offset
+            if (acked.get(rtp)!=null) {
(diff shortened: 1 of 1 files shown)


srdo:
Good catch. I think I'd prefer disabling the retry logic entirely when auto commit is enabled though. It doesn't really make sense to me that you would want to have Storm guarantee at least once processing for tuples from the Kafka spout, but also don't care about losing tuples if the Kafka spout crashes. It would be good IMO if we added the same type of check to fail() as ack() has, so we skip scheduling tuples for retry if auto commit is enabled.

@hmcl do you have any thoughts on this?

hmcl:
@srdo @cutd looking at it.

hmcl:
Two things.

@cutd I am +1 for your patch.

@srdo I see your point. If I understand correctly what you mean is that either the Spout guarantees at least once or it doesn't. If it doesn't, we shouldn't even bother to retry at all. The only thing I wonder is if a tuple fails occasionally, if one should try to retry it from memory, hoping for a best effort. What do you think?

srdo:
@hmcl Yes, the case I'm thinking about is where the topology has acking enabled, but the spout is set to auto-commit mode. It would probably be fine to do best-effort retrying when using that setup, but it seems like kind of an odd case to support. The retry code needs to be changed to support that case as well, since doSeekRetriableTopicPartitions doesn't work at all when auto-commit is enabled, because it depends on acked, so I'm not really sure it's worth it.

Regarding the patch and my previous comment, I still think moving this check into fail() and just disabling retries if auto commit is enabled is a better option (i.e. if auto-commit is on, the only thing fail() needs to do is remove the message from emitted and nothing else). It has the same effect as this patch, but we can ensure that it only has effect if auto-commit is enabled. The patch as is could end up hiding bugs, since doSeekRetriableTopicPartitions now quietly skips topicpartitions where acked is null. It's also easier to tell from reading the code why failing tuples would be disabled when auto-commit is on, than it is to tell why there's a null check in doSeekRetriableTopicPartitions.

cutd:
@srdo I think you are right!

srdo:
@cutd Great. Could you update the PR so it changes `fail` instead of `doSeekRetriableTopicPartitions`? You might also want to create an issue at https://issues.apache.org/jira/browse/STORM so this fix can be tracked :)

HeartSaVioR:
@cutd Any updates here? I'd like to see this PR updated with @srdo recommendation.

PR #1808, opened 2016-12-01
STORM-2225: change spout config to be simpler.

revans2:
STORM-2228: removed ability to request a single topic go to multiple streams


revans2:
I updated the docs and the examples.  This is a non-backwards compatible change from the 1.x release.  I have some code (not included here) that can maintain most compatibility if we really want to, but because of STORM-2228 I decided it would be better to just not support that at all.  If we want to add that compatibility for 1.x  but deprecated I am OK with that, but would like feedback on it before putting up some kind of a pull request.

d2r:
I see some duplicate code and documentation.  Is the idea to move from /external/storm-kafka to /external/storm-kafka-client?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutConfig.java
```diff
@@ -60,123 +68,197 @@
...
+         */
+        public Builder<K,V> setBootstrapServers(String servers) {
+            return setProp(BOOTSTRAP_SERVERS_CONF, servers);
+        }
+        
+        /**
+         * The minimum amount of data the broker should return for a fetch request.
+         */
+        public Builder<K,V> setFetchMinBytes(int bytes) {
+            return setProp("fetch.min.bytes", bytes);
```
srdo: Nit: Consider using some of the constants in `org.apache.kafka.clients.consumer.ConsumerConfig` instead

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutConfig.java
```diff
@@ -60,123 +68,197 @@
...
+        public Builder<K,V> setMaxPartitionFectchBytes(int bytes) {
+            return setProp("max.partition.fetch.bytes", bytes);
+        }
+        
+        /**
+         * The maximum number of records a poll will return.
+         * Only will work with Kafak 0.10.0 and above.
+         */
+        public Builder<K,V> setMaxPoolRecords(int records) {
+            return setProp("max.poll.recordes", records);
```
srdo: recordes -> records

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutConfig.java
```diff
@@ -60,123 +68,197 @@
...
+        /**
+         * The maximum amount of data per-partition the broker will return.
+         */
+        public Builder<K,V> setMaxPartitionFectchBytes(int bytes) {
+            return setProp("max.partition.fetch.bytes", bytes);
+        }
+        
+        /**
+         * The maximum number of records a poll will return.
+         * Only will work with Kafak 0.10.0 and above.
```
srdo: kafak -> kafka, and "will only work" instead of "only will work"

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/NamedSubscription.java
```diff
@@ -0,0 +1,66 @@
...
+
+    @Override
+    public <K, V> void subscribe(KafkaConsumer<K, V> consumer, ConsumerRebalanceListener listener) {
+        consumer.subscribe(topics, listener);
+        LOG.info("Kafka consumer subscribed topics {}", topics);
+    }
+
+    @Override
+    public String getTopicsString() {
+        StringBuilder ret = new StringBuilder();
```
srdo: Why not use https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#join-java.lang.CharSequence-java.lang.Iterable-?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/RecordTranslator.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.function.Function;
+
+import org.apache.kafka.clients.consumer.ConsumerRecord;
+import org.apache.storm.tuple.Fields;
+
+/**
+ * Translates a Kafka ConsumerRecord into a tuple
```
srdo: Consider adding a period at the end here, when shown as javadoc this is going to look weird otherwise.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/RecordTranslator.java
```diff
@@ -0,0 +1,59 @@
...
+ * use {@link KafkaTuple} if you want to route your tuples
+ * to different streams.
+ */
+public interface RecordTranslator<K, V> extends Serializable, Function<ConsumerRecord<K, V>, List<Object>> {
+    public static final List<String> DEFAULT_STREAM = Collections.unmodifiableList(Arrays.asList("default"));
```
srdo: Collections.singletonList is an easier way to do this

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/RecordTranslator.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.function.Function;
+
+import org.apache.kafka.clients.consumer.ConsumerRecord;
+import org.apache.storm.tuple.Fields;
+
+/**
+ * Translates a Kafka ConsumerRecord into a tuple
+ * use {@link KafkaTuple} if you want to route your tuples
```
srdo: Nit: I think it's a little clearer to say "Translate to KafkaTuple if..." or "Return a KafkaTuple if..."

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -185,7 +178,9 @@ private long doSeek(TopicPartition tp, OffsetAndMetadata committedOffset) {
     }
 
     private Collection<TopicPartition> toArrayList(final TopicPartition tp) {
-        return new ArrayList<TopicPartition>(1){{add(tp);}};
+        ArrayList<TopicPartition> ret = new ArrayList<>(1);
```
srdo: Nit: Unless you need to modify the return value later, you can just return Collections.singleton(tp)

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/trident/KafkaTridentSpoutEmitter.java
```diff
@@ -152,7 +152,9 @@ private long seek(TopicPartition tp, KafkaTridentSpoutBatchMetadata<K, V> lastBa
     }
 
     private Collection<TopicPartition> toArrayList(final TopicPartition tp) {
-        return new ArrayList<TopicPartition>(1){{add(tp);}};
+        ArrayList<TopicPartition> ret = new ArrayList<>(1);
```
srdo: Nit: Same as in KafkaSpout

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/trident/KafkaTridentSpoutManager.java
```diff
@@ -44,29 +40,30 @@
...
-        kafkaSpoutStreams = kafkaSpoutConfig.getKafkaSpoutStreams();
-        tuplesBuilder = kafkaSpoutConfig.getTuplesBuilder();
+        RecordTranslator<K, V> translator = kafkaSpoutConfig.getTranslator();
+        Fields fields = null;
+        for (String stream: translator.streams()) {
+            if (fields == null) {
+                fields = translator.getFieldsFor(stream);
+            } else {
+                if (!fields.equals(translator.getFieldsFor(stream))) {
+                    throw new IllegalArgumentException("Trident Spouts do nut support multiple output Fields");
```
srdo: Nit: nut -> not. Also maybe rephrase as "must have the same fields for all streams"

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/bolt/KafkaBoltTest.java
```diff
@@ -0,0 +1,91 @@
...
+        final KafkaProducer<String, String> producer = mock(KafkaProducer.class);
+        when(producer.send(any(), any())).thenAnswer(new Answer<Object>() {
+            @Override
+            public Object answer(InvocationOnMock invocation) throws Throwable {
+                Callback c = (Callback)invocation.getArguments()[1];
+                c.onCompletion(null, null);
+                return null;
+            }
+        });
+        KafkaBolt<String, String> bolt = new KafkaBolt<String, String>() {
```
srdo: Nit: Can leave off the contents of `<>` on the right hand side in a few places here

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/ByTopicRecordTranslatorTest.java
```diff
@@ -0,0 +1,72 @@
...
+        assertEquals(new Fields("value"), trans.getFieldsFor("value-stream"));
+        assertEquals(Arrays.asList("THE VALUE"), trans.apply(cr2));
+        
+        ConsumerRecord<String, String> cr3 = new ConsumerRecord<>("TOPIC 2", 100, 100, "THE KEY", "THE VALUE");
+        assertEquals(new Fields("key", "value"), trans.getFieldsFor("key-value-stream"));
+        assertEquals(Arrays.asList("THE KEY", "THE VALUE"), trans.apply(cr3));
+    }
+    
+    @Test(expected = IllegalArgumentException.class)
+    public void testFiledCollision() {
```
srdo: filed -> field

srdo:
I like the new configuration design. Backward compatibility is not an issue for us. I'm wondering if enough people have even switched to the new spout yet to make backward compatibility for 1.x a must.

revans2:
@srdo for me backwards compatibility for 1.x is more a question of violating out versioning than anything else.

srdo:
Sure, that makes sense. Maybe storm-kafka-client should have been marked unstable since it's fairly new, it's still in a phase where the kinks are being worked out, even if the version number says differently.

docs/storm-kafka-client.md
```diff
@@ -1,90 +1,222 @@
...
+final TopologyBuilder tp = new TopologyBuilder();
 
+//By default all topics not covered by another rule will be emitted to "STREAM_1" as "topic", "key", and "value"
+ByTopicRecordTranslator<String, String> byTopic = new ByTopicRecordTranslator<>(
+    (r) -> new Values(r.topic(), r.key(), r.value()),
+    new Fields("topic", "key", "value"), "STREAM_1");
+//For topic_2 all events will be emitted to "STREAM_2" as just "key" and "value"
+byTopic.forTopic("topic_2", (r) -> new Values(r.key(), r.value()), new Fields("key", "value"), "STREAM_2");
+
+tp.setSpout("kafka_spout", new KafkaSpout<>(KafkaSpoutConfig.builder("127.0.0.1:" + port, Pattern.compile("topic_.*")).build()), 1);
```
revans2: That all depends on the topics defined in the kafka cluster.  If it only has `topic_2` and `topic` defined then yes, but if it has `topic_1` and `topic_2` defined then it will match both of them.

revans2: I'll try to make the docs clearer

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/bolt/KafkaBolt.java
```diff
@@ -0,0 +1,194 @@
...
+                Callback callback = null;
+
+                if (!fireAndForget && async) {
+                    callback = new Callback() {
+                        @Override
+                        public void onCompletion(RecordMetadata ignored, Exception e) {
+                            synchronized (collector) {
+                                if (e != null) {
+                                    collector.reportError(e);
+                                    collector.fail(input);
```
revans2: collector could always be called from background threads.  There, however, was a race in the original shuffle grouping that could crash your worker if you didn't synchronize around calls to the collector.  I fixed that a long time ago, but people are still very nervous about it, or are still confused about it.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/NamedSubscription.java
revans2: Another 1.8 addition that I didn't know about.

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/bolt/KafkaBoltTest.java
revans2: In this case because I am subclassing KafkaBolt it the compiler(at least in eclipse) actually complains that I am not allowed to do it.

revans2:
@srdo I think I addressed all of your review comments.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/bolt/KafkaBolt.java
srdo: Sounds good then, thanks

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/bolt/KafkaBoltTest.java
srdo: Right, can't do that. Sorry :)

srdo:
I think my only remaining nit is the spelling error in KafkaBolt. Thanks @revans2. I am +1 on this.

hmcl:
@revans2 I would like to take a look at this before this gets merged in.

revans2:
@srdo as part of backporting this to 1.x I am going to need to make a change to not use Function directly, because it is only in java 8.  So to maintain compatibility between 1.x and 2.x I am going to need to make a few changes in this patch too.

external/storm-kafka-client/README.md
```diff
@@ -1,191 +1,3 @@
...
-<br/>
-
-There will be a blog post coming soon analyzing the trade-offs of this tuning parameters, and comparing the performance of the Kafka Spouts using the Kafka client API introduced in 0.9 (new implementation) and in prior versions (prior implementation)
-
-#Future Work
- Implement comprehensive metrics. Trident spout is coming soon.
-
-## Committer Sponsors
- * Sriharsha Chintalapani ([sriharsha@apache.org](mailto:sriharsha@apache.org))
+Please see [here](../../docs/storm-kafka-client.md)
```
HeartSaVioR: We still need to keep the description because this file will be included to binary dist whereas docs is not.

revans2:
I think I have addressed all of the review comments so far.  I will try to get my 1.x version of the patch up shortly.

external/storm-kafka-client/README.md
```diff
@@ -1,191 +1,5 @@
-#Storm Kafka Spout with New Kafka Consumer API
+#Storm Kafka integration using the kafka-client jar
 
-Apache Storm Spout implementation to consume data from Apache Kafka versions 0.10 onwards (please see [Apache Kafka Version Compatibility] (#compatibility)). 
+Spouts and Bolts used to interact with Kafka thought the kafka-client library.
```
srdo: Nit: "Thought" should be "through"

revans2:
All outstanding review comments should be done now.  This and the 1.x port at #1868 should be ready for a final pass and hopefully being merged in.

revans2:
The test failures are unrelated and are around the integration tests that always seem to fail lately.

revans2:
@srdo @harshach sorry to do this to you, but I just fixed the conflicts with STORM-2236.  Sadly the fastest way I could do it was to revert the original code and implement similar functionality, which is the latest commit.  Could you please take a look at it?  I created a few new Subscription implementations that can do the manual partition management.  It only needed a small change to the Spout to support a timeout.  I will try to look at adding some documentation and also the impact to the trident spout.

docs/storm-kafka-client.md
```diff
@@ -1,90 +1,232 @@
...
+tp.setSpout("kafka_spout", new KafkaSpout<>(KafkaSpoutConfig.builder("127.0.0.1:" + port, "topic").build()), 1);
+tp.setBolt("bolt", new myBolt()).shuffleGrouping("kafka_spout");
+...
 
-String STREAM = "test_wildcard_stream";
-String TOPIC_WILDCARD_PATTERN = "test[1|2]";
+```
+
+#### Wildcard Topics
+Wildcard topics will consume from all topics that are on the configured broakers and match the pattern.  So in the following example
```
hmcl: ... that exist in the brokers list specified and...

docs/storm-kafka-client.md
```diff
@@ -1,90 +1,232 @@
...
+"topic", "topic_foo" and "topic_bar" will all match the pattern, but "not_my_topic" would not match.
```
hmcl: the pattern "topic.*"

docs/storm-kafka-client.md
```diff
@@ -1,90 +1,232 @@
...
+### Configuration
+
+The spout implementations are configured by use of the `KafkaSpoutConfig` class.  This class uses a Builder pattern and can be started either by calling one of
+the Builders constructors or by calling the static method builder in the KafkaSpoutConfig class.
+
+The Constructor or static method to create the builder require a few key values (that can be changed later on) but are the minimum config needed to start
+a spout.
+
+`bootstrapServers` is the same as the Kafka Consumer Property "bootstrap.servers".
+`topics` The topics the spout will consume can either be a `Collection` of specific topic names (1 or more) or a regular expression `Pattern` and any
```
hmcl: ... expression, which specifies that any topics ...

docs/storm-kafka-client.md
```diff
@@ -1,90 +1,232 @@
...
+topics that match that regular expression will be consumed.
+
+In the case of the Constructors you may also need to specify a key deserializer and a value deserializer.  This is to help make the java generics happy
```
hmcl: This is to guarantee type safety through the use of Java generics

docs/storm-kafka-client.md
```diff
@@ -1,90 +1,232 @@
...
+
+`setFirstPollOffsetStrategy` allows you to set where to start consuming data from.  This is used both in case of failure recovery and starting the spout
+for the first time. Allowed values include
+
+ * `EARLIEST` means that the kafka spout polls records starting in the first offset of the partition, regardless of previous commits
+ * `LATEST` means that the kafka spout polls records with offsets greater than the last offset in the partition, regardless of previous commits
+ * `UNCOMMITTED_EARLIEST` (DEFAULT) means that the kafka spout polls records from the last committed offset, if any. If no offset has been committed, it behaves as `EARLIEST`.
+ * `UNCOMMITTED_LATEST` means that the kafka spout polls records from the last committed offset, if any. If no offset has been committed, it behaves as `LATEST`.
+
+`setRecordTranslator` allows you to modify how the spout converts a `ConsumerRecord` into a Tuple and which stream that tuple will go to.  By default the "topic",
```
hmcl: Kafka `Consumer Record` into a Tuple, and which stream that tuple will be published into.

docs/storm-kafka-client.md
```diff
@@ -1,90 +1,232 @@
...
+"partition", "offset", "key", and "value" will be emitted to the "default" stream.  If you want to output entries to different streams based on the topic storm
+provides `ByTopicRecordTranslator`.  See below for more examples on how to use these.
+
+`setProp` can be used to set kafka properties that do not have a convenience method.
```
hmcl: "Optional" -  should we rename this method to `setKafkaProp` or `setKafkaBrokerProp`  ?

docs/storm-kafka-client.md
```diff
@@ -1,90 +1,232 @@
...
+
+`setGroupId` lets you set the id of the kafka consumer group property "group.id'
+
+`setSSLKeystore` and `setSSLTruststore` allow you to configure SSL authentication.
+
+### Usage Examples
+
+#### Create a Simple Insecure Spout
+The following will consume all events published to "topic" and send them to MyBolt as "topic", "partition", "offset", "key", "value".
```
hmcl: ... MyBolt using the fields "topic" ...

docs/storm-kafka-client.md
```diff
@@ -1,90 +1,232 @@
...
+tp.setSpout("kafka_spout", new KafkaSpout<>(KafkaSpoutConfig.builder("127.0.0.1:" + port, "topic").build()), 1);
+tp.setBolt("bolt", new myBolt()).shuffleGrouping("kafka_spout");
+...
 
-String STREAM = "test_wildcard_stream";
-String TOPIC_WILDCARD_PATTERN = "test[1|2]";
+```
+
+#### Wildcard Topics
+Wildcard topics will consume from all topics that exist in the specified broakers list and match the pattern.  So in the following example
```
hmcl: typo - brokers

revans2:
The test failures are unrelated

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/bolt/KafkaBolt.java
```diff
@@ -0,0 +1,194 @@
...
+
+    private static final Logger LOG = LoggerFactory.getLogger(KafkaBolt.class);
+
+    public static final String TOPIC = "topic";
+
+    private KafkaProducer<K, V> producer;
+    private OutputCollector collector;
+    private TupleToKafkaMapper<K,V> mapper;
+    private KafkaTopicSelector topicSelector;
+    private Properties boltSpecfiedProperties = new Properties();
```
hmcl: nit: `boltSpecifiedProps` or `boltSpecificProps`

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/bolt/KafkaBolt.java
```diff
@@ -0,0 +1,196 @@
...
+    private boolean async = true;
+
+    public KafkaBolt() {}
+
+    public KafkaBolt<K,V> withTupleToKafkaMapper(TupleToKafkaMapper<K,V> mapper) {
+        this.mapper = mapper;
+        return this;
+    }
+
+    public KafkaBolt<K, V> withTopicSelector(String topic) {
```
hmcl: `withDefaultTopicSelector` - this would make clear to the API user the difference between the overloaded methods

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/bolt/KafkaBolt.java
```diff
@@ -0,0 +1,196 @@
...
+        return withTopicSelector(new DefaultTopicSelector(topic));
+    }
+    
+    public KafkaBolt<K,V> withTopicSelector(KafkaTopicSelector selector) {
+        this.topicSelector = selector;
+        return this;
+    }
+
+    public KafkaBolt<K,V> withProducerProperties(Properties producerProperties) {
```
hmcl: Is this method intended to set KafkaProducerProperties only? If so, wouldn't it be beneficial to call this method `withKafkaProducerProps`?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/bolt/KafkaBolt.java
```diff
@@ -0,0 +1,196 @@
...
+        if(mapper == null) {
+            this.mapper = new FieldNameBasedTupleToKafkaMapper<K,V>();
+        }
+
+        //for backward compatibility.
+        if(topicSelector == null) {
+            if(stormConf.containsKey(TOPIC)) {
+                this.topicSelector = new DefaultTopicSelector((String) stormConf.get(TOPIC));
+            } else {
+                throw new IllegalArgumentException("topic should be specified in bolt's configuration");
```
hmcl: `IllegalStateException` ?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/bolt/KafkaBolt.java
```diff
@@ -0,0 +1,196 @@
...
+    public KafkaBolt<K,V> withProducerProperties(Properties producerProperties) {
+        this.boltSpecfiedProperties = producerProperties;
+        return this;
+    }
+
+    @Override
+    public void prepare(Map stormConf, TopologyContext context, OutputCollector collector) {
+        //for backward compatibility.
+        if(mapper == null) {
+            this.mapper = new FieldNameBasedTupleToKafkaMapper<K,V>();
```
hmcl: perhaps log here
`LOG.info("Mapper not specified. Setting default mapper to {},  FieldNameBasedTupleToKafkaMapper.class.getSimpleName());`

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/bolt/KafkaBolt.java
```diff
@@ -0,0 +1,196 @@
...
+        }
+
+        //for backward compatibility.
+        if(topicSelector == null) {
+            if(stormConf.containsKey(TOPIC)) {
+                this.topicSelector = new DefaultTopicSelector((String) stormConf.get(TOPIC));
```
hmcl: perhaps log here
LOG.info("TopicSelector not specified. Using [{}] for topic [{}] specified in bolt configuration, DefaultTopicSelector.class.getSimpleName(), stormConf.get(TOPIC));

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/bolt/KafkaBolt.java
```diff
@@ -0,0 +1,196 @@
...
+        }
+
+        producer = mkProducer(boltSpecfiedProperties);
+        this.collector = collector;
+    }
+    
+    /**
+     * Intended to be overridden for tests.  Make the producer from props
+     */
+    protected KafkaProducer<K, V> mkProducer(Properties props) {
```
hmcl: `newProducer` ?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/bolt/KafkaBolt.java
```diff
@@ -0,0 +1,196 @@
...
+                throw new IllegalArgumentException("topic should be specified in bolt's configuration");
+            }
+        }
+
+        producer = mkProducer(boltSpecfiedProperties);
+        this.collector = collector;
+    }
+    
+    /**
+     * Intended to be overridden for tests.  Make the producer from props
```
hmcl: Create producer with props.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/bolt/KafkaBolt.java
```diff
@@ -0,0 +1,196 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger(KafkaBolt.class);
+
+    public static final String TOPIC = "topic";
+
+    private KafkaProducer<K, V> producer;
+    private OutputCollector collector;
+    private TupleToKafkaMapper<K,V> mapper;
+    private KafkaTopicSelector topicSelector;
+    private Properties boltSpecfiedProperties = new Properties();
+    private boolean fireAndForget = false;
```
hmcl: Can you please add a comment saying to look at the comments in the set method? I missed those comments at first.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/bolt/KafkaBolt.java
```diff
@@ -0,0 +1,196 @@
...
+    private boolean async = true;
```
hmcl: Can you please add a comment saying to look at the comments in the set method? I missed those comments at first.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/bolt/KafkaBolt.java
```diff
@@ -0,0 +1,196 @@
...
+
+    }
+
+    @Override
+    public void cleanup() {
+        producer.close();
+    }
+
+    /**
+     * If set to true the bold will assume that sending a message to kafka will succeed and will ack
```
hmcl: bolt

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/bolt/KafkaBolt.java
```diff
@@ -0,0 +1,196 @@
...
+     * the tuple as soon as it has handed the message off to the producer API
+     * if false (the default) the message will be acked after it we successfully sent to kafka or
```
hmcl: it was successfully

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/Subscription.java
```diff
@@ -0,0 +1,62 @@
...
+     * @param consumer the Consumer to get.
+     * @param listener the rebalance listener to include in the subscription
+     * @deprecated please use the version with the TopologyContext in it
+     */
+    public <K, V> void subscribe(KafkaConsumer<K, V> consumer, ConsumerRebalanceListener listener) {
+    	throw new IllegalStateException("At least one subscribe method must be overwritten");
+    }
+    
+    /**
+     * @return a string representing the subscribed topics.
```
srdo: Can this string be in any format? It seems to only be used in getComponentConfiguration, not really sure where that's being read.

external/storm-kafka/src/jvm/org/apache/storm/kafka/bolt/KafkaBolt.java
```diff
@@ -51,7 +51,9 @@
  * This bolt uses 0.8.2 Kafka Producer API.
  * <p/>
  * It works for sending tuples to older Kafka version (0.8.1).
+ * @deprecated Please use the KafkaBolt in storm-kafka-client
  */
+@Deprecated
```
srdo: Maybe it would be fine to just remove it here immediately and deprecate it in 1.1.0?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutConfig.java
```diff
@@ -61,129 +66,244 @@
...
+         */
+        public Builder<K,V> setMaxPartitionFectchBytes(int bytes) {
+            return setProp(ConsumerConfig.MAX_PARTITION_FETCH_BYTES_CONFIG, bytes);
+        }
+        
+        /**
+         * The maximum number of records a poll will return.
+         * Will only work with Kafka 0.10.0 and above.
+         */
+        public Builder<K,V> setMaxPoolRecords(int records) {
```
srdo: maxPool -> maxPoll

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/trident/KafkaTridentSpoutManager.java
```diff
@@ -44,29 +43,30 @@
...
-    private KafkaSpoutStreams kafkaSpoutStreams;                // Object that wraps all the logic to declare output fields and emit tuples
-    private KafkaSpoutTuplesBuilder<K, V> tuplesBuilder;        // Object that contains the logic to build tuples for each ConsumerRecord
+    private final Fields fields;
 
     public KafkaTridentSpoutManager(KafkaSpoutConfig<K, V> kafkaSpoutConfig) {
         this.kafkaSpoutConfig = kafkaSpoutConfig;
-        kafkaSpoutStreams = kafkaSpoutConfig.getKafkaSpoutStreams();
-        tuplesBuilder = kafkaSpoutConfig.getTuplesBuilder();
+        RecordTranslator<K, V> translator = kafkaSpoutConfig.getTranslator();
+        Fields fields = null;
```
hmcl: I would put all of this logic in a method called `setFields`. However, this is going to make this class hard to unit test. So we should even create a static factory method just wrapping this logic, and the constructor with two parameters, `KafkaTridentSpoutManager(KafkaSpoutConfig<K, V> kafkaSpoutConfig, Fields fields)`

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/trident/TridentKafkaState.java
```diff
@@ -0,0 +1,114 @@
...
+public class TridentKafkaState implements State {
+    private static final Logger LOG = LoggerFactory.getLogger(TridentKafkaState.class);
+
+    private KafkaProducer producer;
+    private OutputCollector collector;
+
+    private TridentTupleToKafkaMapper mapper;
+    private KafkaTopicSelector topicSelector;
+
+    public TridentKafkaState withTridentTupleToKafkaMapper(TridentTupleToKafkaMapper mapper) {
```
hmcl: Wouldn't it be preferable the class to be immutable ?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/trident/TridentKafkaState.java
```diff
@@ -0,0 +1,114 @@
...
+        LOG.debug("beginCommit is Noop.");
+    }
+
+    @Override
+    public void commit(Long txid) {
+        LOG.debug("commit is Noop.");
+    }
+
+    public void prepare(Properties options) {
+        if (mapper == null) throw new NullPointerException("mapper can not be null");
```
hmcl: Objects.requireNonNull("mapper cannot be null");
..

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/bolt/selector/KafkaTopicSelector.java
```diff
@@ -0,0 +1,26 @@
...
+ * limitations under the License.
+ */
+package org.apache.storm.kafka.bolt.selector;
+
+import org.apache.storm.tuple.Tuple;
+
+import java.io.Serializable;
+
+public interface KafkaTopicSelector extends Serializable {
+    String getTopic(Tuple tuple);
```
hmcl: since in some scenarios it's OK that this value returns null, wouldn't it be much clearer to return `Optional<String>` instead?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/trident/mapper/FieldNameBasedTupleToKafkaMapper.java
```diff
@@ -0,0 +1,41 @@
...
+    public final String msgFieldName;
+
+    public FieldNameBasedTupleToKafkaMapper(String keyFieldName, String msgFieldName) {
+        this.keyFieldName = keyFieldName;
+        this.msgFieldName = msgFieldName;
+    }
+
+    @Override
+    public K getKeyFromTuple(TridentTuple tuple) {
+        return (K) tuple.getValueByField(keyFieldName);
```
hmcl: ClassCastException?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/trident/mapper/FieldNameBasedTupleToKafkaMapper.java
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    @Override
+    public V getMessageFromTuple(TridentTuple tuple) {
```
hmcl: ClassCastException?

docs/storm-kafka-client.md
revans2: We are configuring the kafka consumer, so I would assume that it was implied.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/bolt/KafkaBolt.java
revans2: I don't want to rename public facing APIs right now, because this is a copy of what is in external/storm-kafka.  The other one is deprecated, but I want to maintain compatibility if possible.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/bolt/KafkaBolt.java
revans2: This is "moving" existing code so I want to maintain compatibility if possible.  But we are doing a copy + deprecation because the two libraries are compiled with different versions of Kafka so combining the two libraries in a single topology is difficult in some cases.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/bolt/KafkaBolt.java
revans2: backwards compatibility

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/bolt/KafkaBolt.java
revans2: backwards compatibility

revans2: No sorry wrong place for that comment.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -225,11 +213,7 @@ public void nextTuple() {
                 }
 
                 if (poll()) {
-                    try {
-                        setWaitingToEmit(pollKafkaBroker());
-                    } catch (RetriableException e) {
-                        LOG.error("Failed to poll from kafka.", e);
-                    }
+                    setWaitingToEmit(pollKafkaBroker());
```
revans2: That's right it was a merge error.  Great catch though.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutConfig.java
```diff
@@ -61,129 +66,244 @@
...
+        private Builder(Builder<?, ?> builder, SerializableDeserializer<K> keyDes, Class<? extends Deserializer<K>> keyDesClazz,
+        		SerializableDeserializer<V> valueDes, Class<? extends Deserializer<V>> valueDesClazz) {
+            this.kafkaProps = new HashMap<>(builder.kafkaProps);
+            this.subscription = builder.subscription;
+            this.pollTimeoutMs = builder.pollTimeoutMs;
+            this.offsetCommitPeriodMs = builder.offsetCommitPeriodMs;
+            this.maxRetries = builder.maxRetries;
+            this.firstPollOffsetStrategy = builder.firstPollOffsetStrategy;
+            this.maxUncommittedOffsets = builder.maxUncommittedOffsets;
+            //this could result in a lot of class case exceptions at runtime,
```
revans2: Sorry the comment was not updated.  I will do that.  Originally it was for API compatibility, but now it is to not force someone to reset the translator when they change a key or value type.  Specifically the default translator that should work on any types.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutConfig.java
```diff
@@ -61,129 +66,244 @@
...
+            this.valueDesClazz = valueDesClazz;
         }
 
         /**
          * Specifying this key deserializer overrides the property key.deserializer
          */
-        public Builder<K,V> setKeyDeserializer(SerializableDeserializer<K> keyDeserializer) {
-            this.keyDeserializer = keyDeserializer;
-            return this;
+        public <NK> Builder<NK,V> setKey(SerializableDeserializer<NK> keyDeserializer) {
```
revans2: Because we are changing the key type of the deserializer.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/Subscription.java
revans2: I didn't know either, but the only place I can see it used is in logging so that is what I did.

external/storm-kafka/src/jvm/org/apache/storm/kafka/bolt/KafkaBolt.java
revans2: I would rather do that in a follow on JIRA.  Most of the code is the same between the two so getting the rework from here to make there is a lot simpler if I can do a cherry pick

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/bolt/selector/FieldIndexTopicSelector.java
```diff
@@ -0,0 +1,52 @@
...
+
+import org.apache.storm.tuple.Tuple;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Uses field with a given index to select the topic name from a tuple .
+ */
+public class FieldIndexTopicSelector implements KafkaTopicSelector {
+    private static final long serialVersionUID = -3830575380208166367L;
```
revans2: Because I missed them.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/bolt/selector/FieldNameTopicSelector.java
```diff
@@ -0,0 +1,49 @@
...
+    private final String defaultTopicName;
+
+
+    public FieldNameTopicSelector(String fieldName, String defaultTopicName) {
+        this.fieldName = fieldName;
+        this.defaultTopicName = defaultTopicName;
+    }
+
+    @Override
+    public String getTopic(Tuple tuple) {
```
revans2: I agree but both for backwards compatibility and to limit the scope of this JIRA I would rather see it done in a follow on JIRA

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/ByTopicRecordTranslator.java
```diff
@@ -0,0 +1,93 @@
...
+        topicToTranslator.put(topic, translator);
+        cacheNCheckFields(translator);
+        return this;
+    }
+    
+    private void cacheNCheckFields(RecordTranslator<K, V> translator) {
+        for (String stream : translator.streams()) {
+            Fields fromTrans = translator.getFieldsFor(stream);
+            Fields cached = streamToFields.get(stream);
+            if (cached != null && !fromTrans.equals(cached)) {
```
revans2:
```
ByTopicRecordTranslator<A,B> trans = new ByTopicRecordTranslator((rec) -> Arrays.asList(rec.offset()), new Fields("offset"), "default");
trans.forTopic("specialTopic", (rec) -> Arrays.asList(rec.offset(), rec.message()), new Fields("offset", "message"), "default");
```
At this point we have tried to declare that the "default" stream has Fields ["offset"], and ["offset", "message"]  This is not supported by storm so we should not allow anyone to configure the spout to do this.

streamToFields is not yet updated for the new translator we are adding it yet.  We do it after we have verified that the Fields match for anything we have done already.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/ByTopicRecordTranslator.java
```diff
@@ -0,0 +1,93 @@
...
+                throw new IllegalArgumentException("Stream " + stream + " currently has Fields of " + cached + " which is not the same as those being added in " + fromTrans);
```
revans2: In this case it is the argument being passed in that is bad and we are rejecting it.  They could come back and switch it to a new stream which would be fine.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/Subscription.java
```diff
@@ -0,0 +1,62 @@
...
+    	subscribe(consumer, listener);
+    }
+
+    /**
+     * Subscribe the KafkaConsumer to the proper topics
+     * @param consumer the Consumer to get.
+     * @param listener the rebalance listener to include in the subscription
+     * @deprecated please use the version with the TopologyContext in it
+     */
+    public <K, V> void subscribe(KafkaConsumer<K, V> consumer, ConsumerRebalanceListener listener) {
```
revans2: I didn't feel good about deprecating it in the first place and will remove it.  It just means that I will also pull in the changes I made for STORM-2236 back to 1.1 too.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -325,15 +310,19 @@ private boolean emitTupleIfNotEmitted(ConsumerRecord<K, V> record) {
         } else {
             boolean isScheduled = retryService.isScheduled(msgId);
             if (!isScheduled || retryService.isReady(msgId)) {   // not scheduled <=> never failed (i.e. never emitted) or ready to be retried
-                final List<Object> tuple = tuplesBuilder.buildTuple(record);
-                kafkaSpoutStreams.emit(collector, tuple, msgId);
+                final List<Object> tuple = kafkaSpoutConfig.getTranslator().apply(record);
+                if (tuple instanceof KafkaTuple) {
```
revans2: The javadocs for RecordTranslator state.
```
     * @return the objects in the tuple.  Return a {@link KafkaTuple}
     * if you want to route the tuple to a non-default stream
```

All of the provided implementations support this.  Both Simple and ByTopic, by way of the SimpleRecordTranslator.

I will add a section in the docs to talk about this, but it only really matters if you are writing your own record translator from scratch, instead of using the built in ones that should cover the vast majority of the use cases.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/TopicPartitionComparator.java
```diff
@@ -1,16 +1,35 @@
...
 public class TopicPartitionComparator implements Comparator<TopicPartition> {
-    @Override
-    public int compare(TopicPartition o1, TopicPartition o2) {
-        if (!o1.topic().equals(o2.topic())) {
-            return o1.topic().compareTo(o2.topic());
-        } else {
-            return o1.partition() - o2.partition();
-        }
-    }
+	public static final TopicPartitionComparator INSTANCE = new TopicPartitionComparator();
```
srdo: Enums for things that are not an enumeration feels a little weird. Wouldn't you get the same result by just adding a private no-args constructor?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/RoundRobinManualPartitioner.java
```diff
@@ -15,14 +15,26 @@
...
+public class RoundRobinManualPartitioner implements ManualPartitioner {
 
-public interface KafkaPartitionReader {
-    List<TopicPartition> readPartitions(KafkaConsumer<?, ?> consumer);
+	@Override
+	public List<TopicPartition> partition(List<TopicPartition> allPartitions, TopologyContext context) {
+		int thisTaskIndex = context.getThisTaskIndex();
+		int totalTaskCount = context.getComponentTasks(context.getThisComponentId()).size();
+		Set<TopicPartition> myPartitions = new HashSet<>(allPartitions.size()/totalTaskCount+1);
+		for (int i = thisTaskIndex; i < allPartitions.size(); i += totalTaskCount) {
```
srdo: I don't think so. Either i is in [0, allPartitions.size()-1] or the loop is over.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutConfig.java
srdo: I think what I was asking was "Why would you need to create a Builder with key type A and then later set a deserializer for keys of type B"? :)

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
revans2: Yes, and it is documented in RecordTranslator.  I will add in more documentation on it though.

The reason specifically for this was because the spout is not able to keep track of a single message being emitted to multiple streams.  It would get confused and ack it before it was truly done.  This makes it impossible for that to happen.  What is more the built in record translators should cover 99% of the use cases, so the fact that it is not super well documented should be more of a corner case.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/TopicPartitionComparator.java
revans2: I'm not sure we need to guarantee it.  The old code had the INSTANCE where it was used, I moved it here in hopes that others might use it.  I could make the constructor private if we really want it to be a singleton, but I don't think it is a requirement.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/NamedSubscription.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    @Override
+    public <K, V> void subscribe(KafkaConsumer<K, V> consumer, ConsumerRebalanceListener listener, TopologyContext unused) {
+        consumer.subscribe(topics, listener);
+        LOG.info("Kafka consumer subscribed topics {}", topics);
+    }
+
+    @Override
+    public String getTopicsString() {
```
revans2: It is in the parent javadocs.  That is around the override

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaTuple.java
```diff
@@ -15,22 +15,33 @@
...
+    
+    public KafkaTuple() {
+        super();
+    }
+    
+    public KafkaTuple(Object... vals) {
+        super(vals);
+    }
+    
+    public KafkaTuple routedTo(String stream) {
```
revans2: This is because the constructor is varadic following the Values parent class. It is ambiguous to have

```
    public KafkaTuple(Object... vals)
```

and any other constructor.  Java complains.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/RoundRobinManualPartitioner.java
revans2: No.  `i < allPartitions.size()` guarantees that we will never call get on allPartitions with an index that is out of bounds.  The Set is just setting the initial size to avoid more memory allocation.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/trident/KafkaTridentSpoutManager.java
revans2: I'm a bit confused about what benefit that would give.  The Fields come from the SpoutConfig, by way of the RecordTranslator.  Why would you want to override the Fields but not the RecordTranslator that must conform to those fields?  I just don't see value it separating two things that are highly coupled together.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/trident/TridentKafkaState.java
revans2: This is code that was "moved" like with the KafkaBolt for storm-kafka to storm-kafka-client.  If we really want to make it immutable we can, but I think that is beyond the scope of this JIRA

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/trident/TridentKafkaState.java
revans2: Objects does not exist in java 6 and I would prefer to keep the code compatible as mush as possible to avoid extra rework when pulling these changes back.  If you insist I will do it.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/bolt/selector/KafkaTopicSelector.java
revans2: File another JIRA and we can look into it.  I think it is beyond the scope of this one.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/trident/mapper/FieldNameBasedTupleToKafkaMapper.java
revans2: It is possible but that is the point of the generics.  To try and reduce the likelihood of it happening.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/trident/mapper/FieldNameBasedTupleToKafkaMapper.java
revans2: It is possible but that is the point of the generics. To try and reduce the likelihood of it happening.

revans2:
@hmcl I think I addressed all of your review comments.  I also re-based to deal with the merge conflict. If everything looks good I will squash some of the commits.  There are a lot.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/trident/TridentKafkaState.java
srdo: I thought 1.x was targeting at least Java 7?

revans2:
The travis test failures are unrelated.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/trident/TridentKafkaState.java
revans2: @srdo you are right.

```
                <artifactId>maven-compiler-plugin</artifactId>
                <configuration>
                    <source>1.7</source>
                    <target>1.7</target>
                </configuration>
```
I'll update it.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/bolt/selector/KafkaTopicSelector.java
HeartSaVioR: IMHO we need to try avoiding JDK 8 feature for public API unless we decided to release minor version only for 2.0.0. But anyway I also think this is beyond of the scope.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/TopicPartitionComparator.java
hmcl: @srdo that is not true. See Effective Java Item 3. Yes, private constructor would work.

@revans2 either way is fine. My comment was motivated because I believe that either we should make it a singleton or not. If not, let the users simple instantiate it with new and that's it. I am a believer in consistency.

revans2: @hmcl I agree I went half way to making it a true singleton and I should finish the job.

external/storm-kafka/src/jvm/org/apache/storm/kafka/bolt/KafkaBolt.java
revans2: I filed https://issues.apache.org/jira/browse/STORM-2317 to do this and I assigned it to myself.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/bolt/selector/FieldNameTopicSelector.java
revans2: I filed https://issues.apache.org/jira/browse/STORM-2318 for this

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -447,7 +440,10 @@ private void shutdown() {
 
     @Override
     public void declareOutputFields(OutputFieldsDeclarer declarer) {
-        kafkaSpoutStreams.declareOutputFields(declarer);
+        RecordTranslator<K, V> translator = kafkaSpoutConfig.getTranslator();
```
revans2: I agree I think we need a fairly standard way to Tuple->X and X->Tuple.  But really it would just end up looking like

```java
public interface <X> TupleToX {
    public X fromTuple(Tuple t);
}

public interface X XToTuple {
    public Tuple toTuple(X record);
}
```

But I don't know how much code reuse there could be, because X is so different for each library we are going to interface with.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/TopicPartitionComparator.java
srdo: @hmcl Neat, hadn't considered those arguments for using enum.

srdo:
There are still a few files with tabs mixed in to spaces, so autoformatting all the files might be good. I also still don't really understand why the KafkaSpoutConfig builder needs to support setting a deserializer with a different generic type to the one the builder was instantiated with.

+1 though.

ptgoetz:
+1

hmcl:
+1

My only suggestion at this point is that we file an umbrella JIRA covering some of the follow up comments that haven't been addressed, such as minor refactoring, improved (java) docs, immutability, etc. Basically to just capture the most meaningful comments that makes sense handle later.

Thanks @revans2 for all your patience addressing the many review comments

revans2:
Everyone,

Sorry this took so long.  I kept getting pulled off on critical issues.  I have rebased/squashed everything (except the code revert) and also fixed the issue with tabs vs spaces.  Since it was just white space issues I don't think I need more +1s but if someone wants to take a look again please feel free to.  I will do the same for the 1.x version, and then file the umbrella jira that @hmcl requested.

hmcl:
Thanks @revans2. Can you please just amend the commit log [messages](https://github.com/apache/storm/pull/1808/commits) as I believe they do not really reflect the changes in this patch. One of the messages has STORM-1997 title, and the other has a cryptic merge log message.

Also, can you please do the same for 1.x-branch [patch](https://github.com/apache/storm/pull/1868)

revans2:
@hmcl github like to truncate long messages.

```
$ git log STORM-2225
commit c9f9348e3de0917020a8c903f4c9d026b7dc6204
Author: Robert (Bobby) Evans <evans@yahoo-inc.com>
Date:   Tue Nov 29 21:39:26 2016 -0600

    STORM-1997: copy state/bolt from storm-kafka to storm-kafka-client
    STORM-2225: change spout config to be simpler.
    STORM-2228: removed ability to request a single topic go to multiple streams
    STORM-2236: Reimplemented manual partition management on top of STORM-2225

commit 74a77aa6494957b941df55d66ff64b6561eee4a8
Author: Robert (Bobby) Evans <evans@yahoo-inc.com>
Date:   Wed Jan 18 10:34:30 2017 -0600

    Revert "Merge branch 'STORM-2236' of https://github.com/MediaV/storm into STORM-2236"

    This reverts commit e2f87c304297ce95f01708e0934c44ee385fa0bb, reversing
    changes made to 00fed518d14b485e2a9e9cdd91ade3cc9d68dc8d.

```

If you really want me to squash the code change with the revert and only reference a single JIRA in the commit messages I can.  But I thought it would be better if I am addressing 4 separate (but very closely related) JIRA I should keep them all together.

hmcl:
@revans2 this is perfect. These 4 commits make it much easier to track down. I confused the way GitHub truncates the messages - my bad.

+1

HeartSaVioR:
Thanks for the great work and patience. +1

PR #1813, opened 2016-12-03
STORM-2234: heartBeatExecutorService in shellSpout don't work well with deactivate

itielshwartz:


HeartSaVioR:
+1 Thanks for finding missed spot.

PR #1819, opened 2016-12-09
[STORM-2238] Add Timestamp extractor for windowed bolt

arunmahadevan:
Added a timestamp extractor so that the tuple timestamp can be derived from the tuple than directly
mapping it to a field within the tuple.


HeartSaVioR:
+1

PR #1820, opened 2016-12-09
[STORM-2238] Add Timestamp extractor for windowed bolt

arunmahadevan:
Added a timestamp extractor so that the tuple timestamp can be derived from the tuple than directly
mapping it to a field within the tuple.

https://github.com/apache/storm/pull/1819 applied to 1.x-branch


harshach:
+1

hmcl:
+1

HeartSaVioR:
@arunmahadevan 
Didn't we have document around setting up timestamp field? I think we should have it, or if we did, it should be updated.
Other than that it looks great overall.

arunmahadevan:
@HeartSaVioR good point. Old api is retained, added additional doc for new api in docs/Windowing.md

HeartSaVioR:
@arunmahadevan Thanks for the quick update. +1

PR #1825, opened 2016-12-14
STORM-2236 storm kafka client support manual partition management.

liurenjie1024:
1. Support manual partition assignment, since kafka managed partition assignment may cause unnecessary rebalance if you don't keep polling.
2. A little change to fix STOMR-2077 which may lose failed messages.

### DIFF
--- a/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
+++ b/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
@@ -215,18 +215,22 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
     @Override
     public void nextTuple() {
         if (initialized) {
-            refreshPartitionIfNeeded();
+            try {
+                refreshPartitionIfNeeded();
 
-            if (commit()) {
-                commitOffsetsForAckedTuples();
-            }
+                if (commit()) {
+                    commitOffsetsForAckedTuples();
+                }
 
-            if (poll()) {
-                setWaitingToEmit(pollKafkaBroker());
-            }
+                if (poll()) {
+                    setWaitingToEmit(pollKafkaBroker());
+                }
 
-            if (waitingToEmit()) {
-                emit();
+                if (waitingToEmit()) {
+                    emit();
+                }
+            } catch (Exception e) {
+                LOG.error("Failed to emit tuples.", e);
             }
         } else {
             LOG.debug("Spout not initialized. Not sending tuples until initialization completes");
@@ -308,11 +312,12 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
         final Set<TopicPartition> retriableTopicPartitions = retryService.retriableTopicPartitions();
 
         for (TopicPartition rtp : retriableTopicPartitions) {
-            final OffsetAndMetadata offsetAndMeta = acked.get(rtp).findNextCommitOffset();
-            if (offsetAndMeta != null) {
-                kafkaConsumer.seek(rtp, offsetAndMeta.offset() + 1);  // seek to the next offset that is ready to commit in next commit cycle
-            } else {
-                kafkaConsumer.seekToEnd(toArrayList(rtp));    // Seek to last committed offset
+            KafkaSpout.OffsetEntry entry = acked.get(rtp);
+            if (entry != null) {
+                final OffsetAndMetadata offsetAndMeta = entry.findNextCommitOffset();
+                if (offsetAndMeta != null) {
+                    kafkaConsumer.seek(rtp, offsetAndMeta.offset() + 1);  // seek to the next offset that is ready to commit in next commit cycle
+                }
             }
         }
     }
--- a/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
+++ b/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
@@ -23,6 +23,7 @@ import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.apache.kafka.clients.consumer.ConsumerRecords;
 import org.apache.kafka.clients.consumer.KafkaConsumer;
 import org.apache.kafka.clients.consumer.OffsetAndMetadata;
+import org.apache.kafka.common.PartitionInfo;
 import org.apache.kafka.common.TopicPartition;
 import org.apache.storm.kafka.spout.KafkaSpoutConfig.FirstPollOffsetStrategy;
 import org.apache.storm.spout.SpoutOutputCollector;
@@ -59,6 +60,8 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
 
     // Storm
     protected SpoutOutputCollector collector;
+    protected int thisTaskIndex;
+    protected int taskCount;
 
     // Kafka
     private final KafkaSpoutConfig<K, V> kafkaSpoutConfig;
@@ -66,11 +69,15 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
     private transient boolean consumerAutoCommitMode;
 
 
+
     // Bookkeeping
     private transient int maxRetries;                                   // Max number of times a tuple is retried
     private transient FirstPollOffsetStrategy firstPollOffsetStrategy;  // Strategy to determine the fetch offset of the first realized by the spout upon activation
     private transient KafkaSpoutRetryService retryService;              // Class that has the logic to handle tuple failure
     private transient Timer commitTimer;                                // timer == null for auto commit mode
+    private transient Timer partitionRefreshTimer;                      // partitionRefreshTime != null if in manual partition assign model
+    private transient boolean manualPartitionAssignment;
+    private transient KafkaSpoutConsumerRebalanceListener partitionRebalanceListener;
     private transient boolean initialized;                              // Flag indicating that the spout is still undergoing initialization process.
     // Initialization is only complete after the first call to  KafkaSpoutConsumerRebalanceListener.onPartitionsAssigned()
 
@@ -94,12 +101,16 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
 
         // Spout internals
         this.collector = collector;
+        thisTaskIndex = context.getThisTaskIndex();
+        taskCount = context.getComponentTasks(context.getThisComponentId()).size();
         maxRetries = kafkaSpoutConfig.getMaxTupleRetries();
         numUncommittedOffsets = 0;
 
         // Offset management
         firstPollOffsetStrategy = kafkaSpoutConfig.getFirstPollOffsetStrategy();
         consumerAutoCommitMode = kafkaSpoutConfig.isConsumerAutoCommitMode();
+        manualPartitionAssignment = kafkaSpoutConfig.isManualPartitionAssign();
+        partitionRebalanceListener = new KafkaSpoutConsumerRebalanceListener();
 
         // Retries management
         retryService = kafkaSpoutConfig.getRetryService();
@@ -109,6 +120,10 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
 
         if (!consumerAutoCommitMode) {     // If it is auto commit, no need to commit offsets manually
             commitTimer = new Timer(500, kafkaSpoutConfig.getOffsetsCommitPeriodMs(), TimeUnit.MILLISECONDS);
+        }
+
+        if (manualPartitionAssignment) {
+            partitionRefreshTimer = new Timer(500, kafkaSpoutConfig.getPartitionRefreshPeriodMs(), TimeUnit.MILLISECONDS);
         }
 
         acked = new HashMap<>();
@@ -200,6 +215,8 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
     @Override
     public void nextTuple() {
         if (initialized) {
+            refreshPartitionIfNeeded();
+
             if (commit()) {
                 commitOffsetsForAckedTuples();
             }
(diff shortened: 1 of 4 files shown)


srdo:
Hi. STORM-2077 is being resolved here https://issues.apache.org/jira/browse/STORM-2087. Sorry for the inconvenience.

I'm not sure I understand why adding manual partition management is necessary. Kafka has recently added heartbeating from a background thread, so it's no longer necessary to poll often in order to avoid rebalances (see http://kafka.apache.org/documentation.html#upgrade_1010_notable, search for "heartbeat"). That change decouples necessary poll frequency from the time it would take to detect a crashed consumer, so the `max.poll.interval.ms` can be bumped if a rebalance every 5 minutes is unacceptable (and then only if poll isn't called in that interval).

Could you describe a use case?

liurenjie1024:
I think the behavior of consumer's automatic partition management is complicated and may lead to unpredictable behavior, which is quite important for many purposes like debugging, performance tuning, etc, especially in critical systems like billing system in our case. In fact, even the author of kafka consumer suggests us to use manual partition management when we use streaming process frameworks (please refer to the Manual Partition Assignment part of  [kafka consumer doc (http://kafka.apache.org/0101/javadoc/index.html?org/apache/kafka/clients/consumer/KafkaConsumer.html). Other streaming process frameworks like flink, apex, spark streaming, old storm kafka client also uses manual partition management.

hmcl:
@liurenjie1024 putting aside for a second the discussion about the need to implement partition assignment, if we indeed agree that we should support it, I think that the `KafkaSpout` class is getting a bit too bloated to handle some of these separate requirements. In a sense this is a new feature on its own. I would favor using a more OOD approach if possible. Perhaps subclass or compose the `KafkaSpout` @srdo @revans2 what's your opinion?

liurenjie1024:
Maybe we do not need to maintain two versions of `KafkaSpout` with same functionality. The main goal of automatic partition management is to ease fault tolerant in simple applications, not in frameworks like storm. Automatic partition management may hide extra overhead and unpredictable behavior which is unnecessary in frameworks like storm.

srdo:
@liurenjie1024 Okay, it makes sense that having Kafka manage partitions isn't really necessary since Storm will make sure to keep spout instances running. I can't speak to how much of an overhead and unpredictability there really is to making Kafka manage it though, but it's fine by me if we want to switch to manual management. I can't think of a good reason for supporting both manual and automatic assignment though. 

The reason the old Kafka spout did manual assignment was that automatic assignment wasn't available on the old APIs as far as I know.

I agree that we shouldn't have two variants of the spout, it just adds unnecessary complexity. I'd rather we just do a solid manual assignment implementation and get rid of the automatic code, if we need to switch.

Keep in mind that to be on par with automatic assignment manual assignment needs to support adding partitions without requiring a spout restart, and without the spout instances stepping on each others' toes (e.g. avoid partitions being assigned to two spout instances at once, even while reassigning).

@hmcl Sure, it's more readable if strategies like this can be sectioned off in their own classes. If we really have to support both, I think it's nicer if assignment can be implemented through delegates rather than through subclassing.

liurenjie1024:
@srdo I'll submit a patch for that later.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -262,11 +312,12 @@ private void doSeekRetriableTopicPartitions() {
         final Set<TopicPartition> retriableTopicPartitions = retryService.retriableTopicPartitions();
 
         for (TopicPartition rtp : retriableTopicPartitions) {
-            final OffsetAndMetadata offsetAndMeta = acked.get(rtp).findNextCommitOffset();
-            if (offsetAndMeta != null) {
-                kafkaConsumer.seek(rtp, offsetAndMeta.offset() + 1);  // seek to the next offset that is ready to commit in next commit cycle
-            } else {
-                kafkaConsumer.seekToEnd(toArrayList(rtp));    // Seek to last committed offset
+            KafkaSpout.OffsetEntry entry = acked.get(rtp);
+            if (entry != null) {
```
srdo: I'd like it if you could revert the commit for fixing STORM-2077, since a different fix is being ported back from master.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -360,18 +411,22 @@ private void subscribeKafkaConsumer() {
...
-        if (kafkaSpoutStreams instanceof KafkaSpoutStreamsNamedTopics) {
-            final List<String> topics = ((KafkaSpoutStreamsNamedTopics) kafkaSpoutStreams).getTopics();
-            kafkaConsumer.subscribe(topics, new KafkaSpoutConsumerRebalanceListener());
-            LOG.info("Kafka consumer subscribed topics {}", topics);
-        } else if (kafkaSpoutStreams instanceof KafkaSpoutStreamsWildcardTopics) {
-            final Pattern pattern = ((KafkaSpoutStreamsWildcardTopics) kafkaSpoutStreams).getTopicWildcardPattern();
-            kafkaConsumer.subscribe(pattern, new KafkaSpoutConsumerRebalanceListener());
-            LOG.info("Kafka consumer subscribed topics matching wildcard pattern [{}]", pattern);
+        if (manualPartitionAssignment) {
+            doRefreshPartitions();
```
srdo: Nit, but it would probably be good to log which topic is subscribed as well here

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutConfig.java
```diff
@@ -230,8 +237,37 @@ public Builder(Map<String, Object> kafkaProps, KafkaSpoutStreams kafkaSpoutStrea
         }
 
         public KafkaSpoutConfig<K,V> build() {
+            validate();
             return new KafkaSpoutConfig<>(this);
         }
+
+        /**
+         * Defines whether the consumer manages partition manually.
+         * If set to true, the consumer behaves like a simple consumer, otherwise it will rely on kafka to do partition assignment.
```
srdo: Consider changing this to "If set to true, the spout instances are assigned partitions round-robin", since readers maybe don't know what it means to behave like a simple consumer.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutConfig.java
```diff
@@ -230,8 +237,37 @@ public Builder(Map<String, Object> kafkaProps, KafkaSpoutStreams kafkaSpoutStrea
...
+         * @param manualPartitionAssign Whether use manual partition assignment.
```
srdo: Nit: "True if using manual partition assignment" is clearer IMO

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutConfig.java
```diff
@@ -230,8 +237,37 @@ public Builder(Map<String, Object> kafkaProps, KafkaSpoutStreams kafkaSpoutStrea
...
+        public Builder setPartitionRefreshPeriodMs(int partitionRefreshPeriodMs) {
+            this.partitionRefreshPeriodMs = partitionRefreshPeriodMs;
+            return this;
+        }
+
+        /**
+         * Validate configs before build.
+         */
+        private void validate() {
+            if (this.manualPartitionAssign && kafkaSpoutStreams instanceof KafkaSpoutStreamsWildcardTopics) {
```
srdo: Probably better to check that kafkaSpoutStreams is an instance of KafkaSpoutStreamsNamedTopics, so this doesn't break if more subclasses are added at some point.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutConfig.java
```diff
@@ -230,8 +237,37 @@ public Builder(Map<String, Object> kafkaProps, KafkaSpoutStreams kafkaSpoutStrea
...
+                throw new IllegalArgumentException("Manual partition assign can't be used with wildcard topics!");
```
srdo: assign -> assignment

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/TopicPartitionComparator.java
```diff
@@ -0,0 +1,21 @@
+package org.apache.storm.kafka.spout;
+
+import org.apache.kafka.common.TopicPartition;
+
+import java.util.Comparator;
+
+/**
+ * Created by liurenjie on 12/7/16.
```
srdo: Please remove the created by comment from the files

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/TopicPartitionComparator.java
```diff
@@ -0,0 +1,21 @@
...
+ */
+public enum TopicPartitionComparator implements Comparator<TopicPartition> {
```
srdo: Nit: I don't think there's anything wrong with making this an enum, but it seems conceptually weird to me. Why not just make this a regular class, and then make the field for it in the spout static?

srdo:
Btw @liurenjie1024 you need to make a PR for this against master first. We can't as far as I know merge features to 1.x before they are also on master.

srdo:
Also I was wrong about spout instances interfering during reassignment. Since the task count is constant for the lifetime of the topology, there's no issue since a partition will always be assigned to the same task. My bad :)

liurenjie1024:
Hi, @srdo I've submitted a new patch for this issue with some refactor against master branch. I think most of your comments are great and have fixed them in the new patch. Please help to review [this PR](https://github.com/apache/storm/pull/1835) and this PR will be closed.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -59,18 +60,24 @@
 
     // Storm
     protected SpoutOutputCollector collector;
+    protected int thisTaskIndex;
```
liurenjie1024: Fixed.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -200,16 +215,22 @@ private void setAcked(TopicPartition tp, long fetchOffset) {
...
-            if (waitingToEmit()) {
-                emit();
+                if (poll()) {
+                    setWaitingToEmit(pollKafkaBroker());
+                }
+
+                if (waitingToEmit()) {
+                    emit();
+                }
+            } catch (Exception e) {
```
liurenjie1024: It's not uncommon to encounter an exception when kafka consumer poll data from broker and I don't think we should restart it. But I should not catch the whole function. This has been resolved in the new patch.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -236,6 +257,35 @@ private boolean poll() {
         return poll;
     }
 
+    private void refreshPartitionIfNeeded() {
+        if (!manualPartitionAssignment || !partitionRefreshTimer.isExpiredResetOnTrue()) return;
```
liurenjie1024: Fixed in the new PR.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -236,6 +257,35 @@ private boolean poll() {
...
+        doRefreshPartitions();
+    }
+
+    private void doRefreshPartitions() {
+        KafkaSpoutStreamsNamedTopics streams = KafkaSpoutStreamsNamedTopics.class.cast(kafkaSpoutStreams);
```
liurenjie1024: 1. Wildcard topics support are added.
2. class.cast is preferred because it throws an exception which better explains the failure rather than NPE.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
liurenjie1024: Fixed in the new PR.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutConfig.java
liurenjie1024: Fixed in the new PR.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutConfig.java
liurenjie1024: Fixed in the new PR.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutConfig.java
liurenjie1024: Fixed in the new PR.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/TopicPartitionComparator.java
liurenjie1024: Fixed in the new PR.

PR #1827, opened 2016-12-14
STORM-2243: adds ip address to supervisor id

abellina:
The supervisor id is a simple UUID currently. If there are issues with the heartbeat messages from supervisor, looking at Zookeeper makes it hard to find which node is having issues. Adding the ip address to the supervisor id makes debugging these cases much simpler.


revans2:
+1

HeartSaVioR:
+1

PR #1829, opened 2016-12-16
STORM-2246: Logviewer download link has urlencoding on part of the URL

arunmahadevan:
Changed the URL format for "Download full file" link under worker logs
from "download/URLENCODED(<filename>)" to "download?file=URLENCODED(<filename>)"


HeartSaVioR:
+1

PR #1831, opened 2016-12-17
STORM-2249: Make Distribution Scripts Put Examples to the Correct Locations

hmcl:
- Move all examples under  STORM_HOME/examples/storm-COMPONENT-NAME-examples/COMPONENT-NAME-examples-x.y.x.jar

### DIFF
--- a/storm-dist/binary/final-package/src/main/assembly/binary.xml
+++ b/storm-dist/binary/final-package/src/main/assembly/binary.xml
@@ -114,6 +114,13 @@
             </includes>
         </fileSet>
         <fileSet>
+            <directory>${project.basedir}/../../../examples/storm-kafka-examples/target</directory>
+            <outputDirectory>examples/storm-kafka-examples</outputDirectory>
+            <includes>
+                <include>storm*jar</include>
+            </includes>
+        </fileSet>
+        <fileSet>
             <directory>${project.basedir}/../../../external/storm-kinesis/target</directory>
             <outputDirectory>external/storm-kinesis</outputDirectory>
             <includes>
@@ -142,6 +149,13 @@
             </includes>
         </fileSet>
         <fileSet>
+            <directory>${project.basedir}/../../../examples/storm-hdfs-examples/target</directory>
+            <outputDirectory>examples/storm-hdfs-examples</outputDirectory>
+            <includes>
+                <include>storm*jar</include>
+            </includes>
+        </fileSet>
+        <fileSet>
             <directory>${project.basedir}/../../../external/storm-hbase/target</directory>
             <outputDirectory>external/storm-hbase</outputDirectory>
             <includes>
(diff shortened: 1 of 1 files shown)


hmcl:
Whit these changes the examples distribution looks like this:

screen shot 2016-12-17 at 2 17 26 am

HeartSaVioR:
@hmcl 
The change looks great, but we need to think why shaded jar of examples are necessary to be included to binary distribution. 
AFAIK the examples modules (except storm-starter) are for referring the working example source code rather than executing the topology. Do we expect end-user would run the example topology? If we expect, we should document about topologies to README.md for each modules.

harshach:
@HeartSaVioR I think it makes sense to have the shaded jars. It will be easier to run the code for the users who not familiar with storm packaging and gives insight into building a topology jar as well.

HeartSaVioR:
@harshach @hmcl 
Then we should add available topologies in README for each modules to since users don't know which topologies are available.

hmcl:
@HeartSaVioR do you mean each README of each module should have a section on how to run the JAR, and the main for that JAR? I agree with that. Perhaps I could file a follow-up JIRA for that.

HeartSaVioR:
@hmcl 
Right. But I changed my mind that all examples should contain the code, like storm-starter in binary dist. We're extracting examples to show example **codes** for each modules, so the most important thing is code.
@harshach What do you think?

HeartSaVioR:
Also could you check how big binary dist (tar.gz / zip) will be? It's already over 200MB so IMHO it would be better to reduce/restrict the size of binary dist.

HeartSaVioR:
Before:
-rw-r--r--   1 jlim  staff   3.3K  3  2 19:41 apache-storm-2.0.0-SNAPSHOT.pom
-rw-r--r--   1 jlim  staff   473B  3  2 19:41 apache-storm-2.0.0-SNAPSHOT.pom.asc
-rw-r--r--   1 jlim  staff   264M  3  2 19:41 apache-storm-2.0.0-SNAPSHOT.tar.gz
-rw-r--r--   1 jlim  staff   473B  3  2 19:41 apache-storm-2.0.0-SNAPSHOT.tar.gz.asc
-rw-r--r--   1 jlim  staff   264M  3  2 19:41 apache-storm-2.0.0-SNAPSHOT.zip
-rw-r--r--   1 jlim  staff   473B  3  2 19:41 apache-storm-2.0.0-SNAPSHOT.zip.asc
----------
After:
-rw-r--r--   1 jlim  staff   3.3K  3  2 19:46 apache-storm-2.0.0-SNAPSHOT.pom
-rw-r--r--   1 jlim  staff   473B  3  2 19:46 apache-storm-2.0.0-SNAPSHOT.pom.asc
-rw-r--r--   1 jlim  staff   564M  3  2 19:46 apache-storm-2.0.0-SNAPSHOT.tar.gz
-rw-r--r--   1 jlim  staff   473B  3  2 19:46 apache-storm-2.0.0-SNAPSHOT.tar.gz.asc
-rw-r--r--   1 jlim  staff   565M  3  2 19:46 apache-storm-2.0.0-SNAPSHOT.zip
-rw-r--r--   1 jlim  staff   473B  3  2 19:47 apache-storm-2.0.0-SNAPSHOT.zip.asc

Sorry but for me the final size (exceeding 500M) is hard to accept. Could we re-think about this?

HeartSaVioR:
@hmcl 
We applied opposite approach for this. Could you close this PR?

PR #1832, opened 2016-12-17
STORM-2250: Kafka Spout Refactoring to Increase Modularity and Testability

srdo:
Making these changes based on a discussion here https://github.com/apache/storm/pull/1826 and https://github.com/apache/storm/pull/1825 about the spout getting a bit large.

This PR makes the following changes:
* OffsetEntry was moved into a new class and renamed OffsetManager
* OffsetManager.commit now returns numCommittedOffsets, so it doesn't have to refer to a KafkaSpout internal variable.
* KafkaSpout.commitOffsetsForAckedTuples no longer iterates over acked twice. The second iteration was used to commit offsets to OffsetEntries, but we may as well use nextCommitOffsets for that iteration, since those were the offsets committed to Kafka. This also saves us a null check in OffsetManager.commit.
* OffsetManager.commit no longer checks if the parameter is null. It should no longer be possible that it is null, unless there's a bug, in which case we want to throw an NPE so we can fix it.
* Timer supports Storm's time simulation. This also means that it only supports time units down to milliseconds. If we need nanosecond precision, we need to update Storm's Utils.Time. 
* Cleaned up a few redundant version declarations in the pom. Also switched out hamcrest-all for hamcrest-core + hamcrest-library, since tests were throwing NoSuchMethodError when an assertThat failed.
* Updated the tests to use time simulation, and to check if the spout calls KafkaConsumer.commitSync instead of reading the spout's internal state

I think it would be nice if RetryService also supported time simulation, let me know what you think @hmcl

### DIFF
--- a/external/storm-kafka-client/pom.xml
+++ b/external/storm-kafka-client/pom.xml
@@ -77,7 +77,13 @@
         </dependency>
         <dependency>
             <groupId>org.hamcrest</groupId>
-            <artifactId>hamcrest-all</artifactId>
+            <artifactId>hamcrest-core</artifactId>
+            <version>1.3</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.hamcrest</groupId>
+            <artifactId>hamcrest-library</artifactId>
             <version>1.3</version>
             <scope>test</scope>
         </dependency>
@@ -90,7 +96,6 @@
         <dependency>
             <groupId>org.slf4j</groupId>
             <artifactId>log4j-over-slf4j</artifactId>
-            <version>${log4j-over-slf4j.version}</version>
             <scope>test</scope>
         </dependency>
     </dependencies>
--- a/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
+++ b/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
@@ -25,16 +25,13 @@ import static org.apache.storm.kafka.spout.KafkaSpoutConfig.FirstPollOffsetStrat
 
 import java.util.Collection;
 import java.util.Collections;
-import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
-import java.util.NavigableSet;
 import java.util.Set;
-import java.util.TreeSet;
 import java.util.concurrent.TimeUnit;
 
 import org.apache.kafka.clients.consumer.ConsumerRebalanceListener;
@@ -48,6 +45,7 @@ import org.apache.kafka.common.errors.RetriableException;
 import org.apache.storm.kafka.spout.KafkaSpoutConfig.FirstPollOffsetStrategy;
 import org.apache.storm.kafka.spout.internal.KafkaConsumerFactory;
 import org.apache.storm.kafka.spout.internal.KafkaConsumerFactoryDefault;
+import org.apache.storm.kafka.spout.internal.OffsetManager;
 import org.apache.storm.kafka.spout.internal.Timer;
 import org.apache.storm.spout.SpoutOutputCollector;
 import org.apache.storm.task.TopologyContext;
@@ -58,18 +56,18 @@ import org.slf4j.LoggerFactory;
 
 public class KafkaSpout<K, V> extends BaseRichSpout {
     private static final long serialVersionUID = 4151921085047987154L;
+    //Initial delay for the commit and subscription refresh timers
+    public static final long TIMER_DELAY_MS = 500;
     private static final Logger LOG = LoggerFactory.getLogger(KafkaSpout.class);
-    private static final Comparator<KafkaSpoutMessageId> OFFSET_COMPARATOR = new OffsetComparator();
 
     // Storm
     protected SpoutOutputCollector collector;
 
     // Kafka
     private final KafkaSpoutConfig<K, V> kafkaSpoutConfig;
-    private final KafkaConsumerFactory kafkaConsumerFactory;
+    private KafkaConsumerFactory kafkaConsumerFactory;
     private transient KafkaConsumer<K, V> kafkaConsumer;
     private transient boolean consumerAutoCommitMode;
-
 
     // Bookkeeping
     private transient FirstPollOffsetStrategy firstPollOffsetStrategy;  // Strategy to determine the fetch offset of the first realized by the spout upon activation
@@ -78,7 +76,7 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
     private transient boolean initialized;                              // Flag indicating that the spout is still undergoing initialization process.
     // Initialization is only complete after the first call to  KafkaSpoutConsumerRebalanceListener.onPartitionsAssigned()
 
-    transient Map<TopicPartition, OffsetEntry> acked;           // Tuples that were successfully acked. These tuples will be committed periodically when the commit timer expires, after consumer rebalance, or on close/deactivate. Not used if it's AutoCommitMode
+    private transient Map<TopicPartition, OffsetManager> acked;           // Tuples that were successfully acked. These tuples will be committed periodically when the commit timer expires, after consumer rebalance, or on close/deactivate
     private transient Set<KafkaSpoutMessageId> emitted;                 // Tuples that have been emitted but that are "on the wire", i.e. pending being acked or failed. Not used if it's AutoCommitMode
     private transient Iterator<ConsumerRecord<K, V>> waitingToEmit;         // Records that have been polled and are queued to be emitted in the nextTuple() call. One record is emitted per nextTuple()
     private transient long numUncommittedOffsets;                       // Number of offsets that have been polled and emitted but not yet been committed. Not used if it's AutoCommitMode
@@ -87,13 +85,13 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
 
 
     public KafkaSpout(KafkaSpoutConfig<K, V> kafkaSpoutConfig) {
-        this(kafkaSpoutConfig, new KafkaConsumerFactoryDefault());
+        this(kafkaSpoutConfig, new KafkaConsumerFactoryDefault<>());
     }
     
     //This constructor is here for testing
     KafkaSpout(KafkaSpoutConfig<K, V> kafkaSpoutConfig, KafkaConsumerFactory<K, V> kafkaConsumerFactory) {
-        this.kafkaSpoutConfig = kafkaSpoutConfig;                 // Pass in configuration
         this.kafkaConsumerFactory = kafkaConsumerFactory;
+        this.kafkaSpoutConfig = kafkaSpoutConfig;
     }
 
     @Override
@@ -114,9 +112,9 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
         retryService = kafkaSpoutConfig.getRetryService();
 
         if (!consumerAutoCommitMode) {     // If it is auto commit, no need to commit offsets manually
-            commitTimer = new Timer(500, kafkaSpoutConfig.getOffsetsCommitPeriodMs(), TimeUnit.MILLISECONDS);
+            commitTimer = new Timer(TIMER_DELAY_MS, kafkaSpoutConfig.getOffsetsCommitPeriodMs(), TimeUnit.MILLISECONDS);
         }
-        refreshSubscriptionTimer = new Timer(500, kafkaSpoutConfig.getPartitionRefreshPeriodMs(), TimeUnit.MILLISECONDS);
+        refreshSubscriptionTimer = new Timer(TIMER_DELAY_MS, kafkaSpoutConfig.getPartitionRefreshPeriodMs(), TimeUnit.MILLISECONDS);
 
         acked = new HashMap<>();
         emitted = new HashSet<>();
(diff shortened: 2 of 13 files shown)


srdo:
I wound up keeping KafkaUnit for the tests anyway. It doesn't hurt to have some integration testing, and mocking the consumer is a bit of a hassle for no real gain.

harshach:
@srdo can you please add details on any testing you've done with these changes.

srdo:
@harshach Nothing beyond running the unit tests. Most of the changes are just moving code to new classes verbatim. The OffsetManager commit code changes should be getting hit by the test code.

hmcl:
@srdo What about calling the JIRA/PR Summary/Title: **Kafka Spout Refactoring to Increase Modularity and Testability**

srdo:
@hmcl Renamed. I'll wait until https://github.com/apache/storm/pull/1808 has been merged to fix conflicts here, since it's likely there's some overlap

srdo:
@hmcl Updated. This should be ready for review.

external/storm-kafka-client/pom.xml
```diff
@@ -52,7 +52,6 @@
         <dependency>
             <groupId>org.apache.kafka</groupId>
             <artifactId>kafka-clients</artifactId>
-            <version>${storm.kafka.client.version}</version>
```
HeartSaVioR: What's behind of the change? We're maintaining both `storm.kafka.version` and `storm.kafka.client.version`, and this line is for picking correct version.

srdo: Sure, I put it back.

srdo:
I'm a little unsure about whether the time unit change to Timer is a good idea. It might be better to either implement nanotime support in Storm Time, or to make the constructor parameters just be in milliseconds (I'm not sure how useful nano/microsecond support is for anyone). Anyone have any opinions on this?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -87,13 +84,14 @@
...
-        this(kafkaSpoutConfig, new KafkaConsumerFactoryDefault());
+        this.kafkaSpoutConfig = kafkaSpoutConfig;
+        setKafkaConsumerFactory(new KafkaConsumerFactoryDefault());
     }
     
-    //This constructor is here for testing
-    KafkaSpout(KafkaSpoutConfig<K, V> kafkaSpoutConfig, KafkaConsumerFactory<K, V> kafkaConsumerFactory) {
-        this.kafkaSpoutConfig = kafkaSpoutConfig;                 // Pass in configuration
+    //This setter is here for testing
+    final KafkaSpout setKafkaConsumerFactory(KafkaConsumerFactory<K, V> kafkaConsumerFactory) {
```
hmcl: I would rather leave this with the constructor. We shouldn't be making this class mutable unless we have a very good reason to do so.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -58,27 +56,26 @@
 
 public class KafkaSpout<K, V> extends BaseRichSpout {
     private static final long serialVersionUID = 4151921085047987154L;
+    public static final long INITIAL_TIMER_DELAY_MS = 500;
```
hmcl: DEFAULT_TIMER_DELAY_MS

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -347,9 +345,15 @@ private void commitOffsetsForAckedTuples() {
...
-                final OffsetEntry offsetEntry = tpOffset.getValue();
-                offsetEntry.commit(nextCommitOffsets.get(tpOffset.getKey()));
+            for (Map.Entry<TopicPartition, OffsetAndMetadata> tpOffset : nextCommitOffsets.entrySet()) {
+                //Update the OffsetManager for each committed partition, and update numUncommittedOffsets
+                final TopicPartition tp = tpOffset.getKey();
+                final OffsetManager offsetManager = acked.get(tp);
+                long oldCommittedOffset = offsetManager.getCommittedOffset();
+                long numCommittedOffsets = offsetManager.commit(tpOffset.getValue());
+                numUncommittedOffsets -= numCommittedOffsets;
+                LOG.debug("Committed offsets [{}-{} = {}] for topic-partition [{}]. [{}] uncommitted offsets across all topic partitions",
```
hmcl: For cohesiveness, I believe that this log message should be present in the `OffsetManager` class. If another (client) class creates an instance of `OffsetManager`,  that class will be required to implement this same log message. Furthermore, this log message dives deep in the internal details of `OffsetManager`, which will make it very hard for client classes to print such message without know the internal details.

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/SingleTopicKafkaSpoutTest.java
```diff
@@ -28,21 +28,36 @@
 import org.junit.Test;
 import org.mockito.ArgumentCaptor;
 
-import static org.junit.Assert.*;
-
 import java.util.Map;
 import java.util.stream.IntStream;
+
 import static org.mockito.Mockito.*;
```
hmcl: can you please remove import wildcards.

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/SingleTopicKafkaSpoutTest.java
```diff
@@ -28,21 +28,36 @@
...
 import static org.apache.storm.kafka.spout.builders.SingleTopicKafkaSpoutConfiguration.*;
+import static org.hamcrest.CoreMatchers.is;
```
hmcl: can you please remove import wildcards.

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/SingleTopicKafkaSpoutTest.java
hmcl: can you please remove import wildcards.

hmcl: can you please remove import wildcards.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
srdo: I'll change it back.

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/SingleTopicKafkaSpoutTest.java
srdo: Sure.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
srdo: It isn't a default, but a hardcoded value. It's mostly here so it can be referred to from tests. I don't think it should be in Timer, since it's not a property inherent to Timer, but to how KafkaSpout uses Timer. I could rename it to INITIAL_COMMIT_AND_SUBSCRIPTION_REFRESH_DELAY_MS to make it clear what it's for?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
srdo: I agree. I only moved it here because the log makes use of the numUncommittedOffsets spout field, which we can't pass into offsetManager.commit because it changes based on the return value of commit. I can split the log message into a part that logs how much was committed for which topic partition (inside OffsetManager), and then another log message that logs numUncommittedOffsets (inside KafkaSpout). What do you think?

hmcl: Perhaps we can split this into two log messages. 

The first one `LOG.debug("Committed offsets [{}-{} = {}] for topic-partition [{}]`  goes into `OffsetManager`

The second `LOG.debug("[{}] uncommitted offsets across all topic partitions` stays here.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
hmcl: let's just make it TIMER_DELAY_MS. Sweet and to the point.

hmcl:
@srdo everything looks OK for me. I just would rather leave the `Timer` in ns than in ms, and change the `SimulatedTime` class to support nanos. I think that moving forward that will be useful for other developers as well.

srdo:
@hmcl I gave updating Time to support nanoseconds a shot. Not really sure I'm happy with it, since it requires that we also switch the current millisecond support over to using System.nanoTime (System.nanoTime and System.currentTimeMillis are not on the same timeline, and are incomparable). 

If we're going to keep this change, Time.currentTimeMillis should be changed to a different name, so it doesn't imply that it's the same as System.currentTimeMillis.

@revans2 Do you have an opinion on nanosecond support in `o.a.s.u.Time`? (asking because you have a number of commits on the Time file)

revans2:
@srdo Sorry I am kind of coming into this half way through.  

History: System.nanoTime and System.currentTimeMillis have different uses and are not on the same time line.  nanoTime is guaranteed to be monotonically increasing so long as the box is up.  currentTimeMillis is not, because it is kept in sync with the system time.  currentTimeMillis on most OSes is a very cheap.  It is reading from shared memory, no system call needed.  It can be different if read from different threads even.  nanoTime, at least on x86 boxes end up reading a special register in the processor.  If there is a very small core count this is cheap.  However if you are on a system with many different cores or physical chips they all have to be synced up to guarantee that it is monotonically increasing so it ends up being about as expensive as a memory barrier.  Not super expensive but also not dead cheap.

As far as Time.java is concerned.  Simulated time can be whatever we want.  I don't care if we store nanos internally or millis internally.  It is all simulated anyways.  If we are not simulating I don't think we should mix the two or have one backed by the other.  They are separate for a reason and people most of the time will pick one or the other because of those differences.  If we want nano support then lest have Time support nano, but the milli APIs stay backed by System.currentTimeMillis, and the nano APIs are backed by System.nanoTime.

HeartSaVioR:
Let's back to actual usage of Timer.

```
        if (!consumerAutoCommitMode) {     // If it is auto commit, no need to commit offsets manually
            commitTimer = new Timer(500, kafkaSpoutConfig.getOffsetsCommitPeriodMs(), TimeUnit.MILLISECONDS);
        }
        refreshSubscriptionTimer = new Timer(500, kafkaSpoutConfig.getPartitionRefreshPeriodMs(), TimeUnit.MILLISECONDS);
```

Here we are taking period as milliseconds consistently, and dropping support nanoseconds on Timer means that up to 1 ms can be added to the period. Unless we're expecting users to set these config values to small value like 1 or so, I think it doesn't matter. Indeed we're having 500 ms as default value, which doesn't matter it will be 501 ms.

srdo:
@revans2 I agree, it is best to keep nanos and millis separate. I've updated Time to pass through to System.currrentTimeMillis for millisecond methods, and System.nanoTime for nanosecond methods. When simulating everything gets converted to nanos.

@HeartSaVioR I agree, I'm not really seeing a use case for submillisecond support in storm-kafka-client, because the things we're currently doing with timers is to set retry backoffs or commit periods. I don't think anyone would want either to be less than 1 millisecond.

@hmcl Is there a reason nano/microsecond support is needed?

I've updated Time and Timer to support nanoseconds (hopefully without introducing bugs).

revans2:
@srdo for Time and Timer for a pure correctness perspective the changes look good to me.  I don't see any issues.

I took a quick look at the rest of the patch and it too looks good, but I don't dig in to it so I don't feel comfortable giving a +1 at this point.  I would love to really take a look at it, but I am sadly swamped with other things right now.

hmcl:
@srdo I was also caught up in a couple of things. I will finish reviewing this tonight.

srdo:
Thanks, both of you :)

storm-core/src/jvm/org/apache/storm/utils/Time.java
```diff
@@ -42,10 +46,10 @@ public SimulatedTime() {
         public SimulatedTime(Number ms) {
```
hmcl: advanceTimeMs

HeartSaVioR:
@srdo Could you rebase this?

storm-core/src/jvm/org/apache/storm/utils/Time.java
```diff
@@ -88,59 +92,97 @@ public static boolean isSimulating() {
...
+            return simulatedCurrTimeNanos.get();
+        } else {
+            return System.nanoTime();
+        }
+    }
+    
     public static long currentTimeMillis() {
         if(simulating.get()) {
-            return simulatedCurrTimeMs.get();
+            return nanosToMillis(simulatedCurrTimeNanos.get());
```
srdo: Why does it need to be? It's not updating simulatedCurrTimeNanos, and it's just dividing by a constant.

srdo:
@HeartSaVioR Fixed conflict

hmcl:
+1 for all the KafkaSpout related code
+1 also for the SimulatedTime changes in the sense that the minor refactoring of this class does not change the semantics of the pre-existing code.

@srdo can you please squash all the commits and create a PR for 1.x-branch, such that we can backport these changes. Thanks.

srdo:
Squashed this and pushed 1.x version at https://github.com/apache/storm/pull/1941. Thanks all for good feedback :)

ptgoetz:
+1

PR #1833, opened 2016-12-18
STORM-2251 Integration test refers specific version of Storm which should be project version

HeartSaVioR:
* fix integration-test's pom.xml to have root pom as parent, and remove version and storm version

### DIFF
--- a/integration-test/pom.xml
+++ b/integration-test/pom.xml
@@ -19,17 +19,22 @@
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
 
+    <parent>
+        <artifactId>storm</artifactId>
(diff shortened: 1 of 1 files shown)


integration-test/pom.xml
```diff
@@ -48,33 +53,6 @@
...
-        </repository>
-        <repository>
-            <id>hortonworks</id>
-            <url>http://nexus-private.hortonworks.com/nexus/content/groups/public/</url>
-        </repository>
-    </repositories>
-
     <dependencies>
         <dependency>
             <groupId>org.testng</groupId>
```
srdo: Consider removing joda-time version too, it's defined in the root pom's dependency management.

srdo:
+1

integration-test/pom.xml
HeartSaVioR: Good point. Will address.

HeartSaVioR: Addressed.

harshach:
+1

PR #1835, opened 2016-12-19
STORM-2236: Kafka Spout with manual partition management.

liurenjie1024:
As the title says.

### DIFF
--- a/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
+++ b/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
@@ -24,7 +24,14 @@ import org.apache.kafka.clients.consumer.ConsumerRecords;
 import org.apache.kafka.clients.consumer.KafkaConsumer;
 import org.apache.kafka.clients.consumer.OffsetAndMetadata;
 import org.apache.kafka.common.TopicPartition;
+import org.apache.kafka.common.errors.InterruptException;
+import org.apache.kafka.common.errors.RetriableException;
 import org.apache.storm.kafka.spout.KafkaSpoutConfig.FirstPollOffsetStrategy;
+import org.apache.storm.kafka.spout.internal.KafkaConsumerFactory;
+import org.apache.storm.kafka.spout.internal.KafkaConsumerFactoryDefault;
+import org.apache.storm.kafka.spout.internal.Timer;
+import org.apache.storm.kafka.spout.internal.fetcher.KafkaRecordsFetcher;
+import org.apache.storm.kafka.spout.internal.fetcher.KafkaRecordsFetchers;
 import org.apache.storm.spout.SpoutOutputCollector;
 import org.apache.storm.task.TopologyContext;
 import org.apache.storm.topology.OutputFieldsDeclarer;
@@ -46,16 +53,11 @@ import java.util.NavigableSet;
 import java.util.Set;
 import java.util.TreeSet;
 import java.util.concurrent.TimeUnit;
-import java.util.regex.Pattern;
 
 import static org.apache.storm.kafka.spout.KafkaSpoutConfig.FirstPollOffsetStrategy.EARLIEST;
 import static org.apache.storm.kafka.spout.KafkaSpoutConfig.FirstPollOffsetStrategy.LATEST;
 import static org.apache.storm.kafka.spout.KafkaSpoutConfig.FirstPollOffsetStrategy.UNCOMMITTED_EARLIEST;
 import static org.apache.storm.kafka.spout.KafkaSpoutConfig.FirstPollOffsetStrategy.UNCOMMITTED_LATEST;
-import org.apache.storm.kafka.spout.internal.KafkaConsumerFactory;
-import org.apache.storm.kafka.spout.internal.KafkaConsumerFactoryDefault;
-
-import org.apache.kafka.common.errors.InterruptException;
 
 public class KafkaSpout<K, V> extends BaseRichSpout {
     private static final Logger LOG = LoggerFactory.getLogger(KafkaSpout.class);
@@ -63,6 +65,7 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
 
     // Storm
     protected SpoutOutputCollector collector;
+    private TopologyContext topologyContext;
 
     // Kafka
     private final KafkaSpoutConfig<K, V> kafkaSpoutConfig;
@@ -77,6 +80,7 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
     private transient KafkaSpoutRetryService retryService;              // Class that has the logic to handle tuple failure
     private transient Timer commitTimer;                                // timer == null for auto commit mode
     private transient boolean initialized;                              // Flag indicating that the spout is still undergoing initialization process.
+    private transient KafkaRecordsFetcher<K, V> recordsFetcher;         // Class that encapsulates the logic of managing partitions and fetching records
     // Initialization is only complete after the first call to  KafkaSpoutConsumerRebalanceListener.onPartitionsAssigned()
 
     private KafkaSpoutStreams kafkaSpoutStreams;                        // Object that wraps all the logic to declare output fields and emit tuples
@@ -102,9 +106,9 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
     @Override
     public void open(Map conf, TopologyContext context, SpoutOutputCollector collector) {
         initialized = false;
-
         // Spout internals
         this.collector = collector;
+        this.topologyContext = context;
         maxRetries = kafkaSpoutConfig.getMaxTupleRetries();
         numUncommittedOffsets = 0;
 
@@ -221,7 +225,11 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
                 }
 
                 if (poll()) {
-                    setWaitingToEmit(pollKafkaBroker());
+                    try {
+                        setWaitingToEmit(pollKafkaBroker());
+                    } catch (RetriableException e) {
+                        LOG.error("Failed to poll from kafka.", e);
+                    }
                 }
 
                 if (waitingToEmit()) {
@@ -277,7 +285,7 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
     private ConsumerRecords<K, V> pollKafkaBroker() {
         doSeekRetriableTopicPartitions();
 
-        final ConsumerRecords<K, V> consumerRecords = kafkaConsumer.poll(kafkaSpoutConfig.getPollTimeoutMs());
+        final ConsumerRecords<K, V> consumerRecords = recordsFetcher.fetchRecords(kafkaSpoutConfig.getPollTimeoutMs());
         final int numPolledRecords = consumerRecords.count();
         LOG.debug("Polled [{}] records from Kafka. [{}] uncommitted offsets across all topic partitions", numPolledRecords, numUncommittedOffsets);
         return consumerRecords;
@@ -404,19 +412,8 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
 
     private void subscribeKafkaConsumer() {
         kafkaConsumer = kafkaConsumerFactory.createConsumer(kafkaSpoutConfig);
-
-        if (kafkaSpoutStreams instanceof KafkaSpoutStreamsNamedTopics) {
-            final List<String> topics = ((KafkaSpoutStreamsNamedTopics) kafkaSpoutStreams).getTopics();
-            kafkaConsumer.subscribe(topics, new KafkaSpoutConsumerRebalanceListener());
-            LOG.info("Kafka consumer subscribed topics {}", topics);
-        } else if (kafkaSpoutStreams instanceof KafkaSpoutStreamsWildcardTopics) {
-            final Pattern pattern = ((KafkaSpoutStreamsWildcardTopics) kafkaSpoutStreams).getTopicWildcardPattern();
-            kafkaConsumer.subscribe(pattern, new KafkaSpoutConsumerRebalanceListener());
-            LOG.info("Kafka consumer subscribed topics matching wildcard pattern [{}]", pattern);
-        }
-        // Initial poll to get the consumer registration process going.
-        // KafkaSpoutConsumerRebalanceListener will be called following this poll, upon partition registration
-        kafkaConsumer.poll(0);
+        recordsFetcher = KafkaRecordsFetchers.create(kafkaSpoutConfig, kafkaConsumer, topologyContext,
+            new KafkaSpoutConsumerRebalanceListener());
     }
 
     @Override
@@ -610,62 +607,6 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
                     ", committedOffset=" + committedOffset +
                     ", ackedMsgs=" + ackedMsgs +
                     '}';
-        }
-    }
-
-    // =========== Timer ===========
-
-    private class Timer {
-        private final long delay;
-        private final long period;
-        private final TimeUnit timeUnit;
-        private final long periodNanos;
-        private long start;
-
-        /**
-         * Creates a class that mimics a single threaded timer that expires periodically. If a call to {@link
-         * #isExpiredResetOnTrue()} occurs later than {@code period} since the timer was initiated or reset, this method returns
-         * true. Each time the method returns true the counter is reset. The timer starts with the specified time delay.
-         *
-         * @param delay    the initial delay before the timer starts
-         * @param period   the period between calls {@link #isExpiredResetOnTrue()}
-         * @param timeUnit the time unit of delay and period
-         */
-        public Timer(long delay, long period, TimeUnit timeUnit) {
-            this.delay = delay;
-            this.period = period;
-            this.timeUnit = timeUnit;
-
-            periodNanos = timeUnit.toNanos(period);
-            start = System.nanoTime() + timeUnit.toNanos(delay);
-        }
-
-        public long period() {
-            return period;
-        }
-
-        public long delay() {
-            return delay;
-        }
-
-        public TimeUnit getTimeUnit() {
-            return timeUnit;
-        }
-
-        /**
-         * Checks if a call to this method occurs later than {@code period} since the timer was initiated or reset. If that is the
-         * case the method returns true, otherwise it returns false. Each time this method returns true, the counter is reset
-         * (re-initiated) and a new cycle will start.
-         *
-         * @return true if the time elapsed since the last call returning true is greater than {@code period}. Returns false
-         * otherwise.
-         */
-        public boolean isExpiredResetOnTrue() {
-            final boolean expired = System.nanoTime() - start > periodNanos;
-            if (expired) {
-                start = System.nanoTime();
-            }
-            return expired;
         }
     }
 }
--- a/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutConfig.java
+++ b/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutConfig.java
@@ -36,6 +36,7 @@ public class KafkaSpoutConfig<K, V> implements Serializable {
     public static final long DEFAULT_OFFSET_COMMIT_PERIOD_MS = 30_000;   // 30s
     public static final int DEFAULT_MAX_RETRIES = Integer.MAX_VALUE;     // Retry forever
     public static final int DEFAULT_MAX_UNCOMMITTED_OFFSETS = 10_000_000;    // 10,000,000 records => 80MBs of memory footprint in the worst case
+    public static final long DEFAULT_PARTITION_REFRESH_PERIOD_MS = 2_000; // 2s
 
     // Kafka property names
     public interface Consumer {
@@ -76,6 +77,8 @@ public class KafkaSpoutConfig<K, V> implements Serializable {
     private final long offsetCommitPeriodMs;
     private final int maxRetries;
     private final int maxUncommittedOffsets;
+    private final long partitionRefreshPeriodMs;
+    private final boolean manualPartitionAssignment;
     private final FirstPollOffsetStrategy firstPollOffsetStrategy;
     private final KafkaSpoutStreams kafkaSpoutStreams;
     private final KafkaSpoutTuplesBuilder<K, V> tuplesBuilder;
@@ -91,6 +94,8 @@ public class KafkaSpoutConfig<K, V> implements Serializable {
         this.firstPollOffsetStrategy = builder.firstPollOffsetStrategy;
         this.kafkaSpoutStreams = builder.kafkaSpoutStreams;
         this.maxUncommittedOffsets = builder.maxUncommittedOffsets;
+        this.partitionRefreshPeriodMs = builder.partitionRefreshPeriodMs;
+        this.manualPartitionAssignment = builder.manualPartitionAssignment;
         this.tuplesBuilder = builder.tuplesBuilder;
         this.retryService = builder.retryService;
     }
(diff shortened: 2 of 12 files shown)


external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/fetcher/ManualKafkaRecordsFetcher.java
```diff
@@ -0,0 +1,88 @@
...
+
+        return consumer.poll(fetchTimeoutMs);
+    }
+
+    public interface PartitionAssignmentChangeListener {
+        void onPartitionAssignmentChange(Set<TopicPartition> oldPartitions, Set<TopicPartition> newPartitions);
+    }
+
+    public static PartitionAssignmentChangeListener listenerOf(final ConsumerRebalanceListener consumerRebalanceListener) {
+        return new PartitionAssignmentChangeListener() {
```
srdo: You don't have to change it, but on master we're targeting Java 8, so lambdas are an option for this kind of thing :)

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaPartitionReader.java
```diff
@@ -0,0 +1,10 @@
+package org.apache.storm.kafka.spout;
+
+import org.apache.kafka.clients.consumer.KafkaConsumer;
+import org.apache.kafka.common.TopicPartition;
+
+import java.util.List;
+
+public interface KafkaPartitionReader {
```
srdo: Nit: Some of these classes could probably go in the internal package, or a new subpackage of the internal package.

srdo:
@hmcl You might want to take a look at this too

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaPartitionReaders.java
```diff
@@ -0,0 +1,27 @@
+package org.apache.storm.kafka.spout;
+
+import org.apache.kafka.common.PartitionInfo;
+import org.apache.kafka.common.TopicPartition;
+
+import java.util.HashSet;
+
+/**
+ * Created by liurenjie on 19/12/2016.
```
liurenjie1024: Fixed.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -75,6 +78,7 @@
     private transient KafkaSpoutRetryService retryService;              // Class that has the logic to handle tuple failure
     private transient Timer commitTimer;                                // timer == null for auto commit mode
     private transient boolean initialized;                              // Flag indicating that the spout is still undergoing initialization process.
+    private transient KafkaRecordsFetcher<K, V> recordsFetcher;         // Class that encapsulate the log of partition management
```
liurenjie1024: Sorry, that's a typo. I've fixed in the new patch.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -120,6 +124,8 @@ public void open(Map conf, TopologyContext context, SpoutOutputCollector collect
             commitTimer = new Timer(500, kafkaSpoutConfig.getOffsetsCommitPeriodMs(), TimeUnit.MILLISECONDS);
         }
 
+        // Manual partition assignment
```
liurenjie1024: Fixed.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -218,7 +224,11 @@ public void nextTuple() {
             }
 
             if (poll()) {
-                setWaitingToEmit(pollKafkaBroker());
+                try {
+                    setWaitingToEmit(pollKafkaBroker());
+                } catch (Exception e) {
+                    LOG.error("Failed to poll from kafka.", e);
```
liurenjie1024: We have encountered many exceptions: KafkaFetchException, LeaderNoFoundException, etc. These are caused by temporarily kafka partition migration. I don't think we should restart the process in these cases.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/fetcher/ManualKafkaRecordsFetcher.java
```diff
@@ -0,0 +1,88 @@
...
+        this.thisTaskIndex = thisTaskIndex;
+        this.totalTaskCount = totalTaskCount;
+        this.partitionReader = partitionReader;
+        this.partitionRefreshTimer = partitionRefreshTimer;
+        this.partitionAssignmentChangeListener = partitionAssignmentChangeListener;
+
+        doRefreshMyPartitions();
+    }
+
+    private void refreshMyPartitionsIfNeed() {
```
liurenjie1024: Fixed.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/fetcher/ManualKafkaRecordsFetcher.java
liurenjie1024: Cool, I also love lambda.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaPartitionReader.java
liurenjie1024: Moved to another package.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
srdo: Right, that makes sense. We're not getting them because I think automatic assignment avoids them. I think it's fine to catch them, but you should replace the Exception with the specific exception classes you're seeing.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -218,7 +223,11 @@ public void nextTuple() {
             }
 
             if (poll()) {
-                setWaitingToEmit(pollKafkaBroker());
+                try {
+                    setWaitingToEmit(pollKafkaBroker());
+                } catch (KafkaException e) {
```
srdo: I think you should catch https://kafka.apache.org/0100/javadoc/org/apache/kafka/common/errors/RetriableException.html instead, it seems to cover all the exceptions that are expected to be transient (leader change for example, through InvalidMetadataException). Catching KafkaException will break interrupt handling (see https://github.com/apache/storm/pull/1821) because InterruptException is an instance of KafkaException.

hmcl:
@liurenjie1024 once the review is complete and it incorporates the feedback from all the reviewers (I am going through the review now), please squash the commits into one commit, and make sure that this commit has the title of the JIRA it is addressing, as well as a summary of its contents.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/fetcher/AutomaticKafkaRecordsFetcher.java
```diff
@@ -0,0 +1,52 @@
...
+import org.apache.storm.kafka.spout.KafkaSpoutStreams;
+import org.apache.storm.kafka.spout.KafkaSpoutStreamsNamedTopics;
+import org.apache.storm.kafka.spout.KafkaSpoutStreamsWildcardTopics;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.List;
+import java.util.regex.Pattern;
+
+/**
```
harshach: please remove author names and add a apache license on top of the file

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/fetcher/KafkaRecordsFetchers.java
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * Created by liurenjie on 19/12/2016.
+ */
+public final class KafkaRecordsFetchers {
+    public static <K, V> KafkaRecordsFetcher<K, V> kafkaRecordsFetcherOf(KafkaSpoutConfig kafkaSpoutConfig,
+                                                                         KafkaConsumer<K, V> consumer,
+                                                                         TopologyContext context,
+                                                                         ConsumerRebalanceListener rebalanceListener) {
+        if (kafkaSpoutConfig.getManualPartitionAssignment()) {
```
harshach: Do we need to check these flags. i.e you are already implementing an interface KafkaRecordsFetcher and any code related to Manual or Automatic should go into respective classes. User can configure one or the other . Instead of checking if its manual or not here we should be doing with in the record fetchers .

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/fetcher/AutomaticKafkaRecordsFetcher.java
liurenjie1024: Fixed.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/fetcher/KafkaRecordsFetchers.java
liurenjie1024: I think just a flag is easier to use since the user has to write code to configure it otherwise, especially in the case of using flux to configure storm topology. Though your way is more flexible but I don't think we need the flexibility here since it's an internal implementation and the user do not need to know much about that. Let the user assign the implementation class should be used in cases like tuple builder.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
liurenjie1024: Yes, I've fixed it.

srdo:
+1 once the merge conflict is resolved. It would probably be good to add some unit tests of the new classes, but I think this can be merged without them.

liurenjie1024:
@srdo Do I have the access to resolve the conflicts? Or it should be resolved by those who merge the code?

srdo:
@liurenjie1024 You have access. You need to open your branch in the Git client you usually use and pull the latest commits from master on this repo (https://github.com/apache/storm.git). If you like you can use `git pull --rebase` to avoid a merge commit in the history. Once your branch is up to date, push it back to your fork. The PR will update automatically.

Here's an example of how it can be done from command line (I have a remote called "upstream" that points to this repository, if you don't have one like it use `git remote add upstream https://github.com/apache/storm.git` to add it).
```
git checkout STORM-2236
git pull upstream master --rebase //you'll get a merge conflict
git mergetool //resolve merge conflict (you can also use whatever git GUI you usually use)
git add <the files you resolved conflicts for>
git rebase --continue
//do the above 3 lines until the rebase is over
//At this point you should verify that the branch diff to master looks reasonable. If it looks fine, push to your branch with the following.
git push --force
```

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/fetcher/KafkaRecordsFetchers.java
harshach: @liurenjie1024 internal details or not we will keep adding one implementation or another so it's better to make this cleaner when we have an opportunity to do so.

harshach:
@liurenjie1024 @srdo lets wait till all the pending comments are resolved.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/fetcher/KafkaRecordsFetchers.java
harshach: @liurenjie1024 irrespective of if it's an internal detail or not we should go with the right approach. We have interfaces for a reason. If we keep adding implementations on how the records are being fetched we need to keep adding these if conditions. Given we are already making changes lets go with the right approach.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
hmcl: @srdo good point, but the main point is also that the code that is handling the interrupt exceptions is very fragile. Right now you were able to catch this, but it's very difficult for other devs to code accounting for that.

@liurenjie1024 we are basically catching and logging the exception. Is that really the intent? Is there a scenario that we should not catch the exception, in which we would rather let the process die, and then restart ? The way the code currently works, if no new records are polled anything (i.e. an exception occurred)  `waitingToEmit` will always be false.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/fetcher/ManualKafkaRecordsFetcher.java
```diff
@@ -0,0 +1,85 @@
...
+
+import org.apache.kafka.clients.consumer.ConsumerRebalanceListener;
+import org.apache.kafka.clients.consumer.ConsumerRecords;
+import org.apache.kafka.clients.consumer.KafkaConsumer;
+import org.apache.kafka.common.TopicPartition;
+import org.apache.storm.kafka.spout.internal.partition.KafkaPartitionReader;
+import org.apache.storm.kafka.spout.KafkaSpout;
+import org.apache.storm.kafka.spout.TopicPartitionComparator;
+
+import java.util.*;
```
hmcl: Please remove wildcard imports

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutConfig.java
```diff
@@ -307,6 +333,14 @@ public KafkaSpoutRetryService getRetryService() {
         return retryService;
     }
 
+    public long getPartitionRefreshPeriodMs() {
+        return partitionRefreshPeriodMs;
+    }
+
+    public boolean getManualPartitionAssignment() {
```
hmcl: isManualPartitionAssignment()

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/fetcher/KafkaRecordsFetchers.java
hmcl: Agree. If you need to reuse some code across the manual and automatic mode, put it in a base abstract class, and then provide two implementations, one for auto and another for manual.

liurenjie1024: 1. The if conditions are wrapped in a factory method so adding more cases will not pollute other components.
2. The construction of instances can only happen in the open method since it needs topology context.
3. Interface is used to hide the client from implementation details and I don't think our approach here violated the rule.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -61,6 +64,7 @@
 
     // Storm
     protected SpoutOutputCollector collector;
+    private TopologyContext topologyContext;
```
liurenjie1024: Yes, it's used to create records fetcher.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutConfig.java
liurenjie1024: Fixed.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -591,7 +589,7 @@ public String toString() {
 
     // =========== Timer ===========
 
-    private class Timer {
+    public static class Timer {
```
hmcl: I think that we should make Timer it's own class, rather than a static inner class, since it's used in other objects as well.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/fetcher/ManualKafkaRecordsFetcher.java
```diff
@@ -0,0 +1,85 @@
...
+        if (!partitionRefreshTimer.isExpiredResetOnTrue()) {
+            return;
+        }
+
+        doRefreshMyPartitions();
+    }
+
+    private void doRefreshMyPartitions() {
+        List<TopicPartition> topicPartitions = partitionReader.readPartitions(consumer);
+        Collections.sort(topicPartitions, KAFKA_TOPIC_PARTITION_COMPARATOR);
```
hmcl: Why is this sort necessary ?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/partition/KafkaPartitionReaders.java
```diff
@@ -0,0 +1,27 @@
...
+import org.apache.kafka.common.PartitionInfo;
+import org.apache.kafka.common.TopicPartition;
+import org.apache.storm.kafka.spout.KafkaSpoutStreams;
+import org.apache.storm.kafka.spout.KafkaSpoutStreamsNamedTopics;
+import org.apache.storm.kafka.spout.KafkaSpoutStreamsWildcardTopics;
+
+import java.util.HashSet;
+
+public final class KafkaPartitionReaders {
+    public static KafkaPartitionReader partitionReaderOf(KafkaSpoutStreams kafkaSpoutStreams) {
```
hmcl: these names are very misleading. Can you please call this create or newInstance?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/fetcher/ManualKafkaRecordsFetcher.java
liurenjie1024: This sort ensures that all instances of KafkaSpout see the same ordered list so that the calculation is correct.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/partition/KafkaPartitionReaders.java
liurenjie1024: Fixed.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/fetcher/ManualKafkaRecordsFetcher.java
liurenjie1024: Fixed.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/Timer.java
```diff
@@ -0,0 +1,60 @@
+package org.apache.storm.kafka.spout.internal;
+
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Created by liurenjie on 22/12/2016.
```
srdo: You should really just edit the template in your IDE :)

I'm making the same change here https://github.com/apache/storm/pull/1832/files#diff-73306e32282def8d045e0bcd2d33f59b btw.

liurenjie1024: Sorry for that. I've fixed it. I'll merge your change later after you submitted it.

liurenjie1024:
Hi, all: 
I've resolved the comments and when this patch can be merged into master? I've read trident source code and it seems that there exists some bugs there, and the bug fix may rely on codes in this branch, so I hope it can be merge ASAP.

srdo:
Please copy the Apache license into all the new files. You can just grab it from the top of KafkaSpout.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/fetcher/KafkaRecordsFetchers.java
```diff
@@ -0,0 +1,56 @@
...
+        if (kafkaSpoutConfig.isManualPartitionAssignment()) {
+            int thisTaskIndex = context.getThisTaskIndex();
+            int totalTaskCount = context.getComponentTasks(context.getThisComponentId()).size();
+            KafkaPartitionReader partitionReader = KafkaPartitionReaders.create(
+                kafkaSpoutConfig.getKafkaSpoutStreams());
+            Timer partitionRefreshTimer = new Timer(500,
+                kafkaSpoutConfig.getPartitionRefreshPeriodMs(), TimeUnit.MILLISECONDS);
+
+            ManualKafkaRecordsFetcher.PartitionAssignmentChangeListener partitionAssignmentChangeListener = null;
+            if (rebalanceListener != null) {
```
srdo: Won't this cause an NPE in ManualKafkaRecordsFetcher if the null case is hit here? Why does it make sense to call create without a rebalance listener?

liurenjie1024: I can't figure out a case for a null rebalance listener now, so this maybe an overdesign here.

liurenjie1024:
@harshach @srdo @hmcl Any updates?

srdo:
LGTM. +1

harshach:
@liurenjie1024 thanks for the updates. Overall looks good to me. Can you squash your commits into one commit. +1 after that.

hmcl:
+1. Just one minor nit on a name. can you please squash the commits and put a little summary of the patch in the commit message. Thanks.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/partition/KafkaPartitionReaders.java
```diff
@@ -0,0 +1,27 @@
...
+                KafkaSpoutStreamsNamedTopics.class.cast(kafkaSpoutStreams).getTopics()));
+        } else if (kafkaSpoutStreams instanceof KafkaSpoutStreamsWildcardTopics) {
+            return new WildcardTopicPartitionReader(
+                KafkaSpoutStreamsWildcardTopics.class.cast(kafkaSpoutStreams).getTopicWildcardPattern());
+        } else {
+            throw new IllegalArgumentException("Unrecognized kafka spout stream: " + kafkaSpoutStreams.getClass());
+        }
+    }
+
+    public static TopicPartition convert(PartitionInfo partitionInfo) {
```
liurenjie1024: Fixed.

liurenjie1024:
Hi, all:
All commits have been squashed into one.

liurenjie1024:
Any update about the merge process of this patch?

harshach:
@liurenjie1024 thanks for you patience. Merged into master. I would like to merge this into 1.x-branch as well. You are using FunctionalInterface and 1.x branch we are still on java 7 . Do you want to change that open another PR for 1.x-branch?

liurenjie1024:
@harshach OK I'll submit a PR for that later.

PR #1837, opened 2016-12-21
[STORM-2254] Provide Socket time out for nimbus thrift client

redsanket:
For DRPC we have seen during our integration tests that the connection needs to be up for more than 10 minutes so not adding any timeout for it.

### DIFF
--- a/storm-core/src/jvm/org/apache/storm/security/auth/SaslTransportPlugin.java
+++ b/storm-core/src/jvm/org/apache/storm/security/auth/SaslTransportPlugin.java
@@ -64,8 +64,14 @@ public abstract class SaslTransportPlugin implements ITransportPlugin {
     @Override
     public TServer getServer(TProcessor processor) throws IOException, TTransportException {
         int port = type.getPort(storm_conf);
+        Integer socketTimeout = type.getSocketTimeOut(storm_conf);
         TTransportFactory serverTransportFactory = getServerTransportFactory();
-        TServerSocket serverTransport = new TServerSocket(port);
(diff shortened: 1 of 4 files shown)


HeartSaVioR:
+1

storm-core/src/jvm/org/apache/storm/security/auth/SaslTransportPlugin.java
```diff
@@ -64,8 +64,14 @@ public void prepare(ThriftConnectionType type, Map storm_conf, Configuration log
     @Override
     public TServer getServer(TProcessor processor) throws IOException, TTransportException {
         int port = type.getPort(storm_conf);
+        Integer socketTimeout = type.getSocketTimeOut(storm_conf);
         TTransportFactory serverTransportFactory = getServerTransportFactory();
-        TServerSocket serverTransport = new TServerSocket(port);
+        TServerSocket serverTransport = null;
+        if (socketTimeout != null) {
```
hmcl: if socketTimeout is defined in defaults.yaml, how can it ever be null ?

redsanket: For NIMBUS thrift client it is not null but for DRPC there is a possibility of it being null
as per since the DRPC intialization here is null. Refering to https://github.com/apache/storm/pull/1837/files#diff-8a40b66d685fb8c7b58cfaba6272e14aR82 
and https://github.com/apache/storm/pull/1837/files#diff-8a40b66d685fb8c7b58cfaba6272e14aR82

PR #1843, opened 2016-12-27
Fix storm.py which breaks when client.jartransformer.class is not set

HeartSaVioR:
When such configuration is not set, `confvalue` returns `nil` in 1.x branch, and `null` in master branch.

Such difference is based on whether language storm script runs: 
- for 1.x, it runs `org.apache.storm.command.config_value` which is Clojure file
- for master, it runs `org.apache.storm.command.ConfigValue` which is Java file

I accidentally changed that value from `nil` to `null` without recognized the difference between master and 1.x. 
Since it breaks all deployments on 1.x branch when client.jartransformer.class is not set, so it must be fixed before releasing.


harshach:
+1

PR #1849, opened 2017-01-03
STORM-2267 Use user's local maven repo. directory to local repo.

HeartSaVioR:
* use user's maven local repository if possible
* if it's not available, just use 'local-repo' (it will be created if not exists)


harshach:
+1

PR #1851, opened 2017-01-04
STORM-2270 Kafka spout should consume from latest when ZK partition commit offset bigger than the latest offset

danny0405:
Kafka spout should consume from latest when ZK offset bigger than partition's latest offset[ an TopicOffsetOutOfRangeException thrown out ], especially when Kafka topic change it's leader and some data lost, if we consume from earliest offset, much meaningless duplicate records will be re-consumed. So, we should consume from latest offset instead.
This is the jira task: [STORM-2270](https://issues.apache.org/jira/browse/STORM-2270)

### DIFF
--- a/external/storm-kafka/src/jvm/org/apache/storm/kafka/PartitionManager.java
+++ b/external/storm-kafka/src/jvm/org/apache/storm/kafka/PartitionManager.java
@@ -202,7 +202,7 @@ public class PartitionManager {
             long partitionLatestOffset = KafkaUtils.getOffset(_consumer, _partition.topic, _partition.partition, kafka.api.OffsetRequest.LatestTime());
             if (partitionLatestOffset < offset) {
                 offset = partitionLatestOffset;
-            }else{
+            } else {
                 offset = KafkaUtils.getOffset(_consumer, _partition.topic, _partition.partition, kafka.api.OffsetRequest.EarliestTime());
             }
             // fetch failed, so don't update the fetch metrics
--- a/external/storm-kafka/src/jvm/org/apache/storm/kafka/PartitionManager.java
+++ b/external/storm-kafka/src/jvm/org/apache/storm/kafka/PartitionManager.java
@@ -199,7 +199,12 @@ public class PartitionManager {
         try {
             msgs = KafkaUtils.fetchMessages(_spoutConfig, _consumer, _partition, offset);
         } catch (TopicOffsetOutOfRangeException e) {
-            offset = KafkaUtils.getOffset(_consumer, _partition.topic, _partition.partition, kafka.api.OffsetRequest.EarliestTime());
+            long partitionLatestOffset = KafkaUtils.getOffset(_consumer, _partition.topic, _partition.partition, kafka.api.OffsetRequest.LatestTime());
+            if (partitionLatestOffset < offset) {
+                offset = partitionLatestOffset;
+            }else{
+                offset = KafkaUtils.getOffset(_consumer, _partition.topic, _partition.partition, kafka.api.OffsetRequest.EarliestTime());
+            }
             // fetch failed, so don't update the fetch metrics
             
             //fix bug [STORM-643] : remove outdated failed offsets


external/storm-kafka/src/jvm/org/apache/storm/kafka/PartitionManager.java
```diff
@@ -199,7 +199,12 @@ private void fill() {
         try {
             msgs = KafkaUtils.fetchMessages(_spoutConfig, _consumer, _partition, offset);
         } catch (TopicOffsetOutOfRangeException e) {
-            offset = KafkaUtils.getOffset(_consumer, _partition.topic, _partition.partition, kafka.api.OffsetRequest.EarliestTime());
+            long partitionLatestOffset = KafkaUtils.getOffset(_consumer, _partition.topic, _partition.partition, kafka.api.OffsetRequest.LatestTime());
+            if (partitionLatestOffset < offset) {
+                offset = partitionLatestOffset;
+            }else{
```
erikdw: nit: `}else{` --> `} else {`   (i.e., please be consistent with other code for the spacing style)

danny0405:
@erikdw thx for your nice review, i have change the style ^_^

HeartSaVioR:
Please change the PR's title to add a prefix "STORM-2270" so that it can be automatically linked.

erikdw:
@danny0405 : thanks for the adjustment! :)

danny0405:
@HeartSaVioR thx, i have changed the title

harshach:
+1

external/storm-kafka/src/jvm/org/apache/storm/kafka/PartitionManager.java
```diff
@@ -199,7 +199,12 @@ private void fill() {
...
-            offset = KafkaUtils.getOffset(_consumer, _partition.topic, _partition.partition, kafka.api.OffsetRequest.EarliestTime());
+            long partitionLatestOffset = KafkaUtils.getOffset(_consumer, _partition.topic, _partition.partition, kafka.api.OffsetRequest.LatestTime());
+            if (partitionLatestOffset < offset) {
+                offset = partitionLatestOffset;
+            } else {
+                offset = KafkaUtils.getOffset(_consumer, _partition.topic, _partition.partition, kafka.api.OffsetRequest.EarliestTime());
+            }
             // fetch failed, so don't update the fetch metrics
             
             //fix bug [STORM-643] : remove outdated failed offsets
```
HeartSaVioR: @danny0405
if processingNewTuples is false, we still need to know earliest available offset since discarding logic assumes offset is the earliest available offset. If not it might discard some tuples which offsets are still available on topic.

harshach: @HeartSaVioR  Given that this code is inside TopicOffsetOutOfRangeException, which can only happen if the spout requested for an offset that's not in Kafka Topic. So in which case do you see that we might disregard the valid offsets.

HeartSaVioR: @harshach 
Either is fine for me, but if we would want to keep this, comments below the if statement needs to be reflected since comment is based on precondition that offset is equal to earliest time.

harshach:
@danny0405 can you add comments in the TopicOffsetOutOfRangeException block on when that case will encounter.

external/storm-kafka/src/jvm/org/apache/storm/kafka/PartitionManager.java
HeartSaVioR: @danny0405 @harshach 
Sorry My bad. I found that I confused earliest and latest... We're good to go.

HeartSaVioR:
+1 Thanks for the great contribution.

HeartSaVioR:
@harshach I found my mistake, and now I'm +1 to merge. Do you want to wait @danny0405 for adding comments?

PR #1852, opened 2017-01-04
[STORM-2271] ClosedByInterruptException should be handled in few cases and removing a confusing debug statement

redsanket:
There are several places in the code where we catch and ignore an interrupted exception (because it meant the process is shutting down).  We should also include ClosedbyInterruptException which is an IOException that can also be triggered by an interruptedexception in a few cases.


revans2:
+1

HeartSaVioR:
+1

PR #1853, opened 2017-01-04
STORM-2264 OpaqueTridentKafkaSpout failing after STORM-2216

HeartSaVioR:
* use JSONValue.parse() instead of JSONValue.parseWithException() in TransactionState
  * this just rolls back to previous, doesn't provide better approach

Please note that this is just a quick and dirty fix given that the priority of STORM-2264 is critical.
Even after merging this, we should find better solution and apply soon. (#1844 is one of)

### DIFF
--- a/storm-core/src/jvm/org/apache/storm/trident/topology/state/TransactionalState.java
+++ b/storm-core/src/jvm/org/apache/storm/trident/topology/state/TransactionalState.java
@@ -162,7 +162,9 @@ public class TransactionalState {
         try {
             Object data;
             if(_curator.checkExists().forPath(path)!=null) {
-                data = JSONValue.parseWithException(new String(_curator.getData().forPath(path), "UTF-8"));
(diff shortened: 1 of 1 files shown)


hmcl:
+1
@HeartSaVioR [STORM-2265](https://github.com/apache/storm/pull/1844) is a duplicate of this that I found independently. We can go ahead and merge this (revert) fix. We can keep STORM-2265 open to track a more solid way to address this issue.

HeartSaVioR:
@revans2 Could you please take a look?

revans2:
+1 seems fine to me

revans2:
I am also OK with reverting STORM-2216 if it is causing a lot of issues.

HeartSaVioR:
@revans2 I think the direction of STORM-2216 is right, and we just found a spot which is not trivial to fix to the right way.

PR #1861, opened 2017-01-05
STORM-2277: Support given for storm-druid connector shaded jar 

omkreddy:
Support given for storm-druid connector shaded jar and added some debug logs to DruidBolt

### DIFF
--- a/external/storm-druid/pom.xml
+++ b/external/storm-druid/pom.xml
@@ -40,16 +40,26 @@
             <groupId>io.druid</groupId>
             <artifactId>tranquility-core_2.11</artifactId>
             <version>0.8.2</version>
-        </dependency>
-        <dependency>
-            <groupId>io.druid</groupId>
-            <artifactId>druid-server</artifactId>
-            <version>0.9.1</version>
+            <exclusions>
+                <exclusion>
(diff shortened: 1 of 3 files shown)


external/storm-druid/pom.xml
```diff
@@ -40,16 +40,26 @@
...
+                  <groupId>mysql</groupId>
+                  <artifactId>mysql-connector-java</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
         <dependency>
             <groupId>org.scala-lang</groupId>
             <artifactId>scala-library</artifactId>
             <version>2.11.8</version>
+            <scope>provided</scope>
```
HeartSaVioR: Should users include scala-language manually? Then it would be better to be documented.

HeartSaVioR:
Looks good to me overall. Left minor comment.

external/storm-druid/pom.xml
omkreddy: @HeartSaVioR  Thanks for the review.  Yes, Scala dependency can be provided by users. Updated the document.

HeartSaVioR:
+1

PR #1862, opened 2017-01-05
STORM-2278: Allow max number of disruptor queue flusher threads to be configurable

revans2:


HeartSaVioR:
+1
Build failure on storm-core is not related.

storm-core/src/jvm/org/apache/storm/utils/DisruptorQueue.java
```diff
@@ -63,6 +63,18 @@
     private static final String PREFIX = "disruptor-";
     private static final FlusherPool FLUSHER = new FlusherPool();
 
+    private static int getNumFlusherPoolThreads() {
+        int numThreads = 100;
+        try {
+            String threads = System.getProperty("num_flusher_pool_threads", "100");
```
HeartSaVioR: Rethinking this a bit: this will be going to be undocumented feature.
Given that we don't reuse this for outside of Storm project, how about adding property to Config and calling Utils.readStormConfig() to read the value?

revans2: I agree that it does not add a lot of value in it's current form.  If you want me to document it I am happy to.

revans2: The issue with using a Config for this is that readStormConfig inside the worker would read the system config, not the topology config, and would not let us override it on a per topology bases.  I will add in a config for the system default, but in the documents there indicate the system property that can be set to override it for a topology.

revans2:
@HeartSaVioR I rebased and I think I addressed your review comments.  Please have a look and see if it is what you wanted.

HeartSaVioR:
+1 Thanks for the fix.

PR #1863, opened 2017-01-05
fix: KafkaSpout is blocked in AutoCommitMode

XuMingmin:
What's the issue?
When Storm topology is run on 'At-Most-Once' mode, with ack executor=0, KafkaSpout cannot emit records after some time.

What's the cause?
Function poll() controls whether it need to pull more data from Kafka cluster. With the condition `!waitingToEmit() && numUncommittedOffsets < maxUncommittedOffsets`, it's always `FALSE` after reaching threshold, as `numUncommittedOffsets` is creasing and never reset on 'At-Most-Once' mode.  

What's the solution?
I change the condition to `!waitingToEmit() && (numUncommittedOffsets < maxUncommittedOffsets || consumerAutoCommitMode)`, that return `TRUE` with AutoCommitMode regarding of UnCommittedOffset, see the comments inline.
Also, it's not required to track emitted(msgId) and numUncommittedOffsets with AutoCommitMode .


srdo:
Nice find :) You should go to https://issues.apache.org/jira/ and create an issue for this. Then rename this PR to mention the jira issue number (see the other open PRs for an example). The commit message should probably also mention the issue number.

The fix itself looks fine to me.

HeartSaVioR:
@XuMingmin Could you rebase this? STORM-2225 touches many places on storm-kafka-client so you might need to check it's still valid.

XuMingmin:
@srdo I create https://issues.apache.org/jira/browse/STORM-2340, 

@HeartSaVioR , let me check with the latest code, and open a new pull request

XuMingmin:
@HeartSaVioR The issue if still there, so I create a new pull request https://github.com/apache/storm/pull/1919 after rebase,

XuMingmin:
close this one, use https://github.com/apache/storm/pull/1919 instead.

PR #1864, opened 2017-01-06
STORM-2176 Workers do not shutdown cleanly and worker hooks don't run when a topology is killed

HeartSaVioR:
* increase supervisor.worker.shutdown.sleep.secs to let workers kill themselves first even stuck
* disable shutdown hook for log4j2 to make sure logs are written after shutdown is started

Please note that I didn't touch JVM's shutdown hook (leaving it to sleep 1 sec and halt) given that 1 sec looks enough if there's no issue on the worker. In fact shutdown hook shouldn't take too much time.

Given that this is blocker for releasing 1.0.3 and 1.1.0, please review soon so that we can go ahead.

### DIFF
--- a/conf/defaults.yaml
+++ b/conf/defaults.yaml
@@ -150,7 +150,7 @@ supervisor.worker.start.timeout.secs: 120
 #how long between heartbeats until supervisor considers that worker dead and tries to restart it
 supervisor.worker.timeout.secs: 30
 #how many seconds to sleep for before shutting down threads on worker
-supervisor.worker.shutdown.sleep.secs: 1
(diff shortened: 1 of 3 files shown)


HeartSaVioR:
cherry-picking will work for 1.x and 1.0.x branches.

HeartSaVioR:
@ptgoetz Could you test my patch to see it can resolve STORM-2176 pretty well?

HeartSaVioR:
@ptgoetz Ping again. I think we can start release process for 1.0.3 after this, so would like to review this soon.

ptgoetz:
+1

In local mode, I found that the JVM never exits, but that's likely a different issue.

PR #1866, opened 2017-01-07
STORM-2279: Unable to open bolt page of storm ui

raghavgautam:
I have added logic to wrap the negative indexes around effectively giving it a python like behaviour.

### DIFF
--- a/storm-core/src/jvm/org/apache/storm/daemon/nimbus/Nimbus.java
+++ b/storm-core/src/jvm/org/apache/storm/daemon/nimbus/Nimbus.java
@@ -3600,7 +3600,7 @@ public class Nimbus implements Iface, Shutdownable, DaemonCommon {
                 List<Integer> tasks = compToTasks.get(StormCommon.EVENTLOGGER_COMPONENT_ID);
                 tasks.sort(null);
                 // Find the task the events from this component route to.
-                int taskIndex = TupleUtils.listHashCode(Arrays.asList(componentId)) %
+                int taskIndex = (TupleUtils.listHashCode(Arrays.asList(componentId)) % tasks.size() + tasks.size()) %
                         tasks.size();
                 int taskId = tasks.get(taskIndex);
                 String host = null;


HeartSaVioR:
+1

HeartSaVioR:
@raghavgautam Could you please submit a patch for 1.x as well? After that I'll merge. Thanks.

raghavgautam:
@HeartSaVioR Done.

storm-core/src/jvm/org/apache/storm/daemon/nimbus/Nimbus.java
```diff
@@ -3600,7 +3600,7 @@ public ComponentPageInfo getComponentPageInfo(String topoId, String componentId,
                 List<Integer> tasks = compToTasks.get(StormCommon.EVENTLOGGER_COMPONENT_ID);
                 tasks.sort(null);
                 // Find the task the events from this component route to.
-                int taskIndex = TupleUtils.listHashCode(Arrays.asList(componentId)) %
+                int taskIndex = (TupleUtils.listHashCode(Arrays.asList(componentId)) % tasks.size() + tasks.size()) %
```
arunmahadevan: This should match the task selected by fields grouping https://github.com/apache/storm/blob/master/storm-core/src/jvm/org/apache/storm/daemon/GrouperFactory.java#L159. 
The modulo technique will return a different value than Math.abs. Probably we should move it to some common utility function and use it in both places.

In 1.x branch the fields grouper doesn't seem to account for negative hashCode, so not sure why didn't it come up before.

raghavgautam: @arunmahadevan Thanks for pointing this out, I will send another pull request with the change.

HeartSaVioR: @raghavgautam Any updates on this? I'll fix this if you're busy with other things.

HeartSaVioR: Just filed this: https://issues.apache.org/jira/browse/STORM-2305
I'll start crafting the fix tomorrow if @raghavgautam is busy to get this done.

HeartSaVioR: @arunmahadevan In 1.x branch the field grouper does care about negative hashCode, via using [mod](https://clojuredocs.org/clojure.core/mod) instead of [rem](https://clojuredocs.org/clojure.core/rem).

http://stackoverflow.com/a/13060774/3480122

So it turned out we missed this point while porting to Java.

HeartSaVioR:
@arunmahadevan @raghavgautam 
To summary, this issue was not valid for 1.x and 1.0.x branch, since field grouper and also nimbus take care of negative hashCode via using `mod`. My bad. I'll revert them.

raghavgautam:
👍 for reverting from 1.x & 1.0.x branches.

storm-core/src/jvm/org/apache/storm/daemon/nimbus/Nimbus.java
arunmahadevan: @HeartSaVioR thanks for digging deeper and nice finding.

PR #1867, opened 2017-01-09
STORM-2280 Upgrade Calcite version to 1.11.0

HeartSaVioR:
* also update freemarker version since Calcite parser file requires that
* apply the change of Calcite parser file

Announcement will be held on tomorrow or so, but artifacts are already available in Maven.


arunmahadevan:
+1

PR #1869, opened 2017-01-09
STORM-2279: Unable to open bolt page of storm ui

raghavgautam:
I have added logic to wrap the negative indexes around effectively giving it a python like behaviour.
Same as pull #1866


HeartSaVioR:
+1

PR #1872, opened 2017-01-11
[STORM-2283] Fix multi threading issues in DefaultStateSerializer

arunmahadevan:
Raising a separate patch for the state issue fixed in https://github.com/apache/storm/pull/1693

### DIFF
--- a/storm-core/src/jvm/org/apache/storm/state/DefaultStateSerializer.java
+++ b/storm-core/src/jvm/org/apache/storm/state/DefaultStateSerializer.java
@@ -30,8 +30,21 @@ import java.util.List;
  * the state.
  */
 public class DefaultStateSerializer<T> implements Serializer<T> {
-    private final Kryo kryo;
-    private final Output output;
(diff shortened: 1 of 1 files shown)


HeartSaVioR:
+1 Confirmed this works well from #1693.

HeartSaVioR:
@arunmahadevan 
Just realized you're using Optional from JDK 8. 
Given that it's a bug across 1.x version lines, could you craft a patch for 1.x? Thanks in advance.

arunmahadevan:
@HeartSaVioR guava's Optional did not have a public default ctor and Kryo was throwing errors during deserialization. I will try to address it and raise a different patch for 1.x branch.

arunmahadevan:
@HeartSaVioR I made some change so that it works without any changes to RedisEncoder and compatible with 1.x-branch as well. Please take a look. I also applied the same changes to https://github.com/apache/storm/pull/1693

HeartSaVioR:
+1 Thanks for the quick fix.

PR #1873, opened 2017-01-12
STORM-2296 Kafka spout - no duplicates on leader changes

ernisv:
Current behavior of Kafka spout emits duplicate tuples whenever Kafka topic leader's change.
In case of exception caused by leader changes, PartitionManagers are simply recreated losing the state about which tuples were already emitted and new PartitionManager re-emits them again.

This is fine as at-least-once is fulfilled, but still it would be better to not emit duplicate data if possible.
Moreover this could be easily avoided by moving the state related to emitted tuples from old PartitionManager to new one.

### DIFF
--- a/external/storm-kafka/src/jvm/org/apache/storm/kafka/PartitionManager.java
+++ b/external/storm-kafka/src/jvm/org/apache/storm/kafka/PartitionManager.java
@@ -101,11 +101,12 @@ public class PartitionManager {
         if (previousManager != null) {
             _failedMsgRetryManager = previousManager._failedMsgRetryManager;
             _committedTo = previousManager._committedTo;
+            _emittedToOffset = previousManager._emittedToOffset;
             _waitingToEmit = previousManager._waitingToEmit;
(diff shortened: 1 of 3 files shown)


harshach:
@ernisv can you file JIRA under STORM here https://issues.apache.org/jira/browse/STORM-1757

ernisv:
Created JIRA issue: https://issues.apache.org/jira/browse/STORM-2296

harshach:
@ernisv please add to PR title as well. Also can you open this against 1.x-branch or master so that we can cherry-pick from there to other branches.

ernisv:
Created another pull request for 1.x-branch:

https://github.com/apache/storm/pull/1888

HeartSaVioR:
@ernisv Hi, I merged this patch to the master and 1.x branch, and I don't think this should be applied to 1.0.x branch since it's not a bugfix. If you agree please close this PR, and if you want to include this to bugfix version line please let me know.

ernisv:
That's fine - i'll close this PR

PR #1880, opened 2017-01-17
STORM-2297 [storm-opentsdb] Support Flux for OpenTSDBBolt

HeartSaVioR:
* Expose OpenTsdbClient.Builder's constructor to public
  * This allows Flux to initialize Builder instance which is needed for OpenTSDBBolt
* Add constructor for having single mapper instance to OpenTsdbBolt
  * Flux doesn't support array of reference for now
* Also fix a bug on OpenTsdbBolt: it should refer the mapper interface for flexibility, not implemented one

Ideally it would be better for Flux to support these features, but given that it's easy fix and no harm, I decided to make this work first.

@satishd I saw you reviewed previous pull request. Could you review this again? Thanks in advance!


harshach:
+1 looks good to me.

PR #1881, opened 2017-01-17
STORM-2298 Don't kill Nimbus when ClusterMetricsConsumer is failed to initialize

HeartSaVioR:
Please refer [STORM-2298](https://issues.apache.org/jira/browse/STORM-2298) for more details.

### DIFF
--- a/storm-core/src/jvm/org/apache/storm/metric/ClusterMetricsConsumerExecutor.java
+++ b/storm-core/src/jvm/org/apache/storm/metric/ClusterMetricsConsumerExecutor.java
@@ -17,7 +17,6 @@
  */
 package org.apache.storm.metric;
 
-import org.apache.storm.Config;
 import org.apache.storm.metric.api.DataPoint;
 import org.apache.storm.metric.api.IClusterMetricsConsumer;
 import org.slf4j.Logger;
@@ -27,6 +26,10 @@ import java.util.Collection;
 
 public class ClusterMetricsConsumerExecutor {
     public static final Logger LOG = LoggerFactory.getLogger(ClusterMetricsConsumerExecutor.class);
+    private static final String ERROR_MESSAGE_PREPARATION_CLUSTER_METRICS_CONSUMER_FAILED =
+            "Preparation of Cluster Metrics Consumer failed. " +
+            "Please check your configuration and/or corresponding systems and relaunch Nimbus. " +
+            "Skipping handle metrics.";
 
     private IClusterMetricsConsumer metricsConsumer;
     private String consumerClassName;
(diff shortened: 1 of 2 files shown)


storm-core/src/jvm/org/apache/storm/metric/ClusterMetricsConsumerExecutor.java
```diff
@@ -40,31 +39,49 @@ public ClusterMetricsConsumerExecutor(String consumerClassName, Object registrat
...
+            metricsConsumer.prepare(registrationArgument);
         } catch (Exception e) {
-            throw new RuntimeException("Could not instantiate a class listed in config under section " +
-                    Config.STORM_CLUSTER_METRICS_CONSUMER_REGISTER + " with fully qualified name " + consumerClassName, e);
-        }
+            LOG.error("Could not instantiate or prepare Cluster Metrics Consumer with fully qualified name " +
+                    consumerClassName, e);
 
-        metricsConsumer.prepare(registrationArgument);
+            if (metricsConsumer != null) {
```
harshach: do we want to move this to finally clause

HeartSaVioR: This is to clean up broken metrics consumer instance, so only needed for error case.

storm-core/src/jvm/org/apache/storm/metric/ClusterMetricsConsumerExecutor.java
```diff
@@ -40,31 +39,49 @@ public ClusterMetricsConsumerExecutor(String consumerClassName, Object registrat
...
+            metricsConsumer = null;
+        }
     }
 
     public void handleDataPoints(final IClusterMetricsConsumer.ClusterInfo clusterInfo, final Collection<DataPoint> dataPoints) {
-        try {
-            metricsConsumer.handleDataPoints(clusterInfo, dataPoints);
-        } catch (Throwable e) {
-            LOG.error("Error while handling cluster data points, consumer class: " + consumerClassName, e);
+        if (metricsConsumer == null) {
```
harshach: do we need to attempt to re-initialize the metricsConsumer in handleDataPoints. Given that this method gets executed in a schedule-recurring loop. 
If the metrics reporter isn't initialized it means that either the user is mis-configured or didn't supply the required class. Any of these errors means user need to fix and restart the nimbus. 
So in that case it would be save to initialize the metrics reporter only once on nimbus start and we can probably add a debug message in the handleDataPoints if metricsConsumer is null so that users can figure out if the metrics are not getting published.

HeartSaVioR: Some modules (at least what I've reported) could throw error on initialization phase when backend storage is not working properly. This could be intermittent so I thought it might be better to retry when needed, but I'm also OK to remove retry logic and leave error log if we don't want to retry automatically.

harshach: @HeartSaVioR I am worried about the state where its badly configured and its retrying every scheduled interval. I think we can move it to try initializing only once and log an error if its fails to do so.

HeartSaVioR: @harshach OK. I'll just make it leaving error log repeatedly on handleDataPoints and don't try initializing again.

HeartSaVioR: @harshach Addressed.

HeartSaVioR:
@harshach Thanks for reviewing. I addressed your comment.

HeartSaVioR:
@harshach ping for reviewing.

harshach:
+1

PR #1883, opened 2017-01-18
STORM-2300 [Flux] support list of references

HeartSaVioR:
Please refer [STORM-2300](https://issues.apache.org/jira/browse/STORM-2300) for more details.

* introduce reflist
* introduce BeanListReference which stores list of id of references
* handle BeanListReference properly
* modify unit test for testing reflist

This feature supports method which arguments contain List<Type>, Type[], Type... (varargs).


harshach:
+1. Looks good.

PR #1887, opened 2017-01-19
STORM-2303 [storm-opentsdb] Fix list invariant issue for JDK 7

HeartSaVioR:
* use bounded wildcard type to fix invariant issue

Succeed to build this with JDK 7, but let's see the result of Travis CI build again.


harshach:
+1

PR #1888, opened 2017-01-19
STORM-2296 Kafka spout no dup on leader changes,  1.x-branch

ernisv:
Current behavior of Kafka spout emits duplicate tuples whenever Kafka topic leader's change.
In case of exception caused by leader changes, PartitionManagers are simply recreated losing the state about which tuples were already emitted and new PartitionManager re-emits them again.

This is fine as at-least-once is fulfilled, but still it would be better to not emit duplicate data if possible.
Moreover this could be easily avoided by moving the state related to emitted tuples from old PartitionManager to new one.

### DIFF
--- a/external/storm-kafka/src/test/org/apache/storm/kafka/ZkCoordinatorTest.java
+++ b/external/storm-kafka/src/test/org/apache/storm/kafka/ZkCoordinatorTest.java
@@ -28,8 +28,7 @@ import org.mockito.MockitoAnnotations;
 
 import java.util.*;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNotEquals;
+import static org.junit.Assert.*;
 import static org.mockito.Matchers.any;
 import static org.mockito.Matchers.anyInt;
 import static org.mockito.Mockito.when;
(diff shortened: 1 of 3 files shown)


external/storm-kafka/src/test/org/apache/storm/kafka/ZkCoordinatorTest.java
```diff
@@ -114,6 +113,33 @@ public void testPartitionsChange() throws Exception {
...
+    public void testPartitionManagerRecreate() throws Exception {
+        final int totalTasks = 2;
+        int partitionsPerTask = 2;
+        List<ZkCoordinator> coordinatorList = buildCoordinators(totalTasks / partitionsPerTask);
+        when(reader.getBrokerInfo()).thenReturn(TestUtils.buildPartitionInfoList(TestUtils.buildPartitionInfo(totalTasks, 9092)));
+        List<List<PartitionManager>> partitionManagersBeforeRefresh = getPartitionManagers(coordinatorList);
+        waitForRefresh();
+        when(reader.getBrokerInfo()).thenReturn(TestUtils.buildPartitionInfoList(TestUtils.buildPartitionInfo(totalTasks, 9093)));
+        List<List<PartitionManager>> partitionManagersAfterRefresh = getPartitionManagers(coordinatorList);
+        assertEquals(partitionManagersAfterRefresh.size(), partitionManagersAfterRefresh.size());
```
HeartSaVioR: It compares exactly same thing. Did you miss here, or it's redundant line?

external/storm-kafka/src/test/org/apache/storm/kafka/ZkCoordinatorTest.java
```diff
@@ -114,6 +113,33 @@ public void testPartitionsChange() throws Exception {
...
+        Iterator<List<PartitionManager>> iterator = partitionManagersAfterRefresh.iterator();
+        for (List<PartitionManager> partitionManagersBefore : partitionManagersBeforeRefresh) {
+            List<PartitionManager> partitionManagersAfter = iterator.next();
+            for (PartitionManager before : partitionManagersBefore)
```
HeartSaVioR: I'd rather create a new hash map for before or after which key is partition, and compare based on the map and other list. The depth of `if` statement is already 4.

external/storm-kafka/src/test/org/apache/storm/kafka/ZkCoordinatorTest.java
```diff
@@ -114,6 +113,33 @@ public void testPartitionsChange() throws Exception {
...
+                for (PartitionManager after: partitionManagersAfter)
+                    if (before.getPartition().partition == after.getPartition().partition)
+                        assertStateIsTheSame(before, after);
+        }
+    }
+
+    private void assertStateIsTheSame(PartitionManager managerBefore, PartitionManager managerAfter) {
+        // check if state was actually moved from old PartitionManager
+        assertNotSame(managerBefore, managerAfter);
+        assertSame(managerBefore._waitingToEmit, managerAfter._waitingToEmit);
```
HeartSaVioR: Let's check all the copying fields, not only _waitingToEmit.

HeartSaVioR:
@ernisv 
Looks good. Minor nits. Build failing is unrelated.

external/storm-kafka/src/test/org/apache/storm/kafka/ZkCoordinatorTest.java
ernisv: Right, there's the same error in another test that got copy-pasted here :)
Fixed in both places.

external/storm-kafka/src/test/org/apache/storm/kafka/ZkCoordinatorTest.java
ernisv: Right, simplified the test.

external/storm-kafka/src/test/org/apache/storm/kafka/ZkCoordinatorTest.java
ernisv: I've added checks for _emittedToOffset and _committedTo fields.
However _failedMsgRetryManager and _pending are private, so we either do not check them or  increase the fields visibility to "package". Do you think it warrants visibility change ?

HeartSaVioR: No I think that's enough.

HeartSaVioR:
+1

HeartSaVioR:
@ernisv 
Could you squash your commits into one? 
`STORM-2296 Kafka spout no dup on leader changes` would be sufficient to new commit title, and you can describe current commits' titles to the body.

I can handle it if you mind, but then you need to close this PR manually after merging, since Github auto-close doesn't work against non-master branch.

ernisv:
Ok, I've squashed all the changes to single commit.

PR #1890, opened 2017-01-20
[STORM-2311]A browser which is outside the cluster cant's access stor…

liu-zhaokun:
JIRA Issue: [https://issues.apache.org/jira/browse/STORM-2311](url)

when storm cluster is in a security mode such as kerberos,We can not use a browser which is outside the cluster to access storm ui,even if we remove the configuration item which named ui.filter.There is a mistake like "server 500",because those methods to get cluster's info for ui can't access the cluster which is in a security mode.
It's not necessary to limit users to get the general information when storm cluster is in a security mode,so it's not necessary to limit a Get-Request.What should we do is to limit the Post-Request,preventing the external users to kill the topology,deactivate the topology and so on... We only should ensure users can't operator the cluster with the RestApi.

### DIFF
--- a/storm-core/src/clj/org/apache/storm/ui/core.clj
+++ b/storm-core/src/clj/org/apache/storm/ui/core.clj
@@ -1140,7 +1140,6 @@
   (GET "/api/v1/cluster/summary" [:as {:keys [cookies servlet-request]} & m]
     (.mark ui:num-cluster-summary-http-requests)
     (populate-context! servlet-request)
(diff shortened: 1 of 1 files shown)


harshach:
@liu-zhaokun we shouldn't be removing authorize methods. For UI to work in secure mode you need to go through the configuration section here https://github.com/apache/storm/blob/master/SECURITY.md#uilogviewer . This is not a issue. -1.

PR #1891, opened 2017-01-23
STORM-2315 Storm kafka client does not commit offsets when ack is disabled.

liurenjie1024:
Here is why this bug happens: when ack is disabled, the ack method of spout will be called in the emit method, and that's why emitted message ids can not be found and moved to acked tuples.

### DIFF
--- a/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
+++ b/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
@@ -336,18 +336,17 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
                             collector.emit(tuple);
                         }
                     } else {
-                        if (tuple instanceof KafkaTuple) {
-                            collector.emit(((KafkaTuple) tuple).getStream(), tuple, msgId);
-                        } else {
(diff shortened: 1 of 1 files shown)


harshach:
@liurenjie1024  If I understand correctly this bug happens when user runs topology.num.ackers=0 ?

liurenjie1024:
@harshach Yes that's the case.

harshach:
@liurenjie1024 I am not sure how this code changes will help though. If acking is not enabled adding auto.commit.enable to true on producer properties should solve this right?

liurenjie1024:
@harshach Even auto commit is enabled, the acked tuples can not be removed from emitted sets. Here is the calling sequence when ack is disabled:
1. outputCollector.emit(msgId)
    1.1 spout.ack(msgId) is called and the message id is not found in the emitted set
2. msgId is added into the emitted set

And that's why emitted message ids will never be cleared.

harshach:
@liurenjie1024  thanks for the details. +1.

HeartSaVioR:
@liurenjie1024 Could you rebase this? STORM-2225 touches many places on storm-kafka-client so you might need to check it's still valid.

liurenjie1024:
@harshach @HeartSaVioR Change rebased.

HeartSaVioR:
+1

PR #1894, opened 2017-01-25
STORM-2325 Logviewer doesn't consider 'storm.local.hostname' (1.x)

HeartSaVioR:
* consider storm.local.hostname first for creating link url in Logviewer

This should be easy to port back to 1.0.x branch.


ptgoetz:
+1

PR #1895, opened 2017-01-25
STORM-2323 Precondition for Leader Nimbus should check all topology blobs and also corresponding dependencies

HeartSaVioR:
* change the precondition for leader Nimbus
  * it should have all active topology blobs locally
  * it should also have all corresponding dependency blobs locally
  * corresponding dependencies will be extracted from topology codes (via getBlob)

Since it tries to get blobs, probably leader listener can try to access other Nimbuses, and someone might think this is a bit dangerous.
I also have other branch which only checks all active topology blobs, not corresponding dependency blobs, so please feel free to share your opinions.
https://github.com/HeartSaVioR/storm/tree/STORM-2323-no-check-dependencies

I'll create pull request for 1.x when we select one of these branch and review and ready to merge.

### DIFF
--- a/storm-core/src/jvm/org/apache/storm/zookeeper/Zookeeper.java
+++ b/storm-core/src/jvm/org/apache/storm/zookeeper/Zookeeper.java
@@ -20,6 +20,7 @@ package org.apache.storm.zookeeper;
 import com.google.common.base.Joiner;
 import com.google.common.collect.Sets;
 
+import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang.StringUtils;
 import org.apache.curator.framework.CuratorFramework;
 import org.apache.curator.framework.api.CuratorEvent;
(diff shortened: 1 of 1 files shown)


storm-core/src/jvm/org/apache/storm/zookeeper/Zookeeper.java
```diff
@@ -336,29 +340,53 @@ public static NimbusInfo toNimbusInfo(Participant participant) {
...
-                } else {
+                if (!diffTopology.isEmpty()) {
                     LOG.info("code for all active topologies not available locally, giving up leadership.");
                     try {
                         leaderLatch.close();
                     } catch (IOException e) {
                         throw new RuntimeException(e);
                     }
+
+                    return;
```
harshach: instead of adding a return can't we make this as if else

harshach: Also if the nimbus has incomplete set of topology resources, wouldn't it be better to cleanup the resources and let it sync from the newly elected leader.

HeartSaVioR: OK I'll try to remove `return` at this place.

Btw, synchronizing blobs are done for each, so I don't see benefit to clean up. If we're worried about inconsistency between versions of the topology blobs, I think it should be handled properly from BlobStore.

HeartSaVioR: Addressed.

harshach:
+1

HeartSaVioR:
Thanks @harshach for reviewing. I also created PR (#1903) for 1.x (and possibly 1.0.x) branch.

PR #1896, opened 2017-01-26
STORM-2326 Upgrade log4j and slf4j

jnioche:
Ran ` mvn clean install -DskipTests` without problems


jnioche:
As far as I understand, Storm uses SLF4J + LOG4J. I could not find a way to run a topology where one of the dependencies relies on a more recent version of log4j (Elasticsearch 5). A similar problem was reported on https://issues.apache.org/jira/browse/STORM-1386.
Upgrading the dependencies in Storm solved the problem.

jnioche:
See issue with #1900 - closing for now

PR #1897, opened 2017-01-26
STORM-2327 ConfigurableTopology

jnioche:
See https://issues.apache.org/jira/browse/STORM-2327


HeartSaVioR:
@jnioche 
Thanks for the contribution. It looks good.
Btw, it would be great if we apply ConfigurableTopology to all of topology implementations on example directory or at least storm-starter to expose this class to more users. What do you think?

jnioche:
Thanks @HeartSaVioR. Good idea, I'll have a look at the topology classes in storm-starter. Just came across the (StormRunner)[https://github.com/apache/storm/blob/master/examples/storm-starter/src/jvm/org/apache/storm/starter/util/StormRunner.java] class, which the ConfigurableTopology could replace.

HeartSaVioR:
@jnioche 
Yes I also think you can replace StormRunner with ConfigurableTopology. 
Btw, there're lots of topologies so you might want to only apply some of them. If that's the case please let me know. I think that would be also OK.

jnioche:
@HeartSaVioR I removed StormRunner and modified some of the topologies so that they extend ConfigurableTopology. Please let me know your thoughts on this.

examples/storm-starter/src/jvm/org/apache/storm/starter/RollingTopWords.java
```diff
@@ -17,114 +17,89 @@
...
+   * "slidingWindowCounts".
    *
    * Examples:
    *
    * ```
    *
-   * # Runs in local mode (LocalCluster), with topology name "slidingWindowCounts"
-   * $ storm jar storm-starter-jar-with-dependencies.jar org.apache.storm.starter.RollingTopWords
+   * # Runs in local mode (LocalCluster), with topology name
+   * "slidingWindowCounts" $ storm jar storm-starter-jar-with-dependencies.jar
```
HeartSaVioR: needs new line before `$`, and next line should be added after `jar`.

HeartSaVioR: same things are repeated below

HeartSaVioR:
@jnioche 
Could you also modify README.md on storm-starter to apply the change for topologies? Looks great otherwise.

examples/storm-starter/src/jvm/org/apache/storm/starter/RollingTopWords.java
jnioche: done!

HeartSaVioR:
+1 
@jnioche 
Could you squash commits into one? `STORM-2327 ConfigurableTopology` would be good as commit title for squashed commit.

jnioche:
I made a mess with this branch and have pushed a clean, squashed PR as a replacement -> #1907

PR #1899, opened 2017-01-27
STORM-2330: Fix storm sql code generation for UDAF with non standard sql types

arunmahadevan:
Storm-sql uses cacite's code generator. For UDAF with non standard return types, the generated code does not compile.

E.g. For an UDAF that returns a `List<T>` cacite converts the return type to SQL type "OTHER" and finally inserts a cast to Object[] in the returned code which fails to compile with "java.lang.ClassCastException: java.util.ArrayList cannot be cast to [Ljava.lang.Object;"

Use a custom type factory to return "ANY" as the type (instead of OTHER) if the java type cannot be mapped to a sql type.


HeartSaVioR:
+1

PR #1901, opened 2017-01-28
STORM-2331: Emitting from JavaScript should work when not anchoring

EitZei:
Currently emitting from JavaScript multilang bolt fails if trying to emit without anchor (ShellBolt will receive anchor list with one null item) or when trying to emit with multiple anchors (ShellBolt will receive nested array).

This fixes the handling of anchors in storm.js in a backwards compatible way.


HeartSaVioR:
+1 Nice catch.

PR #1902, opened 2017-01-30
STORM-2281: Running Multiple Kafka Spouts (Trident) Throws Illegal State Exception

hmcl:
To review this PR please review only the commit that has STORM-2281. It is currently sitting on top of the multiple commits addressing STORM-2225, waiting for them to be squashed in and eventually merged.

@harshach can you please review. Thanks.

### DIFF
--- a/examples/storm-kafka-client-examples/src/main/java/org/apache/storm/kafka/trident/TridentKafkaClientWordCountNamedTopics.java
+++ b/examples/storm-kafka-client-examples/src/main/java/org/apache/storm/kafka/trident/TridentKafkaClientWordCountNamedTopics.java
@@ -18,12 +18,6 @@
 
 package org.apache.storm.kafka.trident;
 
-import static org.apache.storm.kafka.spout.KafkaSpoutConfig.FirstPollOffsetStrategy.EARLIEST;
-
-import java.util.Arrays;
-import java.util.List;
-import java.util.concurrent.TimeUnit;
-
 import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.apache.storm.Config;
 import org.apache.storm.StormSubmitter;
@@ -39,6 +33,13 @@ import org.apache.storm.kafka.spout.trident.KafkaTridentSpoutOpaque;
 import org.apache.storm.tuple.Fields;
 import org.apache.storm.tuple.Values;
 
+import java.io.Serializable;
+import java.util.Arrays;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+import static org.apache.storm.kafka.spout.KafkaSpoutConfig.FirstPollOffsetStrategy.EARLIEST;
+
 public class TridentKafkaClientWordCountNamedTopics {
     private static final String TOPIC_1 = "test-trident";
     private static final String TOPIC_2 = "test-trident-1";
@@ -48,16 +49,21 @@ public class TridentKafkaClientWordCountNamedTopics {
         return new KafkaTridentSpoutOpaque<>(newKafkaSpoutConfig());
     }
 
-    private static Func<ConsumerRecord<String, String>, List<Object>> JUST_VALUE_FUNC = new Func<ConsumerRecord<String, String>, List<Object>>() {
+    private static Func<ConsumerRecord<String, String>, List<Object>> JUST_VALUE_FUNC = new JustValueFunc();
+
+    /**
+     * Needs to be serializable
+     */
+    private static class JustValueFunc implements Func<ConsumerRecord<String, String>, List<Object>>, Serializable {
         @Override
         public List<Object> apply(ConsumerRecord<String, String> record) {
             return new Values(record.value());
         }
-    };
-    
+    }
+
     protected KafkaSpoutConfig<String,String> newKafkaSpoutConfig() {
         return KafkaSpoutConfig.builder(KAFKA_LOCAL_BROKER, TOPIC_1, TOPIC_2)
-                .setGroupId("kafkaSpoutTestGroup")
+                .setGroupId("kafkaSpoutTestGroup_" + System.nanoTime())
                 .setMaxPartitionFectchBytes(200)
                 .setRecordTranslator(JUST_VALUE_FUNC, new Fields("str"))
                 .setRetry(newRetryService())
@@ -87,7 +93,7 @@ public class TridentKafkaClientWordCountNamedTopics {
 
             System.out.printf("Running with broker_url: [%s], topics: [%s, %s]\n", brokerUrl, topic1, topic2);
 
-            Config tpConf = LocalSubmitter.defaultConfig();
+            Config tpConf = LocalSubmitter.defaultConfig(true);
 
             if (args.length == 4) { //Submit Remote
                 // Producers
@@ -107,11 +113,15 @@ public class TridentKafkaClientWordCountNamedTopics {
                     localSubmitter.submit(topic1Tp, tpConf, KafkaProducerTopology.newTopology(brokerUrl, topic1));
                     localSubmitter.submit(topic2Tp, tpConf, KafkaProducerTopology.newTopology(brokerUrl, topic2));
                     // Consumer
-                    localSubmitter.submit(consTpName, tpConf, TridentKafkaConsumerTopology.newTopology(
-                            localSubmitter.getDrpc(), newKafkaTridentSpoutOpaque()));
+                    try {
+                        localSubmitter.submit(consTpName, tpConf, TridentKafkaConsumerTopology.newTopology(
+                                localSubmitter.getDrpc(), newKafkaTridentSpoutOpaque()));
+                        // print
+                        localSubmitter.printResults(15, 1, TimeUnit.SECONDS);
+                    } catch (Exception e) {
+                        e.printStackTrace();
+                    }
 
-                    // print
-                    localSubmitter.printResults(15, 1, TimeUnit.SECONDS);
                 } finally {
                     // kill
                     localSubmitter.kill(topic1Tp);
--- a/examples/storm-kafka-examples/src/main/java/org/apache/storm/kafka/trident/LocalSubmitter.java
+++ b/examples/storm-kafka-examples/src/main/java/org/apache/storm/kafka/trident/LocalSubmitter.java
@@ -43,10 +43,13 @@ public class LocalSubmitter {
     }
 
     public static Config defaultConfig() {
+        return defaultConfig(false);
+    }
+
+    public static Config defaultConfig(boolean debug) {
         final Config conf = new Config();
         conf.setMaxSpoutPending(20);
-        conf.setMaxTaskParallelism(1);
-        conf.setNumWorkers(1);
+        conf.setDebug(debug);
         return conf;
     }
 
--- a/examples/storm-kafka-examples/src/main/java/org/apache/storm/kafka/trident/TridentKafkaWordCount.java
+++ b/examples/storm-kafka-examples/src/main/java/org/apache/storm/kafka/trident/TridentKafkaWordCount.java
@@ -133,7 +133,7 @@ public class TridentKafkaWordCount implements Serializable {
         TridentKafkaConfig config = new TridentKafkaConfig(hosts, "test");
         config.scheme = new SchemeAsMultiScheme(new StringScheme());
 
-        // Consume new data from the topicÓÓ
+        // Consume new data from the topic
         config.startOffsetTime = kafka.api.OffsetRequest.LatestTime();
         return config;
     }
--- a/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/trident/KafkaTridentSpoutBatchMetadata.java
+++ b/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/trident/KafkaTridentSpoutBatchMetadata.java
@@ -34,8 +34,8 @@ public class KafkaTridentSpoutBatchMetadata<K,V> implements Serializable {
     private static final Logger LOG = LoggerFactory.getLogger(KafkaTridentSpoutBatchMetadata.class);
 
     private TopicPartition topicPartition;  // topic partition of this batch
-    private long firstOffset;   // first offset of this batch
-    private long lastOffset;    // last offset of this batch
+    private long firstOffset;               // first offset of this batch
+    private long lastOffset;                // last offset of this batch
 
     public KafkaTridentSpoutBatchMetadata(TopicPartition topicPartition, long firstOffset, long lastOffset) {
         this.topicPartition = topicPartition;
@@ -74,8 +74,8 @@ public class KafkaTridentSpoutBatchMetadata<K,V> implements Serializable {
 
     @Override
     public String toString() {
-        return "KafkaTridentSpoutBatchMetadata{" +
-                "topicPartition=" + topicPartition +
+        return super.toString() +
+                "{topicPartition=" + topicPartition +
                 ", firstOffset=" + firstOffset +
                 ", lastOffset=" + lastOffset +
                 '}';
--- a/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/trident/KafkaTridentSpoutEmitter.java
+++ b/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/trident/KafkaTridentSpoutEmitter.java
@@ -18,20 +18,6 @@
 
 package org.apache.storm.kafka.spout.trident;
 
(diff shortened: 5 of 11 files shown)


external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/trident/KafkaTridentSpoutEmitter.java
```diff
@@ -161,31 +187,84 @@ private long seek(TopicPartition tp, KafkaTridentSpoutBatchMetadata<K, V> lastBa
         return fetchOffset;
     }
 
-    // returns paused topic partitions
+    // returns paused topic partitions.
     private Collection<TopicPartition> pauseTopicPartitions(TopicPartition excludedTp) {
-        final Set<TopicPartition> pausedTopicPartitions  = new HashSet<>(kafkaConsumer.assignment());
-        LOG.debug("Currently assigned topic partitions [{}]", pausedTopicPartitions);
+        final Set<TopicPartition> pausedTopicPartitions = new HashSet<>(kafkaConsumer.assignment());
+        LOG.trace("Currently assigned topic partitions [{}]", pausedTopicPartitions);
```
HeartSaVioR: Do we need to change this to trace?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/trident/KafkaTridentSpoutEmitter.java
```diff
@@ -47,70 +33,111 @@
...
+        this.pollTimeoutMs = kafkaSpoutConfig.getPollTimeoutMs();
+        this.firstPollOffsetStrategy = kafkaSpoutConfig.getFirstPollOffsetStrategy();
         LOG.debug("Created {}", this);
     }
 
+    /**
+     * Creates instance of this class with default 500 millisecond refresh subscription timer
+     */
+    public KafkaTridentSpoutEmitter(KafkaTridentSpoutManager<K,V> kafkaManager, TopologyContext topologyContext) {
+        this(kafkaManager, topologyContext, new Timer(500,
```
harshach: do we want make the timer here configurable instead of 500

HeartSaVioR: This is for backward compatible. Another constructor which receives timer is available.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/trident/KafkaTridentSpoutEmitter.java
```diff
@@ -47,70 +33,111 @@
...
-            pausedTopicPartitions = pauseTopicPartitions(topicPartition);
+        if (assignments == null || !assignments.contains(currBatchPartition.getTopicPartition())) {
+            LOG.debug("SKIPPING processing batch: [transaction = {}], [currBatchPartition = {}], [lastBatchMetadata = {}], " +
+                            "[collector = {}] because it is not assigned [{}] to consumer instance [{}] of consumer group [{}]",
+                    tx, currBatchPartition, lastBatch, collector, assignments, kafkaConsumer,
+                    kafkaManager.getKafkaSpoutConfig().getConsumerGroupId());
+        } else {
+            try {
+                // pause other topic partitions to only poll from current topic partition
+                pausedTopicPartitions = pauseTopicPartitions(currBatchTp);
```
harshach: how this will work in case of multiple executors running the spout. It looks like we are making sure there is only executor/kafka-consumer that can poll the topic partitions. It looks like the intent here is to keep data from 1 partition per batch.

pom.xml
```diff
@@ -267,7 +267,7 @@
         <storm.kafka.artifact.id>kafka_2.10</storm.kafka.artifact.id>
 
         <!-- kafka version used by new storm-kafka-client spout code -->
-        <storm.kafka.client.version>0.10.0.0</storm.kafka.client.version>
+        <storm.kafka.client.version>0.10.1.1</storm.kafka.client.version>
```
harshach: Lets keep the version to 0.10.0.1 here. Its better to support lowest version as much as possible. As of now old clients can connect to new brokers but not other way round.

harshach:
@hmcl overall looks good to me. Left few questions. @hmcl has some good points around the topic names.

examples/storm-kafka-examples/src/main/java/org/apache/storm/kafka/trident/TridentKafkaWordCount.java
```diff
@@ -95,7 +95,8 @@ public static void main(String[] args) throws Exception {
                 localSubmitter.submit(prodTpName, tpConf, KafkaProducerTopology.newTopology(zkBrokerUrl[1], topicName));
                 // Consumer
                 localSubmitter.submit(consTpName, tpConf, TridentKafkaConsumerTopology.newTopology(localSubmitter.getDrpc(),
-                        new TransactionalTridentKafkaSpout(newTridentKafkaConfig(zkBrokerUrl[0]))));
+                        new OpaqueTridentKafkaSpout(newTridentKafkaConfig(zkBrokerUrl[0]))));
+//                        new TransactionalTridentKafkaSpout(newTridentKafkaConfig(zkBrokerUrl[0]))));
```
hmcl: I meant to leave it such that it's pretty trivial, and also brings awareness, of using this test to test both Opaque and Transactional Trident. I think it's an overdo to write code supporting both (e.g. with a CLI option), that's why I left this as an easy way to do so. I can put a comment saying why this code is there. If you still think it should be removed, I can do it. Let me know.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/trident/KafkaTridentSpoutEmitter.java
```diff
@@ -47,70 +33,111 @@
...
+        final TopicPartition currBatchTp = currBatchPartition.getTopicPartition();
+        final Set<TopicPartition> assignments = kafkaConsumer.assignment();
         KafkaTridentSpoutBatchMetadata<K, V> currentBatch = lastBatch;
         Collection<TopicPartition> pausedTopicPartitions = Collections.emptySet();
 
-        try {
-            // pause other topic partitions to only poll from current topic partition
-            pausedTopicPartitions = pauseTopicPartitions(topicPartition);
+        if (assignments == null || !assignments.contains(currBatchPartition.getTopicPartition())) {
+            LOG.debug("SKIPPING processing batch: [transaction = {}], [currBatchPartition = {}], [lastBatchMetadata = {}], " +
```
hmcl: I agree with WARN. I believe the the code can only get here if there is a rebalance while the tasks are being assigned. This logic will guarantee that we retry.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/trident/KafkaTridentSpoutEmitter.java
hmcl: Done

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/trident/KafkaTridentSpoutBatchMetadata.java
```diff
@@ -74,8 +74,8 @@ public TopicPartition getTopicPartition() {
 
     @Override
     public String toString() {
-        return "KafkaTridentSpoutBatchMetadata{" +
-                "topicPartition=" + topicPartition +
+        return super.toString() +
```
hmcl: All this is doing is that the method `toString()` instead of printing `KafkaTridentSpoutBatchMetadata` it will print `org.apache.storm.kafka.spout.trident.KafkaTridentSpoutBatchMetadata@6b74f986`. This was useful while I was debugging and writing the code because the code is multi threaded for multiple executors, and I needed to see which object instance (e.g `@6b74f986`) was being passed around. I can revert it back, but for me there was benefit in having it like this.

examples/storm-kafka-client-examples/src/main/java/org/apache/storm/kafka/trident/TridentKafkaClientWordCountNamedTopics.java
```diff
@@ -107,8 +113,12 @@ protected void run(String[] args) throws AlreadyAliveException, InvalidTopologyE
...
                     // Consumer
+                    try {
                     localSubmitter.submit(consTpName, tpConf, TridentKafkaConsumerTopology.newTopology(
                             localSubmitter.getDrpc(), newKafkaTridentSpoutOpaque()));
+                    } catch (Exception e) {
+                        e.printStackTrace();
+                    }
 
                     // print
                     localSubmitter.printResults(15, 1, TimeUnit.SECONDS);
```
hmcl: Done

examples/storm-kafka-client-examples/src/main/java/org/apache/storm/kafka/trident/TridentKafkaClientWordCountNamedTopics.java
```diff
@@ -107,8 +113,12 @@ protected void run(String[] args) throws AlreadyAliveException, InvalidTopologyE
                     localSubmitter.submit(topic1Tp, tpConf, KafkaProducerTopology.newTopology(brokerUrl, topic1));
                     localSubmitter.submit(topic2Tp, tpConf, KafkaProducerTopology.newTopology(brokerUrl, topic2));
                     // Consumer
+                    try {
                     localSubmitter.submit(consTpName, tpConf, TridentKafkaConsumerTopology.newTopology(
```
hmcl: Done

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/trident/KafkaTridentSpoutEmitter.java
```diff
@@ -196,8 +275,8 @@ public void close() {
 
     @Override
     public String toString() {
-        return "KafkaTridentSpoutEmitter{" +
-                ", kafkaManager=" + kafkaManager +
+        return super.toString() +
```
hmcl: Same comment as above.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/trident/KafkaTridentSpoutManager.java
```diff
@@ -107,9 +90,25 @@ Fields getFields() {
 
     @Override
     public String toString() {
-        return "KafkaTridentSpoutManager{" +
-                "kafkaConsumer=" + kafkaConsumer +
+        return super.toString() +
```
hmcl: Same comment as above.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/trident/KafkaTridentSpoutOpaque.java
```diff
@@ -79,8 +72,8 @@ public Fields getOutputFields() {
 
     @Override
     public String toString() {
-        return "KafkaTridentSpoutOpaque{" +
-                "kafkaManager=" + kafkaManager +
+        return super.toString() +
```
hmcl: Same comment as above.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/trident/KafkaTridentSpoutOpaqueCoordinator.java
```diff
@@ -57,8 +57,8 @@ public void close() {
 
     @Override
     public String toString() {
-        return "KafkaTridentSpoutOpaqueCoordinator{" +
-                "kafkaManager=" + kafkaManager +
+        return super.toString() +
```
hmcl: Same comment as above.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/trident/KafkaTridentSpoutTopicPartition.java
```diff
@@ -43,7 +43,7 @@ public TopicPartition getTopicPartition() {
 
     @Override
     public String getId() {
-        return topicPartition.topic() + "/" + topicPartition.partition();
+        return topicPartition.topic() + "_" + topicPartition.partition();
```
hmcl: so which one do you think it should be? This is new code that I wrote, so it should be a matter of choice. Before this change I had **/** but I felt that there was no need to create a new zookeeper level, hence I changed it to **_**. **_** is what exists in the old kafka consumer code. Please take a look at method `org.apache.storm.kafka.Partition#getId()` and let me know which one we should go with. Currently the code works using either **/** or **_**

examples/storm-kafka-examples/src/main/java/org/apache/storm/kafka/trident/TridentKafkaWordCount.java
HeartSaVioR: Example topology doesn't really need to provide cover features exhaustively. Users still can see TransactionalTridentKafkaSpout is available, even not looking at this class. If users can't find, IMHO it's a documentation issue.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/trident/KafkaTridentSpoutBatchMetadata.java
HeartSaVioR: Thanks for the details. I'm OK to leave this.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/trident/KafkaTridentSpoutTopicPartition.java
HeartSaVioR: Yes it will work if no one is willing to extract the information via parsing the return value. OK to leave as it is, but if possible I'd like to avoid the possible confusing case. 

If you're worrying about creating redundant znode because of '/', how about choosing the character which is not legal char for kafka topic but legal char for znode name, for example, `@`?

examples/storm-kafka-examples/src/main/java/org/apache/storm/kafka/trident/LocalSubmitter.java
```diff
@@ -43,10 +43,15 @@ public static LocalSubmitter newInstance() {
...
     public static Config defaultConfig() {
+        return defaultConfig(false);
+    }
+
+    public static Config defaultConfig(boolean debug) {
         final Config conf = new Config();
         conf.setMaxSpoutPending(20);
-        conf.setMaxTaskParallelism(1);
-        conf.setNumWorkers(1);
+//        conf.setMaxTaskParallelism(2);
```
hmcl: Done

examples/storm-kafka-examples/src/main/java/org/apache/storm/kafka/trident/TridentKafkaWordCount.java
hmcl: Done

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/trident/KafkaTridentSpoutEmitter.java
hmcl: I think it will work fine. Each executor will have one instance of `KakfkaConsumer`. The org.apache.storm.trident.spout.OpaquePartitionedTridentSpoutExecutor.Emitter class has some logic that emits one batch per topic-partition. I believe the same happens in the Kafka Spout with the old consumer. In case I am misunderstanding, can you please elaborate a bit more. Thanks.

pom.xml
hmcl: Done

hmcl:
@harshach I should add this to master branch as well, right ?

hmcl:
@harshach @HeartSaVioR I believe I have addressed all your review comments. I temporarily left one commit just with the code review diff. I will squash it once we agree the patch is good to go. Can you please do one last pass. Thanks.

hmcl:
@harshach @HeartSaVioR can you please take a look. Thanks.

harshach:
+1.

harshach:
@hmcl yes we need one for master as well.

HeartSaVioR:
@hmcl Thanks for the update. +1 from me. There's a conflict while cherry-picking into master, so we might need to have following PR for master.

hmcl:
@harshach @HeartSaVioR thanks for the review. I have squashed the commits. It should be good to go.

HeartSaVioR:
+1 again.

PR #1904, opened 2017-01-31
STORM-2321 Handle blobstore zk key deletion in KeySequenceNumber

HeartSaVioR:
* If NoNodeException is thrown in getKeySequenceNumber, treat it as KeyNotFoundException
* Change callers to handle KeyNotFoundException accordingly

Race condition with a deletion is rare case, so instead of adding some locks to prevent race condition, it just allows error and handle it accordingly.

For 1.x branch: #1905


HeartSaVioR:
@redsanket I'd really appreciated it if you can take a look. Thanks in advance.

harshach:
+1

PR #1905, opened 2017-01-31
STORM-2321 Handle blobstore zk key deletion in KeySequenceNumber (1.x)

HeartSaVioR:
* If NoNodeException is thrown in getKeySequenceNumber, treat it as KeyNotFoundException
* Change callers to handle KeyNotFoundException accordingly

This can be easily cherry picked to 1.0.x branch.

For master branch: #1904


HeartSaVioR:
@harshach Could you review this as well? Thanks in advance.

PR #1834, opened 2016-12-18
STORM-2244 Some shaded jars doesn't exclude dependency signature files

HeartSaVioR:
* Add filter option to all of shade plugin configurations

PR #1913, opened 2017-02-02
STORM-2244 Some shaded jars doesn't exclude dependency signature files (1.x)

HeartSaVioR:
* Add filter option to all of shade plugin configurations

PR for master branch (#1834) is already merged.

@harshach Could you review this as well? Thanks!


harshach:
+1. looks good.

PR #1914, opened 2017-02-02
STORM-2334 - Join Bolt implementation with unit tests

roshannaik:
For details see : 
https://issues.apache.org/jira/browse/STORM-2334


storm-core/src/jvm/org/apache/storm/bolt/JoinBolt.java
```diff
@@ -0,0 +1,428 @@
...
+        return this;
+    }
+
+    /**
+     * Performs inner Join.
+     *  SQL    :   from priorStream inner join newStream on newStream.key = priorStream.key1
+     *  same as:   new WindowedQueryBolt(priorStream,key1). join(newStream, key, priorStream);
+     *
+     *  Note: priorStream must be previously joined.
+     *    Valid ex:    new WindowedQueryBolt(s1,k1). join(s2,k2, s1). join(s3,k3, s2);
```
raghavgautam: It would be probably better if the user could specify joins by saying:
new WindowedQueryBolt(s1,k1). join(s2,k2). join(s3,k3);
(note missing s1, s2 parameters in the join method)
This will be less error prone and more in line with SQL expectation where user specifies:
table1.key1 = table2.key2

storm-core/src/jvm/org/apache/storm/bolt/JoinBolt.java
```diff
@@ -0,0 +1,428 @@
...
+     *
+     *  Note: priorStream must be previously joined
+     *    Valid ex:    new WindowedQueryBolt(s1,k1). leftJoin(s2,k2, s1). leftJoin(s3,k3, s2);
+     *    Invalid ex:  new WindowedQueryBolt(s1,k1). leftJoin(s3,k3, s2). leftJoin(s2,k2, s1);
+     */
+    public JoinBolt leftJoin(String newStream, String key, String priorStream) {
+        return join_common(newStream, key, priorStream, JoinType.LEFT);
+    }
+
+    private JoinBolt join_common(String newStream, String key, String priorStream, JoinType joinType) {
```
harshach: can we keep the method names to camelcase like joinCommon

storm-core/src/jvm/org/apache/storm/bolt/JoinBolt.java
roshannaik: It is necessary to specify which stream s3 should be joined with (s2 or s1?) .... as the results will differ accordingly.

storm-core/src/jvm/org/apache/storm/bolt/JoinBolt.java
roshannaik: indeed.

ptgoetz:
@roshannaik Can you add some documentation for this similar to what we have for Windowing?

roshannaik:
@ptgoetz yes ... I was thinking about that myself... thinking of creating a Join.md under the top level docs directory.

storm-core/src/jvm/org/apache/storm/bolt/JoinBolt.java
raghavgautam: Discussed this offline with @roshannaik , this is necessary when we have left outer join. Consider a join of tables T1, T2, T3. If join of T1 and T2 is left outer join it can lead to intermediate table which looks like:

T1.K1| T2.K1
-----|-------
 Y     | NULL 
Now consider the case, when we have to do inner join with T3.K1 (with value Y) the resulting table might change depending on whether it is join on T1.K1 or T2.K1

docs/Joins.md
```diff
@@ -0,0 +1,102 @@
...
+title: Joining Streams in Storm Core
+layout: documentation
+documentation: true
+---
+
+Storm core supports joining multiple data streams into one with the help of `JoinBolt`.
+`JoinBolt` is a Windowed bolt, i.e. it waits for the configured window duration to match up the
+tuples among the streams being joined. This helps align the streams within the Window boundary.
+
+Each of `JoinBolt`'s incoming data streams must be Fields Grouped on a single field. A stream
```
arunmahadevan: How is it enforced ? 
RuntimeException at topology construction time ?
Best if the field groupings on the join keys are automatically handled by the api than relying on user to define it correctly.

docs/Joins.md
```diff
@@ -0,0 +1,102 @@
...
+```
+
+This could be expressed using `JoinBolt` over 4 similarly named streams as: 
+
+```java
+new JoinBolt(JoinBolt.Selector.STREAM, "stream1", "key1")               // from stream1  
+                            .join     ("stream2", "userId",  "stream1") // join      stream2  on stream2.userId = stream1.key1
+                            .join     ("stream3", "key3",    "stream2") // join      stream3  on stream3.key3   = stream2.userId   
+                            .leftjoin ("stream4", "key4",    "stream3") // left join stream4  on stream4.key4   = stream3.key3
+                            .select("userId, key4, key2, key3")         // chose output fields
```
arunmahadevan: 1. why is it not a `List<String>`
2. How does one select keys having the same name from different streams (e.g. stream2.key1, stream3.key1) ?

docs/Joins.md
```diff
@@ -0,0 +1,102 @@
...
+from    stream1 
+join       stream2  on stream2.userId =  stream1.key1
+join       stream3  on stream3.key3   =  stream2.userId
+left join  stream4  on stream4.key4   =  stream3.key3
+```
+
+This could be expressed using `JoinBolt` over 4 similarly named streams as: 
+
+```java
+new JoinBolt(JoinBolt.Selector.STREAM, "stream1", "key1")               // from stream1
```
arunmahadevan: How do you ensure "stream1" exists ? i.e. some spout or bolt should have declared "stream1" and JoinBolt should subscribe to that spout/bolt's output via fields grouping on the join key. Assume it would throw RuntimeException at topology construction time.

docs/Joins.md
```diff
@@ -0,0 +1,102 @@
...
+join       stream2  on stream2.userId =  stream1.key1
+join       stream3  on stream3.key3   =  stream2.userId
+left join  stream4  on stream4.key4   =  stream3.key3
+```
+
+This could be expressed using `JoinBolt` over 4 similarly named streams as: 
+
+```java
+new JoinBolt(JoinBolt.Selector.STREAM, "stream1", "key1")               // from stream1  
+                            .join     ("stream2", "userId",  "stream1") // join      stream2  on stream2.userId = stream1.key1
```
arunmahadevan: How to Join on a combination of fields (e.g. `s1.f1 == s2.f1 && s1.f2 == s2.f2`) ? Similarly on partial values of fields.

The above could be done if users are writing custom bolts for join, see "Join" section - https://github.com/apache/storm/blob/master/docs/Common-patterns.md, so a good idea to support it.

docs/Joins.md
roshannaik: Absence of the definition of a specific stream upstream is equivalent to not receiving any data from that stream from the Bolt. Conversely, if a upstream component sends it data on a stream that the bolt is not aware of, it will ignore that stream. Bolt doesn't examine the entire topology definition scanning for incorrect wiring.

docs/Joins.md
roshannaik: answered in next comment...

docs/Joins.md
roshannaik: 1. Conversion it to list is done internally.  Chose CSV string instead of list to make it easier for end users (less typing) ... without loss of functionality. 

2. I think .. as of now its not supported.. let me take a look and add it if not there already.

roshannaik:
@arunmahadevan  yes..  good idea... i will add an example into storm-starter.

docs/Joins.md
roshannaik: Yes good question. As of now, grouping on single key and consequently joining on 1 key only is supported ...  noted in the doc, 

To get the effect of joining on multiple fields you need to combine two/more fields into one, somewhere upstream. Then you can join on that merged field. Perhaps worth mentioning this in docs.

To be more inline with SQL abilities, I started supporting streams grouped on multiple fields for joins initially, but later scaled it back to single key. Saw a few issues with opening up grouping and joins on multiple fields. Let me try to summarize...

It easily gives users the wrong impression that they can do many SQL-type things .. like

Assuming S1 is grouped on both f1 & f2:
```
from s1
join s2   s2.f1 == s1.f1 && s2.f2 == s1.f2     // uses both f1 & f2 form s1
join s3   s3.f1 == s1.f1 && s3.f2 == s2.f2     // one field from s1 and another from s2
```
which cannot work correctly due to FG.

Another example is  grouping S1 on f1 and f2 and then trying to do this...
```
from s1
join s2   s2.f1 == s1.f1    // use s1.f1 
join s3   s3.f1 == s1.f2   // but s1.f2 here
```

**In short:** The extent to which multi field joins can be correctly supported is limited to whatever can be achieved by merging those fields into one before joining. Going beyond it complicated the interface and the internal checking... in addition to easily setting up wrong user expectations.

The downside of current approach is an additional step to merge the fields in such cases. Thats the trade off I settled for.

docs/Joins.md
arunmahadevan: based on the next comment, checks are not  done and applying the right grouping is up to the user. 
IMO, adding as many validations as possible and automatically handling the groupings would make it less error prone. But this may require changes in the topology builder, so for now may be just add some relevant code that shows the fields grouping (`builder.setBolt(joinBolt).fieldsGrouping(..)`) in the docs.

docs/Joins.md
arunmahadevan: may be for now just add the relevant code in the docs to show the upstream component declaring the stream to make it clearer.

docs/Joins.md
arunmahadevan: the sql like syntax is not supported correct? So assume joining on subset of join fields (the "not ok" case in your example) is not possible with the current api. 
Its only about allowing users to have multiple fields as the join key without the extra merge step. May be accepting a type like `JoinKey` than string as the key parameter in the api would make it clear.

docs/Joins.md
roshannaik: sure... good idea.

docs/Joins.md
roshannaik: yes.

docs/Joins.md
roshannaik: Yes, not supported. Joining on a subset of grouping fields is not supported as it **will** lead to wrong results in a streaming system... unless you restrict it to 1 running instance of the Join bolt.  

This is not a limitation of the JoinBolt per say, it is a limitation that it inherits from the nature of in which streaming data is routed.  Its a bit tricky to express, but let me give it a try:


When we join with f1, the system **must** route the tuples based on hash(f1) to the right Join bolt instance.  If we want to join on both f1&f2, then routing **must** be based on hash(f1,f2).  
Basically, if the join fields and the grouping/routing fields are not same, the bolt instance will not receive all the right tuples for the join.

So there is an inherent contradiction when you try to route on hash(f1,f2),  then try to join on anything but those two fields.

arunmahadevan: @roshannaik may be the earlier comment was not clear. Since the join key is also the group by field in the api, choosing a subset of those keys in the join condition is not possible.

E.g.
```java
new JoinBolt(JoinBolt.Selector.STREAM, "stream1", new Fields("key1", "key2")) 
         // here its not possible to join on stream1.key1 or stream1.key2 separately
        .join("stream2", new Fields("key3", "key4"), "stream1");
```

roshannaik: right.

roshannaik:
@arunmahadevan  i have update the PR with
- Updated docs
- Added sample topo
- Added support to qualify keys with stream names for disambiguation
- Added more UT

Have addressed all the outstanding review comments.

docs/Joins.md
```diff
@@ -0,0 +1,126 @@
...
+                    .withTumblingWindow( new Duration(10, TimeUnit.MINUTES) ) ;
+
+topoBuilder.setBolt("joiner", jbolt, 1)
+            .fieldsGrouping("spout1", new Fields("key1") )
+            .fieldsGrouping("spout2", new Fields("userId") )
+            .fieldsGrouping("spout3", new Fields("key3") )
+            .fieldsGrouping("spout4", new Fields("key4") );
+```
+
+The bolt constructor takes two arguments. The 1st argument introduces the data from `spout1`'
```
arunmahadevan: extra char (') after `spout1`

docs/Joins.md
```diff
@@ -0,0 +1,126 @@
...
+to be the first stream and specifies that it will always use field `key1` when joining this with the others streams.
+The name of the component specified must refer to the spout or bolt that is directly connected to the Join bolt. 
+Here data received from `spout1` must be fields grouped on `key1`. Similarly, each of the `leftJoin()` and `join()` method
+calls introduce a new stream along with the field to use for the join. As seen in above example, the same FieldsGrouping
+requirement applies to these streams as well. The 3rd argument to the join methods refers to another stream with which
+to join.
+
+The `select()` method is used to specify the output fields. The argument to `select` is a comma separated list of fields.
+Individual field names can be prefixed with a stream name to disambiguate between the same field name occurring in
+multiple streams as follows:  `.select("spout3:key3, spout4:key3")`. Nested tuple types are supported if the
```
arunmahadevan: why not **dot** syntax which is closer to the sql syntax, for example `select("spout3.key3, spout4.key4.nested1, key5")` ? Is the syntax same if stream name is used instead of component name?

docs/Joins.md
```diff
@@ -0,0 +1,126 @@
...
+1. Stream names must be introduced (in constructor or as 1st arg to various join methods) before being referred
+to (in the 3rd argument of the join methods). Forward referencing of stream names, as shown below, is not allowed:
+
+```java
+new JoinBolt( "spout1", "key1")                 
+  .join     ( "spout2", "userId",  "spout3") //not allowed. spout3 not yet introduced
+  .join     ( "spout3", "key3",    "spout1")
+```
+
+2. Internally, the joins will be performed in the order expressed by the user.
```
arunmahadevan: Its showing up as 1. in the rich diff mode.

docs/Joins.md
roshannaik: ya... i started with the dot syntax... then realized that it still leaves ambiguity .. if there is a field and stream with same name. I wanted to avoid putting code to guess whether the first part is a stream name or a key name. So its just to avoid that corner case. 

Standard SQL does not have this issue as it doesn't allow a mix of table names and nested field names in  a simple x.y.z format.

docs/Joins.md
roshannaik: fixed. thanks.

docs/Joins.md
roshannaik: fixed. thanks.

roshannaik:
Revised the PR with the fixes. Thanks.

roshannaik:
@arunmahadevan ... PR has been revised. thanks.

arunmahadevan:
+1, will wait for a day to see if there are any comments from other reviewers and will merge this after that.

ptgoetz:
+1

roshannaik:
@arunmahadevan can you please commit this ?

arunmahadevan:
Thanks @roshannaik merged to master and 1.x-branch. You can close this PR.

PR #1916, opened 2017-02-03
fixed storm-kafka-client test failed

lukess:
in order to finish tests for storm-kafka-client


srdo:
It looks like Fields has an equals() now to do this https://github.com/apache/storm/blob/1.x-branch/storm-core/src/jvm/org/apache/storm/tuple/Fields.java

lukess:
thanks @srdo let me re pr

PR #1922, opened 2017-02-04
STORM-2344 Flux YAML File Viewer for Nimbus UI

ambud:
Adding capabilities to visualize the Storm Flux YAML files before submitting the topology would be helpful for developers and operations.


storm-core/src/ui/public/flux.html
```diff
@@ -0,0 +1,139 @@
+<html>
```
vesense: Missing apache license. And I guess we may need a link from storm ui homepage to flux.html.

ambud: @vesense I thought someone will ask for that :D 

Just added both.

ambud: Also, should we plan on back porting this to 0.10.x and 1.x?

HeartSaVioR: I think @vesense means that Storm UI index page should have a link to go to flux page. You're adding reverse link, which is already missing.

And I think we might want to port back to 1.x but not necessary to 1.0.x. Let's keep the bugfix version line.

ambud: Sorry I misunderstood. @HeartSaVioR not sure what location in the homepage do we want to put the forward link? 

One option could be to skip the link in homepage and provide it via Ambari View.

HeartSaVioR: No it shouldn't be only for Ambari. I'll see where's good location to put. Welcome any ideas.

HeartSaVioR: If we can have icon then we can place it to the left side of the upper right magnifier icon. If we can't, Text link `[Flux Viewer]` or so may be fine given that we can replace it anytime if we find better alternatives.

vesense: I have the same idea as @HeartSaVioR :smile:

ambud: Added a link on the top right hand side after the username place holder. Here's the icon:
[Icon](https://github.com/ambud/storm/blob/af8ed339cb3896100dabe3619c0fba2d19165dff/storm-core/src/ui/public/images/flux.png)

Y for YAML and also for Flux Capacitor from Back to the future 💯 

Please let me know if this works.

HeartSaVioR: Not sure which is better between F (for Flux) and Y (for Yaml). I guess the question could be changed to is flux yaml viewer **flux** yaml viewer, or flux **yaml** viewer. What do you think?

HeartSaVioR: And posting screenshot to the comment would help us to see.

ambud: storm-ui

HeartSaVioR:
@ambud 
We need to update LICENSE file to reflect newly added js files.
https://github.com/apache/storm/blob/master/LICENSE

Could you update it?

HeartSaVioR:
And some javascript files like esprima.js are not minimized. Is there any reason not to do?

ambud:
- Added license
- Squashed commits. 

Couldn't find a min.js for dagre

HeartSaVioR:
js files are licensed under BSD-2 and MIT, seems OK.
Regarding min.js, I think you're saying about cytoscape-dagre.js, which is under 200 lines so it would be OK.

+1 from me. Thanks!

PR #1924, opened 2017-02-05
STORM-2343: New Kafka spout can stop emitting tuples if more than maxUncommittedOffsets tuples fail at once

srdo:
This builds on https://github.com/apache/storm/pull/1832, sorry for the mixed diff

This makes the spout emit failed tuples even when numUncommittedOffsets has reached maxUncommittedOffsets. Previously the spout would fail to progress in that case. I haven't really been able to enforce maxUncommittedOffsets strictly without having to request extra messages from Kafka and throwing them away. The consumer doesn't allow the spout to limit how many tuples it requests dynamically, and I'd prefer that the spout doesn't truncate the list of records it receives.

Instead, maxUncommittedOffsets is now a soft cap on the number of tuples that can be emitted before some must be committed. In some cases (e.g. 10 partitions with 1 failed message emitted on each), the spout will exceed the limit. It should never be by more than 1 maxPollRecords size though.

This also makes KafkaSpoutRetryExponentialBackoff use Storm Time, and it fixes a bug where messages could be lost if they were scheduled for retry at the same timestamp. It also fixes double counting failed tuples in numUncommittedOffsets when retrying, since the counter isn't decreased when the tuple is scheduled for retry.

maxPollRecords is now capped by maxUncommittedOffsets, since if maxPollRecords is higher the spout will exceed the limit on any poll where Kafka can return maxPollRecords messages.

### DIFF
--- a/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
+++ b/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
@@ -31,6 +31,7 @@ import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 
@@ -79,7 +80,7 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
     private transient Map<TopicPartition, OffsetManager> acked;         // Tuples that were successfully acked. These tuples will be committed periodically when the commit timer expires, or after a consumer rebalance, or during close/deactivate
     private transient Set<KafkaSpoutMessageId> emitted;                 // Tuples that have been emitted but that are "on the wire", i.e. pending being acked or failed. Not used if it's AutoCommitMode
     private transient Iterator<ConsumerRecord<K, V>> waitingToEmit;     // Records that have been polled and are queued to be emitted in the nextTuple() call. One record is emitted per nextTuple()
-    private transient long numUncommittedOffsets;                       // Number of offsets that have been polled and emitted but not yet been committed. Not used if it's AutoCommitMode
+    private transient long numUncommittedOffsets;                       // Number of offsets that have been polled and emitted but not yet been committed. Not used if auto commit mode is enabled.
     private transient Timer refreshSubscriptionTimer;                   // Triggers when a subscription should be refreshed
     private transient TopologyContext context;
 
@@ -250,9 +251,13 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
 
     private boolean poll() {
         final int maxUncommittedOffsets = kafkaSpoutConfig.getMaxUncommittedOffsets();
-        final boolean poll = !waitingToEmit()
-                && (numUncommittedOffsets < maxUncommittedOffsets || consumerAutoCommitMode);
-
+        final int readyMessageCount = retryService.readyMessageCount();
+        final boolean poll = !waitingToEmit() &&
+            //Check that the number of uncommitted, nonretriable tuples is less than the maxUncommittedOffsets limit
+            //Accounting for retriable tuples this way still guarantees that the limit is followed on a per partition basis, and prevents locking up the spout when there are too many retriable tuples
+            (numUncommittedOffsets - readyMessageCount < maxUncommittedOffsets ||
+            consumerAutoCommitMode);
+        
         if (!poll) {
             if (waitingToEmit()) {
                 LOG.debug("Not polling. Tuples waiting to be emitted. [{}] uncommitted offsets across all topic partitions", numUncommittedOffsets);
@@ -290,15 +295,11 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
     }
 
     private void doSeekRetriableTopicPartitions() {
-        final Set<TopicPartition> retriableTopicPartitions = retryService.retriableTopicPartitions();
+        final Map<TopicPartition, Long> retriableTopicPartitions = retryService.earliestRetriableOffsets();
 
-        for (TopicPartition rtp : retriableTopicPartitions) {
-            final OffsetAndMetadata offsetAndMeta = acked.get(rtp).findNextCommitOffset();
-            if (offsetAndMeta != null) {
-                kafkaConsumer.seek(rtp, offsetAndMeta.offset() + 1);  // seek to the next offset that is ready to commit in next commit cycle
-            } else {
-                kafkaConsumer.seek(rtp, acked.get(rtp).getCommittedOffset() + 1);    // Seek to last committed offset
-            }
+        for (Entry<TopicPartition, Long> retriableTopicPartitionAndOffset : retriableTopicPartitions.entrySet()) {
+            //Seek directly to the earliest retriable message for each retriable topic partition
+            kafkaConsumer.seek(retriableTopicPartitionAndOffset.getKey(), retriableTopicPartitionAndOffset.getValue());
         }
     }
 
@@ -318,7 +319,6 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
     private boolean emitTupleIfNotEmitted(ConsumerRecord<K, V> record) {
         final TopicPartition tp = new TopicPartition(record.topic(), record.partition());
         final KafkaSpoutMessageId msgId = new KafkaSpoutMessageId(record);
-
         if (acked.containsKey(tp) && acked.get(tp).contains(msgId)) {   // has been acked
             LOG.trace("Tuple for record [{}] has already been acked. Skipping", record);
         } else if (emitted.contains(msgId)) {   // has been emitted and it's pending ack or fail
--- a/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutConfig.java
+++ b/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutConfig.java
@@ -269,12 +269,9 @@ public class KafkaSpoutConfig<K, V> implements Serializable {
         
         /**
          * The maximum number of records a poll will return.
-         * Will only work with Kafka 0.10.0 and above.
          */
         public Builder<K,V> setMaxPollRecords(int records) {
-            //to avoid issues with 0.9 versions that technically still work
-            // with this we do not use ConsumerConfig.MAX_POLL_RECORDS_CONFIG
-            return setProp("max.poll.records", records);
+            return setProp(ConsumerConfig.MAX_POLL_RECORDS_CONFIG, records);
         }
         
         //Security Related Configs
@@ -330,11 +327,12 @@ public class KafkaSpoutConfig<K, V> implements Serializable {
             this.offsetCommitPeriodMs = offsetCommitPeriodMs;
             return this;
         }
-        
+
         /**
          * Defines the max number of polled offsets (records) that can be pending commit, before another poll can take place.
          * Once this limit is reached, no more offsets (records) can be polled until the next successful commit(s) sets the number
-         * of pending offsets bellow the threshold. The default is {@link #DEFAULT_MAX_UNCOMMITTED_OFFSETS}.
+         * of pending offsets below the threshold. The default is {@link #DEFAULT_MAX_UNCOMMITTED_OFFSETS}.
+         * Note that this limit can in some cases be exceeded, but no partition will exceed this limit by more than maxPollRecords - 1.
          * @param maxUncommittedOffsets max number of records that can be be pending commit
          */
         public Builder<K,V> setMaxUncommittedOffsets(int maxUncommittedOffsets) {
--- a/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutRetryExponentialBackoff.java
+++ b/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutRetryExponentialBackoff.java
@@ -25,15 +25,18 @@ import org.slf4j.LoggerFactory;
 import java.io.Serializable;
 import java.util.Collection;
 import java.util.Comparator;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.Map;
 import java.util.Set;
 import java.util.TreeSet;
 import java.util.concurrent.TimeUnit;
+import org.apache.storm.utils.Time;
 
 /**
  * Implementation of {@link KafkaSpoutRetryService} using the exponential backoff formula. The time of the nextRetry is set as follows:
- * nextRetry = failCount == 1 ? currentTime + initialDelay : currentTime + delayPeriod^(failCount-1)    where failCount = 1, 2, 3, ...
+ * nextRetry = failCount == 1 ? currentTime + initialDelay : currentTime + delayPeriod*2^(failCount-1)    where failCount = 1, 2, 3, ...
  * nextRetry = Min(nextRetry, currentTime + maxDelay)
  */
 public class KafkaSpoutRetryExponentialBackoff implements KafkaSpoutRetryService {
@@ -54,7 +57,14 @@ public class KafkaSpoutRetryExponentialBackoff implements KafkaSpoutRetryService
     private static class RetryEntryTimeStampComparator implements Serializable, Comparator<RetrySchedule> {
         @Override
         public int compare(RetrySchedule entry1, RetrySchedule entry2) {
-            return Long.valueOf(entry1.nextRetryTimeNanos()).compareTo(entry2.nextRetryTimeNanos());
+            int result = Long.valueOf(entry1.nextRetryTimeNanos()).compareTo(entry2.nextRetryTimeNanos());
+            
+            if(result == 0) {
+                //TreeSet uses compareTo instead of equals() for the Set contract
+                //Ensure that we can save two retry schedules with the same timestamp
+                result = entry1.hashCode() - entry2.hashCode();
+            }
+            return result;
         }
     }
 
@@ -62,13 +72,13 @@ public class KafkaSpoutRetryExponentialBackoff implements KafkaSpoutRetryService
         private final KafkaSpoutMessageId msgId;
         private long nextRetryTimeNanos;
 
-        public RetrySchedule(KafkaSpoutMessageId msgId, long nextRetryTime) {
+        public RetrySchedule(KafkaSpoutMessageId msgId, long nextRetryTimeNanos) {
             this.msgId = msgId;
-            this.nextRetryTimeNanos = nextRetryTime;
+            this.nextRetryTimeNanos = nextRetryTimeNanos;
             LOG.debug("Created {}", this);
         }
 
-        public void setNextRetryTime() {
+        public void setNextRetryTimeNanos() {
             nextRetryTimeNanos = nextTime(msgId);
             LOG.debug("Updated {}", this);
         }
@@ -81,7 +91,7 @@ public class KafkaSpoutRetryExponentialBackoff implements KafkaSpoutRetryService
         public String toString() {
             return "RetrySchedule{" +
                     "msgId=" + msgId +
-                    ", nextRetryTime=" + nextRetryTimeNanos +
+                    ", nextRetryTimeNanos=" + nextRetryTimeNanos +
                     '}';
         }
 
@@ -96,19 +106,19 @@ public class KafkaSpoutRetryExponentialBackoff implements KafkaSpoutRetryService
 
     public static class TimeInterval implements Serializable {
         private final long lengthNanos;
-        private final long length;
         private final TimeUnit timeUnit;
+        private final long length;
 
         /**
          * @param length length of the time interval in the units specified by {@link TimeUnit}
          * @param timeUnit unit used to specify a time interval on which to specify a time unit
          */
         public TimeInterval(long length, TimeUnit timeUnit) {
-            this.length = length;
-            this.timeUnit = timeUnit;
             this.lengthNanos = timeUnit.toNanos(length);
+            this.timeUnit = timeUnit;
+            this.length = length;
         }
-
+        
         public static TimeInterval seconds(long length) {
             return new TimeInterval(length, TimeUnit.SECONDS);
         }
@@ -116,19 +126,15 @@ public class KafkaSpoutRetryExponentialBackoff implements KafkaSpoutRetryService
         public static TimeInterval milliSeconds(long length) {
             return new TimeInterval(length, TimeUnit.MILLISECONDS);
         }
-
+        
         public static TimeInterval microSeconds(long length) {
             return new TimeInterval(length, TimeUnit.MICROSECONDS);
         }
-
+        
         public long lengthNanos() {
             return lengthNanos;
         }
-
-        public long length() {
-            return length;
-        }
-
+        
         public TimeUnit timeUnit() {
             return timeUnit;
         }
@@ -165,26 +171,27 @@ public class KafkaSpoutRetryExponentialBackoff implements KafkaSpoutRetryService
     }
 
     @Override
-    public Set<TopicPartition> retriableTopicPartitions() {
-        final Set<TopicPartition> tps = new HashSet<>();
-        final long currentTimeNanos = System.nanoTime();
+    public Map<TopicPartition, Long> earliestRetriableOffsets() {
+        final Map<TopicPartition, Long> tpToEarliestRetriableOffset = new HashMap<>();
+        final long currentTimeNanos = Time.nanoTime();
         for (RetrySchedule retrySchedule : retrySchedules) {
             if (retrySchedule.retry(currentTimeNanos)) {
                 final KafkaSpoutMessageId msgId = retrySchedule.msgId;
-                tps.add(new TopicPartition(msgId.topic(), msgId.partition()));
+                final TopicPartition tpForMessage = new TopicPartition(msgId.topic(), msgId.partition());
+                tpToEarliestRetriableOffset.merge(tpForMessage, msgId.offset(), Math::min);
             } else {
                 break;  // Stop searching as soon as passed current time
             }
         }
-        LOG.debug("Topic partitions with entries ready to be retried [{}] ", tps);
-        return tps;
+        LOG.debug("Topic partitions with entries ready to be retried [{}] ", tpToEarliestRetriableOffset);
+        return tpToEarliestRetriableOffset;
     }
 
     @Override
     public boolean isReady(KafkaSpoutMessageId msgId) {
         boolean retry = false;
         if (toRetryMsgs.contains(msgId)) {
-            final long currentTimeNanos = System.nanoTime();
+            final long currentTimeNanos = Time.nanoTime();
             for (RetrySchedule retrySchedule : retrySchedules) {
                 if (retrySchedule.retry(currentTimeNanos)) {
                     if (retrySchedule.msgId.equals(msgId)) {
@@ -265,13 +272,27 @@ public class KafkaSpoutRetryExponentialBackoff implements KafkaSpoutRetryService
             return true;
         }
     }
+    
+    @Override
+    public int readyMessageCount() {
+        int count = 0;
+        final long currentTimeNanos = Time.nanoTime();
+        for (RetrySchedule retrySchedule : retrySchedules) {
+            if (retrySchedule.retry(currentTimeNanos)) {
+                ++count;
+            } else {
+                break; //Stop counting when past current time
+            }
+        }
+        return count;
+    }
 
     // if value is greater than Long.MAX_VALUE it truncates to Long.MAX_VALUE
     private long nextTime(KafkaSpoutMessageId msgId) {
-        final long currentTimeNanos = System.nanoTime();
+        final long currentTimeNanos = Time.nanoTime();
         final long nextTimeNanos = msgId.numFails() == 1                // numFails = 1, 2, 3, ...
                 ? currentTimeNanos + initialDelay.lengthNanos
-                : (currentTimeNanos + delayPeriod.timeUnit.toNanos((long) Math.pow(delayPeriod.length, msgId.numFails() - 1)));
+                : currentTimeNanos + delayPeriod.lengthNanos * (long)(Math.pow(2, msgId.numFails()-1));
         return Math.min(nextTimeNanos, currentTimeNanos + maxDelay.lengthNanos);
     }
 
--- a/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutRetryService.java
+++ b/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutRetryService.java
@@ -22,6 +22,7 @@ import org.apache.kafka.common.TopicPartition;
 
 import java.io.Serializable;
 import java.util.Collection;
+import java.util.Map;
 import java.util.Set;
 
 /**
@@ -54,10 +55,11 @@ public interface KafkaSpoutRetryService extends Serializable {
     boolean retainAll(Collection<TopicPartition> topicPartitions);
 
     /**
-     * @return set of topic partitions that have offsets that are ready to be retried, i.e.,
-     * for which a tuple has failed and has retry time less than current time
+     * @return The earliest retriable offset for each TopicPartition that has
+     * offsets ready to be retried, i.e. for which a tuple has failed
+     * and has retry time less than current time
      */
-    Set<TopicPartition> retriableTopicPartitions();
+    Map<TopicPartition, Long> earliestRetriableOffsets();
 
     /**
      * Checks if a specific failed {@link KafkaSpoutMessageId} is ready to be retried,
@@ -75,4 +77,6 @@ public interface KafkaSpoutRetryService extends Serializable {
      * Returns false is this message is not scheduled for retrial
      */
     boolean isScheduled(KafkaSpoutMessageId msgId);
+    
+    int readyMessageCount();
 }
--- a/external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/KafkaSpoutEmitTest.java
+++ b/external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/KafkaSpoutEmitTest.java
@@ -1,0 +1,250 @@
+/*
+ * Copyright 2017 The Apache Software Foundation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.storm.kafka.spout;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.junit.Assert.assertThat;
+import static org.mockito.Matchers.anyCollection;
+import static org.mockito.Matchers.anyLong;
+import static org.mockito.Matchers.anyObject;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.reset;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+import org.apache.kafka.clients.consumer.ConsumerRebalanceListener;
+import org.apache.kafka.clients.consumer.ConsumerRecord;
+import org.apache.kafka.clients.consumer.ConsumerRecords;
+import org.apache.kafka.clients.consumer.KafkaConsumer;
+import org.apache.kafka.common.TopicPartition;
+import org.apache.storm.kafka.spout.builders.SingleTopicKafkaSpoutConfiguration;
+import org.apache.storm.kafka.spout.internal.KafkaConsumerFactory;
+import org.apache.storm.spout.SpoutOutputCollector;
+import org.apache.storm.task.TopologyContext;
+import org.junit.Test;
+import org.mockito.ArgumentCaptor;
+
+import static org.apache.storm.kafka.spout.builders.SingleTopicKafkaSpoutConfiguration.getKafkaSpoutConfigBuilder;
+import static org.mockito.Mockito.inOrder;
+import static org.mockito.Mockito.never;
+
+import java.util.HashSet;
+import org.apache.storm.utils.Time;
+import org.apache.storm.utils.Time.SimulatedTime;
+import org.mockito.InOrder;
+
+public class KafkaSpoutEmitTest {
+
+    private final long offsetCommitPeriodMs = 2_000;
+    private final TopologyContext contextMock = mock(TopologyContext.class);
+    private final SpoutOutputCollector collectorMock = mock(SpoutOutputCollector.class);
+    private final Map<String, Object> conf = new HashMap<>();
+    private final TopicPartition partition = new TopicPartition(SingleTopicKafkaSpoutConfiguration.TOPIC, 1);
+    private KafkaConsumer<String, String> consumerMock;
+    private KafkaSpout<String, String> spout;
+    private KafkaSpoutConfig spoutConfig;
+
+    private void setupSpout(Set<TopicPartition> assignedPartitions) {
+        spoutConfig = getKafkaSpoutConfigBuilder(-1)
+            .setOffsetCommitPeriodMs(offsetCommitPeriodMs)
+            .build();
+
+        consumerMock = mock(KafkaConsumer.class);
+        KafkaConsumerFactory<String, String> consumerFactory = (kafkaSpoutConfig) -> consumerMock;
+
+        //Set up a spout listening to 1 topic partition
+        spout = new KafkaSpout<>(spoutConfig, consumerFactory);
+
+        spout.open(conf, contextMock, collectorMock);
+        spout.activate();
+
+        ArgumentCaptor<ConsumerRebalanceListener> rebalanceListenerCapture = ArgumentCaptor.forClass(ConsumerRebalanceListener.class);
+        verify(consumerMock).subscribe(anyCollection(), rebalanceListenerCapture.capture());
+
+        //Assign partitions to the spout
+        ConsumerRebalanceListener consumerRebalanceListener = rebalanceListenerCapture.getValue();
+        consumerRebalanceListener.onPartitionsAssigned(assignedPartitions);
+    }
+
+    @Test
+    public void testNextTupleEmitsAtMostOneTuple() {
+        //The spout should emit at most one message per call to nextTuple
+        //This is necessary for Storm to be able to throttle the spout according to maxSpoutPending
+        setupSpout(Collections.singleton(partition));
+        Map<TopicPartition, List<ConsumerRecord<String, String>>> records = new HashMap<>();
+        List<ConsumerRecord<String, String>> recordsForPartition = new ArrayList<>();
+        for (int i = 0; i < 10; i++) {
+            recordsForPartition.add(new ConsumerRecord(partition.topic(), partition.partition(), i, "key", "value"));
+        }
+        records.put(partition, recordsForPartition);
+
+        when(consumerMock.poll(anyLong()))
+            .thenReturn(new ConsumerRecords(records));
+
+        spout.nextTuple();
+
+        verify(collectorMock, times(1)).emit(anyObject(), anyObject(), anyObject());
+    }
+
+    @Test
+    public void testNextTupleEmitsFailedMessagesEvenWhenMaxUncommittedOffsetsIsExceeded() {
+        //The spout must reemit failed messages waiting for retry even if it is not allowed to poll for new messages due to maxUncommittedOffsets being exceeded
+        
+        //Emit maxUncommittedOffsets messages, and fail all of them. Then ensure that the spout will retry them when the retry backoff has passed
+        try (SimulatedTime simulatedTime = new SimulatedTime()) {
+            setupSpout(Collections.singleton(partition));
+            Map<TopicPartition, List<ConsumerRecord<String, String>>> records = new HashMap<>();
+            List<ConsumerRecord<String, String>> recordsForPartition = new ArrayList<>();
+            for (int i = 0; i < spoutConfig.getMaxUncommittedOffsets(); i++) {
+                //This is cheating a bit since maxPollRecords would normally spread this across multiple polls
+                recordsForPartition.add(new ConsumerRecord(partition.topic(), partition.partition(), i, "key", "value"));
+            }
+            records.put(partition, recordsForPartition);
+
+            when(consumerMock.poll(anyLong()))
+                .thenReturn(new ConsumerRecords(records));
+
+            for (int i = 0; i < recordsForPartition.size(); i++) {
+                spout.nextTuple();
+            }
+
+            ArgumentCaptor<KafkaSpoutMessageId> messageIds = ArgumentCaptor.forClass(KafkaSpoutMessageId.class);
+            verify(collectorMock, times(recordsForPartition.size())).emit(anyObject(), anyObject(), messageIds.capture());
+
+            for (KafkaSpoutMessageId messageId : messageIds.getAllValues()) {
+                spout.fail(messageId);
+            }
+
+            reset(collectorMock);
+
+            Time.advanceTime(50);
+            //No backoff for test retry service, just check that messages will retry immediately
+            for (int i = 0; i < recordsForPartition.size(); i++) {
+                spout.nextTuple();
+            }
+
+            ArgumentCaptor<KafkaSpoutMessageId> retryMessageIds = ArgumentCaptor.forClass(KafkaSpoutMessageId.class);
+            verify(collectorMock, times(recordsForPartition.size())).emit(anyObject(), anyObject(), retryMessageIds.capture());
+
+            //Verify that the poll started at the earliest retriable tuple offset
+            List<Long> failedOffsets = new ArrayList<>();
+            for(KafkaSpoutMessageId msgId : messageIds.getAllValues()) {
+                failedOffsets.add(msgId.offset());
+            }
+            InOrder inOrder = inOrder(consumerMock);
+            inOrder.verify(consumerMock).seek(partition, failedOffsets.get(0));
+            inOrder.verify(consumerMock).poll(anyLong());
+        }
+    }
+    
+    @Test
+    public void testNextTupleEmitsAtMostMaxUncommittedOffsetsPlusMaxPollRecordsWhenRetryingTuples() {
+        /*
+        The spout must reemit failed messages waiting for retry even if it is not allowed to poll for new messages due to maxUncommittedOffsets being exceeded.
+        numUncommittedOffsets is equal to numNonRetriableEmittedTuples + numRetriableTuples.
+        The spout will only emit if numUncommittedOffsets - numRetriableTuples < maxUncommittedOffsets (i.e. numNonRetriableEmittedTuples < maxUncommittedOffsets)
+        This means that the latest offset a poll can start at for a retriable partition,
+        counting from the last committed offset, is maxUncommittedOffsets,
+        where there are maxUncommittedOffsets - 1 uncommitted tuples "to the left".
+        If the retry poll starts at that offset, it at most emits the retried tuple plus maxPollRecords - 1 new tuples.
+        The limit on uncommitted offsets for one partition is therefore maxUncommittedOffsets + maxPollRecords - 1.
+        
+        It is only necessary to test this for a single partition, because partitions can't contribute negatively to numNonRetriableEmittedTuples,
+        so if the limit holds for one partition, it will also hold for each individual partition when multiple are involved.
+        
+        This makes the actual limit numPartitions * (maxUncommittedOffsets + maxPollRecords - 1)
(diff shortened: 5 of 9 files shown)


srdo:
I think it would be best if we merged maxPollOffsets and maxUncommittedOffsets, since having them be different has some undesirable side effects. @hmcl do you have an opinion.

Also this drops support for Kafka 0.9 in KafkaSpoutConfig, since I believe https://github.com/apache/storm/pull/1556 already means that this release won't work with 0.9.

srdo:
This also changes the exponential backoff formula from currentTime + delayPeriod^(failCount-1) to currentTime + delayPeriod*2^(failCount-1). Multiplying the delay by itself causes the delay to grow extremely quickly, and probably wasn't intended.

It might make sense to add jitter to the backoff as well.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutRetryExponentialBackoff.java
```diff
@@ -27,13 +27,15 @@
...
 import java.util.Set;
 import java.util.TreeSet;
 import java.util.concurrent.TimeUnit;
+import org.apache.storm.utils.Time;
 
 /**
  * Implementation of {@link KafkaSpoutRetryService} using the exponential backoff formula. The time of the nextRetry is set as follows:
- * nextRetry = failCount == 1 ? currentTime + initialDelay : currentTime + delayPeriod^(failCount-1)    where failCount = 1, 2, 3, ...
+ * nextRetry = failCount == 1 ? currentTime + initialDelay : currentTime + delayPeriod*2^(failCount-1)    where failCount = 1, 2, 3, ...
  * nextRetry = Min(nextRetry, currentTime + maxDelay)
```
hmcl: The exponential backoff is a geometric progression with rate `delayPeriod`, hence has a very well known formula. Wy change the formula? Why the constant factor of 2 ?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutRetryExponentialBackoff.java
```diff
@@ -54,59 +56,92 @@
...
+                result = entry1.hashCode() - entry2.hashCode();
+            }
+            return result;
         }
     }
 
     private class RetrySchedule {
         private final KafkaSpoutMessageId msgId;
-        private long nextRetryTimeNanos;
+        private long nextRetryTimeMs;
```
hmcl: Any special reason to change this for ms ?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutRetryExponentialBackoff.java
```diff
@@ -54,59 +56,92 @@
...
         public String toString() {
             return "RetrySchedule{" +
                     "msgId=" + msgId +
-                    ", nextRetryTime=" + nextRetryTimeNanos +
+                    ", nextRetryTimeMs=" + nextRetryTimeMs +
                     '}';
         }
 
+        @Override
+        public int hashCode() {
```
hmcl: why did you decide to implement  equals and hashcode ?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutRetryExponentialBackoff.java
```diff
@@ -54,59 +56,92 @@
...
+         * @param timeUnit unit used to specify a time interval on which to specify a time unit. Smallest supported unit is milliseconds
          */
         public TimeInterval(long length, TimeUnit timeUnit) {
-            this.length = length;
-            this.timeUnit = timeUnit;
-            this.lengthNanos = timeUnit.toNanos(length);
+            
+            if(timeUnit == TimeUnit.MICROSECONDS || timeUnit == TimeUnit.NANOSECONDS) {
+                throw new IllegalArgumentException("TimeInterval does not support time units smaller than milliseconds");
+            }
```
hmcl: Why?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutConfig.java
```diff
@@ -329,11 +328,13 @@ private Builder(Builder<?, ?> builder, SerializableDeserializer<K> keyDes, Class
...
             return this;
         }
-        
+
         /**
          * Defines the max number of polled offsets (records) that can be pending commit, before another poll can take place.
          * Once this limit is reached, no more offsets (records) can be polled until the next successful commit(s) sets the number
-         * of pending offsets bellow the threshold. The default is {@link #DEFAULT_MAX_UNCOMMITTED_OFFSETS}.
+         * of pending offsets below the threshold. The default is {@link #DEFAULT_MAX_UNCOMMITTED_OFFSETS}.
+         * Note that this limit can in some cases be exceeded, but no partition will exceed this limit by more than 2*maxPollRecords.
```
hmcl: where is the number 2*maxPollRecords coming from?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -78,7 +78,7 @@
     private transient Map<TopicPartition, OffsetManager> acked;           // Tuples that were successfully acked. These tuples will be committed periodically when the commit timer expires, after consumer rebalance, or on close/deactivate
     private transient Set<KafkaSpoutMessageId> emitted;                 // Tuples that have been emitted but that are "on the wire", i.e. pending being acked or failed
     private transient Iterator<ConsumerRecord<K, V>> waitingToEmit;         // Records that have been polled and are queued to be emitted in the nextTuple() call. One record is emitted per nextTuple()
-    private transient long numUncommittedOffsets;                       // Number of offsets that have been polled and emitted but not yet been committed
+    private transient int numUncommittedOffsets;                       // Number of offsets that have been polled and emitted but not yet been committed
```
hmcl: Why impose a more restricted data type? What if the user desires a large number?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutRetryExponentialBackoff.java
srdo: The current implementation will multiply the delay by delayPeriod for each retry after the first. Let's say I set initialDelay to 1000ms and retryPeriod to 1000ms. The first retry is then scheduled at 1 second after failure. The second at 1 second after that. The third retry is scheduled 1000 seconds after that. The usual exponential backoff function would multiply the backoff by 2 for each failure, not by the backoff itself. This is a geometric progression where you can only set ratio, not scale factor, which isn't very useful (for anything except tiny backoffs, it grows much too fast). If we want to have backoffs with multipliers other than 2, we could add another parameter for that? (in that case that parameter should be an int, not a time period IMO)

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutRetryExponentialBackoff.java
srdo: Storm's time simulation doesn't support values smaller than ms. I'd be happy to implement nanosecond support in Time instead if anyone actually needs sub-millis support.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutRetryExponentialBackoff.java
srdo: Classes of this type are put in TreeSet. If you add a value to TreeSet that equals an object already in the set with respect to compareTo, the new value is dropped. This is a problem in one of the tests I added that uses time simulation, because with simulated time the clock is stopped (i.e. the schedules have the same timestamps). I suppose it could occur outside tests, but it would be extremely rare. No reason not to fix it though.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutRetryExponentialBackoff.java
srdo: The class doesn't support time units less than milliseconds. I'd prefer to warn about misconfiguration rather than just clamping to 0 or 1 millisecond.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
srdo: maxUncommittedOffsets is an int. This just makes it consistent.

srdo: Actually you're right, since the spout may emit more than maxUncommittedOffsets this should be changed back.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutConfig.java
srdo: Sorry, this is a mistake. The worst case is one where maxUncommittedOffsets - 1 messages have been emitted. In that case, the spout may poll and emit another maxPollRecords messages. So the comment should say maxPollRecords, not 2*maxPollRecords.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -240,7 +240,9 @@ private boolean commit() {
 
     private boolean poll() {
         final int maxUncommittedOffsets = kafkaSpoutConfig.getMaxUncommittedOffsets();
-        final boolean poll = !waitingToEmit() && numUncommittedOffsets < maxUncommittedOffsets;
+        final boolean poll = !waitingToEmit() &&
+            (numUncommittedOffsets < maxUncommittedOffsets ||
+            !retryService.retriableTopicPartitions().isEmpty());
```
hmcl: `!retryService.retriableTopicPartitions().isEmpty()` can't this condition cause unbounded  polling (i.e. continue forever), if some specific tuples are always failing?

The goal is to have a limit in how much can be polled (hence kept in memory) in case of recurring failures.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -289,6 +293,13 @@ private void doSeekRetriableTopicPartitions() {
                 kafkaConsumer.seek(rtp, acked.get(rtp).getCommittedOffset() + 1);    // Seek to last committed offset
             }
         }
+        
+        if(!retriableTopicPartitions.isEmpty()) {
+            //Pause other partitions for this poll so maxUncommittedPartitions isn't exceeded by too much
```
hmcl: I am a bit confused about this. When can it be exceeded? By how much? This should be a deterministic number.

I read your comment in the email thread about  maxSpoutPending. Is this related to that ?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
srdo: No, I don't think so. The tuples need to be both failed and ready for retry for this to be true, and if it's true, the consumer will seek back to the last committed offset on the relevant topic partitions. The consumer pause in doSeekRetriableTopicPartitions ensures that we don't get messages for the partitions that don't have failed tuples. maxPollRecords then caps how far past the commit offset we can read on those partitions. If the same tuples are always failing, we'll never get more than maxPollRecords from the commit offset on those partitions as far as I can tell.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
srdo: No, I made a mistake there. maxSpoutPending is being enforced correctly.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutRetryExponentialBackoff.java
srdo: I'll remove these two again, the Comparator change should work even with the default hashcode.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutRetryExponentialBackoff.java
hmcl: OK.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
hmcl: I am still not quite understanding why we need to pause all other topic partitions in the case of a retrial. This could potentially lead to the scenario that one topic partition may be causing a lot of failures, but all others are running fine. If that's the case, we will the throttling all other topic-partitions for no particular reason.

Are you doing this pause to handle the case when `nunUncommittedOffsets` may exceed `maxUncommittedOffsets`? If that is the case, then in the case of a retrial we should only only when  `nunUncommittedOffsets >= maxUncommittedOffsets`. Otherwise perhaps we should poll from all topic-partitions.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutConfig.java
hmcl: Agree

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutRetryExponentialBackoff.java
srdo: I'll take a look at putting nanosecond support into o.a.s.Time.

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/KafkaSpoutEmitTest.java
```diff
@@ -0,0 +1,156 @@
...
+        ArgumentCaptor<ConsumerRebalanceListener> rebalanceListenerCapture = ArgumentCaptor.forClass(ConsumerRebalanceListener.class);
+        verify(consumerMock).subscribe(anyCollection(), rebalanceListenerCapture.capture());
+
+        //Assign partitions to the spout
+        ConsumerRebalanceListener consumerRebalanceListener = rebalanceListenerCapture.getValue();
+        consumerRebalanceListener.onPartitionsAssigned(Collections.singleton(partition));
+    }
+
+    @Test
+    public void shouldEmitAtMostOneMessagePerNextTupleCall() {
```
srdo: Will rename.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
srdo: I thought about it some more, and I don't think it's necessary. As long as we only have to guarantee that maxUncommittedOffsets isn't exceeded by more than maxPollRecords per partition, it should be fine to consume from all partitions. I'll remove this.

srdo:
@hmcl I think this is ready for another look when you get a chance.

hmcl:
@srdo Thanks for the note. I will take a look later today.

hmcl:
@srdo is this patch still relevant in face of the other patches that got merged in ?

srdo:
@hmcl Yes, I believe so. There is still a problem where the spout will stop emitting tuples if more than numUncommittedOffsets tuples have failed.

hmcl:
@srdo what you mean is that if `numUncommittedOffsets  > max UncommittedOffsets` the Spout should still emit the retries, but not poll new records. Is that it ?

srdo:
@hmcl yes, more or less.

srdo:
I think it would be good to get some of the assumptions I've made down in writing, so I just wrote down a more thorough explanation of the issues I think this PR should fix, and what changes I think it should make:

The intention of this PR is to fix an issue where the spout has emitted maxUncommittedOffsets (or more) tuples, and some of them fail, and the spout then hangs because the `numUncommittedOffsets < maxUncommittedOffsets` check when deciding whether to poll doesn't account for retriable tuples. Blindly polling when there are retriable tuples will fix this issue, but leads to us not being able to bound the number of offsets the consumer has read past the last committed offset.

For an example of a case where we can't put a bound on uncommitted offsets, say that partition 0 has maxUncommittedOffsets + maxPollRecords emitted tuples past the commit offset, and partition 1 has no emitted tuples (or any number below maxUncommittedOffsets, it doesn't matter).

When tuples fail and become retriable on partition 0, the spout would blindly poll for more tuples. If it gets tuples from partition 0 it's okay, since it seeks back to the committed offset for that partition, so it won't go beyond committedOffset + maxPollRecords (since the retry logic seeks back to the committed offset on every poll, the spout can't go more than maxPollRecords beyond that offset). If it gets tuples from partition 1 we're effectively just emitting new tuples while ignoring maxUncommittedOffsets. Since we aren't controlling which partitions the polled tuples may come from, we can't say anything meaningful about a cap on uncommitted offsets. While I think it would probably work out to being capped anyway, due to the consumer consuming partitions round robin (AFAIK), I'd prefer if the spout implementation doesn't make assumptions beyond those guaranteed by the consumer API (in which case we should assume any call to `KafkaConsumer.poll` could return messages for any assigned non-paused partition).

So the way to fix that issue is to ensure that if we're polling for retriable tuples, we only poll on those partitions that have retriable tuples (i.e. we pause the others when doing `doSeekRetriableTopicPartitions`, then resume all assigned after calling `KafkaConsumer.poll`). Pausing the other partitions should only affect that poll cycle, since once retriable tuples get emitted they're no longer retriable, and we won't hit `doSeekRetriableTopicPartitions` for those tuples again right away. In most polls (no failed tuples, or tuples are failed but not ready for retry), we won't hit the pausing case.

If we pause partitions with no retriable tuples when polling on partitions with retriable tuples, we should be able to guarantee that any partition never gets more than maxUncommittedOffsets + maxPollRecords - 1 past the last committed offset. In the case where there are no failed tuples, we can reach the limit by having maxUncommittedOffsets - 1 emitted offsets, and polling once, getting up to maxPollRecords more. If there are retriable tuples, pausing will stop us from ignoring the maxUncommittedOffsets cap for partitions with no retriable tuples, and the partitions with retriable tuples won't get more than maxPollRecords beyond the last committed offset, since the consumer seeks back to that offset when polling for retriable offsets.

There's a second minor issue I'd like this PR to address:
If maxPollRecords isn't exactly equal to maxUncommittedOffsets, the spout can behave in some undesirable ways.
* If maxPollRecords is greater than maxUncommittedOffsets, the maxUncommittedOffsets limit may be exceeded on any one poll. In this case there's no reason to have 2 separate variables, since the net effect is the same as setting maxUncommittedOffsets to be equal to maxPollRecords. 
* If maxPollRecords is less than maxUncommittedOffsets, there's a risk of the spout getting stuck on some tuples for a while when it is retrying tuples.
  Say there are 10 retriable tuples following the last committed offset, and maxUncommittedOffsets is 10. If maxPollRecords is 5 and the first 5 retriable tuples are reemitted in the first batch, the next 5 tuples can't be emitted until (some of) the first 5 are acked. This is because the spout will seek the consumer back to the last committed offset any time there are failed tuples, which will lead to it getting the first 5 tuples out of the consumer, checking that they are emitted, and skipping them. This will repeat until the last committed offset moves. If there are other partitions with tuples available, those tuples may get emitted, but the "blocked" partition won't progress until some tuples are acked on it.
  
I think it might make sense to remove maxUncommittedOffsets entirely, and have the spout use maxPollRecords instead?

The tl;dr is that I think we should allow polling if there are retriable offsets, even if `numUncommittedOffsets >= maxUncommittedOffsets`, we should pause non-retriable partitions when polling for retries, and we should maybe merge maxUncommittedOffsets and maxPollRecords.

@hmcl Could you take a look at this explanation and let me know if I got something wrong, or if you'd prefer to solve the issues in another way? I think it got a bit hypothetical.

srdo:
@hmcl I made the changes mentioned in my last comment. 

As far as I can tell there's no way to get the default max.poll.records out of Kafka, so for now I've made do with adding comments in KafkaSpoutConfig recommending that maxUncommittedOffsets and maxPollRecords should be equal.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -270,17 +272,36 @@ public void setWaitingToEmit(ConsumerRecords<K,V> consumerRecords) {
 
     // ======== poll =========
     private ConsumerRecords<K, V> pollKafkaBroker() {
-        doSeekRetriableTopicPartitions();
+        Set<TopicPartition> retriableTopicPartitions = doSeekRetriableTopicPartitions();
+        Set<TopicPartition> pausedTopicPartitions = pauseNonRetriableTopicPartitionsIfAtUncommittedOffsetsLimit(retriableTopicPartitions);
         if (refreshSubscriptionTimer.isExpiredResetOnTrue()) {
             kafkaSpoutConfig.getSubscription().refreshAssignment();
         }
         final ConsumerRecords<K, V> consumerRecords = kafkaConsumer.poll(kafkaSpoutConfig.getPollTimeoutMs());
```
HeartSaVioR: Minor: I'd rather apply defensive programming here (to guarantee calling `resume` at any chance) like below: 
```
try {
    final ConsumerRecords<K, V> consumerRecords = kafkaConsumer.poll(kafkaSpoutConfig.getPollTimeoutMs());
    final int numPolledRecords = consumerRecords.count();
    LOG.debug("Polled [{}] records from Kafka. [{}] uncommitted offsets across all topic partitions", numPolledRecords, numUncommittedOffsets);
    return consumerRecords;
} finally {
    if (!pausedTopicPartitions.isEmpty()) {
        //Ensure that we resume all partitions in case some were paused when processing retries
        kafkaConsumer.resume(kafkaConsumer.assignment());
    }
}
```

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -329,11 +350,13 @@ private boolean emitTupleIfNotEmitted(ConsumerRecord<K, V> record) {
                     }
                 
                     emitted.add(msgId);
-                    numUncommittedOffsets++;
                 
                     if (isScheduled) { // Was scheduled for retry, now being re-emitted. Remove from schedule.
                         retryService.remove(msgId);
-                    }
+                    } else {
+                    //New tuple, increment the uncommitted offset counter
```
HeartSaVioR: nits: indentation seemed off here and below lines.

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/KafkaSpoutEmitTest.java
```diff
@@ -0,0 +1,169 @@
+/*
```
HeartSaVioR: We should have separate PR for 1.x branch as well, or change current implementation to be compatible with JDK 7.

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/MaxUncommittedOffsetTest.java
```diff
@@ -0,0 +1,245 @@
...
+        });
+    }
+
+    private void initializeSpout(int msgCount) {
+        populateTopicData(SingleTopicKafkaSpoutConfiguration.TOPIC, msgCount);
+        spout.open(conf, topologyContext, collector);
+        spout.activate();
+    }
+
+    public ArgumentCaptor<KafkaSpoutMessageId> emitMaxUncommittedOffsetsMessagesAndCheckNoMoreAreEmitted(int messageCount) {
```
HeartSaVioR: Minor: Might be better to include assertion for `messageCount >= maxUncommittedOffsets`.

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/MaxUncommittedOffsetTest.java
```diff
@@ -0,0 +1,245 @@
...
+            messageIds.capture());
+        return messageIds;
+    }
+
+    @Test
+    public void testNextTupleCanEmitMoreMessagesWhenDroppingBelowMaxUncommittedOffsetsDueToCommit() {
+        //The spout must respect maxUncommittedOffsets after committing a set of records
+        try (Time.SimulatedTime simulatedTime = new Time.SimulatedTime()) {
+            //First check that maxUncommittedOffsets is respected when emitting from scratch
+            ArgumentCaptor<KafkaSpoutMessageId> messageIds = emitMaxUncommittedOffsetsMessagesAndCheckNoMoreAreEmitted(numMessages);
```
HeartSaVioR: Minor: Precondition here is that `numMessages >= 2 * (maxUncommittedOffsets)`.

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/MaxUncommittedOffsetTest.java
```diff
@@ -0,0 +1,245 @@
...
+                spout.nextTuple();
+            }
+            ArgumentCaptor<KafkaSpoutMessageId> thirdRunMessageIds = ArgumentCaptor.forClass(KafkaSpoutMessageId.class);
+            verify(collector, times(maxPollRecords-1)).emit(
+                anyObject(),
+                anyObject(),
+                thirdRunMessageIds.capture());
+            reset(collector);
+
+            failAllExceptTheFirstMessageThenCommit(thirdRunMessageIds);
```
HeartSaVioR: Minor: might be better to add comparison between captured message ids and first run message ids but excluding first one to make sure failed tuples are reemitted.

HeartSaVioR:
IMHO, looks like `maxUncommittedOffsets` makes thing complicated.

We've two throttling mechanism here: maxUncommittedOffsets and max spout pending.
Since KafkaSpout guarantees that only one tuple emitted per one nextTuple() call, max spout pending will work perfectly.
So if it's just for throttling, why not relying on max spout pending? At-least-once is guaranteed whether we have maxUncommittedOffsets or not.

If max spout pending is unset, users should have known that Spout will emit tuples as many and fast as possible unless backpressure is not in place.

hmcl:
@srdo apologies for a little delay. Had a lot of deadlines for last week. I will finish reviewing this today.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
srdo: Will fix

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
srdo: Will fix

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/KafkaSpoutEmitTest.java
srdo: I'll make a PR for 1.x soon

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/MaxUncommittedOffsetTest.java
srdo: Will add

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/MaxUncommittedOffsetTest.java
srdo: Added a comment to numMessages

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/MaxUncommittedOffsetTest.java
srdo: Will add a comparison

srdo:
@HeartSaVioR I agree, I think it would be much simpler to rely on maxSpoutPending. maxUncommittedOffsets does have a slightly different meaning though. With only maxSpoutPending, the spout might progress arbitrarily far past the last committed offset, since there's only a limit on how many tuples are in-flight at once, rather than on how many tuples have been emitted past the committed offset. If a user needs to cap how far the spout can read ahead for some reason, they can't do that with maxSpoutPending.

I can't tell if that's a real need anyone has though. @hmcl any opinion?

HeartSaVioR:
+1
Once @hmcl finished the review and no outstanding comments there I'll merge.

Btw, regarding maxUncommittedOffsets, yes it provides more guarantees but if we guarantee that commit occurs in time periodically (we're providing option too), IMHO that's enough.

The thing I have in mind is that commit in nextTuple() doesn't strictly guarantee commit in time. Two situations come in mind: reached max spout pending, backpressure in place.

If spout reached max spout pending, since we are emitting **at most a tuple** for nextTuple() so when tuple tree for a tuple is complete or tuple fails, nextTuple() can be called. At least nextTuple() will be called after tuple timeout (actually up to 2 * tuple timeout due to underlying implementation).
So if users can tolerate this, it might be OK, but could be far from configured value.

If backpressure is in place, nextTuple() will never be called. We can't guarantee any time frame here.

If we want to strictly guarantee that commit occurs in time, it should be apart from spout's lifecycle, like timer. It might also introduce some handling of thread-safety so a bit more complicated, but IMO simpler than respecting maxUncommittedOffsets.

HeartSaVioR:
@hmcl 
I guess you're busy, but could you continue reviewing this one, or please let me know if you are not available for reviewing soon? I'd like to get this merged sooner since this is necessary for 1.1.0.

hmcl:
@HeartSaVioR @srdo reviewing this right now.

srdo:
@hmcl @HeartSaVioR I think if we can't think of a good use case for maxUncommittedOffsets, we'd be better off removing it.

hmcl:
@srdo I am reading through your detailed explanation again and will let you know my opinion concerning the need, or no need, for the parameter `maxUncommittedOffsets`, based on all the facts you state. 

Prior to that, however, I will share right away what is the rationale behind `maxUncommittedOffsets` and why I think we still need it. 

`maxUncommittedOffsets` is unrelated to `ConsumerConfig#MAX_POLL_RECORDS_CONFIG` and to `Config#TOPOLOGY_MAX_SPOUT_PENDING`. It's purpose is to allow the `KafkaSpout` to keep polling/fetching records from Kafka while imposing a (memory consumption) limit on how many offsets can be **polled and non committed over ANY ARBITRARY NUMBER of polls to Kafka (not just one poll)**, before the KafkaSpout stops polling. 

This limit is necessary because if e.g. tuple for offset **3** keeps failing, and all subsequent offsets **(4, 5, 6, ... 1M, ... 1B, ...)** are acked, the spout won't be able to commit to Kafka any offset greater than 3, unless it reaches **maxNumberOfRetrials**. If there is no limit on **maxNumberOfRetrials** (in order to guarantee at least once delivery), and offset 3 keeps failing forever, the spout won't ever be able to commit any offset > 3. In the limit this would cause the `Map<TopicPartition, OffsetManager> acked` to grow to infinite size (despite it's small footprint, which is constituted of mainly offsets and some bookkeeping small objects - it doesn't have `ConsumerRecord`s). 

In order to put a cap on the size of this map, `maxUncommittedOffset` was created to stop polling if this number is too high, and avoid `OutOfMemoryError`. **The purpose of the KafkaSpout keeping on polling despite a particular offset (tuple) keeping on failing is to increase the overall throughput.** If at some point the offset 3 is successfully acked, the max subset of continuous offsets acked would be committed to Kafka in the next `Timer commitTimer` cycle, and the size of `Map<TopicPartition, OffsetManager> acked` would be reduced (potentially by a lot). The spout could then keep doing its work and the overall throughput would be much higher.

`maxUncommittedOffsets` is not a hard limit parameter, and I don't think that we have to enforce that the KafkaSpout never exceeds this number. It is OK to exceed it as long as the JVM doesn't throw `OutOfMemoryError`. Currently, and prior to these proposed changes, in the worse case **maxUncommittedOffsets** is upper bounded by  **maxUncommittedOffsets + maxFetchRecords - 1**, which is OK as long as there is memory. 

Perhaps the name `maxUncommittedOffsets` is not the most clear name. We can either change the name (which can cause background compatibility issues and/or confusion), or document properly what this parameter really means. The bottom line is that I think that we need a parameter doing the job that `maxUncommittedOffsets` is currently doing, which I don't think can be accomplished with a combination of`ConsumerConfig#MAX_POLL_RECORDS_CONFIG` and/or `Config#TOPOLOGY_MAX_SPOUT_PENDING`

The name `maxUncommittedOffsets` should more precisely be something like `allowPollingIflLessThanMaxUncommittedOffsets`

Please let me know your thoughts. I will still go over your explanation again in the meantime.

srdo:
@hmcl Thanks for the explanation. It makes more sense now.

I don't think we can meaningfully handle the case where the user has specified no cap on retries and one or more tuples keep failing.

In the situation you describe, `maxUncommittedOffsets` isn't really solving the problem. If offset 3 fails and all other tuples can be acked, the spout will emit another `maxUncommittedOffsets` tuples on the same partition as 3, and will then stop emitting because it can't commit any tuples until 3 is acked. At this point, only offset 3 will get re-emitted until it hopefully succeeds at some point, so having a continually failing tuple (or just one that fails many times) will still "clog" the spout.

If `maxUncommittedOffsets` is removed, you're right that the `acked` map size is technically uncapped, but given the amount of information we store, and how the spout handles retries, I think it is a non-issue. Just to illustrate, say we remove `maxUncommittedOffsets`, and let's say again that offset 3 is failing a large number of times, and the last acked offset is far beyond it, e.g. 5000000. When 3 is ready for retry, the consumer is reset to that position on the relevant partition. This means it fetches and emits message 3. It will also have to fetch 4...5000000 (potentially over many calls to `poll()`), because we never seek forward to where we left off). The spout therefore has to spend time fetching and discarding all tuples up to 5000000 before it can finally emit another tuple. It seems likely that it'll hit the point where it is spending all the time fetching and discarding acked tuples earlier than it'll run out of memory to store their offsets.

Disregarding the issue with consumer seeking, because my reasoning relies on an implementation detail of how we do retries, I'm still not sure `maxUncommittedOffsets` is allowing higher throughput compared to not having it. If we allow the spout to fail with an `OutOfMemoryError`, the spout will have had higher throughput up to the crash than if it were being throttled by `maxUncommittedOffsets` (because otherwise it would also have had an OOME in that case). It really seems to me like all `maxUncommittedOffsets` is doing is trading having the spout potentially cause an OOME due to `acked` size, in exchange for making the spout react to the same situation by not emitting any more tuples. I'm not sure that is better, because data flow stops in either case. 

`maxUncommittedOffsets` could have some value in a system where the `KafkaSpout` is a secondary stream source and the messages coming out of it aren't time sensitive. In that case it might be fine to let the `KafkaSpout` stop emitting tuples for a while if some tuples temporarily can't be acked, but having it cause an OOME would be too disruptive because the primary spout could be doing fine. I can't come up with a concrete example of this kind of configuration though.

I'd be fine with keeping `maxUncommittedOffsets` for that kind of situation.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -270,17 +272,40 @@ public void setWaitingToEmit(ConsumerRecords<K,V> consumerRecords) {
 
     // ======== poll =========
     private ConsumerRecords<K, V> pollKafkaBroker() {
-        doSeekRetriableTopicPartitions();
-        if (refreshSubscriptionTimer.isExpiredResetOnTrue()) {
-            kafkaSpoutConfig.getSubscription().refreshAssignment();
+        Set<TopicPartition> retriableTopicPartitions = doSeekRetriableTopicPartitions();
+        Set<TopicPartition> partitionsToPause = getPartitionsToPauseToEnforceUncommittedOffsetsLimit(retriableTopicPartitions);
```
hmcl: let's replace lines 276 to 280 with the code I suggest in [here](https://github.com/apache/storm/pull/1924/files#r106557594)

!partitionsToPause.isEmpty() is not necessary

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutConfig.java
```diff
@@ -268,12 +268,12 @@ private Builder(Builder<?, ?> builder, SerializableDeserializer<K> keyDes, Class
         
         /**
          * The maximum number of records a poll will return.
-         * Will only work with Kafka 0.10.0 and above.
+         * This is limited by maxUncommittedOffsets, since it doesn't make sense to allow larger polls than the spout is allowed to emit.
+         * Please note that when there are retriable tuples on a partition, maxPollRecords is an upper bound for how far the spout will read past the last committed offset on that partition.
+         * It is recommended that users set maxUncommittedOffsets and maxPollRecords to be equal.
```
srdo: @hmcl I agree that it is not ideal, but there's an issue with letting maxUncommittedOffsets be larger than maxPollRecords.

From an earlier response:
"If maxPollRecords is less than maxUncommittedOffsets, there's a risk of the spout getting stuck on some tuples for a while when it is retrying tuples.
Say there are 10 retriable tuples following the last committed offset, and maxUncommittedOffsets is 10. If maxPollRecords is 5 and the first 5 retriable tuples are reemitted in the first batch, the next 5 tuples can't be emitted until (some of) the first 5 are acked. This is because the spout will seek the consumer back to the last committed offset any time there are failed tuples, which will lead to it getting the first 5 tuples out of the consumer, checking that they are emitted, and skipping them. This will repeat until the last committed offset moves. If there are other partitions with tuples available, those tuples may get emitted, but the "blocked" partition won't progress until some tuples are acked on it."

How about we fix this by making doSeekRetriableTopicPartitions seek to the lowest retriable offset per partition for the partitions with failed tuples, instead of seeking to the last committed/committable offset? It seems like seeking to the last committed offset is likely to have some bad interactions with maxUncommittedOffsets and maxPollRecords.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutRetryExponentialBackoff.java
```diff
@@ -54,21 +55,28 @@
     private static class RetryEntryTimeStampComparator implements Serializable, Comparator<RetrySchedule> {
         @Override
         public int compare(RetrySchedule entry1, RetrySchedule entry2) {
-            return Long.valueOf(entry1.nextRetryTimeNanos()).compareTo(entry2.nextRetryTimeNanos());
+            int result = Long.valueOf(entry1.nextRetryTimeNanos()).compareTo(entry2.nextRetryTimeNanos());
+            
+            if(result == 0) {
+                //TreeSet uses compareTo instead of equals() for the Set contract
+                //Ensure that we can save two retry schedules with the same timestamp
```
srdo: It's probably impossible in real life if System.nanoTime has sufficiently high resolution (not guaranteed to be any better than System.currentTimeMillis), but it can happen in tests that use simulated time. There's still potential for collisions if two retry schedules happen to have the same hashCode and timestamp, but that's very unlikely. I mostly made the change for the benefit of the tests, but it doesn't hurt to reduce the chance of collision in general.

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/builders/SingleTopicKafkaSpoutConfiguration.java
```diff
@@ -41,31 +41,22 @@ public static Config getConfig() {
 
     public static StormTopology getTopologyKafkaSpout(int port) {
         final TopologyBuilder tp = new TopologyBuilder();
-        tp.setSpout("kafka_spout", new KafkaSpout<>(getKafkaSpoutConfig(port)), 1);
+        tp.setSpout("kafka_spout", new KafkaSpout<>(getKafkaSpoutConfigBuilder(port).build()), 1);
```
srdo: The SingleTopicKafkaSpoutConfiguration class was beginning to turn into a mess of telescoping constructors, and it was a bit silly given that we already have a nice builder interface underlying that class.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
srdo: I wasn't sure if calling resume() after every poll() might be expensive, but it appears to just set a flag in the consumer. I'll change it.

srdo:
Sorry to keep changing this. I hope these are the last few changes. The spout should now seek to the lowest retriable offset for each partition when retrying tuples, instead of seeking back to the last committed offset. This means there's not as far as I can see any reason to worry about maxPollRecords being set too low compared to maxUncommittedOffsets, so I've reverted that part of KafkaSpoutConfig.

HeartSaVioR:
@hmcl 
I think @srdo addressed your last review comments. Could you finish the review?
Please let me know if there're some points we need to discuss or resolve, so that we can address them now, or file new issues and postpone to next release if possible.

srdo:
@HeartSaVioR I think as long as the spout can seek to the tuples it needs to retry instead of seeking to the committed offset, all of the minor issues I could think of are resolved.

hmcl:
@srdo @HeartSaVioR I will do the last pass on this first thing in the morning such that we can merge this in.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutRetryService.java
```diff
@@ -54,10 +55,11 @@
     boolean retainAll(Collection<TopicPartition> topicPartitions);
 
     /**
-     * @return set of topic partitions that have offsets that are ready to be retried, i.e.,
-     * for which a tuple has failed and has retry time less than current time
+     * @return The earliest retriable offset for each TopicPartition that has
+     * offsets that are ready to be retried, i.e., for which a tuple has failed
```
hmcl: offsets ready to be retried

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/MaxUncommittedOffsetTest.java
```diff
@@ -57,10 +57,9 @@
     //Current tests require that numMessages >= 2*maxUncommittedOffsets
     private final int numMessages = 100;
     private final int maxUncommittedOffsets = 10;
-    //This is set to be the same as maxUncommittedOffsets since it is difficult to test maxUncommittedOffsets when maxPollRecords is not the same
-    //If maxPollRecords is larger, a single call to poll will emit more than maxUncommittedOffsets messages
-    //If maxPollRecords is lower, it will cap how far past the commit offset the spout can read when there are failed tuples ready for retry
-    private final int maxPollRecords = maxUncommittedOffsets;
+    //Current tests require that maxPollRecords is less than maxUncommittedOffsets
```
hmcl: Is this still true? If so, let's put an assertion validating this in the @Before method or something.

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/SingleTopicKafkaSpoutTest.java
```diff
@@ -266,4 +269,36 @@ public void shouldReplayFirstTupleFailedOutOfOrder() throws Exception {
...
+            spout.nextTuple();
+        }
+        ArgumentCaptor<KafkaSpoutMessageId> messageIds = ArgumentCaptor.forClass(KafkaSpoutMessageId.class);
+        verify(collector, times(messageCount)).emit(anyObject(), anyObject(), messageIds.capture());
+        reset(collector);
+        //Fail tuple 5 and 3, call nextTuple, then fail tuple 2
+        List<KafkaSpoutMessageId> capturedMessageIds = messageIds.getAllValues();
+        spout.fail(capturedMessageIds.get(5));
+        spout.fail(capturedMessageIds.get(3));
+        spout.nextTuple();
```
hmcl: What's the relation between this call to nextTuple and the failed messages? Why is it needed?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutRetryExponentialBackoff.java
```diff
@@ -172,19 +174,20 @@ public KafkaSpoutRetryExponentialBackoff(TimeInterval initialDelay, TimeInterval
     }
 
     @Override
-    public Set<TopicPartition> retriableTopicPartitions() {
-        final Set<TopicPartition> tps = new HashSet<>();
+    public Map<TopicPartition, Long> earliestOffsetForEachRetriableTopicPartition() {
```
hmcl: Let's rename this method to `earliestRetriableOffsets`. The javadoc will provide the details. Let's not write very complex and long method names.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutRetryService.java
```diff
@@ -54,10 +55,11 @@
...
+     * and has retry time less than current time
      */
-    Set<TopicPartition> retriableTopicPartitions();
+    Map<TopicPartition, Long> earliestOffsetForEachRetriableTopicPartition();
```
hmcl: Let's rename this method to `earliestRetriableOffsets`. The javadoc will provide the details. Let's not write very complex and long method names.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutRetryExponentialBackoff.java
```diff
@@ -60,17 +61,19 @@ public int compare(RetrySchedule entry1, RetrySchedule entry2) {
             if(result == 0) {
                 //TreeSet uses compareTo instead of equals() for the Set contract
                 //Ensure that we can save two retry schedules with the same timestamp
-                result = entry1.hashCode() - entry2.hashCode();
+                result = entry1.instanceNumber - entry2.instanceNumber;
```
srdo: Will revert to using hashCode(). I'd like to at least keep that check, since it would be annoying in tests using simulated time if we had to add 1 nanosecond between every call to spout.fail().

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutRetryExponentialBackoff.java
srdo: Will rename

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -273,11 +273,9 @@ public void setWaitingToEmit(ConsumerRecords<K,V> consumerRecords) {
     // ======== poll =========
     private ConsumerRecords<K, V> pollKafkaBroker() {
         Set<TopicPartition> retriableTopicPartitions = doSeekRetriableTopicPartitions();
-        Set<TopicPartition> partitionsToPause = getPartitionsToPauseToEnforceUncommittedOffsetsLimit(retriableTopicPartitions);
         try{
-            if(!partitionsToPause.isEmpty()) {
-                kafkaConsumer.pause(partitionsToPause);
-            }
+            //Pause all partitions with no retriable messages
+            pauseNonRetriablePartitionsToEnforceUncommitedOffsetsLimit(retriableTopicPartitions);
```
srdo: I'm not sure that's more readable though

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutRetryService.java
srdo: Will fix

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/MaxUncommittedOffsetTest.java
srdo: It is still true. I'll add an assertion.

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/SingleTopicKafkaSpoutTest.java
srdo: Basically I wanted to test that the spout will emit each retriable tuple exactly once, even if the tuples fail in a weird order or Storm asks for more tuples in the middle of the failures.

Added a comment to the test.

srdo:
Thanks for the reviews guys.

hmcl:
+1. 

@srdo I forgot to say in the last review. Can you please squash the commits. Once that is done as I am concerned, this patch is ready to merge. 

@HeartSaVioR can you please do your last pass? Thanks.

HeartSaVioR:
@srdo 
Thanks for your patience. +1

I can squash this and merge, but seems like #2004 depends on this so it would be clearer if you squash the commits for this PR (and for 1.x branch #2022 too), and rebase #2004 after merging this.

Let me know when you finish squashing. Thanks again!

srdo:
Very sorry about this, but this is still broken :( 

In the current implementation (pre this PR), the spout can be prevented from polling if there are more than maxUncommittedOffsets failed tuples. This is because the check for whether to poll isn't accounting for retriable tuples already being counted in the numUncommittedOffsets count.

The current fix is to allow failed tuples to retry always, even if maxUncommittedOffsets is exceeded, because failed tuples don't contribute further to the numUncommittedOffsets count. The problem has been to ensure that we don't also emit a bunch of new tuples when polling for retriable tuples, and end up ignoring maxUncommittedOffsets entirely. This was partially fixed by pausing partitions that have no retriable tuples.

With the previous seeking behavior, this was a complete fix, since maxPollRecords put a bound on how far the spout could read from the commit offset in a poll that was ignoring maxUncommittedOffsets. With the new seeking behavior this is broken. If the spout has emitted as many tuples as allowed, and the last (highest offset) tuple fails, the spout may now poll for a full batch of new tuples, starting from the failed tuple. This scenario can repeat arbitrarily many times, so maxUncommittedOffsets is completely ineffective. 

We don't want to go back to the old seeking behavior (IMO), because it meant that in the case where maxPollRecords is much lower than maxUncommittedOffsets (almost always), the spout might end up choking on failed tuples. For example, if maxPollRecords is 5, and tuple 0-4 are not ready for retry (they might have been retried already, and are now waiting for retry backoff), but tuple 5-9 are, the spout is unable to retry 5-9 (or anything else on that partition) because it keeps seeking back to 0, and polling out the first 5 tuples. Seeking directly to the retriable tuples should in most cases be more efficient as well, because in the old implementation we'd just be seeking to the last committed offset, polling, and discarding tuples until we reach the ones that can be retried.

We could probably fix the broken behavior by trying really hard not to emit new tuples when we're ignoring maxUncommittedOffsets, but that seems like it would be error prone and complicated to implement.

I think we might be able to fix this by ensuring that we don't "doublecount" retriable tuples. When the spout is deciding whether to poll, it should deduct retriable tuples from numUncommittedOffsets when comparing to maxUncommittedOffsets.

Changing the poll check in this way is the same as enforcing the following constraint per partition, it seems to me:
* Poll only if `numNonRetriableEmittedTuples < maxUncommittedOffsets`. If there are more nonretriable tuples than that, the poll won't be allowed because `numUncommittedOffsets = numRetriableTuples + numNonRetriableEmittedTuples`, so `numUncommittedOffsets - numRetriableTuples >= maxUncommittedOffsets`. 

This should mean that the limit on uncommitted tuples on each partition is going to be `maxUncommittedOffsets + maxPollRecords - 1`, because the latest tuple that can be retried on a partition is the one at offset `maxUncommittedOffsets`, where there are `maxUncommittedOffsets - 1` uncommitted tuples "to the left". If the retry poll starts at that offset, it at most emits the retried tuple plus `maxPollRecords - 1` new tuples.

There shouldn't be any problems when multiple partitions have retriable tuples, where retriable tuples on one partition might be able to cause a different partition to break the uncommitted offset limit. This is because a partition will at minimum contribute 0 to numUncommittedOffsets (e.g. if all uncommitted tuples on that partition are retriable), because any retriable tuples being subtracted were already counted in numUncommittedOffsets when the tuples were originally emitted.

If we can enforce the limit on a per partition basis this way, there's no reason to worry about only emitting retriable tuples when we're exceeding maxUncommittedOffsets. 

I don't think there's a need for pausing partitions anymore either. It was meant to prevent polling for new tuples when there were retriable tuples, but we're no longer trying to prevent that, since the per partition cap is already ensuring we won't emit too many tuples. Pausing in this case would prioritize retriable tuples over new tuples (e.g. in the case where an unpaused consumer might choose to fetch from a nonretriable partition even though there are retriable tuples), but might lead to lower throughput overall (in the case where there are not enough messages on the retriable partitions to fill a batch). I've removed it again.

I've put up what I hope is the fix both here and on the 1.x branch. Sorry again that this has turned into such a moving target.

srdo:
@HeartSaVioR @hmcl Do you guys have time to take another look? (again, very sorry about the number of times I've declared this "done")

pavankumarp89:
Can you please let us know if this fix will be applied to lower version 1.0.4 or 1.0.3

srdo:
@pavankumarp89 It should probably go in 1.2.0 or something like that. I definitely want it backported to 1.x, but I think we're trying to avoid API breakage in patch versions, and this PR changes the RetryService API.

srdo:
Squashed and rebased

harshach:
Thanks @srdo for your patience. Merged into 1.x & master.

PR #1928, opened 2017-02-07
Create  ZippedTextFileReader.java

megha10:
This is to read from a zip file

### DIFF
--- a/external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/ZippedTextFileReader.java
+++ b/external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/ZippedTextFileReader.java
@@ -1,0 +1,241 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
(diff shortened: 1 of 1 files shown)


tibkiss:
Thanks for contributing, nice idea!

Seems like the build has failed due to missing symbols in ZippedTextFileReader.java:
https://travis-ci.org/apache/storm/jobs/199168284

Unit (and possibly integration) tests would be really beneficial so that we can ensure quality in a long run.

It is a protocol that first a JIRA issue is created for the feature/fix, then a PR is posted including the JIRA Issue ID in the commit message.

Could you please address these issues?

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/ZippedTextFileReader.java
```diff
@@ -0,0 +1,241 @@
...
+		// Initialize start and end markers for the current split.
+		start = split.getStart();
+		end = start + split.getLength();
+
+		final Path path = split.getPath(); // Filename of the current split.
+		// key.set(path.getName()); // Set the current filename as key
+		zipFilename = path.getName();
+
+		LOG.info("zipFilename: ============:" + zipFilename);
+		System.out.println("zipFilename: ============:" + zipFilename);
```
pasalkarsachin1: Can you remove syso

megha10:
Sure will do that :)

PR #1930, opened 2017-02-08
Replace kafka-unit by kafka_2.11 and kafka-clients for test

lukess:
Hi:
This pr is for testing kafka-clients:0.10.1.1 on storm-kafka-client. I replace the kafka-unit by kafka_2.11 and kafka-clients.

### DIFF
--- a/external/storm-kafka-client/pom.xml
+++ b/external/storm-kafka-client/pom.xml
@@ -86,9 +86,23 @@
             <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>info.batey.kafka</groupId>
-            <artifactId>kafka-unit</artifactId>
-            <version>0.6</version>
+            <groupId>org.apache.kafka</groupId>
+            <artifactId>kafka_2.11</artifactId>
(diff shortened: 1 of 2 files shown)


srdo:
Hi. You should raise an issue at https://issues.apache.org/jira/browse/STORM so the fix can be tracked. Then rename the PR to mention the jira issue number (take a look at the other open PRs to see what it usually looks like). You should also open a PR for master first, then the fix can be cherry picked back to 1.x-branch once it goes in.

About the contents of this PR: I think getting rid of kafka-unit is a good idea. It uses deprecated Kafka APIs. That said, I'd prefer if we could replace it with something similar since we're likely going to be using embedded Kafka for other tests in the future, and copy-pasting code to other test classes would get tedious. Would you be up for wrapping the setup/teardown code for the Kafka broker in a junit TestRule (https://github.com/junit-team/junit4/wiki/Rules#custom-rules)?

lukess:
@srdo thank you for describe the details of the process. I will do that.

lukess:
create a new pr from master branch - https://github.com/apache/storm/pull/1935

PR #1931, opened 2017-02-08
Convert NoNodeException to KeyNotFoundException in getNimbodesWithLatestSequenceNumberOfBlob

HeartSaVioR:
* since callers are able to handle KeyNotFoundException but not NoNodeException

This patch is addressing missed spot on [STORM-2321](https://issues.apache.org/jira/browse/STORM-2321).

It doesn't mean patch for STORM-2321 is broken (just found a missing case), so it doesn't hurt the release vote for 1.0.3 RC2.


harshach:
+1

HeartSaVioR:
Merged into master, 1.x, 1.0.x (target for 1.0.4) branches.

PR #1934, opened 2017-02-09
STORM-2333: CGroup memory and CPU metrics

revans2:


storm-core/src/jvm/org/apache/storm/metric/SystemBolt.java
```diff
@@ -85,14 +87,15 @@ public Object getValueAndReset() {
...
 
+    @SuppressWarnings({ "unchecked", "rawtypes" })
     @Override
-    public void prepare(final Map stormConf, TopologyContext context, OutputCollector collector) {
-        if(_prepareWasCalled && !"local".equals(stormConf.get(Config.STORM_CLUSTER_MODE))) {
+    public void prepare(final Map topoConf, TopologyContext context, OutputCollector collector) {
+        if(_prepareWasCalled && !"local".equals(topoConf.get(Config.STORM_CLUSTER_MODE))) {
             throw new RuntimeException("A single worker should have 1 SystemBolt instance.");
         }
         _prepareWasCalled = true;
```
pczb: is storm prepare method thread safe?

storm-core/src/jvm/org/apache/storm/metric/cgroup/CGroupCpu.java
```diff
@@ -0,0 +1,68 @@
...
+                    }
+                }   
+            }
+            long user = userHz - previousUser;
+            long sys = systemHz - previousSystem;
+            previousUser = userHz;
+            previousSystem = systemHz;
+            long hz = getUserHZ();
+            HashMap<String, Long> ret = new HashMap<>();
+            ret.put("user-ms", user * 1000/hz); //Convert to millis
```
tibkiss: This is a integer division during ms conversion. A float division would produce better resolution. It is surely not a problem with getUserHz() returning 100, but might cause issues if 100 is changed in the future.

revans2: I did this in part to match the https://github.com/apache/storm/blob/master/external/storm-metrics/src/main/java/org/apache/storm/metrics/sigar/CPUMetric.java format, which offers ms level metrics, not floating point, but just for the JVM, not for the entire container. 

I'll try to add documentation to explain the level of granularity that this supports and some of it's limitations.  I am missing documentation all together after all.

tibkiss: Makes sense, thanks for the explanation.

storm-core/src/jvm/org/apache/storm/metric/cgroup/CGroupMemoryLimit.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+public class CGroupMemoryLimit extends CGroupMetricsBase<Long> {
+
+    public CGroupMemoryLimit(Map<String, Object> conf) {
+        super(conf, "memory.limit_in_bytes");
+    }
+
+    @Override
+    public Long parseFileContents(String contents) {
+        return Long.parseLong(contents.trim());
```
jerrypeng: Many of these cgroup getting functions have been implemented in the cgroup package:

https://github.com/apache/storm/tree/master/storm-core/src/jvm/org/apache/storm/container/cgroup/core


for example there is already a method of get memory.limit_in_bytes:

https://github.com/apache/storm/blob/master/storm-core/src/jvm/org/apache/storm/container/cgroup/core/MemoryCore.java#L133

storm-core/src/jvm/org/apache/storm/metric/cgroup/CGroupCpuGuarantee.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+public class CGroupCpuGuarantee extends CGroupMetricsBase<Long> {
+    long previousTime = -1;
+    
+    public CGroupCpuGuarantee(Map<String, Object> conf) {
+        super(conf, CpuCore.CPU_SHARES);
+    }
+
+    @Override
+    public Long parseFileContents(String contents) {
```
jerrypeng: Function already defined:

https://github.com/apache/storm/blob/master/storm-core/src/jvm/org/apache/storm/container/cgroup/core/CpuCore.java#L50

jerrypeng:
@revans2 this is a really cool feature and thanks for adding it.  My only question is that I see you have implemented some parsers/getters for the cgroup metrics.  There is already a set of them implemented in the cgroup package.  Is there a reason why we shouldn't reuse them?

storm-core/src/jvm/org/apache/storm/metric/cgroup/CGroupMemoryLimit.java
revans2: I'll look into possibly using it, but right now most of the code that would be replaced is the code in CGroupMetricsBase that slurps the file.

revans2: I have been looking into it and it would be a lot of change just to save a few lines of code.  If you really think it is important to make the code common I will, but I would prefer to leave it.

revans2: Actually it might be worth it, just to allow it to support different CGroup hierarchies better.  I'll keep trying

revans2:
@jerrypeng Please take another look.  I refactored the code to reuse the CgroupManager and such.

docs/cgroups_in_storm.md
```diff
@@ -43,13 +43,24 @@ group storm {
...
 For a more detailed explanation of the format and configs for the cgconfig.conf file, please visit:
 
 https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/6/html/Resource_Management_Guide/ch-Using_Control_Groups.html#The_cgconfig.conf_File
 
+To let storm manage the cgroups for indavidual workers you need to make sure that the resources you want to control are mounted under the same directory as in the example above.
+If they are not in the same directory the supervisor will throw an exception.
+
+The perm section needs to be configured so that the user the supervisor is running as can modify the group.
+
+If run as user is enabled so the supervisor spawns other processes as the user that launched the topology make sure that the permissions are such that indavidual users have read access but not write access.
```
jerrypeng: perhaps more clean 

if "run as user"is enabled so **that** the supervisor spawns other...

storm-core/src/jvm/org/apache/storm/metric/cgroup/CGroupCpu.java
```diff
@@ -0,0 +1,70 @@
...
+            String line = in.readLine().trim();
+            userHz = Integer.valueOf(line);
+        }
+        return userHz;
+    }
+
+    @Override
+    public Map<String, Long> getDataFrom(CgroupCore core) throws IOException {
+        CpuacctCore cpu = (CpuacctCore) core;
+        Map<StatType, Long> stat = cpu.getCpuStat();
```
jerrypeng: is there a reason why we not doing a try catch of an IOException here similar for cpu.getCpuShares() below but just propagating the exception?

storm-core/src/jvm/org/apache/storm/metric/cgroup/CGroupMemoryLimit.java
```diff
@@ -0,0 +1,39 @@
...
+ * Reports the current memory limit of the cgroup for this worker
+ */
+public class CGroupMemoryLimit extends CGroupMetricsBase<Long> {
+
+    public CGroupMemoryLimit(Map<String, Object> conf) {
+        super(conf, SubSystemType.memory);
+    }
+
+    @Override
+    public Long getDataFrom(CgroupCore core) throws Exception {
```
jerrypeng: Same question here but also this function throws an Exception except for IOException

storm-core/src/jvm/org/apache/storm/metric/cgroup/CGroupMemoryUsage.java
```diff
@@ -0,0 +1,39 @@
...
+ * Reports the current memory usage of the cgroup for this worker
+ */
+public class CGroupMemoryUsage extends CGroupMetricsBase<Long> {
+
+    public CGroupMemoryUsage(Map<String, Object> conf) {
+        super(conf, SubSystemType.memory);
+    }
+
+    @Override
+    public Long getDataFrom(CgroupCore core) throws Exception {
```
jerrypeng: same question here

jerrypeng:
thanks @revans2 for taking the time to refactor some of the code!  I reviewed it once more.  If there is an example of how to user the metrics in the docs I think that would be extremely helpful to users.

storm-core/src/jvm/org/apache/storm/metric/cgroup/CGroupCpu.java
revans2: All of the CGroupCore subclasses throw an IOException.  I thought it would be better to have the parent class handle it rather then have the children each have the same code copied and pasted.   On some I guess I forgot to clean it up.  Will do.

storm-core/src/jvm/org/apache/storm/metric/cgroup/CGroupMemoryLimit.java
revans2: ^C^V

storm-core/src/jvm/org/apache/storm/metric/cgroup/CGroupMemoryUsage.java
revans2: ^C^V

revans2:
@jerrypeng I added in the extra documentation you wanted, and I spell checked the whole thing.

I also updated the CGroupCpuGuarantee metric to let the parent handle the exception.

jerrypeng:
Thanks @revans2 +1

revans2:
@jerrypeng Actually just found a bug working on a fix now....

revans2:
@jerrypeng Fixed the issue.  We were only checking if the SybSystem was enabled, but not if it was mounted.

jerrypeng:
Thanks @revans2 +1

PR #1936, opened 2017-02-12
STORM-2354: Upgrade to datastax driver 3.1.3

gilcesarf:
implemented some fixes to allow using driver 3.1.3. 
This TODO is also commented in the code as a possible new fix:

// TODO: codecFor(Object) return first codec that accepts value.
// perhaps we could provide a way to the programmer give more info
// to allow other forms of codedFor more suitable to the case we
// known the the correct CSQL types involved. Anyway, does not
// impact when Token-aware Load Balancing Policy is not used.


HeartSaVioR:
As I commented to the JIRA issue, this is duplicated. Could you please close this PR?

gilcesarf:
Closing. Thank you.

PR #1939, opened 2017-02-14
STORM-1363: TridentKafkaState should handle null values from TridentTupleToKafkaMapper.getMessageFromTuple()

pasalkarsachin1:
In case null value comes from the mapper it will print warning messages.
Added log to print the time taken to emit number of messages.


HeartSaVioR:
@pasalkarsachin1 Could you preserve the style? I'm sorry we don't have clear style guide but most of changeset is from spaces.

pasalkarsachin1:
I haven't changed anything it was I guess previous one had style. You can ignore white space by appending ?w=1 to url. e.g. https://github.com/apache/storm/pull/1939/files?w=1 This shows only updated code

HeartSaVioR:
@pasalkarsachin1 
We don't use 8 spaces for indentation from other java files. If the indentation spaces was 4, it should be 4 unless we set brand new style guide. Code style is important as much as code change.

pasalkarsachin1:
@HeartSaVioR  Done

HeartSaVioR:
@pasalkarsachin1 
Indentation is still off. Could you please check the diff on PR?
https://github.com/apache/storm/pull/1939/files

pasalkarsachin1:
@HeartSaVioR  I did my best. Can you take a look at it now

pasalkarsachin1:
Can someone take a look at this?

pasalkarsachin1:
@HeartSaVioR  Can you take a look at it

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/trident/TridentKafkaState.java
```diff
@@ -98,14 +107,19 @@ public void updateState(List<TridentTuple> tuples, TridentCollector collector) {
...
             if(exceptions.size() > 0){
-                String errorMsg = "Could not retrieve result for messages " + tuples + " from topic = " + topic 
-                        + " because of the following exceptions: \n";
-                for (ExecutionException exception : exceptions) {
-                    errorMsg = errorMsg + exception.getMessage() + "\n";
-                }
-                LOG.error(errorMsg);
-                throw new FailedException(errorMsg);
-            }
+		StringBuilder errorMsg = new StringBuilder("Could not retrieve result for messages ").append(tuples)
```
roshannaik: Java auto converts string concats to StringBuilder. There is no need for this. See discussion here for when is useful to use StringBuilder
http://stackoverflow.com/questions/11942368/why-use-stringbuilder-explicitly-if-the-compiler-converts-string-concatenation-t

pasalkarsachin1: If you read statement in stack overflow
"However StringBuilder is useful if you are concatenating a large number of strings in a loop. It requires O(n2) time to run and creates many temporary strings."
 
We also have same case. 
```java
for (ExecutionException exception : exceptions) {
 	errorMsg = errorMsg.append(exception.getMessage()).append(System.lineSeparator()); ;
 }
```
So rather than having String above loop & then recreate StringBuilder object & created it initially.
StringBuilder object

roshannaik: makes sense.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/trident/TridentKafkaState.java
```diff
@@ -98,14 +107,19 @@ public void updateState(List<TridentTuple> tuples, TridentCollector collector) {
...
+			.append(" from topic = ").append(topic).append(" because of the following exceptions: ").append(System.lineSeparator());
+				
+		for (ExecutionException exception : exceptions) {
+			errorMsg = errorMsg.append(exception.getMessage()).append(System.lineSeparator()); ;
+		}
+		String message = errorMsg.toString();
+		LOG.error(message);
+		throw new FailedException(message);
+	    }
+	    long latestTime=System.currentTimeMillis();
```
HeartSaVioR: Code style is not consistent even your changes. spaces before and after '='.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/trident/TridentKafkaState.java
```diff
@@ -98,14 +107,19 @@ public void updateState(List<TridentTuple> tuples, TridentCollector collector) {
...
+	    LOG.info("Emitted record {} sucessfully in {} ms to topic {} ", new Object[]{emittedRecords,latestTime-currentTime,topic});
```
HeartSaVioR: I think it receives varargs, so no necessary to group into `Object []` explicitly.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/trident/TridentKafkaState.java
HeartSaVioR: Personally I'd rather using StringBuilder to only concat exception messages. For me current implementation is better for readability, and compiler is able to care of this. 
(Please note that those lines are not - and should not be - hot spot, since code block will be executed only when there're some exceptions.)

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/trident/TridentKafkaState.java
```diff
@@ -75,20 +75,29 @@ public void prepare(Properties options) {
...
                 topic = topicSelector.getTopic(tuple);
-
+                Object messageFromTuple = mapper.getMessageFromTuple(tuple);
+		 Object keyFromTuple = mapper.getKeyFromTuple(tuple);
+				
                 if(topic != null) {
-                    Future<RecordMetadata> result = producer.send(new ProducerRecord(topic,
-                            mapper.getKeyFromTuple(tuple), mapper.getMessageFromTuple(tuple)));
-                    futures.add(result);
+                   if(messageFromTuple!=null){
```
HeartSaVioR: space between ')' and '{', and left side of '!=' and right side of '!='. may be better to add space between 'f' and '('.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/trident/TridentKafkaState.java
```diff
@@ -75,20 +75,29 @@ public void prepare(Properties options) {
...
+		      Future<RecordMetadata> result = producer.send(new ProducerRecord(topic,keyFromTuple, messageFromTuple));
+		      futures.add(result);
+		   } else {
+		      LOG.warn("skipping Message with Key "+keyFromTuple+" as message was null");
```
HeartSaVioR: spaces between `"` and `+` / and `+` and `"`

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/trident/TridentKafkaState.java
```diff
@@ -75,20 +75,29 @@ public void prepare(Properties options) {
     public void updateState(List<TridentTuple> tuples, TridentCollector collector) {
         String topic = null;
         try {
-            List<Future<RecordMetadata>> futures = new ArrayList<>(tuples.size());
+            long currentTime=System.currentTimeMillis();
```
HeartSaVioR: 1. `startTime` would be better to represent its intention well
2. space between `e` and `=` and `=` and `S`

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/trident/TridentKafkaState.java
HeartSaVioR: Please keep log simple for INFO, since it will be called for every updates. Just leaving record count is enough.
If you really want to leave these records, this should be DEBUG.

pasalkarsachin1: In our production currently there is no way you get information of how many records are getting pushed from logs. This log has turned out to provide at least so information on records pushed to Kafka in what time. This also helps who took time whether Kafka was slow in responding or we have large chunk of messages sent to Kafka

HeartSaVioR: My bad ignore this. I was misunderstanding it as actual payload, but was just count. It should be OK.

pasalkarsachin1:
@HeartSaVioR  I have updated code as per your comment :)

HeartSaVioR:
+1
@pasalkarsachin1 Could you squash commits into one? Thanks in advance!

pasalkarsachin1:
@HeartSaVioR  Squashed into one commit. Commit sha is 7cc1bdbea562a48aa09cc6dc7fd29edd5bd4eb10

pasalkarsachin1:
@HeartSaVioR  Can you merge this so I will close this PR :) Thanks in advance

HeartSaVioR:
@pasalkarsachin1 
Merged into master via 313e960bedde3b2c47520a74c391425d79271727, and 1.x branch via 4fc55b33504d446ef192a25e5164f861e1495291.
Could you close this PR? Thanks in advance.

pasalkarsachin1:
Thanks for merge

PR #1940, opened 2017-02-14
STORM-2361 Kafka spout - after leader change, it stops committing offsets to ZK

ernisv:
Try to route acks/fails to recreated partition managers if manager is not found by exact (broker, partition, topic) combination

### DIFF
--- a/external/storm-kafka/src/jvm/org/apache/storm/kafka/KafkaSpout.java
+++ b/external/storm-kafka/src/jvm/org/apache/storm/kafka/KafkaSpout.java
@@ -157,12 +157,27 @@ public class KafkaSpout extends BaseRichSpout {
         }
     }
 
+    private PartitionManager getManagerForPartition(int partition) {
+        for (PartitionManager partitionManager: _coordinator.getMyManagedPartitions()) {
+            if (partitionManager.getPartition().partition == partition) {
+                return partitionManager;
+            }
+        }
+        return null;
+    }
+
     @Override
     public void ack(Object msgId) {
         KafkaMessageId id = (KafkaMessageId) msgId;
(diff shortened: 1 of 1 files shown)


ptgoetz:
@ernisv Can you update the code to put all for, if-else, etc. blocks curly braces?

ernisv:
Sure - braces added.

external/storm-kafka/src/jvm/org/apache/storm/kafka/KafkaSpout.java
```diff
@@ -172,6 +187,12 @@ public void fail(Object msgId) {
         PartitionManager m = _coordinator.getManager(id.partition);
         if (m != null) {
             m.fail(id.offset);
+        } {
```
HeartSaVioR: missing `else` between `} {`

ernisv: Right - fixed.

HeartSaVioR:
+1 
@ernisv Could you squash commits into one? 
Btw, I feel this patch should be along with STORM-2296, so if we would want to port this back to 1.0.x, STORM-2296 should be ported back first.

ernisv:
Squashed the commits.
About moving to STORM-2296 - I've put the links/relations to STORM-2296.
But as the issue itself is already in "resolved" state - not sure how the patch could be added to it.
Is registering the relation not enough?

HeartSaVioR:
@ernisv 
Could you change your commit title to be meaningful? We often set commit title to same as JIRA title.
I can take care of other things so you don't need to worry about that.

ernisv:
Ok, changed the title

PR #1950, opened 2017-02-21
STORM-2369 [storm-redis] Use binary type for State management (1.x)

HeartSaVioR:
* introduce new command interface and relevant containers
  * new command interface will contain both binary and string commands once we address STORM-2304
* change RedisKeyValue to use byte[] for State key and value management instead of String
  * get rid of Base64 encode/decode on RedisEncoder, and also SafeEncoder.encode() in Jedis internal
* implement some utils for handling Map with byte[]: since byte[] is bad for key type of Map
* users can select either Redis single server or Redis Cluster by passing JedisPoolConfig or JedisClusterConfig

PR for master: #2172

### DIFF
--- a/docs/State-checkpointing.md
+++ b/docs/State-checkpointing.md
@@ -70,6 +70,48 @@ json config with the following properties.
     }
 }
 ```
+ 
+For Redis Cluster state this is a json config with the following properties.
+ 
+```
+ {
+   "keyClass": "Optional fully qualified class name of the Key type.",
+   "valueClass": "Optional fully qualified class name of the Value type.",
+   "keySerializerClass": "Optional Key serializer implementation class.",
+   "valueSerializerClass": "Optional Value Serializer implementation class.",
+   "jedisClusterConfig": {
+     "nodes": ["localhost:7379", "localhost:7380", "localhost:7381"],
+     "timeout": 2000,
+     "maxRedirections": 5
+   }
+ }
+```
+
+NOTE: If you used Redis state with Storm version 1.1.0 or earlier, you would need to also migrate your state since the representation of state has changed  
+from Base64-encoded string to binary to reduce huge overhead. Storm provides a migration tool to help, which is placed on `storm-redis-example` module.
+
+Please download the source from download page or clone the project, and type below command:
+
+```
+mvn clean install -DskipTests
+cd examples/storm-redis-examples
+<storm-installation-dir>/bin/storm jar target/storm-redis-examples-*.jar org.apache.storm.redis.tools.Base64ToBinaryStateMigrationUtil [options]
+```
+
+Supported options are listed here:
+
+```
+ -d,--dbnum <arg>       Redis DB number (default: 0)
+ -h,--host <arg>        Redis hostname (default: localhost)
+ -n,--namespace <arg>   REQUIRED the list of namespace to migrate.
+ -p,--port <arg>        Redis port (default: 6379)
+    --password <arg>    Redis password (default: no password)
+```
+
+You can provide multiple `namespace` options to migrate multiple namespaces at once. 
+(e.g.: `--namespace total-7 --namespace partialsum-3`)
+Other options are not mandatory.
+Please note that you need to also migrate the key starting with "$checkpointspout-" since it's internal namespace of state. 
 
 ## Checkpoint mechanism
 Checkpoint is triggered by an internal checkpoint spout at the specified `topology.state.checkpoint.interval.ms`. If there is
--- a/external/storm-redis/src/main/java/org/apache/storm/redis/common/commands/RedisCommands.java
+++ b/external/storm-redis/src/main/java/org/apache/storm/redis/common/commands/RedisCommands.java
@@ -1,0 +1,60 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.storm.redis.common.commands;
+
+import redis.clients.jedis.ScanParams;
+import redis.clients.jedis.ScanResult;
+
+import java.util.Map;
+
+/**
+ * This interface represents Jedis methods exhaustively which are used on storm-redis.
+ *
+ * This is a workaround since Jedis and JedisCluster doesn't implement same interface for binary type of methods,
+ * and unify binary methods and string methods into one interface.
+ */
+public interface RedisCommands {
+    // common
+    Boolean exists(byte[] key);
+
+    boolean exists(String key);
+
+    Long del(byte[] key);
+
+    Long del(String key);
+
+    String rename(byte[] oldkey, byte[] newkey);
+
+    String rename(String oldkey, String newkey);
+
+    // hash
+    byte[] hget(byte[] key, byte[] field);
+
+    Map<byte[], byte[]> hgetAll(byte[] key);
+
+    Map<String,String> hgetAll(String key);
+
+    String hmset(byte[] key, Map<byte[], byte[]> fieldValues);
+
+    String hmset(String key, Map<String, String> fieldValues);
+
+    Long hdel(byte[] key, byte[]... fields);
+
+    ScanResult<Map.Entry<byte[], byte[]>> hscan(byte[] key, byte[] cursor, ScanParams params);
+}
--- a/external/storm-redis/src/main/java/org/apache/storm/redis/state/RedisKeyValueState.java
+++ b/external/storm-redis/src/main/java/org/apache/storm/redis/state/RedisKeyValueState.java
@@ -17,24 +17,31 @@
  */
 package org.apache.storm.redis.state;
 
+import com.google.common.collect.Maps;
+import com.google.common.primitives.UnsignedBytes;
+import org.apache.storm.redis.common.commands.RedisCommands;
+import org.apache.storm.redis.common.config.JedisClusterConfig;
+import org.apache.storm.redis.common.container.RedisCommandsContainerBuilder;
+import org.apache.storm.redis.common.container.RedisCommandsInstanceContainer;
+import org.apache.storm.state.DefaultStateEncoder;
 import org.apache.storm.state.DefaultStateSerializer;
 import org.apache.storm.state.KeyValueState;
 import org.apache.storm.state.Serializer;
 import org.apache.storm.redis.common.config.JedisPoolConfig;
-import org.apache.storm.redis.common.container.JedisCommandsContainerBuilder;
-import org.apache.storm.redis.common.container.JedisCommandsInstanceContainer;
-import org.apache.storm.redis.utils.RedisEncoder;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import redis.clients.jedis.JedisCommands;
+import redis.clients.util.SafeEncoder;
 
-import java.util.Collections;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.List;
 import java.util.ArrayList;
-import java.util.concurrent.ConcurrentHashMap;
+import java.util.NavigableMap;
+import java.util.TreeMap;
+import java.util.concurrent.ConcurrentNavigableMap;
+import java.util.concurrent.ConcurrentSkipListMap;
 
 /**
  * A redis based implementation that persists the state in Redis.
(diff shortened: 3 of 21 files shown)


HeartSaVioR:
Tested storm-starter's StatefulTopology with Redis Cluster nodes manually.

HeartSaVioR:
Might need to provide migration tool to migrate Base64 encoded states to be conformed to current implementation.

HeartSaVioR:
@arunmahadevan 
Please take a look when you have some time. Thanks in advance!

arunmahadevan:
@HeartSaVioR what is the motivation for using binary instead of string type ? Is this mainly for performance ? I will review this sometime next week.

HeartSaVioR:
@arunmahadevan 
We could save 
- cost of serde. on KV pair
- another serde. for String to byte[] while sending/receiving from/to Redis
- traffic/space of Redis since base64 converts KV to much longer

external/storm-redis/src/main/java/org/apache/storm/redis/state/RedisKeyValueState.java
```diff
@@ -152,70 +170,72 @@ public V get(K key, V defaultValue) {
...
-                for (Map.Entry<String, String> e: pendingCommit.entrySet()) {
+                for (Map.Entry<ByteArrayWrapper, byte[]> e: pendingCommit.entrySet()) {
                     if (!currentPending.containsKey(e.getKey())) {
                         currentPending.put(e.getKey(), e.getValue());
                     }
                 }
             }
             if (!currentPending.isEmpty()) {
-                commands.hmset(prepareNamespace, currentPending);
+                commands.hmset(prepareNamespace, ByteArrayUtil.Maps.newHashMapUnwrappingKey(currentPending));
```
arunmahadevan: maybe you can avoid the wrapping/unwrapping and extra copying if you use a TreeMap with a comparator that compares the byte array. (e.g. `ConcurrentSkipListMap<>(com.google.common.primitives.UnsignedBytes.lexicographicalComparator())` since we want concurrency).

external/storm-redis/src/main/java/org/apache/storm/redis/utils/ByteArrayUtil.java
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    public static class Maps {
+        public static Map<ByteArrayWrapper, byte[]> newHashMapWrappingKey(Map<byte[], byte[]> source) {
+            Map<ByteArrayWrapper, byte[]> destination = new HashMap<>(source.size());
+            wrapEntities(source, destination);
+            return destination;
+        }
+
+        public static Map<ByteArrayWrapper, byte[]> newConcurrentHashMapWrappingKey(Map<byte[], byte[]> source) {
```
HeartSaVioR: Yes I'll remove that.

external/storm-redis/src/main/java/org/apache/storm/redis/common/commands/RedisCommands.java
```diff
@@ -0,0 +1,56 @@
...
+    String hmset(byte[] key, Map<byte[], byte[]> fieldValues);
+
+    Map<byte[], byte[]> hgetAll(byte[] key);
+
+    Long hdel(byte[] key, byte[]... fields);
+
+    ScanResult<Map.Entry<byte[], byte[]>> hscan(byte[] key, byte[] cursor, ScanParams params);
+
+    // string, common
+    boolean exists(String key);
```
HeartSaVioR: Good suggestion. I'll group them per command.

external/storm-redis/src/main/java/org/apache/storm/redis/state/RedisKeyValueState.java
HeartSaVioR: Great suggestion. I thought it is bad but didn't know about alternatives. I'll try to replace that one.

HeartSaVioR:
Agreed to leave a note regarding that, and I'm OK to provide a tool to convert current state since state KVs would be valuable for users. I'll try to handle that.

HeartSaVioR:
@arunmahadevan 
Sorry got really late for addressing review comments. 

Addressed comments, and created migration tool `Base64ToBinaryStateMigrationUtil`.
(Succeed to read State with this patch from migrated State.)

I'll describe how to use to the doc soon.

HeartSaVioR:
@arunmahadevan 
Also updated the doc to explain how to use migration tool. Commits squashed. 
Please review again. Thanks in advance.

HeartSaVioR:
Rebased and updated state iterator implementation.
Also extracted base implementation of binary based state iterator into storm-core which could be used also from HBase state.

@arunmahadevan Please take a look when you have time. Thanks in advance.

storm-core/src/jvm/org/apache/storm/state/DefaultStateEncoder.java
```diff
@@ -15,27 +15,23 @@
...
-import org.apache.commons.codec.binary.Base64;
-import org.apache.storm.state.DefaultStateSerializer;
-import org.apache.storm.state.Serializer;
-
 /**
- * Helper class for encoding/decoding redis key values.
+ * Helper class for encoding/decoding key values.
  */
-public class RedisEncoder<K, V> {
+public class DefaultStateEncoder<K, V> {
```
arunmahadevan: Maybe this can the default implementation of a StateEncoder interface?

external/storm-redis/src/main/java/org/apache/storm/redis/state/RedisKeyValueState.java
```diff
@@ -64,75 +75,83 @@ public RedisKeyValueState(String namespace, JedisPoolConfig poolConfig) {
...
             if (commands.exists(prepareNamespace)) {
                 LOG.debug("Loading previously prepared commit from {}", prepareNamespace);
-                pendingCommit = Collections.unmodifiableMap(commands.hgetAll(prepareNamespace));
+                NavigableMap<byte[], byte[]> pendingCommitMap = createPendingCommitMap();
+                pendingCommitMap.putAll(commands.hgetAll(prepareNamespace));
+                pendingCommit = Maps.unmodifiableNavigableMap(pendingCommitMap);
             } else {
                 LOG.debug("No previously prepared commits.");
-                pendingCommit = Collections.emptyMap();
+                pendingCommit = createPendingCommitMap();
```
arunmahadevan: nit: can be `Collections.emptyNavigableMap()` (java8) or a static final variable initialized with an empty TreeMap

external/storm-redis/src/main/java/org/apache/storm/redis/state/RedisKeyValueState.java
```diff
@@ -187,70 +206,73 @@ public void prepareCommit(long txid) {
...
+                    commands.hdel(namespace, keysToDelete.toArray(new byte[0][]));
                 }
             } else {
                 LOG.debug("Nothing to save for commit, txid {}.", txid);
             }
             txIds.put(COMMIT_TXID_KEY, String.valueOf(txid));
             commands.hmset(txidNamespace, txIds);
             commands.del(prepareNamespace);
-            pendingCommit = Collections.emptyMap();
+            pendingCommit = createPendingCommitMap();
```
arunmahadevan: nit: can be `Collections.emptyNavigableMap()` (java8) or a static final variable initialized with an empty TreeMap

external/storm-redis/src/main/java/org/apache/storm/redis/state/RedisKeyValueState.java
```diff
@@ -266,10 +288,10 @@ public void rollback() {
                 LOG.debug("hmset txidNamespace {}, txIds {}", txidNamespace, txIds);
                 commands.hmset(txidNamespace, txIds);
             }
-            pendingCommit = Collections.emptyMap();
-            pendingPrepare = new ConcurrentHashMap<>();
+            pendingCommit = createPendingCommitMap();
```
arunmahadevan: nit: can be `Collections.emptyNavigableMap()` (java8) or a static final variable initialized with an empty TreeMap

storm-core/src/jvm/org/apache/storm/state/BaseBinaryStateIterator.java
```diff
@@ -0,0 +1,162 @@
...
+  public void remove() {
+    throw new UnsupportedOperationException();
+  }
+
+  /**
+   * Load some part of state KVs from storage and returns iterator of cached data from storage.
+   *
+   * @return Iterator of loaded state KVs
+   */
+  protected abstract Iterator<Map.Entry<byte[],byte[]>> loadChunkFromStateStorage();
```
arunmahadevan: Here it assumes the state implementations store in binary format? Can't it return `Iterator<Map.Entry<K, V>>` instead and not assume any specific implementation and let sub-classes do the decoding ?

storm-core/src/jvm/org/apache/storm/state/BaseBinaryStateIterator.java
```diff
@@ -0,0 +1,162 @@
...
+   */
+  protected abstract boolean isEndOfDataFromStorage();
+
+  /**
+   * Decode key to convert byte array to state key type.
+   *
+   * @param key byte array encoded key
+   * @return Decoded value of key
+   */
+  protected abstract K decodeKey(byte[] key);
```
arunmahadevan: Can we let the subclass deal with the specifics of decoding than assuming data is encoded as `byte[]`?

storm-core/src/jvm/org/apache/storm/state/BaseBinaryStateIterator.java
```diff
@@ -0,0 +1,162 @@
...
+
+  /**
+   * Decode value to convert byte array to state value type.
+   *
+   * @param value byte array encoded value
+   * @return Decoded value of value
+   */
+  protected abstract V decodeValue(byte[] value);
```
arunmahadevan: Can we let the subclass deal with the specifics of decoding than assuming data is encoded as `byte[]`?

docs/State-checkpointing.md
```diff
@@ -70,6 +70,46 @@ json config with the following properties.
...
+   "valueSerializerClass": "Optional Value Serializer implementation class.",
+   "jedisClusterConfig": {
+     "nodes": ["localhost:7379", "localhost:7380", "localhost:7381"],
+     "timeout": 2000,
+     "maxRedirections": 5
+   }
+ }
+```
+
+NOTE: If you used Redis state with Storm version 1.1.0 or earlier, you would need to also migrate your state since the representation of state has changed
```
HeartSaVioR: I think it will throw exception because state will not be converted properly, but there's no code to determine State version for now. Adding state version to the checkpoint spout state might be the one way to do this. I'll look into more about how we can handle this.

docs/State-checkpointing.md
```diff
@@ -70,6 +70,46 @@ json config with the following properties.
...
+
+```
+ -d,--dbnum <arg>       Redis DB number (default: 0)
+ -h,--host <arg>        Redis hostname (default: localhost)
+ -n,--namespace <arg>   REQUIRED the list of namespace to migrate.
+ -p,--port <arg>        Redis port (default: 6379)
+    --password <arg>    Redis password (default: no password)
+```
+
+You can provide multiple `namespace` options to migrate multiple namespaces at once. Other options are not mandatory.
```
HeartSaVioR: Yes that would be nice to add it. Will add.

storm-core/src/jvm/org/apache/storm/state/DefaultStateEncoder.java
HeartSaVioR: We're already exposing Serializer for key and value, so not sure we want to have another interface and implementations here. Do we want to have encoder for storage which doesn't support binary?

external/storm-redis/src/main/java/org/apache/storm/redis/state/RedisKeyValueState.java
HeartSaVioR: Good point. Will address. Btw I'll try to avoid JDK 8 thing right now unless it is only applied to the master branch.

external/storm-redis/src/main/java/org/apache/storm/redis/state/RedisKeyValueState.java
HeartSaVioR: Will address.

external/storm-redis/src/main/java/org/apache/storm/redis/state/RedisKeyValueState.java
HeartSaVioR: Will address.

storm-core/src/jvm/org/apache/storm/state/DefaultStateEncoder.java
arunmahadevan: Was thinking the `byte[]` is specific to Redis. Do we expect to encode in similar format for other implementations as well? Maybe we can refactor it later if we need different formats for other implementations.

storm-core/src/jvm/org/apache/storm/state/BaseBinaryStateIterator.java
HeartSaVioR: The type of pendingPrepare and pendingCommit, and storage should be same, so the type is bound to also constructor parameters. So in order to make this generic, we need to have two additional types for storage raw type. I'll see how it will be changed when it is generalized.

storm-core/src/jvm/org/apache/storm/state/BaseBinaryStateIterator.java
HeartSaVioR: Same above.

storm-core/src/jvm/org/apache/storm/state/BaseBinaryStateIterator.java
HeartSaVioR: Same above.

storm-core/src/jvm/org/apache/storm/state/DefaultStateEncoder.java
HeartSaVioR: HBase uses `byte[]` as raw type so we could use the same encoder to HBase State, too.

docs/State-checkpointing.md
HeartSaVioR: I'm trying to address this via adding `version` field to CheckpointState class, but stuck on migrating old checkpoint spout state to newer. Kyro doesn't support forward compatibility, and even seemed not be able to read with copied old class. I have no idea how to do it right now.

HeartSaVioR:
@arunmahadevan Thanks for the further review comments. I addressed your comments without versioning state.

external/storm-redis/src/main/java/org/apache/storm/redis/state/RedisKeyValueState.java
```diff
@@ -316,10 +340,18 @@ private Long lastPreparedTxid() {
...
+            lastId = Long.valueOf(txId);
         }
         return lastId;
     }
+
+    private ConcurrentNavigableMap<byte[], byte[]> createPendingPrepareMap() {
+        return new ConcurrentSkipListMap<>(UnsignedBytes.lexicographicalComparator());
+    }
+
+    private NavigableMap<byte[], byte[]> createPendingCommitMap() {
```
arunmahadevan: Seems this is not used. If so can be removed.

HeartSaVioR:
@arunmahadevan Thanks I've addressed latest review comment and also squashed.

HeartSaVioR:
@arunmahadevan Ah, please review #2172 as well when you revisit this. Thanks in advance!

arunmahadevan:
+1

PR #1952, opened 2017-02-21
STORM-2372: Pacemaker client doesn't clean up heartbeats properly. (1.x)

knusbaum:
Paths are not deleted correctly. Pacemaker's delete-path operates by matching a prefix against all the keys in the map.

The issue here is that the prefix is given a '/' on the end, but keys don't have a trailing '/' if there is no 'subkey'.

i.e. delete path `/foo/bar/baz/` doesn't match the key `/foo/bar/baz`
The path has to have the trailing '/' so that delete path `/foo/bar/baz` doesn't also delete `/foo/bar/bazoo`

The solution here is to tack on a '/' to every key when checking against the prefix.

We also want to send the delete command to every pacemaker server rather than just the normal write client.


ptgoetz:
+1

PR #1955, opened 2017-02-22
Add Streams API and Storm SQL docs to index page

HeartSaVioR:
Rearrange a bit and add documentations about Streams API and some missing Storm SQL.

PR #1956, opened 2017-02-22
Add Storm SQL docs to index page for 1.x branch

HeartSaVioR:
Rearrange a bit and add documentations about some missing Storm SQL.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -31,9 +31,6 @@ documentation: true
 
 ### Layers on Top of Storm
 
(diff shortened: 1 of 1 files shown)


HeartSaVioR:
It also needs to be addressed for 1.1.0, but since release vote is ongoing, I'll handle it manually if vote passes and site doc for 1.1.0 is up.

HeartSaVioR:
Since #1955 is superset of this, and it contains comments which can be also applied to this, I'll update this as well, and merge.

PR #1959, opened 2017-02-22
STORM-2374: Storm Kafka Client Test Topologies Must be Serializable

hmcl:


external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/test/KafkaSpoutTopologyMainNamedTopics.java
```diff
@@ -94,10 +98,10 @@ protected StormTopology getTopologyKafkaSpout() {
 
     protected KafkaSpoutConfig<String,String> getKafkaSpoutConfig() {
         ByTopicRecordTranslator<String, String> trans = new ByTopicRecordTranslator<>(
-                (r) -> new Values(r.topic(), r.partition(), r.offset(), r.key(), r.value()),
```
revans2: I find it surprising that the lambda function does not work here.  I remember explicitly checking to see if it would produce a static lambda or not, and I thought it would create a static one.

revans2:
I am OK with the change but if this is really causing issues then we need to update the docs too.

revans2:
OK Now I understand what is happening.  I think what we really want is to mark Func as Serializable.  Could you try the following patch instead?

```
diff --git a/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/Func.java b/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/Func.java
index a631d9608..041453393 100644
--- a/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/Func.java
+++ b/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/Func.java
@@ -17,10 +17,12 @@
  */
 package org.apache.storm.kafka.spout;

+import java.io.Serializable;
+
 /**
  * A simple interface to allow compatibility with non java 8
  * code bases
  */
-public interface Func<V, R> {
+public interface Func<V, R> extends Serializable {
     R apply(V record);
 }
```

It should work on both master and 1.x

hmcl:
@revans2 yes, I knew what was the problem. However, it's very strong and dangerous (as for backwards compatibility) to enforce base types (e.g interfaces) to implement `java.io.Serializable`. We do that a lot across the codebase, so I guess it is ok in this case as well. 

I was not sure we wanted to make `Func` to extend `Serializable` as it would automatically expose all the internals of all the implementations of this interface. That's the reason I did the proposed fix. If we agree that it's ok to have `Func` extend `Serializable`, I will do as you suggested. Let me know.

revans2:
@hmcl This code has not been released yet, or if it was we need to fix this bug in it.  This particular Func class was created specifically to make it simple to override specific places in the Kafka spout.  They will always need to be serialized out and sent across the wire to the worker, so making it Serializable is a desired thing.

hmcl:
@revans2 done. @ptgoetz this should be ready to merge. There is also a PR for 1.x-branch.

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/test/KafkaSpoutTopologyMainNamedTopics.java
HeartSaVioR: `(Values & Serializable)(r) -> new Values(r.topic(), r.partition(), r.offset(), r.key(), r.value()` might work, but still think Func should be Serializable.

HeartSaVioR:
@hmcl 
minor: We can remove `Serializable` from RecordTranslator since `Func` already extends it.

With or without the change mentioned above, +1.

I agree that adding Serializable to the existing interface is not safe, but given that `Func` is only used for RecordTranslator, then this approach is easier and clearer since most of implementations of RecordTranslator will have Func as field, and in that case `Func` should be serializable.

revans2:
+1

ptgoetz:
+1

PR #1962, opened 2017-02-24
STORM-2381: Add logging of JDBC connection string in storm-jdbc integration for debugging failures

tkuwabara789:
…ilures

Minor addition to expose JDBC connection string through logging in Hikari CP implementation.  For debugability purposes only.

### DIFF
--- a/external/storm-jdbc/src/main/java/org/apache/storm/jdbc/common/HikariCPConnectionProvider.java
+++ b/external/storm-jdbc/src/main/java/org/apache/storm/jdbc/common/HikariCPConnectionProvider.java
@@ -25,7 +25,11 @@ import java.sql.SQLException;
 import java.util.Map;
 import java.util.Properties;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 public class HikariCPConnectionProvider implements ConnectionProvider {
+    private static final Logger LOG = LoggerFactory.getLogger(HikariCPConnectionProvider.class);
 
     private Map<String, Object> configMap;
     private transient HikariDataSource dataSource;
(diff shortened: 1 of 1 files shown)


external/storm-jdbc/src/main/java/org/apache/storm/jdbc/common/HikariCPConnectionProvider.java
```diff
@@ -40,6 +44,10 @@ public synchronized void prepare() {
             Properties properties = new Properties();
             properties.putAll(configMap);
             HikariConfig config = new HikariConfig(properties);
+            if(properties.containsKey("dataSource.url"))
+                LOG.info("DataSource Url: " + properties.getProperty("dataSource.url"));
+            else if (config.getJdbcUrl() != null)
+                LOG.info("JDBC Url: " + config.getJdbcUrl());
```
tibkiss: Shouldn't this printout be independent of the presence of dataSource.url?
i.e. just an {{if}} instead of {{else if}}

tibkiss:
Thanks for this addition, looks reasonable. 

I'd recommend opening an issue first in jira, then posting this PR with the associated issue id.

external/storm-jdbc/src/main/java/org/apache/storm/jdbc/common/HikariCPConnectionProvider.java
tkuwabara789: The else-if is there because with datasource.className specified (e.g. com.mysql.jdbc.jdbc2.optional.MysqlDataSource), you either specify datasource.url (which takes precedence) or specify the jdbc property.

ptgoetz: @tkuwabara789 can you put the if-else block in curly braces? And also create a JIRA and update the title of this PR?

tkuwabara789: Sure thing, will do that now

ptgoetz:
+1

harshach:
+1

PR #1963, opened 2017-02-24
STORM-2380 worker.childopts with whitespace inside one param

nilday:
This pattern matches
1.everything starts with -XX:\w+ or -D[\w.]+ and followed with quoted (both ' and ") strings
2.everything without \s, ' and " in it
This will solve the problem which params like -XX:OnError="pstack %p >~/pstack%p.log" will be split into pieces

### DIFF
--- a/storm-core/src/jvm/org/apache/storm/daemon/supervisor/BasicContainer.java
+++ b/storm-core/src/jvm/org/apache/storm/daemon/supervisor/BasicContainer.java
@@ -33,6 +33,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.stream.Collectors;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import org.apache.commons.lang.StringUtils;
 import org.apache.storm.Config;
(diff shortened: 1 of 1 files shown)


storm-core/src/jvm/org/apache/storm/daemon/supervisor/BasicContainer.java
```diff
@@ -411,10 +413,17 @@ private String substituteChildOptsInternal(String string, int memOnheap) {
...
-                String[] strings = string.split("\\s+");
-                for (String s: strings) {
-                    if (StringUtils.isNotBlank(s)) {
-                        rets.add(s);
+                /* This pattern matches
+                 * 1.everything starts with -XX:\w+ or -D[\w.]+ and followed with quoted (both ' and ") strings
+                 * 2.everything without \s, ' and " in it
+                 * This will solve the problem which params like -XX:OnError="pstack %p >~/pstack%p.log" will be split into pieces
+                 */
+                Matcher m= Pattern.compile("(?:-XX:\\w+|-D[\\w.]+)=((?<![\\\\])['\\\"])((?:.(?!(?<![\\\\])\\1))*.?)\\1|[^\\s\\\"']+")
```
HeartSaVioR: Could we have various test cases to cover whether it's working?

HeartSaVioR:
@nilday Any updates?

PR #1968, opened 2017-02-25
STORM-2290: Upgrade zookeeper to get some critical upgrades

sachingsachin:
Complements https://github.com/apache/storm/pull/1967


r0perice:
In my job we are currently using storm 1.0.3. It will be great if in version 1.0.X of storm the last Zookeeper 3.4.X can be integrated. Their is a lot of bug fixes that we will enjoy to have :)

PR #1970, opened 2017-02-27
STORM-2383 Support HBase as state backend (1.x)

HeartSaVioR:
* Implement HBase state backend
  * picked Redis state backend implementation
  * Current implementation stores prepare/commit/txids into a column family in a table
    * prepare, commit, txids have its row
    * they're stored to qualifier:value map (not scalable but guarantee atomicity)
    * we can address atomicity vs scalable later
* Added 'how to set up' to State-checkpointing doc

Please note that `jackson-core` and `jackson-databind` are added. If we would want to shade this away I'll try to manage.

NOTE: Currently I can't run or debug tests on IntelliJ for master branch (so worked on 1.x-branch). I also heard many folks are struggling, so might be better to address it soon.


HeartSaVioR:
I chose this approach (all KVs to one cf in one row) because this is only guaranteed way for HBase to insert/delete atomically.

If we're OK to apply the changeset (in commit phase) without atomicity, we can make state scalable via having each row. Changeset still need to be put with atomic manner so that we don't have partial changeset as `prepared changeset`. Normally changeset don't grow too much so it doesn't hurt much for scalability.

HeartSaVioR:
The compilation failure is from STORM-2388. I'll rebase after STORM-2388 is merged.

external/storm-hbase/src/main/java/org/apache/storm/hbase/state/HBaseKeyValueState.java
```diff
@@ -0,0 +1,399 @@
...
+        LOG.debug("delete key '{}'", key);
+        byte[] columnKey = encoder.encodeKey(key);
+        V curr = get(key);
+        pendingPrepare.put(new ByteArrayWrapper(columnKey), HBaseEncoder.TOMBSTONE);
+        return curr;
+    }
+
+    @Override
+    public Iterator<Map.Entry<K, V>> iterator() {
+        // TODO: is it possible to implement this correctly?
```
HeartSaVioR: @arunmahadevan 
In Redis implementation, it can return same key multiple time, with different values, which is not what we expect from Iterator.

Maybe we can make Iterator not returning KV pair if key is already provided before. If you think this approach fixes above issue, I'll apply this to Redis implementation, and try to migrate it to HBase state.

HeartSaVioR:
@arunmahadevan 
OK I can change the implementation to store KV to each row. Will update the patch and let you know.

external/storm-hbase/src/main/java/org/apache/storm/hbase/state/HBaseKeyValueState.java
arunmahadevan: @HeartSaVioR nice observation. Is it because the entries in the pending iterators can get committed to the db (redis) while we are iterating ? If so we may be we need to fix it. While returning the values from the pending iterators we could store the keys into a set and use this to filter already returned keys while scanning the db (redis).

We put 'tombstone' entries into the pending maps during delete, but I don't see this handled during the iteration. May be we need to handle this as well.

Once the redis implementation is fixed, may be you can refactor and re-use the common parts for hbase as well.

HeartSaVioR: OK. Filed [STORM-2449](https://issues.apache.org/jira/browse/STORM-2449).
I'll apply the approach to this PR, too.

HeartSaVioR:
@arunmahadevan 
Thanks for reviewing. I addressed your review comments.
For iterator, let's handle it at once for Redis and HBase from STORM-2449 since it should be similar approach.

HeartSaVioR:
I'll implement HBase state iterator after #1950 is merged.

HeartSaVioR:
I'll also craft PR against master branch when implementing HBase state iterator.

HeartSaVioR:
I updated this PR to apply the last change of STORM-2369. I'll simply rebase again after STORM-2369 is merged. And will also craft the PR for master branch.

HeartSaVioR:
Just rebased with latest 1.x-branch.

HeartSaVioR:
@arunmahadevan Kindly request for review.

external/storm-hbase/src/main/java/org/apache/storm/hbase/state/HBaseKeyValueStateIterator.java
```diff
@@ -0,0 +1,153 @@
...
+    public HBaseKeyValueStateIterator(String namespace, byte[] columnFamily, HBaseClient hBaseClient,
+                                      Iterator<Map.Entry<byte[], byte[]>> pendingPrepareIterator,
+                                      Iterator<Map.Entry<byte[], byte[]>> pendingCommitIterator,
+                                      int chunkSize, Serializer<K> keySerializer,
+                                      Serializer<V> valueSerializer) {
+        super(pendingPrepareIterator, pendingCommitIterator);
+        this.columnFamily = columnFamily;
+        this.keyNamespace = (namespace + "$key:").getBytes();
+        this.cursorKey = (namespace + "$key:").getBytes();
+        this.endScanKey = advanceRow(this.cursorKey);
```
arunmahadevan: Is this the final end key or the end key for the first chunk ? Can you add some clarifying comment here?

external/storm-hbase/src/main/java/org/apache/storm/hbase/state/HBaseKeyValueState.java
```diff
@@ -0,0 +1,419 @@
...
+        LOG.debug("delete key '{}'", key);
+        byte[] columnKey = encoder.encodeKey(key);
+        V curr = get(key);
+        pendingPrepare.put(columnKey, encoder.getTombstoneValue());
+        return curr;
+    }
+
+    @Override
+    public Iterator<Map.Entry<K, V>> iterator() {
+        throw new UnsupportedOperationException("Not supported yet.");
```
HeartSaVioR: Ah I missed to replace the line with HBaseKeyValueStateIterator. Will change.

external/storm-hbase/src/main/java/org/apache/storm/hbase/state/HBaseKeyValueStateIterator.java
```diff
@@ -0,0 +1,153 @@
...
+import org.apache.storm.state.Serializer;
+
+import org.apache.storm.state.StateEncoder;
+
+import static org.apache.storm.hbase.state.HBaseKeyValueState.STATE_QUALIFIER;
+
+/**
+ * An iterator over {@link HBaseKeyValueState}.
+ */
+public class HBaseKeyValueStateIterator<K, V> extends BaseBinaryStateIterator<K, V> {
```
HeartSaVioR: Yes I intended to do that, but forgot to replace it.

external/storm-hbase/src/main/java/org/apache/storm/hbase/state/HBaseKeyValueStateIterator.java
HeartSaVioR: That's an end key for whole scan. I'll put comment on this line.

HeartSaVioR:
@arunmahadevan 
Thanks for reviewing. I addressed review comments, and squashed and rebased.
I'll also craft a PR for master branch.

arunmahadevan:
+1, LGTM

PR #1974, opened 2017-02-27
STORM-2038: Disable symlinks with a config option

revans2:
This is built on top of #1972 but I have not ported it to other branches yet, so I would like to keep the two separate.  Once reviews are done on this version I will port it to 1.x and 1.0.x (as it apparently is a real pain for windows users)

### DIFF
--- a/conf/defaults.yaml
+++ b/conf/defaults.yaml
@@ -58,6 +58,7 @@ storm.codedistributor.class: "org.apache.storm.codedistributor.LocalFileSystemCo
 storm.workers.artifacts.dir: "workers-artifacts"
 storm.health.check.dir: "healthchecks"
 storm.health.check.timeout.ms: 5000
+storm.disable.symlinks: false
 
 ### nimbus.* configs are for the master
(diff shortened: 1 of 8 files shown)


HeartSaVioR:
Applied the option and submit topology to remote cluster.
No symbolic link created. (`find . -type l` on storm root doesn't show any.)

+1

@revans2 Great work. Could you resolve conflict? And please create pull requests for 1.x and 1.0.x.

revans2:
@HeartSaVioR I have rebased the changes to remove the conflict (it was just in the documentation).  I'll start working on backporting it.

ptgoetz:
+1 @revans2 Was #1972 cherry-picked to the 1.x and 1.0.x branches?

revans2:
@ptgoetz yes #1972 was a clean cherry pick so I pulled it into 1.x and 1.0.x.  This is not so I have #2014 and #2015.

The only real changes for them is that I had to translate the code in Nimbus.java to clojure and modify the many getOrDefault calls to be java 7 compatible.

PR #1975, opened 2017-02-28
STORM-2386 Fail-back Blob deletion also fails in BlobSynchronizer.syncBlobs (1.x)

HeartSaVioR:
Please refer [STORM-2386](https://issues.apache.org/jira/browse/STORM-2386) for more details.

I'll submit patches for master soon. If we need to port back to 1.0.x as well, it should be easy to cherry-pick.


HeartSaVioR:
For master branch: #1976

harshach:
+1

PR #1976, opened 2017-02-28
STORM-2386 Fail-back Blob deletion also fails in BlobSynchronizer.syncBlobs

HeartSaVioR:
Please refer [STORM-2386](https://issues.apache.org/jira/browse/STORM-2386) for more details.

This is against master branch, 

* 1.x-branch: #1975


harshach:
+1

PR #1978, opened 2017-02-28
STORM-2387 Handle tick tuples properly for Bolts in external modules

HeartSaVioR:
* also remove ack for tick tuples since it's not necessary

I also fixed some mixing indentations. We should introduce code style soon.

For 1.x branch: #1979

### DIFF
--- a/external/storm-druid/src/main/java/org/apache/storm/druid/bolt/DruidBeamBolt.java
+++ b/external/storm-druid/src/main/java/org/apache/storm/druid/bolt/DruidBeamBolt.java
@@ -26,9 +26,11 @@ import org.apache.storm.task.OutputCollector;
 import org.apache.storm.task.TopologyContext;
 import org.apache.storm.topology.OutputFieldsDeclarer;
 import org.apache.storm.topology.base.BaseRichBolt;
+import org.apache.storm.topology.base.BaseTickTupleAwareRichBolt;
 import org.apache.storm.tuple.Fields;
 import org.apache.storm.tuple.Tuple;
 import org.apache.storm.tuple.Values;
+import org.apache.storm.utils.TupleUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -47,7 +49,7 @@ import java.util.Map;
  * <p/>
  *
  */
-public class DruidBeamBolt<E> extends BaseRichBolt {
+public class DruidBeamBolt<E> extends BaseTickTupleAwareRichBolt {
     private static final Logger LOG = LoggerFactory.getLogger(DruidBeamBolt.class);
 
     private volatile  OutputCollector collector;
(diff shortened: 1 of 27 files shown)


external/storm-druid/src/main/java/org/apache/storm/druid/bolt/DruidBeamBolt.java
```diff
@@ -76,30 +77,33 @@ public void prepare(Map stormConf, TopologyContext context, OutputCollector coll
 
     @Override
     public void execute(final Tuple tuple) {
-      Future future = tranquilizer.send((druidEventMapper.getEvent(tuple)));
-        LOG.debug("Sent tuple : [{}]" , tuple);
+        if (TupleUtils.isTick(tuple)) {
```
satishd: Most of these connector bolts extend `BaseRichBolt`. It may be good to have a class extending BaseRichBolt and handle ticktuples. This class can be used across other connector bolts instead of handling them in respective connector's base classes.

```
// you can have better class name if found any
public class BaseRichBoltWithTickTuples extends BaseRichBolt {
  @Override
  public void execute(final Tuple tuple) {
    if(TupleUtils.isTick(tuple)) {
      onTickTuple(tuple);
    } else {
      process(tuple);
    }
  }

  protected void onTickTuple(final Tuple tuple) {
  }

}

```

dossett: @satishd I agree in principle.  I once experimented with that approach and what I found difficult was declaring the `throws` for each of those methods since bolts can generate all sorts of exceptions.

+1

priyank5485: I agree with @satishd  Right now, onTickTuple method is in abstract base classes of different connectors and not even used. Taking Satish's approach is a little more work since we will have to change the bolts that actually process the tick tuples to separate out onProcess and onTickTuple logic. But i think its worth. @dossett I did not exactly get the problem with exceptions. Can't we have a catch all and rethrow as a RuntimeException?

HeartSaVioR: @satishd 
Make sense.
About class name I don't see good name for that. I thought `TickTupleAwareBaseRichBolt` inspired of Spring but the naming `~Aware` is actually used for interfaces. 
If we would want to make it public API (instead of using internally), it needs to have better class name.

@dossett 
`execute()` also doesn't have `throws` definitions. If we follow @satishd  `execute()` implementation, `execute()` doesn't care about exceptions and it's up to  subclasses.
So handling exceptions for two separated methods (`onTickTuple()` and `process()`) could happen, but it doesn't hurt. It just makes some duplicate codes.

Do you have other considerations or point of view?

satishd: @HeartSaVioR I am fine with `TickTupleAwareBaseRichBolt` unless a better name is found.

HeartSaVioR:
@satishd @priyank5485 Applied review comments.

satishd:
@HeartSaVioR +1 LGTM

PR #1979, opened 2017-02-28
STORM-2387 Handle tick tuples properly for Bolts in external modules (1.x)

HeartSaVioR:
* also remove ack for tick tuples since it's not necessary

I also fixed some mixing indentations. We should introduce code style soon.

For master branch: #1978


HeartSaVioR:
I'll also merge this given that #1978 got +1 and this is for porting back.

PR #1818, opened 2016-12-08
STORM-2104 1.x

srdo:
1.x version of https://github.com/apache/storm/pull/1696

PR #1980, opened 2017-02-28
Storm-2104 1.0.x

srdo:
Putting https://github.com/apache/storm/pull/1818 on 1.0.x as well since @qiozas asked about it here https://github.com/apache/storm/pull/1696

Only change from 1.x version is that `KafkaSpout.emitTupleIfNotEmitted` returns void instead of boolean here.


external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutConfig.java
```diff
@@ -164,15 +164,15 @@ public Builder(Map<String, Object> kafkaProps, KafkaSpoutStreams kafkaSpoutStrea
         /**
          * Specifying this key deserializer overrides the property key.deserializer
          */
-        public Builder<K,V> setKeyDeserializer(Deserializer<K> keyDeserializer) {
+        public Builder<K,V> setKeyDeserializer(SerializableDeserializer<K> keyDeserializer) {
```
HeartSaVioR: I'm not 100% sure it keeps binary compatibility. `Serializable ` doesn't have methods so it might work, but just not sure.
Once who makes me clear that this is also safe for binary compatibility, I'll give +1.

srdo: It doesn't. I'll revert this part.

HeartSaVioR:
Thanks for the quick fix. +1

PR #1983, opened 2017-03-02
(1.x) STORM-2389 Avoid instantiating Event Logger when topology.eventlogger…

HeartSaVioR:
….executors=0

Manually tested with cluster mode: 4 workers in local machine.


harshach:
+1

PR #1988, opened 2017-03-03
STORM-2395: storm.cmd supervisor calls the wrong class name

alexlehm:
use org.apache.storm.daemon.supervisor.Supervisor instead

Signed-off-by: alexlehm <alexlehm@gmail.com>


HeartSaVioR:
+1 Nice finding.

PR #1989, opened 2017-03-04
STORM-2396: setting interrupted status back before throwing a RuntimeException

ikashperskyi:
As described in [JIRA](https://issues.apache.org/jira/browse/STORM-2396) the interrupted flag is not set back after catching InterruptedException and wrapping it into a RuntimeException so any code dependent on checking the interrupted status will not work properly in case the RuntimeException doesn't make it up the stack.

### DIFF
--- a/storm-core/src/jvm/org/apache/storm/utils/Utils.java
+++ b/storm-core/src/jvm/org/apache/storm/utils/Utils.java
@@ -356,6 +356,7 @@ public class Utils {
         try {
             Time.sleep(millis);
         } catch(InterruptedException e) {
+            Thread.currentThread().interrupt();
             throw new RuntimeException(e);
         }
     }


storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -134,6 +134,8 @@
 import clojure.lang.Keyword;
 import clojure.lang.RT;
 
+import static java.lang.Thread.currentThread;
```
HeartSaVioR: `Thread.currentThread()` is widely used so it's more readable than `currentThread()` for me.

storm-core/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -356,13 +358,14 @@ public static void sleep(long millis) {
         try {
             Time.sleep(millis);
         } catch(InterruptedException e) {
+            currentThread().interrupt();
```
HeartSaVioR: How it will collaborate with throwing RuntimeException? Are we assuming that caller catches RuntimeException? Just not clear that how it works together.

ikashperskyi: Here's a suboptimal example of a loop that will never exit as intended
`while(!currentThread.isInterrupted()) {
	execute();
}
...

void execute() {
	try {
		... // RE may be thrown here
		Utils.sleep(1000);
		... // and here
	} catch (RuntimeException e)
		log.error(e);
	}
}`

storm-core/src/jvm/org/apache/storm/utils/Utils.java
ikashperskyi: Okay, will change.

srdo:
+1

storm-core/src/jvm/org/apache/storm/utils/Utils.java
HeartSaVioR: Got it. Thanks.

HeartSaVioR:
+1

PR #1993, opened 2017-03-09
STORM-2402: KafkaSpout sub-classes should be able to customize tuple processing

sakanaou:
We need a `KafkaSpout` that writes unprocessable records to a "dead-letter-topic". For this to function we sub-classed `KafkaSpout` and added the corresponding code. Without this patch sub-classses can not have access to the actual tuples/records but just the `KafkaSpoutMessageId` in `ack()` and `fail()`.

### DIFF
--- a/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
+++ b/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
@@ -304,6 +304,7 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
             boolean isScheduled = retryService.isScheduled(msgId);
             if (!isScheduled || retryService.isReady(msgId)) {   // not scheduled <=> never failed (i.e. never emitted) or ready to be retried
                 final List<Object> tuple = tuplesBuilder.buildTuple(record);
+                emittingTuple(msgId, tuple, record);
                 kafkaSpoutStreams.emit(collector, tuple, msgId);
                 emitted.add(msgId);
                 numUncommittedOffsets++;
@@ -313,6 +314,11 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
                 LOG.trace("Emitted tuple [{}] for record [{}]", tuple, record);
             }
         }
+    }
+
+    // Allows for custom processing in sub-classes
+    protected void emittingTuple(KafkaSpoutMessageId msgId, List<Object> tuple, ConsumerRecord<K, V> record) {
+        // Yet empty
     }
 
     private void commitOffsetsForAckedTuples() {


harshach:
@sakanaou Instead of providing a method like you've in PR, I recommend you to provide TupleBuilder interface with a default implementation that can be pluggable via config option. This will allow users to write their own implementation and configure it.

hmcl:
@sakanaou I am just about to commit a fix that gives the option to emit or not emit null tuples as configured in `KafkaSpoutConfig`. The case of non processable records is already supported if you implement a suitable [RecordTranslator](https://github.com/apache/storm/blob/master/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/RecordTranslator.java) that has its [apply](https://github.com/apache/storm/blob/master/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/RecordTranslator.java#L39) method return [KafkaTuple](https://github.com/apache/storm/blob/master/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaTuple.java) instances. You can have an instance of `KafkaTuple` that is associated with the "dead-letter-topic" stream to handle invalid records. 

Based on the above,  don't think the changes proposed in this pull request are necessary.

sakanaou:
Thanks for your suggestions, @harshach & @hmcl. Maybe I didn't find the correct way to handle our request yet in the provided architecture. But to make it more clear: The records/tuples are not per se invalid and should go straight to a "dead-letter-topic". The system is more about resilience in such a way that it should not be a big problem if a system, that lies behind a downward bolt, is temporarily unavailable. Instead of simply "skipping" such failed records (after their retry count is reached) they should then - and only then - be written to the mentioned "dead-letter-topic". We did so in the overwritten `fail()` of our `KafkaSpout` and built a map from `KafkaSpoutMessageId` to the `ConsumerRecord` with the help of the proposed `emittingTuple()`.

But I assume we could work out a solution solely based on the mentioned `RecordTranslator` (or `KafkaSpoutTupleBuilder` in our case as we are still on Storm 1.0.2).

srdo:
If `RecordTranslator.apply` was given access to the `KafkaSpoutMessageId` (or current number of retries), you could redirect tuples that have been retried sufficiently many times to another stream using `KafkaTuple` as @hmcl mentioned. That allows you to write the failed tuples to the dead letter topic from a KafkaBolt instead of doing it from inside the spout. It would be cleaner IMO.

sakanaou:
Then this PR can be safely closed and discarded. Thanks for your input.

srdo:
@sakanaou The `RecordTranslator.apply` function isn't currently getting the `KafkaSpoutMessageId/numRetries`, so I'm not sure you can implement the redirect with the code as it is now. We might still want to make some sort of change here.

PR #1994, opened 2017-03-09
STORM-2403 Fix KafkaBolt test failure: tick tuple should not be acked

HeartSaVioR:
* More detail: [STORM-2403](http://issues.apache.org/jira/browse/STORM-2403)
* This bug is introduced from [STORM-2387](http://issues.apache.org/jira/browse/STORM-2387)

**NOTE: It should be easy to cheery-pick to 1.x / 1.0.x branches, and we must cherry-pick to those branches since STORM-2387 is merged to those branches.**


vesense:
+1

PR #1995, opened 2017-03-10
STORM-2407: KafkaTridentSpoutOpaque Doesn't Poll Data From All Topic-…

hmcl:
…Partitions When Parallelism Hint Not a Multiple Total Topic-Partitions

 - Introduce logic to poll data from the topic partitions assigned to each  task

### DIFF
--- a/storm-core/src/jvm/org/apache/storm/trident/spout/IOpaquePartitionedTridentSpout.java
+++ b/storm-core/src/jvm/org/apache/storm/trident/spout/IOpaquePartitionedTridentSpout.java
@@ -52,13 +52,26 @@ public interface IOpaquePartitionedTridentSpout<Partitions, Partition extends IS
          * This method is called when this task is responsible for a new set of partitions. Should be used
          * to manage things like connections to brokers.
          */        
-        void refreshPartitions(List<Partition> partitionResponsibilities);        
+        void refreshPartitions(List<Partition> partitionResponsibilities);
+
+        /**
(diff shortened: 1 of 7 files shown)


storm-core/src/jvm/org/apache/storm/trident/spout/IOpaquePartitionedTridentSpout.java
```diff
@@ -52,13 +52,26 @@
...
+
+        /**
+         * @return The oredered list of partitions being processed by all the tasks
+         */
         List<Partition> getOrderedPartitions(Partitions allPartitionInfo);
+
+        /**
+         * @return The list of partitions that are to be processed by the task with id {@code taskId}
+         */
+        List<Partition> getPartitionsForTask(int taskId, int numTasks, Partitions allPartitionInfo);
```
harshach: if we are making changes to interface here, why keep getPartitionsForTask. Given the partition assignment depends on Kafka consumer api, Why not let it take over the assignment. All of this assignment magic came about because we ended up using the lower-level API. We should look at a way to remove this all together as the consumer API does a better handling of distributing partitions.

harshach: @hmcl my bad, missed the kafkaConsumer.assignments() part. This looks good.

harshach:
+1.

PR #1999, opened 2017-03-14
STORM-2409: Storm-Kafka-Client KafkaSpout Support for Failed and NullTuples

hmcl:
- Created config property to make emit null tuples configurable
  - Ack directly null tuples that are not emitted

### DIFF
--- a/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
+++ b/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
@@ -76,18 +76,18 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
     private transient boolean initialized;                              // Flag indicating that the spout is still undergoing initialization process.
     // Initialization is only complete after the first call to  KafkaSpoutConsumerRebalanceListener.onPartitionsAssigned()
 
-    private transient Map<TopicPartition, OffsetManager> acked;           // Tuples that were successfully acked. These tuples will be committed periodically when the commit timer expires, after consumer rebalance, or on close/deactivate
+    private transient Map<TopicPartition, OffsetManager> acked;         // Tuples that were successfully acked. These tuples will be committed periodically when the commit timer expires, or after a consumer rebalance, or during close/deactivate
     private transient Set<KafkaSpoutMessageId> emitted;                 // Tuples that have been emitted but that are "on the wire", i.e. pending being acked or failed. Not used if it's AutoCommitMode
-    private transient Iterator<ConsumerRecord<K, V>> waitingToEmit;         // Records that have been polled and are queued to be emitted in the nextTuple() call. One record is emitted per nextTuple()
+    private transient Iterator<ConsumerRecord<K, V>> waitingToEmit;     // Records that have been polled and are queued to be emitted in the nextTuple() call. One record is emitted per nextTuple()
     private transient long numUncommittedOffsets;                       // Number of offsets that have been polled and emitted but not yet been committed. Not used if it's AutoCommitMode
+    private transient Timer refreshSubscriptionTimer;                   // Triggers when a subscription should be refreshed
     private transient TopologyContext context;
-    private transient Timer refreshSubscriptionTimer;                   // Used to say when a subscription should be refreshed
 
 
     public KafkaSpout(KafkaSpoutConfig<K, V> kafkaSpoutConfig) {
         this(kafkaSpoutConfig, new KafkaConsumerFactoryDefault<K, V>());
     }
-    
+
     //This constructor is here for testing
     KafkaSpout(KafkaSpoutConfig<K, V> kafkaSpoutConfig, KafkaConsumerFactory<K, V> kafkaConsumerFactory) {
         this.kafkaConsumerFactory = kafkaConsumerFactory;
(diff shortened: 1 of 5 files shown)


srdo:
Is this necessary to handle in the spout? Couldn't it be handled by putting a null filtering bolt between the spout and the rest of the topology?

harshach:
@srdo its better to have this on kafka spout itself, for one it reduces unnecessary hope or another bolt that user need to add to filter out unnecessary tuples.

srdo:
@harshach Okay. I don't think filtering null tuples should be the default though. We might be hiding errors in user code. I'd like them to have to actively enable null filtering.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -383,23 +403,24 @@ private void commitOffsetsForAckedTuples() {
...
         final KafkaSpoutMessageId msgId = (KafkaSpoutMessageId) messageId;
-        if(!emitted.contains(msgId)) {
-            LOG.debug("Received ack for tuple this spout is no longer tracking. Partitions may have been reassigned. Ignoring message [{}]", msgId);
-            return;
-        }
-        
-        if (!consumerAutoCommitMode) {  // Only need to keep track of acked tuples if commits are not done automatically
-            acked.get(msgId.getTopicPartition()).add(msgId);
+        if (!emitted.contains(msgId)) {
+            LOG.debug("Received ack for message [{}], which is associated with either a null tuple that was never emitted, " +
```
srdo: This message is ambiguous now. Could you add the msgId to emitted before acking (i.e. basically fake a "real" ack) instead? If people want to log that they're dropping an invalid tuple they can do it from the RecordTranslator without ambiguity.

hmcl: I don't understand what you mean with this comment.

hmcl:
@srdo the changes proposed in this PR handle null tuples in identical fashion that what is done in the old Kafka Spout as can be seen in the following links.

https://github.com/apache/storm/blob/master/external/storm-kafka/src/jvm/org/apache/storm/kafka/PartitionManager.java#L197

https://github.com/apache/storm/blob/master/external/storm-kafka/src/jvm/org/apache/storm/kafka/PartitionManager.java#L208-L210

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
srdo: I mean that the debug log here is now getting written both when the spout receives an ack for a tuple from a partition that has been reassigned since it was emitted, and when the RecordTranslator emits a null tuple. If you add msgId to emitted in L357 before calling ack(), we don't have to change the log here to also mention null tuples.

I think logging something when null tuples are being dropped might fit better in RecordTranslator.apply (or in L357) than here, because then the log can be clear about what is happening.

srdo:
@hmcl Googled it, it appears that null messages can be sent to delete messages in Kafka (see https://kafka.apache.org/documentation/#design_compactionbasics). It makes sense to drop messages quietly if their values are null. 

It's good to be able to automatically drop them if the RecordTranslator returns null as well, I'm mostly worried that bugs in RecordTranslator implementations could be hidden by the null drop being enabled by default. It's more of a nitpick though.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
hmcl: @srdo The reason I decided not to add msgId to emitted was because that is a big hack. We will be setting the internal state of the spout like a message was emitted, where it really wasn't. 

We can probably fix this by creating and setting a boolean field `emitted` in msgId, which by default is true, but in this particular scenario that we ack directly, we set it to false.

We cannot log null tuples message in RecordTranslator because that interface is implemented by the user. This message is internal to the spout, and should not depend on third party (user) implementations.

hmcl:
@srdo we can make de default emit null. That's a trivial change. As I said, I decided not to emit null because that is the default behavior in the old spout, and users are either happy, or used to it. I am not in favor or one way or another. It's just that emitting null tuples, unforeseen problems can potentially happen downstream.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
srdo: Adding the `emitted` field to msgId would be a good solution. The log can be done in L357 instead then.

hmcl: OK, will do.

hmcl:
@srdo so what will we decide to do as far as null? send or not send by default ?

srdo:
@hmcl Either one is fine by me. Keeping it consistent with the old spout is probably the best solution. The only case where defaulting to emitting nulls makes sense is if the user supplies a RecordTranslator that sometimes returns null erroneously, because they'll then see the bug when they get unexpected nulls, but I don't think it's a big problem.

harshach:
@srdo Log compaction is only used very specific case like consumer_offsets or streams topics. It only useful if you want to keep the latest data. So we are not addressing those here. 
@hmcl  lets add a debug statement that will log this when we drop the null value

srdo:
+1

@harshach I don't know about that. Can the Kafka consumer randomly return null? Users should know if their Kafka producers are sending nulls, or if their RecordTranslator is returning nulls. If we don't know why nulls are getting returned and the user didn't manually enable filtering, we might end up making it harder for users to debug missing messages.

harshach:
+1

satishd:
+1

PR #2000, opened 2017-03-14
STORM-2411 Setting topology.eventlogger.executors=0 in defaults.yaml

roshannaik:


HeartSaVioR:
+1 This makes consistent with 1.x branch.

Ethanlm:
Just realized that https://github.com/apache/storm/blob/master/storm-client/src/jvm/org/apache/storm/Config.java#L398 should be updated because of this patch

roshannaik:
Good point. Will provide a fix.

Ethanlm:
I think I can include it in my PR for https://issues.apache.org/jira/browse/STORM-3021

roshannaik:
thanks.

PR #2003, opened 2017-03-14
STORM-2412: Nimbus isLeader check while waiting for max replication

arunmahadevan:
While using local FS blob store, nimbus goes into a state where it indefinitely waits for max replication (with max replication wait time set to -1). At this time its also observed that the nimbus that received the topology submission is no longer the leader.
While waiting for max replication, nimbus can also do the `isLeader` check and fail the topology submission if its no longer the leader.

Also added some debug logs to better troubleshoot the race conditions when it happens.


satishd:
+1 LGTM

HeartSaVioR:
+1

satishd:
@arunmahadevan You may want to raise PR on master also.

PR #2005, opened 2017-03-15
STORM-2414 Skip checking meta's ACL when subject has write privileges for any blobs (1.x)

HeartSaVioR:
Syncing up blobs throws KeyNotFoundException when deleting blobs which are already deleted from ZK.
Currently deleting blob requires pulling the latest change first to check ACL. That makes sense, but in above case we already deleted the blob from ZK so it just fails. We need to have "force delete" when request for deletion comes from syncing up blobs.

* add new method: clearRemovedBlob
  * skip checking ACL while deleting when blob is already removed from ZK

This patch is easy to be cherry-picked to master and 1.0.x branches as well.

### DIFF
--- a/storm-core/src/jvm/org/apache/storm/blobstore/LocalFsBlobStore.java
+++ b/storm-core/src/jvm/org/apache/storm/blobstore/LocalFsBlobStore.java
@@ -129,9 +129,7 @@ public class LocalFsBlobStore extends BlobStore {
     @Override
     public AtomicOutputStream updateBlob(String key, Subject who) throws AuthorizationException, KeyNotFoundException {
         validateKey(key);
-        checkForBlobOrDownload(key);
-        SettableBlobMeta meta = getStoredBlobMeta(key);
-        _aclHandler.hasPermissions(meta.get_acl(), WRITE, who, key);
+        checkPermission(key, who, WRITE);
         try {
             return new BlobStoreFileOutputStream(fbs.write(DATA_PREFIX+key, false));
         } catch (IOException e) {
@@ -219,14 +217,47 @@ public class LocalFsBlobStore extends BlobStore {
     @Override
     public void deleteBlob(String key, Subject who) throws AuthorizationException, KeyNotFoundException {
         validateKey(key);
-        checkForBlobOrDownload(key);
-        SettableBlobMeta meta = getStoredBlobMeta(key);
-        _aclHandler.hasPermissions(meta.get_acl(), WRITE, who, key);
+
+        if (!_aclHandler.checkForValidUsers(who, WRITE)) {
+            // need to get ACL from meta
+            LOG.debug("Retrieving meta to get ACL info... key: {} subject: {}", key, who);
+
+            try {
+                checkPermission(key, who, WRITE);
+            } catch (KeyNotFoundException e) {
+                LOG.error("Error while retrieving meta from ZK or local... key: {} subject: {}", key, who);
+                throw e;
+            }
+        } else {
+            // able to delete the blob without checking meta's ACL
+            // skip checking everything and continue deleting local files
+            LOG.debug("Given subject is eligible to delete key without checking ACL, skipping... key: {} subject: {}",
+                    key, who);
+        }
+
         try {
-            fbs.deleteKey(DATA_PREFIX+key);
-            fbs.deleteKey(META_PREFIX+key);
+            deleteKeyIgnoringFileNotFound(DATA_PREFIX + key);
+            deleteKeyIgnoringFileNotFound(META_PREFIX + key);
         } catch (IOException e) {
             throw new RuntimeException(e);
+        }
+    }
+
+    private void checkPermission(String key, Subject who, int mask) throws KeyNotFoundException, AuthorizationException {
+        checkForBlobOrDownload(key);
+        SettableBlobMeta meta = getStoredBlobMeta(key);
+        _aclHandler.hasPermissions(meta.get_acl(), mask, who, key);
+    }
+
+    private void deleteKeyIgnoringFileNotFound(String key) throws IOException {
+        try {
+            fbs.deleteKey(key);
+        } catch (IOException e) {
+            if (e instanceof FileNotFoundException) {
+                LOG.debug("Ignoring FileNotFoundException since we're about to delete such key... key: {}", key);
+            } else {
+                throw e;
+            }
         }
     }
 


harshach:
+1

storm-core/src/jvm/org/apache/storm/blobstore/BlobStore.java
```diff
@@ -128,6 +128,15 @@
     public abstract void deleteBlob(String key, Subject who) throws AuthorizationException, KeyNotFoundException;
 
     /**
+     * Clear the blob data and metadata which is already removed from ZK.
+     * Implementation should just remove the blob as it is.
+     * This method should be used ONLY for syncing up blobs in Nimbus.
+     * @param key Key for the blob.
+     * @throws AuthorizationException
+     */
+    public abstract void clearRemovedBlob(String key) throws AuthorizationException;
```
satishd: Is `BlobStore` a public API, which users can write their own implementations? If so, then this change would not be a backward compatible change for those users.

storm-core/src/jvm/org/apache/storm/blobstore/LocalFsBlobStore.java
```diff
@@ -231,6 +231,18 @@ public void deleteBlob(String key, Subject who) throws AuthorizationException, K
     }
 
     @Override
+    public void clearRemovedBlob(String key) throws AuthorizationException {
+        validateKey(key);
+        // not pulling recent change and checking ACL
```
arunmahadevan: if you handle this in deleteBlob itself, we don't need to add new api to blobstore. i.e. if getStoredBlobMeta throws KeyNotFoundException, just log it and skip the ACL check and proceed with deleting the keys.

storm-core/src/jvm/org/apache/storm/blobstore/BlobStore.java
HeartSaVioR: @satishd 
Yes but no one would implement their own BlobStore. 
Btw, as @arunmahadevan suggested, we can do it without this change. I'll change it.

storm-core/src/jvm/org/apache/storm/blobstore/LocalFsBlobStore.java
HeartSaVioR: @arunmahadevan 
Yes I was thinking about that way but over-thought about false-positive (when ZK is down, throwing Exception, etc). KeyNotFoundException clearly states that ZK is accessible and key is not exist so no false-positive. Thanks I'll address this.

storm-core/src/jvm/org/apache/storm/blobstore/BlobStore.java
satishd: @HeartSaVioR :) May not want to assume no one implements their store when it is a public API. Anyways, good that  `deleteBlob` can be changed to handle this scenario.

storm-core/src/jvm/org/apache/storm/blobstore/LocalFsBlobStore.java
```diff
@@ -219,9 +219,20 @@ public void setBlobMeta(String key, SettableBlobMeta meta, Subject who) throws A
     @Override
     public void deleteBlob(String key, Subject who) throws AuthorizationException, KeyNotFoundException {
         validateKey(key);
-        checkForBlobOrDownload(key);
-        SettableBlobMeta meta = getStoredBlobMeta(key);
-        _aclHandler.hasPermissions(meta.get_acl(), WRITE, who, key);
+
+        boolean checkPermission = true;
+        try {
+            checkForBlobOrDownload(key);
```
HeartSaVioR: Here KeyNotFoundException is thrown when blob is removed from ZK.

storm-core/src/jvm/org/apache/storm/blobstore/LocalFsBlobStore.java
```diff
@@ -219,9 +219,20 @@ public void setBlobMeta(String key, SettableBlobMeta meta, Subject who) throws A
...
+        } catch (KeyNotFoundException e) {
+            // not exist in ZK so this blob should be deleted regardless of permission
+            checkPermission = false;
+        }
+
+        if (checkPermission) {
+            SettableBlobMeta meta = getStoredBlobMeta(key);
```
HeartSaVioR: Here KeyNotFoundException is thrown when blob file is not found.

HeartSaVioR:
@harshach @satishd @arunmahadevan Addressed. Please review again. Thanks in advance!

storm-core/src/jvm/org/apache/storm/blobstore/LocalFsBlobStore.java
```diff
@@ -219,9 +219,20 @@ public void setBlobMeta(String key, SettableBlobMeta meta, Subject who) throws A
...
         validateKey(key);
-        checkForBlobOrDownload(key);
-        SettableBlobMeta meta = getStoredBlobMeta(key);
-        _aclHandler.hasPermissions(meta.get_acl(), WRITE, who, key);
+
+        boolean checkPermission = true;
+        try {
+            checkForBlobOrDownload(key);
+        } catch (KeyNotFoundException e) {
+            // not exist in ZK so this blob should be deleted regardless of permission
```
arunmahadevan: maybe just add some logs here.

satishd:
+1

storm-core/src/jvm/org/apache/storm/blobstore/LocalFsBlobStore.java
HeartSaVioR: Sorry I was mistaken... Even if there's no key on ZK, KeyNotFoundException is not thrown.
KeyNotFoundException is thrown only when there's race condition between ZK node update and removing node from ZK.

HeartSaVioR:
@arunmahadevan @satishd 
Sorry please take a look again. Actually origin issue was observed when meta blob file is already deleted. (I've mistaken the codebase.) Changed to ignore the FNF exception when deleting file.

storm-core/src/jvm/org/apache/storm/blobstore/LocalFsBlobStore.java
satishd: If it does not throw `KeyNotFoundException` when there is no key on ZK then how do you find out whether the given key exists or not on ZK so that you can set `checkPermission` as `false`.

storm-core/src/jvm/org/apache/storm/blobstore/LocalFsBlobStore.java
```diff
@@ -219,17 +219,44 @@ public void setBlobMeta(String key, SettableBlobMeta meta, Subject who) throws A
...
+            LOG.warn("Detected deletion from ZK for the key {} while pulling the latest change. Needs to be deleted. key: {}", key);
+            checkPermission = false;
+        }
+
+        if (checkPermission) {
+            try {
+                SettableBlobMeta meta = getStoredBlobMeta(key);
+                _aclHandler.hasPermissions(meta.get_acl(), WRITE, who, key);
+            } catch (KeyNotFoundException e) {
+                // meta file is not exist, skip handling error to delete data file
```
satishd: nit: meta file does not exist

storm-core/src/jvm/org/apache/storm/blobstore/LocalFsBlobStore.java
```diff
@@ -219,17 +219,44 @@ public void setBlobMeta(String key, SettableBlobMeta meta, Subject who) throws A
...
+
+        boolean checkPermission = true;
+        try {
+            checkForBlobOrDownload(key);
+        } catch (KeyNotFoundException e) {
+            LOG.warn("Detected deletion from ZK for the key {} while pulling the latest change. Needs to be deleted. key: {}", key);
+            checkPermission = false;
+        }
+
+        if (checkPermission) {
```
arunmahadevan: Instead of this can you combine the two `try...catch` into something like
```java
try {
 checkForBlobOrDownload(key);
 SettableBlobMeta meta = getStoredBlobMeta(key);
 _aclHandler.hasPermissions(meta.get_acl(), WRITE, who, key);
} catch (KeyNotFoundException e) {
 LOG.warn(...);
}
```

HeartSaVioR: The meaning for KeyNotFoundException is different between `checkForBlobOrDownload` and `getStoredBlobMeta`. First one indicates ZK node is not exist and second one indicates META blob file in local is not exist.

storm-core/src/jvm/org/apache/storm/blobstore/LocalFsBlobStore.java
HeartSaVioR: Thanks addressed.

storm-core/src/jvm/org/apache/storm/blobstore/LocalFsBlobStore.java
HeartSaVioR: Maybe we would need to check ZK and set checkPermission to false. 
If there's race condition so ZK node didn't exist but added later, we might skip pulling latest information and delete the blob in local anyway, but it should be occurred only from standby Nimbus and the blob will be re-pulled at next sync up.

I'll make appropriate change. Thanks.

storm-core/src/jvm/org/apache/storm/blobstore/LocalFsBlobStore.java
arunmahadevan: Unless you want to apply different logic in the catch block for the two methods. If we just want to log a generic message having single try..catch should be sufficient.

HeartSaVioR:
Looking into the logic deeply and realize `deleteBlob` is not that simple as it seems especially with H/A... Investigating more on this.

HeartSaVioR:
@arunmahadevan @satishd 
I have updated the problem definition and fix for the problem.

If deleting blob is requested from superuser (say, Nimbus, Admin. Supervisor is not since Supervisor is only superuser for reading) we don't need to get meta since the user already get WRITE privileges for any blobs. Syncing up blobs is done via Nimbus privilege so it will pass downloading and/or reading meta.

If deleting blob is requested from user, the logic will be run via leader Nimbus. Given that the scenario, both `ZK node not found` and `meta file not found` are all errors which users (maybe admin) need to know about the thing. 
Ideally we might want to consider repairing, but it makes the logic more complicated so I'd rather define Blobstore H/A as not fully reliable and recommend HDFS BlobStore for H/A environment.

Please take a look again. Thanks in advance!

HeartSaVioR:
Build failure is not related to this one: `storm-core` and `!storm-core` succeed for both JDK7 and JDK8.

storm-core/src/jvm/org/apache/storm/blobstore/LocalFsBlobStore.java
```diff
@@ -219,17 +219,37 @@ public void setBlobMeta(String key, SettableBlobMeta meta, Subject who) throws A
     @Override
     public void deleteBlob(String key, Subject who) throws AuthorizationException, KeyNotFoundException {
         validateKey(key);
-        checkForBlobOrDownload(key);
-        SettableBlobMeta meta = getStoredBlobMeta(key);
-        _aclHandler.hasPermissions(meta.get_acl(), WRITE, who, key);
+
+        if (!_aclHandler.checkForValidUsers(who, WRITE)) {
+            // need to get ACL from meta
```
arunmahadevan: add LOG

storm-core/src/jvm/org/apache/storm/blobstore/LocalFsBlobStore.java
```diff
@@ -219,17 +219,37 @@ public void setBlobMeta(String key, SettableBlobMeta meta, Subject who) throws A
...
+            checkForBlobOrDownload(key);
+            SettableBlobMeta meta = getStoredBlobMeta(key);
+            _aclHandler.hasPermissions(meta.get_acl(), WRITE, who, key);
+        } else {
+            // able to delete the blob without checking meta's ACL
```
arunmahadevan: LOG

storm-core/src/jvm/org/apache/storm/blobstore/LocalFsBlobStore.java
```diff
@@ -219,17 +219,37 @@ public void setBlobMeta(String key, SettableBlobMeta meta, Subject who) throws A
...
             throw new RuntimeException(e);
         }
     }
 
+    private void deleteKeyIgnoringFileNotFound(String key) throws IOException {
+        try {
+            fbs.deleteKey(key);
+        } catch (IOException e) {
+            if (e instanceof FileNotFoundException) {
+                // ignoring
```
arunmahadevan: LOG

storm-core/src/jvm/org/apache/storm/blobstore/LocalFsBlobStore.java
HeartSaVioR: @arunmahadevan Will leave log message with debug level since it's useful for debugging. Please let me know if you would like to raise level to INFO or so.

storm-core/src/jvm/org/apache/storm/blobstore/LocalFsBlobStore.java
HeartSaVioR: Same as above.

storm-core/src/jvm/org/apache/storm/blobstore/LocalFsBlobStore.java
HeartSaVioR: Same as above.

satishd:
+1

PR #2009, opened 2017-03-16
STORM-2407: KafkaTridentSpoutOpaque Doesn't Poll Data From All Topic-…

hmcl:
…Partitions When Parallelism Hint Not a Multiple Total Topic-Partitions

 - Introduce logic to poll data from the topic partitions assigned to each  task


hmcl:
@harshach can you please review. This is the master branch equivalent to https://github.com/apache/storm/pull/1995

PR #2010, opened 2017-03-16
STORM-2240: STORM PMML Bolt - Add Support to Load Models from Blob Store

hmcl:
- Created factory to create ModelRunner and ModelOutputs from Blobstore
   - Updated example topology to support loading PMML model from file and from Blobstore


hmcl:
@satishd this is the 1.x-branch equivalent to master.

satishd:
merged to 1.x-branch, Thanks @hmcl

PR #2012, opened 2017-03-16
STORM-2419 - Fix storm-perf module's parent module version and path

satishd:
@harshach ping


HeartSaVioR:
I intentionally let this 1.1.0 because it shouldn't be rollbacked to 1.1.0 when rolling back release. RC change is not rolled back yet.

cc. @ptgoetz

PR #2014, opened 2017-03-16
STORM-2038: Disable symlinks with a config option (1.x)

revans2:


HeartSaVioR:
+1

revans2:
@HeartSaVioR could you also take a look at #2015 it is almost identical, except some white space changes in nimbus.clj made it so the cherry-pick was not 100% clean.

PR #2016, opened 2017-03-17
STORM-2422: Reduce the size of a serialized trident topology

revans2:
For now I only plan to do this for master.  If someone else runs into this issue I am happy to backport it to other versions of storm.


revans2:
@arunmahadevan yes this is showing up in a real topology (part of monitoring and alerting at Yahoo).  And yes some of these topologies are really large.  They are machine generated from a number of different user provided configurations, which is why it can grow to be so large.  Having fewer big topologies instead of many small topologies makes them more efficient because they only have to read/process the metrics once.  There are ways to work around it on the user side, but this felt like a cleaner solution overall.

HeartSaVioR:
+1

PR #2019, opened 2017-03-18
STORM-2423 - Join Bolt should use explicit instead of default window anchoring for emitted tuples

roshannaik:
Default anchoring will anchor each emitted tuple to every tuple in current window. This requires a very large numbers of ACKs from any downstream bolt. If topology.debug is enabled, it also worsens the load on the system significantly.
Letting the topo run in this mode (in particular with max.spout.pending disabled), could lead to the worker running out of memory and crashing.

*Fix* Join Bolt should avoid using default window anchoring, and explicitly anchor each emitted tuple with the exact matching tuples form each inputs streams. This reduces the complexity of the tuple trees and consequently the reduces burden on the ACKing & messaging subsystems.

Also added additional documentation.

### DIFF
--- a/docs/Joins.md
+++ b/docs/Joins.md
@@ -122,4 +122,18 @@ is treated as a single stream and joined against `stream2` from `bolt3`.
 2. Duplication of joined records across windows is possible when using Sliding Windows. This is because the tuples continue to exist
    across multiple windows when using Sliding Windows.
 
+3. If message timeouts are enabled, ensure the timeout setting (topology.message.timeout.secs) is large enough to comfortably
+   accommodate the window size, plus the additional processing by other spouts and bolts.
 
+4. Joining a window of two streams with M and N elements each, *in the worst case*, can produce MxN elements with every output tuple
+   anchored to one tuple from each input stream. This can mean a lot of output tuples from JoinBolt and even more ACKs for downstream bolts
+   to emit. This can place a substantial pressure on the messaging system and dramatically slowdown the topology if not careful.
+   To manage the load on the messaging subsystem, it is advisable to:
+   * Increase the worker's heap (topology.worker.max.heap.size.mb).
+   * **If** ACKing is not necessary for your topology, disable ACKers (topology.acker.executors=0).
+   * Disable event logger (topology.eventlogger.executors=0).
+   * Turn of topology debugging (topology.debug=false).
+   * Set topology.max.spout.pending to a value large enough to accommodate an estimated full window worth of tuples plus some more for headroom.
+     This helps mitigate the possibility of spouts emitting excessive tuples when messaging subsystem is experiencing excessive load. This situation
+     can occur when its value is set to null.
+   * Lastly, keep the window size to the minimum value necessary for solving the problem at hand.
--- a/storm-core/src/jvm/org/apache/storm/bolt/JoinBolt.java
+++ b/storm-core/src/jvm/org/apache/storm/bolt/JoinBolt.java
@@ -191,9 +191,11 @@ public class JoinBolt extends BaseWindowedBolt {
         for (ResultRecord resultRecord : joinResult.getRecords()) {
             ArrayList<Object> outputTuple = resultRecord.getOutputFields();
             if ( outputStreamName==null )
-                collector.emit( outputTuple );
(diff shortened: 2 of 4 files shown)


docs/Joins.md
```diff
@@ -122,4 +122,18 @@ is treated as a single stream and joined against `stream2` from `bolt3`.
 2. Duplication of joined records across windows is possible when using Sliding Windows. This is because the tuples continue to exist
    across multiple windows when using Sliding Windows.
 
-
+3. If message timeouts are enabled, ensure the timeout setting (topology.message.timeout.secs) is large enough to comfortably
+   accommodate the window size, plus the additional processing by other spouts and bolts.
+
+4. Joining a window of two streams with M and N elements each, *in the worst case*, can produce MxN elements with every output tuple
+   anchored to one tuples from each input stream. This can mean a lot of output tuples from JoinBolt and even more ACKs for downstream bolts
```
satishd: Should not this be
produce MxN elements with every output tuple anchored to <s>one</s> tuples  from each input stream.

storm-core/src/jvm/org/apache/storm/bolt/JoinBolt.java
```diff
@@ -191,9 +191,9 @@ public void execute(TupleWindow inputWindow) {
         for (ResultRecord resultRecord : joinResult.getRecords()) {
             ArrayList<Object> outputTuple = resultRecord.getOutputFields();
             if ( outputStreamName==null )
-                collector.emit( outputTuple );
+                collector.emit( resultRecord.tupleList, outputTuple );
```
satishd: It may be good to add a comment here, saying that anchoring with only matched input tuples as .
`WindowedOutputCollector` anchors with all the incoming tuples.

For others to understand the issue, collector used here is [`WindowedOutputCollector`](https://github.com/apache/storm/blob/master/storm-core/src/jvm/org/apache/storm/topology/WindowedBoltExecutor.java#L31) and it emits  output tuple by anchoring with all the input tuples with in that window.

storm-core/src/jvm/org/apache/storm/bolt/JoinBolt.java
```diff
@@ -191,9 +191,9 @@ public void execute(TupleWindow inputWindow) {
...
             else
-                collector.emit( outputStreamName, outputTuple );
+                collector.emit( outputStreamName, resultRecord.tupleList, outputTuple );
```
satishd: same as above comment.

docs/Joins.md
```diff
@@ -122,4 +122,18 @@ is treated as a single stream and joined against `stream2` from `bolt3`.
 2. Duplication of joined records across windows is possible when using Sliding Windows. This is because the tuples continue to exist
    across multiple windows when using Sliding Windows.
 
-
+3. If message timeouts are enabled, ensure the timeout setting (topology.message.timeout.secs) is large enough to comfortably
+   accommodate the window size, plus the additional processing by other spouts and bolts.
+
+4. Joining a window of two streams with M and N elements each, *in the worst case*, can produce MxN elements with every output tuple
```
satishd: Nice to see this case added in the doc here.

satishd:
+1, LGTM. 

@roshannaik You may want to address those minor nits.

docs/Joins.md
roshannaik: actually it should be  "one tuple from each input stream" ... a joined tuple will only be formed out of merging two input tuples (if joining two streams)

storm-core/src/jvm/org/apache/storm/bolt/JoinBolt.java
roshannaik: good idea.

roshannaik:
Thanks @satishd for the quick turnaround. Updated the PR with the fixes.

satishd:
@roshannaik TestJoinBolt is filed with errors in travis runs. Can you take a look into those failures?

revans2:
+1 for the concept, would like to understand the test failures though.

roshannaik:
@satishd 
I have fixed the UTs. The MockCollector.emit used in the UT needed to be updated to correspond with  the emit method used by the JoinBolt.  Have squashed the commits.

PR #2021, opened 2017-03-21
[STORM-2427] Fix event logger enable disable UI buttons

arunmahadevan:
Applying relevant commits from 1.x and some minor fix.


satishd:
+1, travis test failures does not seem to be related to this change.

PR #2023, opened 2017-03-21
STORM-2429: Properly validate supervisor.scheduler.meta

djudd-stripe:
https://issues.apache.org/jira/browse/STORM-2429


HeartSaVioR:
+1 Nice catch.

PR #2030, opened 2017-03-27
STORM-2435: Logging levels and consistency with console.log etc

XC-:
Logging now supports logging levels (trace, debug, info, …warn and error).
Logging methods now accept multiple arguments and argument types are not restricted.
Some style fixes to make it more consistent.


EitZei:
Could the new logging functions (`trace`, `debug` etc.) be somehow namespaced so that they don't mix up with tuple handling methods like `fail`.

Either create subject `storm.logger` or similar where you bind the functions or prefix the function names with `log` i.e. `logError`.

XC-:
Good point. In the Python version there is logTrace, logDebug etc. so it having it in the similar fashion in the storm.js would be more consistent.

PR #1932, opened 2017-02-08
[STORM-2194]  Stop ignoring socket timeout error from executor

ppoulosk:
This is a translation of the fix recommended by @revans2 here:

https://github.com/apache/storm/pull/1767

PR #2036, opened 2017-03-31
[STORM-2440] Prevent KafkaSpout from hanging under certain Kafka failure scenarios

nicom:
Issue: https://issues.apache.org/jira/browse/STORM-2440

### DIFF
--- a/storm-core/src/clj/org/apache/storm/daemon/executor.clj
+++ b/storm-core/src/clj/org/apache/storm/daemon/executor.clj
@@ -268,8 +268,10 @@
                                (catch Exception e
                                  (log-message "Error while reporting error to cluster, proceeding with shutdown")))
                              (if (or
-                                    (exception-cause? InterruptedException error)
-                                    (exception-cause? java.io.InterruptedIOException error))
+                                   (exception-cause? InterruptedException error)
+                                   (and
+                                     (exception-cause? java.io.InterruptedIOException error)
+                                     (not (exception-cause? java.net.SocketTimeoutException error))))
                                (log-message "Got interrupted excpetion shutting thread down...")
                                ((:suicide-fn <>))))
      :sampler (mk-stats-sampler storm-conf)


srdo:
+1, nice catch.

Here's the master version, for reference https://github.com/apache/storm/pull/1932.

storm-core/src/clj/org/apache/storm/daemon/executor.clj
```diff
@@ -267,9 +267,11 @@
                                ((:report-error <>) error)
                                (catch Exception e
                                  (log-message "Error while reporting error to cluster, proceeding with shutdown")))
-                             (if (or
-                                    (exception-cause? InterruptedException error)
-                                    (exception-cause? java.io.InterruptedIOException error))
+                             (if (and
```
HeartSaVioR: https://github.com/apache/storm/blob/master/storm-core/src/jvm/org/apache/storm/executor/error/ReportErrorAndDie.java#L41-L43

According to current implementation for master branch, it should be

```
(if (or
  (exception-cause? InterruptedException error)
  (and
    (exception-cause? java.io.InterruptedIOException error)
    (not (exception-cause? java.net.SocketTimeoutException error))))))
```

HeartSaVioR:
Please also change title to STORM-2440 instead of DEV-2440.

nicom:
I just found out that this is a duplicate of #1932

HeartSaVioR:
@nicom 
#1932 just addressed it for master branch so this PR is still valid. 
Could you double check my review comment and reopen / address?

storm-core/src/clj/org/apache/storm/daemon/executor.clj
nicom: Fixed

HeartSaVioR:
@nicom +1 Thanks for the quick fix.

HeartSaVioR:
This PR is merged into asf git but asf git seems have some issue which might be related to mirror trigger. I'll try sync up mirror tomorrow. This will be closed automatically after sync up.

PR #2038, opened 2017-04-01
STORM-2450:  Write resources into correct local directory

revans2:


HeartSaVioR:
+1 Nice finding.

revans2:
@HeartSaVioR it actually took me way too long beating my head against it thinking I broke the thing as part of trying to make local mode transparent.

PR #2039, opened 2017-04-02
fix md formatting for windows storm.local.dir example 

alexlehm:
jekyll needs a newline before the code quotes, while e.g. the markdown preview in atom and the github md renderer doesn't

### DIFF
--- a/docs/Setting-up-a-Storm-cluster.md
+++ b/docs/Setting-up-a-Storm-cluster.md
@@ -17,11 +17,11 @@ Here's a summary of the steps for setting up a Storm cluster:
 
 ### Set up a Zookeeper cluster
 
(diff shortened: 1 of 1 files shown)


alexlehm:
looks like atom has caused a few whitespace diffs since it removes trailing spaces, hope that is ok

docs/Setting-up-a-Storm-cluster.md
```diff
@@ -58,11 +58,12 @@ If the port that your Zookeeper cluster uses is different than the default, you
 ```yaml
 storm.local.dir: "/mnt/storm"
 ```
-If you run storm on windows,it could be:
+If you run storm on windows, it could be:
+
```
HeartSaVioR: Actual change would be only this line, right?

HeartSaVioR:
Btw I'm OK to adopt whitespace diff. You don't need to modify it.

docs/Setting-up-a-Storm-cluster.md
alexlehm: yes (actually the blank line, the other change is only cosmetic)

HeartSaVioR:
+1

HeartSaVioR:
Thanks @alexlehm I merged into master, 1.x, 1.0.x branches.
I've other works to do so site update will be followed up later.

PR #2040, opened 2017-04-02
STORM-2451: windows storm.cmd does not set log4j2 config file correct…

alexlehm:
…ly by default

the dir from the default config is log4j2 and that should be relative to STORM_HOME, however this is check was missing

Signed-off-by: alexlehm <alexlehm@gmail.com>

### DIFF
--- a/bin/storm-config.cmd
+++ b/bin/storm-config.cmd
@@ -102,6 +102,16 @@ FOR /F "delims=" %%i in (%CMD_TEMP_FILE%) do (
 )
 
 @rem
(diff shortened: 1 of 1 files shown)


HeartSaVioR:
I don't have Windows machine but code change looks good to me, so +1 assuming you tested this manually.

alexlehm:
yes, i am using the modified start script now

HeartSaVioR:
This PR is merged into asf git but asf git seems have some issue which might be related to mirror trigger. I'll try sync up mirror tomorrow. This will be closed automatically after sync up.

PR #2041, opened 2017-04-03
STORM-2452: Storm Metric classes are not thread safe

knusbaum:
Synchronizing metrics methods


HeartSaVioR:
I think thread-unsafety of metrics is by design. They're guaranteed to be called only a thread at once, so unless we get or update the value on other than Spout or Bolt's event thread, they should be safe. Updating them is occurred from critical path, and metrics doesn't need to have synchronization overhead with current design.

Btw, thread-unsafety on metrics is why I gave up metrics aggregation for worker side. Instead of struggling about metrics more, I'm just waiting for Metrics V2.

HeartSaVioR:
Please do some benchmark and check performance regression, and please let me know if it doesn't hurt performance.

PR #2047, opened 2017-04-06
STORM-2447: add in storm local to avoid having server on worker classpath

revans2:

PR #2049, opened 2017-04-09
STORM-2448: Add in Storm and JDK versions when submitting a topology

revans2:
This is based off of the local mode patch #2047 once it goes in I'll rebase it again, so it is simpler to see the changes.  All of the changes are in the last commit 7d1f346.

I'll be putting up pull requests to the other branches so that they can submit topologies with a version in them.

I have tested this with workers on the latest versions of branch-1.x, branch-1.0.x, and branch-0.10.x.

I am not totally sure how useful this will be for other groups so I would like feedback if we want the UI changes to stay or not, and how much documentation should I add.  Right now the docs are only in Config.java

### DIFF
--- a/storm-client/src/jvm/org/apache/storm/Config.java
+++ b/storm-client/src/jvm/org/apache/storm/Config.java
@@ -1321,6 +1321,51 @@ public class Config extends HashMap<String, Object> {
     public static final String SUPERVISOR_WORKER_LAUNCHER = "supervisor.worker.launcher";
 
     /**
+     * Map a version of storm to a worker classpath that can be used to run it.
+     * This allows the supervisor to select an available version of storm that is compatible with what a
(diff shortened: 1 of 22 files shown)


revans2:
@ptgoetz @HeartSaVioR I would appreciate a review of this.  I would like to know what people think about it.  If it is good I am happy to add in more documentation and to rip out the rename hack code, as it would no longer really be needed.

revans2:
@ptgoetz @HeartSaVioR ping again. This should give us the ability to do a rolling upgrade from 1.x to 2.x

revans2:
@kishorvpatil @knusbaum Could you two take a look at this and also the patches for 1.x etc.

knusbaum:
+1 LGTM.

PR #2052, opened 2017-04-09
STORM-2448: Report storm and jdk versions to nimbus on topology submission

revans2:
0.10.x support for #2049


HeartSaVioR:
I'd rather not support under 1.0 version at all, especially when we're starting the version line for 2.0.0.
Please let me know if you have users or customers still on 0.10.x.

revans2:
@HeartSaVioR We do have a large number on 0.10 and we are hoping that this will be a good way to transition them off to 2.x when we are ready to go to that version.  The plan is not to have 0.10 run under a 1.x cluster, but to have 0.10 or 1.x run under a 2.x cluster.  I have manually tested this and it does work.

revans2:
But I am happy to only add this into our own release of storm if you don't think this is something you want to try and support.

HeartSaVioR:
OK. Personally I'm in favor of dropping support too-old versions to require users trying to catch up the latest whenever possible, but if there're large number of 0.10.x users and you think this approach helps lowering transition curve, that should be great.

+1

PR #2053, opened 2017-04-10
[STORM-2455] Expose the window start and end timestamp in TupleWindow

arunmahadevan:


harshach:
@arunmahadevan this looks like backward incompatible change?

arunmahadevan:
@harshach, the earlier change (that added getTimestamp) was checked in to only the master branch. Since its not part of any released versions of Storm may be its not an issue if we can get this merged before the 2.0 release?

storm-client/src/jvm/org/apache/storm/windowing/Window.java
```diff
@@ -47,9 +47,17 @@
     List<T> getExpired();
 
     /**
-     * If processing based on event time, returns the watermark time otherwise the current timestamp.
+     * If processing based on event time, returns the window end time based on watermark otherwise
+     * returns the current timestamp.
```
satishd: Returning the current timestamp(at which this method is invoked) for non event-timestamp scenarios does not look to be intuitive. Users may assume `endTimeStamp` should return the timestamp at which the window is completed.

arunmahadevan: Thanks for noticing. Its actually returning the timestamp at which window ends for both event and processing time. I am invoking `System.currentTimeMillis` in `getEndTimestamp` which is un-necessary. Will reword the doc and refactor a bit to make it clear.

harshach:
+1

satishd:
+1 LGTM

PR #2054, opened 2017-04-10
STORM-2462 Adding regex mapper to KerberosPrincipalToLocal class

ambud:
In some environments it may be the case that the local unix usernames are different than the ones in Kerberos account. This feature lets you override the username and map it to the one locally available.

Added settings:
storm.principal.mapper.regex = what to replace in the existing principal
storm.principal.mapper.replacement = what to replace in the existing principal with

e.g.

Kerberos principal is 0001@storm.apache.org 
storm.principal.mapper.regex=^
storm.principal.mapper.replacement=A

Would result in local principal=A0001

https://issues.apache.org/jira/browse/STORM-2462

### DIFF
--- a/storm-client/src/jvm/org/apache/storm/security/auth/RegexKerberosPrincipalToLocal.java
+++ b/storm-client/src/jvm/org/apache/storm/security/auth/RegexKerberosPrincipalToLocal.java
@@ -1,0 +1,56 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.storm.security.auth;
+
+import java.util.Map;
+import java.security.Principal;
+
+/**
+ * Map a kerberos principal to a local user
+ */
+public class RegexKerberosPrincipalToLocal extends KerberosPrincipalToLocal {
+
+    private String inputRegex;
+	private String replacementString;
+
+	/**
+     * Invoked once immediately after construction
+     * @param storm_conf Storm configuration
+     */
+    public void prepare(@SuppressWarnings("rawtypes") Map storm_conf) {
+        Object object = storm_conf.get("storm.principal.mapper.regex");
+        if (object != null) {
+            inputRegex = object.toString();
+        }
+        object = storm_conf.get("storm.principal.mapper.replacement");
+        if (object != null) {
+            replacementString = object.toString();
+        }
+    }
+    
+    /**
+     * Convert a Principal to a local user name.
+     * @param principal the principal to convert
+     * @return The local user name.
+     */
+    public String toLocal(Principal principal) {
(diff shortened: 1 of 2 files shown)


storm-client/src/jvm/org/apache/storm/security/auth/RegexKerberosPrincipalToLocal.java
```diff
@@ -0,0 +1,56 @@
...
+import java.util.Map;
+import java.security.Principal;
+
+/**
+ * Map a kerberos principal to a local user
+ */
+public class RegexKerberosPrincipalToLocal extends KerberosPrincipalToLocal {
+
+    private String inputRegex;
+	private String replacementString;
```
revans2: nit: The indentation her and in general in some of the files  appears to be off.

storm-client/src/jvm/org/apache/storm/security/auth/RegexKerberosPrincipalToLocal.java
```diff
@@ -0,0 +1,56 @@
...
+
+package org.apache.storm.security.auth;
+
+import java.util.Map;
+import java.security.Principal;
+
+/**
+ * Map a kerberos principal to a local user
+ */
+public class RegexKerberosPrincipalToLocal extends KerberosPrincipalToLocal {
```
revans2: I would prefer to see this be inside KerberosPrincipalToLocal instead of having a separate class.

storm-client/src/jvm/org/apache/storm/security/auth/RegexKerberosPrincipalToLocal.java
```diff
@@ -0,0 +1,56 @@
...
+	/**
+     * Invoked once immediately after construction
+     * @param storm_conf Storm configuration
+     */
+    public void prepare(@SuppressWarnings("rawtypes") Map storm_conf) {
+        Object object = storm_conf.get("storm.principal.mapper.regex");
+        if (object != null) {
+            inputRegex = object.toString();
+        }
+        object = storm_conf.get("storm.principal.mapper.replacement");
```
revans2: Could we put all of the configs here inside of Config.java and then tag them with the annotations to make it a string so we don't need to call .toString on the things passed in.

ambud: Can we do conditional not null checks in Config.java? @revans2 

```
    @isString
    @NotNull
    public static final String STORM_PRINCIPAL_MAPPER_REGEX = "storm.principal.mapper.regex";
    
    @isString
    @NotNull
    public static final String STORM_PRINCIPAL_MAPPER_REPLACEMENT = "storm.principal.mapper.replacement";
```

storm-client/src/jvm/org/apache/storm/security/auth/RegexKerberosPrincipalToLocal.java
ambud: This implementation is useful for a few environments that have complications of usename mapping.

The standard KerberosPrincipalToLocal works for majority of the environments so my approach was to not alter it.

revans2: Really there are two ways to make this optional.  You can do it like this and making it a separate class or you can have the configs be optional, and if they are not set then no modification is done.  I kind of prefer the other one, as it feels more intuitive to me.

revans2:
I have been thinking about this, and I am not sure if a REGEX is really what we want.  Hadoop does this in a semi-standard way using auth_to_local rules like PAM does.

It might be worth lifting some of that code from hadoop and reusing it here instead of just a regular expression.

https://github.com/apache/hadoop/blob/trunk/hadoop-common-project/hadoop-auth/src/main/java/org/apache/hadoop/security/authentication/util/KerberosName.java

harshach:
@ambud we should add the auth_to_local rules an make that as part of storm.yaml config option . Adding regex will not be helpful here.

ambud:
@harshach got it, let me work on the implementation and update the code.

ambud:
Feature no longer needed

PR #2055, opened 2017-04-11
[STORM-2467] Use explicit charset when decoding from array backed buffer

marcust:
We had the case that this broke encoding for us because it used the default
system locale. Given the fact that this code uses an explicit encoding for the
other case and that it evolved from something that always used an explicit
encoding I believe this is more correct.


srdo:
Nice find. The change LGTM. You should create an issue for this at https://issues.apache.org/jira/browse/STORM so the bugfix gets tracked. Then rename this PR and the commit message to include the issue number (example https://github.com/apache/storm/pull/2057/commits).

marcust:
@srdo Done

HeartSaVioR:
+1 Nice finding.

PR #2059, opened 2017-04-11
STORM-2463: fix DRPCTest.testDequeueAfterTimeout test failure

vesense:
This is the hot fix for travis-ci build failure.

error info:
```
classname: org.apache.storm.daemon.drpc.DRPCTest / testname: testDequeueAfterTimeout

java.lang.AssertionError: null
	at org.junit.Assert.fail(Assert.java:86)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at org.junit.Assert.assertTrue(Assert.java:52)
	at org.apache.storm.daemon.drpc.DRPCTest.testDequeueAfterTimeout(DRPCTest.java:129)
```


vesense:
@revans2 100ms dosen't work, but 1s works well. 500ms sometimes still get failure.
 I changed the value to 1s.

revans2:
OK Still +1

PR #2064, opened 2017-04-12
STORM-1114: Handle race condition in Storm/Trident transactional stat…

ptgoetz:
Handle race condition in Storm/Trident transactional state when ZK nodes have already been created/deleted. This impacts certain Trident topologies, causing the worker to crash due to the propagation of runtime exceptions.

This change should be back ported to all branches we intend to release from.


revans2:
@ptgoetz and yes lets please do port this back

vesense:
+1

satishd:
+1

PR #2065, opened 2017-04-13
STORM-832: Allow config validation to be used by plugins/etc.

revans2:
I did this because hard coding the usage of DaemonConf.java into a few places just didn't look right, and I thought it would be better to just do this feature and make it generic so others can start using this too, and config settings for a plugin can be with the plugin, and not forced to be in Config or DaemonConfig


HeartSaVioR:
+1
Travis failing is only from integration test which seems unrelated.

PR #2068, opened 2017-04-13
Fix documentation code formatting

regisb:
Code blocks should always follow an empty line; otherwise, jekyll will
fail to properly format the code block. Also, github's code fencing (with triple backticks) does not work in the middle of item lists.

### DIFF
--- a/docs/State-checkpointing.md
+++ b/docs/State-checkpointing.md
@@ -22,31 +22,30 @@ For example a word count bolt could use the key value state abstraction for the 
 last committed by the framework during the previous run.
 3. In the execute method, update the word count.
 
- ```java
- public class WordCountBolt extends BaseStatefulBolt<KeyValueState<String, Long>> {
- private KeyValueState<String, Long> wordCounts;
- private OutputCollector collector;
- ...
-     @Override
-     public void prepare(Map stormConf, TopologyContext context, OutputCollector collector) {
-       this.collector = collector;
-     }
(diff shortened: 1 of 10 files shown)


docs/State-checkpointing.md
```diff
@@ -22,31 +22,30 @@ For example a word count bolt could use the key value state abstraction for the
...
-       Integer count = wordCounts.get(word, 0);
-       count++;
-       wordCounts.put(word, count);
-       collector.emit(tuple, new Values(word, count));
-       collector.ack(tuple);
-     }
- ...
- }
- ```
+    public class WordCountBolt extends BaseStatefulBolt<KeyValueState<String, Long>> {
```
HeartSaVioR: Maybe missed wrapping to code block.

HeartSaVioR:
@regisb Looks good overall without minor. Left comment.

docs/State-checkpointing.md
HeartSaVioR: Ah never mind. I got what you mean that `code fencing (with triple backticks) does not work in the middle of item lists`.

HeartSaVioR:
+1

HeartSaVioR:
Thanks @regisb, merged into master and 1.x-branch.

I've had other works to do, so website update will be done later.

regisb:
Great, thanks for the quick review and merge!

PR #2072, opened 2017-04-16
STORM-2477: Add generics to Config types

revans2:
Because this touches so much of the code to avoid having to constantly upmerge I have a script that will transform most of the code automatically. transform.sh . I don't plan on checking it in.  If everyone is OK with the change I will run the script and just check those changes in.  The other changes and small updates to make the result of running the script not produce compile errors.  Overall this drops more than 800 warnings from the code base (according to eclipse).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,7 @@ before_install:
   - rvm use 2.1.5 --install
   - nvm install 0.12.2
   - nvm use 0.12.2
+  - ./transform.sh
 install: /bin/bash ./dev-tools/travis/travis-install.sh `pwd`
 script:
   - /bin/bash ./dev-tools/travis/travis-script.sh `pwd` $MODULES
--- a/external/storm-hbase/src/main/java/org/apache/storm/hbase/bolt/AbstractHBaseBolt.java
+++ b/external/storm-hbase/src/main/java/org/apache/storm/hbase/bolt/AbstractHBaseBolt.java
@@ -51,11 +51,11 @@ public abstract class AbstractHBaseBolt extends BaseRichBolt {
     }
 
     @Override
-    public void prepare(Map map, TopologyContext topologyContext, OutputCollector collector) {
+    public void prepare(Map topoConf, TopologyContext topologyContext, OutputCollector collector) {
         this.collector = collector;
         final Configuration hbConfig = HBaseConfiguration.create();
 
-        Map<String, Object> conf = (Map<String, Object>)map.get(this.configKey);
+        Map<String, Object> conf = (Map<String, Object>)topoConf.get(this.configKey);
         if(conf == null) {
             throw new IllegalArgumentException("HBase configuration not found using key '" + this.configKey + "'");
         }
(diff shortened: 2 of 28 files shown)


external/storm-hbase/src/main/java/org/apache/storm/hbase/bolt/AbstractHBaseBolt.java
```diff
@@ -51,11 +51,11 @@ public AbstractHBaseBolt(String tableName, HBaseMapper mapper) {
     }
 
     @Override
-    public void prepare(Map map, TopologyContext topologyContext, OutputCollector collector) {
+    public void prepare(Map topoConf, TopologyContext topologyContext, OutputCollector collector) {
         this.collector = collector;
         final Configuration hbConfig = HBaseConfiguration.create();
 
-        Map<String, Object> conf = (Map<String, Object>)map.get(this.configKey);
+        Map<String, Object> conf = (Map<String, Object>)topoConf.get(this.configKey);
```
revans2: No because of the order of operations.  It is casting the value returned by `topoConf.get`, not `topoConf` itself.

revans2:
@srdo I addressed your review comments about empty map.

external/storm-hbase/src/main/java/org/apache/storm/hbase/bolt/AbstractHBaseBolt.java
srdo: Right, read a bit too fast there.

srdo:
@revans2 I meant you could get rid of the emptyMap variables entirely.

revans2:
@srdo can do. Wow you really don't like ugly code :)

revans2:
@srdo done

srdo:
@revans2 Sorry, didn't mean to nitpick you to death :) +1

revans2:
@srdo please continue to nit pick.  I like good clean code and I am happy to see people call me out on things that should be fixed.  I find that people who read the code closely and nit pick find more serious bugs during the review which saves me a lot of time debugging and fixing them later on.

revans2:
@ptgoetz @HeartSaVioR if either of you could take a look at this I would appreciate it.

.travis.yml
```diff
@@ -27,6 +27,7 @@ before_install:
   - rvm use 2.1.5 --install
   - nvm install 0.12.2
   - nvm use 0.12.2
+  - ./transform.sh
```
HeartSaVioR: Do we want to include running `transform.sh` to the build process?

HeartSaVioR:
Left a comment to know intention of changing build process. Others look great.

.travis.yml
revans2: Sorry Maybe I didn't make myself clear in the initial comments.  Because this is touching a lot of different places in the code I thought it would be best to have the transformations be done by a script, that way if other stuff is merged in, we don't have to worry as much about merge conflicts.

If I get the needed +1s on the change I will run the script, remove it and any mention of it in the code and check in the result before merging it to master.

I added this line so that travis could test the fully transformed code.

revans2: If you just want to see the fully transformed code I can do that, but it is as of right now it ends up being

```
 593 files changed, 1481 insertions(+), 1481 deletions(-)
```

HeartSaVioR: OK then please go ahead. I already run the script and have a look.

HeartSaVioR:
+1 Nice work.

revans2:
@HeartSaVioR Thanks I'll make the change

PR #2073, opened 2017-04-16
STORM-2478: Fix BlobStoreTest.testDeleteAfterFailedCreate on Windows

srdo:
See https://issues.apache.org/jira/browse/STORM-2478

I couldn't think of a way to fix this test without exposing the getKeyDir method. When the AtomicOutputStream in BlobStoreTest L181 is closed, the temporary file is renamed to `data`, which is deleted explicitly when the blob store is deleted. This makes the test meaningless since it's checking that the blob store directory can be deleted when there are temporary files in it.


HeartSaVioR:
+1 assuming that the test succeeded on Windows.

PR #2075, opened 2017-04-16
STORM-2479: Fix port assignment race condition in storm-webapp tests

srdo:
See https://issues.apache.org/jira/browse/STORM-2479

I think we should get rid of Utils.getAvailablePort entirely, but removing it from storm-webapp first seemed like a good opportunity to figure out how difficult it is.


srdo:
Earlier test failure seems unrelated, was a windowing bolt test failure.

storm-client/src/jvm/org/apache/storm/security/auth/ThriftServer.java
```diff
@@ -45,34 +48,50 @@ public ThriftServer(Map storm_conf, TProcessor processor, ThriftConnectionType t
...
+        } catch (IOException | TTransportException ex) {
+            handleServerException(ex);
+        }
+
     }
 
     public void stop() {
-        if (_server != null)
-            _server.stop();
+        _server.stop();
```
HeartSaVioR: `_server` is still able to be null, so checking null would be safer.

storm-client/src/jvm/org/apache/storm/security/auth/ThriftServer.java
```diff
@@ -45,34 +48,50 @@ public ThriftServer(Map storm_conf, TProcessor processor, ThriftConnectionType t
...
     }
 
     /**
      * @return true if ThriftServer is listening to requests?
      */
     public boolean isServing() {
-        return _server != null && _server.isServing();
+        return _server.isServing();
```
HeartSaVioR: same here.

HeartSaVioR:
Change looks good. Minor comments.

storm-client/src/jvm/org/apache/storm/security/auth/ThriftServer.java
srdo: I checked, AuthUtils.GetTransportPlugin can't return null, and neither included implementation of ITransportPlugin can return null :)

HeartSaVioR: Heh great. Thanks for explaining. :)

HeartSaVioR:
+1

PR #2076, opened 2017-04-17
[STORM-1642] Catch Exception when deserialization failed for 1.x branch.

liurenjie1024:
A new version for 1.x branch, the origin discussion can be found [here](https://github.com/apache/storm/pull/1316)

### DIFF
--- a/storm-core/src/jvm/org/apache/storm/serialization/KryoTupleDeserializer.java
+++ b/storm-core/src/jvm/org/apache/storm/serialization/KryoTupleDeserializer.java
@@ -22,11 +22,15 @@ import org.apache.storm.tuple.MessageId;
 import org.apache.storm.tuple.Tuple;
 import org.apache.storm.tuple.TupleImpl;
 import com.esotericsoftware.kryo.io.Input;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import java.io.IOException;
 import java.util.List;
 import java.util.Map;
 
 public class KryoTupleDeserializer implements ITupleDeserializer {
+    private static final Logger LOG = LoggerFactory.getLogger(KryoTupleDeserializer.class);
     GeneralTopologyContext _context;
     KryoValuesDeserializer _kryo;
     SerializationFactory.IdDictionary _ids;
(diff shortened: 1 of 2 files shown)


liurenjie1024:
@knusbaum Can you help to merge this?

storm-core/src/jvm/org/apache/storm/serialization/KryoTupleDeserializer.java
```diff
@@ -50,7 +54,8 @@ public Tuple deserialize(byte[] ser) {
             List<Object> values = _kryo.deserializeFrom(_kryoInput);
             return new TupleImpl(_context, values, taskId, streamName, id);
         } catch(IOException e) {
-            throw new RuntimeException(e);
+            LOG.error("Failed to deserialize tuple.", e);
```
revans2: Storm is designed to be a fail fast system, this fix appears to be doing the opposite of that.  If we cannot deserialize a tuple we need to make it very very clear to the user that something bad is happening.  Sadly in my experience even crashing the worker is not always enough, but in most cases it is.  Can we do a System.exit after logging the error instead of returning null?

liurenjie1024: In our case this happens because storm's internal message exchange protocol is vulnerable to foreign  attack. If deserialization fails, you should warn the user and go on with processing.

revans2: Then we need to secure the connection against attach and not mask real errors.

govind-menon:
@liurenjie1024 I have a pull-request https://github.com/apache/storm/pull/2095 that should handle this better. Specifically for your issue to protect against foreign attack, you can turn on SSL by following the instructions here - http://storm.apache.org/releases/0.10.0/SECURITY.html

revans2:
@govind-menon Actually I think it is SASL not SSL that you want.
if you just set
```
storm.messaging.netty.authentication=true
```

Either as the default for your cluster or even on a per topology bases (I think) each of the netty connections between workers will do a handshake and validate that each of them have the same secret, that is generated when the topology is submitted.

govind-menon:
@revans2 Oops - thanks.

PR #2080, opened 2017-04-18
STORM-2481 Upgrade Aether version to resolve Aether bug BUG-451566

HeartSaVioR:
* this resolves NPE issue on Aether and helps showing proper error message
* this also fixes the bug which dependencies root directories does not exist
  * after the patch it will create the directory if not found


harshach:
+1

PR #2081, opened 2017-04-21
[STORM-2482] Refactor the Storm auto credential plugins to be more usable

arunmahadevan:
1. Create a new storm module storm-autocreds
2. Move AutoHDFS and AutoHBase to storm-autocreds
3. Refactor code and accepts config keys for customizing the hadoop configuration for the plugins
4. Package the auto cred and dependency jars in the storm binary and deploy to external/storm-autocreds


docs/SECURITY.md
```diff
@@ -423,16 +423,18 @@ nimbus.impersonation.acl:
...
-To hide this from them in the common case plugins can be used to populate the credentials, unpack them on the other side into a java Subject, and also allow Nimbus to renew the credentials if needed.
-These are controlled by the following configs. topology.auto-credentials is a list of java plugins, all of which must implement IAutoCredentials interface, that populate the credentials on gateway 
-and unpack them on the worker side. On a kerberos secure cluster they should be set by default to point to org.apache.storm.security.auth.kerberos.AutoTGT.  
-nimbus.credential.renewers.classes should also be set to this value so that nimbus can periodically renew the TGT on behalf of the user.
+To hide this from them in the common case plugins can be used to populate the credentials, unpack them on the other side into a java Subject, and also allow Nimbus to renew the credentials if needed. These are controlled by the following configs.
+ 
+`topology.auto-credentials` is a list of java plugins, all of which must implement the `IAutoCredentials` interface, that populate the credentials on gateway
```
priyank5485: Is gateway here nimbus or namenode like entity? Should we rename it to nimbus?

docs/storm-hbase.md
```diff
@@ -56,22 +56,43 @@ The approach described above requires that all potential worker hosts have "stor
...
+Instead of specifying key values you may also directly specify the resource files for e.g.,
+
+```
+cluster1: [{"resources": ["/path/to/core-site1.xml", "/path/to/hbase-site1.xml"]}]
+cluster2: [{"resources": ["/path/to/core-site2.xml", "/path/to/hbase-site2.xml"]}]
+```
+
+Storm will download the tokens separately for each of the clusters and populate it into the subject and also renew the tokens periodically. 
+This way it would be possible to run multiple bolts connecting to separate HBase cluster within the same topology.
+
```
priyank5485: Just a few lines below it says "As nimbus is impersonating..... and it mentions storm.kerberos.principal. Is that right or should we change it to hbase.kerberos.principal ?

external/storm-autocreds/src/main/java/org/apache/storm/common/AbstractAutoCreds.java
```diff
@@ -0,0 +1,250 @@
...
+                }
+            }
+            LOG.info("Resources to load {}", resourcesToLoad);
+            // add configs from resources like hdfs-site.xml
+            for (String pathStr : resourcesToLoad) {
+                configuration.addResource(new Path(Paths.get(pathStr).toUri()));
+            }
+        }
+        LOG.info("Initializing UGI with config {}", configuration);
+        UserGroupInformation.setConfiguration(configuration);
```
priyank5485: Will this work if tow hbase or hdfs bolts connecting to different clusters are running in the same jvm? I think this was always the problem but re-factoring in this PR aims to be able to connect to two different clusters. Just checking if issue still exists.

priyank5485: Never mind. It seems that this will always be executed in sequence for hdfs and hbase. So it will be overwriting the configuration and then login and get the tokens.

priyank5485:
+1(non-binding)

external/storm-autocreds/src/main/java/org/apache/storm/common/AbstractAutoCreds.java
```diff
@@ -0,0 +1,250 @@
...
+                    LOG.info("configKey = {}, config = {}", key, config);
+                }
+            }
+        }
+    }
+
+    @Override
+    public void populateCredentials(Map<String, String> credentials, Map conf) {
+        try {
+            if (configKeys != null) {
```
omkreddy: configKeys is initialised to ArrayList. we need to use configKeys.isEmpty() check here.

external/storm-autocreds/src/main/java/org/apache/storm/common/AbstractAutoCreds.java
```diff
@@ -0,0 +1,250 @@
...
+            }
+            LOG.info("Tokens added to credentials map.");
+        } catch (Exception e) {
+            LOG.error("Could not populate credentials.", e);
+        }
+    }
+
+    private Map<String, Object> updateConfigs(Map topologyConf) {
+        Map<String, Object> res = new HashMap<>(topologyConf);
+        if (configKeys != null) {
```
omkreddy: same as above

external/storm-autocreds/src/main/java/org/apache/storm/common/AbstractAutoCreds.java
```diff
@@ -0,0 +1,250 @@
...
+     */
+    @Override
+    public void updateSubject(Subject subject, Map<String, String> credentials) {
+        addCredentialToSubject(subject, credentials);
+        addTokensToUGI(subject);
+    }
+
+    protected Set<Pair<String, Credentials>> getCredentials(Map<String, String> credentials) {
+        Set<Pair<String, Credentials>> res = new HashSet<>();
+        if (configKeys != null) {
```
omkreddy: same as above

docs/SECURITY.md
arunmahadevan: Gateway is the host from where the topology is submitted.

docs/storm-hbase.md
arunmahadevan: I think it should be hbase.kerberos.principal. Will update.

harshach:
still +1 after the above comments.

harshach:
@arunmahadevan we need a PR for master

PR #2082, opened 2017-04-21
expose Tuple for node.js 

daluu:
in case people have uses for it, like in unit testing for example (that's what I was using it for)


harshach:
+1

vesense:
+1

PR #2084, opened 2017-04-24
STORM-2488: The UI user Must be HTTP

MichealShin:
The UI user Must be HTTP. Otherwise, the UI page can not be authorized


harshach:
+1

PR #2089, opened 2017-04-25
STORM-2490: Lambda support

vesense:
Write spout / bolt using Java8 lambda.
https://issues.apache.org/jira/browse/STORM-2490


storm-client/src/jvm/org/apache/storm/topology/TopologyBuilder.java
```diff
@@ -316,6 +322,68 @@ public BoltDeclarer setBolt(String id, IWindowedBolt bolt, Number parallelism_hi
     }
 
     /**
+     * Define a new bolt in this topology. This defines a lambda basic bolt, which is a
+     * simpler to use but more restricted kind of bolt. Basic bolts are intended
+     * for non-aggregation processing and automate the anchoring/acking process to
+     * achieve proper reliability in the topology.
+     *
+     * @param id the id of this component. This id is referenced by other components that want to consume this bolt's outputs.
+     * @param biConsumer lambda expression which is the instance of functional interface BiConsumer
```
srdo: Nitpick: This comment doesn't really say much about what the consumer is/does. Maybe something like "lambda expression that implements tuple processing for this bolt"?

examples/storm-starter/src/jvm/org/apache/storm/starter/LambdaTopology.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    @Override
+    protected int run(String[] args) throws Exception {
+        TopologyBuilder builder = new TopologyBuilder();
+
+        // example. spout1: generate random strings
+        // bolt1: get the first part of a string
+        // bolt2: output the tuple
+        builder.setSpout("spout1", () -> UUID.randomUUID().toString());
```
srdo: I'm wondering if it makes sense to add a note here to help people avoid the case where they refer to a field on LambdaTopology from the lambda and hit a NotSerializableException?

harshach:
@vesense this looks good. +1.

storm-client/src/jvm/org/apache/storm/lambda/LambdaBiConsumerBolt.java
```diff
@@ -0,0 +1,36 @@
...
+package org.apache.storm.lambda;
+
+import org.apache.storm.topology.BasicOutputCollector;
+import org.apache.storm.tuple.Tuple;
+
+public class LambdaBiConsumerBolt extends AbstractLambdaBolt {
+
+    private SerializableBiConsumer<Tuple,BasicOutputCollector> biConsumer;
+
+    public LambdaBiConsumerBolt(SerializableBiConsumer<Tuple,BasicOutputCollector> biConsumer) {
```
vesense: I prefer to keep the current names. Both Consumer and BiConsumer are functional interfaces accepting parameters and returning nothing, these two bolts are corresponding processor. And users don't access them directly when writing spout/bolt by lambda.

storm-client/src/jvm/org/apache/storm/topology/TopologyBuilder.java
vesense: Good catch. Will update.

examples/storm-starter/src/jvm/org/apache/storm/starter/LambdaTopology.java
vesense: Variable used in lambda expression should be final or effectively final (or it will cause compilation error), and variable type should implement the Serializable interface if it isn't primitive type (or it will cause not serializable exception).

@srdo Thanks for your reminding. I will add some NOTE to remind users.

vesense:
@srdo Addressed your comments. Please take a look again.

srdo:
@vesense Looks good, thanks. +1

storm-client/src/jvm/org/apache/storm/topology/TopologyBuilder.java
```diff
@@ -353,6 +421,31 @@ public void setStateSpout(String id, IRichStateSpout stateSpout, Number parallel
     }
 
     /**
+     * Define a new spout in this topology.
+     *
+     * @param id the id of this component. This id is referenced by other components that want to consume this spout's outputs.
+     * @param supplier lambda expression that implements tuple generating for this spout
+     * @throws IllegalArgumentException if {@code parallelism_hint} is not positive
+     */
+    public SpoutDeclarer setSpout(String id, SerializableSupplier<Object> supplier) throws IllegalArgumentException {
```
arunmahadevan: make it a generic method
`public <T> SpoutDeclarer setSpout(String id, SerializableSupplier<T> supplier) throws IllegalArgumentException {`

Otherwise following wont work 
```java
        SerializableSupplier<String> supplier = () -> UUID.randomUUID().toString();
        builder.setSpout("spout1", supplier);
```

storm-client/src/jvm/org/apache/storm/topology/TopologyBuilder.java
```diff
@@ -353,6 +421,31 @@ public void setStateSpout(String id, IRichStateSpout stateSpout, Number parallel
...
+     * Define a new spout in this topology with the specified parallelism. If the spout declares
+     * itself as non-distributed, the parallelism_hint will be ignored and only one task
+     * will be allocated to this component.
+     *
+     * @param id the id of this component. This id is referenced by other components that want to consume this spout's outputs.
+     * @param parallelism_hint the number of tasks that should be assigned to execute this spout. Each task will run on a thread in a process somewhere around the cluster.
+     * @param supplier lambda expression that implements tuple generating for this spout
+     * @throws IllegalArgumentException if {@code parallelism_hint} is not positive
+     */
+    public SpoutDeclarer setSpout(String id, SerializableSupplier<Object> supplier, Number parallelism_hint) throws IllegalArgumentException {
```
arunmahadevan: same as above

examples/storm-starter/src/jvm/org/apache/storm/starter/LambdaTopology.java
```diff
@@ -0,0 +1,74 @@
...
+        // (or it will cause compilation error),
+        // and variable type should implement the Serializable interface if it isn't primitive type
+        // (or it will cause not serializable exception).
+        Prefix prefix = new Prefix("Hello lambda:");
+        String suffix = ":so cool!";
+
+        builder.setSpout("spout1", () -> UUID.randomUUID().toString());
+        builder.setBolt("bolt1", (tuple, collector) -> {
+            String[] parts = tuple.getStringByField("lambda").split("\\-");
+            collector.emit(new Values(prefix + parts[0] + suffix));
```
arunmahadevan: What happens if the bolt emit multiple fields ? Explicitly call this out in the javadoc or wherever appropriate. If you expect only one field better accept a `BiFunction` than a BiConsumer and emit the return value of the function.

storm-client/src/jvm/org/apache/storm/topology/TopologyBuilder.java
arunmahadevan: may be better:
```java 
public SpoutDeclarer setSpout(String id, SerializableSupplier<?> supplier) throws IllegalArgumentException
```

vesense: Nice catch. Fixed.

storm-client/src/jvm/org/apache/storm/topology/TopologyBuilder.java
vesense: Fixed.

examples/storm-starter/src/jvm/org/apache/storm/starter/LambdaTopology.java
vesense: I think we should support multiple fields. I added a parameter `fields` to support user defined output fields, and updated the example.
```
setBolt(String id, SerializableBiConsumer<Tuple,BasicOutputCollector> biConsumer, String... fields)
```

vesense:
@arunmahadevan Addressed your comments. Please take a look again.

examples/storm-starter/src/jvm/org/apache/storm/starter/LambdaTopology.java
```diff
@@ -0,0 +1,75 @@
...
+
+        // example. spout1: generate random strings
+        // bolt1: get the first part of a string
+        // bolt2: output the tuple
+
+        // NOTE: Variable used in lambda expression should be final or effectively final
+        // (or it will cause compilation error),
+        // and variable type should implement the Serializable interface if it isn't primitive type
+        // (or it will cause not serializable exception).
+        Prefix prefix = new Prefix("Hello lambda:");
```
arunmahadevan: Any reason why this is not just a String?

vesense: This is just an example to prove:
>variable type should implement the Serializable interface if it isn't primitive type (or it will cause not serializable exception).

PR #2090, opened 2017-04-25
STORM-2489: Overlap and data loss on WindowedBolt based on Duration

arunmahadevan:
- Fixed time eviction to not evict events beyond current timestamp.
- Modified time trigger to compute the reference tim based on window duration.


harshach:
+1

storm-client/src/jvm/org/apache/storm/windowing/TimeTriggerPolicy.java
```diff
@@ -62,7 +63,9 @@ public void reset() {
 
     @Override
     public void start() {
-        executorFuture = executor.scheduleAtFixedRate(newTriggerTask(), duration, duration, TimeUnit.MILLISECONDS);
+        // initial delay is slightly less than the duration so that the initial tuples wont't expire due to time drift
+        long initialDelay = duration - Math.min((long) (duration * .1), 100);
+        executorFuture = executor.scheduleAtFixedRate(newTriggerTask(), initialDelay, duration, TimeUnit.MILLISECONDS);
```
satishd: If the initial delay is slightly less than the given duration then its subsequent scheduling will happen after the given period of initialDelay which can be earlier than expected. Is not it?

storm-client/src/jvm/org/apache/storm/windowing/TimeTriggerPolicy.java
```diff
@@ -109,13 +112,22 @@ private Runnable newTriggerTask() {
         return new Runnable() {
             @Override
             public void run() {
+                // compute the current trigger timestamp based on prevTriggerTimestamp,
+                // since the calculation based on System.currentTimeMillis might have a slight drift
+                long now = System.currentTimeMillis();
+                long triggerTs = prevTriggerTimestamp == 0 ? now : prevTriggerTimestamp + duration;
+                prevTriggerTimestamp = triggerTs;
+                if (Math.abs(triggerTs - now) > 1000) {
```
satishd: Did we think about an option to execute a bolt in a different thread to avoid timing issues in trigger because of user code?

storm-client/src/jvm/org/apache/storm/windowing/TimeTriggerPolicy.java
arunmahadevan: Subsequent scheduling will happen after "duration" which would be the window sliding interval.

storm-client/src/jvm/org/apache/storm/windowing/TimeTriggerPolicy.java
arunmahadevan: The timing issues are due to the scheduled executor (which uses nanoTime) which does not exactly trigger after the duration when measured with System.currentTimeMillis. Theres a mismatch of a few ms, which is addressed here. In general user code is expected to return before the next trigger. If user code takes more time than the trigger interval still we set the next trigger timestamp correctly and execute as soon as they return from execute, but this should be very rare and users are not expected to do this, so we log the warning. I don't think we want multiple concurrent executes (i.e trigger execute when they haven't returned from the first execute)

storm-client/src/jvm/org/apache/storm/windowing/TimeTriggerPolicy.java
satishd: subsequentTrigger(x) = initialDelay + x*(delay)
Right, but the subsequent triggers may happen earlier as the initialDelay is slightly reduced.

storm-client/src/jvm/org/apache/storm/windowing/TimeTriggerPolicy.java
satishd: OK, Throwing warning incase of large drifts looks fine for now.

Multiple concurrent executes will not happen as there is only one thread which executes the bolt and the subsequent bolt executions wait till the current execution is done.

storm-client/src/jvm/org/apache/storm/windowing/TimeTriggerPolicy.java
arunmahadevan: We cant't guarantee the absolute time at which the trigger will fire since it depends on when and how long storm takes to initialize the bolt. Only the relative time difference between two window firings can be guaranteed. The issue was that the first trigger was firing a few milliseconds after the duration which was causing the initial tuples to directly expire. I have refactored the logic to not depend on the initial delay being slightly lesser.

storm-client/src/jvm/org/apache/storm/windowing/TimeTriggerPolicy.java
```diff
@@ -62,7 +62,9 @@ public void reset() {
 
     @Override
     public void start() {
-        executorFuture = executor.scheduleAtFixedRate(newTriggerTask(), duration, duration, TimeUnit.MILLISECONDS);
+        // initial delay is slightly less than the duration so that the initial tuples wont't expire due to time drift
+        long initialDelay = duration - Math.min((long) (duration * .1), 10);
+        executorFuture = executor.scheduleAtFixedRate(newTriggerTask(), initialDelay, duration, TimeUnit.MILLISECONDS);
```
srdo: Won't this still be vulnerable to losing initial tuples if the first execution is delayed longer than expected for some reason?

I'm wondering if we could make this more reliable by tracking the time we consider expired instead of/in addition to tracking what time it is now. Tuples between the prevExpired timestamp and the nowExpired timestamp are tuples we can't be sure have been processed, so they should be included in the window. Since the expired timestamps also move ~duration every onTrigger, we still get the tuples expired we expect. For the first onTrigger, we can just process all the tuples because we can tell it's the first call since the prevExpired timestamp isn't set yet.

srdo:
I'm wondering if you guys with some experience with windowing would mind taking a look at the question posted in https://github.com/apache/storm/pull/2127? It also has to do with tuples possibly getting expired before they're passed to the user's code, and affects the same policies as this PR.

storm-client/src/jvm/org/apache/storm/windowing/TimeTriggerPolicy.java
srdo: Also since this is likely going to rely on in-bolt state to avoid expiring tuples before the user sees them, I think we should add to the documentation that this set of policies don't guarantee at-least-once processing.

arunmahadevan: @srdo, I removed the initial delay and changed the logic so that the initial events are processed in the first window. Subsequent windows will be adjusted so that even if the trigger is slightly off, the events are considered. 

The processing time windows collects the events into windows based on system timestamp so the user may not see all the events. However the processed events are still guaranteed at-least once based on storm's acking mechanisms. 

With the proposed patch we do a best effort to slightly adjust the window sizes if the triggers are off and log warnings if it exceeds a threshold.

storm-client/src/jvm/org/apache/storm/windowing/TimeEvictionPolicy.java
```diff
@@ -54,7 +62,22 @@ public void track(Event<T> event) {
 
     @Override
     public void setContext(EvictionContext context) {
-        this.evictionContext = context;
+        EvictionContext prevContext = evictionContext;
+        evictionContext = context;
+        // compute window length adjustment (delta) to account for time drift
+        if (context != null && context.getSlidingInterval() != null) {
```
srdo: Nitpick: I'd rather not check for null unless there's a good reason. As far as I can tell there's no reason to reset the EvictionContext to null, so I'd rather we get an NPE if it happens unexpectedly, so we can fix what is likely a bug.

storm-client/src/jvm/org/apache/storm/windowing/TimeEvictionPolicy.java
```diff
@@ -41,8 +47,10 @@ public TimeEvictionPolicy(int windowLength) {
     public Action evict(Event<T> event) {      
         long now = evictionContext == null ? System.currentTimeMillis() : evictionContext.getReferenceTime();
         long diff = now - event.getTimestamp();
-        if (diff >= windowLength) {
+        if (diff >= (windowLength + delta)) {
             return Action.EXPIRE;
+        } else if (diff < 0) { // do not process events beyond current ts
+            return Action.KEEP;
```
srdo: Nitpick: I think the events are ordered when iterated over in WindowManager, so returning Action.STOP might be marginally faster.

srdo:
+1

storm-client/src/jvm/org/apache/storm/windowing/TimeEvictionPolicy.java
arunmahadevan: yes, ideally events should be ordered since its processing time, unless `System.currentTimeMillis` goes backwards, so I assumed it could be out of order.

storm-client/src/jvm/org/apache/storm/windowing/TimeEvictionPolicy.java
arunmahadevan: Addressed

PR #2092, opened 2017-04-27
STORM-2493: update documents to reflect the changes

vesense:


docs/storm-pmml.md
```diff
@@ -0,0 +1,37 @@
+#Storm PMML Bolt
```
erikdw: nit: spacing is not consistent here with the other docs that you've added (and the existing ones).  i.e., `#Foo` should be `# Foo`.

vesense: Nice catch. Will fix.

erikdw: I'm anal like that. ;-)

vesense:
@harshach / @revans2 Can you take a look?

PR #2093, opened 2017-04-28
STORM-2495:Integrate checkstyle check during build

vinodkc:
Integrated checkstyle in maven validate phase.Added google java style check

### DIFF
--- a/examples/storm-elasticsearch-examples/pom.xml
+++ b/examples/storm-elasticsearch-examples/pom.xml
@@ -98,7 +98,7 @@
                 <artifactId>maven-checkstyle-plugin</artifactId>
                 <!--Note - the version would be inherited-->
                 <configuration>
-                    <maxAllowedViolations>105</maxAllowedViolations>
+                    <maxAllowedViolations>235</maxAllowedViolations>
                 </configuration>
             </plugin>
         </plugins>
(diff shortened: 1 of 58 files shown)


vinodkc:
Due to new checkstyle warnings, size of build log file exceeded 4mb limit. 

Error from build 
"The log length has exceeded the limit of 4 MB (this usually means that the test suite is raising the same exception over and over)."

Can we increase the log size limit?

revans2:
@vinodkc We cannot increase the size of the log without paying money to travisci, or switching the build to the internal Jenkins servers at apache.  The later has been proposed but is not likely to happen soon.

My advice would be to set the useFile config to move the warnings to a text file that we can output the first X lines of if the build fails.  Then add in a maxAllowedViolations for each project with their current number of violations along with a JIRA to fix the warnings in the project and remove the limit.

vinodkc:
@revans2 , I've redirected the warnings to  files and added maxAllowedViolations for each project with their current number of violations.
Shall add an umbrella JIRA which contains multiple JIRA sub tasks to handle errors and warnings in each project?  
Then I shall  add JIRA reference in corresponding  pom.xml files

revans2:
@vinodkc Looks good we are likely going to need to modify the total number of violations when this is ready to be merged in.

I would love to see something that can find the checkstyle-result.xml that caused a failure (or at least has some of the failures in it) and will output some of the violations.  But that may be above and beyond.

revans2:
Still +1

PR #2100, opened 2017-05-04
STORM-2503: Fix lgtm.com alerts on equality and comparison operations.

adityasharad:
This PR addresses STORM-2503, by fixing a number of alerts found by lgtm.com at https://lgtm.com/projects/g/apache/storm/alerts/ that indicate genuine bugs in comparison operations in the Storm code.

These rules are:
- Equals on incomparable types (due to typos, fixed here)
- Hashed value without hashCode definition
- Inconsistent equals and hashCode (together with the previous item, this indicates some missing hashCode implementations, which have been added here)
- Comparison of identical values (indicating a comparator logic typo in `DefaultResourceAwareStrategy`, which is fixed here along with a fix to the overall comparator logic, to consider components with fewer connections as being smaller)
- Useless comparison test (indicating a variable name typo, which is fixed here)
- Suspicious reference equality test of boxed types (indicating a spurious use of `==`, replaced here by `equals`, on boxed types)

lgtm.com also provides PR integration, in case it would be useful to catch such alerts directly from incoming PRs.

### DIFF
--- a/storm-client/src/jvm/org/apache/storm/scheduler/resource/strategies/scheduling/DefaultResourceAwareStrategy.java
+++ b/storm-client/src/jvm/org/apache/storm/scheduler/resource/strategies/scheduling/DefaultResourceAwareStrategy.java
@@ -541,7 +541,7 @@ public class DefaultResourceAwareStrategy implements IStrategy {
                     connections2 += (componentMap.get(childId).execs.size() * o2.execs.size());
                 }
 
-                if (connections1 > connections1) {
+                if (connections1 > connections2) {
                     return -1;
                 } else if (connections1 < connections2) {
                     return 1;
@@ -567,9 +567,9 @@ public class DefaultResourceAwareStrategy implements IStrategy {
             public int compare(Component o1, Component o2) {
                 int connections1 = o1.execs.size() * thisComp.execs.size();
                 int connections2 = o2.execs.size() * thisComp.execs.size();
-                if (connections1 > connections2) {
+                if (connections1 < connections2) {
                     return -1;
-                } else if (connections1 < connections2) {
+                } else if (connections1 > connections2) {
                     return 1;
                 } else {
                     return o1.id.compareTo(o2.id);
--- a/examples/storm-starter/src/jvm/org/apache/storm/starter/TransactionalWords.java
+++ b/examples/storm-starter/src/jvm/org/apache/storm/starter/TransactionalWords.java
@@ -125,7 +125,7 @@ public class TransactionalWords {
             for (String key : _counts.keySet()) {
                 CountValue val = COUNT_DATABASE.get(key);
                 CountValue newVal;
-                if (val == null || !val.txid.equals(_id)) {
+                if (val == null || !val.txid.equals(_id.getTransactionId())) {
                     newVal = new CountValue();
                     newVal.txid = _id.getTransactionId();
                     if (val != null) {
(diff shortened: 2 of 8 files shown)


harshach:
Thanks @adityasharad  this looks good. +1. Can you please squash your commits into single one.

revans2:
@adityasharad Some of the tests are failing now specifically TestDefaultResourceAwareStrategy.

I'll try and spend some time looking into it to see if I can see why it is failing, because the test itself is not giving any useful information at all.

revans2:
@adityasharad so the test itself was testing for the wrong behavior, and when you fixed it the test stated to fail.  I will try to make a pull request to your repo with a fix for the test.

Please let me know if you want any help getting this in as I really would like to see these fixed.

revans2:
@adityasharad I put up https://github.com/adityasharad/storm/pull/1 to your repo that should fix the failing test.

adityasharad:
Thank you both!
@harshach - I have squashed my commits into one as requested.
@revans2 - I have merged your PR into this branch, so the commit is now included to resolve the test failure.

revans2:
I am +1 on the change, but because my fix for the test is a part of this please consider my +1 non-binding.

HeartSaVioR:
+1

HeartSaVioR:
@adityasharad 
Could you raise PR also for 1.x-branch? I think you're fixing some bugs which should be backported to 1.x version line but now both branches are diverged so can't cherry-pick it easily.

storm-client/src/jvm/org/apache/storm/scheduler/resource/strategies/scheduling/DefaultResourceAwareStrategy.java
```diff
@@ -567,9 +567,9 @@ public int compare(Component o1, Component o2) {
             public int compare(Component o1, Component o2) {
                 int connections1 = o1.execs.size() * thisComp.execs.size();
                 int connections2 = o2.execs.size() * thisComp.execs.size();
-                if (connections1 > connections2) {
```
jerrypeng: hey guys this change is not right.  This is suppose to sort by the number of connections in descending order so that tasks with most connections to each other can be attempted to be colocated first.  What is the reason for this change?

jerrypeng:
@adityasharad thanks for addressing these issues but can you tell me why you decided to change the compare function that I commented above?

@revans2 do you believe the sorting mentioned was incorrect before?

revans2:
@jerrypeng The reason that particular piece of code was flagged was because the logic and the return values were not consistent with each other.

returning -1 means less than (but the code was checking for a greater than).

As for me I was mostly looking at the tests and as all of the tests passed (except for one that was not actually impacted by this change but another change that was a really serious bug).  If we need to switch this particular part of the change back I am happy to do it.

adityasharad:
@jerrypeng: as @revans2 says the existing comparator logic was incorrect, so I put it back in a way that seemed consistent with the intent. I see that my changes make the `sortNeighbours` and `sortConnections` methods use opposite comparison logic -- if you think these should be identical then I am happy to make them match.

@HeartSaVioR I will port this to the 1.x branch.

jerrypeng:
@adityasharad is I agree with your change for the sortComponents on line 544.  Thank you for correcting that, but can you change the sorting logic of sortNeighbors to as before since I believe that is the correct sorting logic.  Thanks!

PR #2102, opened 2017-05-08
STORM-2496 Dependency artifacts should be uploaded to blobstore with …

HeartSaVioR:
…READ permission for all

* When uploading dependencies, set ACL properly so that it can be shared to other users as well
  * but allows only READ so that it can't be deleted from others

This is needed for Supervisor with secured cluster, since dependency artifacts are uploaded once and shared for all topologies.


harshach:
+1

PR #2104, opened 2017-05-08
[STORM-2505] Spout to support topic compaction

vivekmittal:
Kafka maintains the spout progress (offsets for partitions) which can hold a value which no longer exists (or offset+1 doesn't exist) in the topic due to following reasons
Topology stopped processing (or died) & topic got compacted (cleanup.policy=compact) leaving offset voids in the topic.

Topology stopped processing (or died) & Topic got cleaned up (cleanup.policy=delete) and the offset.
When the topology starts processing again (or restarted), the spout logic suggests that the next offset has to be (committedOffset+1) for the spout to make progress, which will never be the case as (committedOffset+1) has been removed from the topic and will never be acked.

**OffsetManager.java**
```
if (currOffset == nextCommitOffset + 1) {
        found = true;
        nextCommitMsg = currAckedMsg;
        nextCommitOffset = currOffset;
} else if (currOffset > nextCommitOffset + 1) {
        LOG.debug("topic-partition [{}] has non-continuous offset [{}]. It will be processed in a subsequent batch.", tp, currOffset);
}
```

This pull request maintains a emitted offset set in OffsetManager & smartly forward the offset to the next logical offset

### DIFF
--- a/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/OffsetManager.java
+++ b/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/OffsetManager.java
@@ -39,6 +39,8 @@ public class OffsetManager {
     private final long initialFetchOffset;
     // Last offset committed to Kafka. Initially it is set to fetchOffset - 1
     private long committedOffset;
+    // Emitted Offsets List
+    private final NavigableSet<Long> emittedOffsets = new TreeSet<>();
     // Acked messages sorted by ascending order of offset
     private final NavigableSet<KafkaSpoutMessageId> ackedMsgs = new TreeSet<>(OFFSET_COMPARATOR);
 
@@ -49,8 +51,12 @@ public class OffsetManager {
         LOG.debug("Instantiated {}", this);
     }
 
-    public void add(KafkaSpoutMessageId msgId) {          // O(Log N)
+    public void addToAckMsgs(KafkaSpoutMessageId msgId) {          // O(Log N)
         ackedMsgs.add(msgId);
+    }
+
+    public void addToEmitMsgs(long offset) {
+        this.emittedOffsets.add(offset);                  // O(Log N)
     }
 
     /**
@@ -68,13 +74,34 @@ public class OffsetManager {
         KafkaSpoutMessageId nextCommitMsg = null;     // this is a convenience variable to make it faster to create OffsetAndMetadata
 
         for (KafkaSpoutMessageId currAckedMsg : ackedMsgs) {  // complexity is that of a linear scan on a TreeMap
-            if ((currOffset = currAckedMsg.offset()) == nextCommitOffset + 1) {            // found the next offset to commit
+            currOffset = currAckedMsg.offset();
+            if (currOffset == nextCommitOffset + 1) {            // found the next offset to commit
                 found = true;
                 nextCommitMsg = currAckedMsg;
                 nextCommitOffset = currOffset;
-            } else if (currAckedMsg.offset() > nextCommitOffset + 1) {    // offset found is not continuous to the offsets listed to go in the next commit, so stop search
-                LOG.debug("topic-partition [{}] has non-continuous offset [{}]. It will be processed in a subsequent batch.", tp, currOffset);
-                break;
+            } else if (currOffset > nextCommitOffset + 1) {
+                if (emittedOffsets.contains(nextCommitOffset + 1)) {
+                    LOG.debug("topic-partition [{}] has non-continuous offset [{}]. It will be processed in a subsequent batch.", tp, currOffset);
+                    break;
+                } else {
+                    /*
+                        This case will arise in case of non contiguous offset being processed.
+                        So, if the topic doesn't contain offset = committedOffset + 1 (possible
+                        if the topic is compacted or deleted), the consumer should jump to
+                        the next logical point in the topic. Next logical offset should be the
+                        first element after committedOffset in the ascending ordered emitted set.
+                     */
+                    LOG.debug("Processed non contiguous offset. (committedOffset+1) is no longer part of the topic. Committed: [{}], Processed: [{}]", committedOffset, currOffset);
+                    final Long nextEmittedOffset = emittedOffsets.ceiling(nextCommitOffset);
+                    if (nextEmittedOffset != null && currOffset == nextEmittedOffset) {
+                        found = true;
+                        nextCommitMsg = currAckedMsg;
+                        nextCommitOffset = currOffset;
+                    } else {
+                        LOG.debug("topic-partition [{}] has non-continuous offset [{}]. Next Offset to commit should be [{}]", tp, currOffset, nextEmittedOffset);
+                        break;
+                    }
+                }
             } else {
                 //Received a redundant ack. Ignore and continue processing.
                 LOG.warn("topic-partition [{}] has unexpected offset [{}]. Current committed Offset [{}]",
@@ -113,6 +140,15 @@ public class OffsetManager {
                 break;
             }
         }
+
+        for (Iterator<Long> iterator = emittedOffsets.iterator(); iterator.hasNext();) {
+            if (iterator.next() <= committedOffset.offset()) {
+                iterator.remove();
+            } else {
+                break;
+            }
+        }
+
         LOG.trace("{}", this);
         
         LOG.debug("Committed offsets [{}-{} = {}] for topic-partition [{}].",
@@ -143,6 +179,7 @@ public class OffsetManager {
             + "topic-partition=" + tp
             + ", fetchOffset=" + initialFetchOffset
             + ", committedOffset=" + committedOffset
+            + ", emittedOffsets=" + emittedOffsets
             + ", ackedMsgs=" + ackedMsgs
             + '}';
     }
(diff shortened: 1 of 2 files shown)


srdo:
LGTM. Good solution :)

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/OffsetManager.java
```diff
@@ -68,13 +74,34 @@ public OffsetAndMetadata findNextCommitOffset() {
...
+                    break;
+                } else {
+                    /*
+                        This case will arise in case of non contiguous offset being processed.
+                        So, if the queue doesn't contain offset = committedOffset + 1 (possible
+                        if the queue is compacted or deleted), the consumer should jump to
+                        the next logical point in the queue. Next logical offset should be the
+                        first element after committedOffset in the ascending ordered emitted set.
+                     */
+                    LOG.debug("Processed non contiguous offset, the previously committed offset has been deleted from the topic. Committed: [{}], Processed: [{}]", committedOffset, currOffset);
```
hmcl: "the previously committed offset has been deleted from the topic" - isn't this incorrect? If the offset has been committed, it has to be in the topic. I believe you meant to say something else in this message. Can you please rephrase it. Thanks.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/OffsetManager.java
```diff
@@ -113,6 +140,15 @@ public long commit(OffsetAndMetadata committedOffset) {
                 break;
             }
         }
+
+        for (Iterator<Long> iterator = emittedOffsets.iterator(); iterator.hasNext();) {
```
hmcl: which scenario is this cleanup code accounting for ?

srdo: The emitted offsets have to get deleted at some point, or the set grows indefinitely

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/OffsetManager.java
```diff
@@ -68,13 +74,34 @@ public OffsetAndMetadata findNextCommitOffset() {
...
+                } else {
+                    /*
+                        This case will arise in case of non contiguous offset being processed.
+                        So, if the queue doesn't contain offset = committedOffset + 1 (possible
+                        if the queue is compacted or deleted), the consumer should jump to
+                        the next logical point in the queue. Next logical offset should be the
+                        first element after committedOffset in the ascending ordered emitted set.
+                     */
+                    LOG.debug("Processed non contiguous offset, the previously committed offset has been deleted from the topic. Committed: [{}], Processed: [{}]", committedOffset, currOffset);
+                    final Long nextEmittedOffset = emittedOffsets.ceiling(nextCommitOffset);
```
srdo: Nice catch, L96 seems like it'll throw an NPE when Java tries to unbox nextEmittedOffset

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/OffsetManager.java
```diff
@@ -49,10 +51,14 @@ public OffsetManager(TopicPartition tp, long initialFetchOffset) {
         LOG.debug("Instantiated {}", this);
     }
 
-    public void add(KafkaSpoutMessageId msgId) {          // O(Log N)
+    public void ack(KafkaSpoutMessageId msgId) {          // O(Log N)
```
srdo: Isn't add more ambiguous now that there are two different methods for adding offsets (emit())?

askprasanna:
Want to add the following note. This issue could potentially happen with non-compacted topics also. We had multiple topologies getting stuck due to this issue including a topic with "delete" config. Based on the learnings from the debugging myself and @vivekmittal did, even "delete" topics could encounter this issue under the following circumstance

- A Kafka topic with relatively short retention period
- A Storm topology that consumes from such a topic with a slow processing rate
- A fallback strategy to earliest offset OR uncommitted_earliest

Given the above, it is possible that while the spout/consumer is busy processing messages fetched from a particular partition, the async cleaner has run and cleaned up expired logs on the Kafka broker managing that partition. Now when the spout fetches the next batch during a subsequent poll it is likely to see message offsets that are not sequential to the ones it received in the previous batch.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/OffsetManager.java
```diff
@@ -68,13 +74,34 @@ public OffsetAndMetadata findNextCommitOffset() {
         KafkaSpoutMessageId nextCommitMsg = null;     // this is a convenience variable to make it faster to create OffsetAndMetadata
 
         for (KafkaSpoutMessageId currAckedMsg : ackedMsgs) {  // complexity is that of a linear scan on a TreeMap
-            if ((currOffset = currAckedMsg.offset()) == nextCommitOffset + 1) {            // found the next offset to commit
+            currOffset = currAckedMsg.offset();
```
vivekmittal: Using the currOffset at multiple places, so just extracted it in a variable.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/OffsetManager.java
vivekmittal: Yes the offset is there because it was part of the topic, but there is no guarantee that it will always be there. The offsets might be wiped off owing to compaction/deletion process in a kafka topic. This case will only arrive when the previously (committedOffset + 1) is no longer present in the topic. I will rephrase the message to explicitly specify (committedOffset+1) is no longer part of the topic.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/OffsetManager.java
vivekmittal: The OffsetManager now maintains acked as well as emitted set. As @srdo  pointed **add** will be a little ambiguous.
I can make them more verbose if needed. Something like **addAck** and **addEmit**

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/OffsetManager.java
vivekmittal: Will fix it.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/OffsetManager.java
```diff
@@ -68,13 +74,34 @@ public OffsetAndMetadata findNextCommitOffset() {
...
-                break;
+            } else if (currOffset > nextCommitOffset + 1) {
+                if (emittedOffsets.contains(nextCommitOffset + 1)) {
+                    LOG.debug("topic-partition [{}] has non-continuous offset [{}]. It will be processed in a subsequent batch.", tp, currOffset);
+                    break;
+                } else {
+                    /*
+                        This case will arise in case of non contiguous offset being processed.
+                        So, if the queue doesn't contain offset = committedOffset + 1 (possible
+                        if the queue is compacted or deleted), the consumer should jump to
```
harshach: Minor nit: "queue" => "topic"

vivekmittal: Thanks for pointing it out. Will change the text.

harshach:
@vivekmittal can you squash your commits into singe one.
"Topology stopped processing (or died) & topic got compacted (cleanup.policy=compact) leaving offset voids in the topic.

Topology stopped processing (or died) & Topic got cleaned up (cleanup.policy=delete) and the offset."

In both of these cases are we not getting OffsetOutofRange exception?

askprasanna:
We have set the auto.offset.reset property to 'latest' which is the default value in Kafka. The exception is thrown only when it is set to 'none'. Verified the same in Fetcher.java in Kafka clients library.

We do see kafka client log message indicating offset being reset.

vivekmittal:
Squashed the commits into one.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/OffsetManager.java
harshach: @vivekmittal I think "addAck" is better than just "ack" as its part of the storm apis might confuse the method name. Probably better name would be "addToAckMsgs" and "addToEmitMsgs"

harshach:
@vivekmittal over LGTM. I am +1 once the method name is addressed.
Thanks for finding & addressing the bug.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/OffsetManager.java
vivekmittal: Make sense. I will change it.

vivekmittal:
@harshach Method name is addressed. Squashed all my commits into one.

harshach:
+1. Thanks @vivekmittal

harshach:
@vivekmittal I think you need to  open another PR against 1.x-branch. Don't think this can be cherry-picked onto 1.x-branch.

vivekmittal:
@harshach Will do.

vivekmittal:
Raised PR to pull same changes in 1.x-branch 
https://github.com/apache/storm/pull/2107

PR #2108, opened 2017-05-10
STORM-2508:storm-solr enhancement: update solrj to 5.5, support custom SolrClientFactory and commit operation

aylei:
I have a case that the SolrCloud in my organization is protected by SSL + Basic Auth, so I need to provide customized SolrClient instance to SolrUpdateBolt. Likewise, the RestJsonSchemaBuilder cannot access Schema API without auth.
In addition, for "near real-time search", soft commit is preferred, so it is better to expose an interface for user to customize commit operation.
I think those features will also be useful for other people who needs to use custom SolrClient implementation and control the detail of commit operation. So, finally, I plan an enhancement for storm-solr:
1. update solrj and related dependencies to 5.5;
2. improve SolrConfig to support custom SolrClientFactory and CommitCallBack;
3. provide a SchemaBuilder implementation which use custom SolrClient to request Schema API;
4. elder SolrUpdateBolts intend to commit when recieve a tick, but by default BaseTickTupleAwareRichBolt will ignore tick, this PR also fix it.

### DIFF
--- a/external/storm-solr/pom.xml
+++ b/external/storm-solr/pom.xml
@@ -32,7 +32,7 @@
         <id>hmcl</id>
         <name>Hugo Louro</name>
         <email>hmclouro@gmail.com</email>
-    </developer>
+        </developer>
     </developers>
 
     <dependencies>
@@ -46,13 +46,13 @@
         <dependency>
             <groupId>org.apache.solr</groupId>
             <artifactId>solr-solrj</artifactId>
-            <version>5.2.1</version>
+            <version>5.5.0</version>
             <scope>compile</scope>
         </dependency>
         <dependency>
             <groupId>commons-codec</groupId>
             <artifactId>commons-codec</artifactId>
-            <version>1.3</version>
+            <version>1.4</version>
         </dependency>
         <dependency>
             <groupId>commons-httpclient</groupId>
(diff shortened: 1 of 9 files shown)


external/storm-solr/pom.xml
```diff
@@ -46,13 +46,13 @@
         <dependency>
             <groupId>org.apache.solr</groupId>
             <artifactId>solr-solrj</artifactId>
-            <version>5.2.1</version>
+            <version>5.5.0</version>
```
sachingsachin: 5.5.0 version is affected by a security problem and is also more than 1.5 years old.
http://lucene.apache.org/solr/news.html
Please consider moving to Solr 6.5.1 or 6.6.0

external/storm-solr/pom.xml
```diff
@@ -67,13 +67,13 @@
...
             <artifactId>solr-core</artifactId>
-            <version>5.2.1</version>
+            <version>5.5.0</version>
             <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.apache.solr</groupId>
             <artifactId>solr-test-framework</artifactId>
-            <version>5.2.1</version>
+            <version>5.5.0</version>
```
sachingsachin: To avoiding adding version everywhere, this should be made a property

PR #2112, opened 2017-05-12
[STORM-2510] update checkstyle configuration to lower violations

erikdw:
1. Update from checkstyle-6.11.2 to checkstyle-7.7.
2. Tweak google_checks.xml from checkstyle-7.7 to have the following changes that
are more inline with the way the storm code is written:
   * 4 space indents instead of 2
   * line-length limit is 120 instead of 100
3. Exclude the generated thrift code in storm-client from being checked.
4. Update all maxAllowedViolations to be in-sync with the current number of
violations, through use of the `update-all-pom-violations.bash` script that
is attached to STORM-2510.

NOTE: this also fixes STORM-2507.

### DIFF
--- a/examples/storm-elasticsearch-examples/pom.xml
+++ b/examples/storm-elasticsearch-examples/pom.xml
@@ -98,7 +98,7 @@
                 <artifactId>maven-checkstyle-plugin</artifactId>
                 <!--Note - the version would be inherited-->
                 <configuration>
-                    <maxAllowedViolations>235</maxAllowedViolations>
+                    <maxAllowedViolations>20</maxAllowedViolations>
                 </configuration>
             </plugin>
         </plugins>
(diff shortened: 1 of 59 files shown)


erikdw:
@vinodkc & @revans2 & @hmcc & @srishtyagrawal : please 👀 when you have a chance.  This is a follow-up to the [discussions](https://github.com/apache/storm/commit/63567ae4e9fe573467db41e13696cbc50176b27a#commitcomment-22096038) in PR #2083, and also an offline email thread about the changes in PR #2093.

revans2:
Like I said before I am fine with whatever convention we have, so long as we have one.

+1

revans2:
Oops we need a license in `storm-buildtools/storm_checkstyle.xml` or we need to white list it for rat.

erikdw:
@revans2 : ah, thanks for pointing out the need for license.  I copied how google-cloud-intellij is doing it: 
* https://github.com/GoogleCloudPlatform/google-cloud-intellij/blob/ba8c85810f65443d9ff411ad30448b377afa6012/google_checks.xml

I also increased the line limit to 140 and adjusted the maxAllowedViolations accordingly.

harshach:
+1

PR #2117, opened 2017-05-15
STORM-2515: Fix most checkstyle violations in storm-kafka-client

srdo:
Most of the remaining violations are a couple of methods missing javadocs in the Trident code, which I don't feel familiar enough with Trident to write. The last couple are because there's a method in KafkaSpoutConfig that contains the SSL abbreviation. The rules only allow 2 capital letters in an abbreviation. Renaming is a breaking change, but maybe this isn't an issue for 2.0?

I also added a commit making a few other changes I thought made sense. The KafkaSpoutMessageId.getMetadata method was including the TopicPartition 3 times. The found variable in OffsetManager.findNextCommitOffset can be replaced with a null check of nextCommitMsg. They are in a separate commit for now, so this is easier to review.

### DIFF
--- a/external/storm-kafka-client/pom.xml
+++ b/external/storm-kafka-client/pom.xml
@@ -152,7 +152,7 @@
                 <artifactId>maven-checkstyle-plugin</artifactId>
                 <!--Note - the version would be inherited-->
                 <configuration>
-                    <maxAllowedViolations>15</maxAllowedViolations>
+                    <maxAllowedViolations>9</maxAllowedViolations>
                 </configuration>
             </plugin>
         </plugins>
--- a/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutRetryService.java
+++ b/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutRetryService.java
@@ -21,8 +21,8 @@ package org.apache.storm.kafka.spout;
 import java.io.Serializable;
 import java.util.Collection;
 import java.util.Map;
-import org.apache.kafka.common.TopicPartition;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
+import org.apache.kafka.common.TopicPartition;
 
 /**
  * Represents the logic that manages the retrial of failed tuples.
@@ -78,6 +78,7 @@ public interface KafkaSpoutRetryService extends Serializable {
     boolean isScheduled(KafkaSpoutMessageId msgId);
 
     /**
+     * Get the number of messages ready for retry.
      * @return The number of messages that are ready for retry
      */
     int readyMessageCount();
@@ -85,7 +86,8 @@ public interface KafkaSpoutRetryService extends Serializable {
     /**
      * Gets the {@link KafkaSpoutMessageId} for the given record.
      * @param record The record to fetch the id for
-     * @return The id the record was scheduled for retry with, or a new {@link KafkaSpoutMessageId} if the record was not scheduled for retry.
+     * @return The id the record was scheduled for retry with,
+     *     or a new {@link KafkaSpoutMessageId} if the record was not scheduled for retry.
      */
     KafkaSpoutMessageId getMessageId(ConsumerRecord<?, ?> record);
 }
--- a/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/trident/KafkaTridentSpoutOpaque.java
+++ b/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/trident/KafkaTridentSpoutOpaque.java
@@ -46,7 +46,8 @@ public class KafkaTridentSpoutOpaque<K,V> implements IOpaquePartitionedTridentSp
     }
 
     @Override
-    public Emitter<List<TopicPartition>, KafkaTridentSpoutTopicPartition, KafkaTridentSpoutBatchMetadata<K,V>> getEmitter(Map<String, Object> conf, TopologyContext context) {
+    public Emitter<List<TopicPartition>, KafkaTridentSpoutTopicPartition, KafkaTridentSpoutBatchMetadata<K,V>> getEmitter(
+            Map<String, Object> conf, TopologyContext context) {
         return new KafkaTridentSpoutEmitter<>(kafkaManager, context);
     }
 
--- a/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/trident/KafkaTridentSpoutTransactional.java
+++ b/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/trident/KafkaTridentSpoutTransactional.java
@@ -25,14 +25,15 @@ import org.apache.storm.trident.spout.ISpoutPartition;
 import org.apache.storm.tuple.Fields;
 
 // TODO
-public class KafkaTridentSpoutTransactional<Ps, P extends ISpoutPartition, T> implements IPartitionedTridentSpout<Ps, P, T> {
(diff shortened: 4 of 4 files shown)


srdo:
If someone wants to take a stab at writing the last bit of javadoc, and renaming the methods in KafkaSpoutConfig is fine, we could get rid of the last few warnings.

srdo:
The test failure is due to https://github.com/apache/storm/pull/2118

HeartSaVioR:
@srdo 
Since we're already struggling with checkstyle violations, I would like to merge this in, but we should take "breaking change" into consideration. 
I'm +1 to 7fdc48a and 49784bf and would like to merge them in via cherry-picking, but would like to see others opinions for f9e3331d with such conventions.

srdo:
@erikdw @revans2 @vinodkc You guys might have an opinion on this, since I've seen you discuss checkstyle on other PRs. How do you feel about the method renames in KafkaSpoutConfig in f9e331ddb3e8c1d917a2c9e27c2afbce7ba3dad0?

I think we shouldn't make this kind of breaking change in 1.x, but 2.0 isn't released yet so it's probably OK here?

HeartSaVioR:
@srdo
I cherry-picked two commits and lessened max checkstyle violations to 15 because violation count is still 11 after applying two commits.

Could we initiate discussion around `f9e331d` via dev@ mailing list?

srdo:
@HeartSaVioR Sure. Could you start the thread, I'm not sure what the discussion should be? Is it about making breaking changes in 2.0, the method name convention or something else?

HeartSaVioR:
@srdo 
I think both. 
Main issue is "are we OK to introduce breaking change because of method convention?", and I also would like to hear that "is setSslKeystore better than setSSLKeystore".
We may reach consensus easily for latter but maybe not former. Breaking backward compatibility is a bit annoying to the user, and we already did it once for 1.1.0.

srdo:
@HeartSaVioR Okay. I started a thread on the mailing list about those issues. Thanks for reviewing.

HeartSaVioR:
@srdo Thanks for the quick action!

srdo:
@HeartSaVioR How do you feel about merging this? It doesn't seem like anyone was overly worried about the API change in 2.x

erikdw:
@srdo : my opinion runs contrary to others it seems.  For every breaking API change my team & company incur a bunch of coordination work.  We have many independent internal engineering teams that are entirely responsible for their own topology code, so if there is a breaking API change we need to work with each of them to get them to modify their code appropriately.

Rather than breaking an API to adhere to a belatedly-applied code standard, we could take the approach of allowing explicit exemptions from the uppercase-in-identifier standard via annotations:
* https://stackoverflow.com/a/22556386

This would require another modification to the base checkstyle.xml file, since Google's defaults don't allow this explicit one-off suppression via annotations.

If we don't have the stomach for using this suppression mechanism, then we should at the minimum backport the changed API identifier as an available method in every active storm version branch.  That will help minimize upgrade pains since people can make the changes when upgrading to 1.1.x, and not have to make this particular change when upgrading to 2.x.

srdo:
@erikdw Sorry, didn't get much response on the mailing list thread, so expected that people just didn't care. I'd be happy with either solution, but prefer deprecating the existing methods and adding the renamed ones in 1.0.x and 1.1.x. Then your teams should be able to fix the deprecation warnings when they have time instead of having to do it immediately.

I don't really feel strongly about it so if you'd prefer going the suppression route let me know, and I'll do that instead. I'd just like to have a general rule for how we resolve this type of violation, since it's an issue in other PRs too (e.g. https://github.com/apache/storm/pull/2146/files)

erikdw:
@srdo : I might be in the minority (though it's hard to tell, since as you noted the email thread didn't get much traction), but I don't think there's much advantage in changing existing ***external*** APIs just for adherence with this specific coding style standard.  I should note that we're not actually using `storm-kafka-client` in our users' topologies yet because we haven't been able to upgrade to Storm 1.0+ yet (partially because of backwards-incompatible changes in code and behavior).

In my view, breaking of backwards compatibility should be done only when actually necessary; e.g., changing package from `backtype` to `org.apache` was a good reason, since we needed to make the project properly namespaced.  Whereas coding style adherence is not a very convincing reason.

Maybe someone with a view on the number & scale of breaking changes in Storm 2.0 as compared to 1.0 could speak up?  I'm fairly ignorant about the "gestalt" of what is happening in Storm 2.0 and so don't know the answer to that question.  i.e., if everyone is going to have to make a bunch of changes *anyways* then it's probably "no big deal" to slap in this method name casing change.  But if 2.0 isn't highly divergent as compared to 1.0 for topology authors, then I'd argue this isn't worth it.

srdo:
@erikdw Doesn't seem like anyone else is going to chime in. I can't speak for most of the Storm components, but I believe at least storm-kafka-client has a fair number of breaking API changes between 1.x and 2.x. It also seems likely to me that the Java port work would have broken some APIs. Looking at some topology code I have lying around, there's a decent number of compile errors after switching to 2.0.0-SNAPSHOT from 1.1.0

I think if we're going to rename methods it would be good to get them all out of the way in a single release. That way it isn't going to become a regular nuisance. We could make sure to get them all fixed now, and then put deprecations for all of them in 1.x?

I agree that this isn't that important of a change, but my impression is that people are probably going to need to change their code anyway when migrating to 2.x, so I don't know that it adds that much of an additional cost.

erikdw:
@srdo : thanks for your perspective on this.   I wonder if we could do this:
* we adhere to the no-uppercase-acronyms rule we are debating now, thus creating some breaking changes
** as stated previously, we make an effort to backport these changes into storm-1.0.x and/or storm-1.x, keeping both the new name and the old name for the methods, and then we mark the previous methods as deprecated.
* maybe we can add the feature to `storm_checkstyle.xml` that *allows* checkstyle-suppression annotations?  Or maybe we just defer that for now.

srdo:
@erikdw That sounds good to me. I'd like to get all the changes done in one PR (plus two for 1.x, 1.0.x) if possible, so we don't have to deal with this again later. Will probably take a look at it sometime this week.

It makes sense to wait to allow suppression until we need it IMO. We can always add it later if we need it.

srdo:
Reverted the KafkaSpoutConfig method name changes. I'd rather do them in the PR mentioned above.

HeartSaVioR:
@srdo @erikdw 
Thanks for the great discussion and like the approach. Please go ahead for the follow up items.
+1 for current PR.

harshach:
@srdo @HeartSaVioR @erikdw 
I understand this PR is merged. But we should be extremely careful when we break the backward incompatibility , if it justifies better implementation of a connector yes but I don't agree on  doing this for check style issues. I think we need to propose a better way to get the changes in. We can adopt the KIP style proposals in Storm community that can be discussed and voted upon before we agree. We can skip small changes/bug fixes out of this process and use it bigger changes and back-ward incompatible changes. I started discussion thread we can continue there. Thanks.

srdo:
@harshach This PR doesn't (as far as I can tell) contain any breaking changes anymore.

I agree that it would be good to have a better process for polling opinion regarding some changes. This question went on the mailing list and was also discussed here, but the number of participants was very small IMO.

harshach:
@srdo Sorry not meant as negative for the PR. But want to get a better exposure to everyone on changes that we make and for users/devs who might not be able to follow dev list day-in/day-out and its easy to get buried in the new emails in mailing list. I started discussion around having KIP (Kafka) proposal docs, so that there is a central we are documenting the critical fixes and changes.

srdo:
@harshach Sounds good. Is the discussion on the dev list? I'm not seeing it.

harshach:
@srdo thought I sent and replied your earlier emails as well. Looks like some issue with gmail they are showing as sent but didn't reached the mailing list. I sent them again.

PR #2118, opened 2017-05-16
STORM-2516: Fix timing issues with testPrepareLateTupleStreamWithoutB…

srdo:
…uilder

The test was sleeping a bit and hoping that WaterMarkEventGenerator.run() had been run by the executor in that class before the test woke up. Sometimes that wasn't happening. Instead we're calling the method manually now, like the rest of the tests involving that class.

I also fixed the checkstyle violations in WaterMarkEventGenerator.

As far as I can tell, the lastWatermarkTs variable in WaterMarkEventGenerator should be volatile or an AtomicLong, since it is being written to in run() with the expectation that it can be seen by the thread calling track().

### DIFF
--- a/storm-client/src/jvm/org/apache/storm/windowing/WaterMarkEventGenerator.java
+++ b/storm-client/src/jvm/org/apache/storm/windowing/WaterMarkEventGenerator.java
@@ -15,12 +15,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.apache.storm.windowing;
 
-import org.apache.storm.generated.GlobalStreamId;
-import org.apache.storm.topology.FailedException;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
+package org.apache.storm.windowing;
 
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


storm-client/src/jvm/org/apache/storm/windowing/WaterMarkEventGenerator.java
```diff
@@ -45,17 +46,25 @@
     private final Map<GlobalStreamId, Long> streamToTs;
     private final ScheduledExecutorService executorService;
     private final int interval;
+    private final AtomicLong lastWaterMarkTs;
```
arunmahadevan: this could be a volatile instead of AtomicLong

srdo: Yes. Why do you prefer volatile? (I'll be happy to change it, I'm just wondering if there's a drawback to using AtomicX)

arunmahadevan: It doesn't make much difference except the extra overhead of the AtomicLong object and we don't require the additional atomic operations from AtomicLong.

srdo: Replaced with volatile

arunmahadevan:
+1

PR #2119, opened 2017-05-16
STORM-2509 Close and perform rotation actions for retired writers in …

ryanpersaud:
…AbstractHDFSBolt.


HeartSaVioR:
@ryanpersaud Thanks for contributing. We introduced checkstyle and the build is failing with checkstyle violation. Could you fix the violation from your changeset and update the patch?

PR #2120, opened 2017-05-17
STORM-2518 Handles empty name for "USER type" ACL when normalizing ACLs

HeartSaVioR:
* if ACLs have both empty name and user, discard empty name ACLs
* add default ACL for user when neither empty name nor user ACL exists
* populate empty name ACLs to users in Subject: set name to user for each

PR for 1.x : #2121


HeartSaVioR:
Tested both secure and non-secure cluster.
Build error comes from checkstyle limit. Build result for 1.x PR seems OK.

HeartSaVioR:
@harshach Please take a look. Thanks in advance.

harshach:
+1

PR #2121, opened 2017-05-17
(1.x) STORM-2518 Handles empty name for "USER type" ACL when normalizing ACLs

HeartSaVioR:
* if ACLs have both empty name and user, discard empty name ACLs
* add default ACL for user when neither empty name nor user ACL exists
* populate empty name ACLs to users in Subject: set name to user for each

PR for master : #2120


harshach:
+1

PR #2122, opened 2017-05-17
STORM-2519: Modify AbstractAutoCreds to look for configKeys in both nimbus and topology configs.

arunmahadevan:
@HeartSaVioR , please take a look.


harshach:
+1

PR #2126, opened 2017-05-19
Add the option to set client.id to storm-kafka

carl34:
Add the ability to set the Kafka client.id property to storm-kafka and storm-kafka-client (https://issues.apache.org/jira/browse/STORM-2524).

storm-kafka example:
```
	SpoutConfig spoutConfig = new SpoutConfig(
		brokers,
		topicConfs.topic,
+		"client_id-" + confs.consumerGroupId + "-" + topicConfs.topic,
		"/consumers",
		confs.consumerGroupId + "-" + topicConfs.topic
	);
```

storm-kafka-client example:
```
	KafkaSpoutConfig<String,String> kafkaSpoutConfig = KafkaSpoutConfig.builder(KAFKA_LOCAL_BROKER, TOPIC)
		.setGroupId(confs.consumerGroupId + "-" + topicConfs.topic)
+		.setClientId("client_id-" + confs.consumerGroupId + "-" + topicConfs.topic)
		.build();
```

### DIFF
--- a/docs/storm-kafka.md
+++ b/docs/storm-kafka.md
@@ -57,11 +57,14 @@ The optional ClientId is used as a part of the ZooKeeper path where the spout's 
 
 There are 2 extensions of KafkaConfig currently in use.
 
-Spoutconfig is an extension of KafkaConfig that supports additional fields with ZooKeeper connection info and for controlling
-behavior specific to KafkaSpout. The Zkroot will be used as root to store your consumer's offset. The id should uniquely
-identify your spout.
+SpoutConfig is an extension of KafkaConfig that supports additional fields with ZooKeeper connection info and for controlling
+behavior specific to KafkaSpout.
+The clientId will be used to identify requests which are made using the Kafka Protocol.
+The zkRoot will be used as root to store your consumer's offset.
+The id should uniquely identify your spout.
 
 ```java
+public SpoutConfig(BrokerHosts hosts, String topic, String clientId, String zkRoot, String id);
 public SpoutConfig(BrokerHosts hosts, String topic, String zkRoot, String id);
 ```
 
--- a/external/storm-kafka/src/jvm/org/apache/storm/kafka/SpoutConfig.java
+++ b/external/storm-kafka/src/jvm/org/apache/storm/kafka/SpoutConfig.java
@@ -42,8 +42,20 @@ public class SpoutConfig extends KafkaConfig implements Serializable {
     public long retryDelayMaxMs = 60 * 1000;
     public int retryLimit = -1;
 
+    /**
+     * Create a SpoutConfig without setting client.id, which can make the source application ambiguous when tracing Kafka calls.
+     */
     public SpoutConfig(BrokerHosts hosts, String topic, String zkRoot, String id) {
         super(hosts, topic);
+        this.zkRoot = zkRoot;
+        this.id = id;
+    }
+
+    /**
+     * Create a SpoutConfig with a client.id value.
+     */
+    public SpoutConfig(BrokerHosts hosts, String topic, String clientId, String zkRoot, String id) {
+        super(hosts, topic, clientId);
         this.zkRoot = zkRoot;
         this.id = id;
     }


priyank5485:
@carl34  We already have setProp method in KafkaSpoutConfig in storm-kafka-client that can be used to set any producer property. I dont think this adds any value. Can you elaborate on why you need that method? As far as the storm-kafka module is concerned that property is already public. I dont see a reason for adding one more constructor argument.

carl34:
@priyank5485, setting client.id is useful for identifying requests Kafka protocol packets.  For instance, we have many topologies which run on a shared Storm cluster using storm-kafka and all of the fetch requests contain blank "Client ID:" values.  We would like to have a clear way for our developers to set client.id so that we can quickly determine the source of Kafka packets.  

For storm-kafka-client, I get your point that setProp can be used, but we would like to encourage our users to set this parameter so we would appreciate if it was clearly available with the builder pattern via intellisense.

Here is an example fetch request from storm-kafka without a client.id:
```
Kafka (Fetch Request)
    Length: 54
    API Key: Fetch (1)
    API Version: 0
    Correlation ID: 0
    String Length: 0
    Client ID:
    Replica ID: -1
    Max Wait Time: 10000
    Min Bytes: 0
    Array Count: 1
    Fetch Request Topic (1 partitions)
        String Length: 6
        Topic Name: canary
        Array Count: 1
        Fetch Request Partition (Partition-ID=2, Offset=162438425)
            Partition ID: 2
            Offset: 162438425
            Max Bytes: 1048576
```

priyank5485:
@carl34 https://github.com/apache/storm/blob/master/external/storm-kafka/src/jvm/org/apache/storm/kafka/KafkaConfig.java#L32 is already a public property. Are you saying passing it as a constructor argument is more clear than setting a public property? If yes, I dont agree with you on that. That extra constructor does not add any value. Plus, community is encouraged to use new spout. So unless there is a blocker or critical bug in the old spout i prefer not changing any code there.

For storm-kafka-client I plan to raise a PR to get rid of the builder pattern. Builder pattern relies on builder.build() and pass the returned object to the constructor. Unfortunately, that does not work well with flux. Hence setProp method, either on the builder (or may be in future directly on KafkaSpoutConfig) is encouraged. There are so many other properties on kafka consumer api. It does not make sense to expose each one of those as a setXXX java method. Without those setXXX methods it is a much cleaner api for spout and its config object. Intellisense should not be the top priority here. I would prefer a cleaner api.

carl34:
Thanks for the additional context.  Our primary objective is to find a way to update our legacy topologies which use storm-kafka, but it looks like the clientId member is final in 1.1.0, correct?

https://github.com/apache/storm/blob/a4afacd9617d620f50cf026fc599821f7ac25c79/external/storm-kafka/src/jvm/org/apache/storm/kafka/KafkaConfig.java#L32

We were initially just going to submit a PR for storm-kafka, but we're hoping to make it clear to new topology creators that client.id is available because we feel that it is valuable.  If the builder pattern is going to be removed, would it be a big deal to add this change in the mean-time?

priyank5485:
@carl34 Good point. I did not realize it was final until you brought it up. In that case i am okay with your change for storm-kafka module. For storm-kafka-client module i still prefer avoiding the method setCliendId. I would rather use setProp instead. Can you use setProp instead on the builder for now? And when i create a PR for refactoring storm-kafka-client you can comment about any issues that you might run in to because of the changes in the PR.

external/storm-kafka/pom.xml
```diff
@@ -57,7 +57,7 @@
                 <artifactId>maven-checkstyle-plugin</artifactId>
                 <!--Note - the version would be inherited-->
                 <configuration>
-                    <maxAllowedViolations>557</maxAllowedViolations>
+                    <maxAllowedViolations>558</maxAllowedViolations>
```
revans2: I would prefer to see the violations fixed instead of adding in new ones.  In this case I suspect that it is missing javadocs for the new SpoutConfig constructor.

revans2:
Also please upmerge there is a merge conflict.

carl34:
Added javadocs and rebased

carl34:
@priyank5485 sounds good, I went ahead and removed the storm-kafka-client change so that only storm-kafka is updated and we will plan to use setProp.

carl34:
I can try rebasing again for the CHANGELOG.md updates, please let me know if I should add a new entry or to that file or if there are any other actions for me to take.

PR #2128, opened 2017-05-21
update EsPercolateBolt.java

wynhbh:
update elasticsearch client api


wynhbh:
public Response performRequest(String method, String endpoint, Map<> params, Header... headers)
->
public Response performRequest(String method, String endpoint, Map<String, String> params, Header... headers)

external/storm-elasticsearch/src/main/java/org/apache/storm/elasticsearch/bolt/EsPercolateBolt.java
```diff
@@ -81,7 +81,7 @@ public void process(Tuple tuple) {
             Map<String, String> indexParams = new HashMap<>();
             indexParams.put(type, null);
             String percolateDoc = "{\"doc\": " + source + "}";
-            Response response = client.performRequest("get", getEndpoint(index, type, "_percolate"), new HashMap<>(), new StringEntity(percolateDoc));
+            Response response = client.performRequest("get", getEndpoint(index, type, "_percolate"), new HashMap<String, String>(), new StringEntity(percolateDoc));
```
HeartSaVioR: What's rationale of this patch? Did it occur compile issue?

PR #2134, opened 2017-05-23
STORM-2527 Initialize java.sql.DriverManager earlier to avoid deadlock

HeartSaVioR:
* also fixed some checkstyle violations

### DIFF
--- a/external/storm-jdbc/src/main/java/org/apache/storm/jdbc/bolt/AbstractJdbcBolt.java
+++ b/external/storm-jdbc/src/main/java/org/apache/storm/jdbc/bolt/AbstractJdbcBolt.java
@@ -15,25 +15,24 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 package org.apache.storm.jdbc.bolt;
 
-import org.apache.storm.Config;
-import org.apache.storm.task.OutputCollector;
-import org.apache.storm.task.TopologyContext;
-import org.apache.storm.topology.base.BaseRichBolt;
+import java.sql.DriverManager;
+import java.util.Map;
 import org.apache.commons.lang.Validate;
+import org.apache.storm.Config;
 import org.apache.storm.jdbc.common.ConnectionProvider;
 import org.apache.storm.jdbc.common.JdbcClient;
+import org.apache.storm.task.OutputCollector;
+import org.apache.storm.task.TopologyContext;
 import org.apache.storm.topology.base.BaseTickTupleAwareRichBolt;
-import org.apache.storm.tuple.Tuple;
-import org.apache.storm.utils.TupleUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.Map;
-
 public abstract class AbstractJdbcBolt extends BaseTickTupleAwareRichBolt {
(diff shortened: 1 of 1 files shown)


external/storm-jdbc/src/main/java/org/apache/storm/jdbc/bolt/AbstractJdbcBolt.java
```diff
@@ -42,26 +41,66 @@
...
+     * <p/>
+     * We should take this workaround since prepare() method is synchronized
+     * but an worker can initialize multiple HikariCPConnectionProviders and
+     * they would make race condition.
+     * <p/>
+     * We just need to ensure that DriverManager class is always initialized
+     * earlier than HikariConfig so below line should be called first
+     * than initializing HikariConfig
+     */
+    private static synchronized void avoidDeadlockFromDriverInitialization() {
```
satishd: I assume this handles when multiple jdbc bolt tasks are getting initialized/prepared with their respective connection providers/drivers.

HeartSaVioR: Yes right. I should modify the comment to be not specific to HikariCP. I'll fix it. Thanks.

HeartSaVioR: Updated the comment.

external/storm-jdbc/src/main/java/org/apache/storm/jdbc/bolt/AbstractJdbcBolt.java
```diff
@@ -42,26 +41,66 @@
...
+     * We should take this workaround since prepare() method is synchronized
+     * but an worker can initialize multiple AbstractJdbcBolt instances and
+     * they would make race condition.
+     * <p/>
+     * We just need to ensure that DriverManager class is always initialized
+     * earlier than provider so below line should be called first
+     * than initializing provider.
+     */
+    private static synchronized void avoidDeadlockFromDriverInitialization() {
+        DriverManager.getDrivers();
```
satishd: Can not we move this to static block with in this abstract class without using static synchronized method?

HeartSaVioR: @satishd This method should be run within worker, not topology submitter process. Hence I put this into start of prepare() and make it class-wide-synchronized.

satishd: Putting it in a static block 
- does not harm on storm submitter as it just loads `DriverManager` class
- it avoids going through synchronized in prepare calls of all JDBC tasks in a worker though it may not involve much cost as prepare is invoked only once for each task's life.

I am fine with either ways. :)

HeartSaVioR: Yeah you're right. I missed that class initialization should also happen on worker side regardless of serde of instance. Better to move this to the static block. Thanks!

HeartSaVioR:
@satishd Addressed review comment. Please review this again. Thanks in advance!

PR #2113, opened 2017-05-12
STORM-2497: Let Supervisor enforce memory and add in support for shared memory regions

revans2:
This is based off of #2112 so that the formatting is simpler.

PR #2135, opened 2017-05-23
Leave checkstyle violation to each module's target directory

HeartSaVioR:
Because of output log size policy for Travis CI, we can't print violation report to the console. Instead of that, we can just leave violation to the file. 

This would help to track and fix checkstyle violations.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1110,7 +1110,7 @@
                                 <encoding>UTF-8</encoding>
                                 <failOnViolation>true</failOnViolation>
                                 <logViolationsToConsole>false</logViolationsToConsole>
-                                <outputFile></outputFile>
+                                <outputFile>target/checkstyle-violation.xml</outputFile>
                                 <violationSeverity>warning</violationSeverity>
                             </configuration>
                             <goals>


HeartSaVioR:
Btw I don't like xml representation but when I set output format to plain, checkstyle plugin complains.
`Output format is 'plain', checkstyle:check requires format to be 'xml'`

srdo:
+1

The check goal doesn't currently support "plain" output, the documentation is wrong. http://svn.apache.org/viewvc/maven/plugins/tags/maven-checkstyle-plugin-2.17/src/main/java/org/apache/maven/plugin/checkstyle/CheckstyleViolationCheckMojo.java?view=markup#l556

It's because the plugin writes the checkstyle results to disk, then parses them back up again to figure out if there were any rule violations so it can fail the build. The parsing code only understands XML.

revans2:
+1

Although I think this is already happening.  As a part of https://github.com/apache/storm/pull/2113 I added in an xslt script to make reading the XML format better (like plain) and a python script to help search through it to help do things like look at violations only for files I touched with this patch.

PR #2136, opened 2017-05-23
[STORM-2206] - replacing visualization with viz.js

Crim:
Replaces the topology visualization with vis.js (http://visjs.org/docs/network/)


revans2:
It is looking good.  A few notes.

Ackers (__acker) are not spouts.  They are bolts (but it might be nice to distinguish between system bolts and other bolts)

Ackers are not showing any instances even when there are some.

The System Bolt (__system) is a bolt but not really, and probably does not need to show up here at all.  It is created as a way to collect metrics and may go away in the future.

Crim:
@revans2 I've updated it to denote components starting with __ as "system" rather than spout.  It would seem like the visualization end point is mis-reporting them as spouts when they should be something else.

Regarding the acker instances showing 0, that's also what the visualization api end point is returning. Unfortunately my clojure isn't so great and I wasn't able to correctly track down how these things get populated within the end point.

Crim:
@revans2 created STORM-2533 and STORM-2534 capturing the two issues w/ the visualization api end point that we uncovered.

PR #2142, opened 2017-05-27
MINOR: Fix pacemaker_state_factory.clj not compile problem

chengxinglin:
Fix pacemaker_state_factory.clj not compile problem

### DIFF
--- a/storm-core/src/clj/org/apache/storm/pacemaker/pacemaker_state_factory.clj
+++ b/storm-core/src/clj/org/apache/storm/pacemaker/pacemaker_state_factory.clj
@@ -106,16 +106,15 @@
 (defn get-pacemaker-write-client [conf servers client-pool]
   ;; Client should be created in case of an exception or first write call
   ;; Shutdown happens in the retry loop
-  (try 
-    (.waitUntilReady
-     (let [client (get @client-pool (first @servers))]
-       (if (nil? client)
-         (do
-           (swap! client-pool merge {(first @servers) (PacemakerClient. conf (first @servers))})
-           (get @client-pool (first @servers)))
-         client)))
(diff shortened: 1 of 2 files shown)


storm-core/src/clj/org/apache/storm/pacemaker/pacemaker_state_factory.clj
```diff
@@ -106,16 +106,15 @@
...
-      (throw e))))
+  (let [client (get @client-pool (first @servers))]
+    (try
+        (.waitUntilReady
+            (let [] (if (nil? client)
+                 (do (swap! client-pool merge {(first @servers) (PacemakerClient. conf (first @servers))})
+                     (get @client-pool (first @servers)))
+                      client)))
+       (catch Exception e (throw e)))
+    (get @client-pool (first @servers))))
```
revans2: waitUntilReady is not actually needed because send will call it internally.  We can just skip calling it here all together.  This then lets us have something like ...

```
(defn get-pacemaker-write-client [conf servers client-pool]
   ;; Client should be created in case of an exception or first write call
   ;; Shutdown happens in the retry loop
   (let [server (first @servers)]
          (or (get @client-pool server)
              (get (swap! client-pool merge {server (PacemakerClient. conf server)}) server))))
```

and we can then remove the update to PacemakerClient.java

HeartSaVioR:
@chengxinglin Do you have any updates?

HeartSaVioR:
@chengxinglin Kindly reminder.
@revans2 If @chengxinglin doesn't respond and if you think this is a blocker, could you craft pull request after waiting a bit?

revans2:
@HeartSaVioR yes.

PR #2143, opened 2017-05-29
STORM-2503: Restore comparator logic in `DefaultResourceAwareStrategy`.

adityasharad:
This PR ensures consistent comparator ordering in the above class, using descending order in both `sortNeighbors` and `sortComponents` and adding comments to indicate descending order.

As requested by @jerrypeng on https://github.com/apache/storm/pull/2100.
@revans2


jerrypeng:
@adityasharad Thanks for changing it back.  What about the unit test that was modified as well?

revans2:
@jerrypeng Like I said in the previous pull request.  The unit test that failed was not cause by this change.  It was cause by an actual fix to the comparison logic for `sortComponents`.  In my tests I saw the output of `sortComponents` having duplicate entries prior to that fix.

I am +1 for this change.

PR #2148, opened 2017-06-02
STORM-2543: Let Topology query if blobs have changed

govind-menon:
WIP

### DIFF
--- a/storm-client/src/jvm/org/apache/storm/daemon/worker/Worker.java
+++ b/storm-client/src/jvm/org/apache/storm/daemon/worker/Worker.java
@@ -29,6 +29,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Function;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 import javax.security.auth.Subject;
@@ -79,6 +81,7 @@ import uk.org.lidalia.sysoutslf4j.context.SysOutOverSLF4J;
 public class Worker implements Shutdownable, DaemonCommon {
 
     private static final Logger LOG = LoggerFactory.getLogger(Worker.class);
+    private static final Pattern BLOB_VERSION_EXTRACTION = Pattern.compile(".*\\.([0-9]+)$");
     private final Map<String, Object> conf;
     private final IContext context;
     private final String topologyId;
@@ -246,7 +249,20 @@ public class Worker implements Shutdownable, DaemonCommon {
                             }
                         }
                     });
-              
+
+                workerState.checkForUpdatedBlobsTimer.scheduleRecurring(0,
+                        (Integer) conf.getOrDefault(Config.WORKER_BLOB_UPDATE_POLL_INTERVAL_SECS, 10), new Runnable() {
+                            @Override public void run() {
+                                try {
+                                    LOG.debug("Checking if blobs have updated");
+                                    updateBlobUpdates();
+                                } catch (IOException e) {
+                                    // IOException from reading the version files to be ignored
+                                    LOG.error(e.getStackTrace().toString());
+                                }
+                            }
+                        });
+
                 // The jitter allows the clients to get the data at different times, and avoids thundering herd
                 if (!(Boolean) topologyConf.get(Config.TOPOLOGY_DISABLE_LOADAWARE_MESSAGING)) {
                     workerState.refreshLoadTimer.scheduleRecurringWithJitter(0, 1, 500, workerState::refreshLoad);
@@ -298,6 +314,35 @@ public class Worker implements Shutdownable, DaemonCommon {
             LOG.error("Worker failed to write heartbeats to ZK or Pacemaker...will retry", ex);
         }
     }
+
+    public Map<String, Long> getCurrentBlobVersions() throws IOException {
+        Map<String, Long> results = new HashMap<>();
+        Map<String, Map<String, Object>> blobstoreMap = (Map<String, Map<String, Object>>) workerState.getTopologyConf().get(Config.TOPOLOGY_BLOBSTORE_MAP);
+        if (blobstoreMap != null) {
+            String stormRoot = ConfigUtils.supervisorStormDistRoot(workerState.getTopologyConf(), workerState.getTopologyId());
+            for (Map.Entry<String, Map<String, Object>> entry : blobstoreMap.entrySet()) {
+                String localFileName = entry.getKey();
+                Map<String, Object> blobInfo = entry.getValue();
+                if (blobInfo != null && blobInfo.containsKey("localname")) {
+                    localFileName = (String) blobInfo.get("localname");
+                }
+
+                String blobWithVersion = new File(stormRoot, localFileName).getCanonicalFile().getName();
+                Matcher m = BLOB_VERSION_EXTRACTION.matcher(blobWithVersion);
+                if (m.matches()) {
+                    results.put(localFileName, Long.valueOf(m.group(1)));
+                }
+            }
+        }
+        return results;
+    }
+
+    public void updateBlobUpdates() throws IOException {
+        Map<String, Long> latestBlobVersions = getCurrentBlobVersions();
+        workerState.blobToLastKnownVersion.putAll(latestBlobVersions);
+        LOG.debug("Latest versions for blobs {}", latestBlobVersions);
+    }
+
 
     public void checkCredentialsChanged() {
         Credentials newCreds = workerState.stormClusterState.credentials(topologyId, null);
@@ -409,6 +454,7 @@ public class Worker implements Shutdownable, DaemonCommon {
             workerState.heartbeatTimer.close();
             workerState.refreshConnectionsTimer.close();
             workerState.refreshCredentialsTimer.close();
+            workerState.checkForUpdatedBlobsTimer.close();
             workerState.refreshBackpressureTimer.close();
             workerState.refreshActiveTimer.close();
             workerState.executorHeartbeatTimer.close();
@@ -437,6 +483,7 @@ public class Worker implements Shutdownable, DaemonCommon {
             && workerState.refreshConnectionsTimer.isTimerWaiting()
             && workerState.refreshLoadTimer.isTimerWaiting()
             && workerState.refreshCredentialsTimer.isTimerWaiting()
+            && workerState.checkForUpdatedBlobsTimer.isTimerWaiting()
             && workerState.refreshBackpressureTimer.isTimerWaiting()
             && workerState.refreshActiveTimer.isTimerWaiting()
             && workerState.executorHeartbeatTimer.isTimerWaiting()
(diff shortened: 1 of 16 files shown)


storm-client/src/jvm/org/apache/storm/daemon/worker/Worker.java
```diff
@@ -246,7 +248,19 @@ public void start() throws Exception {
                             }
                         }
                     });
-              
+
+                workerState.checkForUpdatedBlobsTimer.scheduleRecurring(0,
+                        (Integer) conf.getOrDefault(Config.WORKER_BLOB_UPDATE_POLL_INTERVAL_SECS, 10), new Runnable() {
+                            @Override public void run() {
+                                try {
+                                    LOG.info("Checking if blobs have updated");
```
revans2: This should probably be a debug log.

storm-client/src/jvm/org/apache/storm/daemon/worker/Worker.java
```diff
@@ -246,7 +248,19 @@ public void start() throws Exception {
...
+                                    updateBlobUpdates();
+                                } catch (IOException e) {
+                                    e.printStackTrace();
```
revans2: Can we log this instead of printing it to stderr?  Also it might be nice to have a comment explaining why you are ignoring the exception.

storm-client/src/jvm/org/apache/storm/daemon/worker/Worker.java
```diff
@@ -299,6 +313,33 @@ public void doExecutorHeartbeats() {
         }
     }
 
+    public Map<String, Long> getCurrentBlobVersions() throws IOException {
+        Map<String, Long> results = new HashMap();
```
revans2: The new HashMap should have the generics on it `new HashMap<>();`

storm-client/src/jvm/org/apache/storm/daemon/worker/Worker.java
```diff
@@ -246,7 +248,19 @@ public void start() throws Exception {
                             }
                         }
                     });
-              
+
+                workerState.checkForUpdatedBlobsTimer.scheduleRecurring(0,
```
revans2: Is there another timer we could use here?

storm-client/src/jvm/org/apache/storm/task/TopologyContext.java
```diff
@@ -57,13 +57,14 @@
     private Map<String, Object> _executorData;
     private Map<Integer,Map<Integer, Map<String, IMetric>>> _registeredMetrics;
     private AtomicBoolean _openOrPrepareWasCalled;
-
+    private Map<String, Long> _blobToLastKnownVersion;
```
revans2: Can we remove the `_` at the beginning.  I know it matches the rest of the file but we are moving away from this style.

storm-client/src/jvm/org/apache/storm/utils/ConfigUtils.java
```diff
@@ -27,12 +28,11 @@
 import java.io.File;
 import java.io.IOException;
 import java.net.URLEncoder;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.Random;
+import java.util.*;
```
revans2: * in imports are generally discouraged.

storm-client/src/jvm/org/apache/storm/utils/ConfigUtils.java
```diff
@@ -27,12 +28,11 @@
...
 import java.util.concurrent.Callable;
+import java.util.stream.Collectors;
 
-public class ConfigUtils {
+public class  ConfigUtils {
```
revans2: can we remove the extra ' ' added?

storm-client/src/jvm/org/apache/storm/Constants.java
```diff
@@ -56,5 +56,7 @@
     public static final String STORM_ACTIVE_ATOM = "storm-active-atom";
     public static final String COMPONENT_TO_DEBUG_ATOM = "storm-component->debug-atom";
     public static final Object LOAD_MAPPING = "load-mapping";
+    public static final String USERCACHE = "usercache";
+    public static final String DEFAULT_BLOB_VERSION_SUFFIX = ".version";
```
revans2: I'm not sure these needed to be moved.  They look like they were moved for one possible way of doing this, and then not reverted when they were not used by the client.

storm-client/src/jvm/org/apache/storm/daemon/worker/Worker.java
```diff
@@ -299,6 +313,33 @@ public void doExecutorHeartbeats() {
...
+                File versionFile = new File(blobPath.replaceAll(".[0-9]+$", DEFAULT_BLOB_VERSION_SUFFIX));
+                results.put(localFileName, Utils.getVersionFromBlobVersionFile(versionFile));
+            }
+        }
+        return results;
+    }
+
+    public void updateBlobUpdates() throws IOException {
+        Map<String, Long> latestBlobVersions = getCurrentBlobVersions();
+        workerState.blobToLastKnownVersion.putAll(latestBlobVersions);
```
revans2: We need some kind of synchronization here.  Perhaps we can just make `blobToLastKnownVersion` a ConcurrentMap.  That should be enough.

storm-client/src/jvm/org/apache/storm/daemon/worker/Worker.java
```diff
@@ -299,6 +313,33 @@ public void doExecutorHeartbeats() {
...
+        LOG.debug("Latest versions for blobs " + latestBlobVersions.toString());
```
revans2:
```
LOG.debug("Latest versions for blobs {}", latestBlobVersions);
```

storm-client/src/jvm/org/apache/storm/task/TopologyContext.java
```diff
@@ -57,13 +57,14 @@
...
 
     public TopologyContext(StormTopology topology,
                            Map<String, Object> topoConf,
                            Map<Integer, String> taskToComponent,
                            Map<String, List<Integer>> componentToSortedTasks,
                            Map<String, Map<String, Fields>> componentToStreamToFields,
+                           Map<String, Long> blobToLastKnownVersion,
```
revans2: So this is a little odd in that we send the versions to all of the children by passing in a variable that we expect to change behind the scenes.  I think it would be good to call that out in some comments both in the constructor here, and on field too so it is obvious here in the code what is happening.

storm-server/src/main/java/org/apache/storm/localizer/Localizer.java
```diff
@@ -41,10 +42,7 @@
 import java.io.PrintWriter;
 import java.nio.file.Files;
 import java.nio.file.Paths;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
revans2: Again `*` is discouraged here.

storm-client/src/jvm/org/apache/storm/daemon/worker/Worker.java
```diff
@@ -299,6 +313,33 @@ public void doExecutorHeartbeats() {
...
+            for (Map.Entry<String, Map<String, Object>> entry : blobstoreMap.entrySet()) {
+                String keyName = entry.getKey();
+                String localFileName = keyName;
+                Map<String, Object> blobInfo = blobstoreMap.get(keyName);
+                if (blobInfo != null && blobInfo.containsKey("localname")) {
+                    localFileName = (String) blobInfo.get("localname");
+                }
+                String blobPath = new File(stormRoot, localFileName).getCanonicalPath();
+                File versionFile = new File(blobPath.replaceAll(".[0-9]+$", DEFAULT_BLOB_VERSION_SUFFIX));
+                results.put(localFileName, Utils.getVersionFromBlobVersionFile(versionFile));
```
revans2: Can we please do this slightly differently?  The way it currently works `$whatever.version` can be updated before we officially commit the changes (I know that is not good and I am trying to change it), but for now we really should do it more like.

````
private static final Pattern BLOB_VERSION_EXTRACTION = Pattern.compile(".*\\.([0-9]+)$");
//...
String blobWithVersion = new File("./", localFileName).getCanonicalFile().getName();
Matcher m = BLOB_VERSION_EXTRACTION.matcher(blobWithVersion);
if (m.matches()) {
    results.put(localFileName, Long.valueOf(m.group(1)));
}
````

File: storm-client/src/jvm/org/apache/storm/daemon/worker/Worker.java [hunk 14]
```diff
@@ -299,6 +313,33 @@ public void doExecutorHeartbeats() {
...
 
+    public Map<String, Long> getCurrentBlobVersions() throws IOException {
+        Map<String, Long> results = new HashMap();
+        Map<String, Map<String, Object>> blobstoreMap = (Map<String, Map<String, Object>>) workerState.getTopologyConf().get(Config.TOPOLOGY_BLOBSTORE_MAP);
+        if (blobstoreMap != null) {
+            String stormRoot = ConfigUtils.supervisorStormDistRoot(workerState.getTopologyConf(), workerState.getTopologyId());
+            for (Map.Entry<String, Map<String, Object>> entry : blobstoreMap.entrySet()) {
+                String keyName = entry.getKey();
+                String localFileName = keyName;
+                Map<String, Object> blobInfo = blobstoreMap.get(keyName);
```
revans2: Lets use `entry.getValue()` instead.  It is slightly more efficient.

storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
```diff
@@ -2369,7 +2369,7 @@ public void launchServer() throws Exception {
             for (ClusterMetricsConsumerExecutor exec: clusterConsumerExceutors) {
                 exec.prepare();
             }
-            
+            assertIsLeader();
             if (isLeader()) {
```
revans2: This is a little odd.  Why do we need to be sure that we are the leader here?  Especially with code right below it checking if we are the leader or not.

storm-client/src/jvm/org/apache/storm/daemon/worker/Worker.java
govind-menon: I wanted the period for this check to be it's own value and not tied to other timers

revans2: OK.  We have so many now it might be good to switch it over to a pool in the future, but nothing you need to worry about.

storm-server/src/main/java/org/apache/storm/localizer/Localizer.java
```diff
@@ -226,9 +218,8 @@ public boolean accept(File dir, String name) {
   protected void reconstructLocalizedResources() {
     try {
       LOG.info("Reconstruct localized resource: " + getUserCacheDir().getPath());
-      File[] users = readDirContents(getUserCacheDir().getPath());
-
-      if (users != null) {
+      Collection<File> users = ConfigUtils.readDirFiles(getUserCacheDir().getPath());
+      if (users == null || users.isEmpty()) {
```
revans2: This can be reverted.  The only need was to check for null so we don't NPE.  This inverted the check.

govind-menon:
Tagging @satishd  for another review.

PR #2150, opened 2017-06-05
STORM-2541: Fix storm-kafka-client manual subscription not being able…

srdo:
… to start consuming

I also refactored the manual subscriptions a bit to deduplicate the code. I think there's no real issue with backporting this to 1.x, because clearly no one has been using these subscriptions so the API breakage shouldn't bother anyone.

With regard to testing the manual subscription, I'd like to do that in https://issues.apache.org/jira/browse/STORM-2542 instead, since it'll make the current tests use the manual subscription.


external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/ManualPartitionSubscription.java
```diff
@@ -58,21 +51,21 @@ public ManualPartitionNamedSubscription(ManualPartitioner parter, String ... top
...
         Set<TopicPartition> newAssignment = new HashSet<>(partitioner.partition(allPartitions, context));
         if (!newAssignment.equals(currentAssignment)) {
+            consumer.assign(newAssignment);
             if (currentAssignment != null) {
                 listener.onPartitionsRevoked(currentAssignment);
-                listener.onPartitionsAssigned(newAssignment);
             }
+            listener.onPartitionsAssigned(newAssignment);
             currentAssignment = newAssignment;
-            consumer.assign(currentAssignment);
```
hmcl: @srdo lines 62 and 63 should be switched. We want to call the listener only once all is set.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/TopicFilter.java
```diff
@@ -0,0 +1,38 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.storm.kafka.spout;
+
+import java.util.List;
+import org.apache.kafka.clients.consumer.KafkaConsumer;
+import org.apache.kafka.common.TopicPartition;
+
+public interface TopicFilter {
```
hmcl: must implement Serializable. Otherwise code didn't work for me. The trident [changes](https://github.com/hmcl/storm-apache/blob/35f2c0c71a28a065e64f523b861acd883b2f2101/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/TopicFilter.java#L24) are already do this.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/PatternTopicFilter.java
```diff
@@ -0,0 +1,59 @@
...
+                    allPartitions.add(new TopicPartition(partitionInfo.topic(), partitionInfo.partition()));
+                }
+            }
+        }
+        return allPartitions;
+    }
+
+    @Override
+    public String getTopicsString() {
+        return pattern.pattern();
```
hmcl: This implementation should be for a method called getTopicsPattern. Please make a diff with my [slight modification](https://github.com/hmcl/storm-apache/blob/35f2c0c71a28a065e64f523b861acd883b2f2101/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/PatternTopicFilter.java) of your changes. I suggest that you incorporate that diff in this patch.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/ManualPartitionSubscription.java
srdo: Will fix

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/TopicFilter.java
srdo: Will fix

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/PatternTopicFilter.java
srdo: I'm wondering if this works? storm-kafka-monitor seems to be getting the subscribed topics out of the getComponentConfiguration method, but is that called more than once? If it isn't being called occasionally then the list this method returns is going to be empty for pattern subscriptions.

harshach: @priyank5485 can you answer the storm-kafka-monitor question here.

hmcl: @srdo I believe that the approach I am suggesting in the Trident PR will solve what both you and I have in mind. Please take a look at the following:

https://github.com/hmcl/storm-apache/blob/ca420a4a960c9667262833add74e231fe8d5518b/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/PatternTopicFilter.java#L60-L67

https://github.com/hmcl/storm-apache/blob/ca420a4a960c9667262833add74e231fe8d5518b/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/PatternTopicFilter.java#L35

https://github.com/hmcl/storm-apache/blob/ca420a4a960c9667262833add74e231fe8d5518b/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/PatternTopicFilter.java#L53

srdo: @hmcl I went and tried out storm-kafka-monitor from the https://github.com/apache/storm/pull/2174 branch, and ran into multiple issues.

* The PatternTopicFilter code throws an NPE because the spout calls getTopicString before the Set is initialized (the method is used for logging). This prevents the topology from being submitted.
* It looks like the Kafka consumer code is not being correctly shaded into the storm-kafka-monitor jar, so the bash script didn't work for me. I'm not sure if this was a quirk of my build, since the consumer code is present in the 1.1.0 release storm-kafka-monitor jar.
* After fixing the two other issues, I tried submitting a topology consuming the "test*" pattern from a Kafka instance where topics "test", "test-1" and "test-2" were present. The Storm UI offset lag request returns the following, no matter how long I wait after the topology is running

```
{"kafka-spout":{"spoutId":"kafka-spout","errorInfo":"Unable to get offset lags for kafka. Reason: org.apache.kafka.shaded.common.errors.InvalidTopicException: Topic '' is invalid\n","spoutType":"KAFKA"}}
```

That code (https://github.com/apache/storm/blob/64e29f365c9b5d3e15b33f33ab64e200345333e4/storm-core/src/jvm/org/apache/storm/utils/TopologySpoutLag.java#L91) gets the topic list it should request offset lag for via the spout's getComponentOffsets method, as far as I can tell through this call https://github.com/apache/storm/blob/9e31509d47c4e91c1009f55c7ccf321d7d7e63aa/storm-client/src/jvm/org/apache/storm/topology/TopologyBuilder.java#L541. I don't think we can make the current storm-kafka-monitor implementation work for subscriptions that are not known at topology submission time, unless there is some way to update the component configuration when the subscription changes.

I'd also add that I have a few other issues with storm-kafka-monitor:
* The jar is installed along with the cluster, and depends on the Kafka version specified in Storm's root POM. Kafka guarantees backwards compatible client-server communication for one release only, so there's a potential coupling between Storm cluster version and Kafka version. If users want to update the Kafka version in storm-kafka-monitor, they have to rebuild that module and replace the jar in their Storm install.
* The UI integration uses the storm-kafka-monitor Bash script to start the monitoring code, in order to avoid a dependency between storm-core and storm-kafka-monitor. This prevents the UI integration from working on Windows. We could supply a Windows script as well, but then we'd need to keep the two in sync.

I think it makes sense to make the change you are suggesting simply because it makes the spout log a little nicer, but it still leaves the issue with storm-kafka-monitor not working for pattern subscriptions. It would be good if we could figure out a solution to the other issues as well. Maybe we could track offset lag through the metrics system instead, and just make the spout update it?

I'll create a separate issue for discussing storm-kafka-monitor.

srdo:
Created the issue for storm-kafka-monitor here https://issues.apache.org/jira/browse/STORM-2600

priyank5485:
@srdo @harshach Some comments related to storm-kafka-monitor 

[~Srdo] Regarding first point, because the lag request is a http pull request from UI, as long as kafkaSpoutConfig.getSubscription().getTopicsString(); returns the correct value it will work since the open method would have been called eventually. The only change we would need is that when the rebalance listener is invoked we would need to keep track of the topics currently subscribed. For e.g. PatternSubscription can have an instance variable called Set<String> topics that would be correctly updated anytime onPartitionsRevoked or onPartitionsAssigned is called. We can use that instance variable to return the value when getTopicsString is called on that object. Does that work?

Regarding point 2, we could move the kafka client version to storm-kafka-monitor module. Can you elaborate a bit more on your concern? Is it that if kafka cluster is upgraded, storm-kafka-monitor wont work? In that case storm-kafka-client module will also have to be updated and topology jar rebuilt. Correct? I think in general, we have a certain compatibility restriction in that a given storm release works or has been tested with a certain version of kafka clients. Correct me if i am wrong.

Regarding 3, the main reason for creating a separate module and calling bash script from storm UI is so that storm-core does not have a direct kafka dependency since that made sense. For windows, cygwin can be a workaround. Plus, i dont know how many windows users we have. We can start a thread to see if there is really a user base for which we need to support a use case.

I dont know the details about how metrics would work. We could have opinions from other people. If that is indeed the right way to go about it then I am all for changing it.

srdo:
@priyank5485 Thanks for responding, I replied on the linked issue

srdo:
The issues raised in https://issues.apache.org/jira/browse/STORM-2600 aren't really related to this PR, since the pattern based subscriptions were already unsupported by storm-kafka-monitor prior to this change. I think we should look at those issues separately.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/PatternTopicFilter.java
hmcl: @srdo I think it's good. I still think it's good practice for this class' internal API to return the pattern that it is using to compute the topics string. I am planning on leaving it on my trident [change](https://github.com/hmcl/storm-apache/blob/ca420a4a960c9667262833add74e231fe8d5518b/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/PatternTopicFilter.java#L65-L67).

hmcl:
+1

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/PatternTopicFilter.java
srdo: @hmcl Sure, I don't see a reason not to have a Pattern getter

hmcl:
@srdo I am +1 on this PR. Let's just add the getter method as we agreed on [this comment](https://github.com/apache/storm/pull/2150#discussion_r124855414), and have this PR consist of two commits only. The commit in [STORM-2548 PR](https://github.com/apache/storm/pull/2155) and its own commit.

hmcl:

srdo:
@hmcl Sorry, I accidentally overwrote this PR with the STORM-2548 content. This shouldn't contain the STORM-2548 changes. I added the getter and fixed the content error.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/TopicFilter.java
```diff
@@ -0,0 +1,38 @@
...
+    
+    /**
+     * Get the Kafka TopicPartitions passed by this filter. 
+     * @param consumer The Kafka consumer to use to read the list of existing partitions
+     * @return The Kafka partitions passed by this filter.
+     */
+    List<TopicPartition> getFilteredTopicPartitions(KafkaConsumer<?, ?> consumer);
+    
+    /**
+     * @return A human-readable string representing the subscribed topics.
```
hmcl: NIT: representing the topics that pass the filter

hmcl:
+1. @harshach @priyank5485 can you please take one final look. If you don't have any objection, I suggest that we merge this patch in the next day or so.

PR #2151, opened 2017-06-05
STORM-2542: Remove KafkaConsumer.subscribe API option, make KafkaConsumer.assign the default

srdo:
This builds on https://github.com/apache/storm/pull/2150, which is the first commit in this. The third commit is purely class moves because the kafka.spout package was getting a bit unwieldy.

This should only go on 2.x if it goes in at all.

Please see https://issues.apache.org/jira/browse/STORM-2542 for the justification for why I believe we should make this change.


srdo:
If this goes in I'll follow up with a PR against 1.x to deprecate the classes using `KafkaConsumer.subscribe`

hmcl:
@srdo was this agreed upon ?

srdo:
@hmcl No. I got no response on the mailing list, so now I'm trying here. I figure if anyone objects they'll be able to do so on this PR.

hmcl:
@srdo I am +1 on this PR. Let's just clean the commits that that make it such that we this PR consist of three commits only. The first commit should be [STORM-2548 PR](https://github.com/apache/storm/pull/2155), the 2nd commit [STORM-2541 PR](https://github.com/apache/storm/pull/2150), plus its own commit moving classes into the appropriate packages.

hmcl:

harshach:
+1. Thanks @srdo this looks great.

srdo:
Thanks for the reviews.

harshach:
@srdo are we not planning on pushing this into 1.x-branch?

srdo:
@harshach Yes, sorry I thought I'd make a separate JIRA for that so we could list the deprecation in 1.2.0 and the removal of these classes in 2.0.0. The PR is here https://github.com/apache/storm/pull/2224

PR #2152, opened 2017-06-07
STORM-2639: Kafka Spout incorrectly computes numCommittedOffsets due to voids in the topic (topic compaction)

askprasanna:
The logic currently does a simple subtraction between the first and last offset in the committed range. If the commit spans offset voids then this would be incorrect. This incorrect number would always be on the higher side and impact the logic around maximum uncommitted offsets allowable in the spout.

### DIFF
--- a/external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/KafkaSpoutCommitTest.java
+++ b/external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/KafkaSpoutCommitTest.java
@@ -1,0 +1,146 @@
+/*
+ * Copyright 2017 The Apache Software Foundation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.storm.kafka.spout;
+
+import static org.apache.storm.kafka.spout.builders.SingleTopicKafkaSpoutConfiguration.getKafkaSpoutConfigBuilder;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+import static org.mockito.Matchers.*;
+import static org.mockito.Mockito.*;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import org.apache.kafka.clients.consumer.*;
+import org.apache.kafka.common.TopicPartition;
+import org.apache.storm.kafka.spout.builders.SingleTopicKafkaSpoutConfiguration;
+import org.apache.storm.kafka.spout.internal.KafkaConsumerFactory;
+import org.apache.storm.spout.SpoutOutputCollector;
+import org.apache.storm.task.TopologyContext;
+import org.apache.storm.utils.Time;
+import org.apache.storm.utils.Time.SimulatedTime;
+import org.junit.Test;
+import org.mockito.ArgumentCaptor;
+import org.mockito.Captor;
+import org.mockito.InOrder;
+import org.mockito.MockitoAnnotations;
+
+public class KafkaSpoutCommitTest {
+
+    private final long offsetCommitPeriodMs = 2_000;
+    private final TopologyContext contextMock = mock(TopologyContext.class);
+    private final SpoutOutputCollector collectorMock = mock(SpoutOutputCollector.class);
+    private final Map<String, Object> conf = new HashMap<>();
+    private final TopicPartition partition = new TopicPartition(SingleTopicKafkaSpoutConfiguration.TOPIC, 1);
+    private KafkaConsumer<String, String> consumerMock;
+    private KafkaSpout<String, String> spout;
+    private KafkaSpoutConfig spoutConfig;
+
+    @Captor
+    private ArgumentCaptor<Map<TopicPartition, OffsetAndMetadata>> commitCapture;
+
+    private void setupSpout(Set<TopicPartition> assignedPartitions) {
+        MockitoAnnotations.initMocks(this);
+        spoutConfig = getKafkaSpoutConfigBuilder(-1)
+                .setOffsetCommitPeriodMs(offsetCommitPeriodMs)
+                .build();
+
+        consumerMock = mock(KafkaConsumer.class);
+        KafkaConsumerFactory<String, String> consumerFactory = (kafkaSpoutConfig) -> consumerMock;
+
+        //Set up a spout listening to 1 topic partition
+        spout = new KafkaSpout<>(spoutConfig, consumerFactory);
+
+        spout.open(conf, contextMock, collectorMock);
+        spout.activate();
+
+        ArgumentCaptor<ConsumerRebalanceListener> rebalanceListenerCapture = ArgumentCaptor.forClass(ConsumerRebalanceListener.class);
+        verify(consumerMock).subscribe(anyCollection(), rebalanceListenerCapture.capture());
+
+        //Assign partitions to the spout
+        ConsumerRebalanceListener consumerRebalanceListener = rebalanceListenerCapture.getValue();
+        consumerRebalanceListener.onPartitionsAssigned(assignedPartitions);
+    }
+
+    @Test
+    public void testCommitSuccessWithOffsetVoids() {
+        //Verify that the commit logic can handle offset voids
+        try (SimulatedTime simulatedTime = new SimulatedTime()) {
+            setupSpout(Collections.singleton(partition));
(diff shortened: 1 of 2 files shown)


hmcl:
@askprasanna what do you mean by "topic void" ?

askprasanna:
I am referring to valid gaps in the offsets seen by the kafka consumer under certain conditions.

I am using the same terminology as when we created PR https://github.com/apache/storm/pull/2104 This is a follow-up to that PR. Kindly refer to https://issues.apache.org/jira/browse/STORM-2505 for details.

askprasanna:
Updated to title to mention offset void. Topic void is a typo.

srdo:
This seems reasonable to me. Would be good to have a test too, you could probably fake log compaction by providing non-sequential records to a mocked KafkaConsumer, similar to how KafkaSpoutEmitTest does it. Either way +1

askprasanna:
Missed adding the test. Will do so.

askprasanna:
Added the unit test.

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/KafkaSpoutCommitTest.java
```diff
@@ -0,0 +1,145 @@
...
+            //verify that Offset 4 was last committed offset and 5 offsets were committed
+            //the offset void should be bridged in the next commit
+            assertTrue(committedOffsets.getValue().containsKey(partition));
+            assertEquals(((OffsetAndMetadata) (committedOffsets.getValue().get(partition))).offset(), 4);
+            assertEquals(2, spout.getNumUncommittedOffsets());
+
+            //Trigger second kafka consumer commit
+            reset(consumerMock);
+            when(consumerMock.poll(anyLong()))
+                    .thenReturn(new ConsumerRecords(Collections.EMPTY_MAP));
```
srdo: Nit: You can put generics on ConsumerRecords and use Collections.emptyMap() instead to avoid rawtypes warnings

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/KafkaSpoutCommitTest.java
```diff
@@ -0,0 +1,145 @@
...
+            Time.advanceTime(KafkaSpout.TIMER_DELAY_MS + offsetCommitPeriodMs);
+            spout.nextTuple();
+
+            committedOffsets=ArgumentCaptor.forClass(Map.class);
```
srdo: Nit: You can use the Mock annotation to put generic parameters on this. See https://static.javadoc.io/org.mockito/mockito-core/2.8.9/org/mockito/Mockito.html#mock_annotation

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/KafkaSpoutCommitTest.java
```diff
@@ -0,0 +1,145 @@
...
+            inOrder = inOrder(consumerMock);
+            inOrder.verify(consumerMock).commitSync(committedOffsets.capture());
+            inOrder.verify(consumerMock).poll(anyLong());
+
+            //verify that Offset 9 was last committed offset and no offsets are uncommitted
+            assertTrue(committedOffsets.getValue().containsKey(partition));
+            assertEquals(9, ((OffsetAndMetadata) (committedOffsets.getValue().get(partition))).offset());
```
srdo: It would also allow you to avoid this cast

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -104,6 +104,10 @@ public KafkaSpout(KafkaSpoutConfig<K, V> kafkaSpoutConfig) {
         this.kafkaSpoutConfig = kafkaSpoutConfig;
     }
 
+    long getNumUncommittedOffsets() {
```
srdo: I'm wondering if we can avoid inspecting this directly in the test by unit testing OffsetManager instead of indirectly testing it via the KafkaSpout? Testing that the spout will commit past gaps in the tuple stream is a good idea, but the check of this field is actually testing that OffsetManager.commit isn't buggy, which I feel belongs in a test for that class.

srdo:
I'm not sure if we should have a new JIRA issue for this btw. The issue this links to is already marked fixed.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
askprasanna: Yes, the main purpose of the test is verifying commit behavior over offset gaps. We can remove this getter and the asserts on it if required.

askprasanna:
I picked the existing JIRA since it is a followup to the same issue and addresses something that was missed in the earlier commits. Can we not reopen and re-resolved?

srdo:
It's probably fine since there hasn't been a release containing STORM-2505. It would just be confusing if the same issue got fixed multiple times in https://github.com/apache/storm/blob/master/CHANGELOG.md

askprasanna:
Have made the necessary changes. LMK if anything else is needed for the PR to be merged.

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/KafkaSpoutCommitTest.java
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    @Test
+    public void testCommitSuccessWithOffsetVoids() {
+        //Verify that the commit logic can handle offset voids
+        try (SimulatedTime simulatedTime = new SimulatedTime()) {
+            setupSpout(Collections.singleton(partition));
+            Map<TopicPartition, List<ConsumerRecord<String, String>>> records = new HashMap<>();
+            List<ConsumerRecord<String, String>> recordsForPartition = new ArrayList<>();
+            // Offsets emitted are 0,1,2,3,4,...,8,9
```
srdo: Nit: This comment is a little confusing, the ... would usually imply that 5,6,7 are included.

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/KafkaSpoutCommitTest.java
```diff
@@ -0,0 +1,146 @@
...
+
+            // Advance time and then trigger first call to kafka consumer commit; the commit will progress till offset 4
+            Time.advanceTime(KafkaSpout.TIMER_DELAY_MS + offsetCommitPeriodMs);
+            when(consumerMock.poll(anyLong()))
+                    .thenReturn(new ConsumerRecords<>(Collections.emptyMap()));
+            spout.nextTuple();
+
+            InOrder inOrder = inOrder(consumerMock);
+            inOrder.verify(consumerMock).commitSync(commitCapture.capture());
+            inOrder.verify(consumerMock).poll(anyLong());
```
srdo: Nit: I don't think it matters to the test that poll is called. Consider removing this and the InOrder if you agree

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/KafkaSpoutCommitTest.java
```diff
@@ -0,0 +1,146 @@
...
+            //Trigger second kafka consumer commit
+            reset(consumerMock);
+            when(consumerMock.poll(anyLong()))
+                    .thenReturn(new ConsumerRecords<>(Collections.emptyMap()));
+            Time.advanceTime(KafkaSpout.TIMER_DELAY_MS + offsetCommitPeriodMs);
+            spout.nextTuple();
+
+            inOrder = inOrder(consumerMock);
+            inOrder.verify(consumerMock).commitSync(commitCapture.capture());
+            inOrder.verify(consumerMock).poll(anyLong());
```
srdo: Same as above

srdo:
I forgot to mention it, but this seems applicable to the 1.x branch as well. We can't cherry pick this back since it uses a few lambdas, so would you be willing to make a PR for a backport? This would also be the case for https://github.com/apache/storm/pull/2153.

askprasanna:
Will be happy to do that. Haven't checked fully and being a bit lazy here - we are using lambdas only in the unit tests, right?

srdo:
Yes, it looks like there are only lambdas in the test code for these two PRs.

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/KafkaSpoutCommitTest.java
askprasanna: will tweak the comment.

askprasanna:
Squashed the commits. Also opened https://github.com/apache/storm/pull/2217 for 1.x branch

HeartSaVioR:
@askprasanna 
Maybe we would need to associate this to a new issue to track your effort (given that assignee for STORM-2505 is not you and it can't be overwritten), or remove the issue at all. Except test code I think it's fairly small change which doesn't strictly require filing an issue.

hmcl:
@askprasanna let's just create a new JIRA and associate it with the STORM-2505 JIRA - it's a quick thing and will take care of this clean. 

Another minor detail we have a script to find orphan pull requests, malformed pull requests, etc. The pull request description `[STORM-2505] OffsetManager...` does not conform with the name convention. Can you please rename it to follow the pattern `STORM-2505: OffsetManager...`. Thanks.

askprasanna:
Created a new JIRA and updated PR title as recommended.

srdo:
+1 again

PR #2155, opened 2017-06-10
STORM-2548: Simplify KafkaSpoutConfig to avoid duplicating KafkaConsu…

srdo:
…mer configuration parameters

This is put up for discussion, since a few people on the mailing list expressed a desire for a simpler KafkaSpoutConfig.

@harshach and @priyank5485 I think this is something like what you were asking for?


external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutConfig.java
```diff
@@ -79,16 +78,46 @@
...
+     * @param topics The topic pattern to subscribe to
+     * @return The new builder
+     */
     public static Builder<String, String> builder(String bootstrapServers, Pattern topics) {
-        return new Builder<>(bootstrapServers, StringDeserializer.class, StringDeserializer.class, topics);
+        return setDefaultStringDeserializers(new Builder<>(bootstrapServers, topics));
+    }
+    
+    private static Builder<String, String> setDefaultStringDeserializers(Builder<String, String> builder) {
+        builder.setProp(ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG, StringDeserializer.class);
```
priyank5485: I think ByteArrayDeserializer makes more sense. Not a strong preference though.

priyank5485:
@srdo Overall, this looks good. But I also think we should get rid of the Builder class totally. Reason is that it does not work well with flux yaml since you create the object using builder.build(). Let me know your thoughts on that.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutConfig.java
srdo: This is just here to support the 3 convenience builder() functions above. I'll rename it so it doesn't imply that String is the default type.

srdo:
(copying from the mailing list so it is easier to track the discussion if someone reads this later)

With regard to getting rid of the builder pattern, I think it is a pretty nice pattern for Java. It looks to me like it should be possible to declare and configure the builder with "component:", and then pass it to the KafkaSpoutConfig constructor with "ref:" after (which lets you avoid calling build()). Doesn't this work?

priyank5485:
+1(NB) given that people prefer keeping the Builder class around.

harshach:
+1

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutConfig.java
```diff
@@ -116,217 +141,57 @@
...
+         * Set a Kafka consumer property config. 
          */
         public Builder<K,V> setProp(String key, Object value) {
             kafkaProps.put(key, value);
             return this;
         }
         
         /**
-         * Set multiple Kafka property configs.
+         * Set multiple Kafka consumer property configs.
```
hmcl: Set multiple Kafka consumer properties

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutConfig.java
```diff
@@ -116,217 +141,57 @@
...
          */
         public Builder<K,V> setProp(Map<String, Object> props) {
             kafkaProps.putAll(props);
             return this;
         }
         
         /**
-         * Set multiple Kafka property configs.
+         * Set multiple Kafka consumer property configs.
```
hmcl: Set multiple Kafka consumer properties

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutConfig.java
```diff
@@ -116,217 +141,57 @@
...
          */
         public Builder<K,V> setProp(Properties props) {
             for (String name: props.stringPropertyNames()) {
```
hmcl: This code has one bug as it only picks properties for which both key and value are both strings. It should also pick key/value pairs that are String/Object. I suggest that we refactor this code to be as such:

```java
    public Builder<K,V> setProp(Properties props) {
            props.forEach((key, value) -> {
                if (key instanceof String) {
                    kafkaProps.put((String) key, value);
                }
            });
            return this;
        }
```

srdo: Nice catch. How about we use putAll to put the properties into kafkaProps instead? I'm not sure that I like the method filtering the input properties, I'd rather throw an error if the key is not a String. At the same time I don't think it should be KafkaSpoutConfig's job to validate KafkaConsumer parameters. The KafkaConsumer will handle throwing the error if the keys are not Strings https://github.com/apache/kafka/blob/efb060c57f05d1d586bb14c016b0187c60f8e994/clients/src/main/java/org/apache/kafka/common/config/AbstractConfig.java#L60

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutConfig.java
```diff
@@ -79,16 +78,46 @@
         UNCOMMITTED_EARLIEST,
         UNCOMMITTED_LATEST }
     
+    /**
+     * Convenience method to get a Builder for String key/value spouts. Sets the key/value Deserializers to the StringDeserializer class.
```
srdo: Sure, will shorten this

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutConfig.java
srdo: Nevermind, that won't work since KafkaSpoutConfig enforces the String type on keys. I'd at least like to throw an exception if the input properties contain a non-String key.

srdo:
@hmcl Addressed your comments, and updated the README to reflect the API changes.

hmcl:
@srdo thanks for addressing the code comments. It LGTM, but I forgot to publish the following comment - sorry for the extra overhead. Do you want to address it in this patch as well ?

Most of the defaults in [here](https://github.com/srdo/storm/blob/73adaa983cc421c7974e95011030c54f75e7698a/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutConfig.java#L38-#L51) are also in [here](https://github.com/srdo/storm/blob/73adaa983cc421c7974e95011030c54f75e7698a/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutConfig.java#L135-#L140). I think it would be better to have de defaults only in one place, ideally in the top level class. However, if it makes the code simpler to leave the defaults only in the Builder class, so be it. Neertheless, we should avoid having them in both places.

I also would like to ask if you can squash all the commits once we are done addressing everything, such that we can keep the git log clean. Thanks a lot.

srdo:
@hmcl I'm not sure how we can do that. The fields in Builder are not static, so if we move the default definitions there, we'd have to create a builder and fish out the default values in the tests. The defaults also wouldn't be visible on the KafkaSpoutConfig class, which would probably be a minus. I don't see a way to have the defaults solely in the KafkaSpoutConfig class with no references to them from Builder, since the Builer fields have to get the default values if nothing else is set.

We could move the FirstPollOffsetStrategy default into a static field in KafkaSpoutConfig like the rest for consistency at least?

srdo:
I'll squash the commits soon, I just wanted people to have a chance to review the changes without having to read the entire diff again

hmcl:
Yeah... I also think that's the ideal way to do it. Squash at the end and have a new commit addressing each batch of code review comments.

hmcl:
@srdo I agree with you concerning this [comment](https://github.com/apache/storm/pull/2155#issuecomment-311751331). There is really not a much better way to do this. I agree with your comment "move the FirstPollOffsetStrategy default into a static field in KafkaSpoutConfig like the rest for consistency" - let's do it.

I think that what confused me is the way the whole class is organized. For instance, the state of KafkaSpoutConfig is not at the top of the class as one would expect. Perhaps we should follow the usual convention and move the [constructor and fields](https://github.com/srdo/storm/blob/142d1f8e2609d1a48b6be71084987fe8cdc46a3f/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutConfig.java#L302-#L331) to the top, after the constants, then put the static inner classes ([here](https://github.com/srdo/storm/blob/142d1f8e2609d1a48b6be71084987fe8cdc46a3f/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutConfig.java#L59-#L79) and [here](https://github.com/srdo/storm/blob/142d1f8e2609d1a48b6be71084987fe8cdc46a3f/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutConfig.java#L125-#L300)), then the [factory methods](https://github.com/srdo/storm/blob/142d1f8e2609d1a48b6be71084987fe8cdc46a3f/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutConfig.java#L81-#L123), and at last all the [accessor methods](https://github.com/srdo/storm/blob/142d1f8e2609d1a48b6be71084987fe8cdc46a3f/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutConfig.java#L333-#L398).

On a different note, I noticed that this PR now has the changes from the [STORM-2541 PR](https://github.com/apache/storm/pull/2150) included in its single commit. I think that this PR should have only one commit, with the refactoring of KafkaSpoutConfig. The [STORM-2541 PR](https://github.com/apache/storm/pull/2150) should have two commits, this PR's commit, and it's commit. That would make it easy to merge and the same time not loose Git history.

hmcl:

srdo:
@hmcl Added a field for the offset strategy default and reordered the KafkaSpoutConfig contents as you suggested.

I didn't mean for the branches to contain each other, I accidentally reset a branch to the wrong content, should be fixed now. Once this or STORM-2541 goes in I'll fix any conflicts that pop up in the other PR.

srdo:
Actually you're right, building them on each other would be faster. I'll do as you suggested

srdo:
There were some +1s earlier, but this branch has changed a bunch since. I'd appreciate if someone could give it a skim and check that it's still good.

hmcl:
+1. @harshach @priyank5485 can you please take one final look. If you don't have any objection, I suggest that we merge this patch in the next day or so.

harshach:
still +1 .

srdo:
Thanks for the reviews everyone.

PR #2156, opened 2017-06-10
STORM-2549: Fix broken enforcement mechanism for maxUncommittedOffset…

srdo:
…s in storm-kafka-client spout

Please see https://issues.apache.org/jira/browse/STORM-2549 and maybe also the comments on https://issues.apache.org/jira/browse/STORM-2343

This moves enforcement of maxUncommittedOffsets to the partition level. I don't believe there's a way to enforce the single limit globally without hitting a bunch of issues with it either being too strict regarding retries, or not being strict enough and allowing an unbounded number of new tuples along with the retries.

### DIFF
--- a/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/OffsetManager.java
+++ b/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/OffsetManager.java
@@ -20,6 +20,7 @@ import com.google.common.annotations.VisibleForTesting;
 import java.util.Comparator;
 import java.util.Iterator;
 import java.util.NavigableSet;
+import java.util.NoSuchElementException;
 import java.util.TreeSet;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.apache.kafka.clients.consumer.OffsetAndMetadata;
@@ -64,6 +65,27 @@ public class OffsetManager {
 
     public void addToEmitMsgs(long offset) {
         this.emittedOffsets.add(offset);                  // O(Log N)
+    }
+    
+    public int getNumUncommittedOffsets() {
+        return this.emittedOffsets.size();
+    }
+    
+    /**
+     * Gets the offset of the nth emitted message after the committed offset. 
+     * Example: If the committed offset is 0 and offsets 1, 2, 8, 10 have been emitted,
+     * getNthUncommittedOffsetAfterCommittedOffset(3) returns 8.
+     * 
+     * @param index The index of the message to get the offset for
+     * @return The offset
+     * @throws NoSuchElementException if the index is out of range
+     */
+    public long getNthUncommittedOffsetAfterCommittedOffset(int index) {
+        Iterator<Long> offsetIter = emittedOffsets.iterator();
+        for (int i = 0; i < index - 1; i++) {
+            offsetIter.next();
+        }
+        return offsetIter.next();
     }
 
     /**
--- a/external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/KafkaSpoutEmitTest.java
+++ b/external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/KafkaSpoutEmitTest.java
@@ -26,8 +26,11 @@ import static org.mockito.Mockito.when;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.apache.kafka.clients.consumer.ConsumerRecords;
 import org.apache.kafka.clients.consumer.KafkaConsumer;
@@ -73,11 +76,7 @@ public class KafkaSpoutEmitTest {
         //This is necessary for Storm to be able to throttle the spout according to maxSpoutPending
         KafkaSpout<String, String> spout = SpoutWithMockedConsumerSetupHelper.setupSpout(spoutConfig, conf, contextMock, collectorMock, consumerMock, Collections.singleton(partition));
         Map<TopicPartition, List<ConsumerRecord<String, String>>> records = new HashMap<>();
-        List<ConsumerRecord<String, String>> recordsForPartition = new ArrayList<>();
-        for (int i = 0; i < 10; i++) {
-            recordsForPartition.add(new ConsumerRecord<>(partition.topic(), partition.partition(), i, "key", "value"));
-        }
-        records.put(partition, recordsForPartition);
+        records.put(partition, SpoutWithMockedConsumerSetupHelper.createRecords(partition, 0, 10));
 
         when(consumerMock.poll(anyLong()))
             .thenReturn(new ConsumerRecords<>(records));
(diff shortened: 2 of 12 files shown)


external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/KafkaSpoutEmitTest.java
```diff
@@ -134,110 +130,86 @@ public void testNextTupleEmitsFailedMessagesEvenWhenMaxUncommittedOffsetsIsExcee
...
+            setupSpout(partitions);
+            Map<TopicPartition, List<ConsumerRecord<String, String>>> records = new HashMap<>();
+            //This is cheating a bit since maxPollRecords would normally spread this across multiple polls
+            records.put(partition, createRecords(partition, 0, spoutConfig.getMaxUncommittedOffsets()));
+            records.put(partitionTwo, createRecords(partitionTwo, 0, spoutConfig.getMaxUncommittedOffsets()));
 
             when(consumerMock.poll(anyLong()))
-                .thenReturn(new ConsumerRecords(firstPollRecords))
-                .thenReturn(new ConsumerRecords(secondPollRecords));
+                .thenReturn(new ConsumerRecords(records));
```
askprasanna: Nit: return generic ConsumerRecords to avoid warning around raw type

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/KafkaSpoutEmitTest.java
```diff
@@ -134,110 +130,86 @@ public void testNextTupleEmitsFailedMessagesEvenWhenMaxUncommittedOffsetsIsExcee
...
             inOrder.verify(consumerMock).seek(partition, failedOffsets.get(0));
             inOrder.verify(consumerMock).poll(anyLong());
         }
     }
-    
+
+    private List<ConsumerRecord<String, String>> createRecords(TopicPartition topic, long startingOffset, int numRecords) {
+        List<ConsumerRecord<String, String>> recordsForPartition = new ArrayList<>();
+        for (int i = 0; i < numRecords; i++) {
+            recordsForPartition.add(new ConsumerRecord(topic.topic(), topic.partition(), startingOffset + i, "key", "value"));
```
askprasanna: Nit: return generic ConsumerRecords to avoid warning around raw type

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/KafkaSpoutEmitTest.java
```diff
@@ -134,110 +130,86 @@ public void testNextTupleEmitsFailedMessagesEvenWhenMaxUncommittedOffsetsIsExcee
...
-                spout.nextTuple();
-            }
-
-            ArgumentCaptor<KafkaSpoutMessageId> retryBatchMessageIdsCaptor = ArgumentCaptor.forClass(KafkaSpoutMessageId.class);
-            verify(collectorMock, times(maxPollRecords)).emit(anyObject(), anyObject(), retryBatchMessageIdsCaptor.capture());
-            reset(collectorMock);
             
-            //Check that the consumer started polling at the failed tuple offset
+            when(consumerMock.poll(anyLong()))
+                .thenReturn(new ConsumerRecords(Collections.singletonMap(partition, createRecords(partition, failedMessageId.get().offset(), 1))));
```
askprasanna: return generic ConsumerRecords to avoid warning around raw type

In this case I am getting an incompatible bounds error on the V param to singletonMap()

askprasanna:
+1 for the changes.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/OffsetManager.java
```diff
@@ -64,6 +65,27 @@ public void addToAckMsgs(KafkaSpoutMessageId msgId) {          // O(Log N)
...
+     * 
+     * @param index The index of the message to get the offset for
+     * @return The offset
+     * @throws NoSuchElementException if the index is out of range
+     */
+    public long getNthUncommittedOffsetAfterCommittedOffset(int index) {
+        Iterator<Long> offsetIter = emittedOffsets.iterator();
+        for (int i = 0; i < index - 1; i++) {
+            offsetIter.next();
+        }
```
srdo: Is copying the entire set to an array faster than iterating the set?

askprasanna: Guess we ll need to check the implementation to confirm how the array is created from the data structures behind the Set interface. Agree that it is possible for the iteration to be as fast or faster than the array creation. There may not be a clear cut winner here and it might depend on 'N'.

askprasanna: Looks like an iterator is used under the covers to create the array (which explains the ordering guarantee in the array). No change required.. :-)

hmcl:
@srdo I am reviewing this. Thanks.

hmcl:
@srdo I am doing one last review of this patch.

srdo:
Sorry, had to rebase to fix conflicts.

srdo:
Rebased again to fix conflicts

srdo:
Squashed

WolfeeTJ:
Hi @srdo , I happened to see errors like this in my production after applied your PR:

```
2017-09-06 13:12:01.707 o.a.s.util Thread-15-spout-executor[103 103] [ERROR] Async loop died!
java.lang.IllegalStateException: No current assignment for partition kafka_bd_trigger_action-22
        at org.apache.kafka.clients.consumer.internals.SubscriptionState.assignedState(SubscriptionState.java:231) ~[stormjar.jar:?]
        at org.apache.kafka.clients.consumer.internals.SubscriptionState.resume(SubscriptionState.java:350) ~[stormjar.jar:?]
        at org.apache.kafka.clients.consumer.KafkaConsumer.resume(KafkaConsumer.java:1332) ~[stormjar.jar:?]
        at org.apache.storm.kafka.spout.KafkaSpout.pollKafkaBroker(KafkaSpout.java:323) ~[stormjar.jar:?]
        at org.apache.storm.kafka.spout.KafkaSpout.nextTuple(KafkaSpout.java:234) ~[stormjar.jar:?]
        at org.apache.storm.daemon.executor$fn__9708$fn__9723$fn__9754.invoke(executor.clj:646) ~[storm-core-1.1.1-SNAPSHOT.jar:1.1.1-SNAPSHOT]
        at org.apache.storm.util$async_loop$fn__553.invoke(util.clj:484) [storm-core-1.1.1-SNAPSHOT.jar:1.1.1-SNAPSHOT]
        at java.lang.Thread.run(Thread.java:745) [?:1.8.0_121]
2017-09-06 13:12:01.709 o.a.s.d.executor Thread-15-spout-executor[103 103] [ERROR]
java.lang.IllegalStateException: No current assignment for partition kafka_bd_trigger_action-22
        at org.apache.kafka.clients.consumer.internals.SubscriptionState.assignedState(SubscriptionState.java:231) ~[stormjar.jar:?]
        at org.apache.kafka.clients.consumer.internals.SubscriptionState.resume(SubscriptionState.java:350) ~[stormjar.jar:?]
        at org.apache.kafka.clients.consumer.KafkaConsumer.resume(KafkaConsumer.java:1332) ~[stormjar.jar:?]
        at org.apache.storm.kafka.spout.KafkaSpout.pollKafkaBroker(KafkaSpout.java:323) ~[stormjar.jar:?]

```

I think the issue might happen because it could be possible partition reassignment takes effect at kafkaConsumer.poll(), and as a result the pausedPartitions is different from the actual assignment. How do you think?

srdo:
@WolfeeTJ Thanks for trying this out. I think you are right that it's because partition reassignment is happening at a bad time. I think we should move reassignment to be the first thing in nextTuple instead. Give me a little bit to fix this and maybe add a test.

I've been thinking a bit about partition revocation and shuffling, and we might want to add some warnings to the manual partition assignment API as well. There are a few cases I can think of where shuffling partitions can cause bugs. The Trident spout doesn't support partition shuffling because Trident doesn't expect partitions to move from task to task, as far as I can tell. When we implement at-most-once support for this spout there's also a requirement that partitions don't move between tasks, since otherwise it is possible that tuples are emitted more than once.

srdo:
@WolfeeTJ I think it's fixed now, please take a look at the new commit.

WolfeeTJ:
Hi @srdo , I'm not sure, so please correct me if I were wrong:

So far as I could see, I'm using the default `NamedSubscription`. And I checked the code, it's doing NOOP in `org.apache.storm.kafka.spout.Subscription#refreshAssignment()`.
My understanding is, actual reassignment happens when using 
`final ConsumerRecords<K, V> consumerRecords = kafkaConsumer.poll(kafkaSpoutConfig.getPollTimeoutMs());`
in `org.apache.storm.kafka.spout.KafkaSpout#pollKafkaBroker()`.

So we might need to get the fresh partition assignment for `kafkaConsumer.pause(pausedPartitions);` somehow after a `kafkaConsumer.poll()`, because the previous `pollablePartitions` list might has been changed after `kafkaConsumer.poll()` ?

srdo:
@WolfeeTJ We are dropping support for NamedSubscription and PatternSubscription in 2.0.0 and deprecating them in 1.2.0 because Kafka's mechanism for automatically assigning partitions to consumers is a bad fit for Storm. See https://issues.apache.org/jira/browse/STORM-2542 for details.  

From 1.2.0 (1.x-branch) on, the spout will assign partitions using the https://github.com/apache/storm/blob/master/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/subscription/ManualPartitionSubscription.java class. The default subscription in KafkaSpoutConfig will also change to use this subscription.

Partition assignment happens in refreshPartitions for the ManualPartitionSubscription, and poll won't trigger rebalance anymore, so we shouldn't need to worry about partitions changing when we call poll.

WolfeeTJ:
@srdo great, thank you a lot for your info. This helps a lot. In this case I think it should be ok. I'll apply the commit and try it out.
Thanks a lot!

srdo:
@WolfeeTJ No problem :)

Note that manual partition subscription doesn't work before version 1.1.2 https://issues.apache.org/jira/browse/STORM-2541, so you may want to check out 1.x-branch and apply this PR to it to get it working.

WolfeeTJ:
Thank you @srdo for pointing this out :) :+1: Will do.

srdo:
@HeartSaVioR Thanks for the review.

The 1.x version is here https://github.com/apache/storm/pull/2407. I tried cherry picking to 1.1.x, but there have been a lot of changes in this module since then, and I don't think it's worth the effort to pull back if we're releasing 1.2.0 soon anyway. If someone really needs it I'll pull it back to 1.1.x, but otherwise I'll probably just move this and STORM-2546 to the 1.2.0 epic instead.

HeartSaVioR:
@srdo 
Unfortunately I feel we couldn't bring Storm 1.1.2, and based on the fact, putting less effort to 1.1.x version line and focusing more on 1.2.0 may make sense to me. I guess it is not that trivial to port back indeed.
Let's revisit when someone asks us to port back. I'll take a look at #2407.

PR #2157, opened 2017-06-12
STORM-2517 storm-hdfs writers can't be subclassed

Angus-Slalom:


external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/AbstractHdfsBolt.java
```diff
@@ -297,9 +296,21 @@ protected Path getBasePathForNextFile(Tuple tuple) {
 
     abstract protected String getWriterKey(Tuple tuple);
 
-    abstract protected AbstractHDFSWriter makeNewWriter(Path path, Tuple tuple) throws IOException;
+    abstract protected Writer makeNewWriter(Path path, Tuple tuple) throws IOException;
 
-    static class WritersMap extends LinkedHashMap<String, AbstractHDFSWriter> {
+    public interface Writer {
```
harshach: can you move this into this its own file rather than putting it in HDFSBolt

dossett:
I hadn't thought about subclassing the writers, instead I thought a new bolt would implements its own writer.  @Angus-Slalom what use case do you have in mind?  I'm not opposed to the change, just curious.

Angus-Slalom:
@dossett the AbstractHDFSBolt uses AbstractHDFSWriter throughout, so any implementation of a writer has to inherit from that. For example the makeNewWriter protected function has to return an AbstractHDFSWriter
We've been subclassing the HDFS bolt to add more logging and to make tweaks to our AVRO output

dossett:
@Angus-Slalom Thanks, that makes sense! Feel free to contribute logging changes as well if you think they are generally applicable :-)

I'm at +1 pending the change Harsha suggested.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/AbstractHdfsBolt.java
Angus-Slalom: updated

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/Writer.java
```diff
@@ -0,0 +1,18 @@
+package org.apache.storm.hdfs.bolt;
```
HeartSaVioR: Missing apache license header.

HeartSaVioR:
@Angus-Slalom 
Could you craft the patch against master? We introduced checkstyle in master branch so forward porting via cherry-pick might not work.

Angus-Slalom:
Created a pull request on master at https://github.com/apache/storm/pull/2294

PR #2159, opened 2017-06-14
STORM-2552: KafkaSpoutMessageId should be serializable

askprasanna:
Equivalent to STORM-1848 and https://github.com/apache/storm/pull/1428

### DIFF
--- a/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutMessageId.java
+++ b/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutMessageId.java
@@ -18,13 +18,14 @@
 
 package org.apache.storm.kafka.spout;
 
+import java.io.Serializable;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.apache.kafka.common.TopicPartition;
 
-public class KafkaSpoutMessageId {
-    private final transient TopicPartition topicPart;
-    private final transient long offset;
-    private transient int numFails = 0;
+public class KafkaSpoutMessageId implements Serializable {
+    private final TopicPartition topicPart;
+    private final long offset;
+    private int numFails = 0;
     /**
      * true if the record was emitted using a form of collector.emit(...). false
      * when skipping null tuples as configured by the user in KafkaSpoutConfig


hmcl:
@askprasanna Why does it have to be serializable? It seems that it has been working until now. What changed that is enforcing this class to be serializable?

askprasanna:
My colleague reported an issue with his topology and the issue was a runtime exception around KafkaSpoutMessageID not being serializable. He had debug events enabled in his topology. I found STORM-1848 which references the same symptoms with a fix for the 'storm-kafka' package. Hence opened this PR to apply the same fix to 'storm-kafka-client'.

askprasanna:
Is there a reason why the same fix need not be applied for storm-kafka-client?

srdo:
As I recall (it's been a while), the eventlogging functionality enabled with "topology.eventlogger.executors" in storm.yaml will make both spouts and bolts send whatever they emit to an event logger bolt for logging. Spouts additionally send the message id https://github.com/apache/storm/blob/a4afacd9617d620f50cf026fc599821f7ac25c79/storm-client/src/jvm/org/apache/storm/executor/spout/SpoutOutputCollectorImpl.java#L111. This requires the message id to be serializable, or the spout executor will start throwing exceptions as soon as the button to enable debug logging is pressed in Storm UI.

@priyank5485 
No, I don't see a reason not to do this.

HeartSaVioR:
The build fails with check style violation exceeding with the change. Please fix that.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutMessageId.java
```diff
@@ -21,7 +21,9 @@
 import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.apache.kafka.common.TopicPartition;
 
-public class KafkaSpoutMessageId {
+import java.io.Serializable;
+
+public class KafkaSpoutMessageId implements Serializable{
     private final transient TopicPartition topicPart;
```
HeartSaVioR: @hmcl 
I guess the fields are all Serializable. Is that a reason to make them transient?

srdo: I think they probably shouldn't be. The eventlogger will be more useful if it can log topicpartition and offset.

HeartSaVioR:
While it will fix runtime issue, since fields are transient, it doesn't give any values. I think it's still better than getting runtime error, but we should remove transient if it's not necessary to do so. We may want to file a new issue for that.

@askprasanna 
+1
Could you squash commits into one? Thanks in advance.

askprasanna:
@HeartSaVioR Done. 

LMK if you would like another PR for making the members in KafkaSpoutMessageId non-transient.

srdo:
I wouldn't mind just squeezing that change into this PR honestly. It's still related to making the message id serializable, and it's such a tiny change that it feels like overkill to spin off. Either way, I definitely think we should remove transient.

+1 for this PR with or without the change

askprasanna:
Makes sense. Will update this PR with the change to make the members non-transient.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutMessageId.java
hmcl: @HeartSaVioR OK to remove transient. They were transient when it wasn't Serializable.

hmcl:
+1. LGTM. As @srdo mentioned earlier, I wish there was a backwards compatible way to enforce messageId to be serializable. The documentation is nice and clear, but the same way it happened in here, I don't think it will prevent further runtime exceptions.

HeartSaVioR:
Still +1

PR #2162, opened 2017-06-16
STORM-2449 Ensure same key appears only once in State iterator

HeartSaVioR:
* also don't expose TOMBSTONE keys

This implementation is for current Redis KV state.
I also have two State related pull requests #1970 and #1950 and I need to apply this patch to both of them. I'll update both after this patch is reviewed.

@arunmahadevan Please review and comment. Thanks in advance.


arunmahadevan:
+1 LGTM

Before you apply this to #1970 and #1950, you can refactor the logic to make it re-usable across the different state implementations.

PR #2163, opened 2017-06-16
STORM-2553:  Add support for password in JedisCluster

vyan:
Upgrade Jedis version to 2.9.0, and modify some api for password support in JedisCluster

### DIFF
--- a/external/storm-redis/src/main/java/org/apache/storm/redis/common/config/JedisClusterConfig.java
+++ b/external/storm-redis/src/main/java/org/apache/storm/redis/common/config/JedisClusterConfig.java
@@ -33,6 +33,7 @@ public class JedisClusterConfig implements Serializable {
     private Set<InetSocketAddress> nodes;
     private int timeout;
     private int maxRedirections;
+    private String password;
 
     /**
      * Constructor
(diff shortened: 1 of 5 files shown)


external/storm-redis/src/main/java/org/apache/storm/redis/common/config/JedisClusterConfig.java
```diff
@@ -44,14 +45,16 @@
      * @param nodes list of node information for JedisCluster
      * @param timeout socket / connection timeout
      * @param maxRedirections limit of redirections - how much we'll follow MOVED or ASK
+     * @param password password, if any
      * @throws NullPointerException when you didn't set nodes
      */
-    public JedisClusterConfig(Set<InetSocketAddress> nodes, int timeout, int maxRedirections) {
+    public JedisClusterConfig(Set<InetSocketAddress> nodes, int timeout, int maxRedirections, String password) {
```
HeartSaVioR: Though we also have builder class as well, it would be better to have a separate constructor which doesn't have password as parameter, given that it was `public`.

HeartSaVioR:
Looks good overall. Left comment regarding backward compatibility.

external/storm-redis/src/main/java/org/apache/storm/redis/common/config/JedisClusterConfig.java
```diff
@@ -45,6 +45,26 @@
      * @param nodes list of node information for JedisCluster
      * @param timeout socket / connection timeout
      * @param maxRedirections limit of redirections - how much we'll follow MOVED or ASK
+     * @throws NullPointerException when you didn't set nodes
+     */
+    public JedisClusterConfig(Set<InetSocketAddress> nodes, int timeout, int maxRedirections) {
+        Preconditions.checkNotNull(nodes, "Node information should be presented");
```
HeartSaVioR: Maybe better to call `this(nodes, timeout, maxRedirections, null);` to remove duplicate codes.

HeartSaVioR:
@vyan Please squash commits into one after addressing my comment. Thanks in advance.

vyan:
@HeartSaVioR, thanks for your comments. I have squashed all commits into one now.

HeartSaVioR:
+1

PR #2164, opened 2017-06-17
[STORM-2557]: A bug in DisruptorQueue causing severe underestimation of…

wendyshusband:
Recently, we are tuning the performance of our topology and deploying some theoretical performance models that heavily rely on the metrics of query arrival rates. We found a bug in `DisruptorQueue `that leads to severe underestimation to the queue arrival rates. After further investigation, we finally found that in the current implementation of `DisruptorQueue`, the arrival rates are actually measured as the number of batches of tuples rather than the actual number of tuples, resulting in significant underestimation of the arrival rates.

To be more specific, in` DisruptorQueue.publishDirectSingle()` and `DisruptorQueue.publishDirect()` functions, objects containing tuples are published to the buffer and the metrics are notified by calling `_metric.notifyArrivals(1)`. This works fine when the object is simply a wrapper of a single tuple. However, the object could also be an instance of `ArrayList<AddressedTuple>` or `HashMap<Integer, ArrayList<TaskMessage>>`. In such case, we should get the actual number of tuples in the object and notify the `_metrics `with the right value.

I added some code that determine the type of object to fix this issue at `DisruptorQueue.publishDirectSingle()` and `DisruptorQueue.publishDirect()` functions.

Any comment or suggestion regarding to this pull request is welcomed.

Thanks.

### DIFF
--- a/storm-client/src/jvm/org/apache/storm/utils/DisruptorQueue.java
+++ b/storm-client/src/jvm/org/apache/storm/utils/DisruptorQueue.java
@@ -500,6 +500,22 @@ public class DisruptorQueue implements IStatefulObject {
         return Thread.currentThread().getId();
     }
 
+    private long getTupleCount(Object obj) {
+        //a published object could be an instance of either AddressedTuple, ArrayList<AddressedTuple>, or HashMap<Integer, ArrayList<TaskMessage>>.
+        long tupleCount;
+        if (obj instanceof ArrayList) {
+            tupleCount = ((ArrayList) obj).size();
+        } else if (obj instanceof HashMap) {
+            tupleCount = 0;
+            for (Object value:((HashMap) obj).values()) {
+                tupleCount += ((ArrayList) value).size();
+            }
+        } else {
+            tupleCount = 1;
+        }
+        return tupleCount;
+    }
+
     private void publishDirectSingle(Object obj, boolean block) throws InsufficientCapacityException {
         long at;
         long numberOfTuples;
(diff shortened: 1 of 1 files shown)


storm-client/src/jvm/org/apache/storm/utils/DisruptorQueue.java
```diff
@@ -502,15 +502,26 @@ private static Long getId() {
...
         long at;
+        long numberOfTuples;
         if (block) {
             at = _buffer.next();
         } else {
             at = _buffer.tryNext();
         }
         AtomicReference<Object> m = _buffer.get(at);
         m.set(obj);
+        if (obj instanceof ArrayList) {
```
HeartSaVioR: Seems like the logic for counting number of tuples is duplicated so better to extract method.

HeartSaVioR:
@wendyshusband 
Thanks for the contribution. Looks good. Minor comment.

@revans2 I guess you would want to review this as well, since it's related to disruptor batching.

storm-client/src/jvm/org/apache/storm/utils/DisruptorQueue.java
wendyshusband: Thank you for your suggestion. I have modified accordingly.

wangli1426:
+1

revans2:
The change as is looks OK to me.  My biggest problem with it is that there is now tight coupling between the Disruptor queue and storm itself, although looking at the code that was true before.

Now that we are pure java for both publishers and consumers I really would like to see us add in some generics so it is more obvious what we are storing in the queue, but that can be in a follow on JIRA.

+1

wangli1426:
@revans2 I quite agree with you. The current publish logic in disruptor queue is difficult to follow. It will be great if some one could refine the code in a separate JIRA, so that we can easily know what are storing in the queue.

@HeartSaVioR Given the feedback from revans2, could you merge this PR?

HeartSaVioR:
OK. +1 from me.

wendyshusband:
I apologize for the lack of timely reply. thank you very much.

revans2:
I just realized that because the `sojourn_time_ms` is calculated based off of the request rate, that this updated, and the population (not updated and still based off of slots and not tuples), that this made the sojurn time horribly inaccurate.  I'll file a JIRA to address it.

wangli1426:
@HeartSaVioR @asfgit It seems that we forgot to add @wendyshusband to the contributor list.

Ethanlm:
@wangli1426  It would be great if you could also add me to the contributor list too LOL

wendyshusband:
@HeartSaVioR @wangli1426 @revans2 I guess you forgot to add me to the contributor list, after two years.

Ethanlm:
Hi @wendyshusband  I verified that you are in the contributor list. 

Screen Shot 2020-01-08 at 9 18 06 AM

PR #2165, opened 2017-06-17
STORM-2558: Port storm.sh to Powershell and remove outdated batch scr…

srdo:
…ipts

See https://issues.apache.org/jira/browse/STORM-2558

I'm wondering if it is feasible to port most of the bash script to Python so we have less nonportable code to maintain.


HeartSaVioR:
Sadly I guess no one knows power shell script, and also most of committers/PMCs don't even have Windows PC (including me, for dev. machine).
Would you want to bump this with dev@ to find reviewers?

srdo:
@HeartSaVioR I think powershell runs on Linux now https://github.com/PowerShell/PowerShell

I'm hoping this can be reviewed without knowing much about Powershell, since the script is a line for line translation of the bash script. I'd be happy to post to the mailing list to find reviewers, if you think it would help?

harshach:
+1. Tried on windows 10 looks good to me.

srdo:
Thanks for trying it out :)

HeartSaVioR:
OK then I would also +1 given that the patch is confirmed to be manually tested.

bin/storm.py
```diff
@@ -94,7 +94,7 @@ def init_storm_env():
 CONFFILE = ""
 JAR_JVM_OPTS = shlex.split(os.getenv('STORM_JAR_JVM_OPTS', ''))
 JAVA_HOME = os.getenv('JAVA_HOME', None)
-JAVA_CMD = 'java' if not JAVA_HOME else os.path.join(JAVA_HOME, 'bin', 'java')
+JAVA_CMD = 'java' if not JAVA_HOME else os.path.join(JAVA_HOME, 'bin', 'java' if not is_windows() else "java.exe")
```
HeartSaVioR: This one-liner becomes a bit complicated now. Could we expand this to multiple lines? 
And isn't the difference between non-windows and windows be 'java' or 'java.exe'? I mean `java vs java.exe` or `<JAVA_HOME>/bin/java vs <JAVA_HOME>\bin\java.exe` might be correct whereas this line doesn't do the thing. Please let me know if I'm missing here.

HeartSaVioR:
Oops I left a review comment so please treat this as revoking +1.

bin/storm.py
srdo: I've put the code into a new function. I'm a little uncertain what you're asking in the rest of the comment. The java binary on Windows is called java.exe, so `<JAVA_HOME>/bin/java vs <JAVA_HOME>\bin\java.exe` is correct. The previous implementation should do that?

HeartSaVioR: @srdo 
I meant to say we should set JAVA_CMD to `java.exe` for windows if JAVA_HOME is not provided. The line sets JAVA_CMD to `java` if JAVA_HOME is not provided regardless of OS.
Yes I think previous implementation itself is wrong.

bin/storm.py
```diff
@@ -66,6 +66,12 @@ def init_storm_env():
     for option in options:
         value = config.get('environment', option)
         os.environ[option] = value
+        
+def get_java_cmd():
```
HeartSaVioR: Yes I think this is correct and keeping this to one-liner fails to do this.

HeartSaVioR:
@srdo 
Thanks for the quick fix. +1
Could you squash the commits into one? Thanks in advance!

bin/storm.py
srdo: The one liner should have done the same thing, except when JAVA_HOME wasn't set. But in that case it's okay to just call "java", since you don't have to postfix with ".exe" for binaries on the PATH in Windows. I agree that this is better though :)

HeartSaVioR: Ah didn't know about the magic of Windows. That was why it worked anyway and I just wondered why it works. Thanks for letting me know. :)

PR #2169, opened 2017-06-21
[STORM-2563] Remove the workaround to handle missing UGI.loginUserFromSubject

arunmahadevan:
https://github.com/apache/storm/blob/master/storm-client/src/jvm/org/apache/storm/security/auth/kerberos/AutoTGT.java#L225

The "userCons.setAccessible(true)" invokes constructor of a package private class bypassing the Java access control checks and raising red flags in our internal security scans.

The "loginUserFromSubject(Subject subject)" has been added to UGI (https://issues.apache.org/jira/browse/HADOOP-10164) and available since Hadoop version 2.3 released over three years ago (http://hadoop.apache.org/releases.html).

I think the workaround is no longer required since the case will not happen when using hadoop-common versions >= 2.3

@revans2 , could you take a look ?

### DIFF
--- a/storm-client/src/jvm/org/apache/storm/security/auth/kerberos/AutoTGT.java
+++ b/storm-client/src/jvm/org/apache/storm/security/auth/kerberos/AutoTGT.java
@@ -188,44 +188,8 @@ public class AutoTGT implements IAutoCredentials, ICredentialsRenewer {
                   "in your jar");
                 return;
             }
- 
-            try {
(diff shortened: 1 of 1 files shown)


HeartSaVioR:
+1
We've been setting `hadoop.version` as 2.6.1 for a long period so if it's an workaround for lower version, we could remove it. But it would be better to wait for @revans2 to review this as well.

hmcl:
+1

PR #2170, opened 2017-06-21
[STORM-2564] We should provide a template for storm-cluster-auth.yaml

liu-zhaokun:
[https://issues.apache.org/jira/browse/STORM-2564](https://issues.apache.org/jira/browse/STORM-2564)
As the configuration which named storm.zookeeper.auth.payload should be filled in "storm-cluster-auth.yaml",and there isn't such a file.So, I think we should provide a template for storm-cluster-auth.yaml.

### DIFF
--- a/conf/storm-cluster-auth.yaml.example
+++ b/conf/storm-cluster-auth.yaml.example
@@ -1,0 +1,18 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


liu-zhaokun:
@vesense 
Could you help me review this PR?

liu-zhaokun:
@HeartSaVioR 
Hello,are you available to help me review this PR?

HeartSaVioR:
If this is needed, why not adding storm.yaml.example and commenting it out? I don't think users want to copy a line from template file and paste to their storm.yaml.

liu-zhaokun:
@HeartSaVioR 
It says "Note that if this is set to something with a secret (as when using digest authentication) then it should only be set in the storm-cluster-auth.yaml file." in Config.java,line 1082. So I think we should provide  
a  template.

liu-zhaokun:
@HeartSaVioR 
Users needn't copy the configuration,storm.zookeeper.auth.payload, from storm-cluster-auth.yaml.template and paste to their storm.yaml.They only should add this configuration to  storm-cluster-auth.yaml if they need.

HeartSaVioR:
@liu-zhaokun Thanks for correcting. You're right. I'm not aware of that. +1

liu-zhaokun:
@HeartSaVioR 
Thanks for your reply,if possible, please help me merge it as soon as possible,because this PR has been created for a long time.

PR #2172, opened 2017-06-22
STORM-2369 [storm-redis] Use binary type for State management

HeartSaVioR:
* introduce new command interface and relevant containers
  * new command interface will contain both binary and string commands in near future
* change RedisKeyValue to use byte[] for State key and value management instead of String
  * get rid of Base64 encode/decode on RedisEncoder, and also SafeEncoder.encode() in Jedis internal
* users can select either Redis single server or Redis Cluster by passing JedisPoolConfig or JedisClusterConfig
* implement binary based KV state iterator
* Provide a tool to migrate old base64 state to the new binary state
  * migration tool is placed to storm-redis-examples
  * the module is easy to create uber jar hence easy to run the class
* document how to build and use migration tool

This patch is equivalent to #1950. Please note that there're some other changes for fixing checkstyle. I didn't address all of them, but my patch exceeded the limit so I ended up to fix some in this patch.

### DIFF
--- a/docs/State-checkpointing.md
+++ b/docs/State-checkpointing.md
@@ -70,6 +70,48 @@ json config with the following properties.
     }
 }
 ```
+ 
+For Redis Cluster state this is a json config with the following properties.
+ 
+```
+ {
+   "keyClass": "Optional fully qualified class name of the Key type.",
+   "valueClass": "Optional fully qualified class name of the Value type.",
+   "keySerializerClass": "Optional Key serializer implementation class.",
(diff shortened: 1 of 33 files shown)


HeartSaVioR:
Just update the PR to make equivalent to #1950

HeartSaVioR:
Filed an issue for versioning state: https://issues.apache.org/jira/browse/STORM-2605

arunmahadevan:
+1

PR #2176, opened 2017-06-24
STORM-2598 Add proxy server option for dependency resolver

HeartSaVioR:
* add options to storm.py and dependency resolver: --proxyUrl, --proxyUsername, --proxyPassword
* address some checkstyle issue for storm-submit-tool


harshach:
+1

HeartSaVioR:
@harshach @satishd Thanks for reviewing. Could you review #2175 as well? There're some differences from bin/storm.py between two branches.

PR #2177, opened 2017-06-24
STORM-2568: Fix getTopicsString

haewanj:
- String.valueOf returns square bracket, because 'topics' is Collection
- removed square bracket to use storm-kafka-monitor

### DIFF
--- a/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/NamedSubscription.java
+++ b/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/NamedSubscription.java
@@ -22,6 +22,7 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 
+import org.apache.commons.lang.StringUtils;
 import org.apache.kafka.clients.consumer.ConsumerRebalanceListener;
 import org.apache.kafka.clients.consumer.KafkaConsumer;
 import org.apache.storm.task.TopologyContext;
@@ -56,6 +57,6 @@ public class NamedSubscription extends Subscription {
 
     @Override
     public String getTopicsString() {
-        return String.valueOf(topics);
+        return StringUtils.join(topics, ",");
     }
 }
--- a/external/storm-kafka-client/pom.xml
+++ b/external/storm-kafka-client/pom.xml
@@ -40,6 +40,10 @@
     </developers>
 
     <dependencies>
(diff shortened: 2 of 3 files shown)


HeartSaVioR:
+1
I checked the master branch and confirmed that branch is OK: master branch uses `String.join()`
So don't need to port forward to the master branch.

HeartSaVioR:
@haewanj 
Sorry I revoked +1 because of missing license. Could you add Apache license header to the new file?
And please squash the commits into one after fixing. Thanks in advance!

haewanj:
@HeartSaVioR 
I added license. and squashed the commits into one. Thanks for the advice!

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/NamedSubscription.java
```diff
@@ -56,6 +57,6 @@ public NamedSubscription(String ... topics) {
 
     @Override
     public String getTopicsString() {
-        return String.valueOf(topics);
+        return StringUtils.join(topics, ",");
```
hmcl: I would like to suggest that you use instead Guava's `Joiner.on(",").join(topics);` which will avoid adding a new dependency

haewanj: Thanks for the suggestion. 
As you know, To keep java 1.7+ build, 'String.join()' is not a solution like master branch. And 'storm-core' has dependency both commons-lang and guava. So I choose commons-lang but If to pass the test that travis MODUELS='!storm-core', storm-kafka-client needs the dependency even Guava too. That's why I added commons-lang to storm-kafka-client's pom file.
If avoid adding a new dependency in storm-kafka-client, we can code maybe like this
        StringBuilder sb = new StringBuilder();
        for(String topic: topics) {
            sb.append(topic).append(",");
        }
        sb.deleteCharAt(sb.lastIndexOf(","));
        return sb.toString();
Many of external/packages have commons-lang dependency so, I thought it was fine.

By the way It has not been passed the Integration-test on travis-ci, I have no idea. but in my local integration-test is fine. (ex: mvn -P  integration-tests-only integration-test)

hmcl:
+1

HeartSaVioR:
+1
Thanks for the contribution.

PR #2179, opened 2017-06-26
STORM-2601: add the timeout parameter to the method of getting the nimbus client

MichealShin:
To void the long time waiting, add the timeout parameter to the method of getting the nimbus client.


HeartSaVioR:
+1
@MichealShin Could you change the commit title to include STORM-2601 so that we can track easily? Thanks in advance!

MichealShin:
@HeartSaVioR OK.

PR #2184, opened 2017-06-30
STORM-2610: Fixed the metrics

revans2:
I still need to add these to the docs, but I wanted to capture my progress before the long weekend.

### DIFF
--- a/storm-client/src/jvm/org/apache/storm/daemon/metrics/SpoutThrottlingMetrics.java
+++ b/storm-client/src/jvm/org/apache/storm/daemon/metrics/SpoutThrottlingMetrics.java
@@ -15,6 +15,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 package org.apache.storm.daemon.metrics;
 
 import org.apache.storm.metric.api.CountMetric;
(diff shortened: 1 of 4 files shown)


storm-client/src/jvm/org/apache/storm/utils/NimbusClient.java
```diff
@@ -96,6 +96,10 @@ public static NimbusClient getConfiguredClientAs(Map<String, Object> conf, Strin
         if (override != null) {
             return new NimbusClient(override);
         }
+        Map<String, Object> fullConf = Utils.readStormConfig();
```
HeartSaVioR: I'm not sure this change is related to the origin issue. Could you explain this change?

storm-client/src/jvm/org/apache/storm/daemon/metrics/SpoutThrottlingMetrics.java
```diff
@@ -26,32 +27,20 @@
...
-        return skippedInactive;
+        metricMap.put("skipped-max-spout-ms", skippedMaxSpout);
+        metricMap.put("skipped-throttle-ms", skippedThrottle);
+        metricMap.put("skipped-inactive-ms", skippedInactive);
     }
 
-    public void skippedMaxSpout(CommonStats stats) {
-        this.skippedMaxSpout.incrBy(stats.getRate());
+    public void skippedMaxSpoutMs(long ms) {
+        this.skippedMaxSpout.incrBy(ms);
```
HeartSaVioR: Why not also rename the fields (skippedMaxSpout, skippedThrottle, skippedInactive) to add `Ms` as postfix?

HeartSaVioR:
It looks good overall. This change doesn't touch critical path so don't need to consider performance side.

storm-client/src/jvm/org/apache/storm/utils/NimbusClient.java
revans2: This is something I ran into trying to get the ThroughputVsLatency topology to run, so I could test the metrics.  It simply means that if you pass in an empty config that it will fill in the rest from the command line, storm.yaml, and defaults.yaml, just like when submitting a topology.

storm-client/src/jvm/org/apache/storm/daemon/metrics/SpoutThrottlingMetrics.java
revans2: Good point will do that.

HeartSaVioR:
We didn't document individual builtin metric so maybe no need to document the change or document whole builtin metrics. It is breaking change in point of builtin metrics (for users who use metrics consumer) indeed.

+1

revans2:
@HeartSaVioR I filed STORM-2616 to document all of the built in metrics, and I'll try to get a pull request for that up some time soon.

For now I will leave this as a 2.x change, and if someone really wants/needs this for 1.x I can add the ms metrics and fix the count metrics.

PR #2185, opened 2017-07-03
[STORM-2613] Tuples that aren't sampled shouldn't be considered for latency calcul…

vinodkc:
Set delta to -1 , so that time delay for tuples that aren't sampled will not be added in BoltExecutorStats


HeartSaVioR:
+1 nice finding.

@vinodkc Did you find this spot from 1.x version line? If then could you craft pull request against 1.x-version too?

vinodkc:
Sure, I'll make a PR against 1.x vesrion

PR #2032, opened 2017-03-28
[STORM-2093] Fix permissions in multi-tenant, secure mode

ppoulosk:
Heap dumps created on OOM, when served through the logviewer in secure multi-tenant configurations have permissions set such that the logviewer cannot read them.

This change checks permissions in this case before serving and then runs the worker-launcher to enable the logviewer to serve them.

PR #2187, opened 2017-07-05
[STORM-2093] Fix permissions in multi-tenant, secure mode

Ethanlm:
Heap dumps created on OOM, when served through the logviewer in secure multi-tenant configurations have permissions set such that the logviewer cannot read them.

This change checks permissions in this case before serving and then runs the worker-launcher to enable the logviewer to serve them.  

see https://github.com/apache/storm/pull/2032

### DIFF
--- a/storm-core/src/clj/org/apache/storm/daemon/logviewer.clj
+++ b/storm-core/src/clj/org/apache/storm/daemon/logviewer.clj
@@ -21,6 +21,7 @@
   (:use [org.apache.storm config daemon-config util log])
   (:use [org.apache.storm.ui helpers])
   (:import [org.apache.storm StormTimer]
+           [org.apache.storm.daemon.supervisor ClientSupervisorUtils]
            [org.apache.storm.daemon.supervisor SupervisorUtils]
            [org.apache.storm.metric StormMetricsRegistry])
   (:import [org.apache.storm.utils Time VersionInfo ConfigUtils Utils ServerUtils ServerConfigUtils])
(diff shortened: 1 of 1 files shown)


Ethanlm:
@revans2 Yes, I tested it manually.

storm-core/src/clj/org/apache/storm/daemon/logviewer.clj
```diff
@@ -502,6 +503,20 @@
       (-> (resp/response "Page not found")
           (resp/status 404)))))
 
+(defnk set-log-file-permissions [fname root-dir]
+       (let [file (.getCanonicalFile (File. root-dir fname))
```
knusbaum: Indentation on this function is off.

storm-core/src/clj/org/apache/storm/daemon/logviewer.clj
```diff
@@ -502,6 +503,20 @@
...
+             run-as-user (*STORM-CONF* SUPERVISOR-RUN-WORKER-AS-USER)
+             parent (.getParent (File. root-dir fname))
+             md-file (if (nil? parent) nil (get-metadata-file-for-wroker-logdir parent))
+             topo-owner (if (nil? md-file) nil (get-topo-owner-from-metadata-file (.getCanonicalPath md-file)))]
+         (if (and run-as-user
```
knusbaum: Nit: since there is no 'else' branch, replacing `if` with `when` will allow you to remove the `do`.

storm-core/src/clj/org/apache/storm/daemon/logviewer.clj
Ethanlm: Thanks. Will fix it quickly

storm-core/src/clj/org/apache/storm/daemon/logviewer.clj
Ethanlm: Thanks. Will fix it quickly

knusbaum:
+1

HeartSaVioR:
+1
@Ethanlm Could you squash commits into one? Thanks in advance.

PR #2193, opened 2017-07-09
STORM-2589: Apply new code style to storm-mongodb

vesense:
Reduce the max allowed violation count to 0.


HeartSaVioR:
+1
This patch changes the public class name MongoDBClient -> MongoDbClient, but we're expecting users to use Bolts, not that class directly.

PR #2194, opened 2017-07-09
STORM-2594: Apply new code style to storm-rocketmq

vesense:
Reduce the max allowed violation count to 0.

Changed the class names including MQ to Mq. 
>Since storm-rocketmq is only merged into master branch and storm 2.0 hasn't been released, so the changes should hurt nothing.


HeartSaVioR:
+1

PR #2190, opened 2017-07-09
STORM-2608 Remove any pending offsets that are no longer valid

kevinconaway:
@HeartSaVioR this PR includes fixes for https://issues.apache.org/jira/browse/STORM-2608

PR #2195, opened 2017-07-10
STORM-2608 Remove any pending offsets that are no longer valid (master)

kevinconaway:
@HeartSaVioR  this PR includes fixes for https://issues.apache.org/jira/browse/STORM-2608 targeting the master branch per https://github.com/apache/storm/pull/2190


HeartSaVioR:
+1
It's forward port to #2190 so I think I don't need to follow 24hr rule.

PR #2196, opened 2017-07-10
STORM-2619: Correct the JDK version in Setting-up-a-Storm-cluster.md

vesense:
The Storm 2.0 needs JDK 1.8+


HeartSaVioR:
#2197 covers this and that's more explanatory so would like to pick that. Are you OK with this?

vesense:
OK. No problem. I will close this.

PR #2199, opened 2017-07-10
[STORM-2201] Add dynamic scheduler configuration loading

Ethanlm:
This adds an interface and two implementations, one that will load from a local file, and another
that will load a config from artifactory.

It also modifies the ResourceAwareScheduler and Multitenant schedulers to have a plugin interface and configuration entries for the plugin.

Unit tests are also added for the implementations.

See: https://github.com/apache/storm/pull/1785

Also fixed some checkstyle issues.

### DIFF
--- a/storm-server/pom.xml
+++ b/storm-server/pom.xml
@@ -59,6 +59,11 @@
             <groupId>org.apache.commons</groupId>
             <artifactId>commons-compress</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.google.auto.service</groupId>
+            <artifactId>auto-service</artifactId>
+            <optional>true</optional>
+        </dependency>
 
         <!-- test -->
         <dependency>
--- a/storm-server/src/main/java/org/apache/storm/DaemonConfig.java
+++ b/storm-server/src/main/java/org/apache/storm/DaemonConfig.java
@@ -221,7 +221,7 @@ public class DaemonConfig implements Validated {
     public static final String NIMBUS_TOPOLOGY_VALIDATOR = "nimbus.topology.validator";
 
     /**
-     * Class name for authorization plugin for Nimbus
+     * Class name for authorization plugin for Nimbus.
      */
     @isString
     public static final String NIMBUS_AUTHORIZER = "nimbus.authorizer";
@@ -271,7 +271,7 @@ public class DaemonConfig implements Validated {
     public static final String UI_CENTRAL_LOGGING_URL = "ui.central.logging.url";
 
     /**
-     * HTTP UI port for log viewer
+     * HTTP UI port for log viewer.
      */
     @isInteger
     @isPositiveNumber
@@ -284,46 +284,46 @@ public class DaemonConfig implements Validated {
     public static final String LOGVIEWER_CHILDOPTS = "logviewer.childopts";
 
     /**
-     * How often to clean up old log files
+     * How often to clean up old log files.
      */
     @isInteger
     @isPositiveNumber
     public static final String LOGVIEWER_CLEANUP_INTERVAL_SECS = "logviewer.cleanup.interval.secs";
 
     /**
-     * How many minutes since a log was last modified for the log to be considered for clean-up
+     * How many minutes since a log was last modified for the log to be considered for clean-up.
      */
     @isInteger
     @isPositiveNumber
     public static final String LOGVIEWER_CLEANUP_AGE_MINS = "logviewer.cleanup.age.mins";
 
     /**
-     * The maximum number of bytes all worker log files can take up in MB
+     * The maximum number of bytes all worker log files can take up in MB.
      */
     @isPositiveNumber
     public static final String LOGVIEWER_MAX_SUM_WORKER_LOGS_SIZE_MB = "logviewer.max.sum.worker.logs.size.mb";
 
     /**
-     * The maximum number of bytes per worker's files can take up in MB
+     * The maximum number of bytes per worker's files can take up in MB.
      */
     @isPositiveNumber
     public static final String LOGVIEWER_MAX_PER_WORKER_LOGS_SIZE_MB = "logviewer.max.per.worker.logs.size.mb";
 
     /**
-     * Storm Logviewer HTTPS port
+     * Storm Logviewer HTTPS port.
      */
     @isInteger
     @isPositiveNumber
     public static final String LOGVIEWER_HTTPS_PORT = "logviewer.https.port";
 
     /**
-     * Path to the keystore containing the certs used by Storm Logviewer for HTTPS communications
+     * Path to the keystore containing the certs used by Storm Logviewer for HTTPS communications.
      */
     @isString
     public static final String LOGVIEWER_HTTPS_KEYSTORE_PATH = "logviewer.https.keystore.path";
 
     /**
-     * Password for the keystore for HTTPS for Storm Logviewer
+     * Password for the keystore for HTTPS for Storm Logviewer.
      */
     @isString
     public static final String LOGVIEWER_HTTPS_KEYSTORE_PASSWORD = "logviewer.https.keystore.password";
@@ -342,13 +342,13 @@ public class DaemonConfig implements Validated {
     public static final String LOGVIEWER_HTTPS_KEY_PASSWORD = "logviewer.https.key.password";
 
     /**
-     * Path to the truststore containing the certs used by Storm Logviewer for HTTPS communications
+     * Path to the truststore containing the certs used by Storm Logviewer for HTTPS communications.
      */
     @isString
     public static final String LOGVIEWER_HTTPS_TRUSTSTORE_PATH = "logviewer.https.truststore.path";
 
     /**
-     * Password for the truststore for HTTPS for Storm Logviewer
+     * Password for the truststore for HTTPS for Storm Logviewer.
      */
     @isString
     public static final String LOGVIEWER_HTTPS_TRUSTSTORE_PASSWORD = "logviewer.https.truststore.password";
@@ -370,13 +370,13 @@ public class DaemonConfig implements Validated {
     public static final String LOGVIEWER_HTTPS_NEED_CLIENT_AUTH = "logviewer.https.need.client.auth";
 
     /**
-     * A list of users allowed to view logs via the Log Viewer
+     * A list of users allowed to view logs via the Log Viewer.
      */
     @isStringList
     public static final String LOGS_USERS = "logs.users";
 
     /**
-     * A list of groups allowed to view logs via the Log Viewer
+     * A list of groups allowed to view logs via the Log Viewer.
      */
     @isStringList
     public static final String LOGS_GROUPS = "logs.groups";
@@ -394,19 +394,19 @@ public class DaemonConfig implements Validated {
     public static final String UI_CHILDOPTS = "ui.childopts";
 
     /**
-     * A class implementing javax.servlet.Filter for authenticating/filtering UI requests
+     * A class implementing javax.servlet.Filter for authenticating/filtering UI requests.
      */
     @isString
     public static final String UI_FILTER = "ui.filter";
 
     /**
-     * Initialization parameters for the javax.servlet.Filter
+     * Initialization parameters for the javax.servlet.Filter.
      */
     @isMapEntryType(keyType = String.class, valueType = String.class)
     public static final String UI_FILTER_PARAMS = "ui.filter.params";
 
     /**
-     * The size of the header buffer for the UI in bytes
+     * The size of the header buffer for the UI in bytes.
      */
     @isInteger
     @isPositiveNumber
@@ -555,7 +555,7 @@ public class DaemonConfig implements Validated {
     public static final String DRPC_HTTPS_NEED_CLIENT_AUTH = "drpc.https.need.client.auth";
 
     /**
-     * Class name for authorization plugin for DRPC client
+     * Class name for authorization plugin for DRPC client.
      */
     @isString
     public static final String DRPC_AUTHORIZER = "drpc.authorizer";
@@ -578,7 +578,7 @@ public class DaemonConfig implements Validated {
     public static final String DRPC_CHILDOPTS = "drpc.childopts";
 
     /**
-     * the metadata configured on the supervisor
+     * the metadata configured on the supervisor.
      */
     @isMapEntryType(keyType = String.class, valueType = String.class)
     public static final String SUPERVISOR_SCHEDULER_META = "supervisor.scheduler.meta";
@@ -590,7 +590,7 @@ public class DaemonConfig implements Validated {
      */
     @isNoDuplicateInList
     @NotNull
-    @isListEntryCustom(entryValidatorClasses={ConfigValidation.IntegerValidator.class,ConfigValidation.PositiveNumberValidator.class})
+    @isListEntryCustom(entryValidatorClasses = {ConfigValidation.IntegerValidator.class,ConfigValidation.PositiveNumberValidator.class})
     public static final String SUPERVISOR_SLOTS_PORTS = "supervisor.slots.ports";
 
     /**
@@ -652,14 +652,14 @@ public class DaemonConfig implements Validated {
     public static final String NIMBUS_SLOTS_PER_TOPOLOGY = "nimbus.slots.perTopology";
 
     /**
-     * A class implementing javax.servlet.Filter for DRPC HTTP requests
+     * A class implementing javax.servlet.Filter for DRPC HTTP requests.
      */
     @isString
     public static final String DRPC_HTTP_FILTER = "drpc.http.filter";
 
     /**
      * Initialization parameters for the javax.servlet.Filter of the DRPC HTTP
-     * service
+     * service.
      */
     @isMapEntryType(keyType = String.class, valueType = String.class)
     public static final String DRPC_HTTP_FILTER_PARAMS = "drpc.http.filter.params";
@@ -679,7 +679,7 @@ public class DaemonConfig implements Validated {
     public static final String SUPERVISOR_CHILDOPTS = "supervisor.childopts";
 
     /**
-     * How many seconds to sleep for before shutting down threads on worker
+     * How many seconds to sleep for before shutting down threads on worker.
      */
     @isInteger
     @isPositiveNumber
@@ -739,7 +739,7 @@ public class DaemonConfig implements Validated {
     /**
      * The command launched supervisor with worker arguments
      * pid, action and [target_directory]
-     * Where action is - start profile, stop profile, jstack, heapdump and kill against pid
+     * Where action is - start profile, stop profile, jstack, heapdump and kill against pid.
      *
      */
     @isString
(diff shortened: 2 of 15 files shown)


Ethanlm:
Hi @HeartSaVioR @revans2, could you please review this when you get a chance? Thanks very much!

storm-server/src/test/java/org/apache/storm/scheduler/utils/ArtifactoryConfigLoaderTest.java
```diff
@@ -0,0 +1,352 @@
...
+        tc2.prepare(conf);
+        Map ret2 = tc2.load();
+        Assert.assertNotNull("Unexpectedly returned null", ret2);
+        Assert.assertEquals(1, ret2.get("one"));
+        Assert.assertEquals(2, ret2.get("two"));
+        Assert.assertEquals(3, ret2.get("three"));
+    }
+
+    @Test
+    public void testInvalidConfig() throws Exception {
```
HeartSaVioR: Looks like redundant with `testInvalid()`.

storm-server/src/test/java/org/apache/storm/scheduler/utils/ArtifactoryConfigLoaderTest.java
```diff
@@ -0,0 +1,352 @@
...
+            testMap.put("d", 4);
+
+            Yaml yaml = new Yaml();
+            FileWriter fw = new FileWriter(temp);
+            yaml.dump(testMap, fw);
+            fw.flush();
+            fw.close();
+
+            Config config = new Config();
+            config.put(ArtifactoryConfigLoader.ARTIFACTORY_URI, "file://"+temp.getCanonicalPath());
```
HeartSaVioR: both spaces around `+`

storm-server/src/test/java/org/apache/storm/scheduler/utils/ArtifactoryConfigLoaderTest.java
```diff
@@ -0,0 +1,352 @@
...
+            }
+
+            Time.advanceTime(11*60*1000);
+
+            // Now it should
+            result2 = testClass.load();
+
+            Assert.assertNotNull("Unexpectedly returned null", result2);
+
+            // Shouldn't change yet
```
HeartSaVioR: Remove comment.

Ethanlm:
@HeartSaVioR  Thanks for the review.  I will address your comments ASAP but it will take me a little longer to catch up this PR

storm-server/src/main/java/org/apache/storm/scheduler/utils/ArtifactoryConfigLoader.java
```diff
@@ -0,0 +1,416 @@
...
+        }
+        String thisBase = (String)_conf.get(ARTIFACTORY_BASE_DIRECTORY);
+        if (thisBase != null) {
+            _baseDirectory = thisBase;
+        }
+    }
+
+    @Override
+    public Map load() {
+        // Check for new file every so often
```
Ethanlm: This should be fine. "every so often" means "From time to time; occasionally"

storm-server/src/main/java/org/apache/storm/scheduler/utils/ArtifactoryConfigLoader.java
```diff
@@ -0,0 +1,416 @@
...
+        if (thisBase != null) {
+            _baseDirectory = thisBase;
+        }
+    }
+
+    @Override
+    public Map load() {
+        // Check for new file every so often 
+        if (_lastReturnedValue != null && ((Time.currentTimeSecs() - _lastReturnedTime) < _artifactoryPollTimeSecs)) {
+            LOG.debug("returning our last map");
```
Ethanlm: Are you saying to add _lastReturnedTime, Time.currentTimeSecs() and _artifactoryPoolTimeSecs to cache file?  Sorry if I mis-understand your comments

HeartSaVioR: @Ethanlm Ah I think I was not clear on that. I meant adding some contexts to `log`. It will help to debug.

Ethanlm: Yea sure. Thanks for all the comments. I am addressing them.

Ethanlm:
I rebased the branch and addressed some of your comments. I am still working on the rest of them

Ethanlm:
I refactored the code. It simplifies configurations a lot.  `getConfigLoader()`  will choose `IConfigLoader` implementation based on the scheme of `scheduler.config.loader.uri`. It doesn't require users to set many parameters.

`FileConfigLoader` will be used if `file://` scheme is used; `ArtifactoryHttpConfigLoader`  if  `artifactory+http://`. So we can implement `HttpConfigLoader` for normal `http` scheme (TO DO). 

Doing some more testing, especially for `ArtifactoryHttpConfigLoader`. `HttpConfigLoader` is to be implemented. Hope to get some suggestions first if any

HeartSaVioR:
@Ethanlm 
I would like to be clear about `Artifactory` to continue reviewing. I have no idea what is this, but whether it is open specification or proprietary on specific company, if we can provide it to users, we don't worry about HttpConfigLoader. And if we need to define some interfaces for HttpConfigLoader and have to provide the implementation, maybe we could just implement `Artifactory` and let users spin and use it.

Please mention with cc. to proper folk(s) when you think you couldn't answer this question.

Ethanlm:
@HeartSaVioR  You may want to check with this [website](https://www.jfrog.com/artifactory/). My understanding is that we can get most recent file from artifactory server. It's not the same with normal http server.

HeartSaVioR:
@Ethanlm Thanks for the information. Looks like we can't expect users to use artifactory. If possible we'd be better to have HttpConfigLoader implementation, but may be not easy to implement same for artifactory like pulling most recent file. We can start with just getting full URL for the config file, and pull it periodically.

Ethanlm:
@revans2  Do you have any suggestions?  Thanks.

storm-server/src/main/java/org/apache/storm/scheduler/utils/IConfigLoader.java
```diff
@@ -0,0 +1,138 @@
...
+        Map loaderParams = (Map) conf.get(DaemonConfig.SCHEDULER_USER_POOLS_LOADER_PARAMS);
+        if (loaderParams != null) {
+            String uriString = (String) loaderParams.get(SCHEDULER_CONFIG_LOADER_URI);
+            if (uriString != null) {
+                try {
+                    URI uri = new URI(uriString);
+                    String scheme = uri.getScheme();
+                    switch (SchemeType.toSchemeType(scheme)) {
+                        case FILE:
+                            return new FileConfigLoader(loaderParams);
```
revans2: To me it feels like this would be a great place to use a ServiceLoader instead of having a hard coded FileConfigLoader or ArtifactoryHttpConfigLoader.

storm-server/src/main/java/org/apache/storm/scheduler/utils/ArtifactoryHttpConfigLoader.java
```diff
@@ -0,0 +1,375 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.yaml.snakeyaml.Yaml;
+import org.yaml.snakeyaml.constructor.SafeConstructor;
+
+/**
+ * A dynamic loader that can load scheduler configurations for user resource guarantees from Artifactory (an artifact repository manager).
+ */
+public class ArtifactoryHttpConfigLoader implements IConfigLoader {
+    protected static final String ARTIFACTORY_BASE_DIRECTORY = "artifactory.config.loader.base.directory";
```
revans2: If we have configs in here can we tag them with the config validation annotations?  
```
@isSting
```

And also register it with the config service loader?
```
import org.apache.storm.validation.Validated;

...

ArtifactoryHttpConfigLoader implements IConfigLoader, Validated {
```
and then append `org.apache.storm.scheduler.utils.ArtifactoryHttpConfigLoader` to ./storm-server/src/main/resources/META-INF/services/org.apache.storm.validation.Validated

storm-server/src/main/java/org/apache/storm/scheduler/utils/IConfigLoader.java
```diff
@@ -0,0 +1,138 @@
...
+     * It first retrieves parameters like the location of the scheduler config file from the conf;
+     * and then get the actual scheduler configs associated with the configKey from that location.
+     * @param configKey The key from which we want to get the scheduler config.
+     * @return The scheduler configs
+     */
+    Map<?,?> load(String configKey);
+
+    String SCHEDULER_CONFIG_LOADER_URI = "scheduler.config.loader.uri";
+    String SCHEDULER_CONFIG_LOADER_POLLTIME_SECS = "scheduler.config.loader.polltime.secs";
+    String SCHEDULER_CONFIG_LOADER_TIMEOUT_SECS = "scheduler.config.loader.timeout.secs";
```
revans2: Here too can we annotate these with config validation annotations, and make IConfigLoader part of the Config Validation service loader?

storm-server/src/main/java/org/apache/storm/scheduler/utils/IConfigLoader.java
Ethanlm: I have been thinking about using a ServiceLoader and tried so. But I realized that we actually want to have only one type of loader working in the mean time (one loader for one type of scheme). So I think `switch-case` here seems cleaner.   

The way I can think of applying ServiceLoader here is to have every type of loaders check the `scheme` in their `load()` function and then either directly return `null` (because they don't deal with this type of `scheme`) or return the result. Something like:

For HttpConfigLoader
```
Map load(loaderParams) {
    scheme = getScheme();
    if (!scheme.equalsIgnoreCase("http")) {
        return null;
    }
    //else: process and return result;
}
```
For ArtifactoryHttpConfigLoader,
```
Map load(loaderParams) {
    scheme = getScheme();
    if (!scheme.equalsIgnoreCase("artifactory+http")) {
        return null;
    }
    //else: process and return result;
}
```
Is this what you are talking about? Sorry if I mis-understood it.

revans2: Actually I was thinking we might want to use a factory pattern so we can cache connections and other things in the IConfigLoader implementation.  Something perhaps like.

```java
public interface IConfigLoaderFactory {
    IConfigLoader createIfSupported(URI uri, Map<String, Object> conf);

    private static ServiceLoader<IConfigLoaderFactory> serviceLoader = ServiceLoader.load(IConfigLoaderFactory.class);
    
    public static IConfigLoader open(URI uri, Map<String, Object> conf) {
        for (IConfigLoaderFactory factory: serviceLoader) {
            IConfigLoaderFactory ret = factory.createIfSupported(uri, conf);
            if (ret != null) {
                return ret;
            }
        }
        throw some exception();
    }
}

public class HttpArtifactoryServiceLoaderFactory implements IConfigLoaderFactory{
    public IConfigLoader createIfSupported(URI uri, Map<String, Object> conf);
        if (!"artifactory+http".equals(uri.getScheme())) {
            return null;
        }
        return new ...;
    }
}
```

Ethanlm: This seems a great idea to me. I will try to do so. Thanks

HeartSaVioR:
@Ethanlm 
I guess you're handling some different issues concurrently. (at least this and locality aware shuffle grouping) 
If you're not currently working on HttpConfigLoader, I'm OK to file a new issue to address HttpConfigLoader, and keep my +1.

Ethanlm:
@HeartSaVioR  Yes, please. But this patch is not ready yet. I am going to do some tests manually on ArtifactoryHttpConfigLoader and also add some comments to the source code. Thanks.

HeartSaVioR:
@Ethanlm Sure, please leave a comment when you are done. I'll take a look at further changes too.

storm-server/src/main/java/org/apache/storm/scheduler/resource/ResourceAwareScheduler.java
```diff
@@ -42,14 +43,16 @@
...
     private ISchedulingPriorityStrategy schedulingPrioritystrategy;
     private IEvictionStrategy evictionStrategy;
+    private IConfigLoader configLoader;
 
     @Override
     public void prepare(Map<String, Object> conf) {
         this.conf = conf;
         schedulingPrioritystrategy = (ISchedulingPriorityStrategy) ReflectionUtils.newInstance(
-            (String) conf.get(DaemonConfig.RESOURCE_AWARE_SCHEDULER_PRIORITY_STRATEGY));
+                (String) conf.get(DaemonConfig.RESOURCE_AWARE_SCHEDULER_PRIORITY_STRATEGY));
```
jerrypeng: is there a reason why we need to change the spacing in many places in this file?

Ethanlm: No.. It just accidentally happened..

Ethanlm:
I am done with testing and I think this patch is ready for review. Now I have `ArtifactoryConfigLoader` supports both `artifactory+http` and `artifactory+https` schemes.

@HeartSaVioR  @revans2 @jerrypeng  Could you please let me know if you have any suggestions on this patch when you get a chance. Thanks very much!

--------
I want to explain a bit about `Artifactory` and `ArtifactoryConfigLoader` a bit  if you are not familiar with the idea.

To use `ArtifactoryConfigLoader`, we can set the uri pointing to a directory or a file on Artifactory server. 

We can query the REST API `api/storage` to get the metadata of the artifact. If the URI (set at `scheduler.config.loader.uri` ) points to a directory, we can get the metadata like this:
image

We have `children` which are the files under this directory and `ArtifactoryConfigLoader` will pick the file with the largest lexographic name. 

It the URI points to a file, the metadata will be like this:
image

where we can get the "downloadUri" and then download that file.

storm-server/src/main/java/org/apache/storm/DaemonConfig.java
```diff
@@ -792,6 +792,12 @@
     public static final String ISOLATION_SCHEDULER_MACHINES = "isolation.scheduler.machines";
 
     /**
+     * Configuration elements for scheduler config loader.
+     */
+    @isMapEntryType(keyType = String.class, valueType = String.class)
```
revans2: So I made a mistake.  I thought initially that each of the configs in `ConfigLoaderConfiguration` were regular configs under storm.yaml.  I should have read the code more closely, because they actually are sub-configs under "scheduler.config.loader.params".

I don't really like them being sub configs because they don't get validated.  Which was the entire point of me asking you to add them into the Validated ServiceLoader.  I also don't see a lot of value in having

```yaml
scheduler.config.loader.params:
    scheduler.config.loader.uri: "artifactory+http://artifactory.my.company.com:8000/artifactory/configurations/clusters/my_cluster/ras_pools"
    scheduler.config.loader.timeout.sec: 30
```

over 

```yaml
scheduler.config.loader.uri: "artifactory+http://artifactory.my.company.com:8000/artifactory/configurations/clusters/my_cluster/ras_pools"
scheduler.config.loader.timeout.sec: 30
```

Could we change it?

storm-server/src/main/java/org/apache/storm/scheduler/utils/ArtifactoryConfigLoader.java
```diff
@@ -0,0 +1,395 @@
...
+        }
+        String uriString = (String) loaderParams.get(ConfigLoaderConfiguration.SCHEDULER_CONFIG_LOADER_URI);
+        if (uriString == null) {
+            LOG.error("No URI defined in {} configuration.", ConfigLoaderConfiguration.SCHEDULER_CONFIG_LOADER_URI);
+        } else {
+            try {
+                targetURI = new URI(uriString);
+                scheme = targetURI.getScheme().substring(ARTIFACTORY_SCHEME_PREFIX.length());
+            } catch(URISyntaxException e) {
+                LOG.error("Failed to parse uri={}", uriString);
```
revans2: I think we want to throw an exception here.  I don't think we can or should recover from a bad URI.

storm-server/src/main/java/org/apache/storm/scheduler/utils/ArtifactoryConfigLoader.java
```diff
@@ -0,0 +1,416 @@
...
+            LOG.error("got null metadata");
+            return null;
+        }
+        String downloadURI = (String) json.get("downloadUri");
+
+        // This means we are pointing at a file.
+        if (downloadURI != null) {
+            // Then get it and return the file as string.
+            String returnValue = doGet(null, location, host, port);
+            saveInArtifactoryCache(returnValue);
```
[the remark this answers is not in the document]
revans2: `saveInArtifactoryCache` checks for null.  I know it is not as readable here, but it does look like it works.

storm-server/src/main/java/org/apache/storm/DaemonConfig.java
Ethanlm: You are right. I also felt that `scheduler.config.loader.params` makes things unnecessarily complicated. Changing it

storm-server/pom.xml
```diff
@@ -59,6 +59,10 @@
             <groupId>org.apache.commons</groupId>
             <artifactId>commons-compress</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.google.auto.service</groupId>
+            <artifactId>auto-service</artifactId>
+        </dependency>
```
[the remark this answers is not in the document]
Ethanlm: Addressing it.

storm-server/src/main/java/org/apache/storm/scheduler/utils/ArtifactoryConfigLoader.java
Ethanlm: My movitation here is that if the ConfigLoader doesn't work correctly, we will just return null. Then [scheduler ](https://github.com/apache/storm/pull/2199/files#diff-7f8f870479c8f56254f940ea9fc70c92R65)will fall back to use "multitenant-scheduler.xml" file, or the configuration from storm.yaml. But I'm not sure if it's good

Ethanlm:
Tested it again. Should be good now.

storm-server/src/main/java/org/apache/storm/scheduler/utils/ArtifactoryConfigLoader.java
revans2: I personally would prefer to see it fail fast here and not come up if the URI is not formatted correctly.

revans2:
@HeartSaVioR you took a look at this patch before, do you want to take another look at it?

@Ethanlm could you squash your commits into just one?

HeartSaVioR:
@revans2 still +1.
Looks like we can merge this now.

PR #2200, opened 2017-07-10
STORM-2616: Documentation for built in metrics

revans2:
I am happy to pull this back into 1.x branches too.  I am not sure exactly how the metrics might have changed though, so we might want to be careful about doing so.

### DIFF
--- a/docs/Metrics.md
+++ b/docs/Metrics.md
@@ -4,7 +4,7 @@ layout: documentation
 documentation: true
 ---
 Storm exposes a metrics interface to report summary statistics across the full topology.
-It's used internally to track the numbers you see in the Nimbus UI console: counts of executes and acks; average process latency per bolt; worker heap usage; and so forth.
+The numbers you see on the UI come from some of these built in metrics, but are reported through the worker heartbeats instead of through the IMetricsConsumer described below.
 
 ### Metric Types
 
@@ -20,6 +20,7 @@ Storm gives you these metric types:
   - [MeanReducer]({{page.git-blob-base}}/storm-client/src/jvm/org/apache/storm/metric/api/MeanReducer.java) -- track a running average of values given to its `reduce()` method. (It accepts `Double`, `Integer` or `Long` values, and maintains the internal average as a `Double`.) Despite his reputation, the MeanReducer is actually a pretty nice guy in person.
   - [MultiReducedMetric]({{page.git-blob-base}}/storm-client/src/jvm/org/apache/storm/metric/api/MultiReducedMetric.java) -- a hashmap of reduced metrics.
 
+Be aware that even though `getValueAndReset` can return an object returning any object makes it very difficult for an `IMetricsConsumer` to know how to translate it into something usable.  Also note that because it is sent to the `IMetricsConsumer` as a part of a tuple the values returned need to be able to be [serialized](Serialization.html) by your topology.
 
 ### Metrics Consumer
 
@@ -44,7 +45,7 @@ topology.metrics.consumer.register:
     argument: "http://example.com:8080/metrics/my-topology/"
 ```
 
-Storm appends MetricsConsumerBolt to your topology per each registered metrics consumer internally, and each MetricsConsumerBolt subscribes to receive metrics from all tasks. The parallelism for that Bolt is set to `parallelism.hint` and `component id` for that Bolt is set to `__metrics_<metrics consumer class name>`. If you register same class name more than once, postfix `#<sequence number>` is appended to component id.
+Storm appends MetricsConsumerBolt to your topology per each registered metrics consumer internally, and each MetricsConsumerBolt subscribes to receive metrics from all tasks. The parallelism for that Bolt is set to `parallelism.hint` and `component id` for that Bolt is set to `__metrics_<metrics consumer class name>`. If you register the same class name more than once, postfix `#<sequence number>` is appended to component id.
 
 Storm provides some built-in metrics consumers for you to try out to see which metrics are provided in your topology.
 
@@ -75,7 +76,7 @@ Next, let's initialize and register the metric instance.
 ```java
 @Override
 public void prepare(Map conf, TopologyContext context, OutputCollector collector) {
-	// other intialization here.
+	// other initialization here.
 	countMetric = new CountMetric();
 	context.registerMetric("execute_count", countMetric, 60);
 }
@@ -111,7 +112,7 @@ worker.metrics:
 
 or put `Map<String, String>` (metric name, metric class name) with key `Config.TOPOLOGY_WORKER_METRICS` to config map.
 
-There're some restrictions for worker level metric instances: 
+There are some restrictions for worker level metric instances: 
 
 A) Metrics for worker level should be kind of gauge since it is initialized and registered from SystemBolt and not exposed to user tasks.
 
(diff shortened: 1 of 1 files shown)


docs/Metrics.md
```diff
@@ -125,3 +126,193 @@ The [builtin metrics]({{page.git-blob-base}}/storm-client/src/jvm/org/apache/sto
...
+
+This count metrics is bolt specific.  It counts the number of times that a bolt's `execute` method on a bolt was called.
+
+#### Tuple Latency Metrics
+
+Similar to the tuple counting metrics storm also collects average latency metrics for bolts and spouts.  These follow the same structure as the bolt/spout maps and are sub-sampled in the same way as well.  In all cases the latency is measured in milliseconds.
+
+##### `__complete-latency`
+
+The complete latency is just for spouts.  It is the average amount of time it took for `ack` or `fail` to be called for a tuple after it was emitted.  If acking is disabled this metric is likely to be blank or 0 for all values, but should be ignored.
```
srdo: but -> and

docs/Metrics.md
```diff
@@ -125,3 +126,193 @@ The [builtin metrics]({{page.git-blob-base}}/storm-client/src/jvm/org/apache/sto
...
+
+##### `__execute-latency`
+
+This is just for bolts.  It is the average amount of time that the bolt spent in the call to the `execute` method.  The longer this gets the fewer tuples a single bolt instance can process.
```
srdo: "The longer this gets the fewer tuples a single bolt instance can process" -> "The higher this gets, the lower the throughput of tuples per bolt instance"

docs/Metrics.md
```diff
@@ -125,3 +126,193 @@ The [builtin metrics]({{page.git-blob-base}}/storm-client/src/jvm/org/apache/sto
...
+    "overflow": 0,
+    "read_pos": 103445,
+    "write_pos": 103448,
+    "sojourn_time_ms": 2.440771591407277,
+    "capacity": 1024,
+    "population": 19
+}
+```
+
+NOTE that in the `__receive` and `__transfer` queues a single entry may hold 1 or more tuples in it.  For the `__sendqueue` metrics each slot holds a single tuple.  The batching is an optimization that has been in storm since the beginning, so be careful with how you interpret the metrics.  In older versions of storm all of the metrics represent slots in the queue, and not tuples That has been updated so please be careful when trying to compare metrics between different versions of storm.
```
srdo: Wondering if it would be clearer to replace "entry" with "slot" here for consistency. Is NOTE supposed to be all caps? Shouldn't it say "For the `__sendqueue` queue" instead? Missing period after "and not tuples". I had a little trouble understanding the last half of this. Would "In older versions of Storm all of the queue metrics counted slots in the queue. This has been changed in version x.y.z so the queue metrics now always count tuples. Please be careful when trying to compare metrics between different versions of Storm" be accurate? If so I think it's maybe easier to understand.

docs/Metrics.md
```diff
@@ -125,3 +126,193 @@ The [builtin metrics]({{page.git-blob-base}}/storm-client/src/jvm/org/apache/sto
...
+
+`arrival_rate_secs` is an estimation of the number of tuple that are inserted into the queue in one second, although it is actually the dequeue rate.
```
srdo: tuple -> tuples.

docs/Metrics.md
```diff
@@ -125,3 +126,193 @@ The [builtin metrics]({{page.git-blob-base}}/storm-client/src/jvm/org/apache/sto
...
+`capacity` is the total number of slots in the queue. `population` is the number of slots taken in the queue.
+
+#### System Bolt (Worker) Metrics
+
+The System Bolt `__system` provides lots of metrics for different worker wide things.  The one metric not described here is the `__transfer` queue metric, because it fits with the other disruptor metrics described above.
+
+Be aware that the `__system` bolt is an actual bolt so regular bolt metrics described above also will be reported for it.
+
+##### Receive (NettyServer)
+`__recv-iconnection` under the `__system` bolt reports stats for the netty server on the worker.  This is what gets messages from other workers.  It is of the form
```
srdo: Nit: If all these metrics are all for the `__system` bolt, that part of the description can be omitted since they are under the system bolt header.

docs/Metrics.md
```diff
@@ -125,3 +126,193 @@ The [builtin metrics]({{page.git-blob-base}}/storm-client/src/jvm/org/apache/sto
...
+}
+```
+
+The value is a map where the key is a NodeInfo class for the downstream worker it is sending messages to.  This is the SupervisorId + port.  The value is another map with the fields
+
+`src`  What host/port this client has used to connect to the receiving worker.
+`dest` What host/port this client has connected to.
+`reconnects` the number of reconnections that have happened.
+`pending` the number of messages that have not been sent.  (This corresponds to messages, not tuples)
+`sent` the number of messages that have been send.  (This is messages not tuples)
```
srdo: send -> sent

docs/Metrics.md
```diff
@@ -125,3 +126,193 @@ The [builtin metrics]({{page.git-blob-base}}/storm-client/src/jvm/org/apache/sto
...
+    }
+}
+```
+
+`dequeuedMessages` is a throwback to older code where there was an internal queue between the server and the bolts/spouts.  That is no longer the case and the value can be ignored.
+`enqueued` is a map between the address of the remote worker and the number of tuples that were sent from it to this worker.
+
+##### Send (Netty Client)
+
+The `__send-iconnection` metric of the `__system` bolt holds information about all of the clients for this worker that are sending metrics.  It is of the form
```
srdo: Not sure I understand this. Are there worker clients not shown in this list?

docs/Metrics.md
```diff
@@ -125,3 +126,193 @@ The [builtin metrics]({{page.git-blob-base}}/storm-client/src/jvm/org/apache/sto
...
+`src`  What host/port this client has used to connect to the receiving worker.
+`dest` What host/port this client has connected to.
+`reconnects` the number of reconnections that have happened.
+`pending` the number of messages that have not been sent.  (This corresponds to messages, not tuples)
+`sent` the number of messages that have been send.  (This is messages not tuples)
+`lostOnSend`.  This is the number of messages that were lost because of connection issues. (This is messages not tuples). 
+
+##### JVM Memory
+
+The JVM running the worker has several metrics relating to memory usage.  We pull them out of the JVM and report them here.  The relevant metrics are `memory/nonHeap` for off heap memory and `memory/heap` for on heap memory.  These values come from the [MemoryUsage](https://docs.oracle.com/javase/8/docs/api/index.html?java/lang/management/MemoryUsage.html) mxbean for details.
```
srdo: "for details" doesn't fit the rest of the sentence. Also nit: "report them here" -> "report them as metrics"

docs/Metrics.md
```diff
@@ -125,3 +126,193 @@ The [builtin metrics]({{page.git-blob-base}}/storm-client/src/jvm/org/apache/sto
...
+        "dest": "localhost/127.0.0.1:6701",
+        "sent": 420779,
+        "lostOnSend": 0
+    }
+}
+```
+
+The value is a map where the key is a NodeInfo class for the downstream worker it is sending messages to.  This is the SupervisorId + port.  The value is another map with the fields
+
+`src`  What host/port this client has used to connect to the receiving worker.
```
srdo: Maybe this should be a list, so it's rendered one value per line. Github is appending all these to the same line. Same issue in the JVM Memory section, the GC section and the uptime section

docs/Metrics.md
```diff
@@ -125,3 +126,193 @@ The [builtin metrics]({{page.git-blob-base}}/storm-client/src/jvm/org/apache/sto
...
+`sent` the number of messages that have been send.  (This is messages not tuples)
+`lostOnSend`.  This is the number of messages that were lost because of connection issues. (This is messages not tuples). 
+
+##### JVM Memory
+
+The JVM running the worker has several metrics relating to memory usage.  We pull them out of the JVM and report them here.  The relevant metrics are `memory/nonHeap` for off heap memory and `memory/heap` for on heap memory.  These values come from the [MemoryUsage](https://docs.oracle.com/javase/8/docs/api/index.html?java/lang/management/MemoryUsage.html) mxbean for details.
+
+The fields reported are
+
+`maxBytes`	`memUsage.getMax()`
```
srdo: This doesn't render well. Maybe use colon to indicate which names and values belong together?

docs/Metrics.md
```diff
@@ -125,3 +126,193 @@ The [builtin metrics]({{page.git-blob-base}}/storm-client/src/jvm/org/apache/sto
 
 [BuiltinMetricsUtil.java]({{page.git-blob-base}}/storm-client/src/jvm/org/apache/storm/daemon/metrics/BuiltinMetricsUtil.java) sets up data structures for the built-in metrics, and facade methods that the other framework components can use to update them. The metrics themselves are calculated in the calling code -- see for example [`ackSpoutMsg`]({{page.git-blob-base}}/storm-client/src/jvm/org/apache/storm/executor/Executor.java).
 
+#### Reporting Rate
+
+The rate at which built in metrics are reported is configurable through the `topology.builtin.metrics.bucket.size.secs` metric.  If you set this too low it can overload the consumers
+and some metrics consumers expect metrics to show up at a fixed rate or the numbers could be off, so please use caution when modifying this.
```
HeartSaVioR: If bucket.size.secs is configured to be too small, the update to metrics consumers will happen more frequently and it could make some metrics consumers being overloaded.

docs/Metrics.md
```diff
@@ -125,3 +126,193 @@ The [builtin metrics]({{page.git-blob-base}}/storm-client/src/jvm/org/apache/sto
...
+
+
+#### Tuple Counting Metrics
+
+There are several different metrics related to counting what a bolt or spout does to a tuple. These include things like emitting, transferring, acking, and failing of tuples.
+
+In general all of these tuple count metrics are randomly sub-sampled unless otherwise state.  This means that the counts you see both on the UI and from the built in metrics are not necessarily exact.  In fact by default we sample only 5% of the events and estimate the total number of events from that.  The sampling percentage is configurable per topology through the `topology.stats.sample.rate` config.  Setting it to 1.0 will make the counts exact, but be aware that the more events we sample the slower your topology will run (as the metrics are counted on the critical path).  This is why we have a 5% sample rate as the default.
```
HeartSaVioR: I don't think we introduce 'critical path' anywhere in Storm but we already know about that. Isn't it widely-used word? If there's more widely-used representation we may be better to change.

docs/Metrics.md
srdo: Sure, but wouldn't that just be a performance issue? I'm referring to "and some metrics consumers expect metrics to show up at a fixed rate or the numbers could be off", which seems to say that some consumers will report the wrong results if the bucket size is changed(?)

docs/Metrics.md
srdo: Nevermind, I just googled it and the first result is for this meaning. I was just not aware :)

revans2:
@srdo I think I have addressed all of your concerns.  Great suggestions by the way.  Please have another look.

srdo:
Thanks, it looks great. +1

HeartSaVioR:
+1 Nice documentation!

PR #2204, opened 2017-07-12
STORM-1280 port backtype.storm.daemon.logviewer to java

HeartSaVioR:
* port logviewer to Java using Dropwizard
  * follow implementation of DRPC Server
* add '/api/v1/' to the logviewer API endpoints
  * seems like it is not possible (or not easy) to serve both API endpoints and static resources in same servlet
  * reflect the change to the static resources

Please note that logviewer test should be ported to Java since ported version of Logviewer is placed to storm-webapp module. Below is TODO list:

TODO
* [x] port logviewer-test.clj to Java and delete logviewer-test.clj
* [x] delete logviewer.clj and rename package to remove 'wip'
* [x] apply the change for https://github.com/apache/storm/commit/0201ae79a5f4fc7ba7ee1c524818a29a92f140f1#diff-81a8fa5a47d77bd6cca8d98d9916f172 and rebase this to master.
* [x] debug and fix failing tests
* [x] try best to reduce checkstyle violations

I've added J2HTML and jOOL as dependencies for storm-webapp: J2HTML as a replacement for hiccup, and jOOL as Java 8 lambda helper.

Manually tested. Unit test will be covered with porting logviewer-test.clj.

Please let me know deleting logviewer.clj is more convenient to review.


HeartSaVioR:
Also removed unused imports from other spots. I hope this is OK.

HeartSaVioR:
DirectoryCleaner also can be moved to storm-webapp after porting logviewer-test.clj

HeartSaVioR:
UPDATE:
I also ported logviewer-test.clj to Java. Some tests are failing and I'm looking into it.
I still need to apply recent change of logviewer: 
https://github.com/apache/storm/commit/0201ae79a5f4fc7ba7ee1c524818a29a92f140f1#diff-81a8fa5a47d77bd6cca8d98d9916f172.

HeartSaVioR:
Applied recent change of logviewer and rebased.

HeartSaVioR:
Also fixed the tests.

HeartSaVioR:
Removed WIP though I still need to address checkstyle issues. I think check style violation doesn't hurt review process.

HeartSaVioR:
Test is failing with RAT issue.

```
[WARNING] Files with unapproved licenses:
  storm-webapp/src/test/resources/test-3072.log.test
  storm-webapp/src/test/resources/small-worker.log.test
  storm-webapp/src/test/resources/test-worker.log.test
  storm-webapp/src/test/resources/logviewer-search-context-tests.log.test
```

I'll address it after addressing checkstyle.

HeartSaVioR:
Addressed RAT and checkstyle violation.

storm-buildtools/storm_checkstyle.xml
```diff
@@ -46,6 +46,9 @@
  -->
 
 <module name = "Checker">
+    <!-- Filter out Checkstyle warnings that have been suppressed with the @SuppressWarnings annotation -->
```
HeartSaVioR: I just add allowing SuppressWarnings annotation, since the test code contains long string constant. I didn't recognize that maven checkstyle doesn't check test codes, but it might be useful for specific situations anyway.

Ethanlm:
I did some simple tests manually in non-secure and secure storm cluster. It seems that it works well on non-secure storm cluster except some display issues. I found some issues on secure storm cluster. I am not sure if these only happen in my cluster.  Thanks.

1. "Prev" "First" etc buttons seems to mix together.  (On non-secure and secure cluster)
screen shot 2017-07-18 at 1 38 04 pm

2. Got "page not found" while querying worker.log on **secure** cluster. The file actually exists. 
screen shot 2017-07-18 at 1 41 32 pm

3. Got "Request failed" which querying dumps.
screen shot 2017-07-18 at 1 42 01 pm
The error message from logviewer.log:
screen shot 2017-07-18 at 1 42 39 pm

HeartSaVioR:
@Ethanlm 
> 1. "Prev" "First" etc buttons seems to mix together. (On non-secure and secure cluster)

Looks like failing to load css and also js file. We don't include static resource files to the class path, so you need to build binary distribution and use it to test correctly.
(Please let me know if you tested it with binary distribution.)

Btw, it seems to be by intention to avoid adding static resources to the client side. Maybe we can move static files to storm-webapp after porting UI and Logviewer is completed, since storm-webapp module is not intended to be used (as a dependency) for Storm client. It also helps to run and test Logviewer and UI with IDE.

> 2. Got "page not found" while querying worker.log on secure cluster. The file actually exists.

If you could spend some time to help me debugging, could you check the file's permission? 

> 3. Got "Request failed" which querying dumps on secure cluster

Is it occurred only from secure cluster? Looks like actual error is not relevant to secure, so just curious.

To be honest, I didn't have a chance to deploy the patch to secured cluster, since I feel there should be  no difference between non-secure and secure cluster. I might be wrong. I'll also spin secured cluster and test it again.

Ethanlm:
1. I believe I tested it from binary distribution. I used "mvn clean install" to compile and "cd storm-dist/binary && mvn package" to created the binary distribution. And then I deployed it under /tmp directory on my openstack instance. I do it every time when I need to do some tests

2. Will try to do something about it.

3. It happens only on secure cluster. Non-secure cluster works fine in my tests. I could have done something wrong. It would be great if you could test it on your secure cluster.

Thanks.

revans2:
I was playing around with it and I am seeing `&gt;` in the output log. i.e.

```
2017-07-19 15:00:49.413 o.a.s.d.s.Slot SLOT_6701 [INFO] STATE EMPTY msInState: 38110 -&gt; WAITING_FOR_BASIC_LOCALIZATION msInState: 1
```

It looks like we might be double escaping some things.

I'll play around with it some more, and then start looking through the code.

storm-webapp/src/main/java/org/apache/storm/daemon/logviewer/webapp/LogviewerResource.java
```diff
@@ -0,0 +1,268 @@
...
+    @GET
+    @Path("/searchLogs")
+    public Response searchLogs(@Context HttpServletRequest request) throws IOException {
+        String user = httpCredsHandler.getUserName(request);
+        String topologyId = request.getParameter("topoId");
+        String portStr = request.getParameter("port");
+        String callback = request.getParameter("callback");
+        String origin = request.getHeader("Origin");
+
+        return logviewer.listLogFiles(user, portStr != null ? Integer.parseInt(portStr) : null, topologyId, callback, origin);
```
revans2: In the original code we handled an InvalidRequestException. Do we need to do that here too?

storm-webapp/src/main/java/org/apache/storm/daemon/logviewer/webapp/LogviewerResource.java
```diff
@@ -0,0 +1,268 @@
...
+    public Response listLogs(@Context HttpServletRequest request) throws IOException {
+        meterListLogsHttpRequests.mark();
+
+        String user = httpCredsHandler.getUserName(request);
+        String topologyId = request.getParameter("topoId");
+        String portStr = request.getParameter("port");
+        String callback = request.getParameter("callback");
+        String origin = request.getHeader("Origin");
+
+        return logviewer.listLogFiles(user, portStr != null ? Integer.parseInt(portStr) : null, topologyId, callback, origin);
```
revans2: Do we need to handle an InvalidRequestException?

storm-webapp/src/main/java/org/apache/storm/daemon/logviewer/webapp/LogviewerResource.java
```diff
@@ -0,0 +1,268 @@
...
+     */
+    @GET
+    @Path("/download")
+    public Response downloadLogFile(@Context HttpServletRequest request) throws IOException {
+        meterDownloadLogFileHttpRequests.mark();
+
+        String user = httpCredsHandler.getUserName(request);
+        String file = request.getParameter("file");
+        String decodedFileName = URLDecoder.decode(file);
+        return logDownloadHandler.downloadLogFile(decodedFileName, user);
```
revans2: Do we need to handle InvalidRequestException?

storm-webapp/src/main/java/org/apache/storm/daemon/logviewer/webapp/LogviewerResource.java
```diff
@@ -0,0 +1,268 @@
...
+     */
+    @GET
+    @Path("/daemondownload")
+    public Response downloadDaemonLogFile(@Context HttpServletRequest request) throws IOException {
+        meterDownloadLogDaemonFileHttpRequests.mark();
+
+        String user = httpCredsHandler.getUserName(request);
+        String file = request.getParameter("file");
+        String decodedFileName = URLDecoder.decode(file);
+        return logDownloadHandler.downloadDaemonLogFile(decodedFileName, user);
```
revans2: Again InvalidRequestException?

storm-webapp/src/main/java/org/apache/storm/daemon/logviewer/webapp/LogviewerResource.java
```diff
@@ -0,0 +1,268 @@
...
+                               @Context HttpServletRequest request) throws IOException {
+        String user = httpCredsHandler.getUserName(request);
+        String searchString = request.getParameter("search-string");
+        String numMatchesStr = request.getParameter("num-matches");
+        String portStr = request.getParameter("port");
+        String startFileOffset = request.getParameter("start-file-offset");
+        String startByteOffset = request.getParameter("start-byte-offset");
+        String searchArchived = request.getParameter("search-archived");
+        String callback = request.getParameter("callback");
+        String origin = request.getHeader("Origin");
```
revans2: Just an FYI Headers and parameters can be marked with annotations that are pulled out as parameters.  I am not sure if this is cleaner or if that would be.  Bother are rather long.

storm-webapp/src/main/java/org/apache/storm/daemon/logviewer/handler/LogviewerLogPageHandler.java
```diff
@@ -0,0 +1,467 @@
...
+                            }
+                        }
+                    }
+                }
+            }
+        } else {
+            if (port == null) {
+                fileResults = new ArrayList<>();
+
+                File topoDir = new File(logRoot + Utils.FILE_PATH_SEPARATOR + topologyId);
```
revans2: nit:  I know this is an exact translation, but we could just use `new File(logRoot, topologyId)`

HeartSaVioR: Yes it would be better to fix that. I'll address it.

storm-webapp/src/main/java/org/apache/storm/daemon/logviewer/webapp/LogviewerResource.java
HeartSaVioR: Yes agreed. I'm also not sure which is clearer, so I just applied what we did.

storm-webapp/src/main/java/org/apache/storm/daemon/logviewer/webapp/LogviewerResource.java
HeartSaVioR: InvalidRequestException is a checked exception. Clojure seems not check whether the exception class is checked or runtime so placing InvalidRequestException is possible, but from here it doesn't throw InvalidRequestException so can't place it.

storm-webapp/src/main/java/org/apache/storm/daemon/logviewer/webapp/LogviewerResource.java
HeartSaVioR: Same as above.

storm-webapp/src/main/java/org/apache/storm/daemon/logviewer/webapp/LogviewerResource.java
HeartSaVioR: Same as above.

storm-webapp/src/main/java/org/apache/storm/daemon/logviewer/webapp/LogviewerResource.java
HeartSaVioR: Same as above.

storm-webapp/src/main/java/org/apache/storm/daemon/logviewer/LogviewerServer.java
```diff
@@ -0,0 +1,182 @@
...
+        httpServer.join();
+    }
+
+    @Override
+    public synchronized void close() {
+        if (!closed) {
+            //This is kind of useless...
+            meterShutdownCalls.mark();
+
+            //TODO this is causing issues...
```
HeartSaVioR: Actually most of part for LogviewerServer is copy and paste of DRPCServer, and that comment was from there. Could you revisit DRPCServer and address comments?

storm-webapp/src/main/java/org/apache/storm/daemon/logviewer/LogviewerServer.java
```diff
@@ -0,0 +1,182 @@
...
+
+            closed = true;
+        }
+    }
+
+    /**
+     * @return The port the HTTP server is listening on. Not available until {@link #start() } has run.
+     */
+    public int getHttpServerPort() {
+        assert httpServer.getConnectors().length == 1;
```
HeartSaVioR: Same here: this is from copy and paste, and even DRPCServer it seems to be only used for unit test. We could just remove the method and re-add if needed.

storm-webapp/src/main/java/org/apache/storm/daemon/logviewer/LogviewerServer.java
```diff
@@ -0,0 +1,182 @@
...
+            final String httpsKsPassword = (String) (conf.get(DaemonConfig.LOGVIEWER_HTTPS_KEYSTORE_PASSWORD));
+            final String httpsKsType = (String) (conf.get(DaemonConfig.LOGVIEWER_HTTPS_KEYSTORE_TYPE));
+            final String httpsKeyPassword = (String) (conf.get(DaemonConfig.LOGVIEWER_HTTPS_KEY_PASSWORD));
+            final String httpsTsPath = (String) (conf.get(DaemonConfig.LOGVIEWER_HTTPS_TRUSTSTORE_PATH));
+            final String httpsTsPassword = (String) (conf.get(DaemonConfig.LOGVIEWER_HTTPS_TRUSTSTORE_PASSWORD));
+            final String httpsTsType = (String) (conf.get(DaemonConfig.LOGVIEWER_HTTPS_TRUSTSTORE_TYPE));
+            final Boolean httpsWantClientAuth = (Boolean) (conf.get(DaemonConfig.LOGVIEWER_HTTPS_WANT_CLIENT_AUTH));
+            final Boolean httpsNeedClientAuth = (Boolean) (conf.get(DaemonConfig.LOGVIEWER_HTTPS_NEED_CLIENT_AUTH));
+
+            //TODO a better way to do this would be great.
```
HeartSaVioR: The comment was copied and pasted from DRPCServer.

While porting I thought about how to do it better, but no idea given that LogviewerApplication instance is generated dynamically. It may be possible to not passing config map and re-read storm configuration in LogviewerApplication, but it is duplicated effort and we will not be able to inject other config map in unit test. 

Maybe we could just remove the TODO comment.

storm-core/src/clj/org/apache/storm/ui/core.clj
```diff
@@ -134,12 +134,12 @@
 
 (defn logviewer-link [host fname secure?]
   (if (and secure? (*STORM-CONF* LOGVIEWER-HTTPS-PORT))
-    (UIHelpers/urlFormat "https://%s:%s/log?file=%s"
+    (UIHelpers/urlFormat "https://%s:%s/api/v1/log?file=%s"
```
HeartSaVioR: Yes agreed. There're some responses returning HTML page instead of JSON, so we may need to change it first and try to document it.

storm-webapp/src/main/java/org/apache/storm/daemon/logviewer/LogviewerServer.java
revans2: I thought it looked familiar.  I'll file a follow on JIRA to try and clean it up once we have everything ported over to java.

storm-webapp/src/main/java/org/apache/storm/daemon/logviewer/webapp/LogviewerResource.java
revans2: OK Sounds good.

revans2:
Overall it looks good.  At least as an accurate translation of the clojure code.  I want to play around with it some more, and also try to understand what is happening with some of the escaped charters (just to see if it was happening before too).

revans2:
I did a bunch of testing and I am +1 on merging this in.  There is still the issue with double escaping the contents returned for a file, but I think we can fix that after if we want.

HeartSaVioR:
@revans2 Just to check: did you play around secured cluster as well? I still didn't test this on secured cluster, just now set up secured cluster and plan to set up Storm and test.

Ethanlm:
@HeartSaVioR An update about _"(2)Got "page not found" while querying worker.log on secure cluster. The file actually exists."_

I probably have setup something wrong.

I used "mapredqa" (the nimbus admin account and also the topology submitter in my test) to login and I was able to view the log files on logviewer, i.e. no "page not found" issue. 
screen shot 2017-07-20 at 11 37 14 pm

But if I use "ethan" account to login, I am able to see topology summary, component summary, except log files.  "ethan" is a nimbus.user, ui.user, logs.user and topology user of this topology. I got an warning in logviewer.log while querying for the worker.log

screen shot 2017-07-20 at 11 26 07 pm

I don't know the reason yet. I guess there could be something wrong in my cluster settings. It would be great if someone can test it on secure cluster about it.

Note that I was still testing the version before @revans2 's comments.

HeartSaVioR:
@Ethanlm 
Thanks for the update. I'll try to get it today and put my observation.
Btw, my additional commits didn't address anything about bug fix so it doesn't matter which commit you're testing with.

revans2:
@HeartSaVioR I have not tested on a secure cluster at all.  It was all on a single node cluster.

I verified that as many links as I could worked correctly.  I verified I could switch between files, search through files.  I verified that log files were cleaned up as expected as the size grew.

The only other thing that I just noticed is that the pull down for selecting files does not have the current file selected as the default.  (again this can probably be done in a follow on JIRA)

HeartSaVioR:
@revans2 
I fixed the drop down issue, and also fixed issue regarding authorization check. I think it was the reason of issues what @Ethanlm observed.
@Ethanlm 
Could you test it again if you can spend time on this? If you don't have time to test, I'll just merge this after reviewing the latest commit from @revans2 .

Ethanlm:
@HeartSaVioR  I will test it quickly

Ethanlm:
Just tested (3). Don't know why (3) is still not working (even with nimbus admin account). 
image
(querying worker.log as nimbus admin works.)
image

But I think it's fine to fix it later if it's a real problem (if it doesn't only happen on my cluster) .

HeartSaVioR:
@Ethanlm 
Yeah I think I addressed (2) and forgot to debug (3). I'll test it quickly.

HeartSaVioR:
@Ethanlm 
I realized I missed addressing other places for parameter swapped. Just addressed and tested as well.
You can just build `storm-webapp` module and copy `storm-webapp jar` into `lib-webapp` directory.

Ethanlm:
Yea~ It works now. 
Use nimbus admin account: 
image
Use "ethan" account (it's not authorized. Probably that why I got "page-not-found" while querying worker.log): 
image

A quick question.. "ethan" is nimbus.user, ui.user, and logs.user. But it's not able to view worker.log. Which configuration should I set "ethan" to..? Thanks very much.

HeartSaVioR:
@Ethanlm 
For now you should include `ethan` to `logs.users` in `topology configuration`. Global configuration doesn't take effect. This behavior seems same as current logviewer.
If we need to merge global configuration of `logs.users` and `logs.groups` into topology configuration, we can file a new issue and handle it accordingly.

Ethanlm:
@HeartSaVioR Thanks! It's good to learn

HeartSaVioR:
@Ethanlm 
No problem. Thanks for testing my patch thoughtfully. Please let me know if (2) becomes fixed with setting `logs.users` in the topology configuration.

Ethanlm:
Still not authorized.. But I guess it's my problem.. Will try to figure it out. Thanks.

image
image

revans2:
The change looks good to me and I am +1.  Will wait for @Ethanlm to work through his issues first though.

Ethanlm:
@HeartSaVioR @revans2  I think I figured it out. 
I added "ethan@DEREKD.YSTORM.NET" to logs.users and then it works. I traced down to the [isAuthorizedLogUser](https://github.com/apache/storm/pull/2204/files#diff-aad25f432acc108d98b3126bbd914873R74) function and found that it doesn't translate "principal" to "user" so that "ethan@DEREKD.YSTORM.NET" doesn't match with logs.user "ethan" in the configuration. @revans2 and I think the code is identical to the old code. So this should an existing bug.

revans2:
@Ethanlm since you found this could you file a follow on JIRA to use the PrincipalToLocal plugin in the logviewer when authorizing users?

Ethanlm:
@HeartSaVioR  Yes. Sure. Should I wait until this PR being merged in?

HeartSaVioR:
@Ethanlm Great news for me to figure it out. I just added pair of `username` & `username@domain` while testing so didn't recognize the issue.

PR #2216, opened 2017-07-16
#2645 -- Update storm.py to be python3 compatible

rukaya:
Function `emitBolt()` emits a `map`, which is not json serializeble in python3. It should be changed to return a `list` in order to be json serializeble.

### DIFF
--- a/storm-multilang/python/src/main/resources/resources/storm.py
+++ b/storm-multilang/python/src/main/resources/resources/storm.py
@@ -109,7 +109,7 @@ def emitBolt(tup, stream=None, anchors = [], directTask=None):
     m = {"command": "emit"}
     if stream is not None:
         m["stream"] = stream
-    m["anchors"] = map(lambda a: a.id, anchors)
+    m["anchors"] = [a.id for a in anchors]
     if directTask is not None:
         m["task"] = directTask
     m["tuple"] = tup


revans2:
@rukaya 

I am +1 on the change, but we need to have a the change go into master before it goes into any of the 1.x branches.  Do you have a pull request for master that I missed?

srdo:
Hi @rukaya. Thanks for the fix, it looks good. Could you file an issue on jira at https://issues.apache.org/jira and update this PR and the commit message to reflect the issue number?

revans2:
```
diff --git a/storm-multilang/python/src/main/resources/resources/storm.py b/storm-multilang/python/src/main/resources/resources/storm.py
index 9106390b4..224ed120b 100755
--- a/storm-multilang/python/src/main/resources/resources/storm.py
+++ b/storm-multilang/python/src/main/resources/resources/storm.py
@@ -109,7 +109,7 @@ def emitBolt(tup, stream=None, anchors = [], directTask=None):
     m = {"command": "emit"}
     if stream is not None:
         m["stream"] = stream
-    m["anchors"] = map(lambda a: a.id, anchors)
+    m["anchors"] = [a.id for a in anchors]
     if directTask is not None:
         m["task"] = directTask
     m["tuple"] = tup
```

Is what I did to port it to master for testing on master.

rukaya:
@revans2, I just created a pull request from the master branch: https://github.com/apache/storm/pull/2229

rukaya:
@srdo, I just created a jira issue: https://issues.apache.org/jira/browse/STORM-2645. I also updated the commit message.

PR #2218, opened 2017-07-17
STORM-2614: Enhance stateful windowing to persist the window state

arunmahadevan:
Right now the tuples in window are stored in memory. This limits the usage to windows
that fit in memory. Also the source tuples cannot be acked until the window expiry.
By persisting the window transparently in the state backend and caching/iterating them as needed,
we could support larger windows and also support windowed bolts with user/application state.


roshannaik:
Sounds like a checkpoint mechanism limited to windowed bolts. Could you provide a design document for this ?

arunmahadevan:
This builds on top of the existing state checkpointing mechanism (documented here - https://github.com/apache/storm/blob/master/docs/State-checkpointing.md). Theres nothing extra added to the underlying checkpointing mechanism itself and its pretty straightforward. 

The tuples in window (think a FIFO queue) are split into multiple partitions so that they are more manageable and can be distributed/sharded via the underlying key-value state (redis/hbase etc). The modified partitions are saved during a checkpoint. During iteration the partition are loaded on demand from the underlying state backend as they are accessed. A subset of the most recently used partitions are cached in memory (backed by a guava Loading Cache).

During checkpoint, the following are saved :
1. Any modified or newly created window partitions.
2. Any state needed to recover the Trigger/Eviction policies.
3. State thats exposed to the user where the user may have saved some values.

Since the KV state does not guarantee any specific ordering of the keys during iteration, a separate structure is maintained to store the ordered partition Ids which is used during iteration to retrieve the partitions in order. This is also saved during the checkpoint.

The above mechanism kicks in only if user choses to use the windowed state persistence, otherwise the current behavior (keeping the tuples in an in-memory queue) is retained.

I would like to capture the design details in the PR/Documentation itself as far as possible than adding it in separate docs which tend to get lost.

docs/Windowing.md
```diff
@@ -266,3 +266,105 @@ tuples can be received within the timeout period.
...
+                   .withMaxEventsInMemory(25000), // max number of events to be cached in memory
+                    parallelism)
+               .shuffleGrouping("spout");
+
+```
+
+The `withPersistence` instructs the framework to transparently save the tuples in window along with
+any associated system and user state to the state backend. The `withMaxEventsInMemory` is an optional 
+configuration that specifies the maximum number of tuples that may be kept in memory. The tuples are transparently loaded from 
+the state backend as required and the most recently used tuples are retained in memory (backed by a LRU cache).
```
HeartSaVioR: I feel the definition of 'used' in this sentence needs to be cleared. Yes I think caching tuples in memory would be the right way to do, but just not clear we can apply LRU in this case, especially older tuples are occasionally having more chance to be fired.

arunmahadevan: By 'used' I meant accessed. This could be the latest partition where the events are being added or the partitions that were accessed recently during window activation. Yes, its true that the older partitions needs to be accessed before the recent ones while iterating the window. Right now we invalidate any empty partitions during checkpointing. I will see how we can further optimize to keep the older partitions (queue head) cached.

arunmahadevan: @HeartSaVioR, replaced Guava with Caffeine, which has better cache hit rate with sequential access and is based on LFU. I ran some tests with sequential queue scans (one of the typical operations in windowing) and it gave much better results. Caffeine is also api compatible with guava and is already used elsewhere in storm.

roshannaik:
@arunmahadevan  ... could you  put that info on the design in the JIRA itself. easier for future reference.

docs/Windowing.md
HeartSaVioR: @arunmahadevan 
Great news to hear that we've much better results. 
Actually I forgot to leave comment that we can use Caffeine and less depend on Guava, so the replacement is also preferred.

arunmahadevan:
After some more tests (with Caffeine for caching) I found some instances where the current partition (where entries are added or being removed) could get evicted causing issues in serialization. So I added a simple caching with support for pinning that evicts the latest un-pinned partition and based on tests found it to be better and fits the current use case well. @HeartSaVioR , you may want to re-visit the changes.

@roshannaik , added the high level design notes to the JIRA as well.

satishd:
@HeartSaVioR I will go through this PR in next 1-2 days and put my comments.

ben-manes:
fyi, you can emulate pinning by setting the entry's weight to zero. Then the entry will not be evicted, but is eligible for expiration, etc. Of course zero weight entries means the cache allows more entries overall, whereas pinning might still restrict overall capacity. But its not clear how to a cache should handle evictions when not enough space can be freed from unpinned entries.

The other approach is less elegant. A `CacheWriter` can intercept an eviction and a `CacheLoader` intercepts gets to fetch a missing entry. Since recursive computations are disallowed, the writer could re-put the entry asynchronously. This races with a `get`, so the loader could fetch from the map prior to some other data store. In some ways not too dissimilar to your own pinned cache.

In general pinning breaks the caching model of recomputable transient state and the eviction policy will try to avoid pollution by removing low-value entries early. Generally it can lead to O(n) evictions unless maintained in a dedicated LRU queue so that potential victims are not evaluated. Since pinning vs capacity is confusing, right now Caffeine favors zero weights since the use-cases are unclear.

Hope that helps. I agree your own logic may be more straightforward and preferable.

arunmahadevan:
@ben-manes , thanks for the insights. Right now the custom logic is kind of straightforward. May be I will attempt to provide an alternate implementation on top of Caffeine which could be useful in future.

storm-client/src/jvm/org/apache/storm/topology/SimpleWindowPartitionCache.java
```diff
@@ -0,0 +1,182 @@
...
+    @Override
+    public void invalidate(K key) {
+        if (isPinned(key)) {
+            LOG.debug("Entry '{}' is pinned, skipping invalidation", key);
+        } else {
+            LOG.debug("Invalidating entry '{}'", key);
+            V val = map.remove(key);
+            try {
+                lock.lock();
+                --size;
```
ben-manes: if the key isn't in the cache (e.g. concurrent invalidations) then this will decrease incorrectly. I think you need to check `val` and skip the locking code if not present.

arunmahadevan: Yes, thanks for the observation. Refactored the code a bit to address this and a few other race conditions.

docs/Windowing.md
```diff
@@ -266,3 +266,105 @@ tuples can be received within the timeout period.
...
+due to the complex acking and anchoring requirements.
+ 
+To address the above limitations and to support larger window sizes, storm provides stateful windowing support via `IStatefulWindowedBolt`. 
+User bolts should typically extend `BaseStatefulWindowedBolt` for the windowing operations with the framework automatically 
+managing the state of the window in the background.
+
+If the sources provide a monotonically increasing identifier as a part of the message, the framework can use this
+to periodically checkpoint the last expired and evaluated message ids, to avoid duplicate window evaluations in case of 
+failures or restarts. During recovery, the tuples with message ids lower than last expired id are discarded and tuples with 
+message id between the last expired and last evaluated message ids are fed into the system without activating any triggers.
```
arunmahadevan: yes its the trigger policy instances. will reword.

docs/Windowing.md
```diff
@@ -266,3 +266,105 @@ tuples can be received within the timeout period.
...
+topologyBuilder.setBolt("mybolt",
+                   new MyStatefulWindowedBolt()
+                   .withWindow(...) // windowing configuarations
+                   .withMessageIdField("msgid"), // a monotonically increasing 'long' field in the tuple
+                   parallelism)
+               .shuffleGrouping("spout");
+```
+
+However, this option is feasible only if the sources can provide a monotonically increasing identifier in the tuple and the same is maintained
+while re-emitting the messages in case of failures. Here the tuples in window are still held in memory.
```
arunmahadevan: will reword: With this option the tuples are still buffered in memory until processed and expired from the window.

docs/Windowing.md
```diff
@@ -266,3 +266,105 @@ tuples can be received within the timeout period.
...
+
+For more details take a look at the sample topology in storm starter `StatefulWindowingTopology` which will help you get started.
```
arunmahadevan: yes, will link to the file

examples/storm-starter/src/jvm/org/apache/storm/starter/PersistentWindowingTopology.java
```diff
@@ -0,0 +1,176 @@
...
+        @Override
+        public void prepare(Map<String, Object> topoConf, TopologyContext context, OutputCollector collector) {
+            this.collector = collector;
+        }
+
+        @Override
+        public void initState(KeyValueState<String, Pair<Long, Long>> state) {
+            this.state = state;
+            globalAvg = state.get(STATE_KEY, Pair.of(0L, 0L));
+            LOG.info("initState with global avg [" + (double) globalAvg._1 / globalAvg._2 + "]");
```
arunmahadevan: _1 and _2 are exposed to be consistent with other `TupleN` classes. Will use the getters here.

storm-client/src/jvm/org/apache/storm/topology/PersistentWindowedBoltExecutor.java
```diff
@@ -0,0 +1,563 @@
...
+        }
+        super.execute(input);
+        // StatefulBoltExecutor does the actual ack when the state is saved.
+        outputCollector.ack(input);
+    }
+
+    @Override
+    public void initState(T state) {
+        if (stateInitialized) {
+            LOG.warn("State is already initialized. Ignoring initState");
```
arunmahadevan: If a worker crashes, all worker's state are rolled back and an initState message is sent across the topology so that crashed workers can initialize their state. This is already handled in `StatefulBoltExecutor` that wraps the `PersistentWindowedBoltExecutor` so here its Ok to throw an exception. Will address this and revisit the others.

storm-client/src/jvm/org/apache/storm/topology/PersistentWindowedBoltExecutor.java
```diff
@@ -0,0 +1,563 @@
...
+
+    private KeyValueState<String, Optional<?>> getWindowSystemState(Map<String, Object> topoConf, TopologyContext context) {
+        String namespace = context.getThisComponentId() + "-" + context.getThisTaskId() + "-window-systemstate";
+        return (KeyValueState<String, Optional<?>>) StateFactory.getState(namespace, topoConf, context);
+    }
+
+    // a wrapper around the window related states that are checkpointed
+    private static class WindowState<T> extends AbstractCollection<Event<T>> {
+        // number of events per window-partition
+        private static final int PARTITION_SZ = 1000;
```
arunmahadevan: yes

storm-client/src/jvm/org/apache/storm/topology/PersistentWindowedBoltExecutor.java
```diff
@@ -0,0 +1,563 @@
...
+            windowSystemState.rollback();
+        }
+
+        private void initPartitions() {
+            pids = partitionIds.get(PARTITION_IDS_KEY, new LinkedList<>());
+            if (pids.isEmpty()) {
+                pids.add(0L);
+                partitionIds.put(PARTITION_IDS_KEY, pids);
+            } else {
+                latestPartitionId = pids.peekLast();
```
arunmahadevan: yes, it was initialized to 0 being a member variable, anyways I will remove the else to make it clearer.

storm-client/src/jvm/org/apache/storm/topology/PersistentWindowedBoltExecutor.java
```diff
@@ -0,0 +1,563 @@
...
+            long size = maxEventsInMemory / PARTITION_SZ;
+            LOG.info("maxEventsInMemory: {}, partition size: {}, number of partitions: {}",
+                maxEventsInMemory, PARTITION_SZ, size);
+            cache = SimpleWindowPartitionCache.<Long, WindowPartition<T>>newBuilder()
+                .maximumSize(size)
+                .removalListener(new RemovalListener<Long, WindowPartition<T>>() {
+                    @Override
+                    public void onRemoval(Long pid, WindowPartition<T> p, RemovalCause removalCause) {
+                        LOG.debug("onRemoval for id '{}', WindowPartition '{}'", pid, p);
+                        if (pid == null || p == null) {
```
arunmahadevan: not possible now, will change it to asserts.

storm-client/src/jvm/org/apache/storm/topology/SimpleWindowPartitionCache.java
```diff
@@ -0,0 +1,191 @@
...
+
+        return val;
+    }
+
+    private V load(K key) {
+        V val = map.get(key);
+        if (val == null) {
+            val = cacheLoader.load(key);
+            if (val == null) {
+                throw new NullPointerException("Null value");
```
arunmahadevan: good point.

storm-client/src/jvm/org/apache/storm/topology/SimpleWindowPartitionCache.java
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    private void pin(K key) {
+        LOG.debug("pin '{}'", key);
+        pinned.compute(key, (k, v) -> v == null ? 1L : v + 1);
+        LOG.debug("pinned '{}'", pinned);
+    }
+
+    private boolean isPinned(K key) {
+        return pinned.getOrDefault(key, 0L) > 0;
```
arunmahadevan: containsKey should work here since we are removing the entry in `unpin` if the value <= 0, but prefer to leave it as is to be defensive against future code refactoring.

storm-client/src/jvm/org/apache/storm/topology/WindowedBoltExecutor.java
```diff
@@ -102,6 +109,18 @@ private int getMaxSpoutPending(Map<String, Object> topoConf) {
         return maxPending;
     }
 
+    private int getCheckpointIntervalSecs(Map<String, Object> topoConf) {
```
arunmahadevan: it can be refactored

storm-client/src/jvm/org/apache/storm/topology/WindowedBoltExecutor.java
```diff
@@ -127,26 +154,30 @@ private void validate(Map<String, Object> topoConf, Count windowLengthCount, Dur
...
 
-        if (windowLengthDuration != null && slidingIntervalDuration != null) {
-            ensureDurationLessThanTimeout(windowLengthDuration.value + slidingIntervalDuration.value, topologyTimeout);
-        } else if (windowLengthDuration != null) {
-            ensureDurationLessThanTimeout(windowLengthDuration.value, topologyTimeout);
-        } else if (slidingIntervalDuration != null) {
-            ensureDurationLessThanTimeout(slidingIntervalDuration.value, topologyTimeout);
-        }
+        if (isPersistent()) {
+            ensureCheckpointIntervalLessThanTimeout(getCheckpointIntervalSecs(topoConf), topologyTimeout);
```
arunmahadevan: we don't need to validate `message timeout > window length + sliding interval` here since the tuples are acked during checkpoint. Hence ensuring the `timeout > checkpoint interval` would suffice.

storm-client/src/jvm/org/apache/storm/windowing/TupleWindowIterImpl.java
```diff
@@ -0,0 +1,80 @@
...
+    @Override
+    public List<Tuple> get() {
+        List<Tuple> tuples = new ArrayList<>();
+        tuplesIt.get().forEachRemaining(t -> tuples.add(t));
+        return tuples;
+    }
+
+    @Override
+    public List<Tuple> getNew() {
+        throw new UnsupportedOperationException("Not implemented");
```
arunmahadevan: it not straightforward to return the new and expired tuples iterators without persisting the tuples separately and maintaining the state, so for now its unsupported. The newTuplesIt and expiredTuplesIt receives null values for now

arunmahadevan:
@srdo, addressed your comments, let me know if I missed something.

storm-client/src/jvm/org/apache/storm/topology/WindowPartitionCache.java
```diff
@@ -0,0 +1,142 @@
...
+    V get(K key);
+
+    /**
+     * Get value from the cache or load the value pinning it
+     * so that the entry will never get evicted.
+     *
+     * @param key the key
+     * @return the value
+     */
+    V getPinned(K key);
```
satishd: This method name can be changed to `pinAndGet` as `getPinned` was implies getting the value if it is pinned.

storm-client/src/jvm/org/apache/storm/topology/SimpleWindowPartitionCache.java
```diff
@@ -0,0 +1,191 @@
...
+        if (shouldPin) {
+            try {
+                lock.lock();
+                val = load(key);
+                pin(key);
+            } finally {
+                lock.unlock();
+            }
+        } else {
+            val = map.get(key);
```
satishd: Why are we checking here as this is already checked in `load(key)`?

docs/Windowing.md
```diff
@@ -266,3 +266,108 @@ tuples can be received within the timeout period.
...
+    
+    // emit the results downstream
+    collector.emit(new Values(result));
+  }
+}
+```
+
+Note: In case of persistent windowed bolts, use `TupleWindow.getIter` to retrieve an iterator over the 
+events in the window. If the number of tuples in windows is huge, invoking `TupleWindow.get` would
+try to load all the tuples into memory and may throw an OOM exception.
```
satishd: Do we also want to say `getNew` and `getExpired` are not supported with persistent windowed bolts?

storm-client/src/jvm/org/apache/storm/topology/SimpleWindowPartitionCache.java
```diff
@@ -0,0 +1,191 @@
...
+import java.util.concurrent.ConcurrentSkipListMap;
+import java.util.concurrent.locks.ReentrantLock;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A simple implementation that evicts the largest un-pinned entry from the cache. This works well
+ * for caching window partitions since the access pattern is mostly sequential scans.
+ */
+public class SimpleWindowPartitionCache<K, V> implements WindowPartitionCache<K, V> {
```
satishd: It is good to have unit tests, including with multiple threads.

storm-client/test/jvm/org/apache/storm/topology/PersistentWindowedBoltExecutorTest.java
```diff
@@ -0,0 +1,261 @@
...
+        Mockito.verify(mockOutputCollector, Mockito.times(5)).ack(tupleCaptor.capture());
+        Assert.assertArrayEquals(mockTuples.toArray(), tupleCaptor.getAllValues().toArray());
+
+        Mockito.doAnswer(new Answer<Void>() {
+            @Override
+            public Void answer(InvocationOnMock invocation) throws Throwable {
+                TupleWindow window = (TupleWindow) invocation.getArguments()[0];
+                // iterate the tuples
+                Assert.assertEquals(5, window.get().size());
+                // iterating multiple times should produce same events
```
srdo: This feels like it belongs in a test of the iterator implementation

storm-client/test/jvm/org/apache/storm/topology/PersistentWindowedBoltExecutorTest.java
```diff
@@ -0,0 +1,261 @@
...
+            }
+        }).when(mockBolt).execute(Mockito.any());
+        // trigger the window
+        long activationTs = tupleTs + 1000;
+        executor.getWindowManager().add(new WaterMarkEvent<>(activationTs));
+        executor.prePrepare(0);
+
+        // partition ids
+        Mockito.verify(mockPartitionState, Mockito.times(1)).put(stringCaptor.capture(), partitionValuesCaptor.capture());
+        Assert.assertEquals(PARTITION_KEY, stringCaptor.getAllValues().get(0));
```
srdo: There's a getValue you can use instead I think. You already know there is exactly one value due to the verify above

storm-client/test/jvm/org/apache/storm/topology/PersistentWindowedBoltExecutorTest.java
```diff
@@ -0,0 +1,261 @@
...
+        Assert.assertEquals(Collections.singletonList(0L), partitionValuesCaptor.getAllValues().get(0));
+
+        // window partitions
+        Mockito.verify(mockWindowState, Mockito.times(1)).put(longCaptor.capture(), windowValuesCaptor.capture());
+        Assert.assertEquals(0L, (long) longCaptor.getAllValues().get(0));
```
srdo: Nit: Consider naming these values

storm-client/test/jvm/org/apache/storm/topology/PersistentWindowedBoltExecutorTest.java
```diff
@@ -0,0 +1,261 @@
...
+        Assert.assertEquals(5, windowValuesCaptor.getAllValues().get(0).size());
+        List<Event<?>> tuples = windowValuesCaptor.getAllValues().get(0)
+            .getEvents().stream().map(Event::get).collect(Collectors.toList());
+        Assert.assertArrayEquals(mockTuples.toArray(), tuples.toArray());
+
+        // window system state
+        Mockito.verify(mockSystemState, Mockito.times(2)).put(stringCaptor.capture(), systemValuesCaptor.capture());
+        Assert.assertEquals(EVICTION_STATE_KEY, stringCaptor.getAllValues().get(1));
```
srdo: Nit: Reusing the captor like this is pretty brittle when someone needs to add extra verification to this test. Consider just declaring the captors as local variables and use a captor per verify. I don't think you gain anything by using [at]Captor for non-generic classes anyway.

storm-client/test/jvm/org/apache/storm/topology/PersistentWindowedBoltExecutorTest.java
```diff
@@ -0,0 +1,261 @@
...
+    @Test
+    public void testCacheEviction() {
+        Mockito.when(mockWaterMarkEventGenerator.track(Mockito.any(GlobalStreamId.class), Mockito.anyLong())).thenReturn(true);
+        executor.initState(null);
+        executor.waterMarkEventGenerator = mockWaterMarkEventGenerator;
+        List<Tuple> mockTuples = getMockTuples(20000);
+        mockTuples.forEach(t -> executor.execute(t));
+
+        Mockito.verify(mockWindowState, Mockito.times(10)).put(longCaptor.capture(), windowValuesCaptor.capture());
+        Assert.assertEquals(10, longCaptor.getAllValues().size());
```
srdo: You already know this is the case because of the verify

storm-client/test/jvm/org/apache/storm/topology/PersistentWindowedBoltExecutorTest.java
```diff
@@ -0,0 +1,261 @@
...
+        Assert.assertEquals(10, windowValuesCaptor.getAllValues().size());
+        // number of evicted events
+        Assert.assertEquals(10_000, windowValuesCaptor.getAllValues().stream()
```
srdo: I think you should make MAX_PARTITION_EVENTS and MIN_PARTITIONS public so you can refer to them here. Much of this test might also fit better into a WindowCache test

storm-client/test/jvm/org/apache/storm/topology/PersistentWindowedBoltExecutorTest.java
```diff
@@ -0,0 +1,261 @@
...
+            .mapToInt(x -> x.size()).sum());
+
+        Map<Long, PersistentWindowedBoltExecutor.WindowPartition<Event<Tuple>>> partitionMap = new HashMap<>();
+        windowValuesCaptor.getAllValues().forEach(v -> partitionMap.put(v.getId(), v));
+
+        Mockito.verify(mockPartitionState, Mockito.times(20)).put(stringCaptor.capture(), partitionValuesCaptor.capture());
```
srdo: Also refer to the constants here. 20 is pretty magical at first glance

storm-client/test/jvm/org/apache/storm/topology/PersistentWindowedBoltExecutorTest.java
```diff
@@ -0,0 +1,261 @@
...
+public class PersistentWindowedBoltExecutorTest {
+    public static final String LATE_STREAM = "late_stream";
+    private PersistentWindowedBoltExecutor<KeyValueState<String, String>> executor;
+    private IStatefulWindowedBolt<KeyValueState<String, String>> mockBolt;
+    private OutputCollector mockOutputCollector;
+    private TopologyContext mockTopologyContext;
+    private Map<String, Object> mockStormConf = new HashMap<>();
+    private TimestampExtractor mockTimestampExtractor;
+    private WaterMarkEventGenerator mockWaterMarkEventGenerator;
+    private KeyValueState mockPartitionState;
```
srdo: Nit: Could probably avoid rawtypes complaints by using [at]Mock similar to what you've done with Captor

storm-client/test/jvm/org/apache/storm/topology/PersistentWindowedBoltExecutorTest.java
```diff
@@ -0,0 +1,261 @@
...
+        Assert.assertEquals(10, windowValuesCaptor.getAllValues().size());
+        // number of evicted events
+        Assert.assertEquals(10_000, windowValuesCaptor.getAllValues().stream()
+            .mapToInt(x -> x.size()).sum());
+
+        Map<Long, PersistentWindowedBoltExecutor.WindowPartition<Event<Tuple>>> partitionMap = new HashMap<>();
+        windowValuesCaptor.getAllValues().forEach(v -> partitionMap.put(v.getId(), v));
+
+        Mockito.verify(mockPartitionState, Mockito.times(20)).put(stringCaptor.capture(), partitionValuesCaptor.capture());
+        Assert.assertEquals(LongStream.range(0, 20).boxed().collect(Collectors.toList()), partitionValuesCaptor.getAllValues().get(19));
```
srdo: I can't tell why 0...20 is the expected value here

storm-client/test/jvm/org/apache/storm/topology/PersistentWindowedBoltExecutorTest.java
```diff
@@ -0,0 +1,261 @@
...
+                partitionMap.put((long)args[0], (PersistentWindowedBoltExecutor.WindowPartition<Event<Tuple>>)args[1]);
+                return null;
+            }
+        }).when(mockWindowState).put(Mockito.any(), Mockito.any());
+
+        // trigger the window
+        long activationTs = tupleTs + 1000;
+        executor.getWindowManager().add(new WaterMarkEvent<>(activationTs));
+
+        Mockito.verify(mockBolt, Mockito.times(4000)).execute(Mockito.any());
```
srdo: Why 4000?

storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
```diff
@@ -2363,7 +2363,7 @@ private CommonTopoInfo getCommonTopoInfo(String topoId, String operation) throws
             ret.launchTimeSecs = 0;
         }
         ret.assignment = state.assignmentInfo(topoId, null);
-        ret.beats = Utils.OR(heartbeatsCache.get().get(topoId), Collections.emptyMap());
+        ret.beats = Utils.<Map<List<Integer>, Map<String, Object>>>OR(heartbeatsCache.get().get(topoId), Collections.emptyMap());
```
srdo: Why is this necessary?

storm-client/src/jvm/org/apache/storm/windowing/WindowManager.java
```diff
@@ -111,14 +125,86 @@ public void add(Event<T> windowEvent) {
...
+                    return inner.next();
+                }
+                throw new IllegalStateException("Stale iterator");
+            }
+        };
+    }
+
+    private boolean doOnTriggerStateful() {
+        Supplier<Iterator<T>> scanEventsStateful = this::scanEventsStateful;
+        Iterator<T> it = scanEventsStateful.get();
```
srdo: Why isn't it necessary to lock while accessing this iterator?

storm-client/src/jvm/org/apache/storm/windowing/TupleWindowIterImpl.java
```diff
@@ -0,0 +1,80 @@
...
+                               Long startTimestamp, Long endTimestamp) {
+        this.tuplesIt = tuplesIt;
+        this.newTuplesIt = newTuplesIt;
+        this.expiredTuplesIt = expiredTuplesIt;
+        this.startTimestamp = startTimestamp;
+        this.endTimestamp = endTimestamp;
+    }
+
+    @Override
+    public List<Tuple> get() {
```
srdo: Doesn't this put the entire window in memory?

storm-client/src/jvm/org/apache/storm/windowing/WindowManager.java
```diff
@@ -111,14 +125,86 @@ public void add(Event<T> windowEvent) {
...
+        boolean hasEvents = it.hasNext();
+        if (hasEvents) {
+            final IteratorStatus status = new IteratorStatus();
+            LOG.debug("invoking windowLifecycleListener onActivation with iterator");
+            // reuse the retrieved iterator
+            Supplier<Iterator<T>> wrapper = new Supplier<Iterator<T>>() {
+                Iterator<T> initial = it;
+                @Override
+                public Iterator<T> get() {
+                    if (status.isValid()) {
```
srdo: Could you explain how this mechanism works?

storm-client/src/jvm/org/apache/storm/topology/SimpleWindowPartitionCache.java
```diff
@@ -0,0 +1,191 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger(SimpleWindowPartitionCache.class);
+
+    private final ConcurrentSkipListMap<K, V> map = new ConcurrentSkipListMap<>();
+    private final Map<K, Long> pinned = new HashMap<>();
+    private final long maximumSize;
+    private final RemovalListener<K, V> removalListener;
+    private final CacheLoader<K, V> cacheLoader;
+    private final ReentrantLock lock = new ReentrantLock(true);
+    private int size;
```
satishd: `ConcurrentSkipList#size` is not a constant time operation and it may need to traverse all of them and its value also may change when size() is invoked.

srdo: Thanks for explaining, this makes sense then.

storm-client/src/jvm/org/apache/storm/topology/base/BaseStatefulWindowedBolt.java
```diff
@@ -151,6 +158,38 @@
...
+
+    /**
+     * The maximum number of window events to keep in memory. This is meaningful only if
+     * {@link #withPersistence()} is also set. As the number of events in memory grows close
+     * to the maximum, the events that are less likely to be used again are evicted and persisted.
+     * The default value for this is {@code 1,000,000}.
+     *
+     * @param maxEventsInMemory the maximum number of window events to keep in memory
+     */
+    public BaseStatefulWindowedBolt<T> withMaxEventsInMemory(long maxEventsInMemory) {
```
arunmahadevan: Yes, it would be a good option to provide in future. Will raise a followup JIRA.

storm-client/src/jvm/org/apache/storm/topology/PersistentWindowedBoltExecutor.java
```diff
@@ -0,0 +1,596 @@
...
+        return (KeyValueState<String, Deque<Long>>) StateFactory.getState(namespace, topoConf, context);
+    }
+
+    private KeyValueState<String, Optional<?>> getWindowSystemState(Map<String, Object> topoConf, TopologyContext context) {
+        String namespace = context.getThisComponentId() + "-" + context.getThisTaskId() + "-window-systemstate";
+        return (KeyValueState<String, Optional<?>>) StateFactory.getState(namespace, topoConf, context);
+    }
+
+    // a wrapper around the window related states that are checkpointed
+    private static class WindowState<T> extends AbstractCollection<Event<T>> {
```
arunmahadevan: can be refactored.

storm-client/src/jvm/org/apache/storm/topology/WindowPartitionCache.java
arunmahadevan: makes sense.

storm-client/src/jvm/org/apache/storm/topology/SimpleWindowPartitionCache.java
arunmahadevan: An optimization to avoid the lock if the entry is already present.

docs/Windowing.md
arunmahadevan: will do.

storm-client/src/jvm/org/apache/storm/topology/PersistentWindowedBoltExecutor.java
```diff
@@ -0,0 +1,596 @@
...
+        }
+
+        void rollback() {
+            partitionIdsState.rollback();
+            windowPartitionsState.rollback();
+            windowSystemState.rollback();
+        }
+
+        private void initPartitions() {
+            partitionIds = partitionIdsState.get(PARTITION_IDS_KEY, new LinkedList<>());
```
arunmahadevan: Thanks for noticing, yes this needs to be handled.

storm-client/src/jvm/org/apache/storm/topology/PersistentWindowedBoltExecutor.java
```diff
@@ -0,0 +1,596 @@
...
+                    }
+                );
+        }
+
+        private void deletePartition(long pid) {
+            LOG.debug("Delete partition: {}", pid);
+            try {
+                partitionIdsLock.lock();
+                partitionIds.remove(pid);
+                partitionIdsState.put(PARTITION_IDS_KEY, partitionIds);
```
arunmahadevan: this is to save the initial list of partition ids `[0]` to the state.

storm-client/src/jvm/org/apache/storm/topology/PersistentWindowedBoltExecutor.java
```diff
@@ -0,0 +1,596 @@
...
+                if (pidToInvalidate != null) {
+                    cache.invalidate(pidToInvalidate);
+                }
+            });
+            windowSystemStateSupplier.get().forEach(windowSystemState::put);
+        }
+    }
+
+    // the window partition that holds the events
+    public static class WindowPartition<T> implements Iterable<Event<T>> {
```
arunmahadevan: Will refactor.

storm-client/src/jvm/org/apache/storm/topology/PersistentWindowedBoltExecutor.java
```diff
@@ -0,0 +1,596 @@
...
+        private final ConcurrentLinkedQueue<Event<T>> events = new ConcurrentLinkedQueue<>();
+        private final AtomicInteger size = new AtomicInteger();
```
arunmahadevan: because the `ConcurrentLinkedQueue.size` is not a constant time operation

storm-client/src/jvm/org/apache/storm/topology/PersistentWindowedBoltExecutor.java
```diff
@@ -0,0 +1,596 @@
...
+        }
+
+        void setModified() {
+            if (!modified) {
+                modified = true;
+            }
+        }
+
+        void clearModified() {
+            if (modified) {
```
arunmahadevan: yes, since its not too frequent its not required.

storm-client/src/jvm/org/apache/storm/windowing/WindowManager.java
```diff
@@ -53,28 +61,34 @@
...
-    private TriggerPolicy<T> triggerPolicy;
-
+    private final boolean stateful;
+    private EvictionPolicy<T, ?> evictionPolicy;
+    private TriggerPolicy<T, ?> triggerPolicy;
     public WindowManager(WindowLifecycleListener<T> lifecycleListener) {
+        this(lifecycleListener, new ConcurrentLinkedQueue<>(), false);
+    }
+
+    public WindowManager(WindowLifecycleListener<T> lifecycleListener, Collection<Event<T>> queue, boolean stateful) {
```
arunmahadevan: Ok.

storm-client/src/jvm/org/apache/storm/windowing/WindowManager.java
```diff
@@ -111,14 +125,86 @@ public void add(Event<T> windowEvent) {
...
+            compactWindow();
+        }
     }
 
     /**
      * The callback invoked by the trigger policy.
      */
     @Override
     public boolean onTrigger() {
+        return stateful ? doOnTriggerStateful() : doOnTrigger();
```
arunmahadevan: Strategy is not straightforward without a lot of refactoring. Will move it into sub class to keep the stateful case separate.

storm-client/src/jvm/org/apache/storm/windowing/WindowManager.java
```diff
@@ -289,4 +414,22 @@ public String toString() {
                 ", triggerPolicy=" + triggerPolicy +
                 '}';
     }
+
+    public void restoreState(Map<String, Optional<?>> state) {
+        if (state != null) {
```
arunmahadevan: appears that its not required now.

storm-client/src/jvm/org/apache/storm/windowing/WatermarkCountEvictionPolicy.java
```diff
@@ -17,20 +17,28 @@
...
+import java.util.concurrent.atomic.AtomicLong;
+
 /**
  * An eviction policy that tracks count based on watermark ts and
  * evicts events up to the watermark based on a threshold count.
  *
  * @param <T> the type of event tracked by this policy.
  */
-public class WatermarkCountEvictionPolicy<T> extends CountEvictionPolicy<T> {
+public class WatermarkCountEvictionPolicy<T> implements EvictionPolicy<T, Pair<Long, Long>> {
```
arunmahadevan: Theres some code duplication, but the logic is not the same and delegation might not be correct here. Also it doesn't look very straightforward without a lot of refactoring.  Would like to keep it this way for now.

storm-client/test/jvm/org/apache/storm/topology/PersistentWindowedBoltExecutorTest.java
arunmahadevan: thanks for the input, yes `@Mock` can be used.

storm-client/test/jvm/org/apache/storm/topology/PersistentWindowedBoltExecutorTest.java
arunmahadevan: Am pulling out the WindowState into a top level class and adding additional tests. Maybe this can still be left as is.

storm-client/test/jvm/org/apache/storm/topology/PersistentWindowedBoltExecutorTest.java
arunmahadevan: yes, can use getValue

storm-client/test/jvm/org/apache/storm/topology/PersistentWindowedBoltExecutorTest.java
arunmahadevan: thanks, can use local captor variables here.

storm-client/test/jvm/org/apache/storm/topology/PersistentWindowedBoltExecutorTest.java
arunmahadevan: Here its checking if the `PersistentWindowedBoltExecutor` is writing the evicted values out to the state. Will add separate unit tests for WindowState.

storm-client/test/jvm/org/apache/storm/topology/PersistentWindowedBoltExecutorTest.java
arunmahadevan: will add a comment. its the partition ids `0 .. 19`

storm-client/test/jvm/org/apache/storm/topology/PersistentWindowedBoltExecutorTest.java
arunmahadevan: 20,000 events and a count window of 5. Will address by naming the constants.

storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
arunmahadevan: I am getting the below error when I do a `mvn clean install` in the storm-server module without the type hints. Will check again.

```
[ERROR] COMPILATION ERROR :
[INFO] -------------------------------------------------------------
[ERROR] /Users/aiyer/github/storm/storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java:[2366,29] incompatible types: inferred type does not conform to upper bound(s)
    inferred: java.util.Map<? extends java.lang.Object,? extends java.lang.Object>
    upper bound(s): java.util.Map<java.util.List<java.lang.Integer>,java.util.Map<java.lang.String,java.lang.Object>>,java.lang.Object

$ java -version
java version "1.8.0_51"
Java(TM) SE Runtime Environment (build 1.8.0_51-b16)
Java HotSpot(TM) 64-Bit Server VM (build 25.51-b03, mixed mode)
```

storm-client/src/jvm/org/apache/storm/windowing/WindowManager.java
arunmahadevan: The locks are required in `onTrigger` due to `compactEvents`. Here its not required since compact is not invoked. I am refactoring this into sub class.

storm-client/src/jvm/org/apache/storm/windowing/WindowManager.java
arunmahadevan: The iterator is invalidated after returning from the activation callback. Basically bolts are not supposed to hold a reference to the iterator obtained via `Window.getIter` after it returns from `execute` (say to iterate the values later). The window events can expire and some internal state is cleared so its not feasible and not very meaningful to do this.

storm-client/src/jvm/org/apache/storm/windowing/TupleWindowIterImpl.java
arunmahadevan: yes, its mentioned in the doc. Will add a Javadoc comment.

storm-client/src/jvm/org/apache/storm/topology/PersistentWindowedBoltExecutor.java
srdo: Isn't that guaranteed to happen in initPartitions further up?

storm-client/src/jvm/org/apache/storm/topology/PersistentWindowedBoltExecutor.java
srdo: Sorry, I didn't consider that it was volatile. I'm guessing it's a performance optimization?

storm-client/src/jvm/org/apache/storm/windowing/WatermarkCountEvictionPolicy.java
srdo: Okay

storm-client/src/jvm/org/apache/storm/windowing/WindowManager.java
srdo: Makes sense, thanks. Could you expand the IllegalStateException to mention this? Why is the iterator being updated in line 191 btw?

storm-client/src/jvm/org/apache/storm/topology/PersistentWindowedBoltExecutor.java
arunmahadevan: Ok I wrongly assumed the comment was for `initPartitions`, so discard my previous comment. 
Here we first update the `partitionIds` member variable. The put is to save the value into the kv-state. The partitionIdsState state has a single entry something like `"key" -> List of ids`.

storm-client/src/jvm/org/apache/storm/topology/PersistentWindowedBoltExecutor.java
arunmahadevan: In `clearModified` its probably ok to skip the check since its invoked only during flush. 

`setModified` is invoked in `add` so we can use it to avoid frequent writes to the volatile.

storm-client/src/jvm/org/apache/storm/windowing/WindowManager.java
arunmahadevan: Will update the message. 
Within execute, a bolt may invoke `get`/`getIter` multiple times. So here we have a supplier and return a new iterator each time.

arunmahadevan:
@srdo , @satishd,  thanks for the review. Pushed some changes to address the latest review comments.

storm-client/test/jvm/org/apache/storm/windowing/persistence/WindowStateTest.java
```diff
@@ -0,0 +1,247 @@
...
+            it.next();
+        }
+
+        for (int i = 0; i < numEvents; i++) {
+            Event<Integer> event = getEvent(i);
+            expected.add(event);
+            ws.add(event);
+        }
+
+        // 9 th partition should not have been evicted
```
srdo: Isn't this the 10th?

storm-client/test/jvm/org/apache/storm/windowing/persistence/WindowStateTest.java
```diff
@@ -0,0 +1,247 @@
...
+    @Mock
+    private KeyValueState<String, Optional<?>> systemState;
+    @Mock
+    private Supplier<Map<String, Optional<?>>> supplier;
+    @Captor
+    private ArgumentCaptor<Long> longCaptor;
+    @Captor
+    private ArgumentCaptor<WindowState.WindowPartition<Integer>> windowValuesCaptor;
+
+    private static final int MAX_EVENTS_PER_PARTITION = 1000;
```
srdo: Isn't this the same as the constant in WindowState? If so I think you should refer to that one instead of duplicating it.

storm-client/test/jvm/org/apache/storm/windowing/persistence/WindowStateTest.java
```diff
@@ -0,0 +1,247 @@
...
+        long numEvents = partitions * MAX_EVENTS_PER_PARTITION;
+        List<Event<Integer>> expected = new ArrayList<>();
+        for (int i = 0; i < numEvents; i++) {
+            Event<Integer> event = getEvent(i);
+            expected.add(event);
+            ws.add(event);
+        }
+
+        Iterator<Event<Integer>> it = ws.iterator();
+        for(int i=0; i<9500; i++) {
```
srdo: I think it would be helpful to put a note here about this causing the partition to be pinned. At a glance it's not obvious why it should work this way.

storm-client/src/jvm/org/apache/storm/windowing/persistence/WindowState.java
```diff
@@ -0,0 +1,422 @@
...
+                    LOG.debug("Updating modified partition {}", pid);
+                    p.clearModified();
+                    windowPartitionsState.put(pid, p);
+                }
+            } finally {
+                windowPartitionsLock.unlock(pid);
+            }
+            // invalidate after releasing the lock
+            if (pidToInvalidate != null) {
+                cache.invalidate(pidToInvalidate);
```
srdo: If this id is pinned when this happens, when is it removed?

storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
srdo: I'd guess it's a type inference error in your JDK. Could you try updating and see if that fixes it? I know some of the earlier Java 8 releases had some bugs in this regard.

storm-client/test/jvm/org/apache/storm/topology/SimpleWindowPartitionCacheTest.java
```diff
@@ -0,0 +1,232 @@
...
+                .build(key -> {
+                    loaded.add(key);
+                    return key;
+                });
+
+        cache.get(1);
+        cache.get(2);
+        cache.get(3);
+        Assert.assertEquals(Arrays.asList(1, 2, 3), loaded);
+        Assert.assertEquals(Collections.singletonList(2), removed);
```
arunmahadevan: Since 2 is the largest un-pinned entry before 3 is loaded. The cache evicts the largest unpinned entry.

storm-client/test/jvm/org/apache/storm/windowing/persistence/WindowStateTest.java
```diff
@@ -0,0 +1,247 @@
...
+        Mockito.when(windowState.get(Mockito.any(), Mockito.any())).then(returnsArgAt(1));
+
+        WindowState<Integer> ws = getWindowState(10 * MAX_EVENTS_PER_PARTITION);
+
+        long partitions = 15;
+        long numEvents = partitions * MAX_EVENTS_PER_PARTITION;
+        for (int i = 0; i < numEvents; i++) {
+            ws.add(getEvent(i));
+        }
+        // 5 events evicted to window state
```
arunmahadevan: yes

storm-client/test/jvm/org/apache/storm/windowing/persistence/WindowStateTest.java
arunmahadevan: right.

storm-client/src/jvm/org/apache/storm/windowing/persistence/WindowState.java
arunmahadevan: it will be removed when (1) that partition gets unpinned and the entry is evicted from the cache due to loading other entries or (2) the next time the flush happens and the partition is empty.

storm-client/test/jvm/org/apache/storm/topology/SimpleWindowPartitionCacheTest.java
```diff
@@ -0,0 +1,232 @@
...
+                    return new Object();
+                });
+
+        FutureTask<Object> ft1 = new FutureTask<>(() -> cache.pinAndGet(1));
+        FutureTask<Object> ft2 = new FutureTask<>(() -> cache.pinAndGet(1));
+        Thread t1 = new Thread(ft1);
+        Thread t2 = new Thread(ft2);
+        t1.start();
+        t2.start();
+        t1.join();
```
arunmahadevan: Added

storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
arunmahadevan: Yes thanks, after upgrading the java 8 to the latest release it appears to work. Removed the type hint.

arunmahadevan:
@srdo , thanks for taking time to do an in-depth review. Have addressed your latest comments.

HeartSaVioR:
@arunmahadevan
+1 Nice work.
Could you squash the commits into one? You can merge it by yourself, or please mention me if you want me to merge.

arunmahadevan:
Thanks for the reviews, squashed and merged to master.

PR #2219, opened 2017-07-17
[STORM-2637] fix ClassCastException in logviewer get-log-user-group-whitelist func…

Ethanlm:
In get-log-user-group-whitelist function, the getLogMetaDataFile function returns a File while the readYamlFile function requires a String.

see https://issues.apache.org/jira/browse/STORM-2637

### DIFF
--- a/storm-core/src/clj/org/apache/storm/daemon/logviewer.clj
+++ b/storm-core/src/clj/org/apache/storm/daemon/logviewer.clj
@@ -311,7 +311,7 @@
 
 (defn get-log-user-group-whitelist [fname]
   (let [wl-file (ServerConfigUtils/getLogMetaDataFile fname)
-        m (clojurify-structure (Utils/readYamlFile wl-file))]
+        m (clojurify-structure (Utils/readYamlFile (.toString wl-file)))]
     (if (not-nil? m)
       (do
         (let [user-wl (.get m LOGS-USERS)


storm-core/src/clj/org/apache/storm/daemon/logviewer.clj
```diff
@@ -311,7 +311,7 @@
 
 (defn get-log-user-group-whitelist [fname]
   (let [wl-file (ServerConfigUtils/getLogMetaDataFile fname)
-        m (clojurify-structure (Utils/readYamlFile wl-file))]
+        m (clojurify-structure (Utils/readYamlFile (.toString wl-file)))]
```
HeartSaVioR: Maybe better to use `getAbsolutePath` or `getCanonicalPath` to clearly represent path. I used `getAbsolutePath` from #2204 but I can change it if others claim that `getCanonicalPath` is more preferred.

HeartSaVioR:
Thanks for the patch. I just would like to see this patch against 1.x branch rather than master, given that we should review and apply #2204 to remove Clojure implementation. 
Btw, if you had (or have) a chance to read logviewer clojure implementation, I would like to ask a favor with reviewing #2204.

Ethanlm:
Thanks. I will submit a patch against 1.x branch.  I am not very familiar with logviewer clojure implementation yet. But I will try and see what I can do.

Ethanlm:
Close this PR.

PR #2222, opened 2017-07-18
[STORM-2133] add page-rendered-at timestamp on the UI

Ethanlm:
As a user, I would like a simple timestamp of when a UI page was rendered, so that, I can quickly check if the information is fresh.
This is mainly for the case of keeping browser tabs open for a long while during debugging. Seeing old data in your browser could harm the effort.    -Reported by @d2r 

see: https://issues.apache.org/jira/browse/STORM-2133

This PR adds timestamp on index, topology, supervisor, owner, component pages, etc. I didn't add timestamp on logviewer pages since logs have clear timestamps.  Please let me know if it's also needed. Thanks.

### DIFF
--- a/storm-core/src/ui/public/component.html
+++ b/storm-core/src/ui/public/component.html
@@ -74,6 +74,9 @@
       <p id="toggle-switch" style="display: block;" class="js-only"></p>
     </div>
   </div>
+  <div>
+    <p id="page-rendered-at-timestamp"></p>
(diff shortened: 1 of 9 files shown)


Ethanlm:
The timestamp at the bottom of the page.

screen shot 2017-07-18 at 12 29 02 pm

HeartSaVioR:
+1

PR #2223, opened 2017-07-18
STORM-2541: Fix manual partition assignment (1.x)

srdo:
Same as https://github.com/apache/storm/pull/2150, but for 1.x.


srdo:
The only change from the master version is replacing String.join with StringUtils.join in the topic filters.

HeartSaVioR:
+1
I'm OK to break the API if it didn't work anyway.

srdo:
Thanks for reviewing. The test failure was an unrelated bit of flakiness in the integration test.

PR #2224, opened 2017-07-19
STORM-2640: Deprecate KafkaConsumer.subscribe API option, make KafkaConsumer.assign the default

srdo:
This is the 1.x subtask of https://issues.apache.org/jira/browse/STORM-2542

The first commit is from https://github.com/apache/storm/pull/2223. I'd like us to switch to using manual partitioning in 1.2.0 as the default rather than waiting for 2.0, since the subscribe API seems to cause people issues. It's not something I feel strongly about though, so if it's too big of a change for a minor version release, let me know and I'll revert it.

### DIFF
--- a/docs/storm-kafka-client.md
+++ b/docs/storm-kafka-client.md
@@ -247,12 +247,9 @@ streams.  If you are doing this for Trident a value must be in the List returned
 otherwise trident can throw exceptions.
 
 
-### Manual Partition Control (ADVANCED)
+### Manual Partition Assigment (ADVANCED)
 
-By default Kafka will automatically assign partitions to the current set of spouts.  It handles lots of things, but in some cases you may want to manually assign the partitions.
-This can cause less churn in the assignments when spouts go down and come back up, but it can result in a lot of issues if not done right.  This can all be handled by subclassing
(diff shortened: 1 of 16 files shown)


hmcl:
@srdo since the dynamic partition assignment does not work anyways, and since Storm 2.0 is a major release, I wonder if we shouldn't simply remove these classes. It is not backwards compatible, but if it doesn't really work, what's the harm? I am not a big fan of, from very early releases, leaving deprecated code around, which just becomes a source of confusion and harder to maintain. At the very least we should plan a future release to remove it for good.

hmcl:
The commit with title "Cherry pick STORM-2542" should have the same title as in master. We should also incorporate the Java 7 compatibility changes in the commit itself. Otherwise it's really hard to track which commits really belong to the JIRA.

srdo:
@hmcl The deprecated classes are removed in https://github.com/apache/storm/pull/2151, this is a 1.x PR to deprecate them before they get removed. 

I thought the PR might be easier to review with some separate commits. I'll squash down to one commit with the PR's title once it has been reviewed.

hmcl:
+1

PR #2225, opened 2017-07-19
STORM-2641: Make storm.py print output from subprocess on error on Wi…

srdo:
…ndows, make storm.ps1 return the right exit code


revans2:
The test failure looks unrelated.

PR #2229, opened 2017-07-19
STORM-2645 -- Update storm.py to be python3 compatible

rukaya:
Function `emitBolt()` emits a `map`, which is not json serializeble in python3. It should be changed to return a `list` in order to be json serializeble.


HeartSaVioR:
+1

HeartSaVioR:
Btw I guess you filed STORM-2645. Then could you update the PR title as having STORM-2645 as prefix? Same thing applies to the commit title.

srdo:
+1

PR #2230, opened 2017-07-19
[STORM-2644] Show message when result not found on deep search page

Ethanlm:
Deep search shows a blank area if no results are found. see https://issues.apache.org/jira/browse/STORM-2644

image

Show some message to avoid confusion about whether the searching is finished or not.

### DIFF
--- a/storm-core/src/ui/public/deep_search_result.html
+++ b/storm-core/src/ui/public/deep_search_result.html
@@ -97,11 +97,26 @@ $(document).ready(function() {
             file.count = count;
             file.logviewerPort = logviewerPort;
             file.search_archived = search_archived;
+            if (file.matches == 0) {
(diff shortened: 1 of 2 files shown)


Ethanlm:
image

knusbaum:
I would prefer *one* message if nothing is found. When there are hundreds of workers, this would make users scroll through many "Result not found" messages before getting to results that *are* returned.

Ethanlm:
Thanks. Will fix it

Ethanlm:
Addressed comment from @knusbaum . 
Tested some simple cases on firefox, chrome and safari. The following screenshots are from testing on firefox.
onlyon6702
onlyon6703
notanywhere
onbothworkers

HeartSaVioR:
+1

PR #2231, opened 2017-07-20
Add link to issue tracker on Readme page.

ecararus:


HeartSaVioR:
+1

README.markdown
```diff
@@ -33,6 +33,9 @@ You can subscribe to this list by sending an email to [dev-subscribe@storm.apach
 
 You can also [browse the archives of the storm-dev mailing list](http://mail-archives.apache.org/mod_mbox/storm-dev/).
 
+### Issue tracker
+In case you want to rise a bug/feature or propose an idea, please use [Apache Jira](https://issues.apache.org/jira/projects/STORM)
```
srdo: "rise" should be "raise" I think.

Ethanlm:
Thanks for the patch. I am not very sure if this is necessary since there is already a [link](https://github.com/apache/storm#documentation) pointing to [Developer documentation](https://github.com/apache/storm/blob/master/DEVELOPER.md) where there are more instructions about how to ["report a bug"](https://github.com/apache/storm/blob/master/DEVELOPER.md#report-a-bug) etc, which seems better to me.

HeartSaVioR:
@Ethanlm 
Yes you're right about that. I missed. 
2 cents for me is that non-contributors might not would want to open DEVELOPER.md so maybe better to move only `report a bug` section to README, and keep other sections.

Ethanlm:
@HeartSaVioR  Agreed.

revans2:
@ecararus I pulled this into master, and fixed the one minor spelling comment (its documentation so we don't technically need a vote for it).  Thanks for fixing this.  Keep up the good work.

PR #2232, opened 2017-07-20
STORM-2650: Expand property substitution test for Flux to cover non-s…

srdo:
…tring properties

See https://issues.apache.org/jira/browse/STORM-2650


HeartSaVioR:
+1 great to add the test.

ptgoetz:
+1

PR #2234, opened 2017-07-21
STORM-2652: fix error in open method of JmsSpout

omerhadari:
I open this PR since I was delighted to see that there is an existing JMS spout implementation, but encountered a bug when trying to use it. The bug caused an exception (described below) to be thrown when trying to set up (`open`) the spout, since the value of `Config.TOPOLOGY_MESSAGE_TIMEOUT_SECS` corresponds to a `Long`, and not to an `Integer`.

This is a minor change, hope it's ok :)

```
7317 [Thread-18-a-executor[2 2]] ERROR o.a.s.util - Async loop died!
java.lang.ClassCastException: java.lang.Long cannot be cast to java.lang.Integer
	at org.apache.storm.jms.spout.JmsSpout.open(JmsSpout.java:175) ~[storm-jms-1.1.0.jar:1.1.0]
	at org.apache.storm.daemon.executor$fn__4976$fn__4991.invoke(executor.clj:600) ~[storm-core-1.1.0.jar:1.1.0]
	at org.apache.storm.util$async_loop$fn__557.invoke(util.clj:482) [storm-core-1.1.0.jar:1.1.0]
	at clojure.lang.AFn.run(AFn.java:22) [clojure-1.7.0.jar:?]
	at java.lang.Thread.run(Thread.java:748) [?:1.8.0_131]
7320 [Thread-18-a-executor[2 2]] ERROR o.a.s.d.executor - 
java.lang.ClassCastException: java.lang.Long cannot be cast to java.lang.Integer
	at org.apache.storm.jms.spout.JmsSpout.open(JmsSpout.java:175) ~[storm-jms-1.1.0.jar:1.1.0]
	at org.apache.storm.daemon.executor$fn__4976$fn__4991.invoke(executor.clj:600) ~[storm-core-1.1.0.jar:1.1.0]
	at org.apache.storm.util$async_loop$fn__557.invoke(util.clj:482) [storm-core-1.1.0.jar:1.1.0]
	at clojure.lang.AFn.run(AFn.java:22) [clojure-1.7.0.jar:?]
	at java.lang.Thread.run(Thread.java:748) [?:1.8.0_131]
```

PR #2236, opened 2017-07-23
STORM-2652: fix error in open method of JmsSpout

omerhadari:
I changed the casting to `Number` instead of `Integer`
since it is actually a `Long`, and it caused a `ClassCastException`
to be thrown. Now it is referred to as a `Number` so that it won't
happen again.
Also resolved style issues in `JmsSpout`. Added javadoc, renamed
fields etc...

This is essentially the same as #2234 but compatible with the 1.x-branch

@srdo


srdo:
Thanks

PR #2238, opened 2017-07-24
[STORM-2654] preserve the formatting of the exception stacktrace on UI

Ethanlm:
See https://issues.apache.org/jira/browse/STORM-2654
The stack trace is not formatted on UI.

image


Ethanlm:
Changes:
image

PR #2239, opened 2017-07-24
[STORM-2655] fix: log users cannot view worker.log on Storm UI on sec…

Ethanlm:
See: https://issues.apache.org/jira/browse/STORM-2655

solution: add princialToLocal inside of isAuthorizedLogUser function.


Ethanlm:
Tested it manually.
image

HeartSaVioR:
+1

Maybe better to address this also for 1.x branch as well.

revans2:
@HeartSaVioR the biggest issue with addressing this on the 1.x branch is that the patch uses both a JDK8 specific feature (a default method in an interface) and breaks backwards compatibility with other IPrincipalToLocal plugins.

We could address this by having a static method that will create a principal that wraps the user name instead and passes that into the original PrincipalToLocal instance.  But it is not a clean port.

PR #2240, opened 2017-07-25
[STORM-2657] Update SECURITY.MD

liu-zhaokun:
[https://issues.apache.org/jira/browse/STORM-2657](https://issues.apache.org/jira/browse/STORM-2657)

There are several PRs about storm acl merged into master,so I update SECURITY.md.

### DIFF
--- a/docs/SECURITY.md
+++ b/docs/SECURITY.md
@@ -478,6 +478,44 @@ nimbus.groups:
  
 
 ### DRPC
-Hopefully more on this soon
+ 
+ Storm provides the Access Control List for the DRPC Authorizer.Users can see [org.apache.storm.security.auth.authorizer.DRPCSimpleACLAuthorizer](javadocs/org/apache/storm/security/auth/authorizer/DRPCSimpleACLAuthorizer.html) for more details.
+ 
+ There are several DRPC ACL related configurations.
+ 
+ | YAML Setting | Description |
+ |------------|----------------------|
+ | drpc.authorizer.acl | A class that will perform authorization for DRPC operations. Set this to org.apache.storm.security.auth.authorizer.DRPCSimpleACLAuthorizer when using security.|
+ | drpc.authorizer.acl.filename | This is the name of a file that the ACLs will be loaded from. It is separate from storm.yaml to allow the file to be updated without bringing down a DRPC server. Defaults to drpc-auth-acl.yaml |
+ | drpc.authorizer.acl.strict| It is useful to set this to false for staging where users may want to experiment, but true for production where you want users to be secure. Defaults to false. |
+
+ The file pointed to by drpc.authorizer.acl.filename will have only one config in it drpc.authorizer.acl this should be of the form
+
+ drpc.authorizer.acl:
+   "functionName1":
+     "client.users":
+       - "alice"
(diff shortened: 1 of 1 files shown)


liu-zhaokun:
@revans2 
Hello,I have updated SECURITY.MD,review it,please.

liu-zhaokun:
Can one of the admins verify this patch?

docs/SECURITY.md
```diff
@@ -478,6 +478,35 @@ nimbus.groups:
  
 
 ### DRPC
-Hopefully more on this soon
+ 
+ Storm provides the Access Control List for the DRPC Authorizer.Users can see org.apache.storm.security.auth.authorizer.DRPCSimpleACLAuthorizer for more details.
```
revans2: It would be great if we could turn this into a link.

```
[org.apache.storm.security.auth.authorizer.DRPCSimpleACLAuthorizer](javadocs/org/apache/storm/security/auth/authorizer/DRPCSimpleACLAuthorizer.html)
```

Should work.

docs/SECURITY.md
```diff
@@ -478,6 +478,35 @@ nimbus.groups:
...
+ 
+ There are several DRPC ACL related configurations.
+ 
+ | YAML Setting | Description |
+ |------------|----------------------|
+ | drpc.authorizer.acl | The class for DRPC ACL. |
+ | drpc.authorizer.acl.filename | File name of the DRPC Authorizer ACL.It should be set to "drpc-auth-acl.yaml",users can see drpc-auth-acl.yaml.example for more details. |
+ | drpc.authorizer.acl.strict| Whether the DRPCSimpleAclAuthorizer should deny requests for operations involving functions that have no explicit ACL entry. |
```
revans2: This is kind of confusing, and that is because the config is kind of confusing.  Some configs go in the main storm.yaml

| YAML Setting | Description |
|------------|----------------------|
| drpc.authorizer | A class that will perform authorization for DRPC operations.  Set this to `org.apache.storm.security.auth.authorizer.DRPCSimpleACLAuthorizer` when using security. |
| drpc.authorizer.acl.strict | Whether the DRPCSimpleAclAuthorizer should deny requests for operations involving functions that have no explicit ACL entry. It is useful to set this to false for staging where users may want to experiment, but true for production where you want users to be secure. Defaults to false. |
| drpc.authorizer.acl.filename | This is the name of a file that the ACLs will be loaded from.  It is separate from storm.yaml to allow the file to be updated without bringing down a DRPC server. Defaults to drpc-auth-acl.yaml |

The file pointed to by `drpc.authorizer.acl.filename` will have only one config in it `drpc.authorizer.acl` this should be of the form 

```yaml
drpc.authorizer.acl:
  "functionName1":
    "client.users":
      - "alice"
      - "bob"
    "invocation.user": "bob"
```

In this the users `bob` and `alice` as `client.users` are allowed to run DRPC requests against functionName1, but only `bob` as the `invocation.user` is allowed to run the topology that actually processes those requests.

liu-zhaokun:
@revans2 
Hello,I have modified this PR followed your suggestion.Thanks again for your hard work.Review it,Please.

PR #2242, opened 2017-07-25
[STORM-1347] display topology version on UI

Ethanlm:
Display topology version on UI.

see: https://issues.apache.org/jira/browse/STORM-1347

### DIFF
--- a/storm-client/src/jvm/org/apache/storm/Config.java
+++ b/storm-client/src/jvm/org/apache/storm/Config.java
@@ -170,6 +170,12 @@ public class Config extends HashMap<String, Object> {
     public static final String TOPOLOGY_DEBUG = "topology.debug";
 
     /**
+     * User defined version of this topology
+     */
+    @isString
+    public static final String TOPOLOGY_VERSION = "topology.version";
+
+    /**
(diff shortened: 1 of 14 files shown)


Ethanlm:
image

HeartSaVioR:
Actually this issue is the subtask of topology online update, and topology version is expected to be set automatically to the increasing number.
But I also think this patch doesn't hurt in any way. When we support online update and have version, we can just overwrite `topology.version` only if user doesn't set custom version string.
Moreover, supporting version as label looks like good idea indeed.

+1 on this. Thanks @Ethanlm

Ethanlm:
Thanks for review @HeartSaVioR

HeartSaVioR:
@Ethanlm 
Sorry there's merge conflict now. Could you resolve this?
You may want to just regenerate classes from thrift since the conflict occurs from generated files.

Ethanlm:
Thanks @HeartSaVioR  Regenerated them

Ethanlm:
storm-cassandra failed. Should be not related to this PR.

HeartSaVioR:
+1

PR #2243, opened 2017-07-25
STORM-2658: Extract storm-kafka-client examples to storm-kafka-client-examples

srdo:
See https://issues.apache.org/jira/browse/STORM-2658.

The changes to the Trident examples mainly have to do with the command line arguments not being consistently passed to the topologies, e.g. the broker url was passed to producing topologies but not the consuming topology. The extra parameters have been removed, I think the example code is fine with hard coded topic names.


examples/storm-kafka-client-examples/README.markdown
```diff
@@ -0,0 +1,10 @@
+## Usage
+This module contains example topologies demonstrating storm-kafka-client spout and Trident usage.
+
+The module is built by `mvn clean package`. This will generate the `target/storm-kafka-client-examples-VERSION.jar` file. The jar contains all dependencies and can be submitted to Storm via the Storm CLI. For example:
+```
+storm jar storm-kafka-client-examples-2.0.0-SNAPSHOT.jar org.apache.storm.kafka.spout.test.KafkaSpoutTopologyMainNamedTopics
+```
+will submit the topologies set up by KafkaSpoutTopologyMainNamedTopics to Storm.
+
+Note that this example produces a jar containing all dependencies for ease of use. In a production environment you may want to reduce the jar size by extracting some dependencies (e.g. org.apache.kafka:kafka-clients) from the jar. You can do this by setting the dependencies you don't want to include in the jars to `provided` scope, and then manually copying the dependencies to your Storm extlib directory.
```
HeartSaVioR: Instead of copying dependencies to the extlib, you can achieve the same thing (or more) via using `--artifacts` to add dependencies for specific topology while submitting. I think this is simpler and topology-wide, so would love to guide both, or only `--artifacts`. (We already replaced the guide for how to add dependencies from Storm SQL.)

Please refer https://github.com/apache/storm/blob/master/docs/Command-line-client.md#jar for details.

examples/storm-kafka-client-examples/pom.xml
```diff
@@ -42,7 +42,9 @@
             <groupId>org.apache.storm</groupId>
             <artifactId>storm-kafka</artifactId>
             <version>${project.version}</version>
+            <!-- You can reduce jar size by uncommenting this and putting dependencies in $STORM-HOME/extlib instead of including them in the jar
```
HeartSaVioR: Same above. Maybe we can let users choose how they provide dependencies, change the sentence to `...uncommenting this and providing dependencies manually.` or so. My intention is that we don't recommend putting dependencies to extlib directory, unless they know what they're doing (affecting whole topologies' dependencies)

examples/storm-kafka-client-examples/pom.xml
```diff
@@ -73,19 +75,27 @@
             <groupId>org.apache.storm</groupId>
             <artifactId>storm-kafka-client</artifactId>
             <version>${project.version}</version>
+            <!-- You can reduce jar size by uncommenting this and putting dependencies in $STORM-HOME/extlib instead of including them in the jar
```
HeartSaVioR: Same here.

examples/storm-kafka-client-examples/pom.xml
```diff
@@ -73,19 +75,27 @@
...
             <scope>${provided.scope}</scope>
+            -->
         </dependency>
         <dependency>
             <groupId>org.apache.kafka</groupId>
             <artifactId>${storm.kafka.artifact.id}</artifactId>
             <version>${storm.kafka.client.version}</version>
+            <scope>compile</scope>
+            <!-- You can reduce jar size by uncommenting this and putting dependencies in $STORM-HOME/extlib instead of including them in the jar
```
HeartSaVioR: Same here.

examples/storm-kafka-client-examples/pom.xml
```diff
@@ -73,19 +75,27 @@
...
             <scope>${provided.scope}</scope>
+            -->
         </dependency>
         <dependency>
             <groupId>org.apache.kafka</groupId>
             <artifactId>kafka-clients</artifactId>
             <version>${storm.kafka.client.version}</version>
+            <scope>compile</scope>
+            <!-- You can reduce jar size by uncommenting this and putting dependencies in $STORM-HOME/extlib instead of including them in the jar
```
HeartSaVioR: Same here.

examples/storm-kafka-client-examples/README.markdown
srdo: Thanks, I didn't know about this flag. It's much better, will replace references to extlib.

HeartSaVioR:
+1

examples/storm-kafka-client-examples/README.markdown
```diff
@@ -0,0 +1,10 @@
+## Usage
+This module contains example topologies demonstrating storm-kafka-client spout and Trident usage.
+
+The module is built by `mvn clean package`. This will generate the `target/storm-kafka-client-examples-VERSION.jar` file. The jar contains all dependencies and can be submitted to Storm via the Storm CLI. For example:
```
hmcl: ... built running ... the Storm CLI, e.g.:

examples/storm-kafka-client-examples/pom.xml
```diff
@@ -42,7 +42,9 @@
             <groupId>org.apache.storm</groupId>
             <artifactId>storm-kafka</artifactId>
             <version>${project.version}</version>
+            <!-- You can reduce jar size by uncommenting this and providing the dependencies manually. See the README for details.
             <scope>${provided.scope}</scope>
```
hmcl: Isn't the goal of ${provided.scope} to handle the proper scope according to the profile, e.g. just like it is done with the Intellij profile. I am not quite following why the user has to comment/uncomment the scope configuration.

examples/storm-kafka-client-examples/pom.xml
```diff
@@ -73,19 +75,27 @@
             <groupId>org.apache.storm</groupId>
             <artifactId>storm-kafka-client</artifactId>
             <version>${project.version}</version>
+            <!-- You can reduce jar size by uncommenting this and providing the dependencies manually. See the README for details.
             <scope>${provided.scope}</scope>
```
hmcl: Isn't the goal of ${provided.scope} to handle the proper scope according to the profile, e.g. just like it is done with the Intellij profile. I am not quite following why the user has to comment/uncomment the scope configuration.

examples/storm-kafka-client-examples/src/main/java/org/apache/storm/kafka/trident/TridentKafkaClientWordCountNamedTopics.java
```diff
@@ -42,72 +42,68 @@
...
-            StormSubmitter.submitTopology(topic1 + "-producer", tpConf, KafkaProducerTopology.newTopology(brokerUrl, topic1));
-            StormSubmitter.submitTopology(topic2 + "-producer", tpConf, KafkaProducerTopology.newTopology(brokerUrl, topic2));
-            // Consumer
-            StormSubmitter.submitTopology("topics-consumer", tpConf, TridentKafkaConsumerTopology.newTopology(newKafkaTridentSpoutOpaque()));
-
-            // Print results to console, which also causes the print filter in the consumer topology to print the results in the worker log
-            Thread.sleep(2000);
-            DrpcResultsPrinter.remoteClient().printResults(60, 1, TimeUnit.SECONDS);
-        }
+    protected void run(String[] args) throws AlreadyAliveException, InvalidTopologyException,
```
hmcl: why remove the ability to specify the topic name from the command line ?

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/builders/SingleTopicKafkaSpoutConfiguration.java
```diff
@@ -21,15 +21,10 @@
 import static org.apache.storm.kafka.spout.KafkaSpoutConfig.FirstPollOffsetStrategy.EARLIEST;
 
 import org.apache.kafka.clients.consumer.ConsumerConfig;
```
hmcl: I would call package where this class lives `config.builder` instead of `builders`, which is a bit misleading since this is really a configuration class.

examples/storm-kafka-client-examples/src/main/java/org/apache/storm/kafka/spout/test/KafkaSpoutTopologyMainNamedTopics.java
```diff
@@ -0,0 +1,103 @@
...
+    private static final String KAFKA_LOCAL_BROKER = "localhost:9092";
+    private static final String TOPIC_0 = "kafka-spout-test";
+    private static final String TOPIC_1 = "kafka-spout-test-1";
+    private static final String TOPIC_2 = "kafka-spout-test-2";
+
+    public static void main(String[] args) throws Exception {
+        new KafkaSpoutTopologyMainNamedTopics().runMain(args);
+    }
+
+    protected void runMain(String[] args) throws Exception {
```
hmcl: Isn't this change removing the ability to run this code in LocalCluster mode? I think it is very useful. For example, I use it all the time to run these simple test examples from IntelliJ.

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/builders/SingleTopicKafkaSpoutConfiguration.java
hmcl: I also would call the two getXyz methods in this class createXyz, as they are static factory methods. I know that the name was already like that, but since we are changing it, we should just make it more conventional.

examples/storm-kafka-client-examples/pom.xml
```diff
@@ -73,19 +75,27 @@
...
+            -->
         </dependency>
         <dependency>
             <groupId>org.apache.kafka</groupId>
             <artifactId>kafka-clients</artifactId>
             <version>${storm.kafka.client.version}</version>
+            <scope>compile</scope>
+            <!-- You can reduce jar size by uncommenting this and providing the dependencies manually. See the README for details.
```
hmcl: Isn't the goal of ${provided.scope} to handle the proper scope according to the profile, e.g. just like it is done with the Intellij profile. I am not quite following why the user has to comment/uncomment the scope configuration.

examples/storm-kafka-client-examples/pom.xml
```diff
@@ -73,19 +75,27 @@
...
             <scope>${provided.scope}</scope>
+            -->
         </dependency>
         <dependency>
             <groupId>org.apache.kafka</groupId>
             <artifactId>${storm.kafka.artifact.id}</artifactId>
             <version>${storm.kafka.client.version}</version>
+            <scope>compile</scope>
+            <!-- You can reduce jar size by uncommenting this and providing the dependencies manually. See the README for details.
```
hmcl: Isn't the goal of ${provided.scope} to handle the proper scope according to the profile, e.g. just like it is done with the Intellij profile. I am not quite following why the user has to comment/uncomment the scope configuration.

examples/storm-kafka-client-examples/README.markdown
srdo: Will fix

examples/storm-kafka-client-examples/pom.xml
srdo: Yes, I somehow missed the dollar. I'll try reverting this and update the readme to set the scope to compile

examples/storm-kafka-client-examples/src/main/java/org/apache/storm/kafka/trident/TridentKafkaClientWordCountNamedTopics.java
srdo: The topic name wasn't being passed to the consumer before, only the producers as far as I could tell, so if you used the parameters the example didn't work. Fixing it caused a conflict with the wildcard example, because I'd have to change the newKafkaSpoutConfig signature to take a list of topics. That won't work with the Pattern required by the wildcard example. It seemed easier to just remove the option.

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/builders/SingleTopicKafkaSpoutConfiguration.java
srdo: Sure, will rename

examples/storm-kafka-client-examples/src/main/java/org/apache/storm/kafka/spout/test/KafkaSpoutTopologyMainNamedTopics.java
srdo: Yes. I'll restore that bit

srdo: I remembered why I removed it. LocalCluster is in the storm-server jar, which isn't included by the example projects. I think including it would cause conflict when the jar is deployed to a real cluster. How about I move the ability to run this from a local cluster to a test class? That should still leave people able to run on a local cluster from an IDE, but doesn't interfere with the generated jar.

srdo:
@hmcl I think I addressed everything. Please look again. Thanks.

hmcl:
+1. Thanks @srdo

HeartSaVioR:
@srdo 
I have no idea which branches I need to apply this patch, so please go on merging yourself. Thanks for the patch. :)

srdo:
Will do, thanks @HeartSaVioR :)

PR #2244, opened 2017-07-25
STORM-2659: Add daemon.name variable to storm.cmd to fix log4j logging

srdo:
See https://issues.apache.org/jira/browse/STORM-2659

The python script was updated to handle this (https://github.com/apache/storm/blob/v1.1.1/bin/storm.py#L260), but it seems like storm.cmd was missed.


HeartSaVioR:
+1

PR #2245, opened 2017-07-27
Restore issues missing from changelog since fb2446075de202387658b0cc8…

srdo:
…434ee53ded5d35a

See https://github.com/apache/storm/commit/fb2446075de202387658b0cc8434ee53ded5d35a

I've grepped the log for the added issue numbers on 1.1.x-branch and the associated commits appear to be in 1.x but not 1.1.x, so they likely belong in 1.2.0. On master these issues are listed in 2.0.0, which is maybe also wrong(?)

The two removed issues are duplicate, 2432 isn't fixed in 1.0.0 based on the 1.0.0 tag, and 2343 is listed twice in 1.1.1

ping @HeartSaVioR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,9 @@
 ﻿## 1.2.0
+ * STORM-2520: AutoHDFS should prefer cluster-wise hdfs kerberos principal
+ * STORM-2519: Modify AbstractAutoCreds to look for configKeys in both nimbus and topology configs
+ * STORM-2493: update documents to reflect the changes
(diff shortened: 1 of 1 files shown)


HeartSaVioR:
+1 You can merge this without waiting 24hr.

HeartSaVioR:
As you may know, you can even just made a change if it was wrong. Now you also think it's hard to maintain CHANGELOG, let's see how the discussion goes.

srdo:
Thanks @HeartSaVioR. I'll make similar changes quietly on master.

PR #2247, opened 2017-07-31
STORM-2664: Fix for the distribution packaged on Windows OS can't run on Linux

cluo512:
I made a distribution from source code on windows OS, but when I want to launch this on linux. I found there are some mistake for line break in the scripts. And then I found the settings for line ending in the binary.xml was not set obviously, so it is decided by the OS where run the maven command.
I think its better to set the line ending obviously for unix. because we mostly use linux system to deploy storm and, at the same time, this distribution will run well in the windows OS.


srdo:
Hi Cluo. 

This looks good. I checked, and there doesn't seem to be an issue running the powershell files with LF EOL, so I don't think there's a reason to keep them as CRLF.

Could you open an issue on https://issues.apache.org/jira and rename this PR and commit so it references the JIRA number? See for example https://github.com/apache/storm/pull/2248/commits.

cluo512:
Hi Srdo.
Thank you for your advice. 
I create a issue on [https://issues.apache.org/jira/browse/STORM-2664]. And I will rename this PR later.

srdo:
@cluo512 Could you add the issue number to the commit message? Then I think we're good to merge :)

cluo512:
@srdo The issue number have added.
Thanks for your advice again. I'm new about how to contribute.

cluo512:
Oh..check failed.
I looked the details. It seems that @HeartSaVioR  changed the CHANGELOG.md to DEVELOPER.md, so maven can't found this file.
Should I update the binary.xml to DEVELOPER.md?

HeartSaVioR:
@cluo512 
Sorry we decided to get rid of CHANGELOG.md, but I forgot to remove it from distribution pom as well. I'll do the update and let you know. You can rebase after then.

HeartSaVioR:
@cluo512 Done. Please rebase against master. Thanks in advance.

cluo512:
@HeartSaVioR Get it, thanks.
I updated the code for my fork project just now. 
but it seems that I do a merge not a rebase. Dose it matter?

cluo512:
Ok..Thanks!

HeartSaVioR:
+1

PR #2249, opened 2017-07-31
STORM-2648/STORM-2357: Add storm-kafka-client support for at-most-onc…

srdo:
…e processing and a toggle for whether messages should be emitted with a message id when not using at-least-once

See https://issues.apache.org/jira/browse/STORM-2357 and https://issues.apache.org/jira/browse/STORM-2648.

I'd like to get some opinions on whether this approach is a good idea, or whether I've overlooked a better option, before finishing this up with some tests. I don't love that we'll end up with 3 different committing behaviors.

In 2357 it was noted that the spout doesn't currently support true at-most-once, because using Kafka's auto commit option leaves the possibility that the spout receives a tuple, emits it to the topology, crashes and recovers, and then receives and emits the same tuple. The linked issue suggests solving this by committing polled offsets before emitting them to the topology, which is an option added here.

2648 notes that there is currently no way to make Storm track messages when using auto commit with this spout. This prevents Storm UI from showing the complete latency for the spout, and I would assume also prevents max spout pending from having an effect. I've added a toggle to KafkaSpoutConfig to force the spout to emit messages with message ids, even when using auto commit or the at-most-once option. The spout does nothing on ack or fail when not doing at-least-once.

I'd like to keep the spout config simple for the user, so I've added a processing guarantee setting corresponding to the standard at-least-once code path, the path that uses auto commit, and the path that commits offsets before emitting any tuples.


hmcl:
@srdo can you please assign JIRA's to you and mark them as in progress as you work on them and/or submit a pull request. Thanks.

srdo:
Yes, I forgot.

revans2:
Conceptually the changes look good to me.  I have not dug into it in great detail yet, but I do like the direction of the change.

I would also like to see the documentation and examples updated to reflect the new change.

revans2:
+1

hmcl:
@srdo reviewing it

srdo:
Added some tests and updated the docs.

srdo:
@hmcl Are you reviewing this, or are you satisfied with it?

revans2:
Still +1

srdo:
@hmcl I don't mean to try to rush you, but please let me know if you're still reviewing. If not I'll probably merge in the next few days.

hmcl:
@srdo apologies for the delay. I will finish today.

srdo:
Thanks :)

srdo:
@hmcl Are you still reviewing this? It's fine if you don't have time to look at this, but please say so. I'd like to not keep holding this up.

HeartSaVioR:
This PR has been waiting for about 2 months, and once it gets +1 and no -1, it can be merged.
I'm +1 and will just merge.

@hmcl Please vote -1 later and rollback the merge if you have concern about the patch and would want to vote -1.

HeartSaVioR:
Merged via https://github.com/apache/storm/commit/48f6969027e7b02a5b9220577189d3911aa2226d

@srdo 
Sorry I forgot to add auto close message while squashing commits. Could you close this?
And please craft the patch for 1.x branch since it doesn't looks like a clean cherry-pick.
Thanks in advance!

srdo:
Thanks for reviews. Opened the 1.x version here https://github.com/apache/storm/pull/2353.

PR #2251, opened 2017-08-01
[STORM-2677] consider all sampled tuples which took greater than 0 ms processing time

vinodkc:
Set delta to -1 , so that time delay for tuples that aren't sampled will not be added in BoltExecutorStats.
This is a followup PR of  https://github.com/apache/storm/pull/2185 to fix in  1.x version

### DIFF
--- a/storm-core/src/clj/org/apache/storm/daemon/executor.clj
+++ b/storm-core/src/clj/org/apache/storm/daemon/executor.clj
@@ -672,12 +672,14 @@
 (defn- tuple-time-delta! [^TupleImpl tuple]
   (let [ms (.getProcessSampleStartTime tuple)]
     (if ms
-      (time-delta-ms ms))))
+      (time-delta-ms ms)
+      -1)))
       
 (defn- tuple-execute-time-delta! [^TupleImpl tuple]
   (let [ms (.getExecuteSampleStartTime tuple)]
     (if ms
-      (time-delta-ms ms))))
+      (time-delta-ms ms)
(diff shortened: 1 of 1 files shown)


vinodkc:
retest this

HeartSaVioR:
The failure looks unrelated.
+1

HeartSaVioR:
Btw it is clearly a bug, but in opposite to STORM-2613, because both nil and 0 are logically false. So it doesn't include tuples which are not sampled, but it also doesn't include tuples which are done in 0 ms. Less harmful but still affects the overall latency.

@vinodkc 
I guess we need to file different issue for this, since the bug is affecting differently. Could you please follow up? Thanks in advance!

vinodkc:
@HeartSaVioR 
Sure, filed a new issue STORM-2677.
Can you please change the status of STORM-2613.

HeartSaVioR:
@vinodkc
OK. Please squash commits into one with changing commit title corresponding to STORM-2677. Please also change PR title as well.

HeartSaVioR:
@vinodkc Thanks for changing the PR title. Could you change commit title as well?

vinodkc:
@HeartSaVioR , Thanks for the review comments.
I've changed commit title

HeartSaVioR:
@vinodkc I merged this with some modification (commit title) but auto-close didn't work because it is pointing to non master branch. Could you close this? Thanks again for contribution!

vinodkc:
Thanks.

PR #2250, opened 2017-07-31
STORM-2665: Adapt Kafka's release note generation script for Storm

srdo:
See https://issues.apache.org/jira/browse/STORM-2665

PR #2253, opened 2017-08-01
Remove CHANGELOG.md and update DEVELOPER.md to reflect the change

HeartSaVioR:
This is for master branch, and I'll proceed this for all other version lines without additional pull requests: 1.x-branch, 1.1.x-branch, 1.0.x-branch, 0.10.x-branch, 0.9.x-branch.

No strictly need to get +1 to merge, but just wanting to leave a history on this.


DEVELOPER.md
```diff
@@ -214,14 +214,9 @@ To pull in a merge request you should generally follow the command line instruct
         $ git pull <remote_repo_url> <remote_branch>
     You can use `./dev-tools/storm-merge.py <pull-number>` to produce the above command most of the time.
 
-4.  Assuming that the pull request merges without any conflicts:
-    Update the top-level `CHANGELOG.md`, and add in the JIRA ticket number (example: `STORM-1234`) and ticket
-    description to the change log.  Make sure that you place the JIRA ticket number in the commit comments where
-    applicable.
+4. Run any sanity tests that you think are needed.
```
srdo: Nit: I think it still makes sense to encourage placing the ticket number in the commit message

HeartSaVioR: The commit message refers to the commit on committing change on CHANGELOG.
If we would want to encourage contributors to place ticket number in the commit, it should be placed on `Create a pull request`. If you mean placing the ticket number on merge commit, we can describe it to 3.

srdo: Right, that makes sense.

srdo:
Should we do this before we have the other solution to generating the change log in place?

revans2:
This looks good.  It removes that changelog and updates the docs.  But I want to be sure that we have a script for doing the changelog on a release before we actually remove the changelog. Otherwise we are stuck and cannot do a release.

revans2:
For got to say that I am +1 once we have that script, or at least enough progress on it to feel good that it is going to happen.

HeartSaVioR:
I'm just in favor of #2250, and once we have at least one way to handle the changelog in release, I'd like to get this not too late. The discussion was staled once and it was about to be stale again.

PR #2256, opened 2017-08-02
[STORM-2674] catch NoNodeException when IStormClusterState tries to delete znodes in reportError function

Ethanlm:
See: https://issues.apache.org/jira/browse/STORM-2674

Catch the NoNodeException.

Tested it manually.


HeartSaVioR:
+1

PR #2257, opened 2017-08-02
STORM-2673: add support for prioritizing nodes in scheduling

revans2:
I really would like feedback on the configs and if it makes since to modify other schedulers to support this too.

Config:
I thought about using regexs for nodes instead of just a host name.  Would that be better? Would it make it simple for someone to prefer a large section of a cluster and essentially abuse the feature?

Other Schedulers:
The isolation scheduler possibly, where we could resort The HostAssignableSlots if the topology want to prioritize one host over another.  And then sort them back if they don't

The even scheduler is a little odd as it feels completely against what it stands for.  It would almost turn it into an odd hybrid with the isolation scheduler.

MT scheduler is so complex at this point I could do it, but only if someone really wants it.

### DIFF
--- a/storm-client/src/jvm/org/apache/storm/Config.java
+++ b/storm-client/src/jvm/org/apache/storm/Config.java
@@ -262,6 +262,20 @@ public class Config extends HashMap<String, Object> {
     public static final String TOPOLOGY_SCHEDULER_STRATEGY = "topology.scheduler.strategy";
 
     /**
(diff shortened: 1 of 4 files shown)


HeartSaVioR:
+1 for the change.

@revans2
Regarding accepting regex, since its purpose is only debug, I don't feel we need to apply regex for now. If we are unsure and would like to hear others' voices then we can initiate poll in users@.

Regarding applying other schedulers as well, I guess it would be better to do that, but you can still file new issues addressing them and let this be merged.

PR #2258, opened 2017-08-02
addition of isDebugEnabled() check to improve the performance.

tterstep:


tterstep:
@MichaelTimper

jnioche:
-1 
Why not simply use parameterized messages instead? See https://www.slf4j.org/faq.html#logging_performance

PR #2267, opened 2017-08-07
STORM-2682: Fix issues with null owner on rolling upgrade (1.0.x)

revans2:
This is the 1.0.x version of  #2266 and should probably apply cleanly to all of the 1.x line of releases.

This prevents doing a clean rolling upgrade from 1.0.3 to 1.0.4 and 1.1.0 to 1.1.1.  We should look at getting this in and possibly doing another release ASAP.


HeartSaVioR:
+1

PR #2268, opened 2017-08-08
STORM-2689: Simplify storm-kafka-example…

srdo:
…s, storm-kafka-client-examples and storm-elasticsearch-examples

See https://issues.apache.org/jira/browse/STORM-2689. 

The basic problem is that provided.scope doesn't allow for setting a different scope for storm-client compared to the other dependencies. Right now you can either build a jar with all dependencies + storm-client, which you can only use to run LocalCluster topologies, or a jar without storm-client that is also missing a bunch of other dependencies.

We could add another Maven profile or property like we do with provided.scope for IntelliJ in order to allow users to produce a jar without e.g. kafka-clients, but I think making running the examples as simple as possible is attractive.

I think the easiest solution is to just produce a fat jar by default. This also seems to be how the rest of the examples work.


srdo:
ping @HeartSaVioR and @hmcl since you reviewed https://github.com/apache/storm/pull/2243 and this is related.

srdo:
For some reason the integration test can't download Zookeeper. I'm guessing it's unrelated.

examples/storm-kafka-examples/pom.xml
```diff
@@ -24,6 +24,11 @@
         <version>2.0.0-SNAPSHOT</version>
         <relativePath>../../pom.xml</relativePath>
     </parent>
+    
+    <properties>
+        <!-- Override with -Dkafka.dependency.scope=compile to generate a jar with dependencies -->
+        <kafka.dependency.scope>provided</kafka.dependency.scope>
```
srdo: This workaround is necessary because this module is included as a dependency in storm-kafka-client-examples, so we need a normal install to generate a slim jar.

HeartSaVioR:
I feel that dependencies are going to be tricky in storm-kafka-examples and storm-kafka-client-examples. storm-kafka-client-examples depends on storm-kafka-examples, and storm-kafka-examples depends on storm-starter, and storm-starter depends on lots of dependencies like storm-hdfs, storm-redis, etc.

Looks like storm-kafka-client-examples refers storm-kafka-examples only for referencing `KafkaProducerTopology`. If then I think we shouldn't. There's a side effect: because of relying on KafkaProducerTopology, we don't have any code example to show how to use KafkaBolt in `storm-kafka-client`.

Let's duplicate KafkaProducerTopology with using only `storm-kafka-client` things. @hmcl WDYT?

srdo:
@HeartSaVioR You're right, it's nicer if the example is not relying on storm-kafka. I copied the missing classes to storm-kafka-client-examples, and added a bunch of documentation.

srdo:
We could get rid of the storm-starter dependency if we removed the DRPC demonstration from the Trident examples and replaced the RandomSentenceSpout. There's already a DRPC demo in storm-starter, so I don't know why we also need it here. I think it makes the example more complicated than it needs to be.

HeartSaVioR:
@srdo +1 to remove DRPC demo in other examples rather than storm-starter. Please go ahead. Thanks.

srdo:
@HeartSaVioR Done. I've changed the storm-kafka(-client)-examples projects to just put some random strings in Kafka and retrieve and print them in the Trident examples.

HeartSaVioR:
+1 Nice improvement.

HeartSaVioR:
@srdo Please squash the commits into one. I'll merge after that. Thanks in advance!

srdo:
@HeartSaVioR Thanks, squashed. This should only go on master.

srdo:
Test failure is unrelated, flaky test in storm-jms.

PR #2272, opened 2017-08-08
Try fixing travis build by explicitly setting dist and updating integ…

srdo:
…ration test Zookeeper version

It looks like Travis has begun moving builds to Trusty from Precise (https://blog.travis-ci.com/2017-07-11-trusty-as-default-linux-is-coming). This causes the integration test to fail, because it refers to an older Zookeeper version that isn't available in Trusty.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,9 @@ env:
   - MODULES='!storm-client,!storm-server,!storm-core'
   - MODULES='INTEGRATION-TEST'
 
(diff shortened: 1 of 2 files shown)


HeartSaVioR:
+1 Nice finding.

HeartSaVioR:
Thanks @srdo, merged into master, 1.x, 1.1.x, 1.0.x branches.

PR #2274, opened 2017-08-10
Fix checkstyle file inaccessibility issue on modules

HeartSaVioR:
Before the patch we couldn't build example module in module's directory.

You can reproduce the behavior via building binary distribution, and extract archive file, and change directory to storm-starter, and do `mvn clean package`.

Please note that we will publish storm-checkstyle artifact to Maven repo. when releasing Storm 2.0.0, hence it should be available without running `mvn clean install` afterwards.


pom.xml
```diff
@@ -378,6 +378,7 @@
         <module>examples/storm-jms-examples</module>
         <module>examples/storm-rocketmq-examples</module>
         <module>examples/storm-perf</module>
+        <module>storm-checkstyle</module>
```
srdo: Nit: Move up to the uncategorized modules so this isn't part of the examples list

HeartSaVioR:
@srdo Thanks for the point. I move `storm-checkstyle` to the top. Could you review again?

srdo:
Still +1, thanks

HeartSaVioR:
@srdo Thanks for reviewing!

PR #2278, opened 2017-08-15
Quick fix: remove duplicated IStormClusterState object in ReadClusterState.java

Ethanlm:
It seems to me that `private final IStormClusterState clusterState;` is a dup of ` private final IStormClusterState stormClusterState;`. So I removed `clusterState`.


HeartSaVioR:
+1 Nice finding.

HeartSaVioR:
Thanks @Ethanlm, I merged into master.

PR #2280, opened 2017-08-16
STORM-2500: remove call to waitUntilReady

ZacharyThomas:
In 1.1.x PacemakerClient was refactored so waitUntilReady was private, and called by the send method. But this clojure reference remained, causing pacemaker to fail to initialize. AFAIK this reference can simply be removed, as waitForReady is called on send. However, I'm not very familiar with the code base. I applied this patch to our own cluster with success.

### DIFF
--- a/storm-core/src/clj/org/apache/storm/pacemaker/pacemaker_state_factory.clj
+++ b/storm-core/src/clj/org/apache/storm/pacemaker/pacemaker_state_factory.clj
@@ -106,14 +106,13 @@
 (defn get-pacemaker-write-client [conf servers client-pool]
   ;; Client should be created in case of an exception or first write call
   ;; Shutdown happens in the retry loop
-  (try 
-    (.waitUntilReady
-     (let [client (get @client-pool (first @servers))]
-       (if (nil? client)
-         (do
-           (swap! client-pool merge {(first @servers) (PacemakerClient. conf (first @servers))})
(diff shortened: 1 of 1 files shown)


HeartSaVioR:
+1
This fixes what #2142 already fixed, but no activity is seen on #2142 I'll just merge this.
Travis CI fails on all JDK7 builds, hence not related to this.

HeartSaVioR:
@ZacharyThomas 
Thanks for the patch. I merged this but can't update assignee on issue (https://issues.apache.org/jira/browse/STORM-2500). 

Could you let me know your Apache JIRA ID? Thanks in advance!

ZacharyThomas:
@HeartSaVioR Thanks for the merge. My username on jira should be actking. Sorry I couldn't find anywhere to set assignee on Jira.

HeartSaVioR:
@ZacharyThomas Just updated. Thanks again for contributing!

ZacharyThomas:
@HeartSaVioR It was fun. Encountered this while trying to debug the exclamation topology dying on a new cluster (see: https://mail-archives.apache.org/mod_mbox/storm-user/201708.mbox/%3CCA%2BAmaO0oP5imSgs7nXboM_VQva0PzoQNGGURG%2BfR3rySystm1g%40mail.gmail.com%3E)

Which, if you have any suggestions for, that'd be great.

PR #2282, opened 2017-08-17
STORM-2694: Add KafkaTupleListener to storm-kafka-client

bijanfahimi:
This PR proposes an improvement to the KafkaSpout. A KafkaTupleListener interface can now be implemented to handle state changes of the tuple. 
Our primary use case is to put tuples, which are discarded by the RetryService into a dead letter queue.

### DIFF
--- a/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/EmptyKafkaTupleListener.java
+++ b/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/EmptyKafkaTupleListener.java
@@ -1,0 +1,53 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ *   or more contributor license agreements.  See the NOTICE file
+ *   distributed with this work for additional information
+ *   regarding copyright ownership.  The ASF licenses this file
+ *   to you under the Apache License, Version 2.0 (the
+ *   "License"); you may not use this file except in compliance
+ *   with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+
+
+package org.apache.storm.kafka.spout;
+
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+
+import org.apache.kafka.common.TopicPartition;
+import org.apache.storm.task.TopologyContext;
+
+public final class EmptyKafkaTupleListener implements KafkaTupleListener {
+
+    @Override
+    public void open(Map<String, Object> conf, TopologyContext context) { }
+
+    @Override
+    public void onEmit(List<Object> tuple, KafkaSpoutMessageId msgId) { }
+
+    @Override
+    public void onAck(KafkaSpoutMessageId msgId) { }
+
+    @Override
+    public void onPartitionsReassigned(Collection<TopicPartition> topicPartitions) { }
+
+    @Override
+    public void onRetry(KafkaSpoutMessageId msgId) { }
+
+    @Override
+    public void onMaxRetryReached(KafkaSpoutMessageId msgId) { }
+
+    @Override
+    public String toString() {
+        return "EmptyKafkaTupleListener";
+    }
+}
--- a/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
+++ b/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
@@ -73,7 +73,9 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
     // Strategy to determine the fetch offset of the first realized by the spout upon activation
     private transient FirstPollOffsetStrategy firstPollOffsetStrategy;  
     // Class that has the logic to handle tuple failure
-    private transient KafkaSpoutRetryService retryService;              
+    private transient KafkaSpoutRetryService retryService;
+    // Handles tuple events (emit, ack etc.)
+    private transient KafkaTupleListener tupleListener;
     // timer == null for auto commit mode
     private transient Timer commitTimer;                                
     // Flag indicating that the spout is still undergoing initialization process.
@@ -95,7 +97,7 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
 
 
     public KafkaSpout(KafkaSpoutConfig<K, V> kafkaSpoutConfig) {
-        this(kafkaSpoutConfig, new KafkaConsumerFactoryDefault<K, V>());
+        this(kafkaSpoutConfig, new KafkaConsumerFactoryDefault<>());
     }
 
     //This constructor is here for testing
@@ -121,6 +123,8 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
         // Retries management
         retryService = kafkaSpoutConfig.getRetryService();
 
+        tupleListener = kafkaSpoutConfig.getTupleListener();
+
         if (!consumerAutoCommitMode) {     // If it is auto commit, no need to commit offsets manually
             commitTimer = new Timer(TIMER_DELAY_MS, kafkaSpoutConfig.getOffsetsCommitPeriodMs(), TimeUnit.MILLISECONDS);
         }
@@ -129,6 +133,8 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
         offsetManagers = new HashMap<>();
         emitted = new HashSet<>();
         waitingToEmit = Collections.emptyListIterator();
+
+        tupleListener.open(conf, context);
 
         LOG.info("Kafka Spout opened with the following configuration: {}", kafkaSpoutConfig);
     }
@@ -151,6 +157,7 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
             LOG.info("Partitions reassignment. [task-ID={}, consumer-group={}, consumer={}, topic-partitions={}]",
                     context.getThisTaskId(), kafkaSpoutConfig.getConsumerGroupId(), kafkaConsumer, partitions);
 
+            tupleListener.onPartitionsReassigned(partitions);
             initialize(partitions);
         }
 
@@ -363,6 +370,7 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
                         } else {
                             collector.emit(tuple, msgId);
                         }
+                        tupleListener.onEmit(tuple, msgId);
                     }
                     LOG.trace("Emitted tuple [{}] for record [{}] with msgId [{}]", tuple, record, msgId);
                     return true;
@@ -432,6 +440,7 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
             }
             emitted.remove(msgId);
         }
+        tupleListener.onAck(msgId);
     }
 
     // ======== Fail =======
@@ -449,8 +458,10 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
             LOG.debug("Reached maximum number of retries. Message [{}] being marked as acked.", msgId);
             // this tuple should be removed from emitted only inside the ack() method. This is to ensure
             // that the OffsetManager for that TopicPartition is updated and allows commit progression
+            tupleListener.onMaxRetryReached(msgId);
             ack(msgId);
         } else {
+            tupleListener.onRetry(msgId);
             emitted.remove(msgId);
         }
     }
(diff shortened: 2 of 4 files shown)


external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaTupleListener.java
```diff
@@ -0,0 +1,48 @@
...
+
+    /**
+     * Called when a tuple is acked.
+     *
+     * @param msgId The id of the tuple in the spout.
+     */
+    void onAck(KafkaSpoutMessageId msgId);
+
+    /**
+     * Called when a fail reaches the spout, but the Kafka record does not belong to the spout anymore.
```
srdo: Nit: The kafka record belongs to a partition that is not assigned to the spout anymore.
I'm wondering what this hook would be useful for? The record will get replayed by a different task if using at-least-once, otherwise it'll be dropped. I think if the reason it is here is to allow implementations to clean up after emit, we should just have a general onPartitionsReassigned(assignedPartitions) instead.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaTupleListener.java
```diff
@@ -0,0 +1,48 @@
+package org.apache.storm.kafka.spout;
+
+import java.io.Serializable;
+import java.util.List;
+
+
+/**
+ * The KafkaTupleListener handles state changes of a kafka tuple inside a KafkaSpout.
+ */
+public interface KafkaTupleListener extends Serializable {
```
srdo: It might be useful to have a prepare method on this interface too, in case implementations need to instantiate something that can't be serialized, e.g. a KafkaProducer. I'm also wondering if it would be useful to be able to emit dead tuples to a different stream as well?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/EmptyKafkaTupleListener.java
```diff
@@ -0,0 +1,35 @@
+package org.apache.storm.kafka.spout;
+
+import java.util.List;
+
+public class EmptyKafkaTupleListener implements KafkaTupleListener {
+    @Override
+    public void onEmit(List<Object> tuple, KafkaSpoutMessageId msgId) {
+        // empty method
```
srdo: This comment isn't really necessary.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -73,7 +73,9 @@
     // Strategy to determine the fetch offset of the first realized by the spout upon activation
     private transient FirstPollOffsetStrategy firstPollOffsetStrategy;  
     // Class that has the logic to handle tuple failure
-    private transient KafkaSpoutRetryService retryService;              
+    private transient KafkaSpoutRetryService retryService;
+    // Handles the events off a tuple
```
srdo: Nit: Rephrase as "Handles tuple events (emit, ack etc.)", I don't think it is clear what this means.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaTupleListener.java
bijanfahimi: Good idea, I have added a prepare method or "open" since it is in the context of a spout.  

I am not sure what you mean by "dead tuples". 
If `isEmitTuple(tuple)` is false in line 380 of the KafkaSpouts `emitTupleIfNotEmitted`? But then i would have to change the onAck call as well, since it would be called twice. I can imagine that someone needs more callbacks. However, I tried to keep the interface simple and small. But if you wish I would gladly add new functionality to the interface.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaTupleListener.java
bijanfahimi: I changed the name of the method and updated the comment to reflect your remarks. You were right about the onPartitionsReassigned. That was what I intended with the call.

bijanfahimi:
I have updated the PR according to the comments. The checkstyle errors should be fixed. If there is anything else, please let me know.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaTupleListener.java
srdo: I meant being able to emit the tuples that have exceeded their retry limit (those that hit onMaxRetryReached) to a stream. It's probably not a great idea, since we'd need to consider processing guarantees for those too, so never mind :)

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/EmptyKafkaTupleListener.java
```diff
@@ -0,0 +1,33 @@
...
+    public void open(Map<String, Object> conf, TopologyContext context, SpoutOutputCollector collector) { }
+
+    @Override
+    public void onEmit(List<Object> tuple, KafkaSpoutMessageId msgId) { }
+
+    @Override
+    public void onAck(KafkaSpoutMessageId msgId) { }
+
+    @Override
+    public void onPartitionsReassigned(KafkaSpoutMessageId msgId) { }
```
srdo: I think this should take the list of assigned partitions as a parameter, and be called in the onPartitionsAssigned of the rebalance listener instead. That would allow implementations to delete any saved message ids that are for revoked partitions easily.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/EmptyKafkaTupleListener.java
```diff
@@ -0,0 +1,33 @@
...
+
+    @Override
+    public void onRetry(KafkaSpoutMessageId msgId) { }
+
+    @Override
+    public void onMaxRetryReached(KafkaSpoutMessageId msgId) { }
+
+    @Override
+    public String toString() {
+        return "EmptyKafkaTupleListener";
```
srdo: Nit: What is this toString for?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -363,6 +369,7 @@ private boolean emitTupleIfNotEmitted(ConsumerRecord<K, V> record) {
                         } else {
                             collector.emit(tuple, msgId);
                         }
+                        tupleListener.onEmit(tuple, msgId);
```
srdo: Is it intended that this isn't called in auto commit mode? I would understand why we wouldn't want to call it there, but I think we should maybe add a note to the interface so it's obvious to users.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/EmptyKafkaTupleListener.java
bijanfahimi: The KafkaSpoutConfig has implemented the toString method, which print the current config. I have added that for consistency reasons.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
bijanfahimi: I have added additional comments.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/EmptyKafkaTupleListener.java
bijanfahimi: Done.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/EmptyKafkaTupleListener.java
```diff
@@ -0,0 +1,35 @@
+package org.apache.storm.kafka.spout;
```
srdo: You need to add the Apache license to the top of the new files. You can just copy it from one of the other ones.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/EmptyKafkaTupleListener.java
srdo: Makes sense

srdo:
Thanks for making the changes. +1 once the license issue is fixed

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/EmptyKafkaTupleListener.java
bijanfahimi: I have added the license.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaTupleListener.java
```diff
@@ -0,0 +1,78 @@
...
+import org.apache.storm.task.TopologyContext;
+
+
+/**
+ * The KafkaTupleListener handles state changes of a kafka tuple inside a KafkaSpout.
+ */
+public interface KafkaTupleListener extends Serializable {
+
+
+    void open(Map<String, Object> conf, TopologyContext context, SpoutOutputCollector collector);
```
srdo: I'm wondering if we should pass along the output collector here? If we don't support retrying tuples emitted to an error stream, then I don't know how useful emitting to such a stream will be. What do you think?

bijanfahimi: Thats a good point. So how I see it there are two options. First, add a hook for declareOutputFields to allow streaming to an error stream or, as you suggested, removing the SpoutOutputcollector from the open parameters. 

I tend to your suggestion. I think when this interface is capable of emitting it acts like a spout Itself and not like a simple listener.

If more people need to emit to an user specific error stream the interface can be expanded afterwards, i guess.

srdo: I'm leaning towards removing it too. We can always add in error stream support later if someone needs it.

bijanfahimi: I have updated the PR.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/EmptyKafkaTupleListener.java
```diff
@@ -0,0 +1,54 @@
...
+
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+
+import org.apache.kafka.common.TopicPartition;
+import org.apache.storm.spout.SpoutOutputCollector;
+import org.apache.storm.task.TopologyContext;
+
+public class EmptyKafkaTupleListener implements KafkaTupleListener {
```
hmcl: @bijanfahimi I don't think this class is necessary. We can just put this default, idle, behavior in the interface itself by making the interface methods default methods.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/EmptyKafkaTupleListener.java
hmcl: If there is no config fields to show, in my opinion this method should simply return an empty String.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/EmptyKafkaTupleListener.java
bijanfahimi: @hmcl default methods in interfaces is a java 8 feature. If storm is > java 7 then indeed this class is not needed. I haven't checked.

srdo: I think it is nicer to leave the empty implementation as a separate thing from the interface, so other implementations have to implement all methods. Maybe make this one final too.

bijanfahimi: It's now final.

srdo:
+1

bijanfahimi:
@hmcl and @srdo.  Is there anything else I can do.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/EmptyKafkaTupleListener.java
srdo: @hmcl At the risk of getting into bikeshedding territory, I think it makes sense to leave it this way, since the listener is being printed in the toString of KafkaSpoutConfig. As it is now, the print will show "tupleListener=EmptyKafkaTupleListener", so at least it is obvious what's in the field. Leaving it empty would show "tupleListener=".

I'm happy to merge this either way though.

srdo:
@bijanfahimi I think the last minor thing is the comment left by @hmcl here https://github.com/apache/storm/pull/2282#discussion_r135033155. I'll merge this tomorrow unless @hmcl objects.

srdo:
Thanks @bijanfahimi, I merged to master. Cherry-picking to 1.x-branch hit some conflicts, so please open a PR against that branch as well if you'd like this to go in those releases.

PR #2284, opened 2017-08-18
[STORM-2695] Fix in docs: BlobStore uncompress argument should be Boolean

Ethanlm:
See: https://issues.apache.org/jira/browse/STORM-2695

We need to update the doc for "uncompress" argument from String (e.g. "false") to Boolean (e.g. false)


HeartSaVioR:
+1 Nice finding.

HeartSaVioR:
This should be applied to 1.x version lines as well.

PR #2288, opened 2017-08-21
[STORM-2700] Should not check Blobstore ACL if the validation is disabled

Ethanlm:
See: https://issues.apache.org/jira/browse/STORM-2700

`canUserReadBlob ` function should take `STORM_BLOBSTORE_ACL_VALIDATION_ENABLED` config into consideration. 

Tested it manually.

### DIFF
--- a/storm-server/src/main/java/org/apache/storm/localizer/Localizer.java
+++ b/storm-server/src/main/java/org/apache/storm/localizer/Localizer.java
@@ -523,7 +523,7 @@ public class Localizer {
         out = new FileOutputStream(downloadFile);
         numTries++;
         try {
-          if (!ServerUtils.canUserReadBlob(blobstore.getBlobMeta(key), user)) {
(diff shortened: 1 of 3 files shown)


Ethanlm:
This error should be unrelated.

HeartSaVioR:
First of all, the change looks good.

Btw, the default value of `storm.blobstore.acl.validation.enabled` is `false` so it will affect all of the existing cluster. Moreover I don't think it is good default value for secure cluster. Maybe we should explore to change the default value to `true` and see it also works with insecure cluster.

Ethanlm:
@HeartSaVioR Yes. But changing default value seems not very simple. We probably need to file another JIRA for it if we want to do it.

revans2:
@HeartSaVioR I agree that it should not be false by default for a secure cluster.  But like @Ethanlm said it is not a simple task to make part of security work on an insecure cluster.  The other alternative is to have a way to automatically determine if the cluster is secure or not.

PR #2290, opened 2017-08-22
[STORM-2703] Handle ExecutionException in handleWaitingForBlobLocalization state

Ethanlm:
See: https://issues.apache.org/jira/browse/STORM-2703

Keep supervisor from crashing when exception happens because of blob permission exception (or KeyNotFound exception). The code change here will bring state machine back to `MachineState.EMPTY` when that happens and start over.

So, if the blob ACL permission is fixed (or the key is uploaded) during the runtime, the topology will start to work; if the topology is killed, the Slot is able to continue to deal with other assignment.

### DIFF
--- a/storm-server/src/main/java/org/apache/storm/daemon/supervisor/Slot.java
+++ b/storm-server/src/main/java/org/apache/storm/daemon/supervisor/Slot.java
@@ -27,6 +27,7 @@ import java.util.HashSet;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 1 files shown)


Ethanlm:
Tested it manually.
 
1. Create a blob `key1` with ACL set to `u:mapredqa:rwa`.
2. Submit a topology as `ethan` with `-c topology.blobstore.map='{"key1":{"localname":"test-blob.txt", "uncompress":false}}'`.
3. Check the supervisor.log: `[ethan] does not have access no key1`. And the slot changed to `EMPTY` state. The supervisor didn't crash. 
image
4. Later, set the blob `key1` ACL with command `set-acl -s u:ethan:rwa key1`
5. Check the supervisor.log. The workers started running. Also checked with worker.log which showed normal outputs.
image

HeartSaVioR:
+1 Nice finding.

PR #2291, opened 2017-08-22
[STORM-2704] Check blob permission before submitting the topology

Ethanlm:
See: https://issues.apache.org/jira/browse/STORM-2704

The original code only checks if the keys in `topology.blobstore.map` exist in the BlobStore.  We need to check the ACL as well.

### DIFF
--- a/storm-client/test/jvm/org/apache/storm/TestConfigValidate.java
+++ b/storm-client/test/jvm/org/apache/storm/TestConfigValidate.java
@@ -20,7 +20,6 @@ package org.apache.storm;
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
-import org.apache.storm.generated.InvalidTopologyException;
 import org.apache.storm.utils.Utils;
 import org.apache.storm.validation.ConfigValidation;
 import org.apache.storm.validation.ConfigValidation.*;
@@ -34,11 +33,11 @@ import java.lang.reflect.InvocationTargetException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
-import java.util.HashSet;
 import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+
 
 public class TestConfigValidate {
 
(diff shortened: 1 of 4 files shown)


Ethanlm:
Tested it manually. All the experiments submitted topologies as `ethan`.

1. With `storm.blobstore.acl.validation.enabled: false`
1.1. When the key does not exist: `InvalidTopologyException` 
image
1.2. When the key exists with ACL set to `u:mapredqa:rwa`: submitted successfully.
image

2. With `storm.blobstore.acl.validation.enabled: true`
2.1 When the key does not exist: `InvalidTopologyException` 
image
2.2 When the key exists with ACL set to `u:mapredqa:rwa`: `AuthorizationException`
image
2.3 When the key exists with ACL set to `u:ethan:rwa`: submitted successfully.
image

storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
```diff
@@ -2529,7 +2529,7 @@ public void submitTopologyWithOpts(String topoName, String uploadedJarLocation,
                 }
             }
             validateTopologyWorkerMaxHeapSizeConfigs(topoConf, topology);
-            Utils.validateTopologyBlobStoreMap(topoConf, Sets.newHashSet(blobStore.listKeys()));
+            Utils.validateTopologyBlobStoreMap(topoConf);
```
HeartSaVioR: Maybe better to add method which contains NimbusBlobStore as parameter to avoid creating NimbusBlobStore in Nimbus.

storm-client/test/jvm/org/apache/storm/TestConfigValidate.java
```diff
@@ -84,21 +83,6 @@ public void invalidConfigTest() throws InvocationTargetException, NoSuchMethodEx
         ConfigValidation.validateFields(conf);
     }
 
-    @Test(expected = InvalidTopologyException.class)
```
HeartSaVioR: Could we mock NimbusBlobStore and retain the test?

Ethanlm: I tried to mock it and then decided not to because the logic is really simple in `validateTopologyBlobStoreMap`. But I will add the unit test if you think it's better

storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
Ethanlm: Will see how to do it

storm-client/test/jvm/org/apache/storm/TestConfigValidate.java
HeartSaVioR: Yes please do it if you really don't mind. I think we need to have broader coverage of tests.

Ethanlm: Sure. Will do

Ethanlm:
@HeartSaVioR  Thanks for the review. Addressed your comments. 

Tested it manually again.

Also ran the same tests with commenting out [validateTopologyBlobStoreMap function in validateConfs](https://github.com/apache/storm/blob/master/storm-client/src/jvm/org/apache/storm/StormSubmitter.java#L564-L567) so that I can test the function call  (`Utils.validateTopologyBlobStoreMap(topoConf, blobStore)` ) in `Nimbus.java`. All tests are passed.  Some exception messages like:

image
image

PR #2292, opened 2017-08-23
STORM-2705: DRPCSpout sleeps twice

sakanaou:
When the `DRPCSpout` cannot fetch a request it puts itself to sleep at the end of `nextTuple()`. After leaving this method the case that the spout did not emit any tuple is handled by `ISpoutWaitStrategy` - which in case of `SleepSpoutWaitStrategy` (the default) issues another sleep.

This causes unnecessary context switches in my opinion, especially with a high parallelism setting and many different `DRPCSpout`s.

Only the configured / active `ISpoutWaitStrategy` should decide the action.


HeartSaVioR:
+1 to master branch

sakanaou:
Ok, I will update the PR so it can be applied to master easily.

HeartSaVioR:
+1 again.

PR #2294, opened 2017-08-24
STORM-2517 add interface for Writer, make AbstractHDFSWriter properti…

Angus-Slalom:
…es protected

### DIFF
--- a/external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/AbstractHdfsBolt.java
+++ b/external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/AbstractHdfsBolt.java
@@ -30,7 +30,6 @@ import org.apache.storm.hdfs.bolt.format.FileNameFormat;
 import org.apache.storm.hdfs.bolt.rotation.FileRotationPolicy;
 import org.apache.storm.hdfs.bolt.rotation.TimedRotationPolicy;
 import org.apache.storm.hdfs.bolt.sync.SyncPolicy;
-import org.apache.storm.hdfs.common.AbstractHDFSWriter;
 import org.apache.storm.hdfs.common.NullPartitioner;
 import org.apache.storm.hdfs.common.Partitioner;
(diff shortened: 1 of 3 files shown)


Angus-Slalom:
Created for pull request https://github.com/apache/storm/pull/2157

revans2:
@Angus-Slalom there are some checkstyle violations here, which is why the build is failing.  Other then that it looks good.  +1

revans2:
```
external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/Writer.java
	18: 'package' should be separated from previous statement.
	23: Wrong lexicographical order for 'java.io.IOException' import. Should be before 'org.apache.storm.tuple.Tuple'.
```

are the two issues associated with the new file you added.  Then there is just one more added in, but because it is changes to existing files it is hard to tell, so it is simplest to just pick one or two other issues to "fix" too.

```
external/storm-hdfs/src/main/java/org/apache/storm/hdfs/common/AbstractHDFSWriter.java
	18: 'package' should be separated from previous statement.
	25: Wrong lexicographical order for 'java.io.IOException' import. Should be before 'org.apache.storm.tuple.Tuple'.
	27: Abbreviation in name 'AbstractHDFSWriter' must contain no more than '2' consecutive capital letters.
	27: 'public' modifier out of order with the JLS suggestions.
	31: 'protected' modifier out of order with the JLS suggestions.
	32: 'protected' modifier out of order with the JLS suggestions.
	34: Missing a Javadoc comment.
	40: Missing a Javadoc comment.
	40: 'public' modifier out of order with the JLS suggestions.
	47: 'public' modifier out of order with the JLS suggestions.
	51: 'public' modifier out of order with the JLS suggestions.
	63: 'protected' modifier out of order with the JLS suggestions.
	65: 'protected' modifier out of order with the JLS suggestions.
	67: 'protected' modifier out of order with the JLS suggestions.
```

revans2:
@Angus-Slalom actually because these are all just white space changes I can do it myself when I check it in.

Let me run the tests all the way through to be sure it looks good then I'll pull it in.

PR #2295, opened 2017-08-28
STORM-2712: accept arbitrary number of rows per tuple in storm-cassandra

morokosi:
Current implementation in `TridentResultSetValuesMapper::map` restricts a SELECT query to return one row. In `StateQueryProcessor::finishBatch`, it checks the equality between the result size of `batchRetrieve` and input tuple size. When the number of result rows is less than 1 or greater than 1, it breaks the condition and an exception is thrown.

We should accept arbitrary number of rows by adjusting List dimensions.

### DIFF
--- a/external/storm-cassandra/src/main/java/org/apache/storm/cassandra/trident/state/TridentResultSetValuesMapper.java
+++ b/external/storm-cassandra/src/main/java/org/apache/storm/cassandra/trident/state/TridentResultSetValuesMapper.java
@@ -44,6 +44,8 @@ public class TridentResultSetValuesMapper implements CQLResultSetValuesMapper {
     @Override
     public List<List<Values>> map(Session session, Statement statement, ITuple tuple) {
         List<List<Values>> list = new ArrayList<>();
+        List<Values> innerList = new LinkedList<>();
+        list.add(innerList);
         ResultSet resultSet = session.execute(statement);
         for (Row row : resultSet) {
             final Values values = new Values();
@@ -54,9 +56,7 @@ public class TridentResultSetValuesMapper implements CQLResultSetValuesMapper {
                     values.add(row.getObject(field));
                 }
             }
-            list.add(new LinkedList<Values>() {{
-                add(values);
-            }});
+            innerList.add(values);
         }
         return list;
     }


external/storm-cassandra/src/main/java/org/apache/storm/cassandra/trident/state/TridentResultSetValuesMapper.java
```diff
@@ -44,6 +44,8 @@ public TridentResultSetValuesMapper(Fields outputDeclaredFields) {
     @Override
     public List<List<Values>> map(Session session, Statement statement, ITuple tuple) {
         List<List<Values>> list = new ArrayList<>();
+        List<Values> innerList = new LinkedList<>();
```
HeartSaVioR: Could you explain that how different the changeset is? I don't get it and for me it looks like same.

morokosi:
In a nutshell, the changeset makes utilize of inner list dimensions correctly when a cassandra query result has multiple rows.

Assume we have a cassandra table like the following:

```sql
CREATE TABLE t1 (pk int, ck int, data text, PRIMARY KEY (pk, ck)));
```

| pk | ck | data |
| -- | -- | -- |
| 1 | 0 | a |
| 1 | 1 | b |
| 2 | 0 | c |
| 3 | 0 | d |

and we have a StateQuery as `SELECT data from t1 where pk = ?` with input tuples `1,2,3`.

With current implementation, the return list of `TridentResultSetValuesMapper::map` is as follows:

```
[
  [values(pk=1, ck=0)],
  [values(pk=1, ck=1)],
  [values(pk=2, ck=0)],
  [values(pk=3, ck=0)]
]
```

This throws an exception because the list size differs from the input tuple size.

On the other hand, one with the changeset results

```
[
  [values(pk=1, ck=0), values(pk=1, ck=1)],
  [values(pk=2, ck=0)],
  [values(pk=3, ck=0)]  
]
```

which the list size is same as the input tuple size.

HeartSaVioR:
Ah OK. I missed that difference while looking at the code, and now I realized the difference even in the code.
Thanks for clarifying. +1

PR #2299, opened 2017-08-30
STORM-2443 Fix issues on changing log level

HeartSaVioR:
* Addressed below issues
  * Nimbus throws error when changing log level on UI topology page
  * Log configs in worker are removed even they're not timed out and never reset
  * The unit of epoch between backend and frontend are different
    * second in backend, millisecond in frontend
    * millisecond is right so fixed it

The above issues seem to be only in master branch.


HeartSaVioR:
@satishd Sure. I missed that. Will address.

HeartSaVioR:
@satishd Addressed and squashed commits.

satishd:
It seems `org.apache.storm.st.tests.window.SlidingWindowTest` is getting failed but that looks to be an intermittent failure and it is not related to this change.

PR #2301, opened 2017-08-30
Quick fix: remove storm-rename-hack dependency

Ethanlm:
Since storm-rename-hack is removed, I think we should remove the dependency.


srdo:
+1

HeartSaVioR:
+1 Nice finding.

revans2:
+1 sorry I missed it when I did the other hanges

PR #2303, opened 2017-08-31
STORM-2660 The Nimbus storm-local directory is relative to the workin…

HeartSaVioR:
…g directory of the shell executing "storm nimbus"

* Let Nimbus uses same storm-local directory as Supervisor which is known to be valid

### DIFF
--- a/storm-client/src/jvm/org/apache/storm/utils/ConfigUtils.java
+++ b/storm-client/src/jvm/org/apache/storm/utils/ConfigUtils.java
@@ -215,6 +215,20 @@ public class ConfigUtils {
         }
     }
 
+    public static String absoluteStormBlobStoreDir(Map<String, Object> conf) {
+        String stormHome = System.getProperty("storm.home");
+        String blobStoreDir = (String) conf.get(Config.BLOBSTORE_DIR);
+        if (blobStoreDir == null) {
+            return ConfigUtils.absoluteStormLocalDir(conf);
+        } else {
+            if (new File(blobStoreDir).isAbsolute()) {
+                return blobStoreDir;
+            } else {
+                return (stormHome + FILE_SEPARATOR + blobStoreDir);
+            }
+        }
+    }
+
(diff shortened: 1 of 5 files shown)


srdo:
Thanks for taking a look at this. The fix looks good, but I think this is an issue elsewhere too. At least this path is making the same mistake https://github.com/apache/storm/blob/85486dde9dc5cff8e9a7cbb7c913c5cb68ed3be5/storm-server/src/main/java/org/apache/storm/blobstore/LocalFsBlobStore.java#L90. 

Looking at uses of STORM_LOCAL_DIR, I'm wondering if these uses are also suspicious:
* https://github.com/apache/storm/blob/a4afacd9617d620f50cf026fc599821f7ac25c79/storm-core/src/jvm/org/apache/storm/command/KillWorkers.java#L31
*https://github.com/apache/storm/blob/e1dd247ce97ad560c72554ed612e1ec8e69e9777/storm-server/src/main/java/org/apache/storm/scheduler/utils/ArtifactoryConfigLoader.java#L335

HeartSaVioR:
Nice finding @srdo . Addressed other places as well.

storm-client/src/jvm/org/apache/storm/utils/ConfigUtils.java
```diff
@@ -215,6 +215,20 @@ public static String absoluteStormLocalDir(Map<String, Object> conf) {
         }
     }
 
+    public static String absoluteStormBlobStoreDir(Map<String, Object> conf) {
+        String stormHome = System.getProperty("storm.home");
+        String blobStoreDir = (String) conf.get(Config.BLOBSTORE_DIR);
+        if (blobStoreDir == null) {
+            return null;
```
srdo: I think we can make this a little nicer by having this return absoluteStormLocalDir. There's less risk of us accidentally forgetting to null check the result if this ever gets used elsewhere.

srdo:
+1, verified that `storm nimbus` now puts storm-local in the right directory.

HeartSaVioR:
@srdo Also addressed review comments and squashed commits into one.

srdo:
+1, thanks :)

PR #2306, opened 2017-09-01
[STORM-2717] Output topology Id from storm list command

Ethanlm:
I think we should output topology id from the storm list command for debugging purpose.

JIRA: https://issues.apache.org/jira/browse/STORM-2717


Ethanlm:
Something like this:
image

HeartSaVioR:
+1

PR #2314, opened 2017-09-08
[STORM-2731] - Simple checks in Storm Windowing

jerrypeng:
There is also inconsistent and mixed used of Longs and Ints throughout the windowing code.  Perhaps we should just change all the numeric values to use longs?  Especially since all calculation of time is done in milliseconds.  It would also be nice to make clear in both the code and documentation that all timestamps and time based calculations is in milliseconds.

### DIFF
--- a/storm-client/src/jvm/org/apache/storm/topology/base/BaseWindowedBolt.java
+++ b/storm-client/src/jvm/org/apache/storm/topology/base/BaseWindowedBolt.java
@@ -89,7 +89,14 @@ public abstract class BaseWindowedBolt implements IWindowedBolt {
         public final int value;
 
         public Duration(int value, TimeUnit timeUnit) {
-            this.value = (int) timeUnit.toMillis(value);
+            if (value < 0) {
+                throw new IllegalArgumentException("Duration cannot be negative");
+            }
+            long longVal = timeUnit.toMillis(value);
+            if (longVal > (long) Integer.MAX_VALUE) {
+                throw new IllegalArgumentException("Duration is too long");
+            }
+            this.value = (int)longVal;
         }
 
         /**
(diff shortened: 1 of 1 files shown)


storm-client/src/jvm/org/apache/storm/topology/base/BaseWindowedBolt.java
```diff
@@ -339,6 +364,9 @@ public BaseWindowedBolt withLag(Duration duration) {
      * @param interval the interval at which watermark events are generated
      */
     public BaseWindowedBolt withWatermarkInterval(Duration interval) {
+        if (interval == null) {
+            throw new IllegalArgumentException("Lag duration cannot be set null");
```
srdo: This seems like it belongs in the withLag method.

srdo:
I agree, we should use long for time unit parameters. 

A decent way to make clear that timestamps are in milliseconds is to name the relevant variables so it is obvious, e.g. `timeMs` instead of `time`.

jerrypeng:
@srdo I totally agree with your idea.  I looked into changing all time variables to be long and renaming all variables to have a "Ms".  It will probably be a significant diff

storm-client/src/jvm/org/apache/storm/topology/base/BaseWindowedBolt.java
```diff
@@ -339,6 +364,9 @@ public BaseWindowedBolt withLag(Duration duration) {
      * @param interval the interval at which watermark events are generated
      */
     public BaseWindowedBolt withWatermarkInterval(Duration interval) {
+        if (interval == null) {
+            throw new IllegalArgumentException("watermark interval cannot be set null");
```
srdo: Nit: First letter is lowercase

srdo:
@jerrypeng +1 for the changes in this PR. It's up to you whether you'd like to do the type changes + renames in this PR or a new one.

Please squash to one commit when you think this PR is complete. Thanks for the contribution.

jerrypeng:
@srdo thanks for the review.  Lets do the larger change in a separate PR.

srdo:
Thanks @jerrypeng, merged to master. The windowing classes have moved since 1.x so please open another PR if you want this to go in those versions too.

srdo:
Also opened an issue for tracking the int -> long change and variable renames so we don't forget https://issues.apache.org/jira/browse/STORM-2747.

PR #2315, opened 2017-09-08
[STORM-2732] Close the eldest writer before removing it from WritersMap

Ethanlm:
[JIRA-STORM-2732](https://issues.apache.org/jira/browse/STORM-2732)

`WritersMap` serves as an LRU cache for `maxWriters` number of `Writer`s. But it doesn't close the removed writer, which could lead to potential data loss.

The exception handling is really hard. The bolt behaviors depend on later on situations. I don't know how to address the exception handling currently so I chose to just log the error which is also what we have done elsewhere in this class.

--------

I wanted to discuss about what would happen. Please correct me if I am wrong.
#### Case 1
If `eldest.getValue().close()` failed, assume next tuple invokes SyncPolicy but the sync failed, then the bolt will call `collector.fail()` on all the tuples in `tupleBatch`. In this case, all those tuples will be transferred again, which gives the bolt another chance to write those tuples later.  However, if some tuples has been written to hdfs before `eldest.getValue().close()` failed, these tuples could be written to hdfs **twice**. 

#### Case 2
If `eldest.getValue().close()` failed, assume next tuple invokes SyncPolicy and the sync succeed, then the bolt will call `collector.ack()` on all the tuples in `tupleBatch`. Then the lost data because of  `eldest.getValue().close()` failure will not be transferred to the bolt again, which means they are **lost** permanently .

#### Case 3
If `eldest.getValue().close()` succeeded, assume next tuple invokes SyncPolicy but the sync failed, then the bolt will call `collector.fail()` on all the tuples in `tupleBatch`. Then the tuples have been written to hdfs by `eldest.getValue().close()`  will be transferred to the bolt again. Potentially they could be written **twice**.

The same issues apply to the calling to `doRotationAndRemoveWriter()`. The behaviors are not consistent.  We might want to have a better way to deal with exceptions.

### DIFF
--- a/external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/AbstractHdfsBolt.java
+++ b/external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/AbstractHdfsBolt.java
@@ -106,7 +106,7 @@ public abstract class AbstractHdfsBolt extends BaseRichBolt {
             throw new IllegalStateException("File system URL must be specified.");
         }
 
-        writers = new WritersMap(this.maxOpenFiles);
+        writers = new WritersMap(this.maxOpenFiles, collector);
 
         this.collector = collector;
         this.fileNameFormat.prepare(conf, topologyContext);
@@ -196,7 +196,7 @@ public abstract class AbstractHdfsBolt extends BaseRichBolt {
             }
 
             if (writer != null && writer.needsRotation()) {
-                    doRotationAndRemoveWriter(writerKey, writer);
+                doRotationAndRemoveWriter(writerKey, writer);
             }
         }
     }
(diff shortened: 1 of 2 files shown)


external/storm-hdfs/src/main/java/org/apache/storm/hdfs/bolt/AbstractHdfsBolt.java
```diff
@@ -317,7 +319,18 @@ public WritersMap(long maxWriters) {
...
         @Override
         protected boolean removeEldestEntry(Map.Entry<String, Writer> eldest) {
-            return this.size() > this.maxWriters;
+            if (this.size() > this.maxWriters) {
+                //The writer must be closed before removed from the map.
+                //If it failed, we might lose some data.
+                try {
+                    eldest.getValue().close();
+                } catch (IOException e) {
+                    LOG.error("Failed to close the eldest Writer");
```
revans2: Would it be possible to report the error here like we do in other places in the code?

```
this.collector.reportError(e);
```

It would also be create if you could add it to `doRotationAndRemoveAllWriters` too just so it is consistent everywhere.

Ethanlm:
Thanks @revans2  . I forgot it.  Addressed in the latest commit (and squashed).

PR #2317, opened 2017-09-10
STORM-2675 (1.x): Fix storm-kafka-client Trident spout failing to serialize…

srdo:
… meta objects to Zookeeper

1.x version of https://github.com/apache/storm/pull/2271.

The only changes should be replacing streams with regular for-each, and this version doesn't modify the storm-kafka-client-examples.


HeartSaVioR:
Built in my local and verified it passed with JDK 7. +1

PR #2319, opened 2017-09-11
[STORM-2693] Nimbus assignments promotion

danny0405:
Storm now doesn't support large cluster[ for example thousand of supervisors] very well, for our production, topology submission/killing is very ineffective when cluster grows to be large, i checkout the assignments strategy now and find that actually it can be promoted.

For assignment promotion:
1. nimbus will put the assignments in zk first then to local disk
2. when restart or HA leader trigger nimbus will recover assignments from zk to local disk
3. nimbus will tell supervisor its assignment every time through RPC every scheduling round [ only the assignments changed nodes will be notified ]
4. expect that the nimbus notification, supervisor will sync assignments at fixed time[ rpc request to nimbus ]
5. workers will sync assignments just from local supervisor [ or from zookeeper when local supervisor collapse]

assignments-promotion

I have tested it in our cluster, with the new assignments distribution mode of RPC, supervisor will response to the assignments change very fast[ milliseconds ] and efficiently [ only assignments changed nodes will be notified ], also it has the full robustness of the old zookeeper mode:
1. when nimbus collapse, workers works fine[ like the original ], when leader starts up, it will sync the assignments and start to work again
2. when supervisor goes down, workers still workers fine,[ it will sync connections from zk like the original ], when supervisor goes up, it will just sync the assignments from nimbus
3. when zk is unstable, it will not affect the assignments sync, but only the heartbeats and logical plan[ StormBase or something ]

This is my JIRA task: https://issues.apache.org/jira/browse/STORM-2693

@HeartSaVioR can you please help me to review this?


revans2:
@danny0405 conceptually this is a wonderful improvement.  I will try to take some time looking into this patch.  One of the big things we will need though is a corresponding patch for the master branch.  You don't have to do it now.  Waiting until the review are done is probably best, but I want to set the expectations appropriately.

storm-core/src/clj/org/apache/storm/daemon/nimbus.clj
```diff
@@ -961,14 +1015,25 @@
...
+    ;; grouping assignment by node to see the nodes diff, then notify nodes/supervisors to synchronize its owned assignment
+    ;; because the number of existing assignments is small for every scheduling round,
+    ;; we expect to notify supervisors at almost the same time.
+    (->> new-assignments
+         (map (fn [[tid new-assignment]]
+           (let [existing-assignment (get existing-assignments tid)]
+             (assignment-changed-nodes existing-assignment new-assignment ))))
+         (apply concat)
+         (into #{})
+         (notify-supervisors-assignments conf new-assignments))
```
revans2: I am very concerned about this.  One of our biggest issues is the timing of this loop.  It blocks scheduling a new topology, rebalancing a topology, and recovering from a failed worker or a failed supervisor.

One of the key tenants of Hadoop has been to never let the main daemon reach out to the thousands of other processes that may be on the cluster, because if one of them blocks, or there are network issues or whatever it can slow done the rest of the cluster.

I would really prefer to see a different model here.  At a minimum we need to have a thread pool that is used to notify the supervisors asynchronously, or even better we never actually reach out to the supervisors.  We just set something up in memory on nimbus and have the supervisors poll very frequently.  Nimbus is mostly single threaded so using more CPU is not a big deal.  We know that even 1,000 times a second for the DRPC servers is not really any load at all, so doing it a few times a second would not matter that much.

storm-core/src/clj/org/apache/storm/daemon/supervisor.clj
```diff
@@ -62,26 +63,17 @@
...
           (->>
             (dofor [sid (distinct storm-ids)]
                    (if-let [topo-profile-actions (.get-topology-profile-requests storm-cluster-state sid false)]
                       {sid topo-profile-actions}))
            (apply merge))]
          
-      {:assignments (into {} (for [[k v] new-assignments] [k (:data v)]))
+      {:assignments new-assignments
        :profiler-actions new-profiler-actions
+       ;; TODO: remove versions
```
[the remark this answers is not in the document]
danny0405: yes, i'm going to remove the TODO

storm-core/src/clj/org/apache/storm/daemon/supervisor.clj
```diff
@@ -758,6 +752,35 @@
...
+;; TODO: get leader address from zk directly.
+(defn assignments-from-master
+  [conf supervisor]
+  (let [client (atom nil)]
+    (try
+      (let [master-client (NimbusClient/getConfiguredClientAs conf nil)
+            _ (reset! client master-client) ;; keep a refence so we can close it
+            supervisor-assignments (.getSupervisorAssignments (.getClient master-client) (:my-hostname supervisor))]
+        (assigned-assignments-to-local! supervisor-assignments supervisor))
+      (catch Throwable e
```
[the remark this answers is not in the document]
danny0405: okey, but if we just throw it out, supervisor will restart more frequently, i have no idea now, this need a discussion

storm-core/pom.xml
```diff
@@ -331,6 +331,10 @@
             <artifactId>commons-codec</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.rocksdb</groupId>
+            <artifactId>rocksdbjni</artifactId>
+        </dependency>
```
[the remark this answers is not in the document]
danny0405: okey, this is an issue, i choose rocksdb for its good performance, but the jar conflicts really need a concern.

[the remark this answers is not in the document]
revans2: If you move the rocksdb implementation to a separate jar that people can add in on their own and set the default implementation to be in memory, then it should not be a big deal.

storm-core/src/clj/org/apache/storm/daemon/nimbus.clj
danny0405: Really thx for your suggestions, i agree with you that putting a blocking method into nimbus's main loop is not a good idea, an thread poll is a better choice.
Because only assignments changed nodes will be notified in a round, the nodes num will not be very large[ except that we submit really large topologies or submit large number of topologies at almost the same time ]
Actually supervisors will synchronize its assignments every 10 seconds, if we make nimbus mk-assignments totally asynchronous, it will be okey.

revans2: So then if we can put in a thread pool with a timeout of 5 seconds or so for each push attempt we should be able to get the best of both worlds.

danny0405: Sounds really good, i will try to make it better

HeartSaVioR:
Before looking into detail, according to design doc, as I commented to the JIRA issue, I don't see a reason to store cache to disk or so, since ZK is still a source of truth and whenever process starts it should sync up with ZK because cache may be outdated. It looks like keeping cache in memory looks sufficient.

danny0405:
@HeartSaVioR @revans2 
i have change the default backend to memory one, also i change the distribute mode to  asynchronous
i chose a rocks-db backend because when the cluster goes to large, assignments-info will takes more memory cache[ memory mode ], but if this is not a factor, we can just remote it

revans2:
@danny0405 I am not concerned about the memory for the cache, because we are already keeping all of that in memory while we schedule the topologies anyways.  If it does prove to be an issue we can address it then.

storm-core/src/clj/org/apache/storm/cluster.clj
```diff
@@ -123,6 +133,7 @@
 (def SUPERVISORS-SUBTREE (str "/" SUPERVISORS-ROOT))
 (def WORKERBEATS-SUBTREE (str "/" WORKERBEATS-ROOT))
 (def BACKPRESSURE-SUBTREE (str "/" BACKPRESSURE-ROOT))
+(def LEADERINFO-SUBTREE (str "/" LEADERINFO-ROOT))
```
revans2: Is this needed?  we already have a way of getting the leader that is built into the client.  Why do we need a second way?

storm-core/src/clj/org/apache/storm/cluster.clj
```diff
@@ -244,15 +255,19 @@
...
-  [cluster-state-spec :acls nil :context (ClusterStateContext.)]
-  (let [[solo? cluster-state] (if (instance? ClusterState cluster-state-spec)
-                                [false cluster-state-spec]
-                                [true (mk-distributed-cluster-state cluster-state-spec :auth-conf cluster-state-spec :acls acls :context context)])
+  [conf :cluster-state nil :acls nil :context (ClusterStateContext.) :backend nil]
+  (let [[solo? cluster-state] (if (and (not-nil? cluster-state) (instance? ClusterState cluster-state))
+                                [false cluster-state]
+                                [true (mk-distributed-cluster-state conf :auth-conf conf :acls acls :context context)])
+        assignments-backend (if (nil? backend)
+                              (doto (InMemoryAssignmentBackend.) (.prepare nil nil))
```
revans2: Why do we call `(.prepare nil nil)` only on the InMemoryAssignmentBackend?  It feels like we might be making it difficult to extend it in the future.  From what I have seen the assignments dir is a subdirectory of storm-local.  Could we just make prepare take the conf, and then we can pass in the conf we already have.

storm-core/src/clj/org/apache/storm/cluster.clj
```diff
@@ -244,15 +255,19 @@
 
 ;; Watches should be used for optimization. When ZK is reconnecting, they're not guaranteed to be called.
 (defnk mk-storm-cluster-state
-  [cluster-state-spec :acls nil :context (ClusterStateContext.)]
-  (let [[solo? cluster-state] (if (instance? ClusterState cluster-state-spec)
-                                [false cluster-state-spec]
-                                [true (mk-distributed-cluster-state cluster-state-spec :auth-conf cluster-state-spec :acls acls :context context)])
+  [conf :cluster-state nil :acls nil :context (ClusterStateContext.) :backend nil]
```
revans2: Could we rename backend to assignments-backend?  backend feels too generic for what this actually is.

storm-core/src/jvm/org/apache/storm/assignments/AssignmentDistributionService.java
```diff
@@ -0,0 +1,232 @@
...
+                    } else {
+                        // service is off now just interrupt it.
+                        Thread.currentThread().interrupt();
+                    }
+                }
+            }
+        }
+
+        private void sendAssignmentsToNode(NodeAssignments assignments) {
+            SupervisorClient client = SupervisorClient.getConfiguredClient(service.getConf(), assignments.getNode());
```
revans2: SupervisorClient should be AutoClosable, so we can use java to ensure it is closed.

storm-core/src/jvm/org/apache/storm/assignments/AssignmentDistributionService.java
```diff
@@ -0,0 +1,232 @@
...
+            this.service.awaitTermination(10l, TimeUnit.SECONDS);
+        } catch (InterruptedException e) {
+            LOG.error("Failed to close assignments distribute service");
+        }
+        this.assignmentsQueue = null;
+    }
+
+    public void addAssignmentsForNode(String node, SupervisorAssignments assignments) {
+        try {
+            boolean success = nextQueue().offer(NodeAssignments.getInstance(node, assignments), 5l, TimeUnit.SECONDS);
```
revans2: So we have a 5 second timeout on offering, instead of a 5 second timeout on connecting and sending?

storm-core/src/jvm/org/apache/storm/assignments/AssignmentDistributionService.java
revans2: Can we do something where if the send is too old that we don't do it?

Because we have 2 ways to get this information to the client I can see us having race conditions where a supervisor sees State A followed by State B and then back again to State A.  I would prefer to avoid something like that, and timeout the send if it is more than 10 seconds old, because the supervisor should be checking every 10 seconds anyways.

storm-core/src/clj/org/apache/storm/cluster.clj
danny0405: okey, i will rename it

storm-core/src/clj/org/apache/storm/cluster.clj
danny0405: Actually, this is not really needed, but if we always get leader from nimbus, we will make 2 RPC requests every time we contact with the leader [supervisor to nimbus/], instead get it from zookeeper directly is a better way.

storm-core/src/clj/org/apache/storm/cluster.clj
danny0405: okey, i will make it configurable

storm-core/src/jvm/org/apache/storm/assignments/AssignmentDistributionService.java
danny0405: i also config the SupervisorClient timeout to 5s timeout, which is taken in ThriftConnectionType.

danny0405:
@revans2 
I have removed the rocks db backend

revans2:
@danny0405 the most recent changes look good, but I really want to spend some time to dig into the code.

danny0405:
@revans2 really thx for your review work

danny0405:
@revans2 hi, is this PR okey for merging now? I have another heartbeats promotion which is based on this, and i expect for your review too

HeartSaVioR:
@danny0405 
To be clear, all the pull requests should be merged into master first, and merged back to 1.x version line. So to integrate your great effort into Apache Storm, we need PR against master branch. Actually transition to Java makes me easier to review given that I'm still not familiar with Clojure.

HeartSaVioR:
Also given that the change is significant, we may want to integrate only 2.0.0 instead of having this in 1.x version line.

danny0405:
@HeartSaVioR ok, i will do the 2.0.0 version, the promotion is running on our production cluster for a while and it works great, i will start to work on the 2.0.0 branch now, is the master branch 2.0 version right?

HeartSaVioR:
@danny0405 Yes right. Thank you for your patience and understanding the policy of Storm contribution.

danny0405:
@HeartSaVioR i have made a PR for storm 2.0, really thx for your review work, this is the link:
https://github.com/apache/storm/pull/2431/

danny0405:
@HeartSaVioR @revans2 i have moved all the features to
https://github.com/apache/storm/pull/2389 for 1.1.x-branch
https://github.com/apache/storm/pull/2433 for 2.0-branch
so this is closed now

PR #2325, opened 2017-09-14
[STORM-2738] The number of ackers should default to the number of actual running workers on RAS cluster

Ethanlm:
**Problem**:
If `topology.acker.executors` is not set, the number of ackers will be equal to `topology.workers`. But on RAS cluster, we don't set `topology.workers` because the number of workers will be determined by the scheduler. So in this case, the number of ackers will always be 1.
For example, there are 5 workers but only 1 acker.
image

**Analysis**:
The number of ackers has to be computed before scheduling happens, so it knows how to schedule the topology. The number of workers is not set until the topology is scheduled, so it is a bit of a chicken and egg problem.
**Solution**:
We could probably use the total amount of requested memory when the topology is submitted divided by the memory per worker to get an estimate that is better than 1.

JIRA at https://issues.apache.org/jira/browse/STORM-2738

### DIFF
--- a/storm-client/src/jvm/org/apache/storm/utils/Utils.java
+++ b/storm-client/src/jvm/org/apache/storm/utils/Utils.java
@@ -1501,4 +1501,12 @@ public class Utils {
     public static boolean isLocalhostAddress(String address) {
         return LOCALHOST_ADDRESSES.contains(address);
     }
+
+    public static <K, V> Map<K, V> merge(Map<? extends K, ? extends V> first, Map<? extends K, ? extends V> other) {
+        Map<K, V> ret = new HashMap<>(first);
+        if (other != null) {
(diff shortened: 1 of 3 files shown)


Ethanlm:
Two tests:
**Confg**
`topology.workers` not set;
`topology.acker.executors` not set.
`topology.component.resources.onheap.memory.mb` is default: 128.
`worker.heap.memory.mb` is default: 768.

**Test1**
image
numOfAckerExec = Math.ceil(12 * 128 / 768) = 2. 

**Test2**

image

numOfAckerExec =Math.ceil(25 * 128 / 768) = 5.

HeartSaVioR:
The estimation makes sense. +1
@Ethanlm Could you resolve merge conflict? Thanks in advance.

PR #2324, opened 2017-09-12
[STORM-2730] Add in config options for acker cpu and memory

Ethanlm:
[JIRA-STORM-2730](https://issues.apache.org/jira/browse/STORM-2730)
 
Add config options for acker cpu and memory.

A test:
Config: 
 `topology.acker.resources.onheap.memory.mb: 100` `topology.component.resources.onheap.memory.mb: 128` 

image

PR #2326, opened 2017-09-15
[STORM-2741] Add in config options for metrics consumer cpu and memory

Ethanlm:
[jira-STORM-2741](https://issues.apache.org/jira/browse/STORM-2741)

Similar to https://github.com/apache/storm/pull/2324 , add cpu and memory configs for metrics consumer.

### DIFF
--- a/storm-client/src/jvm/org/apache/storm/Constants.java
+++ b/storm-client/src/jvm/org/apache/storm/Constants.java
@@ -30,7 +30,7 @@ public class Constants {
     public static final List<Long> SYSTEM_EXECUTOR_ID = Arrays.asList(-1L, -1L);
     public static final String SYSTEM_COMPONENT_ID = "__system";
     public static final String SYSTEM_TICK_STREAM_ID = "__tick";
-    public static final String METRICS_COMPONENT_ID_PREFIX = "__metrics";
+    public static final String METRICS_COMPONENT_ID_PREFIX = "__metrics_";
     public static final String METRICS_STREAM_ID = "__metrics";
     public static final String METRICS_TICK_STREAM_ID = "__metrics_tick";
     public static final String CREDENTIALS_CHANGED_STREAM_ID = "__credentials";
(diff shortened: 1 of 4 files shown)


Ethanlm:
A test:
Config: 
```
topology.metrics.consumer.register:
  - class: "org.apache.storm.metric.LoggingMetricsConsumer"
    parallelism.hint: 1
  - class: "org.apache.storm.metric.LoggingMetricsConsumer"
    parallelism.hint: 1

topology.metrics.consumer.resources.onheap.memory.mb: 100
topology.component.resources.onheap.memory.mb: 128
```

image

storm-client/src/jvm/org/apache/storm/Constants.java
```diff
@@ -30,7 +30,7 @@
     public static final List<Long> SYSTEM_EXECUTOR_ID = Arrays.asList(-1L, -1L);
     public static final String SYSTEM_COMPONENT_ID = "__system";
     public static final String SYSTEM_TICK_STREAM_ID = "__tick";
-    public static final String METRICS_COMPONENT_ID_PREFIX = "__metrics";
+    public static final String METRICS_COMPONENT_ID_PREFIX = "__metrics_";
```
revans2: This is a minor change, but it might impact some people.  We probably want to just put this into 2.x for now.  If we really want it in 1.x we can decide how much of an impact it is (probably minor, but I want to see what others thing about it).

Ethanlm: Yea... But I made this change because https://github.com/apache/storm/blob/master/docs/Metrics.md says, 
```
The parallelism for each Bolt is set to parallelism.hint and component id for that Bolt is set to __metrics_<metrics consumer class name>
```

revans2: That is fine, I think this is a really good fix.  I am just saying that it might make it difficult for someone to find say the LoggingMetricsConsumer in an automated way, and so if we do make this change in 1.x we should do it thoughtfully.

Ethanlm: Yes sure. Thanks for the review.

PR #2328, opened 2017-09-18
Quick fix: report cassandra error messages to logger

morokosi:
`CassandraState::batchRetrieve` does not report specific error messages when binding values or retrieving results goes wrong.

I think we should make exception message available to log in order to make debugging easier. Thank you.


HeartSaVioR:
+1 Nice finding.

PR #2330, opened 2017-09-18
STORM-2742: Logviewer leaking file descriptors

knusbaum:


knusbaum:
I just added a try {...} finally {...} block. Not sure why the diff is so ugly.

HeartSaVioR:
+1 Maybe it is due to the reorganization of line sequences, but doesn't matter anyway.

PR #2331, opened 2017-09-19
[STORM-2743] Add logging and metrics to monitor how long scheduling is taking

Ethanlm:
A trivial improvement: Add logging to monitor how long scheduling is taking

https://issues.apache.org/jira/browse/STORM-2743

A test:
image

### DIFF
--- a/storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
+++ b/storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
@@ -18,6 +18,8 @@
 
 package org.apache.storm.daemon.nimbus;
 
+import com.codahale.metrics.ExponentiallyDecayingReservoir;
+import com.codahale.metrics.Histogram;
 import com.codahale.metrics.Meter;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableMap;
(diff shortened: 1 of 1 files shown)


jerrypeng:
+1 but perhaps its work while looking into using something like dropwizard metrics:

http://metrics.dropwizard.io/3.2.3/manual/core.html

HeartSaVioR:
+1 and agree @jerrypeng that better to 'also' record with histogram via StormMetricsRegistry.registerHistogram() like Nimbus records for other operations with Meter.

Ethanlm:
@jerrypeng @HeartSaVioR  Thanks for the review.  Just one question, do we use two Histograms one for the number of the topology and the other one for the time the scheduling takes? I didn't find any way to combine these two together in one Histogram. But I think we might not want to deal with these two separately.

HeartSaVioR:
@Ethanlm 
I'm OK to have two metrics and IMO that is right way since we measure two different things, one for topology count, another one for topology schedule delay.

revans2:
@Ethanlm I am fine with having 2 histograms, but quite honestly the number of topologies is not that interesting or helpful and I would rather drop it.

The number of topologies has relatively little impact on how long scheduling takes, because for the most part we are going to cache the results, and just skip over topologies that are already scheduled.

For me I would like to see two histograms.  One for how long it took to schedule all of the topologies in the loop.  This indicates how responsive we are likely to be when a worker crashes and it needs to be rescheduled or when a topology is launched and it needs to be scheduled the first time.

The second one I would be interested in is a histogram for how long it took to schedule each individual topology.  Sadly that one is much harder, because it would have to be something each scheduler reported on their own.

Ethanlm:
A test:
I monitored the metrics through JMX. And after a while, I submitted a topology.

image

HeartSaVioR:
+1

PR #2337, opened 2017-09-21
[STORM2751] Removing AsyncLoggerContext from Supervisor Configuration

kishorvpatil:
If disk can not keep it up with all logging, the {{AsyncLoggingContext}} causes large heap memory be utilized causing the JVM to churn CPU.

### DIFF
--- a/bin/storm.py
+++ b/bin/storm.py
@@ -739,7 +739,6 @@ def supervisor(klass="org.apache.storm.daemon.supervisor.Supervisor"):
     cppaths = [CLUSTER_CONF_DIR]
     jvmopts = parse_args(confvalue("supervisor.childopts", cppaths)) + [
         "-Dlogfile.name=" + STORM_SUPERVISOR_LOG_FILE,
-        "-DLog4jContextSelector=org.apache.logging.log4j.core.async.AsyncLoggerContextSelector",
         "-Dlog4j.configurationFile=" + os.path.join(get_log4j2_conf_dir(), "cluster.xml"),
     ]
     exec_storm_class(


HeartSaVioR:
@kishorvpatil 
Are you observing the issue only from Supervisor?
I understand what the patch would want to address, but just to be sure about that patch because we have another daemons as well.
If it only occurs from Supervisor, we may want to revisit Supervisor to check whether it is too verbose.

kishorvpatil:
@HeartSaVioR 
I am noticed it on supervisors, but I am suspecting topologies being too verbose making many workers adding load on disk. I am not sure if supervisor is too verbose. But with limited memory ( 256MB) assigned to it, I noticed it only only on supervisor.

kishorvpatil:
@HeartSaVioR I do not notice this on logviewer but clearly it do not have so much activity or logging for that matter.

bin/storm.py
```diff
@@ -739,7 +739,6 @@ def supervisor(klass="org.apache.storm.daemon.supervisor.Supervisor"):
     cppaths = [CLUSTER_CONF_DIR]
     jvmopts = parse_args(confvalue("supervisor.childopts", cppaths)) + [
         "-Dlogfile.name=" + STORM_SUPERVISOR_LOG_FILE,
-        "-DLog4jContextSelector=org.apache.logging.log4j.core.async.AsyncLoggerContextSelector",
```
revans2: Do we want to remove this for the UI and other daemons too?

kishorvpatil: synchronized logging might slow down UI response time. May be if we notice this in future we can remove it.

HeartSaVioR:
+1 from me too.

PR #2338, opened 2017-09-21
[STORM-2744] add in restart timeout for backpressure

Ethanlm:
[JIRA-STORM-2744](https://issues.apache.org/jira/browse/STORM-2744)

Backpressure mechanism relies on zookeeper to transfer signals. 

If at some point, the topology wants to turn off the throttle but zookeeper failed to delete the backpressure node, then the throttle will be on. And in that case, it's very likely that the znode will never be deleted and thus the throttle will always be on. The spouts will stop indefinitely. 

We want to add a restart timeout so that even the above case happens, the spouts can continue to work.

### DIFF
--- a/storm-client/src/jvm/org/apache/storm/cluster/StormClusterStateImpl.java
+++ b/storm-client/src/jvm/org/apache/storm/cluster/StormClusterStateImpl.java
@@ -32,6 +32,7 @@ import org.apache.zookeeper.data.ACL;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.nio.ByteBuffer;
 import java.security.NoSuchAlgorithmException;
 import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
@@ -423,24 +424,26 @@ public class StormClusterStateImpl implements IStormClusterState {
     }
 
     /**
-     * if znode exists and to be not on?, delete; if exists and on?, do nothing; if not exists and to be on?, create; if not exists and not on?, do nothing;
+     * if znode exists and timestamp is 0?, delete; if exists and timestamp is larger than 0?, do nothing;
+     * if not exists and timestamp is larger than 0?, create the node and set the timestamp; if not exists and timestamp is 0?, do nothing;
      * 
      * @param stormId
      * @param node
      * @param port
-     * @param on
+     * @param timestamp
      */
     @Override
-    public void workerBackpressure(String stormId, String node, Long port, boolean on) {
+    public void workerBackpressure(String stormId, String node, Long port, long timestamp) {
         String path = ClusterUtils.backpressurePath(stormId, node, port);
         boolean existed = stateStorage.node_exists(path, false);
         if (existed) {
-            if (on == false)
+            if (timestamp == 0) {
                 stateStorage.delete_node(path);
-
+            }
         } else {
-            if (on == true) {
-                stateStorage.set_ephemeral_node(path, null, acls);
+            if (timestamp > 0) {
+                byte[] data = ByteBuffer.allocate(Long.BYTES).putLong(timestamp).array();
+                stateStorage.set_ephemeral_node(path, data, acls);
             }
         }
     }
@@ -448,7 +451,8 @@ public class StormClusterStateImpl implements IStormClusterState {
     /**
      * Check whether a topology is in throttle-on status or not:
      * if the backpresure/storm-id dir is not empty, this topology has throttle-on, otherwise throttle-off.
-     * 
+     * But if the backpresure/storm-id dir is not empty and has not been updated for more than 30s, we treat it as throttle-off.
+     * This will prevent the spouts from getting stuck indefinitely if something wrong happens.
      * @param stormId
      * @param callback
      * @return
@@ -459,14 +463,15 @@ public class StormClusterStateImpl implements IStormClusterState {
             backPressureCallback.put(stormId, callback);
         }
         String path = ClusterUtils.backpressureStormRoot(stormId);
-        List<String> childrens = null;
+        long mostRecentTimestamp = 0;
         if(stateStorage.node_exists(path, false)) {
-            childrens = stateStorage.get_children(path, callback != null);
-        } else {
-            childrens = new ArrayList<>();
+            List<String> children = stateStorage.get_children(path, callback != null);
+            mostRecentTimestamp = children.stream().map(childPath -> stateStorage.get_data(ClusterUtils.backpressurePath(stormId, childPath), false))
+                    .filter(data -> data != null).mapToLong(data -> ByteBuffer.wrap(data).getLong()).max().orElse(0);
         }
-        return childrens.size() > 0;
-
+        boolean ret =  ((System.currentTimeMillis() - mostRecentTimestamp) < 30000);
+        LOG.debug("topology backpressure is {}", ret ? "on" : "off");
+        return ret;
     }
 
     @Override
--- a/storm-client/src/jvm/org/apache/storm/cluster/StormClusterStateImpl.java
+++ b/storm-client/src/jvm/org/apache/storm/cluster/StormClusterStateImpl.java
@@ -424,13 +424,14 @@ public class StormClusterStateImpl implements IStormClusterState {
     }
 
     /**
-     * if znode exists and timestamp is 0?, delete; if exists and timestamp is larger than 0?, do nothing;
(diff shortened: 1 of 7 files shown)


Ethanlm:
I did some simulation test.

The command I ran is:
```
storm jar /tmp/storm-loadgen-2.0.0-SNAPSHOT.jar org.apache.storm.loadgen.ThroughputVsLatency --rate 20000 --counters  12 --splitters 6 --spouts 3 -t 60 -c topology.workers=3
```

**The behavior before the code change:**

At about timestamp 241, I created a znode under backpressure/topology-id manually. Quickly the rate (tuple/s) dropped and became 0. At about timestamp 391, I deleted the znode manually. The rate increased and the spouts were catching up. This is the expected behavior of original storm.

image



**The behavior after the code change:**

At about timestamp 211, I created a znode (with timestamp set to `System.currentTimeMillis()`) under backpressure/topology-id. The rate dropped quickly. Then after 30s, the spouts were catching up. Even though the znode I created still exists in zookeeper, it's no long valid because its timestamp is out of date.

image

storm-client/src/jvm/org/apache/storm/daemon/worker/Worker.java
```diff
@@ -406,25 +406,36 @@ private WorkerBackpressureCallback mkBackpressureHandler() {
...
+                    long prevBackpressureTimestamp = workerState.backpressure.get();
+                    long currTimestamp = System.currentTimeMillis();
+                    long currBackpressureTimestamp = 0;
+                    // the backpressure flag is true if at least one of the disruptor queues has throttle-on
+                    boolean backpressureFlag = workerState.transferQueue.getThrottleOn() || (executors.stream()
+                            .map(IRunningExecutor::getBackPressureFlag).reduce((op1, op2) -> (op1 || op2)).get());
+
+                    if (backpressureFlag) {
+                        // update the backpressure timestamp every 15 seconds
+                        if ((currTimestamp - prevBackpressureTimestamp) > 15000) {
```
govind-menon: Could we make this configurable? If not a constant somewhere

storm-client/src/jvm/org/apache/storm/cluster/StormClusterStateImpl.java
```diff
@@ -423,32 +424,35 @@ public void supervisorHeartbeat(String supervisorId, SupervisorInfo info) {
     }
 
     /**
-     * if znode exists and to be not on?, delete; if exists and on?, do nothing; if not exists and to be on?, create; if not exists and not on?, do nothing;
+     * if znode exists and timestamp is 0?, delete; if exists and timestamp is larger than 0?, do nothing;
```
srdo: Nit: What are the question marks after 0 for?

storm-client/src/jvm/org/apache/storm/cluster/StormClusterStateImpl.java
```diff
@@ -459,14 +463,15 @@ public boolean topologyBackpressure(String stormId, Runnable callback) {
...
         }
         String path = ClusterUtils.backpressureStormRoot(stormId);
-        List<String> childrens = null;
+        long mostRecentTimestamp = 0;
         if(stateStorage.node_exists(path, false)) {
-            childrens = stateStorage.get_children(path, callback != null);
-        } else {
-            childrens = new ArrayList<>();
+            List<String> children = stateStorage.get_children(path, callback != null);
+            mostRecentTimestamp = children.stream().map(childPath -> stateStorage.get_data(ClusterUtils.backpressurePath(stormId, childPath), false))
```
srdo: Please put a newline for each part of the stream pipeline, it gets a little hard to read in the second line here.

storm-client/src/jvm/org/apache/storm/cluster/StormClusterStateImpl.java
```diff
@@ -459,14 +463,15 @@ public boolean topologyBackpressure(String stormId, Runnable callback) {
...
+                    .filter(data -> data != null).mapToLong(data -> ByteBuffer.wrap(data).getLong()).max().orElse(0);
         }
-        return childrens.size() > 0;
-
+        boolean ret =  ((System.currentTimeMillis() - mostRecentTimestamp) < 30000);
```
srdo: I agree with @govind-menon, it would be good to make some constants (or config if necessary) for these.

storm-client/src/jvm/org/apache/storm/cluster/StormClusterStateImpl.java
```diff
@@ -423,32 +424,35 @@ public void supervisorHeartbeat(String supervisorId, SupervisorInfo info) {
...
-    public void workerBackpressure(String stormId, String node, Long port, boolean on) {
+    public void workerBackpressure(String stormId, String node, Long port, long timestamp) {
         String path = ClusterUtils.backpressurePath(stormId, node, port);
         boolean existed = stateStorage.node_exists(path, false);
         if (existed) {
-            if (on == false)
+            if (timestamp == 0) {
                 stateStorage.delete_node(path);
-
+            }
```
srdo: Shouldn't we update the timestamp if timestamp != 0 here?

storm-client/src/jvm/org/apache/storm/cluster/StormClusterStateImpl.java
```diff
@@ -423,32 +424,35 @@ public void supervisorHeartbeat(String supervisorId, SupervisorInfo info) {
...
     /**
-     * if znode exists and to be not on?, delete; if exists and on?, do nothing; if not exists and to be on?, create; if not exists and not on?, do nothing;
+     * if znode exists and timestamp is 0?, delete; if exists and timestamp is larger than 0?, do nothing;
+     * if not exists and timestamp is larger than 0?, create the node and set the timestamp; if not exists and timestamp is 0?, do nothing;
      * 
      * @param stormId
      * @param node
      * @param port
-     * @param on
+     * @param timestamp
```
srdo: It might be nicer to split this method into a create and delete variant, but we should at least put in the javadoc that timestamp 0 means "turn off"

storm-client/src/jvm/org/apache/storm/cluster/StormClusterStateImpl.java
knusbaum: The question mark on the `on` in the old comment appears to be left-over from the clojure translation where the on argument was spelled `on?`.

It should go away in any new comment.

Ethanlm:
@govind-menon @srdo @knusbaum  Thanks for the review. Will address them soon

Ethanlm:
@govind-menon @srdo  I think I have address all your comments. I tested it with different configs. Could you please take another look when you get a chance? Thanks

storm-client/src/jvm/org/apache/storm/cluster/StormClusterStateImpl.java
```diff
@@ -423,50 +424,63 @@ public void supervisorHeartbeat(String supervisorId, SupervisorInfo info) {
...
+     * @param timestamp The backpressure timestamp. 0 means turning off the worker backpressure
      */
     @Override
-    public void workerBackpressure(String stormId, String node, Long port, boolean on) {
+    public void workerBackpressure(String stormId, String node, Long port, long timestamp) {
         String path = ClusterUtils.backpressurePath(stormId, node, port);
         boolean existed = stateStorage.node_exists(path, false);
         if (existed) {
-            if (on == false)
+            if (timestamp == 0) {
```
revans2: nit: Can we make this `<= 0` just as defensive programming?

storm-client/src/jvm/org/apache/storm/cluster/StormClusterStateImpl.java
```diff
@@ -423,50 +424,63 @@ public void supervisorHeartbeat(String supervisorId, SupervisorInfo info) {
...
-
+            List<String> children = stateStorage.get_children(path, callback != null);
+            mostRecentTimestamp = children.stream()
+                    .map(childPath -> stateStorage.get_data(ClusterUtils.backpressurePath(stormId, childPath), false))
+                    .filter(data -> data != null)
+                    .mapToLong(data -> ByteBuffer.wrap(data).getLong())
+                    .max()
+                    .orElse(0);
+        }
+        boolean ret =  ((System.currentTimeMillis() - mostRecentTimestamp) < timeoutMs);
```
revans2: nit extra space after =

revans2:
@govind-menon @srdo are you OK with the changes?

srdo:
+1, thanks @Ethanlm.

PR #2340, opened 2017-09-21
[STORM-2753] Avoid shutting down netty server on netty exception

Ethanlm:
[JIRA-STORM-2753](https://issues.apache.org/jira/browse/STORM-2753)

`NettyUncaughtExceptionHandler` should only exit if there is an `Error`


Ethanlm:
Test failed in org.apache.storm.st.tests.window.TumblingWindowTest.testTumbleCount seems unrelated.

HeartSaVioR:
Storm is designed as fail-fast, and all exceptions which we should catch and handle should not go there. So if it reaches to NettyUncaughtExceptionHandler, I think there's nothing to do.

Ethanlm:
Per @revans2 's suggestion, I moved this commit to https://github.com/apache/storm/pull/2341

PR #2341, opened 2017-09-21
[STORM-2754] Not killing on exceptions in other threads

Ethanlm:
[JIRA-STORM-2754](https://issues.apache.org/jira/browse/STORM-2754)

We probably don't want to kill the process if the exceptions are from other threads

### DIFF
--- a/storm-client/src/jvm/org/apache/storm/messaging/netty/NettyUncaughtExceptionHandler.java
+++ b/storm-client/src/jvm/org/apache/storm/messaging/netty/NettyUncaughtExceptionHandler.java
@@ -15,6 +15,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 package org.apache.storm.messaging.netty;
 
 import org.apache.storm.utils.Utils;
@@ -33,10 +34,9 @@ public class NettyUncaughtExceptionHandler implements Thread.UncaughtExceptionHa
 
         try {
             Utils.handleUncaughtException(e);
-        } catch (Throwable throwable) {
-            LOG.error("Exception thrown while handling uncaught exception " + throwable.getCause());
+        } catch (Error error) {
+            LOG.error("Exception thrown while handling uncaught exception " + error.getCause());
+            Runtime.getRuntime().exit(1);
         }
-        LOG.info("Received error in netty thread.. terminating server...");
-        Runtime.getRuntime().exit(1);
     }
 }
--- a/storm-client/src/jvm/org/apache/storm/utils/Utils.java
+++ b/storm-client/src/jvm/org/apache/storm/utils/Utils.java
@@ -534,29 +534,43 @@ public class Utils {
                 && !((String)conf.get(Config.STORM_ZOOKEEPER_TOPOLOGY_AUTH_SCHEME)).isEmpty());
     }
 
-    public static void handleUncaughtException(Throwable t) {
-        handleUncaughtException(t, defaultAllowedExceptions);
+    public static void handleUncaughtExceptionWithoutKillingProcess(Throwable t) {
+        handleUncaughtException(t, defaultAllowedExceptions, true);
     }
 
     public static void handleUncaughtException(Throwable t, Set<Class> allowedExceptions) {
+        handleUncaughtException(t, allowedExceptions, false);
+    }
+
+    public static void handleUncaughtException(Throwable t) {
+        handleUncaughtException(t, defaultAllowedExceptions, false);
+    }
+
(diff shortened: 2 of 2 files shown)


HeartSaVioR:
Same thought on this: fail-fast if we can't handle the exception. But I'm OK to see others' opinions as well.

revans2:
@Ethanlm I think we want to combine STORM-2753 with this into a single pull request.

@HeartSaVioR 

I think I need to give a little background on this so it makes since, and I am not sure this is the best solution which is why I would like to see the two combined into a single pull request.

We ran into some issues where pacemaker servers were being killed because of a client being shot at the wrong time, or security scans, etc.  The issue was that the uncaught exception handler was not what we wanted for the pacemaker server.  We then adjusted some things and ended up with an end user that had their topology crashing periodically because an exception that was being thrown on a background thread, and ignored previously, was now fatal.

So the goal with these was to find a balance.  We want the default uncaught exception handler for all threads to be a little bit more lenient.  But for daemon threads we already handle exceptions very strictly so the default UncaughtExceptionHandler is not needed here.

But thinking about it now perhaps what we want is something that is a bit more explicit about what is happening.  For example we can perhaps have a separate default exception handler for the worker from everything else.

storm-client/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -865,7 +874,7 @@ public static void setupDefaultUncaughtExceptionHandler() {
         Thread.setDefaultUncaughtExceptionHandler(new Thread.UncaughtExceptionHandler() {
                 public void uncaughtException(Thread thread, Throwable thrown) {
                     try {
-                        handleUncaughtException(thrown);
+                        handleUncaughtExceptionWithoutKillingProcess(thrown);
```
HeartSaVioR: This will swallow all the `Throwable`s including `Error`s excluding OOME. Is it intended? We may still want to make it crash when there's `Error` thrown from somewhere, given that it normally indicates unrecoverable state.

storm-client/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -550,12 +558,13 @@ public static void handleUncaughtException(Throwable t, Set<Class> allowedExcept
             }
         }
 
-        if(allowedExceptions.contains(t.getClass())) {
+        if(alwaysSwallow || allowedExceptions.contains(t.getClass())) {
```
HeartSaVioR: Even `alwaysSwallow` is set to true, we may still want to throw Error on `Error`s.

Ethanlm:
@HeartSaVioR Sorry for the late reply. I made the changes. Could you please review again? This also fix a  potential NPE issue.

HeartSaVioR:
+1

storm-client/src/jvm/org/apache/storm/messaging/netty/NettyUncaughtExceptionHandler.java
```diff
@@ -33,10 +34,9 @@ public void uncaughtException(Thread t, Throwable e) {
 
         try {
             Utils.handleUncaughtException(e);
-        } catch (Throwable throwable) {
-            LOG.error("Exception thrown while handling uncaught exception " + throwable.getCause());
+        } catch (Error error) {
+            LOG.error("Exception thrown while handling uncaught exception " + error.getCause());
+            Runtime.getRuntime().exit(1);
         }
-        LOG.info("Received error in netty thread.. terminating server...");
```
srdo: If a thread hits the exception handler will it still die? If so who is responsible for replacing it? As far as I can tell this handler is used by a HashedWheelTimer in Context, and it doesn't look to me like that class will replace a dead thread.

storm-client/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -865,7 +879,7 @@ public static void setupDefaultUncaughtExceptionHandler() {
         Thread.setDefaultUncaughtExceptionHandler(new Thread.UncaughtExceptionHandler() {
                 public void uncaughtException(Thread thread, Throwable thrown) {
                     try {
-                        handleUncaughtException(thrown);
+                        handleUncaughtExceptionWithoutKillingProcess(thrown);
```
srdo: Same question, if a thread gets an exception and the process is not shut down, will the thread be replaced?

storm-client/src/jvm/org/apache/storm/messaging/netty/NettyUncaughtExceptionHandler.java
govind-menon: It looks like the HashedWheelTimer will not be replaced - I'll take a second pass at this to confirm

PR #2343, opened 2017-09-25
STORM-2083 Blacklist scheduler

HeartSaVioR:
This is rebased version of #1674 with addressing latest review comments from me.

Sure the credit for providing great stuff should go to @nilday. :)

Since now I can't give +1 to my patch, it would be really appreciated if someone could review and do some manual tests.

Thanks in advance.

This closes #1674

### DIFF
--- a/storm-server/pom.xml
+++ b/storm-server/pom.xml
@@ -136,7 +136,7 @@
                 <artifactId>maven-checkstyle-plugin</artifactId>
                 <!--Note - the version would be inherited-->
                 <configuration>
-                    <maxAllowedViolations>3626</maxAllowedViolations>
+                    <maxAllowedViolations>4000</maxAllowedViolations>
                 </configuration>
             </plugin>
             <plugin>
--- a/storm-server/pom.xml
+++ b/storm-server/pom.xml
@@ -136,7 +136,7 @@
                 <artifactId>maven-checkstyle-plugin</artifactId>
                 <!--Note - the version would be inherited-->
                 <configuration>
-                    <maxAllowedViolations>4000</maxAllowedViolations>
+                    <maxAllowedViolations>3590</maxAllowedViolations>
                 </configuration>
             </plugin>
             <plugin>
--- a/storm-server/src/main/java/org/apache/storm/DaemonConfig.java
+++ b/storm-server/src/main/java/org/apache/storm/DaemonConfig.java
@@ -18,6 +18,13 @@
 
 package org.apache.storm;
 
+import org.apache.storm.scheduler.blacklist.reporters.IReporter;
+import org.apache.storm.scheduler.blacklist.strategies.IBlacklistStrategy;
+import org.apache.storm.container.ResourceIsolationInterface;
+import org.apache.storm.nimbus.ITopologyActionNotifierPlugin;
+import org.apache.storm.scheduler.resource.strategies.eviction.IEvictionStrategy;
+import org.apache.storm.scheduler.resource.strategies.priority.ISchedulingPriorityStrategy;
+import org.apache.storm.scheduler.resource.strategies.scheduling.IStrategy;
 import org.apache.storm.container.ResourceIsolationInterface;
 import org.apache.storm.nimbus.ITopologyActionNotifierPlugin;
 import org.apache.storm.scheduler.resource.strategies.eviction.IEvictionStrategy;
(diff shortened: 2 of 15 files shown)


storm-server/pom.xml
```diff
@@ -136,7 +136,7 @@
                 <artifactId>maven-checkstyle-plugin</artifactId>
                 <!--Note - the version would be inherited-->
                 <configuration>
-                    <maxAllowedViolations>3626</maxAllowedViolations>
+                    <maxAllowedViolations>4000</maxAllowedViolations>
```
revans2: Can we try and fix the violations instead of boosting the number?

storm-server/src/main/java/org/apache/storm/DaemonConfig.java
```diff
@@ -103,6 +110,38 @@
...
     /**
+     * The number of seconds that the blacklist scheduler will concern of bad slots or supervisors
+     */
+    @isInteger
+    public static final String BLACKLIST_SCHEDULER_TOLERANCE_TIME = "blacklist.scheduler.tolerance.time.secs";
+
+    /**
+     * The number of hit count that will trigger blacklist in tolerance time
+     */
+    @isInteger
```
revans2: Is a negative count allowed?  If not we should also mark it as `@isPositiveNumber`

Same for tolerance time and resume time.

storm-server/src/main/java/org/apache/storm/scheduler/blacklist/BlacklistScheduler.java
```diff
@@ -0,0 +1,248 @@
...
+public class BlacklistScheduler implements IScheduler {
+    private static final Logger LOG = LoggerFactory.getLogger(BlacklistScheduler.class);
+
+    public static final int DEFAULT_BLACKLIST_SCHEDULER_RESUME_TIME = 1800;
+    public static final int DEFAULT_BLACKLIST_SCHEDULER_TOLERANCE_COUNT = 3;
+    public static final int DEFAULT_BLACKLIST_SCHEDULER_TOLERANCE_TIME = 300;
+
+    private final IScheduler underlyingScheduler;
+    @SuppressWarnings("rawtypes")
+    private Map _conf;
```
revans2: Can we change this to `conf` which will fix one of the violations.

storm-server/src/main/java/org/apache/storm/scheduler/blacklist/BlacklistScheduler.java
```diff
@@ -0,0 +1,248 @@
...
+                if (slots != null) { // slots will be null while supervisor has been removed from cached supervisors
+                    slots.remove(slot);
+                    cachedSupervisors.put(supervisorKey, slots);
+                }
+                LOG.info("Worker slot {} was never back to normal during tolerance period, probably dead. Will be removed from cache.", workerSlot);
+            }
+        }
+    }
+
+    private Object initializeInstance(String className, String representation) {
```
revans2: I think there is a function in Utils, or ServerUtils that can do this for us.

storm-server/src/main/java/org/apache/storm/scheduler/blacklist/Sets.java
```diff
@@ -0,0 +1,58 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.storm.scheduler.blacklist;
+
+import java.util.HashSet;
+import java.util.Set;
+
+public class Sets {
```
revans2: We do set logic in a number of places could you file a follow on JIRA for someone to go through and make it all common?

storm-server/src/main/java/org/apache/storm/scheduler/blacklist/strategies/DefaultBlacklistStrategy.java
```diff
@@ -0,0 +1,169 @@
...
+import java.util.TreeMap;
+
+public class DefaultBlacklistStrategy implements IBlacklistStrategy {
+
+    private static Logger LOG = LoggerFactory.getLogger(DefaultBlacklistStrategy.class);
+
+    public static final int DEFAULT_BLACKLIST_SCHEDULER_RESUME_TIME = 1800;
+    public static final int DEFAULT_BLACKLIST_SCHEDULER_TOLERANCE_COUNT = 3;
+
+    private IReporter _reporter;
```
revans2: Again if we remove the '_' I think that would fix a lot of the checkstyle violations.

storm-server/pom.xml
revans2: It is actually 3679 violations (not 4000).

revans2:
Testing worked really well.  I am +1 once my nits are addressed.

storm-server/pom.xml
HeartSaVioR: Ah yes I increased the number while rebasing and planned to fix it afterwards, but forgot to fix. I'll address. Thanks for the notice.

storm-server/src/main/java/org/apache/storm/DaemonConfig.java
HeartSaVioR: Negative count doesn't make sense. I'll change it to `@isPositiveNumber`

storm-server/src/main/java/org/apache/storm/scheduler/blacklist/BlacklistScheduler.java
HeartSaVioR: The refactoring is mainly intended for removing duplicated long `catch` statements. Btw I'll change below line to use Utils.

storm-server/src/main/java/org/apache/storm/scheduler/blacklist/Sets.java
HeartSaVioR: OK I'll file an issue. Thanks for pointing out.

HeartSaVioR: Filed STORM-2762: https://issues.apache.org/jira/browse/STORM-2762

HeartSaVioR:
@revans2 Thanks for reviewing. I've addressed your comments, and rebased with latest master.

PR #2345, opened 2017-09-26
 STORM-2438: added in rebalance changes to support RAS

revans2:
This is based off #2339 (which is in the first commit).  If people want to review it all together I am fine with closing the other one.

This does a few things

1) It updates the AsyncLocalizer and Slot so that when the jar, conf, or serialized topology changes the new code is downloaded to a temp location, the workers are shot, the changes are made, and the workers are brought back up.
2) It adds in changes to the rebalance command so you can adjust config options and resources in addition to what was there before.

I plan on adding support for generic blobs to optionally cause the workers to be restarted on update. This would be for things like jars on the classpath.  I plan to do this work in a separate pull request because this is already very large.

### DIFF
--- a/storm-server/src/main/java/org/apache/storm/daemon/supervisor/Slot.java
+++ b/storm-server/src/main/java/org/apache/storm/daemon/supervisor/Slot.java
@@ -115,8 +115,6 @@ public class Slot extends Thread implements AutoCloseable, BlobChangingCallback 
         }
     }
 
-    //TODO go through all of the state transitions and make sure we handle changingBlobs
-    //TODO make sure to add in transition helpers that clean changingBlobs && pendingChangeingBlobs for not the current topology
     static class DynamicState {
         public final MachineState state;
         public final LocalAssignment newAssignment;
@@ -126,9 +124,9 @@ public class Slot extends Thread implements AutoCloseable, BlobChangingCallback 
         public final Future<Void> pendingDownload;
         public final Set<TopoProfileAction> profileActions;
         public final Set<TopoProfileAction> pendingStopProfileActions;
-        public final Set<BlobChangeing> changingBlobs;
+        public final Set<BlobChanging> changingBlobs;
         public final LocalAssignment pendingChangingBlobsAssignment;
-        public final Set<Future<Void>> pendingChangeingBlobs;
+        public final Set<Future<Void>> pendingChangingBlobs;
         
         /**
          * The last time that WAITING_FOR_WORKER_START, KILL, or KILL_AND_RELAUNCH were entered into.
@@ -156,7 +154,7 @@ public class Slot extends Thread implements AutoCloseable, BlobChangingCallback 
             this.pendingStopProfileActions = Collections.emptySet();
             this.changingBlobs = Collections.emptySet();
             this.pendingChangingBlobsAssignment = null;
-            this.pendingChangeingBlobs = Collections.emptySet();
+            this.pendingChangingBlobs = Collections.emptySet();
         }
 
         public DynamicState(final MachineState state, final LocalAssignment newAssignment,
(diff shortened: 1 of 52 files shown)


revans2:
I have done a lot of manual testing with this patch, but if others want to try it out feel free to kick the tires.  Any feedback is appreciated.

revans2:
For anyone interested I just rebased as #2339 was merged into master.

revans2:
For anyone interested I just rebased to fix a very minor merge conflict.

storm-server/src/main/java/org/apache/storm/daemon/supervisor/Slot.java
```diff
@@ -104,9 +111,12 @@
             this.port = port;
             this.iSupervisor = iSupervisor;
             this.localState = localState;
+            this.changingCallback = changingCallback;
         }
     }
-    
+
+    //TODO go through all of the state transitions and make sure we handle changingBlobs
+    //TODO make sure to add in transition helpers that clean changingBlobs && pendingChangeingBlobs for not the current topology
```
kishorvpatil: ??

storm-server/src/main/java/org/apache/storm/daemon/supervisor/Slot.java
```diff
@@ -263,7 +314,41 @@ public String toString() {
             return "{ " + topoId + ": " + request + " }";
         }
     }
-    
+
+    /**
+     * Holds the information about a blob that is changing.
+     */
+    static class BlobChangeing {
```
kishorvpatil: Does this need to be `BlobChanging`?

storm-server/src/main/java/org/apache/storm/localizer/AsyncLocalizer.java
```diff
@@ -156,17 +146,141 @@ public AsyncLocalizer(Map<String, Object> conf, AtomicReference<Map<Long, LocalA
...
+                            failures++;
+                            if (failures > blobDownloadRetries) {
+                                throw new RuntimeException("Could not download...", e);
+                            }
+                            LOG.warn("Failed to download blob {} will try again in {} ms", blob, ATTEMPTS_INTERVAL_TIME, e);
+                            Utils.sleep(ATTEMPTS_INTERVAL_TIME);
+                        }
+                    }
+                }
+                LOG.debug("FINISHED download of {}", blob);
```
kishorvpatil: Should this be info?

storm-server/src/main/java/org/apache/storm/localizer/LocallyCachedBlob.java
```diff
@@ -0,0 +1,220 @@
...
+                //TODO we probably want to not use this, or make it just return something that has less power to modify things
+                cb.blobChanging(pna.getAssignment(), pna.getPort(), this, gtg);
+            } finally {
+                gtg.countDownIfLatchWasNotGotten();
+            }
+        }
+        try {
+            cdl.await(3, TimeUnit.MINUTES);
+        } catch (InterruptedException e) {
+            //TODO need to think about error handling here in general.
```
kishorvpatil: Should we at least log the error being ignored?

storm-server/src/main/java/org/apache/storm/daemon/supervisor/Slot.java
revans2: Yes I left in a lot of TODOs #2363 cleans them up, but I will try to pull back what I can here.

In this particular case I did go through the transitions and I think I have it all covered.

storm-server/src/main/java/org/apache/storm/localizer/AsyncLocalizer.java
revans2: I used it for debugging.  If you see a good reason for it to be info I can make it info, but I didn't see a lot of value in it.

revans2:
@kishorvpatil I think I addressed your review comments.  If you really want the log message to be info I can do it.

kishorvpatil:
@revans2 LGTM.

PR #2350, opened 2017-09-28
STORM-2764:  Only login the user if not already done

revans2:
If the changes look good I am happy to put up pull requests for other branches too.


revans2:
@HeartSaVioR I put it up as #2352

PR #2352, opened 2017-09-29
STORM-2764:  Only login the user if not already done (1.x)

revans2:


revans2:
This is the 1.x version of #2350

PR #2355, opened 2017-10-03
[STORM-2769] Fast-fail if output streamId is null

Ethanlm:
JIRA: https://issues.apache.org/jira/browse/STORM-2769

If streamId is null, we should throw an exception because thrift doesn't support null in map keys.

A test:

image

### DIFF
--- a/storm-client/src/jvm/org/apache/storm/topology/OutputFieldsGetter.java
+++ b/storm-client/src/jvm/org/apache/storm/topology/OutputFieldsGetter.java
@@ -39,6 +39,9 @@ public class OutputFieldsGetter implements OutputFieldsDeclarer {
     }
 
     public void declareStream(String streamId, boolean direct, Fields fields) {
+        if (null == streamId) {
+            throw new IllegalArgumentException("streamId can't be null");
+        }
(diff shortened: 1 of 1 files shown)


storm-client/src/jvm/org/apache/storm/topology/OutputFieldsGetter.java
```diff
@@ -39,6 +39,9 @@ public void declareStream(String streamId, Fields fields) {
     }
 
     public void declareStream(String streamId, boolean direct, Fields fields) {
+        if (null == streamId) {
+            throw new IllegalArgumentException("streamId cann't be null");
```
srdo: Nit: cann't with two n's

srdo:
Seems reasonable to me. +1

storm-client/src/jvm/org/apache/storm/topology/OutputFieldsGetter.java
Ethanlm: Thanks. I made a really silly mistake...

srdo:
Thanks @Ethanlm, merged to master, 1.x and 1.1.x.

PR #2360, opened 2017-10-07
STORM-2716: Fix logviewer tests on Windows.

srdo:
See https://issues.apache.org/jira/browse/STORM-2716.

The tests run on Windows now at least. I haven't tested Logviewer beyond checking that I can search logs on my local machine from Storm UI, so there may still be some corners of Logviewer that don't work on Windows.

I also removed some constants that were making accessing the file.separator property easy. I don't think we need them since File.separator exists.


HeartSaVioR:
+1

PR #2362, opened 2017-10-08
Improve mention of aggregate() in the partitionAggregate() docs

F30:
This is a revival of #1018: It took a while, but I have now rebased the changes onto the current master and adjusted the base branch.

However, #1018 is in "missing common history" mode after force-pushing the new commits. Changing the base branch didn't cause the diff to be recomputed, so the Pull Request cannot be re-opened.

Therefore, this new PR is required.

### DIFF
--- a/docs/Trident-API-Overview.md
+++ b/docs/Trident-API-Overview.md
@@ -484,7 +484,7 @@ public class Count implements CombinerAggregator<Long> {
 }
 ```
 
-The benefits of CombinerAggregators are seen when you use them with the aggregate method instead of partitionAggregate. In that case, Trident automatically optimizes the computation by doing partial aggregations before transferring tuples over the network.
+CombinerAggregators offer high efficiency when used with the aggregate method instead of partitionAggregate ([see below](#aggregation-operations)).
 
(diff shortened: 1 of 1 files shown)


docs/Trident-API-Overview.md
```diff
@@ -484,7 +484,7 @@ public class Count implements CombinerAggregator<Long> {
 }
 ```
 
-The benefits of CombinerAggregators are seen when you use them with the aggregate method instead of partitionAggregate. In that case, Trident automatically optimizes the computation by doing partial aggregations before transferring tuples over the network.
+CombinerAggregators offer high efficiency when used with the aggregate method instead of partitionAggregate (see below).
```
srdo: I'm a little unsure which part of the document below this is referring to. Can you insert a link instead, e.g. using https://gist.github.com/asabaylus/3071099?

F30: I have added a link now.

srdo:
Thanks, +1

srdo:
Merged to master, 1.x and 1.1.x.

PR #2363, opened 2017-10-09
 STORM-2759: Let users indicate if a blob should restart a worker 

revans2:
This is based off of the changes in #2345.  They are all in the first commits.  The last commit has all of the newer changes for this JIRA in it.


revans2:
#2345 was merged so I rebased to make it more clear the new changes.  I actually delete code now :).

storm-server/src/test/java/org/apache/storm/localizer/AsyncLocalizerTest.java
```diff
@@ -286,29 +297,80 @@ public void testRequestDownloadTopologyBlobs() throws Exception {
...
+
+        synchronized void addReferences(List<LocalResource> localresource, PortAndAssignment pna, BlobChangingCallback cb) {
+            String user = pna.getOwner();
+            for (LocalResource blob : localresource) {
+                ConcurrentMap<String, LocalizedResource> lrsrcSet = blob.shouldUncompress() ? userArchives.get(user) : userFiles.get(user);
+                if (lrsrcSet != null) {
+                    LocalizedResource lrsrc = lrsrcSet.get(blob.getBlobName());
+                    if (lrsrc != null) {
+                        lrsrc.addReference(pna, blob.needsCallback() ? cb : null);
+                        lrsrc.addReference(pna, blob.needsCallback() ? cb : null);
```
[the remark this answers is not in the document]
revans2: Good catch looks like a copy and paste error

revans2:
@kishorvpatil I addressed your review comments.

storm-client/src/jvm/org/apache/storm/utils/NimbusClient.java
```diff
@@ -126,7 +137,9 @@ public static NimbusClient getConfiguredClientAs(Map<String, Object> conf, Strin
                 nimbusSummary = client.getClient().getLeader();
                 if (nimbusSummary != null) {
                     String leaderNimbus = nimbusSummary.get_host() + ":" + nimbusSummary.get_port();
-                    LOG.info("Found leader nimbus : {}", leaderNimbus);
+                    if (LOG.isDebugEnabled() || shouldLogLeader(leaderNimbus)) {
```
HeartSaVioR: nit: There's mismatch between checked log level and actual log level of message. I think `shouldLogLeader` will reduce verbose of message so personally I prefer to leave message whenever shouldLogLeader is true.

revans2: I will adjust it so it is clearer.

storm-client/src/jvm/org/apache/storm/utils/NimbusClient.java
```diff
@@ -86,7 +86,22 @@ public static NimbusClient getConfiguredClient(Map<String, Object> conf) {
     public static NimbusClient getConfiguredClient(Map<String, Object> conf, Integer timeout) {
         return getConfiguredClientAs(conf, null, timeout);
     }
-    
+
+    private static String oldLeader = "";
+
+    private static synchronized boolean shouldLogLeader(String leader) {
+        if (LOG.isDebugEnabled()) {
+            //If debug logging is turned on we shoudl just log the leader all the time....
```
HeartSaVioR: nit: shoudl -> should

revans2:
@HeartSaVioR I addressed your review comments.  I didn't change the name of shouldLogLeader, but I added javadocs to make it clear what it does.

HeartSaVioR:
OK. Thanks for quick addressing. +1 again.

srdo:
@revans2 I think something in this PR is causing topology deployment to either fail or be really slow occasionally.

The integration test has been failing fairly consistently since cef450064fa20e2194ef3f51a21c8e6693a285e3. I tried running the test outside a VM with a locally installed Storm setup, and it has failed every time for me. 

Most runs seem to fail in ways that make it look like the integration test is just flaky (e.g. tuple windows not matching the calculated window), but in at least a few tests I saw the topology get submitted to Nimbus followed by about 3 minutes of nothing happening. The workers never started and the supervisor didn't seem aware of the scheduling. The only evidence that the topology was submitted was in the Nimbus log. This still happens even if the test topologies are killed with a timeout of 0, so there should be slots free for the next test immediately.

I tried reverting cef450064fa20e2194ef3f51a21c8e6693a285e3 and it seems to make the integration test pass much more often. Over 5 runs there was still an instance of a supervisor failing to start the workers, but the other 4 passed.

revans2:
@srdo 

3 mins would correspond to a timeout that we have when the supervisor gets confused.

https://github.com/apache/storm/blob/7afd6fbe4603e35114a84e836b02484fe8cda660/storm-server/src/main/java/org/apache/storm/localizer/LocallyCachedBlob.java#L238

If you want to file a JIRA and assign it to me I will take a look at it and see if I can reliably reproduce it.  If not I may ask for your help, because I have not see this yet.  But I also don't typically run the integration tests, I just let travis do it for me.

srdo:
@revans2 Thanks. I opened https://issues.apache.org/jira/browse/STORM-2809. When I have some time I'll probably look at this too, to see if I can figure out what's going on.

Regarding the 3 minutes of doing nothing, the tests also kill the topologies if nothing gets emitted within 180 seconds, so I'm not sure if the supervisor timeout is related. I'll poke at it some more.

PR #2181, opened 2017-06-29
[STORM-2607] Offset consumer + 1

tiodollar:
When i put a message in a partition, the storm-kafka-client consume this message.
But storm-kafka-client commit the offset -1.
storm-kafka-client: 1.1.0
storm-core : 1.1.0
kafka: 0.10.2.0
Steps to bug
#1 - Insert message in kafka
#2 - Read with storm Spout this topic
#3 - Get the offset for the consumer group and the offset is always offset -1

https://issues.apache.org/jira/browse/STORM-2607

PR #2367, opened 2017-10-10
STORM-2607: Storm-kafka-client never commits the last message on a partition.

srdo:
This finishes up https://github.com/apache/storm/pull/2181, which has been blocked on a rebase for a while. It was suggested on the mailing list that we finish this up quickly since the fix should ideally go in 1.1.2.

I've squashed the commits as much as possible without mixing up which changes were made by @tiodollar and which were made by me. Let me know if squashing is not okay in this case, and I'll rebase on the commits in the original PR.

### DIFF
--- a/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/OffsetManager.java
+++ b/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/OffsetManager.java
@@ -76,15 +76,17 @@ public class OffsetManager {
     public OffsetAndMetadata findNextCommitOffset() {
         long currOffset;
         long nextCommitOffset = committedOffset;
+        long lastOffMessageOffset = committedOffset;
         KafkaSpoutMessageId nextCommitMsg = null;     // this is a convenience variable to make it faster to create OffsetAndMetadata
 
         for (KafkaSpoutMessageId currAckedMsg : ackedMsgs) {  // complexity is that of a linear scan on a TreeMap
             currOffset = currAckedMsg.offset();
-            if (currOffset == nextCommitOffset + 1) {            // found the next offset to commit
+            if (currOffset == lastOffMessageOffset + 1) {            // found the next offset to commit
                 nextCommitMsg = currAckedMsg;
-                nextCommitOffset = currOffset;
-            } else if (currOffset > nextCommitOffset + 1) {
-                if (emittedOffsets.contains(nextCommitOffset + 1)) {
+                lastOffMessageOffset = currOffset;
+                nextCommitOffset = lastOffMessageOffset + 1;
+            } else if (currOffset > lastOffMessageOffset + 1) {
+                if (emittedOffsets.contains(lastOffMessageOffset + 1)) {
                     LOG.debug("topic-partition [{}] has non-continuous offset [{}]."
                         + " It will be processed in a subsequent batch.", tp, currOffset);
                     break;
@@ -99,10 +101,11 @@ public class OffsetManager {
                     LOG.debug("Processed non contiguous offset."
                         + " (committedOffset+1) is no longer part of the topic."
                         + " Committed: [{}], Processed: [{}]", committedOffset, currOffset);
-                    final Long nextEmittedOffset = emittedOffsets.ceiling(nextCommitOffset + 1);
+                    final Long nextEmittedOffset = emittedOffsets.ceiling(lastOffMessageOffset);
                     if (nextEmittedOffset != null && currOffset == nextEmittedOffset) {
                         nextCommitMsg = currAckedMsg;
-                        nextCommitOffset = currOffset;
+                        lastOffMessageOffset = currOffset;
+                        nextCommitOffset = lastOffMessageOffset + 1;
                     } else {
                         LOG.debug("topic-partition [{}] has non-continuous offset [{}]."
                             + " Next Offset to commit should be [{}]", tp, currOffset, nextEmittedOffset);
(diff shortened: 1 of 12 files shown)


external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/OffsetManager.java
```diff
@@ -77,57 +78,58 @@ public void addToEmitMsgs(long offset) {
...
                 } else {
                     /*
-                        This case will arise in case of non contiguous offset being processed.
-                        So, if the topic doesn't contain offset = committedOffset + 1 (possible
+                        This case will arise in case of non-sequential offset being processed.
+                        So, if the topic doesn't contain offset = nextCommitOffset (possible
                         if the topic is compacted or deleted), the consumer should jump to
                         the next logical point in the topic. Next logical offset should be the
-                        first element after committedOffset in the ascending ordered emitted set.
+                        first element after nextEarliestUncommittedOffset in the ascending ordered emitted set.
```
HeartSaVioR: The field `nextEarliestUncommittedOffset` looks removed in the patch and this comment refers it.

srdo: Nice catch, fixed.

HeartSaVioR:
+1
Would this be a clean merge to 1.x version lines?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/OffsetManager.java
```diff
@@ -53,8 +54,8 @@
     public OffsetManager(TopicPartition tp, long initialFetchOffset) {
         this.tp = tp;
         this.initialFetchOffset = initialFetchOffset;
-        this.committedOffset = initialFetchOffset - 1;
-        LOG.debug("Instantiated {}", this);
+        this.committedOffset = initialFetchOffset;
+        LOG.debug("Instantiated {}", this.toString());
```
hmcl: toString() is redundant

srdo: Yes, but I get a warning about leaking this from the constructor, which this fixes.

srdo:
@HeartSaVioR No, it doesn't look like it. I'll put up a 1.x and 1.1.x version.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/OffsetManager.java
hmcl: This is not a big deal. But can you tell me what kinda warning you are talking about? I never heard of such a thing. When you concatenate a string with an object reference, the toString() method of that object reference is called. I am not sure why both approaches are any different.

hmcl:
@srdo I understand that you want to keep the commit history for @tiodollar but I don't think we should have 4 commits for such a simple change. This patch should be one commit only, either yours or @tiodollar 's .  Can either you or @tiodollar please squash the commits. Thanks.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/OffsetManager.java
srdo: The warning I get is just "leaking this in constructor", Netbeans has the following description: "Using this as parameter can be dangerous in the contructor because the object is not fully initialized.". IntelliJ and Eclipse have similar warnings I believe. 

It's basically saying not to use `this` as a parameter from the constructor, because the called method may use fields that aren't initialized. I agree that leaking `this` from the constructor is not dangerous here, because sfl4j just calls toString, but I'd still like to get rid of the warning. The reason this is flagged is because `this` is used as a parameter to LOG.debug, and not as part of a string concatenation. The IDE can't tell whether `this` is used unsafely in LOG.debug, so it's flagged as a warning.

srdo:
@hmcl I agree that this many commits is not ideal. The issue is with being able to track who contributed what code. If I squash we can't tell which parts of this PR was contributed by me, and which were contributed by @tiodollar. It's my impression that we need to keep the history of who contributed what intact.

hmcl:
@srdo in my opinion the most important thing is to keep the project, and in particular the git log clean. I often have to cherry-pick changes from Apache branch into my own branch, and multiple commits for a JIRA turn branch maintenance nightmare. Several Apache projects have a strict rule of one JIRA one Git commit, which I believe is very beneficial.

If you feel really strongly about differentiating yours and his contributions, this patch must have at most two commits. One has all his contributions, and another yours. You can squash the commits and put @tiodollar as author as described in [here](https://stackoverflow.com/questions/3042437/change-commit-author-at-one-specific-commit).

srdo:
@hmcl It's not that I feel strongly about keeping the contributions separate, but I think it's a rule for Apache projects that it must be possible to tell who contributed what, even historically. I can't recall where it says so though.

Squashed to two commits.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/OffsetManager.java
hmcl: Thanks for the explanation. I just noticed that `this` is called within its own constructor, and therefore makes sense. However, IntelliJ does not give me this warning.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/OffsetManager.java
```diff
@@ -163,8 +167,9 @@ public long commit(OffsetAndMetadata committedOffset) {
 
         LOG.trace("{}", this);
         
-        LOG.debug("Committed [{}] offsets in the range [{}-{}] for topic-partition [{}].",
-                numCommittedOffsets, preCommitCommittedOffsets + 1, this.committedOffset, tp);
+        LOG.debug("Committed [{}] offsets in the range [{}-{}] for topic-partition [{}]."
+            + " Processing will resume at [{}] if the spout restarts.",
```
hmcl: at offset [{}] in case of a spout restart.

srdo: Good catch, will fix

hmcl:
@srdo are you on Gitter or some sort of chat room where I could ask you a few questions? Thanks.

srdo:
@hmcl Just signed in on Gitter, feel free to message me.

srdo:
I'll merge this weekend if I no one brings up more concerns.

HeartSaVioR:
I read a mail at dev@ from @hmcl and the talk was not relevant to this patch. +1 again.

srdo:
Merged, thanks for the reviews guys.

PR #2368, opened 2017-10-10
STORM-2771: By default don't run any tests as integration tests

revans2:


srdo:
Can we avoid using the dummy annotation by setting http://maven.apache.org/surefire/maven-failsafe-plugin/integration-test-mojo.html#skipTests to true for the default plugin configuration, and setting it to false for the integration-tests-only and all-tests profiles?

revans2:
@srdo I'm not sure I can try it out and see.

revans2:
@srdo I did some tests and it works.  The only problem with it is that if I do something odd like `mvn clean install -Pall-tests -DskipTests` it ends up running all of the java integration tests.  I'm not sure if that is a big deal or not.  I will try to fix it, but I am going to have to see if maven supports some kind of conditionals in their macro substitution.

revans2:
Maven does not support conditional properties.  I would have to do some ugly things with a profile that is activated when a property is set, and I am not sure that even then I could make it work.  Right now I like having the NoTests interface more than having skipTests not work in all situations.  But that is just me, if others disagree I can change it, it is not a big deal to me.

srdo:
Thanks for trying it out. I think we can get the same effect as the annotation by setting `<java.integration.test.include>no.tests</java.integration.test.include>` in the default properties. It's similar to how integration-tests-only disables the unit tests.

+1 for the fix, with or without the annotation.

revans2:
@srdo Made the change you suggested and it works great, thanks.  I squashed the commits and pushed as new version.  Have a look.

srdo:
Looks great, +1

PR #2370, opened 2017-10-12
STORM-2775 Update kafkaPartition metrics to match the same format as …

kevinconaway:
…kafkaOffset

Fixes https://issues.apache.org/jira/browse/STORM-2775

### DIFF
--- a/external/storm-kafka/src/jvm/org/apache/storm/kafka/PartitionManager.java
+++ b/external/storm-kafka/src/jvm/org/apache/storm/kafka/PartitionManager.java
@@ -166,13 +166,22 @@ public class PartitionManager {
     }
 
     public Map getMetricsDataMap() {
+        String[] metricPrefixes = new String[] {
+            _partition.getId(),     // Correct metric prefix, see STORM-2775
+            _partition.toString()   // Old prefix, kept for backwards compatibility
+        };
+
         Map<String, Object> ret = new HashMap<>();
-        ret.put(_partition + "/fetchAPILatencyMax", _fetchAPILatencyMax.getValueAndReset());
-        ret.put(_partition + "/fetchAPILatencyMean", _fetchAPILatencyMean.getValueAndReset());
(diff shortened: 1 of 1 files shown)


kevinconaway:
@HeartSaVioR The build failure appears unrelated to this change.  Is the 1.1.x branch unstable?

srdo:
I think some of the /external modules have flaky tests in 1.x and 1.1.x. They seem to have been fixed in master.

Is this issue not present on the master branch? If it is you should open a PR against master as well.

kevinconaway:
@srdo I can make a PR against master.  Would you like me to do so now or wait until this is reviewed?

srdo:
It's probably best to do now, I think changes have to be merged to master before it goes on the other branches.

srdo:
I'm not sure if the names may be getting parsed by consumers, so on 1.(1.)x we might want to make the new names opt-in with a configuration parameter to avoid breaking changes. What do you think?

kevinconaway:
I don't think consumers are parsing them directly.  They _do_ get passed on as is to the target time series reporting system so any dashboards that used these metrics would "break" (or need to be updated).

If backwards compatibility is a concern here, my preference would just be to send both formats, the old and the new, rather than add a feature toggle.

These particular metrics (kafkaPartition) are not as widely used / useful as the kafkaOffset metrics so I wouldn't think it would be a large impact but I'll defer to you on that.

srdo:
The impact is probably pretty small, but I don't see much harm in returning both the old and new names for 1.x and 1.1.x. I think it's a good idea.

HeartSaVioR:
I'm also with returning both format.

kevinconaway:
Updated to return both formats

srdo:
@kevinconaway This cherry picks fine to 1.x, so if you squash to one commit I'll go ahead and merge. I'd squash them myself, but I think Github won't show the PR as merged then.

Thanks for the contribution :)

kevinconaway:
Done, thank you!

srdo:
Thanks, applied to 1.x and 1.1.x

PR #2375, opened 2017-10-17
STORM-2779 NPE on shutting down WindowedBoltExecutor (1.x)

HeartSaVioR:
* waterMarkEventGenerator could be null when timestamp field is not specified


arunmahadevan:
+1

PR #2378, opened 2017-10-18
STORM-2706: Upgrade to Curator 4.0.0

srdo:
https://issues.apache.org/jira/browse/STORM-2706

I've tested this with storm-starter running on a distributed mode single-node cluster. I attempted to use storm-perf to see if there was a performance difference, but metrics seem broken on Windows. Might try it out on a Linux box.

I'm hoping people will try this out so we can decide if we want to upgrade.

### DIFF
--- a/external/storm-eventhubs/pom.xml
+++ b/external/storm-eventhubs/pom.xml
@@ -75,17 +75,6 @@
         <dependency>
             <groupId>org.apache.curator</groupId>
             <artifactId>curator-framework</artifactId>
-            <version>${curator.version}</version>
-            <exclusions>
-                <exclusion>
-                    <groupId>log4j</groupId>
-                    <artifactId>log4j</artifactId>
-                </exclusion>
-                <exclusion>
-                    <groupId>org.slf4j</groupId>
-                    <artifactId>slf4j-log4j12</artifactId>
-                </exclusion>
-            </exclusions>
         </dependency>
     </dependencies>
 </project>
--- a/external/storm-kafka-monitor/pom.xml
+++ b/external/storm-kafka-monitor/pom.xml
@@ -60,12 +60,6 @@
         <dependency>
             <groupId>org.apache.curator</groupId>
             <artifactId>curator-framework</artifactId>
-            <exclusions>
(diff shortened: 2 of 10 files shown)


HeartSaVioR:
According to https://curator.apache.org/zk-compatibility.html, Curator 4.0.x supports both ZK versions, but we still need to keep an eye of ZK version. If we want to support Zookeeper 3.4.x, we need to **exclude** Zookeeper dependency from `curator-recipes` and add Zookeeper dependency manually.

Does the patch work with Zookeeper 3.4.x? We still need to consider supporting Zookeeper 3.4.x first, cause Zookeeper 3.5.x is still announced as 'beta' and most of our end users don't want to explore the beta in production.

srdo:
@HeartSaVioR The exclusion isn't necessary because we set Zookeeper to version 3.4.6 in the dependencyManagement section of the root pom. It ensures that if any child poms depend on Zookeeper without specifying a version they'll get 3.4.6, and if Zookeeper is included transitively it'll also be version 3.4.6.

revans2:
The simplest way to know for sure what you are going to get is to run `mvn dependency:tree`

```
$ mvn dependency:tree | grep zookeeper | sort -u
[INFO] +- org.apache.zookeeper:zookeeper:jar:3.4.6:compile
[INFO] |     +- org.apache.zookeeper:zookeeper:jar:3.4.6:compile
[INFO] |  +- org.apache.zookeeper:zookeeper:jar:3.4.6:compile
[INFO] |  +- org.apache.zookeeper:zookeeper:jar:3.4.6:provided
[INFO] |  +- org.apache.zookeeper:zookeeper:jar:3.4.6:test
[INFO] |  \- org.apache.zookeeper:zookeeper:jar:3.4.6:compile
[INFO] |  |  +- org.apache.zookeeper:zookeeper:jar:3.4.6:provided
```

I did a diff of the dependency trees both before and after this patch.  They are identical except for the version of curator has changed.

I am +1 for this change

HeartSaVioR:
@revans2 Thanks for making it clear. I'm also +1 then.

PR #2379, opened 2017-10-19
STORM-2782 - refactor partial key grouping to make it more flexible a…

kevpeek:
…nd reusable

### DIFF
--- a/storm-client/test/jvm/org/apache/storm/grouping/partialKeyGrouping/PartialKeyGroupingTest.java
+++ b/storm-client/test/jvm/org/apache/storm/grouping/partialKeyGrouping/PartialKeyGroupingTest.java
@@ -46,6 +46,20 @@ public class PartialKeyGroupingTest {
     @Test
     public void testChooseTasks() {
         PartialKeyGrouping pkg = new PartialKeyGrouping();
+        pkg.prepare(null, null, Lists.newArrayList(0, 1, 2, 3, 4, 5));
+        Values message = new Values("key1");
+        List<Integer> choice1 = pkg.chooseTasks(0, message);
+        assertThat(choice1.size(), is(1));
+        List<Integer> choice2 = pkg.chooseTasks(0, message);
+        assertThat(choice2, is(not(choice1)));
+        List<Integer> choice3 = pkg.chooseTasks(0, message);
+        assertThat(choice3, is(not(choice2)));
+        assertThat(choice3, is(choice1));
+    }
+
+    @Test
+    public void testChooseTasksWithoutConsecutiveTaskIds() {
+        PartialKeyGrouping pkg = new PartialKeyGrouping();
         pkg.prepare(null, null, Lists.newArrayList(9, 8, 7, 1, 2, 3));
         Values message = new Values("key1");
         List<Integer> choice1 = pkg.chooseTasks(0, message);
@@ -62,7 +76,7 @@ public class PartialKeyGroupingTest {
         PartialKeyGrouping pkg = new PartialKeyGrouping(new Fields("test"));
         WorkerTopologyContext context = mock(WorkerTopologyContext.class);
         when(context.getComponentOutputFields(any(GlobalStreamId.class))).thenReturn(new Fields("test"));
-        pkg.prepare(context, mock(GlobalStreamId.class), Lists.newArrayList(9, 8, 7, 1, 2, 3));
+        pkg.prepare(context, mock(GlobalStreamId.class), Lists.newArrayList(0, 1, 2, 3, 4, 5));
         Values message = new Values("key1");
         List<Integer> choice1 = pkg.chooseTasks(0, message);
         assertThat(choice1.size(), is(1));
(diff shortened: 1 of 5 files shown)


storm-client/test/jvm/org/apache/storm/grouping/partialKeyGrouping/PartialKeyGroupingTest.java
```diff
@@ -53,7 +62,7 @@ public void testChooseTasksFields() {
         PartialKeyGrouping pkg = new PartialKeyGrouping(new Fields("test"));
         WorkerTopologyContext context = mock(WorkerTopologyContext.class);
         when(context.getComponentOutputFields(any(GlobalStreamId.class))).thenReturn(new Fields("test"));
-        pkg.prepare(context, mock(GlobalStreamId.class), Lists.newArrayList(0, 1, 2, 3, 4, 5));
+        pkg.prepare(context, mock(GlobalStreamId.class), Lists.newArrayList(9, 8, 7, 1, 2, 3));
```
revans2: Why are we changing this?  The test should still pass just fine with the original values, and if not it is a bug.  So why change them?

kevpeek: I changed these because at one point during my refactoring the test passed purely by coincidence with consecutive, increasing numbers. I wanted to ensure that was not the case, although I don't remember the specifics of which code caused that.

revans2: Sounds reasonable.  Just FYI in reality they are likely to be consecutive increasing numbers because we assign the IDs by component in a consecutive increasing fashion.

revans2:
I am a conditional +1 on this change assuming that the performance numbers are within a few percent of the original.

storm-client/test/jvm/org/apache/storm/grouping/partialKeyGrouping/PartialKeyGroupingTest.java
kevpeek: I can add another test that does use consecutive numbers. I simply wanted to ensure the code did not require consecutive numbers.

kevpeek:
I will go collect some performance numbers and get back to you.

kevpeek:
@revans2 I ran some perf tests, and this refactor did slow things down considerably. I have made a small change that eliminates the main source of slowness, but it's still not great.

The current version of the code runs in about 150% of the time the original code does, so certainly not good. Most of the increase in runtime comes from the more general task selection algorithm. I have been able to speed things up a little bit more, but I suspect the benefits are not really worth it. Perhaps that means this code simply doesn't need to be merged in, or it could be added as a separate grouping, but here are a few observations I made along the way:

* I did not notice the slowdown in our deployment of this code because we know our key set is only a few thousand elements, allowing us to wrap the AssignmentCreator in a caching decorator. This trades a larger memory footprint for better performance. With the caching AssignmentCreator decorator, this grouping runs in 80% of the time the original takes. This is sort of the point of the refactor; customizations like this become possible with just a few lines of code.
* By far, the slowest part of this grouping is the hashing functions in createAssignment. Perhaps this is one of the optimizations you mention above, but replacing these with something faster would be a bigger win than optimizing this refactor.

revans2:
@kevpeek I agree that it might not be worth merging it right now.  But there is room for improvement.

The first thing I would do is to pre-allocate the possible return values as lists.  We should avoid as much memory allocation as possible within chooseTasks.

assignmentCreator.createAssignment is also returning a newly created object.  We could at least turn that into an int array, so there is only one object allocated instead of 3.

You might also consider refactoring createAssignment to take a reference to targetSelector so you don't need to create a class that is an intermediary between the two.  Just have targetSelector take two ints and return one of them, no object allocation at all.

If you are still having performance issues after that I can probably help with some profiling.

kevpeek:
heh. It crossed my mind to change those to arrays, but I figured the jvm would render the difference negligible. A quick test proves that I was very wrong. I'll tweak that tomorrow and run some more legitimate performance numbers, but it looks like that change alone closes the gap to just a few percent.

revans2:
@kevpeek yes the JVM can be very dumb when it comes to arrays vs Lists.

That sounds great if we can be within a few percentage points I would be happy to merge this in.

kevpeek:
current performance is between 3% and 12% slower than the old version.

@revans2 Would you elaborate on what you meant by, "pre-allocate the possible return values as lists?"

Thanks.

HeartSaVioR:
@kevpeek 
You can refer another implementation to see how we optimize it.

https://github.com/apache/storm/blob/master/storm-client/src/jvm/org/apache/storm/grouping/LoadAwareShuffleGrouping.java

For starting point you can see how we preallocate possible return values for `chooseTasks` and return it. That's what @revans2 pointed out.

kevpeek:
I went ahead and replaced the original hash function based logic with something faster. This code now runs in about 60% of the time taken by the original.

As for changing the hash logic, I ran tests to measure the resulting tuple distribution across downstream tasks. There is not a huge difference, but the new code produces a slightly more even distribution. This is primarily because it avoids the case where the two hash functions choose the same task, thus eliminating the power of two choices.

@revans2

kevpeek:
new PR with this squashed to one commit: https://github.com/apache/storm/pull/2395

PR #2380, opened 2017-10-23
STORM-2781: Refactor storm-kafka-client KafkaSpout Processing Guarantees

hmcl:
- Define processing guarantees as AT_LEAST_ONCE, AT_MOST_ONCE, NONE
 - Refactor method name from setForceEnableTupleTracking to setTupleTrackingEnforced
 - Throw IllegalStateException instead of IllegalArgumentException if spout attempts to emit an already committed message
 - Update documentation to reflect these changes

### DIFF
--- a/docs/storm-kafka-client.md
+++ b/docs/storm-kafka-client.md
@@ -266,7 +266,7 @@ use Kafka-clients 0.10.0.0, you would use the following dependency in your `pom.
 You can also override the kafka clients version while building from maven, with parameter `storm.kafka.client.version`
 e.g. `mvn clean install -Dstorm.kafka.client.version=0.10.0.0`
 
-When selecting a kafka client version, you should ensure - 
+When selecting a kafka client version, you should ensure -
  1. kafka api is compatible. storm-kafka-client module only supports **0.10 or newer** kafka client API. For older versions,
  you can use storm-kafka module (https://github.com/apache/storm/tree/master/external/storm-kafka).  
  2. The kafka client selected by you should be wire compatible with the broker. e.g. 0.9.x client will not work with 
@@ -298,25 +298,46 @@ Currently the Kafka spout has has the following default values, which have been 
 * max.uncommitted.offsets = 10000000
 <br/>
 
-# Messaging reliability modes
+# Processing Guarantees
 
-In some cases you may not need or want the spout to guarantee at-least-once processing of messages. The spout also supports at-most-once and any-times modes. At-most-once guarantees that any tuple emitted to the topology will never be reemitted. Any-times makes no guarantees, but may reduce the overhead of committing offsets to Kafka in cases where you truly don't care how many times a message is processed.
+The `KafkaSpoutConfig.ProcessingGuarantee` enum parameter controls when an offset is committed to Kafka. This is
+conceptually equivalent to marking the tuple with the `ConsumerRecord` for that offset as being successfully processed
+because the tuple won't get re-emitted in case of failure or time out.
 
-To set the processing guarantee, use the KafkaSpoutConfig.Builder.setProcessingGuarantee method, e.g.
+For the AT_LEAST_ONCE and AT_MOST_ONCE processing guarantees the spout controls when the commit happens.
+When the guarantee is NONE Kafka controls when the commit happens.
+
+* AT_LEAST_ONCE - an offset is ready to commit only after the corresponding tuple has been processed (at-least-once)
+     and acked. If a tuple fails or times out it will be re-emitted. A tuple can be processed more than once if for instance
+     the ack gets lost.
+
+* AT_MOST_ONCE - Offsets will be committed to Kafka right after being polled but before being emitted to the downstream
+     components of the topology. Offsets are processed at most once because tuples that fail or timeout won't be retried.
+
+* NONE - the polled offsets are committed to Kafka periodically as controlled by the Kafka properties
+     "enable.auto.commit" and "auto.commit.interval.ms". Because the spout does not control when the commit happens
+     it cannot give any message processing guarantees, i.e. a message may be processed 0, 1 or more times.
+     This option requires "enable.auto.commit=true". If "enable.auto.commit=false" an exception will be thrown.
+
+To set the processing guarantee use the `KafkaSpoutConfig.Builder.setProcessingGuarantee` method as follows:
+
 ```java
 KafkaSpoutConfig<String, String> kafkaConf = KafkaSpoutConfig
   .builder(String bootstrapServers, String ... topics)
   .setProcessingGuarantee(ProcessingGuarantee.AT_MOST_ONCE)
 ```
 
-The spout will disable tuple tracking for emitted tuples by default when you use at-most-once or any-times. In some cases you may want to enable tracking anyway, because tuple tracking is necessary for some features of Storm, e.g. showing complete latency in Storm UI, or enabling backpressure through the `Config.TOPOLOGY_MAX_SPOUT_PENDING` parameter.
+# Tuple Tracking
 
-If you need to enable tracking, use the KafkaSpoutConfig.Builder.setForceEnableTupleTracking method, e.g.
+By default the spout only tracks emitted tuples when the processing guarantee is AT_LEAST_ONCE. It may be necessary to track
+emitted tuples with other processing guarantees to benefit from Storm features such as showing complete latency in the UI,
+or enabling backpressure with Config.TOPOLOGY_MAX_SPOUT_PENDING.
+
 ```java
 KafkaSpoutConfig<String, String> kafkaConf = KafkaSpoutConfig
   .builder(String bootstrapServers, String ... topics)
   .setProcessingGuarantee(ProcessingGuarantee.AT_MOST_ONCE)
-  .setForceEnableTupleTracking(true)
+  .setTupleTrackingEnforced(true)
 ```
 
-Note that this setting has no effect in at-least-once mode, where tuple tracking is always enabled.
+Note: This setting has no effect with AT_LEAST_ONCE processing guarantee, where tuple tracking is required and therefore always enabled.
--- a/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
+++ b/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
@@ -78,17 +78,17 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
     private transient KafkaSpoutRetryService retryService;
     // Handles tuple events (emit, ack etc.)
     private transient KafkaTupleListener tupleListener;
-    // timer == null for modes other than at-least-once
+    // timer == null if processing guarantee is none or at-most-once
     private transient Timer commitTimer;
     // Flag indicating that the spout is still undergoing initialization process.
     private transient boolean initialized;
     // Initialization is only complete after the first call to  KafkaSpoutConsumerRebalanceListener.onPartitionsAssigned()
 
     // Tuples that were successfully acked/emitted. These tuples will be committed periodically when the commit timer expires,
-    //or after a consumer rebalance, or during close/deactivate. Always empty if not using at-least-once mode.
+    // or after a consumer rebalance, or during close/deactivate. Always empty if processing guarantee is none or at-most-once.
     private transient Map<TopicPartition, OffsetManager> offsetManagers;
     // Tuples that have been emitted but that are "on the wire", i.e. pending being acked or failed.
-    // Always empty if not using at-least-once mode.
+    // Always empty if processing guarantee is none or at-most-once
     private transient Set<KafkaSpoutMessageId> emitted;
     // Records that have been polled and are queued to be emitted in the nextTuple() call. One record is emitted per nextTuple()
     private transient Iterator<ConsumerRecord<K, V>> waitingToEmit;
@@ -125,8 +125,8 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
 
         tupleListener = kafkaSpoutConfig.getTupleListener();
 
-        if (isAtLeastOnce()) {
-            // Only used if the spout commits offsets for acked tuples
+        if (isAtLeastOnceProcessing()) {
+            // Only used if the spout should commit an offset to Kafka only after the corresponding tuple has been acked.
             commitTimer = new Timer(TIMER_DELAY_MS, kafkaSpoutConfig.getOffsetsCommitPeriodMs(), TimeUnit.MILLISECONDS);
         }
         refreshSubscriptionTimer = new Timer(TIMER_DELAY_MS, kafkaSpoutConfig.getPartitionRefreshPeriodMs(), TimeUnit.MILLISECONDS);
@@ -140,7 +140,7 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
         LOG.info("Kafka Spout opened with the following configuration: {}", kafkaSpoutConfig);
     }
 
-    private boolean isAtLeastOnce() {
+    private boolean isAtLeastOnceProcessing() {
         return kafkaSpoutConfig.getProcessingGuarantee() == KafkaSpoutConfig.ProcessingGuarantee.AT_LEAST_ONCE;
     }
 
@@ -154,7 +154,7 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
             LOG.info("Partitions revoked. [consumer-group={}, consumer={}, topic-partitions={}]",
                     kafkaSpoutConfig.getConsumerGroupId(), kafkaConsumer, partitions);
             previousAssignment = partitions;
-            if (isAtLeastOnce() && initialized) {
+            if (isAtLeastOnceProcessing() && initialized) {
                 initialized = false;
                 commitOffsetsForAckedTuples();
             }
@@ -170,7 +170,7 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
         }
 
         private void initialize(Collection<TopicPartition> partitions) {
-            if (isAtLeastOnce()) {
+            if (isAtLeastOnceProcessing()) {
                 // remove from acked all partitions that are no longer assigned to this spout
                 offsetManagers.keySet().retainAll(partitions);
                 retryService.retainAll(partitions);
@@ -188,7 +188,7 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
                 final OffsetAndMetadata committedOffset = kafkaConsumer.committed(tp);
                 final long fetchOffset = doSeek(tp, committedOffset);
                 // If this partition was previously assigned to this spout, leave the acked offsets as they were to resume where it left off
-                if (isAtLeastOnce() && !offsetManagers.containsKey(tp)) {
+                if (isAtLeastOnceProcessing() && !offsetManagers.containsKey(tp)) {
                     offsetManagers.put(tp, new OffsetManager(tp, fetchOffset));
                 }
             }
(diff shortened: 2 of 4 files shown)


docs/storm-kafka-client.md
```diff
@@ -298,25 +298,44 @@ Currently the Kafka spout has has the following default values, which have been
 * max.uncommitted.offsets = 10000000
 <br/>
 
-# Messaging reliability modes
+# Processing Guarantees
 
-In some cases you may not need or want the spout to guarantee at-least-once processing of messages. The spout also supports at-most-once and any-times modes. At-most-once guarantees that any tuple emitted to the topology will never be reemitted. Any-times makes no guarantees, but may reduce the overhead of committing offsets to Kafka in cases where you truly don't care how many times a message is processed.
+The `KafkaSpoutConfig.ProcessingGuarantee` enum parameter controls when the tuple with the `ConsumerRecord` for an offset is marked
```
srdo: "... when the tuple with the ConsumerRecord for an offset" -> "when an offset", since the offsets are marked as processed, not the tuples or records.

docs/storm-kafka-client.md
```diff
@@ -298,25 +298,44 @@ Currently the Kafka spout has has the following default values, which have been
...
+as processed, i.e. when the offset is committed to Kafka. For AT_LEAST_ONCE and AT_MOST_ONCE guarantees the spout controls when
```
srdo: "For AT_LEAST_ONCE" -> "For the AT_LEAST_ONCE"

docs/storm-kafka-client.md
```diff
@@ -298,25 +298,44 @@ Currently the Kafka spout has has the following default values, which have been
...
+the commit happens. When the guarantee is NONE Kafka controls when the commit happens.
+
+* AT_LEAST_ONCE - an offset is ready to commit only after the corresponding tuple has been processed (at-least-once)
+     and acked. If a tuple fails or times-out it will be re-emitted. A tuple can be processed more than once if for instance
```
srdo: Nit: "times-out" -> "times out".

docs/storm-kafka-client.md
```diff
@@ -298,25 +298,44 @@ Currently the Kafka spout has has the following default values, which have been
...
+     the ack gets lost.
+
+* AT_MOST_ONCE - every offset will be committed to Kafka right after being polled but before being emitted
```
srdo: "Every offset" -> "Offsets"

docs/storm-kafka-client.md
```diff
@@ -298,25 +298,44 @@ Currently the Kafka spout has has the following default values, which have been
...
+     to the downstream components of the topology. It guarantees that the offset is processed at-most-once because it
```
srdo: "It guarantees" -> "This setting guarantees". "because it won't retry" -> "because the spout won't retry". Also nit: "at-most-once" -> "at most once"

docs/storm-kafka-client.md
```diff
@@ -298,25 +298,44 @@ Currently the Kafka spout has has the following default values, which have been
...
+     won't retry tuples that fail or timeout after the commit to Kafka has been done.
```
srdo: Consider dropping " after the commit to Kafka has been done". The spout just won't retry tuples period.

docs/storm-kafka-client.md
```diff
@@ -298,25 +298,44 @@ Currently the Kafka spout has has the following default values, which have been
...
 KafkaSpoutConfig<String, String> kafkaConf = KafkaSpoutConfig
   .builder(String bootstrapServers, String ... topics)
   .setProcessingGuarantee(ProcessingGuarantee.AT_MOST_ONCE)
 ```
 
-The spout will disable tuple tracking for emitted tuples by default when you use at-most-once or any-times. In some cases you may want to enable tracking anyway, because tuple tracking is necessary for some features of Storm, e.g. showing complete latency in Storm UI, or enabling backpressure through the `Config.TOPOLOGY_MAX_SPOUT_PENDING` parameter.
+# Tuple Tracking
+
+By default the spout only tracks emitted tuples when the processing guarantee is AT_LEAST_ONCE. It may be necessary to track
+emitted tuples with other processing guarantees to benefit of Storm features such as showing complete latency in the UI,
```
srdo: "to benefit of" -> "to benefit from"

docs/storm-kafka-client.md
```diff
@@ -298,25 +298,44 @@ Currently the Kafka spout has has the following default values, which have been
...
 KafkaSpoutConfig<String, String> kafkaConf = KafkaSpoutConfig
   .builder(String bootstrapServers, String ... topics)
   .setProcessingGuarantee(ProcessingGuarantee.AT_MOST_ONCE)
-  .setForceEnableTupleTracking(true)
+  .setTupleTrackingEnforced(true)
 ```
 
-Note that this setting has no effect in at-least-once mode, where tuple tracking is always enabled.
\ No newline at end of file
+Note: This setting has no effect with AT_LEAST_ONCE processing guarantees where tuple tracking is required and therefore always enabled.
```
srdo: "with AT_LEAST_ONCE processing guarantees" -> "with the AT_LEAST_ONCE processing guarantee"

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -78,17 +78,17 @@
     private transient KafkaSpoutRetryService retryService;
     // Handles tuple events (emit, ack etc.)
     private transient KafkaTupleListener tupleListener;
-    // timer == null for modes other than at-least-once
+    // timer == null if processing guarantee is other than at-least-once
```
srdo: Probably nicer to write "if processing guarantee is not at-least-once"

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -78,17 +78,17 @@
...
     private transient Timer commitTimer;
     // Flag indicating that the spout is still undergoing initialization process.
     private transient boolean initialized;
     // Initialization is only complete after the first call to  KafkaSpoutConsumerRebalanceListener.onPartitionsAssigned()
 
     // Tuples that were successfully acked/emitted. These tuples will be committed periodically when the commit timer expires,
-    //or after a consumer rebalance, or during close/deactivate. Always empty if not using at-least-once mode.
+    // or after a consumer rebalance, or during close/deactivate. Always empty if not using at-least-once processing guarantee.
```
srdo: "the at-least-once processing guarantee"

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -78,17 +78,17 @@
...
     private transient Map<TopicPartition, OffsetManager> offsetManagers;
     // Tuples that have been emitted but that are "on the wire", i.e. pending being acked or failed.
-    // Always empty if not using at-least-once mode.
+    // Always empty if processing guarantee is other than at-least-once.
```
srdo: Same as for the timer comment

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -125,8 +125,8 @@ public void open(Map<String, Object> conf, TopologyContext context, SpoutOutputC
 
         tupleListener = kafkaSpoutConfig.getTupleListener();
 
-        if (isAtLeastOnce()) {
-            // Only used if the spout commits offsets for acked tuples
+        if (isAtLeastOnceProcessing()) {
+            // Only used if the spout should commit to Kafka an offset only after its tuple has been acked.
```
srdo: "Should commit an offset to Kafka only after the corresponding tuple has been acked"

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -255,26 +255,25 @@ private void throwKafkaConsumerInterruptedException() {
...
     private boolean poll() {
         final int maxUncommittedOffsets = kafkaSpoutConfig.getMaxUncommittedOffsets();
         final int readyMessageCount = retryService.readyMessageCount();
         final boolean poll = !waitingToEmit()
-            //Check that the number of uncommitted, nonretriable tuples is less than the maxUncommittedOffsets limit
-            //Accounting for retriable tuples this way still guarantees that the limit is followed on a per partition basis,
-            //and prevents locking up the spout when there are too many retriable tuples
-            && (numUncommittedOffsets - readyMessageCount < maxUncommittedOffsets
-            || !isAtLeastOnce());
+                // Check that the number of uncommitted, non-retriable tuples is less than the maxUncommittedOffsets limit.
```
srdo: I think we indent to 4 spaces in the rest of the file?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -438,55 +440,53 @@ private void commitOffsetsForAckedTuples() {
...
-        }
-
+        // Only need to keep track of acked tuples if commits to Kafka are done after a tuple ack is received
         final KafkaSpoutMessageId msgId = (KafkaSpoutMessageId) messageId;
-        if (!emitted.contains(msgId)) {
-            if (msgId.isEmitted()) {
-                LOG.debug("Received ack for message [{}], associated with tuple emitted for a ConsumerRecord that "
-                    + "came from a topic-partition that this consumer group instance is no longer tracking "
-                    + "due to rebalance/partition reassignment. No action taken.", msgId);
+        if (isAtLeastOnceProcessing()) {
```
srdo: I think this is worse. Putting a guard clause at the top decreases indentation for the whole function and is more readable IMO.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -336,22 +335,25 @@ private void emit() {
...
         if (offsetManagers.containsKey(tp) && offsetManagers.get(tp).contains(msgId)) {   // has been acked
             LOG.trace("Tuple for record [{}] has already been acked. Skipping", record);
-        } else if (emitted.contains(msgId)) {   // has been emitted and it's pending ack or fail
+        } else if (emitted.contains(msgId)) {   // has been emitted and it is pending ack or fail
             LOG.trace("Tuple for record [{}] has already been emitted. Skipping", record);
         } else {
-            Validate.isTrue(kafkaConsumer.committed(tp) == null || kafkaConsumer.committed(tp).offset() < kafkaConsumer.position(tp),
-                "The spout is about to emit a message that has already been committed."
-                + " This should never occur, and indicates a bug in the spout");
+            if (kafkaConsumer.committed(tp) != null && (kafkaConsumer.committed(tp).offset() >= kafkaConsumer.position(tp))) {
```
srdo: I like the new error message, but why get rid of Validate?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -438,55 +440,53 @@ private void commitOffsetsForAckedTuples() {
...
 
     // ======== Fail =======
     @Override
     public void fail(Object messageId) {
-        if (!isAtLeastOnce()) {
-            // Only need to keep track of failed tuples if commits are done based on acks
-            return;
-        }
+        // Only need to keep track of failed tuples if commits to Kafka are done after a tuple ack is received
+        if (isAtLeastOnceProcessing()) {
```
srdo: Same as for ack

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -438,55 +440,53 @@ private void commitOffsetsForAckedTuples() {
     // ======== Ack =======
     @Override
     public void ack(Object messageId) {
-        if (!isAtLeastOnce()) {
-            // Only need to keep track of acked tuples if commits are done based on acks
-            return;
-        }
-
+        // Only need to keep track of acked tuples if commits to Kafka are done after a tuple ack is received
```
srdo: "after a tuple ack is received" -> "based on received tuple acks"

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -438,55 +440,53 @@ private void commitOffsetsForAckedTuples() {
...
     }
 
     // ======== Fail =======
     @Override
     public void fail(Object messageId) {
-        if (!isAtLeastOnce()) {
-            // Only need to keep track of failed tuples if commits are done based on acks
-            return;
-        }
+        // Only need to keep track of failed tuples if commits to Kafka are done after a tuple ack is received
```
srdo: "after a tuple ack is received" -> "based on received tuple acks"

docs/storm-kafka-client.md
hmcl: Well, this is tricky because Storm does not process offsets, storm processes tuples. More exactly, it processes tuples that contain ConsumerRecord's. The offset is just part of the  ConsumerRecord, which also contains key, val, etc... We commit the offset, but by committing the offset we are technically marking that the tuple was processed because even if the tuple fails, it won't be retried (processed again).

docs/storm-kafka-client.md
hmcl: Done

docs/storm-kafka-client.md
hmcl: Done

docs/storm-kafka-client.md
hmcl: Maybe we can come up with a bit better wording for this, but I really don't think that we should say that an offset is marked as processed. Offsets are committed, not processed. ConsumerRecords, wrapped by Tuples are processed.

docs/storm-kafka-client.md
hmcl: Done

docs/storm-kafka-client.md
hmcl: Done

docs/storm-kafka-client.md
hmcl: Done

docs/storm-kafka-client.md
hmcl: Done

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
hmcl: Done. Removed the double negation.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
hmcl: Done. Removed the double negation.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
hmcl: Done.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
hmcl: Done.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
hmcl: Validate throws an IllegalArgumentException where the correct exception here is IllegalStateException. Furthermore, Validate in my opinion as a confusing API - Validate(true) throws an exception if false. It is misleading to me. I would rather leave it like this.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
hmcl: I disagree. Any method that has only one if condition and nothing else, should be 
```java 
if (condition == true) 
     do_action;
```
imho it is counter natural to have code like 
```java
if (condition = false) 
    do_nothing;
else 
     do_action;
```
which is what basically the early return is doing.

There are also lengthier reasons related to the semantics of OOP, but I just think that in general wherever possible one should have code like if (condtion==true) do action.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
hmcl: Same comment above

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
hmcl: Done.

docs/storm-kafka-client.md
hmcl: Done

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
HeartSaVioR: I know there's a huge debate on if else vs early return, but I'm in favor of early return, especially it is used as some pruning on early stage of logic.
Sure I'm not in favor of second statements (if do nothing else do action) but early return is completely different as do nothing, as it does pruning the logic branch and no need to follow the if statement and statements after end of if statement.

HeartSaVioR: Btw I don't think it is a big deal to decide and stick with specific approach, cause there's no statement after if statement. 
@srdo Could we leave the decision of this review comment to @hmcl?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
HeartSaVioR: For me the behavior of Validate.isTrue() is not confusing (we have more classes from Guava like checking null), but throwing appropriate exception is more important to me.

srdo: Makes sense, thanks.

docs/storm-kafka-client.md
srdo: I like the new wording, thanks.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
srdo: @hmcl I agree that 
```
if condition == false
  do nothing
else
  do something
```
is bad, but as @HeartSaVioR mentions early return (guard clause) isn't the same thing. Instead of writing
```
if precondition == true
  rest of function
return
```
it can make the function more readable to write 
```
if precondition == false
  return
rest of function
```
because the guarding `if` now doesn't cover the rest of the function, and the indendation level is reduced. See https://blog.codinghorror.com/flattening-arrow-code/ or https://en.wikibooks.org/wiki/Computer_Programming/Coding_Style/Minimize_nesting#Early_return.

I agree that since there's no code after the `if` closes it's not that important. Will leave it up to your judgement.

srdo:
+1, good work.

hmcl:
I have squashed the commits and addressed the early return issue.

srdo:
+1 again. The test failure is unrelated. Merged to master. There's a conflict when merging to 1.x, so please open another PR for that branch. Thanks.

PR #2381, opened 2017-10-23
STORM-2784: storm-kafka-client KafkaTupleListener method onPartitions…

hmcl:
…Reassigned() should be called after initialization is complete


srdo:
+1

PR #2382, opened 2017-10-23
STORM-2786: Turn ticks back on for ackers (2.x)

revans2:
This does some minor optimizations for the ackers, but for the most part it just turns Ticks back on for ackers.

I will be putting up a pull request for 1.x as well.  If someone wants a 0.10 or 0.9 pull request I can do it, but I think those lines are mostly not supported any more.

### DIFF
--- a/storm-client/src/jvm/org/apache/storm/daemon/Acker.java
+++ b/storm-client/src/jvm/org/apache/storm/daemon/Acker.java
@@ -22,6 +22,7 @@ import org.apache.storm.task.OutputCollector;
 import org.apache.storm.task.TopologyContext;
 import org.apache.storm.tuple.Tuple;
 import org.apache.storm.tuple.Values;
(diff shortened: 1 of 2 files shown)


revans2:
@srdo I agree surprising.  But I was looking into it.  Each tuple that is leaked is only about 120 bytes in 1.x and about 100 in 0.10 and below, in 2.x it is 40 bytes.  So in order to leak 1 MB in 1.x it would take 9,000 tuples that are truly lost, not just timed out but truly lost.  So my guess is that it is just such a small leak that most of the time it is not a big deal.  And when it is a big deal so much else that is bad is happening that no one debugs the acker because they need to fix the actual cause of the missing tuples.

HeartSaVioR:
+1

revans2:
@srdo could you take a look at https://github.com/apache/storm/pull/2383 too?  it is like this, but for 1.x

PR #2387, opened 2017-10-25
STORM-2787: storm-kafka-client KafkaSpout method onPartitionsRevoked(...) should set initialized flag independently of processing guarantees

hmcl:
The commit for STORM-2787 is on top of STORM-2781 to make it easier to merge later. Please review only STORM-2787. Once the code reviews for STORM-2781 have been agreed upon I will incorporate them in this patch as well.

### DIFF
--- a/docs/storm-kafka-client.md
+++ b/docs/storm-kafka-client.md
@@ -266,7 +266,7 @@ use Kafka-clients 0.10.0.0, you would use the following dependency in your `pom.
 You can also override the kafka clients version while building from maven, with parameter `storm.kafka.client.version`
 e.g. `mvn clean install -Dstorm.kafka.client.version=0.10.0.0`
 
-When selecting a kafka client version, you should ensure - 
+When selecting a kafka client version, you should ensure -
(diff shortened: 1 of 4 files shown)


srdo:
+1 for the change. It might be good to add a test that demonstrates the fix as well if possible.

HeartSaVioR:
Looks like 1.1.x branch also needs the patch, and that branch shouldn't be on top of STORM-2781.
https://github.com/apache/storm/blob/1.1.x-branch/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java

@hmcl Could you craft a new PR against 1.1.x-branch if my understanding is right? Thanks in advance.

hmcl:
@srdo @HeartSaVioR I have incorporated the code review changes of the depending patch. It should be good to merge. Thanks.

HeartSaVioR:
@hmcl 
Thanks, I merged into master. 
Like STORM-2781 (#2380) we need another PR on 1.x branch. Please raise pull requests against 1.x branch. Thanks in advance.

PR #2390, opened 2017-10-27
[STORM-2790] Add nimbus admins groups

kishorvpatil:


Ethanlm:
Looks like there is some checkstyle violatiom issues here.

storm-client/src/jvm/org/apache/storm/blobstore/BlobStoreAclHandler.java
```diff
@@ -45,6 +47,7 @@
 public class BlobStoreAclHandler {
     public static final Logger LOG = LoggerFactory.getLogger(BlobStoreAclHandler.class);
     private final IPrincipalToLocal _ptol;
+    private final IGroupMappingServiceProvider _groupMappingProvider;
```
revans2: We should move away from using '_' as a prefix to member variables.  Please at least fix the new ones to not use it, and if you want to fix all of them that would be great too.

storm-client/src/jvm/org/apache/storm/security/auth/authorizer/SimpleACLAuthorizer.java
```diff
@@ -80,6 +80,7 @@
     }
 
     protected Set<String> _admins;
+    protected Set<String> _adminsGroups;
```
revans2: Same here we should not use '_' as a prefix.

storm-client/src/jvm/org/apache/storm/security/auth/FixedGroupsMapping.java
```diff
@@ -0,0 +1,68 @@
...
+import java.util.Map;
+
+import org.apache.storm.Config;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+public class FixedGroupsMapping implements IGroupMappingServiceProvider {
+
+  public static Logger LOG = LoggerFactory.getLogger(FixedGroupsMapping.class);
```
revans2: The new guidelines have us use 4 spaces not 2 for indentation.

storm-server/src/main/java/org/apache/storm/DaemonConfig.java
```diff
@@ -159,14 +159,6 @@
     public static final String SCHEDULER_DISPLAY_RESOURCE = "scheduler.display.resource";
 
     /**
-     * Initialization parameters for the group mapping service plugin.
-     * Provides a way for a @link{STORM_GROUP_MAPPING_SERVICE_PROVIDER_PLUGIN}
-     * implementation to access optional settings.
-     */
-    @isType(type = Map.class)
-    public static final String STORM_GROUP_MAPPING_SERVICE_PARAMS = "storm.group.mapping.service.params";
```
revans2: Was this just not used anywhere?

kishorvpatil: It was not used anywhere until this PR - where `FixedGroupsMapping` uses it.

external/storm-hdfs/src/test/java/org/apache/storm/hdfs/blobstore/BlobStoreTest.java
```diff
@@ -92,6 +93,19 @@ public static void cleanupAfterClass() throws IOException {
   // Method which initializes nimbus admin
   public static void initializeConfigs() {
     conf.put(Config.NIMBUS_ADMINS,"admin");
+    conf.put(Config.NIMBUS_ADMINS_GROUPS,"adminsGroup");
+
```
Ethanlm: nit: space after comma?

storm-client/src/jvm/org/apache/storm/security/auth/FixedGroupsMapping.java
```diff
@@ -0,0 +1,69 @@
...
+
+    /**
+     * Invoked once immediately after construction
+     *
+     * @param storm_conf Storm configuration
+     */
+    @Override
+    public void prepare(Map storm_conf) {
+        Map<?, ?> params = (Map<?, ?>) storm_conf.get(Config.STORM_GROUP_MAPPING_SERVICE_PARAMS);
+        Map<String, Set<String>> mapping = (Map<String, Set<String>>) params.get(STORM_FIXED_GROUP_MAPPING);
```
Ethanlm: Should we use `stormConf` instead of `storm_conf`? 
Do we want to use Map<String, Object> instead of Map<?, ?>?

PR #2395, opened 2017-10-30
STORM-2782 - refactor partial key grouping

kevpeek:


revans2:
Still +1 this is just a squashed version of #2379 so I am not going to wait the full 24 hours (because it is technically the same code)

PR #2397, opened 2017-10-31
[STORM-2794] Translate backtype to org.apache for topology.scheduler.strategy if scheduling older version topologies

Ethanlm:
https://issues.apache.org/jira/browse/STORM-2794

In order to support running workers of older version, when we schedule a topology of older version, the confs in TopologyDetails could be "backtype.storm". To make scheduler work, we need to translate "backtype.storm" to "org.storm" for Config.TOPOLOGY_SCHEDULER_STRATEGY

Tested it manually.

### DIFF
--- a/storm-server/src/main/java/org/apache/storm/scheduler/resource/ResourceAwareScheduler.java
+++ b/storm-server/src/main/java/org/apache/storm/scheduler/resource/ResourceAwareScheduler.java
@@ -109,7 +109,14 @@ public class ResourceAwareScheduler implements IScheduler {
         IStrategy rasStrategy = null;
         String strategyConf = (String) td.getConf().get(Config.TOPOLOGY_SCHEDULER_STRATEGY);
         try {
-            rasStrategy = ReflectionUtils.newSchedulerStrategyInstance((String) td.getConf().get(Config.TOPOLOGY_SCHEDULER_STRATEGY), conf);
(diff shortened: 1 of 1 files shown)


Ethanlm:
The build passed on my VM. Will try to build it on travis one more time.

HeartSaVioR:
Looks like supporting 0.10.x in Storm 2.0.0 brings unnecessary pain. I'm not sure we only missed this place. Maybe more, and finding more looks against the intention for what we dropped supporting client jar hack.

@revans2 Would like to hear your opinion about this.

revans2:
@HeartSaVioR I agree that we might miss some things and maintaining that backwards compatibility long term is probably not ideal.  But we have some clusters that we need to migrate off of 0.10 so we are going to make this work, at least in the short term.  When we are done migrating we are happy to have a follow on JIRA to rip out the 0.10 support.

Also I am not too concerned about missing things because we really only need configs that are topology configs that have class names in them that are read by either nimbus or the supervisor.  There are really almost no configs where we do this, because there are security concerns when we do this.  I could be wrong, but I think it is manageable.  If you still feel like it is too much of a burden, I don't have a problem with us just doing it internally.

HeartSaVioR:
@revans2 
My intention is forcing users to migrate their Storm version to at least 1.x when we release Storm 2.0.0. While I'm still not in favor of supporting too old version, if it can be removed in short term, I'm OK to add it for now. +1

Ethanlm:
Last commit resolved conflicts.

PR #2398, opened 2017-11-01
STORM-2795: Race in downloading resources can cause failure

revans2:


satishd:
@revans2 You may want to change commit message with JIRA description.

PR #2399, opened 2017-11-01
STORM-2793 Track network data metrics 

jmartell7:
Add metrics to count data transfer to account for tuples of differing sizes.  @revans2, please review when you get a chance. 

https://issues.apache.org/jira/browse/STORM-2793

### DIFF
--- a/storm-client/src/jvm/org/apache/storm/Config.java
+++ b/storm-client/src/jvm/org/apache/storm/Config.java
@@ -451,6 +451,13 @@ public class Config extends HashMap<String, Object> {
     public static final String TOPOLOGY_METRICS_CONSUMER_REGISTER = "topology.metrics.consumer.register";
 
     /**
+     * Enable tracking of network message byte counts per source-destination task. This is off by default as it
+     * creates tasks^2 metric values, but is useful for debugging as it exposes data skew when tuple sizes are uneven.
+     */
+    @isBoolean
+    public static final String TOPOLOGY_SERIALIZED_MESSAGE_SIZE_METRICS = "topology.serialized.message.size.metrics";
+
+    /**
      * A map of metric name to class name implementing IMetric that will be created once per worker JVM
      */
     @isMapEntryType(keyType = String.class, valueType = String.class)
--- a/storm-client/src/jvm/org/apache/storm/messaging/DeserializingConnectionCallback.java
+++ b/storm-client/src/jvm/org/apache/storm/messaging/DeserializingConnectionCallback.java
@@ -17,34 +17,50 @@
  */
 package org.apache.storm.messaging;
 
+import org.apache.storm.Config;
 import org.apache.storm.daemon.worker.WorkerState;
+import org.apache.storm.metric.api.IMetric;
+import org.apache.storm.serialization.KryoTupleDeserializer;
 import org.apache.storm.task.GeneralTopologyContext;
 import org.apache.storm.tuple.AddressedTuple;
-import org.apache.storm.serialization.KryoTupleDeserializer;
+import org.apache.storm.tuple.Tuple;
+import org.apache.storm.utils.ObjectReader;
 
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicLong;
+
 
 /**
  * A class that is called when a TaskMessage arrives.
  */
-public class DeserializingConnectionCallback implements IConnectionCallback {
-    private final WorkerState.ILocalTransferCallback _cb;
-    private final Map _conf;
-    private final GeneralTopologyContext _context;
-    private final ThreadLocal<KryoTupleDeserializer> _des =
-         new ThreadLocal<KryoTupleDeserializer>() {
-             @Override
-             protected KryoTupleDeserializer initialValue() {
-                 return new KryoTupleDeserializer(_conf, _context);
-             }
-         };
+public class DeserializingConnectionCallback implements IConnectionCallback, IMetric {
+    private final WorkerState.ILocalTransferCallback cb;
+    private final Map conf;
+    private final GeneralTopologyContext context;
 
-    public DeserializingConnectionCallback(final Map<String, Object> conf, final GeneralTopologyContext context, WorkerState.ILocalTransferCallback callback) {
-        _conf = conf;
-        _context = context;
-        _cb = callback;
+    private final ThreadLocal<KryoTupleDeserializer> _des =
+        new ThreadLocal<KryoTupleDeserializer>() {
+            @Override
+            protected KryoTupleDeserializer initialValue() {
+                return new KryoTupleDeserializer(conf, context);
+            }
+        };
+
+    // Track serialized size of messages.
+    private final boolean sizeMetricsEnabled;
+    private final ConcurrentHashMap<String, AtomicLong> byteCounts = new ConcurrentHashMap<>();
+
+
+    public DeserializingConnectionCallback(final Map conf, final GeneralTopologyContext context, WorkerState.ILocalTransferCallback callback) {
+        this.conf = conf;
+        this.context = context;
+        cb = callback;
+        sizeMetricsEnabled = ObjectReader.getBoolean(conf.get(Config.TOPOLOGY_SERIALIZED_MESSAGE_SIZE_METRICS), false);
+
     }
 
     @Override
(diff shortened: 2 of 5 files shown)


storm-client/src/jvm/org/apache/storm/Config.java
```diff
@@ -451,6 +451,12 @@
     public static final String TOPOLOGY_METRICS_CONSUMER_REGISTER = "topology.metrics.consumer.register";
 
     /**
+     * Enable tracking of network message byte counts per source-destination task
```
revans2: It would be good to add this in to defaults.yaml.  Just so it is clear what the default is.  It would also be nice to explain why it is disabled by default.  At least from the code that is what it looks like is the intention.

storm-client/src/jvm/org/apache/storm/messaging/DeserializingConnectionCallback.java
```diff
@@ -17,44 +17,102 @@
...
+        new ThreadLocal<KryoTupleDeserializer>() {
+            @Override
+            protected KryoTupleDeserializer initialValue() {
+                return new KryoTupleDeserializer(_conf, _context);
+            }
+        };
+
+    // Track serialized size of messages
+    private final boolean _sizeMetricsEnabled;
+    private final ConcurrentHashMap<String, AtomicLong> _byteCounts = new ConcurrentHashMap<>();
```
revans2: I know this fits the convention we have been using for member variables in this file, but it would be good to follow the new style guides and remove the '_'.  If you want to update the other member variables in the file too that would be great.

storm-client/src/jvm/org/apache/storm/messaging/DeserializingConnectionCallback.java
```diff
@@ -17,44 +17,102 @@
...
+
+
+    public DeserializingConnectionCallback(final Map conf, final GeneralTopologyContext context, WorkerState.ILocalTransferCallback callback) {
         _conf = conf;
         _context = context;
         _cb = callback;
+        _sizeMetricsEnabled = _conf.get(Config.TOPOLOGY_SERIALIZED_MESSAGE_SIZE_METRICS) instanceof Boolean &&
```
revans2: It might be simpler to use.

```
sizeMetricsEnabled = ObjectReader.getBoolean(_conf.get(Config.TOPOLOGY_SERIALIZED_MESSAGE_SIZE_METRICS), false);
```

storm-client/src/jvm/org/apache/storm/messaging/DeserializingConnectionCallback.java
```diff
@@ -17,44 +17,102 @@
...
+            ret.add(addrTuple);
         }
         _cb.transfer(ret);
     }
 
+    @Override
+    public Object getValueAndReset() {
+        HashMap<String, Long> outMap = new HashMap<>();
+
+        if (_sizeMetricsEnabled) { // Possible race conditions
```
revans2: Where is the race and what are we doing to mitigate it?  The comment is a bit confusing.

storm-client/src/jvm/org/apache/storm/messaging/DeserializingConnectionCallback.java
```diff
@@ -17,44 +17,102 @@
...
+    /**
+     * Update serialized byte counts for each message
+     * @param sourceTaskId source task
+     * @param message serialized message
+     */
+    protected void updateMetrics(int sourceTaskId, TaskMessage message) {
+        if (_sizeMetricsEnabled) { // Possible race conditions
+            int dest = message.task();
+            int len = message.message().length;
+            String key = Integer.toString(sourceTaskId) + "-" + Integer.toString(dest);
```
jmartell7: Should this use the ComponentId instead?

storm-client/src/jvm/org/apache/storm/messaging/DeserializingConnectionCallback.java
```diff
@@ -17,44 +17,102 @@
...
+        return outMap;
+    }
+
+    /**
+     * Update serialized byte counts for each message
+     * @param sourceTaskId source task
+     * @param message serialized message
+     */
+    protected void updateMetrics(int sourceTaskId, TaskMessage message) {
+        if (_sizeMetricsEnabled) { // Possible race conditions
```
revans2: Here too, the comment is a bit confusing.

storm-client/src/jvm/org/apache/storm/messaging/DeserializingConnectionCallback.java
```diff
@@ -17,44 +17,102 @@
...
+                if (count.get() > 0) {
+                    outMap.put(ent.getKey(), count.getAndSet(0L));
+                }
+            }
+        }
+        return outMap;
+    }
+
+    /**
+     * Update serialized byte counts for each message
```
revans2: nit: please put a '.' at the end of the first sentence in the javadoc.

storm-client/src/jvm/org/apache/storm/messaging/DeserializingConnectionCallback.java
```diff
@@ -17,44 +17,102 @@
...
+            AddressedTuple addrTuple = new AddressedTuple(message.task(), tuple);
+            updateMetrics(tuple.getSourceTask(), message);
+            ret.add(addrTuple);
         }
         _cb.transfer(ret);
     }
 
+    @Override
+    public Object getValueAndReset() {
+        HashMap<String, Long> outMap = new HashMap<>();
```
revans2: Can we return a null if metrics are not enabled instead of an empty map?  A null means the metric will not be sent to the metrics collector, but an empty map might mean that the metrics are disabled or that there was no activity at all.

storm-client/src/jvm/org/apache/storm/messaging/DeserializingConnectionCallback.java
```diff
@@ -17,44 +17,102 @@
...
+     * @param sourceTaskId source task
+     * @param message serialized message
+     */
+    protected void updateMetrics(int sourceTaskId, TaskMessage message) {
+        if (_sizeMetricsEnabled) { // Possible race conditions
+            int dest = message.task();
+            int len = message.message().length;
+            String key = Integer.toString(sourceTaskId) + "-" + Integer.toString(dest);
+
+            AtomicLong count = _byteCounts.get(key);
```
revans2: In java 8 there is now a computeIfAbsent that would make this a lot simpler.

```
byteCounts.computeIfAbsent(key, (key) -> new AtomicLong(0)).addAndGet(len);
```

storm-client/src/jvm/org/apache/storm/messaging/netty/Server.java
```diff
@@ -249,6 +250,15 @@ public Object getState() {
             }
         }
         ret.put("enqueued", enqueued);
+        
+        // Report messageSizes metric
+        if (_cb instanceof IMetric) {
+            Object metrics = ((IMetric) _cb).getValueAndReset();
+            if(metrics instanceof Map && !((Map) metrics).isEmpty()) {
```
revans2: nit: space after the if

revans2: Also fitting with what we had previously.  It would be nice to allow an empty map, but not do anything on null.  This would match what we do in other places working with IMetrics.

HeartSaVioR:
Looks like the patch addresses a bit different point rather than issue description in STORM-2793. 

Below is the issue description in STORM-2793:

> Existing Storm metrics track tuple counts, but don't account for the size of the tuple payload.  If a task always gets large tuples, it can be slower than a task that gets small tuples without showing any reason why.  It would be good to track the byte counts as well so data skew can be observed directly.

from the description I imagined that the issue would want to track the tuples in bytes for every tasks (maybe with sampling), but this patch only addresses tracking `transferred` tuples, which ignores local tasks.

I can still see the benefit with measuring transferred tuples, but just like to check the patch fits origin intention for the issue.

In addition, currently we allow to put map to value in metric and let metrics collector handles it in any way. It should be changed, but we could resolve it along with other metrics later.

Overall looks good to me, and it would be much better if we could see the performance impact and confirm it's small enough.

HeartSaVioR:
And please change the title to 'STORM-2793 Track network data metrics' to conform to https://github.com/apache/storm/blob/master/DEVELOPER.md#create-a-pull-request

jmartell7:
@HeartSaVioR , I've updated the title.  The overhead of serializing the local tuples to measure the size didn't seem worth it, and local tuples generally become a smaller fraction of the total as the number of workers increases.  

I'm working on performance numbers using ThroughputVsLatency, but am open to other ideas.

jmartell7:
Performance didn't seem to be affected by adding the metric. Using ThroughputVsLatency on my MBP (2.3 GHz i7, 16GB RAM) with 35k tuples / sec, 4 workers, 4 splitters, 4 counters, 4 spouts, I saw the following metrics:  Avg latency was 14.3ms with master and 14.0ms with the metric enabled, 99% latency was 17.0 and 16.9 respectively, and cpu was 3.05 and 3.09 respectively.  Additional details are available for the asking.

jmartell7:
Thank you, everyone for the reviews.  @HeartSaVioR , I've squashed the commits.

PR #2401, opened 2017-11-05
STORM-2799: Exclude jdk.tools for JDK 9 compatibility

srdo:
See https://issues.apache.org/jira/browse/STORM-2799.

This excludes jdk.tools from a few dependencies. I also upgraded hbase to the latest patch release of the current version, since the 1.1.0 version leaks jdk.tools https://issues.apache.org/jira/browse/HBASE-13963 and versions prior to 1.1.11 don't work with Java 9 https://issues.apache.org/jira/browse/HBASE-17944.

### DIFF
--- a/external/storm-hbase/pom.xml
+++ b/external/storm-hbase/pom.xml
@@ -55,6 +55,11 @@
                     <groupId>org.apache.zookeeper</groupId>
                     <artifactId>zookeeper</artifactId>
                 </exclusion>
+                <exclusion>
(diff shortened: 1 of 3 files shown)


srdo:
The test failure seems unrelated. I've run the tests a few times on JDK 8 and haven't been able to reproduce the failure.

revans2:
The changes look fine, but the test failure is something I have been seeing quite a bit recently, but only on travis.  I think it must be a race condition.  I am going to file a JIRA for it, but I have no idea when I will have time to look into it.

PR #2403, opened 2017-11-07
STORM-2803: Fix leaking threads from Nimbus/TimeCacheMap, slightly refactor Time to use more final fields, replaced uses of deprecated classes/methods and added a few tests.

srdo:
I'll squash once review is done. Ran this a few times on Travis, and it looks like SlotTest is stable now. The integration test is still failing occasionally. 

The flaky SlotTest was caused by Nimbus not cleaning up 3 TimeCacheMap instances, which have threads that occasionally call Time.sleep. SlotTest depends on the simulated time having certain values, so when the leaked threads slept the tests broke.

Changes: 
* TimeCacheMap threads were leaking from Nimbus, which caused time simulation to be unreliable in later tests, because the leaked threads call Time.sleep. Made sure to kill the threads.
* Replaced TimeCacheMap with RotatingMap in ShellBasedGroupsMapping, since that class has the same problem as Nimbus but also can't be cleaned up. Also added some tests for this class.
* As part of adding tests for ShellBasedGroupsMapping, moved some static functions from ShellUtils to an interface so they can be stubbed in tests.
* Removed uses of deprecated Time.startSimulating/stopSimulating in storm-server.
* Removed duplicate sudo in travis.yml
* Slightly refactored Time so all fields are final, and field resetting happens in the constructor. It seems easier to me to reason about thread safety this way.
* Add check to SlotTest.testResourcesChanged to ensure that the worker heartbeat hasn't expired in the step from WAITING_FOR_WORKER_START to RUNNING. It's a little weird that the test is reusing the original worker heartbeat in that step, so maybe changing the test to create a second heartbeat would be better. Let me know what you think.
* Removed non-functional (as far as I could tell) Windows groups command from ShellUtils and replaced it with an exception. Running "groups" on Windows doesn't do anything to my knowledge.

### DIFF
--- a/storm-client/src/jvm/org/apache/storm/utils/ShellCommandRunner.java
+++ b/storm-client/src/jvm/org/apache/storm/utils/ShellCommandRunner.java
@@ -1,0 +1,63 @@
+/*
+ * Copyright 2017 The Apache Software Foundation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.storm.utils;
+
+import java.io.IOException;
+import java.util.Map;
+
+/**
+ * Contains convenience functions for running shell commands for cases that are too simple to need a full {@link ShellUtils} implementation.
+ */
+public interface ShellCommandRunner {
+    
+    /**
+     * Method to execute a shell command.
+     * Covers most of the simple cases without requiring the user to implement
+     * the {@link ShellUtils} interface.
+     * @param cmd shell command to execute.
+     * @return the output of the executed command.
+     */
+    String execCommand(String ... cmd) throws IOException;
+
+    /**
+     * Method to execute a shell command.
+     * Covers most of the simple cases without requiring the user to implement
+     * the {@link ShellUtils} interface.
+     * @param env the map of environment key=value
+     * @param cmd shell command to execute.
+     * @param timeout time in milliseconds after which script should be marked timeout
+     * @return the output of the executed command.
+     */
+
+    String execCommand(Map<String, String> env, String[] cmd,
+                                     long timeout) throws IOException;
+
+    /**
+     * Method to execute a shell command.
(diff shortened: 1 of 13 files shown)


storm-client/src/jvm/org/apache/storm/utils/ShellCommandRunner.java
```diff
@@ -0,0 +1,63 @@
...
+    String execCommand(String ... cmd) throws IOException;
+
+    /**
+     * Method to execute a shell command.
+     * Covers most of the simple cases without requiring the user to implement
+     * the {@link ShellUtils} interface.
+     * @param env the map of environment key=value
+     * @param cmd shell command to execute.
+     * @param timeout time in milliseconds after which script should be marked timeout
+     * @return the output of the executed command.o
```
revans2: nit: looks like there is an extra 'o' at the end.

storm-client/src/jvm/org/apache/storm/utils/ShellUtils.java
```diff
@@ -432,47 +437,47 @@ private void setTimedOut() {
...
+//     * Covers most of the simple cases without requiring the user to implement
+//     * the <code>Shell</code> interface.
+//     * @param env the map of environment key=value
+//     * @param cmd shell command to execute.
+//     * @return the output of the executed command.
+//     */
+//    public static String execCommand(Map<String,String> env, String ... cmd)
+//        throws IOException {
+//        return execCommand(env, cmd, 0L);
+//    }
```
revans2: Should we remove this?

storm-client/src/jvm/org/apache/storm/utils/ShellUtils.java
```diff
@@ -135,8 +135,10 @@ protected void setWorkingDirectory(File dir) {
 
     /** a Unix command to get the current user's groups list */
     public static String[] getGroupsCommand() {
-        return (WINDOWS)? new String[]{"cmd", "/c", "groups"}
-        : new String[]{"bash", "-c", "groups"};
```
revans2: Is this a regression or did it really not work?

storm-client/src/jvm/org/apache/storm/utils/ShellUtils.java
srdo: Oops, yes.

storm-client/src/jvm/org/apache/storm/utils/ShellUtils.java
srdo: Running "cmd /c groups" or just "groups" on my Windows 10 install just gives me a command not found error. I googled it a bit too, and didn't turn up any references to a "groups" command on Windows. I don't see any references to this method in the codebase either.

storm-client/src/jvm/org/apache/storm/utils/ShellCommandRunner.java
srdo: Thanks, will remove

PR #2404, opened 2017-11-08
[STORM-2804] Fix TopoCache is not caching ACLs correctly problem

Ethanlm:
https://issues.apache.org/jira/browse/STORM-2804

The ACLs might change during `store.createBlob()` but it's not populated to `topos.put()`, which leads to permission problem.

Tested it manually.


Ethanlm:
Don't know why it failed. The build passed on my VM.

PR #2408, opened 2017-11-10
Fix pom dependency indent

jacobtolar:
I noticed that the indent of this dependency doesn't match the rest of the pom.


revans2:
@jacobtolar because this is just a white space change I didn't wait the usual 24 hours.  Thanks for the patch and keep up the good work.

PR #2409, opened 2017-11-10
STORM-2796: Flux: Provide means for invoking static factory methods

ptgoetz:
Note: The guts of this change are in the first commit. The last commit simply addresses check style violations.

### DIFF
--- a/flux/flux-core/src/test/resources/configs/config-methods-test.yaml
+++ b/flux/flux-core/src/test/resources/configs/config-methods-test.yaml
@@ -72,7 +72,14 @@ bolts:
             - "baz"
   - id: "bolt-2"
     className: "org.apache.storm.flux.test.TestBolt"
-    factory: "factory"
+    factory: "newInstance"
+    parallelism: 1
+
+  - id: "bolt-3"
+    className: "org.apache.storm.flux.test.TestBolt"
+    factory: "newInstance"
+#    factoryArgs:
+#      - FOO
     parallelism: 1
 #stream definitions
 # stream definitions define connections between spouts and bolts.
--- a/flux/flux-core/src/test/resources/configs/config-methods-test.yaml
+++ b/flux/flux-core/src/test/resources/configs/config-methods-test.yaml
@@ -81,6 +81,12 @@ bolts:
     factoryArgs:
       - FOO
     parallelism: 1
+
+  - id: "bolt-4"
+    className: "org.apache.storm.flux.test.TestBolt"
+    constructorArgs:
+      - 10
+    parallelism: 1
 #stream definitions
 # stream definitions define connections between spouts and bolts.
 # note that such connections can be cyclical
--- a/flux/flux-core/src/test/resources/configs/config-methods-test.yaml
+++ b/flux/flux-core/src/test/resources/configs/config-methods-test.yaml
@@ -78,8 +78,8 @@ bolts:
   - id: "bolt-3"
     className: "org.apache.storm.flux.test.TestBolt"
     factory: "newInstance"
-#    factoryArgs:
-#      - FOO
+    factoryArgs:
+      - FOO
     parallelism: 1
 #stream definitions
 # stream definitions define connections between spouts and bolts.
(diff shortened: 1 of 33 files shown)


ptgoetz:
I can squash the commits later, but for now keeping the check style commit separate will make review a lot easier.

flux/flux-core/src/test/resources/configs/config-methods-test.yaml
```diff
@@ -70,8 +70,17 @@ bolts:
...
+  - id: "bolt-2"
+    className: "org.apache.storm.flux.test.TestBolt"
+    factory: "newInstance"
+    parallelism: 1
 
-
+  - id: "bolt-3"
+    className: "org.apache.storm.flux.test.TestBolt"
+    factory: "newInstance"
+#    factoryArgs:
```
HeartSaVioR: What if we don't comment this two lines? If it just works, I think this two lines should not be commented out. If it doesn't work, we may need to make it work, cause it is not actual tested.

ptgoetz:
@HeartSaVioR Thanks for the review. Addressed your comment and fixed another issue discovered in testing.

HeartSaVioR:
@ptgoetz Could we add assertions in `TCKtest.testTopologySourceWithConfigMethods()` for `bolt-2`, `bolt-3`, `bolt-4` as well? We've added the configuration but there's no test added.

roshannaik:
I am attaching a patch here on top of this PR that adds some (failing) tests for factory method invocation. Can you please take a look. There are 4 new tests. 

- invoking a config method on the testbolt that takes the same arg as above
- invoking a config method on the testbolt that takes the array of args
- 2 similar tests for setting properties..  1 taking object and another a reflist.

The verified first test is working. Can you look into the other rest.. not sure if got something wrong there. The 

[test.patch.txt](https://github.com/apache/storm/files/1476952/test.patch.txt)

ptgoetz:
@HeartSaVioR Yes, I will add assertions, although the tests would still fail without them (the constructor/method won't be found and a IllegalArgumentException will be thrown).

@roshannaik Thanks for the patch. I'll take a look.

ptgoetz:
@roshannaik See latest commit. I merged/fixed your patch. 

Most of the issues were typos (e.g. "refList" != "reflist"). The other issue was that properties in flux do not (currently) support ref lists. However, you can work around it by invoking the setter as a config method:

```yaml
    configMethods:
      - name: "setTimeLenArr"
        args:
          - reflist: ["time1", "time2"]
```

We can probably handle property ref list support as a separate issue since there is a fairly easy workaround.

roshannaik:
I am ok with handling the property limitation in a separate issue. Perhaps we could document what is not yet supported and open a jira for it.

hmcl:
@ptgoetz I would suggest putting in different patches the that the minor refactoring around coding style and adding adding Javadoc. Also, can you please squash the commits once you have the necessary +1 to get the patches merged. Thanks.

HeartSaVioR:
+1
The PR is authored by one, so it doesn't matter merger squashes commits into one with proper commit message. I don't like having more interaction phases between reviewers (+ merger) and author, especially the interaction doesn't take place in time. Let's be less synchronized.

HeartSaVioR:
Filed issue for supporting reflist in properties.
https://issues.apache.org/jira/browse/STORM-2849

HeartSaVioR:
Also ported back to 1.x-branch via https://github.com/apache/storm/commit/b48421e71258b670062d280efd1084bb52738e9c

(Resolved merge conflicts and replace java.util.Duration with storm-core's one.)

roshannaik:
Thanks @ptgoetz !

hmcl:
@HeartSaVioR there is not one major project that does not require contributors to merge commits. It takes a few minutes and it makes a world of difference in terms of making the git log easy to understand, and most importantly, easy to cherry-pick.

HeartSaVioR:
@hmcl 
I'm on the same page regarding needs for squashing: even in this PR I squashed the commits before merging. I also support merge script and I have been considering to initiate discussion again.

Most of cases it's just a matter of who to squash, cause the result of squashing is just same regardless of who to squash. Ideally we can request contributors to do that, but in reality I have been encountered so many times I've visited the PR too late (say more than 1 month) and contributor becomes not active as it was, or even gone. In this case requiring something would take several days or even weeks, and in worst case PR becomes forgotten again.

PR #2413, opened 2017-11-11
STORM-2810: Fix resource leaks in storm-hdfs tests

srdo:
https://issues.apache.org/jira/browse/STORM-2810

The tests are leaking at least threads and HDFS file locks. Instead of cleaning up everything manually, I've made sure threads are cleaned up, and have switched the tests that use locks to start a new DFS cluster per test instead of per test class. 

I ran the tests on Travis 3 times without failure, and 100 times on a local Ubuntu VM.


srdo:
Reviewers might want to add ?w=1 to the diff url to disable whitespace diffing, some of the modified files were autoformatted from 2-space indentation to 4-space.

PR #2414, opened 2017-11-11
Fixing Java unit tests in storm-core to run under Windows (1.x branch)

lawrencecraft:
While fixing STORM-2797 I noticed a bunch of the storm-core tests didn't run under Windows.

I've fixed them up to use Windows-friendly asserts. All are done in a generic way through framework classes except one particular case where the behavior actually differs based on operating system. I only modified tests, not any actual application code. I tested this change on Windows, OS X, and Linux: all pass. On Windows, I still have to run Maven as administrator as some of the tests write to temp folders in AppData.

I'll follow up with a pull request for STORM-2797. There are still some Clojure tests failing but a lot of them are around the logviewer, which will be fixed by my next PR.

### DIFF
--- a/storm-core/test/jvm/org/apache/storm/daemon/supervisor/BasicContainerTest.java
+++ b/storm-core/test/jvm/org/apache/storm/daemon/supervisor/BasicContainerTest.java
@@ -370,7 +370,7 @@ public class BasicContainerTest {
         final String workerConf = ContainerTest.asAbsPath(log4jdir, "worker.xml");
         final String workerRoot = ContainerTest.asAbsPath(stormLocal, "workers", workerId);
         final String workerTmpDir = ContainerTest.asAbsPath(workerRoot, "tmp");
-        
+
         final StormTopology st = new StormTopology();
         st.set_spouts(new HashMap<String, SpoutSpec>());
         st.set_bolts(new HashMap<String, Bolt>());
@@ -407,7 +407,7 @@ public class BasicContainerTest {
             assertListEquals(Arrays.asList(
                     "java",
                     "-cp",
-                    "FRAMEWORK_CP:" + stormjar.getAbsolutePath(),
+                    "FRAMEWORK_CP" + File.pathSeparator + stormjar.getAbsolutePath(),
                     "-Dlogging.sensitivity=S3",
                     "-Dlogfile.name=worker.log",
                     "-Dstorm.home=" + stormHome,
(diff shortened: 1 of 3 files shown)


storm-core/test/jvm/org/apache/storm/localizer/AsyncLocalizerTest.java
```diff
@@ -62,15 +62,15 @@ public void testRequestDownloadBaseTopologyBlobs() throws Exception {
         final String jarKey = topoId + "-stormjar.jar";
         final String codeKey = topoId + "-stormcode.ser";
         final String confKey = topoId + "-stormconf.ser";
-        final String stormLocal = "/tmp/storm-local/";
-        final String stormRoot = stormLocal+topoId+"/";
+        final String stormLocalAbsolutePath = new File("/tmp/storm-local/").getAbsolutePath();
```
srdo: Use `Files.createTempFile` instead of `/tmp`.

storm-core/test/jvm/org/apache/storm/daemon/supervisor/ContainerTest.java
```diff
@@ -125,7 +125,8 @@ public void testKill() throws Exception {
     private static final Joiner PATH_JOIN = Joiner.on(File.separator).skipNulls();
     private static final String DOUBLE_SEP = File.separator + File.separator;    
     static String asAbsPath(String ... parts) {
-        return (File.separator + PATH_JOIN.join(parts)).replace(DOUBLE_SEP, File.separator);
+        String path = (File.separator + PATH_JOIN.join(parts)).replace(DOUBLE_SEP, File.separator);
```
srdo: I think starting a path with `/` or `\` is inherently going to be fragile on Windows. It might be better to remove the initial file separator from here, and throw an exception if the first path part is `File.separator` . Then you can probably also get rid of the `DOUBLE_SEP` replacement. Same check should be in `asPath`. In order to work on Windows, paths should start from e.g. `java.io.tmpdir` or the current working directory instead of the filesystem root I think.

storm-core/test/jvm/org/apache/storm/daemon/supervisor/BasicContainerTest.java
```diff
@@ -370,7 +370,7 @@ public void testLaunch() throws Exception {
         final String workerConf = ContainerTest.asAbsPath(log4jdir, "worker.xml");
```
srdo: If you start the paths from `Files.createTempFile` instead of ``tmp` for `stormHome` and `stormLocal`, you should be able to get rid of the `prependFilePrefixIfOnWindows` method.

lawrencecraft:
Thanks Stig, will do. 

I might submit a fix for STORM-2797 (the LogViewer issue) first, if that's alright, as I've got that almost ready to go. I can track the fixing of the unit tests separately.

srdo:
Sounds good.

PR #2415, opened 2017-11-12
STORM-2811: Fix NPE in Nimbus when killing the same topology multiple…

srdo:
… times, fix integration test killing the same topology multiple times

https://issues.apache.org/jira/browse/STORM-2811

Some of the integration tests don't create a TopoWrap, which causes the cleanup code to kill the same topology multiple times.

The Nimbus client's behavior when the server throws exceptions was a little surprising to me. It looks like the client lost connection to the server after the NPE was thrown, so the subsequent atttempts to kill the topology just got TTransportException (I'm guessing the socket was closed). I'm not sure if this is expected behavior?

### DIFF
--- a/integration-test/src/main/java/org/apache/storm/st/utils/StringDecorator.java
+++ b/integration-test/src/main/java/org/apache/storm/st/utils/StringDecorator.java
@@ -17,7 +17,6 @@
 
 package org.apache.storm.st.utils;
 
-import java.nio.charset.StandardCharsets;
 import org.apache.commons.lang.StringUtils;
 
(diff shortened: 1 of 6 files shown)


srdo:
The test failure is likely due to https://issues.apache.org/jira/browse/STORM-2809. 

I ran this change + a revert of https://github.com/apache/storm/pull/2363 6 times on Travis, and the only failure was when Travis somehow failed to start Maven. 14/15 runs locally (with the same revert) have passed, will investigate why the last run failed.

PR #2417, opened 2017-11-13
STORM-2809:  Always create the resources directory so we can check for it

revans2:
The issue was actually a few lines below this.  In the isFullyDownloaded function.  It would check if we wanted to extract things, like with the resources jar, that the destination directory existed.  But that directory would only be created if there was something to extract.  Some topology jars didn't have anything to extract so the directory would never be created and the worker would be relaunched every 30 seconds.


srdo:
Thanks for looking into this. I'll try running the integration test locally on it a number of times tomorrow.

HeartSaVioR:
I'm also running the build in my fork. Will run several times and see it's good.
Btw, Stig stated that integration test was failing consistently, then successful result of build in this PR may indicate the patch will resolve the integration test.

HeartSaVioR:
First trial starts here: https://travis-ci.org/HeartSaVioR/storm/builds/301792774

srdo:
+1.

Ran the integration tests 15 times locally, and 5 failed because the worker didn't start. It looks unrelated to both this and https://github.com/apache/storm/pull/2363. Will try enabling more logs to see if I can figure out what's happening.

srdo:
Thanks @revans2, merged to master.

PR #2419, opened 2017-11-14
STORM-2805: Clean up confs in TopologyBuilders

revans2:


storm-client/src/jvm/org/apache/storm/coordination/BatchSubtopologyBuilder.java
```diff
@@ -84,24 +84,24 @@ private BoltDeclarer setBolt(String id, IRichBolt bolt, Number parallelism) {
...
     public void extendTopology(TopologyBuilder builder) {
-        BoltDeclarer declarer = builder.setBolt(_masterId, new CoordinatedBolt(_masterBolt.bolt), _masterBolt.parallelism);
-        for (InputDeclaration decl: _masterBolt.declarations) {
+        BoltDeclarer declarer = builder.setBolt(masterId, new CoordinatedBolt(masterBolt.bolt), masterBolt.parallelism);
+        for (InputDeclaration decl: masterBolt.declarations) {
             decl.declare(declarer);
         }
-        for (Map<String, Object> conf: _masterBolt.componentConfs) {
-            declarer.addConfigurations(conf);
+        if (!masterBolt.componentConf.isEmpty()) {
```
srdo: Nit: addConfigurations seems like it does this check too, so it seems unnecessary here

storm-client/src/jvm/org/apache/storm/coordination/BatchSubtopologyBuilder.java
```diff
@@ -446,31 +446,26 @@ public String getComponent() {
...
         
         private void addDeclaration(InputDeclaration declaration) {
-            _component.declarations.add(declaration);
+            component.declarations.add(declaration);
         }
 
         @Override
         public BoltDeclarer addConfigurations(Map<String, Object> conf) {
-            _component.componentConfs.add(conf);
+            if (conf != null && !conf.isEmpty()) {
```
srdo: Nit: I think putAll is a no-op when the conf is empty, so you probably don't need to check.

storm-client/src/jvm/org/apache/storm/coordination/BatchSubtopologyBuilder.java
```diff
@@ -446,31 +446,26 @@ public String getComponent() {
...
-                if (conf.containsKey(Config.TOPOLOGY_COMPONENT_RESOURCES_MAP)) {
-                    return conf;
-                }
-            }
-            Map<String, Object> newConf = new HashMap<>();
-            newConf.put(Config.TOPOLOGY_COMPONENT_RESOURCES_MAP, new HashMap());
-            _component.componentConfs.add(newConf);
-            return newConf;
+        public Map<String, Object> getRASConfiguration() {
+            //TODO this should not be modifiable
```
srdo: Would be good to fix this TODO before merging, or at least remove the TODO and raise an issue. I think it'll be forgotten if it's left here.

storm-client/src/jvm/org/apache/storm/drpc/LinearDRPCTopologyBuilder.java
```diff
@@ -53,12 +53,12 @@
 // Trident subsumes the functionality provided by this class, so it's deprecated
 @Deprecated
```
srdo: Not directly related to this PR, but this class has been deprecated for a while. Maybe we should consider removing some of the deprecated classes for 2.0.0?

storm-client/src/jvm/org/apache/storm/trident/topology/TridentTopologyBuilder.java
```diff
@@ -51,8 +51,8 @@
 
 // based on transactional topologies
 public class TridentTopologyBuilder {
-    Map<GlobalStreamId, String> _batchIds = new HashMap();
-    Map<String, TransactionalSpoutComponent> _spouts = new HashMap();
+    Map<GlobalStreamId, String> batchIds = new HashMap();
```
srdo: Nit: Raw types here and in the following lines

storm-client/src/jvm/org/apache/storm/trident/topology/TridentTopologyBuilder.java
```diff
@@ -104,8 +104,8 @@ public static String spoutIdFromCoordinatorId(String coordId) {
     }
     
     Map<GlobalStreamId, String> fleshOutStreamBatchIds(boolean includeCommitStream) {
-        Map<GlobalStreamId, String> ret = new HashMap<>(_batchIds);
-        Set<String> allBatches = new HashSet(_batchIds.values());
+        Map<GlobalStreamId, String> ret = new HashMap<>(batchIds);
+        Set<String> allBatches = new HashSet(batchIds.values());
```
srdo: Also raw here.

storm-client/src/jvm/org/apache/storm/coordination/BatchSubtopologyBuilder.java
revans2: It is probably unnecessary and I will pull it out.  I left it in initially for compatibility, why try to add an empty config, which will be a no-op.

storm-client/src/jvm/org/apache/storm/coordination/BatchSubtopologyBuilder.java
revans2: Yes thank you I forgot to grep the patch for TODOs.

storm-client/src/jvm/org/apache/storm/drpc/LinearDRPCTopologyBuilder.java
revans2: I'll file some separate JIRAs to remove LinerarDRPCTopologyBuilder and TransactionalTopologies that too has been deprecated for a while too.

revans2:
@srdo I think I have addressed all of your comments.  I also found a few places were an NPE could happen, and one where the TopologyBuilder was not setting the resource configs correctly.  I fixed it and the corresponding test failures.

storm-server/src/test/java/org/apache/storm/scheduler/resource/strategies/scheduling/TestGenericResourceAwareStrategy.java
```diff
@@ -129,23 +130,34 @@ public void testGenericResourceAwareStrategySharedMemory() {
...
-        assertEquals(totalExpectedCPU, resources.get_cpu(), 0.01);
-        assertEquals(totalExpectedOnHeap, resources.get_mem_on_heap(), 0.01);
-        assertEquals(totalExpectedWorkerOffHeap, resources.get_mem_off_heap(), 0.01);
-        assertEquals(sharedOnHeap, resources.get_shared_mem_on_heap(), 0.01);
-        assertEquals(sharedOffHeapWorker, resources.get_shared_mem_off_heap(), 0.01);
+        Set<WorkerSlot> slots = assignment.getSlots();
+        Map<String, Double> nodeToTotalShared = assignment.getNodeIdToTotalSharedOffHeapMemory();
+        LOG.info("NODE TO SHARED OFF HEAP {}", nodeToTotalShared);
+        Map<WorkerSlot, WorkerResources> scheduledResources = assignment.getScheduledResources();
+        assertEquals(2, slots.size());
```
srdo: The comment in line 126 says everything should fit in a single slot. Why does the assignment use two slots now?

storm-server/src/test/java/org/apache/storm/scheduler/resource/strategies/scheduling/TestGenericResourceAwareStrategy.java
```diff
@@ -187,16 +199,21 @@ public void testGenericResourceAwareStrategy() {
         rs.prepare(conf);
         rs.schedule(topologies, cluster);
 
+        //We need to have 3 slots on 3 separate hosts to make the GPU situation work
+
         HashSet<HashSet<ExecutorDetails>> expectedScheduling = new HashSet<>();
-        expectedScheduling.add(new HashSet<>(Arrays.asList(new ExecutorDetails(0, 0)))); //Spout
+        expectedScheduling.add(new HashSet<>(Arrays.asList(new ExecutorDetails(3, 3)))); //bolt-3 - 500 MB, 50% CPU, 2 GPU
```
srdo: Why is bolt 3 task 3 and 4 instead of 5 and 6 as before?

storm-client/src/jvm/org/apache/storm/trident/topology/MasterBatchCoordinator.java
```diff
@@ -308,7 +308,7 @@ public String toString() {
                 ", _maxTransactionActive=" + _maxTransactionActive +
                 ", _coordinators=" + _coordinators +
                 ", _managedSpoutIds=" + _managedSpoutIds +
-                ", _spouts=" + _spouts +
+                ", spouts=" + _spouts +
```
Ethanlm: Do we want to get rid of the underscores of other variables? I wonder why only `spouts`

storm-client/src/jvm/org/apache/storm/topology/TopologyBuilder.java
```diff
@@ -562,43 +560,54 @@ private void initCommon(String id, IComponent component, Number parallelism) thr
...
-
-            getRASConfiguration().put(Config.TOPOLOGY_COMPONENT_RESOURCES_MAP, resourcesMap);
+        public T addResources(Map<String, Double> resources) {
+            if (resources != null && !resources.isEmpty()) {
+                String currConf = commons.get(id).get_json_conf();
+                Map<String, Object> conf = parseJson(currConf);
+                Map<String, Double> currentResources =
+                    (Map<String, Double>) conf.computeIfAbsent(Config.TOPOLOGY_COMPONENT_RESOURCES_MAP, (k) -> new HashMap<>());
+                currentResources.putAll(resources);
+                commons.get(id).set_json_conf(JSONValue.toJSONString(currConf));
```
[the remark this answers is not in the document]
revans2: You are right.  Fixing it...

storm-server/src/test/java/org/apache/storm/scheduler/resource/strategies/scheduling/TestGenericResourceAwareStrategy.java
revans2: I'll fix the comment.  The test was wrong because the GPU resources were not being recorded properly.  

Each supervisor has a single GPU.  Each spout needs a GPU, but there are 2 spouts, so it cannot be on a single node, and hence cannot be on a single slot.  I can either up the number of CPUs per node to 2 and leave the rest of the test alone, or I can update the comment.

storm-client/src/jvm/org/apache/storm/trident/topology/MasterBatchCoordinator.java
revans2: This was probably the IDE getting overly ambitious when I refactored something else.  I'll fix it.

storm-server/src/test/java/org/apache/storm/scheduler/resource/strategies/scheduling/TestGenericResourceAwareStrategy.java
revans2: This test was also wrong because it was not taking the GPUs into account.   That is why I called out all of the resources in the comments for the scheduling.  bolt-3 requires 2 GPUs.  Each node has 2 GPUs on it.  Having 2 instances of a bolt-3 on the same node, which is what the scheduling was before, would go over the GPU count for the node.

storm-server/src/test/java/org/apache/storm/scheduler/resource/strategies/scheduling/TestGenericResourceAwareStrategy.java
srdo: Thanks for explaining.

storm-server/src/test/java/org/apache/storm/scheduler/resource/strategies/scheduling/TestGenericResourceAwareStrategy.java
srdo: That makes sense. What I'm unsure about is what determines the task numbering, I'm not understanding why the numbering is spout - 0, bolt1 - 1, 2, bolt2 - 5, 6, bolt3 - 3, 4. I would have thought they'd be numbered in declaration order?

`genExecsAndComps` enumerates the tasks by counting the parallelism hint for each bolt but it iterates the bolts via a map entryset. I think iteration order for those is undefined except for specific map types, and it looks like TopologyBuilder is using a HashMap. It might make this test unstable, or at least be inconsistent for different JVMs?

revans2:
@srdo I have fixed the issues you called out

@Ethanlm I fixed the issue you saw.

srdo:
Thanks, +1

revans2:
@srdo thanks for the review you found a lot of bugs and made this a lot better.

srdo:
Thanks, happy to help :)

PR #2420, opened 2017-11-14
[STORM-2814] Logviewer HTTP server should return 403 instead of 200 if the user is unauthorized

Ethanlm:
A trivial fix. Change the return HTTP Code from 200 to 403.
Also fixed the typos

https://issues.apache.org/jira/browse/STORM-2814


Ethanlm:
Last commit changed other place from using 401 to 403 so that we can have a consistent http code for AuthorizedExceptions.

HeartSaVioR:
Forgot to +1 here. +1

PR #2422, opened 2017-11-15
[STORM-2814] Logviewer HTTP server should return 403 if the user is unauthorized

Ethanlm:
Similar to https://github.com/apache/storm/pull/2420 . This is for  1.x-branch.


Ethanlm:
It seems that this patch should be able to merge into 1.1.x-branch easily. If there is a merge issue, I would like to file a separate pull request for it.

PR #2427, opened 2017-11-19
MINOR: Use booleans instead of strings for 'enable.auto.commit' setting

srdo:
Kafka understands booleans in this property just fine, there's no reason to use strings instead.


hmcl:
+1

srdo:
Thanks, applied to master and 1.x

PR #2430, opened 2017-11-20
[STORM-2827] fix Logviewer search returning incorrect logviewerUrl problem

Ethanlm:
https://issues.apache.org/jira/browse/STORM-2827

The logviewer search returns "http" url no matter whether logviewer is using https or not, which could lead to 404 not found error.

Tested it manually

### DIFF
--- a/storm-webapp/src/main/java/org/apache/storm/daemon/logviewer/handler/LogviewerLogSearchHandler.java
+++ b/storm-webapp/src/main/java/org/apache/storm/daemon/logviewer/handler/LogviewerLogSearchHandler.java
@@ -82,6 +82,7 @@ public class LogviewerLogSearchHandler {
     private final String daemonLogRoot;
     private final ResourceAuthorizer resourceAuthorizer;
     private final Integer logviewerPort;
+    private final String scheme;
 
     /**
      * Constructor.
@@ -97,8 +98,14 @@ public class LogviewerLogSearchHandler {
         this.logRoot = logRoot;
         this.daemonLogRoot = daemonLogRoot;
         this.resourceAuthorizer = resourceAuthorizer;
-
-        this.logviewerPort = ObjectReader.getInt(stormConf.get(DaemonConfig.LOGVIEWER_PORT));
+        Object httpsPort = stormConf.get(DaemonConfig.LOGVIEWER_HTTPS_PORT);
+        if (httpsPort == null) {
+            this.logviewerPort = ObjectReader.getInt(stormConf.get(DaemonConfig.LOGVIEWER_PORT));
+            this.scheme = "http";
+        } else {
+            this.logviewerPort = ObjectReader.getInt(httpsPort);
+            this.scheme = "https";
+        }
     }
 
     /**
@@ -662,7 +669,7 @@ public class LogviewerLogSearchHandler {
         parameters.put("start", Math.max(0, offset - (LogviewerConstant.DEFAULT_BYTES_PER_PAGE / 2) - (needle.length / -2)));
         parameters.put("length", LogviewerConstant.DEFAULT_BYTES_PER_PAGE);
 
-        return UrlBuilder.build(String.format("http://%s:%d/api/v1/log", host, port), parameters);
+        return UrlBuilder.build(String.format(this.scheme + "://%s:%d/api/v1/log", host, port), parameters);
     }
 
     @VisibleForTesting
@@ -675,7 +682,7 @@ public class LogviewerLogSearchHandler {
         parameters.put("start", Math.max(0, offset - (LogviewerConstant.DEFAULT_BYTES_PER_PAGE / 2) - (needle.length / -2)));
         parameters.put("length", LogviewerConstant.DEFAULT_BYTES_PER_PAGE);
 
-        return UrlBuilder.build(String.format("http://%s:%d/api/v1/daemonlog", host, port), parameters);
+        return UrlBuilder.build(String.format(this.scheme + "://%s:%d/api/v1/daemonlog", host, port), parameters);
     }
 
     @VisibleForTesting
--- a/storm-webapp/src/main/java/org/apache/storm/daemon/logviewer/LogviewerServer.java
+++ b/storm-webapp/src/main/java/org/apache/storm/daemon/logviewer/LogviewerServer.java
@@ -155,7 +155,7 @@ public class LogviewerServer implements AutoCloseable {
      */
     public static void main(String [] args) throws Exception {
         Utils.setupDefaultUncaughtExceptionHandler();
-        Map<String, Object> conf = Utils.readStormConfig();
+        Map<String, Object> conf = ConfigUtils.readStormConfig();
 
         String logRoot = ConfigUtils.workerArtifactsRoot(conf);
         File logRootFile = new File(logRoot);


Ethanlm:
The travis test passed on my repo: https://travis-ci.org/Ethanlm/storm/builds/304993981
Don't know why it failed here.

storm-webapp/src/main/java/org/apache/storm/daemon/logviewer/handler/LogviewerLogSearchHandler.java
```diff
@@ -97,8 +98,14 @@ public LogviewerLogSearchHandler(Map<String, Object> stormConf, String logRoot,
         this.logRoot = logRoot;
         this.daemonLogRoot = daemonLogRoot;
         this.resourceAuthorizer = resourceAuthorizer;
-
-        this.logviewerPort = ObjectReader.getInt(stormConf.get(DaemonConfig.LOGVIEWER_PORT));
+        Integer httpsPort = ObjectReader.getInt(stormConf.get(DaemonConfig.LOGVIEWER_HTTPS_PORT), 0);
```
srdo: It seems a little odd to me to use getInt with a default, and then check if the value is the default. Why not check if stormConf.get(...) returns null instead?

srdo:
I think some of the tests are still flaky.

storm-webapp/src/main/java/org/apache/storm/daemon/logviewer/handler/LogviewerLogSearchHandler.java
Ethanlm: It's possible that users set it to a non-positive number. We need to check them both.

srdo: You're right. I thought Logviewer was using the same validation code as storm-client and storm-server to ensure that stormConf was valid according to the annotations in DaemonConfig (e.g. LOGVIEWER_HTTPS_PORT is configured to use validation for "@isInteger" and "@isPositiveNumber".

Do you know if there's a reason why we aren't using the usual validation (this method https://github.com/apache/storm/blob/master/storm-client/src/jvm/org/apache/storm/validation/ConfigValidation.java#L710)?

revans2: @srdo I think the issue is around converting the number to an Integer.  If `ObjectReader.getInt` does not have a default option it will throw an exception on a null value which is the default.  But I think we could make it cleaner by doing something like

```
Object httpsPort = stormConf.get(DaemonConfig.LOGVIEWER_HTTPS_PORT);
if (httpsPort == null) {
    this.logviewerPort = ObjectReader.getInt(stormConf.get(DaemonConfig.LOGVIEWER_PORT));
    this.scheme = "http";
} else {
    this.logviewerPort = ObjectReader.getInt(httpsPort);
    this.scheme = "https";
}
```

revans2: @Ethanlm 

It can only be `null` or a non-zero positive number.

https://github.com/apache/storm/blob/61a944fd15e564482a5c55655daad9736948d961/storm-server/src/main/java/org/apache/storm/DaemonConfig.java#L353

srdo: @revans2 Your code snippet is more or less what I had in mind. I think @Ethanlm has a point too though. I don't think the validation annotations are checked anywhere in the Logviewer process.

The stormConf seems to be read here https://github.com/apache/storm/blob/master/storm-webapp/src/main/java/org/apache/storm/daemon/logviewer/LogviewerServer.java#L158 but I don't see any references to ConfigValidation in storm-webapp. 

I think we should consider running ConfigValidation.validateFields in the LogviewerServer main, so we can trust that the annotation constraints are enforced.

Ethanlm: @srdo @revans2  Thanks for the comments.  It seems to me that `ConfigValidation.validateFields` is only used to validate topology configurations. It's not used to validate the confs daemons use before launching the daemons.

revans2: Actually in the old clojure code we did validate it every time we loaded it.

https://github.com/apache/storm/blob/67c65e774f20d3091caed6da82d155887e966772/storm-core/src/clj/org/apache/storm/config.clj#L72-L80

Not sure what happened with the translation but it is a regression if we are not doing it any more.

Ethanlm: We have two `readStormConfig()`. One is https://github.com/apache/storm/blob/master/storm-client/src/jvm/org/apache/storm/utils/ConfigUtils.java#L135
The other one is 

https://github.com/apache/storm/blob/master/storm-client/src/jvm/org/apache/storm/utils/Utils.java#L262

It seems like we are using the second one everywhere. But we are supposed to use the first one

HeartSaVioR: @Ethanlm 
You're right about using which method to use: Only first one validates the fields. We should fix it, maybe better to rename `Utils.readStormConfig()` or even move the logic to ConfigUtils  if we don't find any place which should use Utils.readStormConfig() instead of ConfigUtils'. 

Btw, `ConfigValidation.validateFields` also validates DaemonConfig, if we have storm-server in classpath. Please refer https://issues.apache.org/jira/browse/STORM-832 and its patch https://github.com/apache/storm/pull/2065.

Ethanlm: It still seems not right to me. If we change `Utils.readStormConfig()` to `ConfigUtil.readStormConfig()` in `LogviewerServer`, we require the `LOGVIEWER_HTTPS_PORT` to be `@isInteger` and `@isPositiveNumber`. Then we will have to use https for logviewer in any cases and never be able to use http. See: https://github.com/apache/storm/blob/master/storm-core/src/jvm/org/apache/storm/ui/UIHelpers.java#L239

srdo: @Ethanlm I don't understand why? As far as I can tell both `@isInteger` (https://github.com/apache/storm/blob/master/storm-client/src/jvm/org/apache/storm/validation/ConfigValidation.java#L160) and `@isPositiveNumber` (https://github.com/apache/storm/blob/master/storm-client/src/jvm/org/apache/storm/validation/ConfigValidation.java#L450) allow nulls to pass the validation check. 

The configuration we'd disallow would be setting the https port to a negative number, which seems reasonable enough to me.

Ethanlm: Then it should be fine. I thought `@isInteger` and `isPositiveNumber` don't allow `null` value

Ethanlm:
I filed a separate jira for readStormConfig() related issue because it looks like there are many places using wrong function. We need to identify them and change them.
JIRA: https://issues.apache.org/jira/browse/STORM-2832

srdo:
+1, thanks for filing the followup issue too.

Ethanlm:
@srdo Thanks for the review.
@revans2 @HeartSaVioR  Could you please take a second look? Thanks

srdo:
Thanks @Ethanlm, merged to master.

PR #2432, opened 2017-11-21
[STORM-2829] fix logviewer deepSearch direct self-reference leading to cycle exception

Ethanlm:
https://issues.apache.org/jira/browse/STORM-2829

```
2017-11-21 21:06:19.369 o.e.j.s.HttpChannel qtp1471948789-17 [WARN] /api/v1/deepSearch/wc-1-1511188542
javax.servlet.ServletException: java.lang.RuntimeException: com.fasterxml.jackson.databind.JsonMappingException: Direct self-reference leading to cycle (through reference chain: org.apache.storm.daemon.logviewer.handler.Matched["matches"]->java.util.ArrayList[0]->java.util.HashMap["port"]->sun.nio.fs.UnixPath["fileSystem"]->sun.nio.fs.LinuxFileSystem["rootDirectories"]->sun.nio.fs.UnixPath["root"])
```

We are having the above exception because 
`truncatePathToLastElements(firstLogAbsPath, 2).getName(0)` returns a Path object. Then `OBJECT_MAPPER.writeValueAsString()` (in `Matched` `toJSONString()`) somehow interprets the path to `sun.nio.fs.UnixPath["fileSystem"]` and etc... which causes the exception. 

Tested it manually. With this patch, the deepSearch should works fine.
 

image

### DIFF
--- a/storm-webapp/src/main/java/org/apache/storm/daemon/logviewer/handler/LogviewerLogSearchHandler.java
+++ b/storm-webapp/src/main/java/org/apache/storm/daemon/logviewer/handler/LogviewerLogSearchHandler.java
@@ -414,7 +414,7 @@ public class LogviewerLogSearchHandler {
             } catch (IOException e) {
                 throw new RuntimeException(e);
             }
-            currentFileMatch.put("port", truncatePathToLastElements(firstLogAbsPath, 2).getName(0));
+            currentFileMatch.put("port", truncatePathToLastElements(firstLogAbsPath, 2).getName(0).toString());
             newMatches.add(currentFileMatch);
 
             int newCount = matchCount + ((List<?>)theseMatches.get("matches")).size();


srdo:
+1

My bad, the change to Path was made in https://github.com/apache/storm/pull/2360. @lawrencecraft you might want to be aware of this if you're still working on backporting logviewer changes to 1.x.

srdo:
I think we can avoid this issue in the future by upgrading Jackson https://github.com/FasterXML/jackson-databind/issues/373. The fix should be included as of version 2.7, which is when JDK7 basic classes became part of core jackson-databind.

Ethanlm:
@srdo @HeartSaVioR Thanks for the review

srdo:
Thanks @Ethanlm, merged to master

PR #2437, opened 2017-11-28
[STORM-2834] getOwnerResourceSummaries not working properly because scheduler is wrapped as BlacklistScheduler

Ethanlm:
https://issues.apache.org/jira/browse/STORM-2834

The condition of `scheduler instanceof ResourceAwareScheduler` or `scheduler instanceof MultitenantScheduler` will never be true because the scheduler is wrapped as BlacklistScheduler

Tested it manually.


govind-menon:
@revans2  There is now - https://issues.apache.org/jira/browse/STORM-2840

PR #2438, opened 2017-11-29
STORM-2835: storm-kafka-client KafkaSpout can fail to remove all tuples from waitingToEmit

hmcl:


srdo:
Could you clarify what the bug is here? I'm having trouble understanding why this could cause issues. `waitingToEmit` is an iterator on a list that we don't use for anything else, so I'm not clear on why we even need to call `waitingToEmit.remove()`?

hmcl:
Yes, you are right. Technically it is not necessary to remove the elements from the collection. Nevertheless, this code change addresses the following, important, issues:

1. Calling next() before hasNext() can throw NoSuchElementException
2. It is good practice remove the elements from the iterator (list) for two reasons
   a.  to relief memory pressure. Otherwise, until a new poll happens, all the tuples will remain in memory
   b. It makes sense from a system internals standpoint: Once you emit the record/tuple, it should no longer be in the waitingToEmit. Why should it be on this list, if it has already been emitted?
3. The code has is is a much more common pattern, and hence better practice as compared to the previous one.

The only reason I can think not to remove the elements from this list is performance. If we feel it gives that much more throughput, we can can consider not removing the elements. However, even if we don't remove the elements the code should be along the lines (perhaps refactoring a bit to avoid the empty body in the while loop)

```java
while (waitingToEmit.hasNext() && !emitTupleIfNotEmitted(waitingToEmit.next()));
```

HeartSaVioR:
So there's `hasNext()` in `waitingToEmit()` which made me confused...

I agree technically the patch may not be required, but I'd like to propose removing `hasNext()` from `waitingToEmit()` and adopt this patch, which helps to read `emit()` method.

And I prefer throwing out the list instead of removing elements if we discard the list just after loop end, but this is not the case so I'm +1 to remove element when element is not needed.

srdo:
@hmcl Thanks for elaborating.

As @HeartSaVioR mentions, there's a `hasNext` call in `waitingToEmit()`, which is called before this loop, so I don't think it's possible that we call `next` without calling `hasNext` first. That said, I agree that this change makes the code easier to read, so we should make the change regardless.

Regarding `waitingToEmit.remove`, we aren't just failing to remove the last element from the list. We end up back in this loop in the next call to `nextTuple` if there are still unemitted records, so we're failing to remove any element that results in a tuple getting emitted, since we skip `waitingToEmit.remove` when `emitTupleIfNotEmitted` returns true. Since we never read any element from `waitingToEmit` more than once, I think we should remove elements as soon as they're read, i.e. something like 
```java
while (waitingToEmit.hasNext()) {
  boolean emitted = emitTupleIfNotEmitted(waitingToEmit.next());
  waitingToEmit.remove();
  if (emitted) {
    break;
  }
}
```

@HeartSaVioR We might need to keep the `hasNext` call in `waitingToEmit()` since `waitingToEmit()` is also used to decide whether to poll for new records. I think if we remove it the spout will poll for more tuples even if there are still pending records in the `waitingToEmit` list.

HeartSaVioR:
@srdo `waitingToEmit()` is also called from `poll()` to determine whether it should poll or not. Got it. Thanks for letting me know. :)

hmcl:
@srdo @HeartSaVioR thanks for the review. Can you please take another look. Thanks.

HeartSaVioR:
@hmcl 
I'm still +1 and I think @srdo agreed to make change. Let's merge in.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -242,9 +242,7 @@ public void nextTuple() {
                     }
                 }
 
-                if (waitingToEmit()) {
-                    emit();
-                }
+                emitIfWaitingNotEmitted();
```
srdo: If you're removing the waitingToEmit check, I think you should check for null in emitIfWaitingNotEmitted.

HeartSaVioR:
Sorry I didn't recognize the code has changed. I'll revoke my +1 and take a look. Sorry about that.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
HeartSaVioR: +1 Nice finding.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -372,19 +370,23 @@ private void ackRetriableOffsetsIfCompactedAway(Map<TopicPartition, Long> earlie
...
             waitingToEmit.remove();
+            if (emitted) {
+                break;
+            }
         }
     }
 
     /**
-     * Creates a tuple from the kafka record and emits it if it was not yet emitted.
+     * Creates a tuple from the kafka record and emits it if it was never emitted or is ready to be retried
```
HeartSaVioR: nit: If my memory is right, checksytle will check the ending '.' in javadoc. It is missing.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java

srdo:
+1

arunmahadevan:
@HeartSaVioR , can we look at merging this if your comments have been addressed ?

hmcl:
@arunmahadevan @HeartSaVioR @srdo this patch has been merged into master and 1.x-branch

PR #2440, opened 2017-11-30
STORM-2784: storm-kafka-client KafkaTupleListener method onPartitionsReassigned() should be called after initialization is complete

hmcl:
Somehow this change got reverted during a merge in 1.x-branch. It is fine in master, so we should just merge this patch here.


srdo:
+1, nice catch

HeartSaVioR:
Thanks @hmcl I merged this into 1.x-branch again.

PR #2442, opened 2017-11-30
STORM-2837: ConstraintSolverStrategy

revans2:


HeartSaVioR:
I have been feeling that many schedulers co-exist today, and some schedulers occupy config keys which don't contain scheduler name or so to determine the usage, and so as newly added configurations.

Given that there're lots of efforts on RAS, I would like to see which schedulers are also covered via RAS, and if it covers most of schedulers existing today, I think we can change the default scheduler to RAS.

If we don't want to make RAS default yet, the configuration keys should contain the mark for determining RAS, or at least documented in javadoc.

jerrypeng:
cool this got open sourced!

revans2:
@HeartSaVioR sorry I have been out for the past 2 weeks.  I will update the configs accordingly.

HeartSaVioR:
@revans2 No problem. Please take your time.

revans2:
@HeartSaVioR I rebased it and made the changes.  My updates are in the last commit f5e9532

HeartSaVioR:
@revans2 Thanks for the quick fix. I can see you've addressed my comment. Btw, I didn't see actual patch so would take time to review the code.

@jerrypeng I think you've got context of the codebase, then could you help reviewing the change? Thanks in advance.

storm-server/src/main/java/org/apache/storm/scheduler/multitenant/NodePool.java
```diff
@@ -98,7 +98,7 @@ public RoundRobinSlotScheduler(TopologyDetails td, int slotsToUse,
       }
       
       _spreadToSchedule = new HashMap<>();
-      List<String> spreadComps = (List<String>)td.getConf().get(Config.TOPOLOGY_SPREAD_COMPONENTS);
+      List<String> spreadComps = (List<String>)td.getConf().get(Config.TOPOLOGY_RAS_CONSTRAINT_SPREAD_COMPONENTS);
```
jerrypeng: I don't know about changing TOPOLOGY_SPREAD_COMPONENTS to TOPOLOGY_RAS_CONSTRAINT_SPREAD_COMPONENTS.  It seems kind of weird to me that the multitenant scheduler would have a config that references RAS.

storm-server/src/main/java/org/apache/storm/scheduler/resource/strategies/scheduling/ConstraintSolverStrategy.java
```diff
@@ -0,0 +1,623 @@
...
+                matrix.get(comp).put(comp2, 0);
+            }
+        }
+        List<List<String>> constraints = (List<List<String>>) topo.getConf().get(Config.TOPOLOGY_CONSTRAINTS);
+        if (constraints != null) {
+            for (List<String> constraintPair : constraints) {
+                String comp1 = constraintPair.get(0);
+                String comp2 = constraintPair.get(1);
+                if (!matrix.containsKey(comp1)) {
+                    LOG.warn("Comp: {} declared in constraints is not valid!", comp1);
```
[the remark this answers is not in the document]
jerrypeng: This pretty much means that the user specified (probably unintentionally) a component that doesn't exist.  In my experience users probably won't check the logs if there is not something obviously wrong with their topology so they will probably miss this error.   We should probably display some kind of warning for the topology status or even better if we can have client side checking to validate that all  components in Config.TOPOLOGY_RAS_CONSTRAINTS are valid components of the topology

storm-server/src/main/java/org/apache/storm/scheduler/resource/strategies/scheduling/ConstraintSolverStrategy.java
```diff
@@ -0,0 +1,623 @@
...
+    private static HashSet<String> getSpreadComps(TopologyDetails topo) {
+        HashSet<String> retSet = new HashSet<>();
+        List<String> spread = (List<String>) topo.getConf().get(Config.TOPOLOGY_SPREAD_COMPONENTS);
+        if (spread != null) {
+            Set<String> comps = topo.getComponents().keySet();
+            for (String comp : spread) {
+                if (comps.contains(comp)) {
+                    retSet.add(comp);
+                } else {
+                    LOG.warn("Comp {} declared for spread not valid", comp);
```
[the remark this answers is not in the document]
jerrypeng: Similar situation as TOPOLOGY_RAS_CONSTRAINTS.  I would give the same recommendation.

storm-server/src/main/java/org/apache/storm/scheduler/resource/strategies/scheduling/GenericResourceAwareStrategy.java
```diff
@@ -129,7 +129,15 @@ public SchedulingResult schedule(Cluster cluster, TopologyDetails td) {
     protected TreeSet<ObjectResources> sortObjectResources(
             final AllResources allResources, ExecutorDetails exec, TopologyDetails topologyDetails,
             final ExistingScheduleFunc existingScheduleFunc) {
+        return sortObjectResourcesImpl(allResources, exec, topologyDetails, existingScheduleFunc);
+    }
 
+    /**
+     * Implelemtation of the sortObjectResources method so other strategies can reuse it.
+     */
+    public static TreeSet<ObjectResources> sortObjectResourcesImpl(
```
jerrypeng: I would suggest putting this in some sort of Utility class.  Its kind of awkward for the ConstraintSovlerStrategy to call a static method in the GenericResourceAwareStrategy

jerrypeng:
Generally look good just some minor comments.  @revans2 you can also implement a version of the algorithm using heap space in the JVM instead of stack stack space so you wouldn't need to tinker with the -Xss settings.  May simplify operations.

jerrypeng:
In terms of the configs, we may want to explore splitting the Configs into multiple pieces.  Config.java has become a monolith in a sense. Configs.java in the storm-client module should probably only contain configs users can set in their topology.

storm-server/src/main/java/org/apache/storm/scheduler/multitenant/NodePool.java
revans2: good point I forgot that I moved that in config instead of adding it.

revans2:
@jerrypeng I reverted the one rename you asked about and rebased to fix a merge conflict.  I also squashed because I thought we were really close on this.

I don't understand your comment about using heap space instead of stack space with the -Xss setting.  We don't have Xss anywhere in the code base.  If you could point out the code that has issues I am happy to look at it.

As for Configs.java we already support that.  We use a service loader for org.apache.storm.validation.Validated to enable config validation. That is how we made org.apache.storm.DaemonConfig and org.apache.storm.hdfs.spout.Configs work.  If you really want me to try and clean things up for 2.x I can move some of the configs into separate classes, but that might be better as a separate jira.

jerrypeng:
@revans2 Sorry for not being more clear about what I meant in regards to using the heap instead of stack space.  Currently, the algorithm uses recursion to backtrack which means it is going to use space on the call stack.  This can cause a stackoverflow if the stack size is not large enough to handle TOPOLOGY_RAS_CONSTRAINT_MAX_STATE_SEARCH. Thus, you might need to configure the jvm stack space (-Xss) in Nimbus to appropriately match what is expected to be set for TOPOLOGY_RAS_CONSTRAINT_MAX_STATE_SEARCH.  What I am suggesting is that if there is an iterative implementation of the recursive algorithm, we would not need to put some many frames on the stack but in heap so that you don't need to configure the stack space for the Nimbus Daemon.  Not something critical, but rather something to think about.

jerrypeng:
LGTM +1. @revans2 up to you if you want to do something more about whether those warnings should dealt with in a more serious manner

revans2:
@jerrypeng TOPOLOGY_RAS_CONSTRAINT_MAX_STATE_SEARCH is not an effective limit on the max stack depth.  The max stack depth is the number of bolt and spout instances in a topology.  If we want to limit the depth then we need to have a separate config that limits the maximum number of components that can be scheduled, limiting the number of states searched is more around not spending too much time/CPU on something that is not likely to happen.

I agree that having an iterative algorithm would be best, but the search is not tail recursive, so making the switch was not that simple and I thought it would be better to just get something out.  I did place some of the stack variables in SearcherState instead of on the call stack.  I could move more there, but the code would be rather ugly and not use as much of the nice java syntactic sugar.  We have been running with something like this in production for a while and have not had to set the stack size at all to make this work.

jerrypeng:
@revans2 thanks for the explanantion

PR #2445, opened 2017-12-05
STORM-2843: [Flux] properties file not found when loading resources...

vesense:
…from classpath

https://issues.apache.org/jira/browse/STORM-2843

### DIFF
--- a/storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
+++ b/storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
@@ -471,6 +471,12 @@ public class Nimbus implements Iface, Shutdownable, DaemonCommon {
         return ret;
     }
 
+    private static IScheduler wrapAsBlacklistScheduler(Map<String, Object> conf, IScheduler scheduler) {
+        BlacklistScheduler blacklistWrappedScheduler = new BlacklistScheduler(scheduler);
+        blacklistWrappedScheduler.prepare(conf);
+        return blacklistWrappedScheduler;
+    }
+
     private static IScheduler makeScheduler(Map<String, Object> conf, INimbus inimbus) {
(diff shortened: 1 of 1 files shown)


flux/flux-core/src/main/java/org/apache/storm/flux/parser/FluxParser.java
```diff
@@ -39,51 +39,57 @@
...
     public static TopologyDef parseResource(String resource, boolean dumpYaml, boolean processIncludes,
     	String propertiesFile, boolean envSub) throws IOException {
         
         InputStream in = FluxParser.class.getResourceAsStream(resource);
-        TopologyDef topology = parseInputStream(in, dumpYaml, processIncludes, propertiesFile, envSub);
+        InputStream propertiesIn = null;
+        if (propertiesFile != null) {
+            propertiesIn = FluxParser.class.getResourceAsStream(propertiesFile);
+        }
+        TopologyDef topology = parseInputStream(in, dumpYaml, processIncludes, propertiesIn, propertiesFile, envSub);
```
HeartSaVioR: There's a corner case here:

1. `--resource` option is provided hence this method called.
2. yaml file has an include statement which sets resource as `false`.
3. `parseFile` is called which doesn't handle propertiesFile as resource.

So it needs more fine-grained approach regarding dealing with propertiesFile.

If my understanding is right, given that we only read and don't modify properties file after loading it, we could just create Properties instance and pass it instead of passing properties file name. Makes sense?

vesense: Make sense. Yaml file and included yaml file should share the same properties instance. PR for master and 1.x are updated. Please take a look again.

vesense:
@HeartSaVioR PR updated. Could you take a look again?

vesense:
@HeartSaVioR Created new PR for this.

PR #2451, opened 2017-12-08
STORM-2850: Make ManualPartitionSubscription call rebalance listener…

srdo:
… on revoke hook before assigning new partitions to the consumer

See https://issues.apache.org/jira/browse/STORM-2850.


hmcl:
+1

PR #2452, opened 2017-12-10
STORM-2850 1.x

srdo:
https://github.com/apache/storm/pull/2451

Will squash before merge. Only changes are replacing Mockito 2 with Mockito 1 in the test.


srdo:
Squashed and merged. Sorry, the ManualPartitionSubscription class has moved between master and 1.x, so I had to redo the fix after the cherry-pick, which is why the fix was in the second commit.

PR #2453, opened 2017-12-10
STORM-2851: Fix ConcurrentModificationException in doSeekRetriablePar…

srdo:
…titions

See https://issues.apache.org/jira/browse/STORM-2851.

The bug is caused by the call to `retriableTopicPartitions.remove(tp)` in doSeekRetriablePartitions. Rather than just rewriting it to use an iterator I thought it was better to get rid of the loop. It was meant to ensure that we didn't seek the consumer on partitions that would be paused in the current call to nextTuple. I've moved this filtering into PollingInfo, and have made poll() return it. This also allows us to only get earliestRetriableOffsets once for the call to nextTuple, rather than once in poll and once in doSeek.

The added test is not related to the fix, it just adds more verification that the maxUncommittedOffsets limit is enforced.

### DIFF
--- a/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
+++ b/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
@@ -34,6 +34,7 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
 import org.apache.commons.lang.Validate;
 import org.apache.kafka.clients.consumer.ConsumerRebalanceListener;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
@@ -229,14 +230,14 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
                     kafkaSpoutConfig.getSubscription().refreshAssignment();
                 }
 
-                if (commit()) {
+                if (shouldCommit()) {
                     commitOffsetsForAckedTuples();
                 }
 
-                Set<TopicPartition> pollablePartitions = poll();
-                if (!pollablePartitions.isEmpty()) {
+                PollablePartitionsInfo pollablePartitionsInfo = getPollablePartitionsInfo();
+                if (pollablePartitionsInfo.shouldPoll()) {
                     try {
-                        setWaitingToEmit(pollKafkaBroker(pollablePartitions));
+                        setWaitingToEmit(pollKafkaBroker(pollablePartitionsInfo));
                     } catch (RetriableException e) {
                         LOG.error("Failed to poll from kafka.", e);
                     }
@@ -257,23 +258,24 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
         throw new RuntimeException(new InterruptedException("Kafka consumer was interrupted"));
     }
 
-    private boolean commit() {
+    private boolean shouldCommit() {
         return isAtLeastOnceProcessing() && commitTimer.isExpiredResetOnTrue();    // timer != null for non auto commit mode
     }
-
-    private Set<TopicPartition> poll() {
-        final int maxUncommittedOffsets = kafkaSpoutConfig.getMaxUncommittedOffsets();
-
+    
+    private PollablePartitionsInfo getPollablePartitionsInfo() {
         if (isWaitingToEmit()) {
             LOG.debug("Not polling. Tuples waiting to be emitted.");
-            return Collections.emptySet();
+            return new PollablePartitionsInfo(Collections.emptySet(), Collections.emptyMap());
         }
+        
         Set<TopicPartition> assignment = kafkaConsumer.assignment();
         if (!isAtLeastOnceProcessing()) {
-            return assignment;
+            return new PollablePartitionsInfo(assignment, Collections.emptyMap());
         }
+        
         Map<TopicPartition, Long> earliestRetriableOffsets = retryService.earliestRetriableOffsets();
         Set<TopicPartition> pollablePartitions = new HashSet<>();
+        final int maxUncommittedOffsets = kafkaSpoutConfig.getMaxUncommittedOffsets();
         for (TopicPartition tp : assignment) {
             OffsetManager offsetManager = offsetManagers.get(tp);
             int numUncommittedOffsets = offsetManager.getNumUncommittedOffsets();
@@ -292,7 +294,7 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
                 }
             }
         }
-        return pollablePartitions;
+        return new PollablePartitionsInfo(pollablePartitions, earliestRetriableOffsets);
     }
 
     private boolean isWaitingToEmit() {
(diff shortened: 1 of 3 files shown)


external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -625,4 +619,15 @@ public String toString() {
...
         return kafkaSpoutConfig.getSubscription().getTopicsString();
     }
+    
+    private static class PollingInfo {
+        private final Set<TopicPartition> pollablePartitions;
+        private final Map<TopicPartition, Long> pollableEarliestRetriableOffsets;
+        
+        public PollingInfo(Set<TopicPartition> pollablePartitions, Map<TopicPartition, Long> earliestRetriableOffsets) {
+            this.pollablePartitions = pollablePartitions;
+            earliestRetriableOffsets.keySet().removeIf(tp -> !pollablePartitions.contains(tp));
```
HeartSaVioR: minor: might be better to copy the map first, and remove the elements to avoid side-effect to the parameter.

srdo:
@HeartSaVioR Thanks, I agree about the map copy. I've made the change you suggested, and also very slightly expanded a test to check that we only seek on partitions that are both pollable and retriable when retrying tuples.

I'll squash before merging if it looks good to you.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
hmcl: agree

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -625,4 +619,15 @@ public String toString() {
     private String getTopicsString() {
         return kafkaSpoutConfig.getSubscription().getTopicsString();
     }
+    
+    private static class PollingInfo {
+        private final Set<TopicPartition> pollablePartitions;
+        private final Map<TopicPartition, Long> pollableEarliestRetriableOffsets;
```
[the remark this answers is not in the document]
hmcl: OK, the name is fine after understanding it better. Can you put a line comment saying what is in this var. Thanks.

srdo:
@hmcl Thanks, made most of the changes you mentioned. I thought PollInfo was a little vague, how about PollablePartitionInfo?

Also the tests are failing for some reason, give me a minute.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -257,23 +258,24 @@ private void throwKafkaConsumerInterruptedException() {
...
-    private boolean commit() {
+    private boolean isCommitAllowed() {
         return isAtLeastOnceProcessing() && commitTimer.isExpiredResetOnTrue();    // timer != null for non auto commit mode
     }
-
-    private Set<TopicPartition> poll() {
-        final int maxUncommittedOffsets = kafkaSpoutConfig.getMaxUncommittedOffsets();
-
+    
+    private PollablePartitionInfo getPollablePartitionInfo() {
```
hmcl: getPollablePartitions

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -625,4 +620,21 @@ public String toString() {
...
+        private final Map<TopicPartition, Long> pollableEarliestRetriableOffsets;
+        
+        public PollablePartitionInfo(Set<TopicPartition> pollablePartitions, Map<TopicPartition, Long> earliestRetriableOffsets) {
+            this.pollablePartitions = pollablePartitions;
+            this.pollableEarliestRetriableOffsets = earliestRetriableOffsets.entrySet().stream()
+                .filter(entry -> pollablePartitions.contains(entry.getKey()))
+                .collect(Collectors.toMap(entry -> entry.getKey(), entry -> entry.getValue()));
+        }
+        
+        public boolean isPollAllowed() {
```
hmcl: why not simply `isEmpty()` ? 

I really don't thing that _allowed_ is a very suitable name here because we are not checking for permissions or anything like that. If it is in the pollable set it is always allowed.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -230,14 +230,14 @@ public void nextTuple() {
                     kafkaSpoutConfig.getSubscription().refreshAssignment();
                 }
 
-                if (commit()) {
+                if (isCommitAllowed()) {
```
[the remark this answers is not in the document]
srdo: How about shouldCommit()?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
srdo: I agree that allowed can be misunderstood. I'd prefer not calling it isEmpty, since this isn't a collection and it's not obvious which internal collection it would be referring to. How about shouldPoll?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
srdo: Will rename

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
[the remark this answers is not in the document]
hmcl: That would work. However, I would say that most conventions suggest that booleans methods should be of the for isPredicate. Of course there are variants such as hasPredicate, shouldPredicate, mustPredicate. 

When I suggested is* names above was with the intent of saying exactly should commit, but starting with is. Up to you what to pick, and I am +1 after that.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
hmcl: Building on the comment above, I would still call it something along the lines `isReadyToPoll()` or `isPollableRetryReady`, because that's what it is, a retry that is ready to poll. However, it's up to you and I am +1 after this.

srdo:
Updated with renames. I went with should* because I think the is* convention gets a little awkward here.

hmcl:
+1. Can you please squash such that I can merge it in. Thanks.

srdo:
Squashed. I'll merge it in a few minutes, and put up a 1.x version soon.

srdo:
Postponing merge a bit until Travis works again.

hmcl:
@srdo do you know when you will get the chance to create the PR for 1.x? Not to rush, but I was hoping we could start converging towards getting 1.2.0 out. Thanks.

srdo:
@hmcl I'll put up a 1.x version later today. It would be good to get 1.2.0 out soon, but my impression is that we're still waiting for metrics v2, as well as a few other storm-kafka-client fixes (https://github.com/apache/storm/pull/2454 and https://issues.apache.org/jira/browse/STORM-2844).

srdo:
Thanks for the reviews.

PR #2455, opened 2017-12-13
Upgrade Travis ruby to latest teeny release to fix Travis build

srdo:
It seems like the Travis build is failing because the Trusty image has been updated. See for example https://travis-ci.org/apache/storm/jobs/315946342.

Upgrading ruby to the latest teeny version of the current line seems to fix it. According to https://www.ruby-lang.org/en/news/2013/12/21/ruby-version-policy-changes-with-2-1-0/ this should be safe.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ language: java
 jdk:
   - oraclejdk8
 before_install:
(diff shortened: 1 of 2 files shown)


srdo:
Nevermind this, will reopen once I have a fix.

PR #2457, opened 2017-12-14
STORM-2854 Expose IEventLogger to make event logging pluggable

HeartSaVioR:
* expose option to register IEventLogger similar to metrics consumer
* change the interface of IEventLogger slightly
  * allow argument
  * the change is technically not backward compatible but in real we can treat it's OK
    * cause we don't provide a chance to implement custom IEventLogger and plug to topology
* Fix EventInfo to contain origin type of values instead of String-converted values
* open possibility to extend FileBasedEventLogger and provide different format of log message
* document the change

This is in reality backward compatible change (though it is 'technically' not backward compatible), and I'd like to get it in 1.x version as well.

For 1.x-branch: #2458

### DIFF
--- a/docs/Eventlogging.md
+++ b/docs/Eventlogging.md
@@ -68,7 +68,8 @@ Eventlogging works by sending the events (tuples) from each component to an inte
 
 
 ## Extending eventlogging
-Storm provides an `IEventLogger` interface which is used by the event logger bolt to log the events. The default implementation for this is a FileBasedEventLogger which logs the events to an events.log file ( `logs/workers-artifacts/<topology-id>/<worker-port>/events.log`). Alternate implementations of the `IEventLogger` interface can be added to extend the event logging functionality (say build a search index or log the events in a database etc)
+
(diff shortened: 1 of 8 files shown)


HeartSaVioR:
Tested manually with FileBasedEventLogger.

docs/Eventlogging.md
```diff
@@ -94,3 +95,29 @@ public interface IEventLogger {
...
 }
 ```
+
+The default implementation for this is a FileBasedEventLogger which logs the events to an events.log file ( `logs/workers-artifacts/<topology-id>/<worker-port>/events.log`).
+Alternate implementations of the `IEventLogger` interface can be added to extend the event logging functionality (say build a search index or log the events in a database etc)
+
+To register event logger to your topology, add to your topology's configuration like:
+
+```java
+conf.registerEventLogger(org.apache.storm.metric.FileBasedEventLogger.class);
```
arunmahadevan: nit: remove the metric package name reference

arunmahadevan:
+1

PR #2458, opened 2017-12-14
(1.x) STORM-2854 Expose IEventLogger to make event logging pluggable

HeartSaVioR:
* expose option to register IEventLogger similar to metrics consumer
* change the interface of IEventLogger slightly
  * allow argument
  * the change is technically not backward compatible but in real we can treat it's OK
    * cause we don't provide a chance to implement custom IEventLogger and plug to topology
* open possibility to extend FileBasedEventLogger and provide different format of log message
* document the change

For master branch: #2457

### DIFF
--- a/conf/storm.yaml.example
+++ b/conf/storm.yaml.example
@@ -73,3 +73,10 @@
 #       - endpoint: "metrics-collector.mycompany.org"
 #
 # storm.cluster.metrics.consumer.publish.interval.secs: 60
+
+# Event Logger
+# topology.event.logger.register:
+#   - class: "org.apache.storm.metric.FileBasedEventLogger"
+#   - class: "org.mycompany.MyEventLogger"
+#     arguments:
+#       endpoint: "event-logger.mycompany.org"
--- a/storm-core/src/jvm/org/apache/storm/metric/IEventLogger.java
+++ b/storm-core/src/jvm/org/apache/storm/metric/IEventLogger.java
@@ -20,6 +20,7 @@ package org.apache.storm.metric;
 import org.apache.storm.task.TopologyContext;
 
 import java.util.Date;
+import java.util.List;
 import java.util.Map;
 
 /**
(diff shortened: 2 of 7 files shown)


HeartSaVioR:
Tested manually with FileBasedEventLogger.

conf/storm.yaml.example
```diff
@@ -72,4 +72,11 @@
...
 #
-# storm.cluster.metrics.consumer.publish.interval.secs: 60
\ No newline at end of file
+# storm.cluster.metrics.consumer.publish.interval.secs: 60
+
+# Event Logger
+# topology.event.logger.register:
+#   - class: "org.apache.storm.metric.FileBasedEventLogger"
+#   - class: "org.mycompany.MyEventLogger"
+#     argument:
```
ptgoetz: Nit: May want to make this "arguments" if it is a list.

storm-core/src/jvm/org/apache/storm/metric/IEventLogger.java
```diff
@@ -31,32 +32,54 @@
...
+
         /**
          * Returns a default formatted string with fields separated by ","
          *
          * @return a default formatted string with fields separated by ","
          */
         @Override
         public String toString() {
-            return new Date(Long.parseLong(ts)).toString() + "," + component + "," + task + "," + messageId + "," + values;
+            return new Date(ts).toString() + "," + component + "," + String.valueOf(task) + ","
```
ptgoetz: Do we want to support configurable date formats or default to `Date.toString()`?

conf/storm.yaml.example
```diff
@@ -72,4 +72,11 @@
 #     argument:
 #       - endpoint: "metrics-collector.mycompany.org"
 #
-# storm.cluster.metrics.consumer.publish.interval.secs: 60
\ No newline at end of file
+# storm.cluster.metrics.consumer.publish.interval.secs: 60
+
+# Event Logger
+# topology.event.logger.register:
+#   - class: "org.apache.storm.metric.FileBasedEventLogger"
```
arunmahadevan: nit: remove the metric package name reference

arunmahadevan:
+1, once the minor nits are addressed.

conf/storm.yaml.example
HeartSaVioR: I'm following the name for metrics consumer, but I think it is worth to fix. Moreover we're deprecating metrics consumer so it looks OK. I'll fix it.

conf/storm.yaml.example
HeartSaVioR: Eventlogging classes are in org.apache.storm.metric package. I don't know why, but if we can't move them to another package, I think it should be this value since this requires full class name.

storm-core/src/jvm/org/apache/storm/metric/IEventLogger.java
HeartSaVioR: Instead of doing that, I just open possibility to extend FileBasedEventLogger easily. Users can simply extend FileBasedEventLogger and override `buildLogMessage(EventInfo)` to provide the log line.

ptgoetz: Fair enough. I think it would be helpful to document that.

HeartSaVioR: Yeah good suggestion. I'll address.

conf/storm.yaml.example
HeartSaVioR: Ah also it is a `Map<String,Object>`. I think I should remove `-`. Will remove.

HeartSaVioR:
@ptgoetz @arunmahadevan 
Addressed or left comments. Please take a look again.

arunmahadevan:
+1

PR #2462, opened 2017-12-15
STORM-2858: Fix worker-launcher build by erroring out if asprintf fai…

srdo:
…ls to allocate memory. Replace make-maven-plugin with a build shell script.

This is part of making the build run on the new Travis Ubuntu image. 

I got an error when building storm-core with -Pnative because GCC has marked asprintf as a function where you shouldn't ignore the return value. I've put in return value handling that exits the worker-launcher process if asprintf fails to allocate memory. As part of this I replaced make-maven-plugin, because it's fairly old (last release in 2009, there doesn't seem to be a page for the plugin anymore) and opaque (had to decompile the jar to figure out what it was doing). I think a build script + exec-maven-plugin is less magical. The make process output now also goes in the terminal when building, which it didn't with make-maven-plugin.

### DIFF
--- a/storm-core/pom.xml
+++ b/storm-core/pom.xml
@@ -31,6 +31,7 @@
 
     <properties>
         <worker-launcher.conf.dir>/etc/storm</worker-launcher.conf.dir>
+        <worker-launcher.build.dir>${project.build.directory}/native/worker-launcher</worker-launcher.build.dir>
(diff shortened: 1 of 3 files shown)


srdo:
Here's the run from a branch that contains this branch, STORM-2859 and some minor travis.yml changes. https://travis-ci.org/srdo/storm/builds/317153261

ptgoetz:
+1

PR #2463, opened 2017-12-15
STORM-2859: Fix a number of issues with NormalizedResources, and prev…

srdo:
…ent it from leaking static state in tests.

This is part of making the build run on the new Travis Ubuntu image. Travis is probably running the tests in a slightly different order than before.

NormalizedResources had a static field used to track the generic resource types registered for supervisors. It was leaking from the GenericResourceAwareSchedulerTest to DefaultResourceAwareSchedulerTest, where it was causing a test failure. When the unused resource type showed up in the test, NormalizedResources would return 0.0 from calculateMinPercentageUsedBy for all racks because there was a resource type where the cluster had 0 of that resource in total. This breaks rack sorting and causes the test to fail.

I'm not entirely sure why NormalizedResources used the static field + double arrays for tracking resources, but I've replaced it with regular resourceName -> resourceValue maps. Let me know if this is a problem.

Other issues fixed:
* calculateMinPercentageUsedBy threw division by 0 error if total cpu or memory was 0.
* Removed static map between resource names and array indices. It was causing tests to leak information to each other. Rewrote parts of NormalizedResources to make this map unnecessary.
* Added check that all resources in a rack/node are also present in the total.
* calculateAvg was being inconsistent about the divisor in the average between cpu/memory and other resources. Fixed so skipped resources are never counted in the divisor. 
* calculateMin didn't handle resource totals being 0 very well. The method defined (anyNumber)/0 to result in 0, so if any resource in the total had a 0 value, all racks/nodes would be prioritized equally. Change calculateMin and calculateAvg to simply skip any resources where the total is 0 in the calculation. I think it makes sense to disregard such resources and fall back on prioritizing by any remaining non-0 resources. It was already done for CPU and memory in calculateAvg. In case there are no non-0 resources, we can just return 100% for both average and minimum to disable prioritization, since any rack/node is as good as any other.
* Fixed up NormalizedResources constructor so it doesn't call an overridable method.

### DIFF
--- a/storm-server/src/main/java/org/apache/storm/blobstore/BlobStoreUtils.java
+++ b/storm-server/src/main/java/org/apache/storm/blobstore/BlobStoreUtils.java
@@ -196,10 +196,10 @@ public class BlobStoreUtils {
                 // Catching and logging KeyNotFoundException because, if
                 // there is a subsequent update and delete, the non-leader
                 // nimbodes might throw an exception.
-                LOG.info("KeyNotFoundException {}", knf);
+                LOG.info("KeyNotFoundException", knf);
             } catch (Exception exp) {
                 // Logging an exception while client is connecting
-                LOG.error("Exception {}", exp);
+                LOG.error("Exception", exp);
             }
         }
 
--- a/storm-server/src/main/java/org/apache/storm/scheduler/resource/NormalizedResourceOffer.java
+++ b/storm-server/src/main/java/org/apache/storm/scheduler/resource/NormalizedResourceOffer.java
@@ -34,27 +34,23 @@ public class NormalizedResourceOffer extends NormalizedResources {
     private double totalMemory;
 
     /**
-     * Create a new normalized set of resources.  Note that memory is not covered here becasue it is not consistent in requests vs offers
+     * Create a new normalized set of resources.  Note that memory is not covered here because it is not consistent in requests vs offers
      * because of how on heap vs off heap is used.
      *
      * @param resources the resources to be normalized.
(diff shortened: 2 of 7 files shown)


srdo:
There's also a TODO I'd like some input on https://github.com/apache/storm/pull/2463/files#diff-0f94945ef866045b8717f781620fd59dR190

srdo:
Here's the run from a branch that contains this branch, STORM-2858 and some minor travis.yml changes. https://travis-ci.org/srdo/storm/builds/317153261

revans2:
@srdo One of the issues I was trying to address with using the array was both performance and memory usage.  Maps are very slow and use lots of memory compared to an array.  Hadoop did a very similar thing as part of their generic resource work for similar reasons.  They tend to do a lot more computation on their resources looking at fairness etc.

We are fining that the performance of the scheduler loop in general is our biggest bottleneck to scaling our clusters larger.  It directly correlates to how responsive a cluster is to user requests.  So for cases where they want to kill a topology and launch a new one, if they cannot do it quickly because of scheduling they are not very happy.

srdo:
@revans2 That makes sense. How about we fix the issue by doing as you suggest in https://github.com/apache/storm/pull/2477, and I'll try to pick out the changes from this I think would still be relevant for a later PR?

revans2:
@srdo that sounds fine to me.

srdo:
Closing in favor of https://github.com/apache/storm/pull/2477, will figure out which parts of this are still relevant and reopen at some point.

PR #2464, opened 2017-12-16
STORM-2847 1.x

srdo:
1.x for https://github.com/apache/storm/pull/2454

PR #2465, opened 2017-12-16
STORM-2844: KafkaSpout Throws IllegalStateException After Committing to Kafka When First Poll Strategy Set to EARLIEST

hmcl:
- Serialize KafkaSpoutMessageId to JSON and add it as part of OffsetAndMetadata
 - Check if commits to Kafka were committed by this topology to correctly enforce FirstPollOffsetStrategy


hmcl:
@srdo can you please do a first pass review. In the meantime I will address the 2 or 3 TODO with minor cleanups. The following tests are also failing. I am working on fixing them, but if you know from the top of your head how to fix these failures (I don't think they are bugs added by the patch), I would appreciate your tips. Thanks.

``` java
Tests in error:
  KafkaSpoutRebalanceTest.spoutMustIgnoreFailsForTuplesItIsNotAssignedAfterRebalance:184 InvalidUseOfMatchers
  KafkaSpoutRebalanceTest.testReassignPartitionSeeksForOnlyNewPartitions:244 » Runtime
  MaxUncommittedOffsetTest.testNextTupleCanEmitMoreMessagesWhenDroppingBelowMaxUncommittedOffsetsDueToCommit:120 » NullPointer
  MaxUncommittedOffsetTest.testNextTupleWillNotEmitMoreThanMaxUncommittedOffsetsPlusMaxPollRecordsMessages:187->failAllExceptTheFirstMessageThenCommit:170 » NullPointer
```

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -225,6 +237,23 @@ private long doSeek(TopicPartition tp, OffsetAndMetadata committedOffset) {
...
+    /**
+     * Checks If {@link OffsetAndMetadata} was committed by this topology, either by this or another spout instance.
+     * This info is used to decide if {@link FirstPollOffsetStrategy} should be applied
+     *
+     * @param committedOffset {@link OffsetAndMetadata} info committed to Kafka
+     * @return true if this topology committed this {@link OffsetAndMetadata}, false otherwise
+     */
+    private boolean isOffsetCommittedByThisTopology(OffsetAndMetadata committedOffset) {
+        try {
+            return committedOffset != null && JSON_MAPPER.readValue(committedOffset.metadata(), KafkaSpoutMessageId.class)
```
srdo: Nit: Move the null check outside this method, it's already checked before using this function at one call site, and in the other I don't think it makes the code more readable that it's moved in here. Also consider putting the message id in a variable before doing the .getTopologyId, that way it's easier to tell what's happening in case we get an NPE from this method.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -225,6 +237,23 @@ private long doSeek(TopicPartition tp, OffsetAndMetadata committedOffset) {
         }
     }
 
+    /**
+     * Checks If {@link OffsetAndMetadata} was committed by this topology, either by this or another spout instance.
```
srdo: Nit: Something like "Checks if OffsetAndMetadata was committed by a KafkaSpout in this topology." might be clearer.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
srdo: Also I think this will crash the spout when starting up on a topic that already has commits from before this change. We should probably warn in the log and pick some reasonable default instead of erroring out if there's no message id in the metadata.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -399,14 +429,16 @@ private void emitIfWaitingNotEmitted() {
...
 
         if (offsetManagers.containsKey(tp) && offsetManagers.get(tp).contains(msgId)) {   // has been acked
             LOG.trace("Tuple for record [{}] has already been acked. Skipping", record);
         } else if (emitted.contains(msgId)) {   // has been emitted and it is pending ack or fail
             LOG.trace("Tuple for record [{}] has already been emitted. Skipping", record);
         } else {
-            if (kafkaConsumer.committed(tp) != null && (kafkaConsumer.committed(tp).offset() >= kafkaConsumer.position(tp))) {
+            final OffsetAndMetadata committedOffset = kafkaConsumer.committed(tp);
+            if (isOffsetCommittedByThisTopology(committedOffset) && committedOffset.offset() > kafkaConsumer.position(tp)) {
+                // this check should pass if commit was done by another topology such that FirstPollOffsetStrategy == EARLIEST is honored (STORM-2844)
```
srdo: Nit: Describe when this check is valid instead. e.g. "It should never be possible for the consumer position to fall behind the offset committed by this topology"

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutRetryExponentialBackoff.java
```diff
@@ -290,8 +291,8 @@ public int readyMessageCount() {
     }
 
     @Override
-    public KafkaSpoutMessageId getMessageId(ConsumerRecord<?, ?> record) {
-        KafkaSpoutMessageId msgId = new KafkaSpoutMessageId(record);
+    public KafkaSpoutMessageId getMessageId(ConsumerRecord<?, ?> record, TopologyContext context) {
```
srdo: I'm fine with this in 2.x, but for 1.2.0 I think we should try to avoid breaking changes (RetryService is a public API, it's the same reason this method still takes a ConsumerRecord instead of just the partition and offset).

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutMessageId.java
```diff
@@ -18,36 +18,88 @@
...
+                                @JsonProperty("thread") String thread,
+                                @JsonProperty("topologyId") String topologyId,
+                                @JsonProperty("numFails") int numFails) {
+
         this.topicPart = topicPart;
         this.offset = offset;
         this.emitted = emitted;
+        this.thread = thread;
+        this.topologyId = topologyId;
+        this.numFails = numFails;
```
srdo: We don't need to store/restore numFails and emitted in metadata.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutMessageId.java
```diff
@@ -18,36 +18,88 @@
 
 package org.apache.storm.kafka.spout;
 
+import java.io.IOException;
```
srdo: I think maybe we should split this object. We don't need half of the fields for metadata, and several fields aren't needed for emit/ack handling either. Maybe have an object with thread name and topology id for metadata, and have KafkaSpoutMessageId only contain the data we need for ack handling? That way we can also avoid doing stuff like storing the topic partition and offset in metadata, which IMO makes no sense.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
srdo: Also we might want to add notes to EARLIEST and LATEST to clarify that they only restart when the topology is redeployed.

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/KafkaSpoutRebalanceTest.java
```diff
@@ -181,16 +181,16 @@ public void spoutMustIgnoreFailsForTuplesItIsNotAssignedAfterRebalance() throws
...
+            .thenReturn(new KafkaSpoutMessageId(partitionThatWillBeRevoked, 0, contextMock))
+            .thenReturn(new KafkaSpoutMessageId(assignedPartition, 0, contextMock));
 
         //Emit a message on each partition and revoke the first partition
         List<KafkaSpoutMessageId> emittedMessageIds = emitOneMessagePerPartitionThenRevokeOnePartition(
             spout, partitionThatWillBeRevoked, assignedPartition, rebalanceListenerCapture);
 
         //Check that only two message ids were generated
-        verify(retryServiceMock, times(2)).getMessageId(Mockito.any(ConsumerRecord.class));
+        verify(retryServiceMock, times(2)).getMessageId(Mockito.any(ConsumerRecord.class), contextMock);
```
srdo: You need to put eq around contextmock here or you'll get an error

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
hmcl: Agree. Up to this point this was a bug. I guess we should default to the previous behavior before this fix, both in 1.x and 2.x. What do you think?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutRetryExponentialBackoff.java
hmcl: I guess the only non-backwards compatible changes are in case the user implemented their own version of KafkaSpoutRetryService. Right? I will check if it is possible to make this fix without this change. If possible I will do.

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/KafkaSpoutRebalanceTest.java
hmcl: OK. I think I tried that but it didn't quite work. I will address the tests after I make the changes you suggest. I will keep you posted on how it goes.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
hmcl: You mean EARLIEST, LATEST only apply when a topology (with a given id) is started for the first time, and not when it is activated/deactivated?

hmcl:
@srdo I plan on submitting a patch to master, but why does the patch need to go on master first? For example, there could be bugs in 1.x that no longer exist in master because of other changes, and in that case the change will only go in 1.x.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
srdo: Sort of. I believe the storm id will be new every time you deploy a topology. According to the javadoc for getStormId it's the topology name + a nonce. So EARLIEST and LATEST will apply when you deploy the topology, but not when activating/deactivating the topology, and it won't apply if the spout crashes either.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
srdo: I'd probably just return false from this method. Since the topology id is new for each deployment we know for certain that if there's no message id in the meta, then it wasn't committed by this topology.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutRetryExponentialBackoff.java
srdo: I think if we split KafkaSpoutMessageId we won't need the context here, so hopefully it won't be a problem to fix.

srdo:
@hmcl I don't know that there's a hard rule that it has to be that way, but it seems like a good way to me to ensure that master is the "most fixed"/best state of the code. It's been done this way in the other PRs I've seen.

This doesn't apply to patches that fix issues that don't exist in later branches, but this issue isn't one of those.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
hmcl: We said the same thing and that is the intended behavior

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
hmcl: That's what I did

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutMessageId.java
hmcl: Done

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutRetryExponentialBackoff.java
hmcl: Done

hmcl:
@srdo I created the PR here because some topologies with latest master were not working for me and I worked off 1.x-branch to be able to make progress. I will submit a PR for master as well.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
srdo: Great. We might still want to expand the javadoc for the constants in KafkaSpoutConfig, because right now it doesn't say anything about when the setting has an effect, just that it applies to "the first poll".

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
srdo: I'm a little unsure what you mean. Won't this currently throw an IOException (or NPE) when you start from a committed offset that doesn't have a KafkaSpoutMessageId in the metadata?

srdo:
@hmcl Great, thanks.

hmcl:
@srdo I am working on making unit tests pass and then will squash and create master PR. Can you take another look in the meantime. Thanks.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -636,4 +675,45 @@ public String toString() {
     private String getTopicsString() {
         return kafkaSpoutConfig.getSubscription().getTopicsString();
     }
+
+    public static class Info {
```
srdo: Nit: Put this in the internal package since it's used by multiple classes. Maybe call it CommitMetadata or something like that.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -636,4 +675,45 @@ public String toString() {
...
+        private final String topologyId;
+        private final int taskId;
+        private final String thread;
```
srdo: Nit: threadName

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -636,4 +675,45 @@ public String toString() {
...
+        public Info(@JsonProperty("topologyId") String topologyId,
+                    @JsonProperty("taskId") int taskId,
+                    @JsonProperty("thread") String thread) {
+
+            this.topologyId = topologyId;
+            this.taskId = taskId;
+            this.thread = thread;
+        }
+
+        public Info(Thread thread, TopologyContext context) {
```
srdo: Unless we're creating Infos in a lot of places, it might be better to remove this and use the other constructor instead. It makes it more obvious what's happening at the call site.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutConfig.java
```diff
@@ -134,7 +134,12 @@ public KafkaSpoutConfig(Builder<K, V> builder) {
         EARLIEST,
         LATEST,
         UNCOMMITTED_EARLIEST,
-        UNCOMMITTED_LATEST
+        UNCOMMITTED_LATEST;
+
+        @Override
+        public String toString() {
+            return "FirstPollOffsetStrategy{" + super.toString() + "}";
```
srdo: This seems a little unnecessary, enums already have a good toString implementation?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/OffsetManager.java
```diff
@@ -140,10 +140,16 @@ public OffsetAndMetadata findNextCommitOffset() {
 
         OffsetAndMetadata nextCommitOffsetAndMetadata = null;
         if (found) {
-            nextCommitOffsetAndMetadata = new OffsetAndMetadata(nextCommitOffset,
-                nextCommitMsg.getMetadata(Thread.currentThread()));
+            try {
+                nextCommitOffsetAndMetadata = new OffsetAndMetadata(nextCommitOffset,
+                    JSON_MAPPER.writeValueAsString(new KafkaSpout.Info(Thread.currentThread(), context)));
```
srdo: Consider moving the JSON serialization into KafkaSpout commitOffsets method instead, then you can pass it as a parameter to this method and we can get away with having only one ObjectMapper. Currently we can just create the info once (e.g. in open) and reuse rather than creating and serializing a new one for each commit and partition.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -225,6 +243,25 @@ private long doSeek(TopicPartition tp, OffsetAndMetadata committedOffset) {
...
+     *
+     * @param committedOffset {@link OffsetAndMetadata} info committed to Kafka
+     * @return true if this topology committed this {@link OffsetAndMetadata}, false otherwise
+     */
+    private boolean isOffsetCommittedByThisTopology(OffsetAndMetadata committedOffset) {
+        try {
+            final KafkaSpout.Info info = JSON_MAPPER.readValue(committedOffset.metadata(), KafkaSpout.Info.class);
+            return info.getTopologyId().equals(context.getStormId());
+        } catch (IOException e) {
+            LOG.trace("Failed to deserialize {}. Error likely occurred because the last commit " +
```
srdo: I don't think anyone is going to see this log if it's left at trace level.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -225,6 +243,25 @@ private long doSeek(TopicPartition tp, OffsetAndMetadata committedOffset) {
...
+                "for this topic-partition was done using an earlier version of Storm. " +
+                "Default behavior compatible with earlier version", committedOffset, e);
```
srdo: Nit: Defaulting to behavior compatible with earlier version

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
hmcl: If we put it at WARN level, it will write this message potentially thousands of times (equal to the number of records that get polled), until the first commit is done. I put it at TRACE because this is technically for debugging purposes, and I am printing the exception stack trace as well. I can put it at DEBUG, but I don't think it should be WARN. I can put the message at  DEBUG and stack trade at TRACE. 

On a side note, I don't think TRACE level is used enough, but I consider it is very useful to print more detail after the problem is identified.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/OffsetManager.java
hmcl: I think that approach is less efficient. Currently we only create and serialize the commit metadata when there are offsets ready to commit. Your proposed solution will always serialize, regardless if there are offsets ready to commit or not. Furthermore, ObjectMapper is a public static field, which means that per JVM we will be creating only two of them, on for the spout, and another for OffsetManager. We can even use the same instance on both classes if we want to.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutConfig.java
hmcl: This toString() makes logging cleaner and more consistent. if you do log.debug("using {}",firstPollOffsetStrategy) it will print  "using FirstPollOffsetStrategy{EARLIEST}", which follows the pattern of every other toString() method. If we don't override toString() it will simply print "using EARLIEST".

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutConfig.java
srdo: Makes sense.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/OffsetManager.java
srdo: I think maybe I was being a little unclear. I'm saying that right now the information contained in the Info object is constant for the lifetime of the spout. We should be able to just create one Info object in KafkaSpout.open and reuse it. This means a single serialization for the lifetime of the spout, which is more efficient than doing it on every commit.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutConfig.java
hmcl:
```
Seeking offset for topic-partition a1p-0 with FirstPollOffsetStrategy{UNCOMMITTED_EARLIEST} and OffsetAndMetadata{offset=3, metadata='{"topologyId":"tpn-1-1513534208","taskId":3,"threadName":"Thread-20-ks-executor[3 3]"}'}
```

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
srdo: Good point, hadn't thought of this code being hit on every emit. The reason I think WARN level is appropriate is that this is indicating a potential error, and we want it to show up in the log if an exception is caught here. Putting it at TRACE level means no one will see it. It's not solely for debugging purposes, it's informational telling the user that the spout is defaulting to a certain behavior because a feature failed to work. TRACE and DEBUG are IMO for logs that happen "normally" and not special cases like this.

I'm wondering if we should cache the committed offset and metadata in a field between commits instead, i.e. add a partition -> lastCommitedOffsetAndMetadata map that we update in the commit method? It seems a little inefficient to run this code on every single emit, in addition to the issue with lots of logging.

hmcl:
@srdo once I get +1 I will create the PR for master.

srdo:
@hmcl There were a few more comments, just making sure they weren't missed.

https://github.com/apache/storm/pull/2465#discussion_r157350384

https://github.com/apache/storm/pull/2465#discussion_r157375639

https://github.com/apache/storm/pull/2465#discussion_r157375809

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
hmcl: The log messages only get printed in the case of a commit made by a topology running an older version of Storm and until the first commit is done by this topology for each topic-partition. Ater that the first commit logs will no longer occur. I will look into if it makes sense caching this info.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/OffsetManager.java
hmcl: Will do.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
srdo: Sure, but we're still deserializing the metadata on every emit. I think we can very likely get away with deserializing once per commit, or even once per spout activation, since we don't need to serialize + deserialize if we only update the metadata when committing.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
hmcl: It returns false if it throws exception when attempting to deserialize OffsetAndMetadata committed using a previous version of Storm.

hmcl:
@srdo it should be good now. Can you please take a look. Thanks.

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/internal/OffsetManagerTest.java
```diff
@@ -55,12 +58,12 @@ public void testSkipMissingOffsetsWhenFindingNextCommitOffsetWithGapInMiddleOfAc
...
         manager.addToAckMsgs(getMessageId(initialFetchOffset + 6));
         
-        assertThat("The offset manager should not skip past offset 5 which is still pending", manager.findNextCommitOffset().offset(), is(initialFetchOffset + 3));
+        assertThat("The offset manager should not skip past offset 5 which is still pending", manager.findNextCommitOffset(COMMIT_METADATA).offset(), is(initialFetchOffset + 3));
         
         manager.addToAckMsgs(getMessageId(initialFetchOffset + 5));
         
         assertThat("The offset manager should skip past the gap in acked messages, since the messages were not emitted", 
-            manager.findNextCommitOffset().offset(), is(initialFetchOffset + 7));
+            manager.findNextCommitOffset(COMMIT_METADATA).offset(), is(initialFetchOffset + 7));
```
srdo: It would be good to add a new test that verifies that the OffsetAndMetadata that is returned from findNextCommitOffset has the right metadata set.

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/OffsetAndMetadataMocks.java
```diff
@@ -0,0 +1,54 @@
...
+        OffsetAndMetadata oam = mock(OffsetAndMetadata.class);
+        when(consumerMock.committed(topicPartition))
+            .thenReturn(oam);
+
+        when(oam.metadata())
+            .thenReturn(COMMIT_METADATA);
+
+        when(topologyContext.getStormId()).thenReturn("tp1");
+
+        ObjectMapper om = mock(ObjectMapper.class);
```
srdo: These last three lines don't appear to do anything, because om isn't returned. Also I don't think there's any reason to mock the ObjectMapper, you should be able to use a real one.

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/OffsetAndMetadataMocks.java
```diff
@@ -0,0 +1,54 @@
...
+import org.apache.storm.kafka.spout.internal.CommitMetadata;
+import org.apache.storm.task.TopologyContext;
+
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class OffsetAndMetadataMocks {
+    public static final String COMMIT_METADATA = "{\"topologyId\":\"tp1\",\"taskId\":3,\"thread\":\"Thread-20\"}";
+
+    public static OffsetAndMetadata createMocksTree(KafkaConsumer<String, String> consumerMock,
```
srdo: I think you should leave off the topicPartition. You can use the any(TopicPartition.class) matcher in L39 instead.

It might also be good to rename to stubCommitted, and split it up a bit. The consumerMock/oam stub makes sense to have in a method together, but I think the topologyContext stubbing should be separated. Otherwise there's no way to test the case where you want to check what happens when the spout starts up with a commit metadata from an older topology (which I think we should add tests for to demonstrate that EARLIEST/LATEST work the way we say they do)

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutConfig.java
```diff
@@ -115,26 +115,30 @@ public KafkaSpoutConfig(Builder<K, V> builder) {
...
+     * in the partition, regardless of previous commits. If the topology is activated/deactivated this setting has no effect</li>
      * <li>UNCOMMITTED_EARLIEST means that the kafka spout polls records from the last committed offset, if any. If no offset has been
      * committed, it behaves as EARLIEST.</li>
      * <li>UNCOMMITTED_LATEST means that the kafka spout polls records from the last committed offset, if any. If no offset has been
      * committed, it behaves as LATEST.</li>
      * </ul>
-     *
      */
-    public static enum FirstPollOffsetStrategy {
+    public enum FirstPollOffsetStrategy {
```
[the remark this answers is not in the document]
hmcl: Enums are inherently static - it's redundant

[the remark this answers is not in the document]
srdo: Thanks, didn't know that.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutConfig.java
```diff
@@ -115,26 +115,30 @@ public KafkaSpoutConfig(Builder<K, V> builder) {
...
-     * records returned in the first poll. By default this parameter is set to UNCOMMITTED_EARLIEST. <br/><br/>
+     * Defines the offset used by the {@link KafkaSpout} in the first poll to Kafka broker. The choice of this parameter will affect
+     * the number of consumer records returned in the first poll. By default this parameter is set to UNCOMMITTED_EARLIEST. <br/><br/>
      * The allowed values are EARLIEST, LATEST, UNCOMMITTED_EARLIEST, UNCOMMITTED_LATEST. <br/>
      * <ul>
-     * <li>EARLIEST means that the kafka spout polls records starting in the first offset of the partition, regardless of previous
-     * commits</li>
-     * <li>LATEST means that the kafka spout polls records with offsets greater than the last offset in the partition, regardless of
-     * previous commits</li>
+     * <li>EARLIEST when the topology is first deployed the kafka spout polls records starting in the first offset of the
```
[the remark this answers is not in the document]
hmcl: OK, however this statement is correct because if a topology is deployed again it is a new topology because it has a new id and creates new object instances.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutConfig.java
```diff
@@ -115,26 +115,30 @@ public KafkaSpoutConfig(Builder<K, V> builder) {
     }
 
     /**
-     * The offset used by the Kafka spout in the first poll to Kafka broker. The choice of this parameter will affect the number of consumer
-     * records returned in the first poll. By default this parameter is set to UNCOMMITTED_EARLIEST. <br/><br/>
+     * Defines the offset used by the {@link KafkaSpout} in the first poll to Kafka broker. The choice of this parameter will affect
+     * the number of consumer records returned in the first poll. By default this parameter is set to UNCOMMITTED_EARLIEST. <br/><br/>
```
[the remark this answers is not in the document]
hmcl: OK.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -204,33 +226,70 @@ private void initialize(Collection<TopicPartition> partitions) {
...
          * Sets the cursor to the location dictated by the first poll strategy and returns the fetch offset.
          */
-        private long doSeek(TopicPartition tp, OffsetAndMetadata committedOffset) {
-            if (committedOffset != null) {             // offset was committed for this TopicPartition
-                if (firstPollOffsetStrategy.equals(EARLIEST)) {
-                    kafkaConsumer.seekToBeginning(Collections.singleton(tp));
-                } else if (firstPollOffsetStrategy.equals(LATEST)) {
-                    kafkaConsumer.seekToEnd(Collections.singleton(tp));
+        private long doSeek(TopicPartition newTp, OffsetAndMetadata committedOffset) {
+            LOG.trace("Seeking offset for topic-partition {} with {} and {}", newTp, firstPollOffsetStrategy, committedOffset);
```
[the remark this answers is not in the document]
hmcl: OK

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/internal/OffsetManagerTest.java
hmcl: OK.

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/OffsetAndMetadataMocks.java
hmcl: Done

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/OffsetAndMetadataMocks.java
hmcl: Done

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/MaxUncommittedOffsetTest.java
```diff
@@ -71,6 +75,10 @@
             1, KafkaSpoutRetryExponentialBackoff.TimeInterval.seconds(initialRetryDelaySecs))) //Retry once after a minute
         .build();
     private KafkaSpout<String, String> spout;
+    private final TopicPartition topicPartition = new TopicPartition(SingleTopicKafkaSpoutConfiguration.TOPIC, 1);
+    private KafkaConsumer<String, String> consumerMock;
```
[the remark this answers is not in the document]
hmcl: Done

hmcl:
@srdo can you please do one last review. Thanks.

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/KafkaSpoutEmitTest.java
```diff
@@ -100,6 +103,8 @@ public void testNextTupleEmitsFailedMessagesEvenWhenMaxUncommittedOffsetsIsExcee
             when(consumerMock.poll(anyLong()))
                 .thenReturn(new ConsumerRecords<>(records));
 
+            OffsetAndMetadataStub.committed(consumerMock);
```
srdo: I tried checking out the branch and running the test without this line. It seems to still pass. Maybe we should just get rid of it (and OffsetAndMetadataStub)?

hmcl:
These two changes address your comment: "_test that when you call findNextCommitOffset, it returns the metadata you put in)_"

https://github.com/hmcl/storm-apache/blob/baaf1fe6ab725ec50ecb09238f8957bc22a4e290/external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/internal/OffsetManagerTest.java#L67

https://github.com/hmcl/storm-apache/blob/baaf1fe6ab725ec50ecb09238f8957bc22a4e290/external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/internal/OffsetManagerTest.java#L82

As for the other tests, I started working on adding tests for scenarios around EARLIEST/LATEST and topology id. However, after working on it for awhile, it didn't feel natural that to test such a simple thing I had to mock so many objects (including KafkaSpoutConfig) and run the whole spout code. Then I decided that I would evaluate if the tests, code, or both should be considered for refactoring to make it easier to test. I need to dig into this further, but it seems to me that system tests are coupled with unit tests, and that makes it hard to write pure unit tests.

Let's merge https://github.com/apache/storm/pull/2464 and I will put my patch on top of it.

srdo:
Thanks, I missed those changes.

Regarding testing EARLIEST/LATEST, I don't understand why you'd need to mock KafkaSpoutConfig. I think it can be tested as an integration test, like the ones in SingleTopicKafkaSpoutTest, where you start a spout, emit and commit a tuple, deactivate/reactivate it, verify that it can emit and commit the next tuple, close the spout and create a new one and verify that the new spout starts over/starts at the end.

Alternatively you can do it with mocks like many of the other tests, basically doing the same flow but verifying that the spout calls seekToBeginning/seekToEnd instead of checking which tuples it emits.

Could you put a +1 on https://github.com/apache/storm/pull/2464 so I can merge?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/OffsetManager.java
```diff
@@ -185,29 +188,25 @@ public long commit(OffsetAndMetadata committedOffset) {
...
+     */
+    public boolean hasCommitted() {
+        return committed;
     }
 
     public boolean contains(KafkaSpoutMessageId msgId) {
         return ackedMsgs.contains(msgId);
     }
     
-    //VisibleForTesting
```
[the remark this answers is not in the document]
hmcl: I did when I cherry-picked the code from master. However, It was like this before, and the code with the annotation does not compile. I guess I would have to add the library dependencies for that. I just left it unchanged.

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/MaxUncommittedOffsetTest.java
```diff
@@ -71,6 +76,9 @@
             1, KafkaSpoutRetryExponentialBackoff.TimeInterval.seconds(initialRetryDelaySecs))) //Retry once after a minute
         .build();
     private KafkaSpout<String, String> spout;
+    private KafkaConsumer<String, String> consumer;
```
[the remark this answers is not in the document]
hmcl: Done

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/KafkaSpoutEmitTest.java
hmcl: Done

hmcl:
@srdo can you pls take a look. Thanks.

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/KafkaSpoutAbstractTest.java
```diff
@@ -0,0 +1,160 @@
...
+    @Rule
+    public KafkaUnitRule kafkaUnitRule = new KafkaUnitRule();
+
+    @Captor
+    ArgumentCaptor<Map<TopicPartition, OffsetAndMetadata>> commitCapture;
+
+    final TopologyContext topologyContext = mock(TopologyContext.class);
+    final Map<String, Object> conf = new HashMap<>();
+    final SpoutOutputCollector collector = mock(SpoutOutputCollector.class);
+    final long commitOffsetPeriodMs = 2_000;
```
srdo: A few of these parameters (commitOffsetPeriodMs, maxRetries, maxPollRecords) belong in the subclasses, since they're set as part of the KafkaSpoutConfig.

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/KafkaSpoutAbstractTest.java
```diff
@@ -0,0 +1,160 @@
...
+     *     <ul>spout.nexTuple()</ul>
+     *     <ul>verify messageId</ul>
+     *     <ul>spout.ack(msgId)</ul>
+     *     <ul>reset(collector) to be able to reuse mock</ul>
+     * </li>
+     *
+     * @param offset offset of message to be verified
+     * @return {@link ArgumentCaptor} of the messageId verified
+     */
+    ArgumentCaptor<Object> nextTuple_verifyEmitted_ack_resetCollectorMock(int offset) {
```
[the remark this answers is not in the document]
hmcl: This is the proper way to name tests, please see:
https://stackoverflow.com/questions/155436/unit-test-naming-best-practices. 

Even Google coding style mentions it:
https://google.github.io/styleguide/javaguide.html#s5.2.3-method-names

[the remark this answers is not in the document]
hmcl: Very long camel case names make it very hard to read, and properly asses what is being test.

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/KafkaSpoutAbstractTest.java
```diff
@@ -0,0 +1,160 @@
...
+
+        spout.ack(messageId.getValue());
+
+        reset(collector);
+
+        return messageId;
+    }
+
+    // offset and messageId are used interchangeably
+    ArgumentCaptor<Object> verifyMessageEmitted(int offest) {
```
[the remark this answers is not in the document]
hmcl: Done

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/KafkaSpoutTopologyDeployActivateDeactivateTest.java
```diff
@@ -0,0 +1,123 @@
...
+import static org.mockito.Mockito.when;
+
+public class KafkaSpoutTopologyDeployActivateDeactivateTest extends KafkaSpoutAbstractTest {
+    @Override
+    KafkaSpoutConfig<String, String> createSpoutConfig() {
+        return SingleTopicKafkaSpoutConfiguration.setCommonSpoutConfig(
+            KafkaSpoutConfig.builder("127.0.0.1:" + kafkaUnitRule.getKafkaUnit().getKafkaPort(),
+                Pattern.compile(SingleTopicKafkaSpoutConfiguration.TOPIC)))
+            .setOffsetCommitPeriodMs(commitOffsetPeriodMs)
+            .setRetry(new KafkaSpoutRetryExponentialBackoff(KafkaSpoutRetryExponentialBackoff.TimeInterval.seconds(0), KafkaSpoutRetryExponentialBackoff.TimeInterval.seconds(0),
```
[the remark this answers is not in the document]
hmcl: Done

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/KafkaSpoutTopologyDeployActivateDeactivateTest.java
```diff
@@ -0,0 +1,123 @@
...
+                maxRetries, KafkaSpoutRetryExponentialBackoff.TimeInterval.seconds(0)))
+            .setFirstPollOffsetStrategy(KafkaSpoutConfig.FirstPollOffsetStrategy.EARLIEST)
+            .setProp(ConsumerConfig.MAX_POLL_RECORDS_CONFIG, maxPollRecords)
```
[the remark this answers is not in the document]
hmcl: Done

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/KafkaSpoutTopologyDeployActivateDeactivateTest.java
```diff
@@ -0,0 +1,123 @@
...
+
+    @Test
+    public void test_FirstPollStrategy_Earliest_NotEnforced_OnTopologyActivateDeactivate() throws Exception {
+        try (Time.SimulatedTime simulatedTime = new Time.SimulatedTime()) {
+            final int messageCount = 2;
+            prepareSpout(messageCount);
+
+            nextTuple_verifyEmitted_ack_resetCollectorMock(0);
+
+            doNothing().when(consumerSpy).close();
```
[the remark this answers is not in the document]
hmcl: In the test you mention the consumer is mock. What you do is you create two mocks and you return them in succession. That does not solve the problem here, which is to poll the data from the same topic-partition where the last commit happen. Without this doNothing there are issues when restarting the spout, reading the partition with the last commit (which has the offsetAndMetadata we want to see enforced).

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/KafkaSpoutTopologyDeployActivateDeactivateTest.java
```diff
@@ -0,0 +1,123 @@
...
+            .setRetry(new KafkaSpoutRetryExponentialBackoff(KafkaSpoutRetryExponentialBackoff.TimeInterval.seconds(0), KafkaSpoutRetryExponentialBackoff.TimeInterval.seconds(0),
+                maxRetries, KafkaSpoutRetryExponentialBackoff.TimeInterval.seconds(0)))
+            .setFirstPollOffsetStrategy(KafkaSpoutConfig.FirstPollOffsetStrategy.EARLIEST)
+            .setProp(ConsumerConfig.MAX_POLL_RECORDS_CONFIG, maxPollRecords)
+            .build();
+    }
+
+    @Test
+    public void test_FirstPollStrategy_Earliest_NotEnforced_OnTopologyActivateDeactivate() throws Exception {
+        try (Time.SimulatedTime simulatedTime = new Time.SimulatedTime()) {
```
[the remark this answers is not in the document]
hmcl: Which method needs simulation time?

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/KafkaSpoutTopologyDeployActivateDeactivateTest.java
```diff
@@ -0,0 +1,123 @@
...
+            spout.activate();
+
+            nextTuple_verifyEmitted_ack_resetCollectorMock(1);
+
+            commitAndVerifyMessagesCommitted(messageCount, 2);
+        }
+    }
+
+    @Test
+    public void test_FirstPollStrategy_Earliest_NotEnforced_OnPartitionReassignment() throws Exception {
```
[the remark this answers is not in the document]
hmcl: This is technically correct. The spout is being restarted although it has the same partitions as before. Since this partitions have been committed to before, this case is indistinguishable from when a spout is running and gets new (this ones)  partitions assigned.

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/KafkaSpoutAbstractTest.java
hmcl: Done

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/KafkaSpoutTopologyDeployActivateDeactivateTest.java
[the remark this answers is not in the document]
hmcl: Done

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/KafkaSpoutAbstractTest.java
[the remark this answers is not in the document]
srdo: The stackoverflow link isn't about Java, and the answer seems to be talking about C#. The names here also don't follow the pattern outlined in the answer. You're right that the Google style guide allows you to use underscores in test method names.

I don't want to bikeshed you to death, and the test method names are fine, but nextTuple_verifyEmitted_ack_resetCollectorMock isn't a test method, and the name for it could be better I think. It's just emitting a tuple and acking it, why do we need to list out the function names it calls?

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/KafkaSpoutTopologyDeployActivateDeactivateTest.java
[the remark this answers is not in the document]
srdo: The consumer is not a mock in that test. It's a spy wrapping a real consumer (essentially a mock object that just tracks which methods were called on it, while passing the calls on to the real object). The reason you need the doNothing here is because deactivate closes the consumer. When you reactivate the spout, your mock KafkaConsumerFactory returns the same consumer as before, which is now closed. The test I linked makes sure to return a fresh consumer on the second call to the factory, so the reactivation happens with the new consumer.

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/KafkaSpoutTopologyDeployActivateDeactivateTest.java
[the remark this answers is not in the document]
srdo: The cases are really not equivalent, especially not if there were to be bugs in the reassignment code.

When you create a new spout you get rid of all the internal spout state and essentially emulate what would happen if the spout process was restarted. To emulate a reassignment, we would trigger the rebalance listener without replacing the spout instance, so the reassignment code would be using the same old KafkaConsumer and have access to all the previous spout state.

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/KafkaSpoutTopologyDeployActivateDeactivateTest.java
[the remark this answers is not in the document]
srdo: You probably spotted it, but just in case: commitAndVerifyMessagesCommitted advances simulated time to force a commit.

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/KafkaSpoutAbstractTest.java
```diff
@@ -0,0 +1,160 @@
...
+    /*
+     * Asserts that commitSync has been called once,
+     * that there are only commits on one topic,
+     * and that the committed offset covers messageCount messages
+     */
+    void verifyAllMessagesCommitted(long messageCount) {
+        verifyAllMessagesCommitted(messageCount, 1);
+    }
+
+    void verifyAllMessagesCommitted(long messageCount, int numTimes) {
```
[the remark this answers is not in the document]
hmcl: Done

hmcl:
@srdo It should be OK now. If you have further requests lest's file a refactoring JIRA and include in it refactoring some of the unit tests for better code reuse.

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/KafkaSpoutAbstractTest.java
srdo: This parameter should be initialized from the subclasses, otherwise it will be confusing when a subclass uses a different value and `commitAndVerifyAllMessagesCommitted` doesn't work. You can set the field here by reading it from the spout config in setUp.

hmcl:
@srdo done. Pls check and I will squash the commits right away. I would like to try to merge this in today. I will update the master PR with everything squashed already. Thanks.

srdo:
+1

hmcl:
@srdo Thanks for the review. I have squashed the commits and it is ready to merge.

PR #2469, opened 2017-12-19
STORM-2861: Explicit reference kafka-schema-registry-client

vesense:
https://issues.apache.org/jira/browse/STORM-2861

### DIFF
--- a/external/storm-hdfs/pom.xml
+++ b/external/storm-hdfs/pom.xml
@@ -199,11 +199,16 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>org.apache.avro</groupId>
+            <artifactId>avro</artifactId>
+            <version>1.7.4</version>
+        </dependency>
+        <dependency>
             <groupId>io.confluent</groupId>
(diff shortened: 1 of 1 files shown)


ptgoetz:
@vesense I tried to reproduce this by deleting Kafka-avro-serializer from my local maven repository, but the build succeeded. Can you elaborate a bit?

satishd:
@vesense kafka-avro-serializer module depends on kafka-schema-registry-client. This change does not seem to be really required. What issue are you facing with the existing dependency?

I also do not see any issues in building with the existing dependency like @ptgoetz.

vesense:
@ptgoetz @satishd 
Yes, I know that kafka-avro-serializer depends on kafka-schema-registry-client and avro.
There were io.confluent.kafka.schemaregistry.client.XXX not found errors when compiling the storm-hdfs code, and this PR fixed the issue. (To tell the truth, it's a bit strange.)
Another benefit is that users can know clearly about which avro version they are using. By default the avro module is from hadoop-common dependencies.

PR #2473, opened 2017-12-19
[STORM-2868] Address handling activate/deactivate in multilang module files

bitcloud:
As it is now required to answer to the activation and deactivation, I added those to the javascript files because the examples stopped working.

I updated ruby and python as well, but they don't have that issue because they always send sync without waiting for a callback.

But now you should be able to implement those as well in your code depending on those libs.

### DIFF
--- a/storm-multilang/ruby/src/main/resources/resources/storm.rb
+++ b/storm-multilang/ruby/src/main/resources/resources/storm.rb
@@ -205,6 +205,10 @@ module Storm
 
     def open(conf, context); end
 
+    def activate; end
+
+    def deactivate; end
+
     def nextTuple; end
 
     def ack(id); end
@@ -219,6 +223,10 @@ module Storm
         while true
           msg = read_command
           case msg['command']
+            when 'activate'
+              activate
+            when 'deactivate'
+              deactivate
             when 'next'
               nextTuple
             when 'ack'
--- a/examples/storm-starter/multilang/resources/randomsentence.js
+++ b/examples/storm-starter/multilang/resources/randomsentence.js
@@ -62,6 +62,14 @@ RandomSentenceSpout.prototype.nextTuple = function(done) {
     },100);
 }
 
+RandomSentenceSpout.prototype.activate = function(done) {
+    done();
+}
+
+RandomSentenceSpout.prototype.deactivate = function(done) {
+    done();
+}
+
 RandomSentenceSpout.prototype.createNextTupleId = function() {
     var id = this.runningTupleId;
     this.runningTupleId++;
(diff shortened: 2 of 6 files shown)


bitcloud:
I stumbled over that while playing around with the 1.1.x branch, so you probably need to merge that to those branches as well. Not quite sure since when the new protocol is in place.

storm-multilang/ruby/src/main/resources/resources/storm.rb
```diff
@@ -205,6 +205,10 @@ class Spout
 
     def open(conf, context); end
 
+    def avtivate; end
+
+    def deavtivate; end
```
HeartSaVioR: deactivate

storm-multilang/ruby/src/main/resources/resources/storm.rb
```diff
@@ -205,6 +205,10 @@ class Spout
 
     def open(conf, context); end
 
+    def avtivate; end
```
HeartSaVioR: activate

HeartSaVioR:
@bitcloud 
Nice finding. Could you please follow-up workflow for contribution? Guide link is here: https://github.com/apache/storm/blob/master/DEVELOPER.md#workflows

We encourage making commits squashed into one, with commit title as same as JIRA number + summary (which will be same or at least similar to PR's title.)

Thanks in advance!

storm-multilang/ruby/src/main/resources/resources/storm.rb
bitcloud: good catch. fill fix it right away.

bitcloud:
hey @HeartSaVioR I checked those before and thought this finding applies to
`For small patches, feel free to submit pull requests directly for those patches.`
so there is no need for a jira ticket?
Misst that about squashing so gonna fix that right away :)
marry xmas! :)

HeartSaVioR:
@bitcloud 
I think it is always good to have associated issue (unless it is quick fix) but we could treat this as small patch too. Either is fine for me.
Btw, even small patch, you are encouraged to write both commit title and title of PR to summarize which things you're addressing. You could imagine the case finding a commit in commit list when we have encountered specific issue.

bitcloud:
You are very right. It just felt that those modules pretty much fly unter the radar, as there are no testcases as well currently. Will create an jira account and submit an issue.

I checked jira first as well before fixing and then there would be some documentation about it.

At some point it felt like the github issues are somehow synchronized to jira ;-)

Do you think the titel of commit and PR is now suitable enough?

HeartSaVioR:
@bitcloud Sorry to modify the comment. I should've added a new comment. Please re-check my latest comment above.

bitcloud:
way better. updated and created issue. hope we are all set now ;-)

PR #2474, opened 2017-12-20

hmcc:
Connected to https://issues.apache.org/jira/browse/STORM-2862.
Put contextual information in the ThreadContext for multilang logging.

I've updated the Python and Ruby resources in multilang, but not the JS one. That's because the JS one is written to support variable arguments (probably because of [STORM-2435](https://issues.apache.org/jira/browse/STORM-2435)). I can't see a way to support optional IDs without breaking the current behaviour.


HeartSaVioR:
+1 prior to merge. (It's kind of porting back.)

PR #2320, opened 2017-09-11
STORM-2736: fix o.a.s.b.BlobStoreUtils [ERROR] "Could not update the blob with key"

hmcc:
Connected to https://issues.apache.org/jira/browse/STORM-2736.
Check for empty nimbus info list before attempting to update the blobstore.

PR #2475, opened 2017-12-20

hmcc:
Connected to https://issues.apache.org/jira/browse/STORM-2862; `master` version of apache/storm#2474.

### DIFF
--- a/storm-client/src/jvm/org/apache/storm/spout/ShellSpout.java
+++ b/storm-client/src/jvm/org/apache/storm/spout/ShellSpout.java
@@ -25,7 +25,9 @@ import org.apache.storm.multilang.ShellMsg;
 import org.apache.storm.multilang.SpoutMsg;
 import org.apache.storm.task.TopologyContext;
 import org.apache.storm.utils.ObjectReader;
+import org.apache.storm.utils.ShellLogHandler;
 import org.apache.storm.utils.ShellProcess;
+import org.apache.storm.utils.ShellUtils;
 
 import java.util.Arrays;
 import java.util.Map;
@@ -50,6 +52,7 @@ public class ShellSpout implements ISpout {
     private SpoutOutputCollector _collector;
     private String[] _command;
     private Map<String, String> env = new HashMap<>();
+    private ShellLogHandler _logHandler;
     private ShellProcess _process;
     private volatile boolean _running = true;
     private volatile RuntimeException _exception;
@@ -110,6 +113,9 @@ public class ShellSpout implements ISpout {
 
         Number subpid = _process.launch(topoConf, context, changeDirectory);
         LOG.info("Launched subprocess with pid " + subpid);
+
+        _logHandler =  ShellUtils.getLogHandler(topoConf);
+        _logHandler.setUpContext(ShellSpout.class, _process, _context);
 
         heartBeatExecutorService = MoreExecutors.getExitingScheduledExecutorService(new ScheduledThreadPoolExecutor(1));
     }
@@ -191,7 +197,7 @@ public class ShellSpout implements ISpout {
                 if (command.equals("sync")) {
                     return;
                 } else if (command.equals("log")) {
-                    handleLog(shellMsg);
+                    _logHandler.log(shellMsg);
                 } else if (command.equals("error")) {
                     handleError(shellMsg.getMsg());
                 } else if (command.equals("emit")) {
@@ -218,34 +224,6 @@ public class ShellSpout implements ISpout {
             throw new RuntimeException(processInfo, e);
         } finally {
             completedWaitingSubprocess();
-        }
-    }
-
-
-    private void handleLog(ShellMsg shellMsg) {
-        String msg = shellMsg.getMsg();
-        msg = "ShellLog " + _process.getProcessInfoString() + " " + msg;
-        ShellMsg.ShellLogLevel logLevel = shellMsg.getLogLevel();
-
-        switch (logLevel) {
-            case TRACE:
-                LOG.trace(msg);
-                break;
-            case DEBUG:
-                LOG.debug(msg);
-                break;
-            case INFO:
-                LOG.info(msg);
-                break;
-            case WARN:
-                LOG.warn(msg);
-                break;
-            case ERROR:
-                LOG.error(msg);
-                break;
-            default:
-                LOG.info(msg);
-                break;
         }
     }
 
--- a/storm-client/src/jvm/org/apache/storm/task/ShellBolt.java
+++ b/storm-client/src/jvm/org/apache/storm/task/ShellBolt.java
@@ -24,12 +24,14 @@ import org.apache.storm.metric.api.IMetric;
 import org.apache.storm.metric.api.rpc.IShellMetric;
 import org.apache.storm.multilang.BoltMsg;
 import org.apache.storm.multilang.ShellMsg;
-import org.apache.storm.topology.ReportedFailedException;
 import org.apache.storm.tuple.Tuple;
 import org.apache.storm.utils.ConfigUtils;
 import org.apache.storm.utils.ObjectReader;
 import org.apache.storm.utils.ShellBoltMessageQueue;
+import org.apache.storm.utils.ShellLogHandler;
 import org.apache.storm.utils.ShellProcess;
+import org.apache.storm.utils.ShellUtils;
+
 import com.google.common.util.concurrent.MoreExecutors;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -78,6 +80,7 @@ public class ShellBolt implements IBolt {
 
     private String[] _command;
     private Map<String, String> env = new HashMap<>();
+    private ShellLogHandler _logHandler;
     private ShellProcess _process;
     private volatile boolean _running = true;
     private volatile Throwable _exception;
@@ -153,6 +156,9 @@ public class ShellBolt implements IBolt {
         //subprocesses must send their pid first thing
         Number subpid = _process.launch(topoConf, context, changeDirectory);
         LOG.info("Launched subprocess with pid " + subpid);
+
+        _logHandler =  ShellUtils.getLogHandler(topoConf);
+        _logHandler.setUpContext(ShellBolt.class, _process, _context);
 
         // reader
         _readerThread = new Thread(new BoltReaderRunnable());
@@ -246,34 +252,6 @@ public class ShellBolt implements IBolt {
         } else {
             _collector.emitDirect((int) shellMsg.getTask(),
                     shellMsg.getStream(), anchors, shellMsg.getTuple());
-        }
-    }
-
-    private void handleLog(ShellMsg shellMsg) {
-        String msg = shellMsg.getMsg();
-        msg = "ShellLog " + _process.getProcessInfoString() + " " + msg;
-        ShellMsg.ShellLogLevel logLevel = shellMsg.getLogLevel();
-
-        switch (logLevel) {
-            case TRACE:
-                LOG.trace(msg);
-                break;
-            case DEBUG:
-                LOG.debug(msg);
-                break;
-            case INFO:
-                LOG.info(msg);
-                break;
-            case WARN:
-                LOG.warn(msg);
-                break;
-            case ERROR:
-                LOG.error(msg);
-                _collector.reportError(new ReportedFailedException(msg));
-                break;
-            default:
-                LOG.info(msg);
-                break;
         }
     }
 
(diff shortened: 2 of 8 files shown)


storm-client/src/jvm/org/apache/storm/utils/ShellLogHandler.java
```diff
@@ -0,0 +1,113 @@
...
+     * @param context
+     *            - the current {@link TopologyContext}.
+     */
+    private static void updateContext(ShellMsg shellMsg, ShellProcess process, TopologyContext context) {
+        putIfNotNull(ID, shellMsg.getId());
+        // Calling this only once allows the same parent ID to be attached to
+        // all log messages from a tuple tree
+        if (!ThreadContext.containsKey(PARENT)) {
+            putIfNotNull(PARENT, shellMsg.getId());
+        }
```
revans2: This really needs some kind of documentation if we are going to start using id with logging.  I just don't know how you expect to use this, so I have no idea how to say if it works properly or not.

Also you missed javascript support.

https://github.com/apache/storm/blob/a18657d0cd01601dfd193dccade1f601fbef94b4/storm-multilang/javascript/src/main/resources/resources/storm.js#L51-L57

storm-client/src/jvm/org/apache/storm/utils/ShellLogHandler.java
```diff
@@ -0,0 +1,113 @@
...
+     *            - the current {@link ShellProcess}. Optional.
+     * @param context
+     *            - the current {@link TopologyContext}. Optional.
+     */
+    public static void handleLog(ShellMsg shellMsg, ShellProcess process, TopologyContext context) {
+        if (shellMsg == null) {
+            throw new IllegalArgumentException("shellMsg is required");
+        }
+
+        updateContext(shellMsg, process, context);
```
revans2: Could we try and optimize the overhead on this?  Users should not be logging much if anything, but it might be nice to check if the current log level is enabled before we try to update the context, as I don't know how expensive it is.

We can probably make that happen by making some changes to the ShellMsg.ShellLogLevel enum to know more about log4j

storm-client/src/jvm/org/apache/storm/utils/ShellLogHandler.java
hmcc: Thanks for looking at this @revans2.

Our use case is that we're running a Storm topology written in Python. As well as human-readable logs, the topology is also configured to write JSON logs which are sent to ELK. 

The parent ID is so that we can easily search for all log messages associated with a particular originating tuple.

@HeartSaVioR has already made some comments on the [issue](https://issues.apache.org/jira/browse/STORM-2862), saying that modifying the format of the logs could break backwards compatibility. I've asked him about maybe making the logging output more configurable instead. If we went that way, we could drop the Python and Ruby part of this change, and handle that in our code.

Sorry, I'd explained the missing JS support on the #2474 but it should have been here -  the JS is written to support variable arguments (probably because of [STORM-2435](https://issues.apache.org/jira/browse/STORM-2435)). I can't see a way to support optional IDs without breaking the current behaviour.

HeartSaVioR: I think letting users leave the log message with providing enough information is most flexible way. `handleLog` is adding some context which some users might not want and in some case could disturb parsing the message.
So the ideal approach from my side is publishing all the necessary context to multi-lang running subprocess and let subprocess build the full message (without context which log4j2 format provides), but it might bring some backward incompatibility like I said, or even some overhead because it should be passed to multilang and longer message should be passed to ShellSpout/ShellBolt.

I'm OK to follow proposal provided by @hmcc, which may not require huge change. ID of the originating tuple is not available (we're sending generated new ID to multilang) so need to be added in some way though.

revans2: @hmcc 

If parent id is intended to be used to search for all log messages about a given tuple tree then there is a bug.  You set parent ID once and only once.  This means that "parent" will be set to the ID of the first log message with an ID ever sent.  Every single tuple is likely to have a different parent so it will end up being a totally useless value.


Also I agree with @HeartSaVioR I think there is a much cleaner way to get the same results that you want, and not having all of the overhead for those that don't want it.  Because the logging is thread specific.

I would propose a ShellLogHandler interface that can be set for implementations of ShellBolt and ShellSpout.  That way you can override it, and there is no overhead for people who don't want it.  I would propose something like

```
interface ShellLogHandler {
  void setupContext(ShellProcess process, TopologyContext context);
  void log(ShellMessage msg);
}
```

`setupContext` would be called from the BoltReaderRunnable thread.  It could set the ThreadContext for that thread for things like the PID, etc.  These are things that will not change during the lifetime of the Bolt.

For the spout it is going to be more overhead because there is no way currently to know if you are on a thread by yourself or sharing a thread with another spout.  So for each log message the spout would call `setupContext` followed by `log`.

You also don't need to modify storm.py as you can just use sendMsgToParent yourself directly and have your own custom helper log methods with the ID.

hmcc: Thanks, that sounds like a plan.

What do you prefer - close this and create a new PR, or leave this open, revert the commit and add a new commit with those changes?

revans2: Either is fine with me.

hmcc:
@revans2, @HeartSaVioR, is this any better? Thanks.

storm-client/src/jvm/org/apache/storm/spout/ShellSpout.java
```diff
@@ -50,6 +51,7 @@
     private SpoutOutputCollector _collector;
     private String[] _command;
     private Map<String, String> env = new HashMap<>();
+    private ShellLogHandler _logHandler;
```
HeartSaVioR: You seems missing initialization here on spout.

storm-client/src/jvm/org/apache/storm/task/ShellBolt.java
```diff
@@ -145,6 +148,8 @@ public void prepare(Map<String, Object> topoConf, TopologyContext context,
             workerTimeoutMills = 1000 * ObjectReader.getInt(topoConf.get(Config.SUPERVISOR_WORKER_TIMEOUT_SECS));
         }
 
+        _logHandler = ShellUtils.getLogHandler(topoConf);
```
HeartSaVioR: This should be also placed to spout as I commented above.

storm-client/src/jvm/org/apache/storm/spout/ShellSpout.java
```diff
@@ -177,6 +179,7 @@ private void handleMetrics(ShellMsg shellMsg) {
     private void querySubprocess() {
         try {
             markWaitingSubprocess();
+            _logHandler.setUpContext(_process, _context);
```
HeartSaVioR: This can be on-time setup, and maybe better to move this to `open()`, since we are more familiar with initializing everything in it.

storm-client/src/jvm/org/apache/storm/task/ShellBolt.java
```diff
@@ -351,6 +328,7 @@ public void run() {
 
     private class BoltReaderRunnable implements Runnable {
         public void run() {
+            _logHandler.setUpContext(_process, _context);
```
HeartSaVioR: This is effectively one-time setup, but if we really want to have one-time setup, maybe better to initialize it in prepare(), since we are more familiar with initializing everything in it.

storm-client/src/jvm/org/apache/storm/utils/DefaultShellLogHandler.java
```diff
@@ -0,0 +1,100 @@
...
+
+package org.apache.storm.utils;
+
+import org.apache.storm.multilang.ShellMsg;
+import org.apache.storm.task.TopologyContext;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Handle output from non-JVM processes.
```
HeartSaVioR: 1. This explanation looks suitable in ShellLogHandler, not this class. And it handles log, not output.
2. Maybe better to describe that this is default implementation.

storm-client/src/jvm/org/apache/storm/utils/DefaultShellLogHandler.java
```diff
@@ -0,0 +1,100 @@
...
+ */
+public class DefaultShellLogHandler implements ShellLogHandler {
+    public static final Logger LOG = LoggerFactory.getLogger(DefaultShellLogHandler.class);
```
HeartSaVioR: Looks like we're losing class information which is logging. (I mean we no longer write log with ShellSpout/ShellBolt.) 

Given that we are creating ShellLogHandler instance per task, we can pass class name in `setUpContext()` (interface needs to be changed) and initialize logger based on the class name. Makes sense?

storm-client/src/jvm/org/apache/storm/utils/ShellLogHandler.java
```diff
@@ -0,0 +1,49 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.storm.utils;
+
+import org.apache.storm.multilang.ShellMsg;
+import org.apache.storm.task.TopologyContext;
+
+/**
+ * Handles output from multilang processes.
```
HeartSaVioR: output -> log

storm-client/src/jvm/org/apache/storm/utils/ShellUtils.java
```diff
@@ -464,4 +465,18 @@ public void run() {
...
+    public static ShellLogHandler getLogHandler(Map<String, Object> topoConf) {
+        if (topoConf == null) {
+            throw new IllegalArgumentException("Config is required");
+        }
+
+        if (topoConf.containsKey(Config.TOPOLOGY_MULTILANG_LOG_HANDLER)) {
+            try {
+                return (ShellLogHandler) Class.forName(topoConf.get(Config.TOPOLOGY_MULTILANG_LOG_HANDLER).toString()).newInstance();
+            } catch (ClassCastException | InstantiationException | IllegalAccessException | ClassNotFoundException e) {
+                LOG.warn(e.toString());
```
HeartSaVioR: Let's throw exception (propagation with wrapping RuntimeException) and let worker crash instead of leaving warning message. This is easy to be missed, and there's a principle of Storm: fail-fast.

HeartSaVioR: And also modify tests as well.

storm-client/test/resources/log4j2-test.xml
```diff
@@ -0,0 +1,32 @@
...
+-->
+
+<configuration monitorInterval="60">
+    <Appenders>
+        <Console name="Console" target="SYSTEM_OUT" follow="true">
+            <PatternLayout pattern="%-4r [%t] %-5p %c{1.} - %msg%n"/>
+        </Console>
+    </Appenders>
+    <Loggers>
+        <!-- suppress INFO o.a.s.u.ShellLogHandler - null when testing -->
```
HeartSaVioR: it shouldn't be o.a.s.u.ShellLogHandler, and you would be OK to not add this if new tests doesn't produce huge logs.

storm-client/src/jvm/org/apache/storm/utils/DefaultShellLogHandler.java
hmcc: Yep, makes sense - thanks!

HeartSaVioR:
@hmcc 
Could you squash commits into one? #2474 as well. Thanks in advance!

hmcc:
@HeartSaVioR, I understand that the 1.2.0 and 1.1.2 releases might be coming soon?

Do you know if 1.1.2 is going to be much before 1.2.0? If so, would it be OK if I made a PR for the 1.1.x branch too?

HeartSaVioR:
@hmcc 
Could you change your commit title to include issue number as well? Like `STORM-2862: Move multilang logging to a shared class and make this class configurable`.

I guess I could port back your PR against 1.x-branch into 1.1.x-branch if there's no (or small) conflict.

hmcc:
@HeartSaVioR done.

If 1.1.2 is the next release then in an ideal world we'd get both this and #2320 in there. Is that possible? I don't want to add to your workload so I'm more than happy to create the PRs if so. 

If it's too late for #2320 then we'll wait for 1.2.0 (and I'll make sure I create 1.1.x PRs as well in future).

HeartSaVioR:
@hmcc Really appreciated to your help! Please feel free to submit one. It would be easy if cherry-pick doesn't make any conflict so I'm normally OK with porting back between 1.x and 1.1.x.

hmcc:
1.1.x PR: #2521. It looks like #2320 is already in 1.1.x branch, thanks!

PR #2477, opened 2017-12-21
STORM-2863: Reset NormalizedResources for tests

revans2:


srdo:
@revans2 Could you take a look at https://github.com/apache/storm/pull/2463? I think it will resolve this issue too.

revans2:
@srdo yes #2463 would resolve the issue too, but it loses most of the reasons why I put in NormalizedResources in the first place.  Lets move the discussion over there for now.

storm-server/src/test/java/org/apache/storm/scheduler/resource/strategies/scheduling/TestDefaultResourceAwareStrategy.java
```diff
@@ -65,11 +67,33 @@
...
+            return result;
+        }
+    };
+
     /**
      * test if the scheduling logic for the DefaultResourceAwareStrategy is correct
      */
     @Test
     public void testDefaultResourceAwareStrategySharedMemory() {
+        NormalizedResources.resetResourceNames();
```
srdo: Isn't it better to put this in `@Before` or `@After` (or ideally a `TestRule`)? I think it's likely we'll forget to copy this to new tests.

revans2: Good point I will update it

srdo:
Does https://github.com/revans2/incubator-storm/blob/6aaf8ee7cb042960831547a3ef07ce68a1105970/storm-server/src/test/java/org/apache/storm/scheduler/resource/strategies/scheduling/TestGenericResourceAwareStrategy.java also affect the static fields?

revans2:
@srdo I moved it over to being a Rule annotation

srdo:
Thanks, +1 assuming TestGenericResourceAwareStrategy doesn't touch the static fields.

PR #2480, opened 2017-12-25
STORM-2867: Add consumer lag metrics to KafkaSpout

omkreddy:


external/storm-kafka-client/src/main/java/org/apache/storm/kafka/KafkaUtils.java
```diff
@@ -0,0 +1,114 @@
...
+import org.apache.storm.metric.api.IMetric;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+import java.util.TreeMap;
+
+public class KafkaUtils {
```
hmcl: Why the need of the KafkaUtils wrapper class? Why not just have the KafkaOffsetMetric class ?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/KafkaUtils.java
```diff
@@ -0,0 +1,114 @@
...
+                    Set<TopicPartition> topicPartitions = offsetManagers.keySet();
+
+                    Map<TopicPartition, Long> beginningOffsets= kafkaConsumer.beginningOffsets(topicPartitions);
+                    Map<TopicPartition, Long> endOffsets= kafkaConsumer.endOffsets(topicPartitions);
+
+                    for (Map.Entry<TopicPartition, OffsetManager> entry : offsetManagers.entrySet()) {
+                        TopicPartition topicPartition = entry.getKey();
+                        OffsetManager offsetManager = entry.getValue();
+
+                        long latestTimeOffset = beginningOffsets.get(topicPartition);
```
HeartSaVioR: Looks like twos are swapped: beginning -> earliest, end -> latest

https://kafka.apache.org/0101/javadoc/org/apache/kafka/clients/consumer/KafkaConsumer.html#beginningOffsets(java.util.Collection)

https://kafka.apache.org/0101/javadoc/org/apache/kafka/clients/consumer/KafkaConsumer.html#endOffsets(java.util.Collection)

examples/storm-kafka-client-examples/pom.xml
```diff
@@ -73,12 +73,6 @@
             <version>${project.version}</version>
             <scope>${provided.scope}</scope>
         </dependency>
-        <dependency>
```
HeartSaVioR: Do you run example topology after this change? If then I'm OK. Just to verify the change doesn't hurt anything.

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/KafkaSpoutConfigTest.java
```diff
@@ -49,6 +49,7 @@ public void testBasic() {
         expected.put(ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG, StringDeserializer.class);
         expected.put(ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG, StringDeserializer.class);
         assertEquals(expected, conf.getKafkaProps());
+        assertEquals(KafkaSpoutConfig.DEFAULT_METRICS_TIME_BUCKET_SIZE_SECONDS, conf.getMetricsTimeBucketSizeInSecs());
```
satishd: This may be in plan to add in next commits. You may want to add a test to see whether the configured property is also taken into account.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/KafkaUtils.java
satishd: May be a copy/paste issue, taken from https://github.com/apache/storm/blob/master/external/storm-kafka/src/jvm/org/apache/storm/kafka/KafkaUtils.java

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutConfig.java
```diff
@@ -182,6 +186,7 @@ public KafkaSpoutConfig(Builder<K, V> builder) {
         private boolean emitNullTuples = false;
         private ProcessingGuarantee processingGuarantee = DEFAULT_PROCESSING_GUARANTEE;
         private boolean tupleTrackingEnforced = false;
+        public int metricsTimeBucketSizeInSecs = DEFAULT_METRICS_TIME_BUCKET_SIZE_SECONDS;
```
satishd: access of this field should be private instead of public.

examples/storm-kafka-client-examples/pom.xml
srdo: Why was this removed?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/KafkaUtils.java
srdo: It would be better to make a package for this (e.g. org.apache.storm.kafka.spout.metrics?), instead of using the KafkaUtils class for namespacing.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/KafkaUtils.java
```diff
@@ -0,0 +1,114 @@
...
+import java.util.Set;
+import java.util.TreeMap;
+
+public class KafkaUtils {
+
+    private static final Logger LOG = LoggerFactory.getLogger(KafkaUtils.class);
+
+    public static class KafkaOffsetMetric implements IMetric {
+        Map<TopicPartition, OffsetManager> offsetManagers;
+        KafkaConsumer kafkaConsumer;
```
srdo: This will break. The spout replaces the consumer when it is deactivated, so cycling topology activation will break this metric. I would either use a KafkaConsumer supplier here, or add a setter to this metric.

Also why not private?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/KafkaUtils.java
```diff
@@ -0,0 +1,114 @@
...
+        private class TopicMetrics {
+            long totalSpoutLag = 0;
+            long totalEarliestTimeOffset = 0;
+            long totalLatestTimeOffset = 0;
+            long totalLatestEmittedOffset = 0;
+            long totalLatestCompletedOffset = 0;
+        }
+
+        @Override
+        public Object getValueAndReset() {
```
srdo: Remind me if this method is called by a different thread from the executor thread that runs nextTuple and the other spout methods? If so we need to add some synchronization on the offset managers. If not I don't think we need a ConcurrentHashMap in the spout.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/OffsetManager.java
```diff
@@ -57,7 +58,8 @@ public void addToAckMsgs(KafkaSpoutMessageId msgId) {          // O(Log N)
     }
 
     public void addToEmitMsgs(long offset) {
-        this.emittedOffsets.add(offset);                  // O(Log N)
+        this.emittedOffsets.add(offset);  // O(Log N)
+        this.latestEmittedOffset = offset;
```
srdo: I don't think this works. This will just set the latest emitted offset to the offset that was most recently emitted. I think you want the highest offset that has been emitted.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/KafkaUtils.java
```diff
@@ -0,0 +1,114 @@
...
+                        ret.put(topic + "/" + "totalLatestTimeOffset", topicMetrics.totalLatestTimeOffset);
+                        ret.put(topic + "/" + "totalLatestEmittedOffset", topicMetrics.totalLatestEmittedOffset);
+                        ret.put(topic + "/" + "totalLatestCompletedOffset", topicMetrics.totalLatestCompletedOffset);
+                    }
+
+                    return ret;
+                } else {
+                    LOG.info("Metrics Tick: Not enough data to calculate spout lag.");
+                }
+            } catch (Throwable t) {
```
srdo: Why are we catching all exceptions here?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -125,12 +129,22 @@ public void open(Map conf, TopologyContext context, SpoutOutputCollector collect
...
         refreshSubscriptionTimer = new Timer(TIMER_DELAY_MS, kafkaSpoutConfig.getPartitionRefreshPeriodMs(), TimeUnit.MILLISECONDS);
 
-        offsetManagers = new HashMap<>();
+        offsetManagers = new ConcurrentHashMap<>();
         emitted = new HashSet<>();
         waitingToEmit = Collections.emptyListIterator();
 
         tupleListener.open(conf, context);
 
+        context.registerMetric("kafkaOffset", new IMetric() {
```
srdo: Why do we need the wrapper metric?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -553,9 +567,11 @@ public void fail(Object messageId) {
     public void activate() {
         try {
             subscribeKafkaConsumer();
+            metricsConsumer = new KafkaConsumer<>(kafkaSpoutConfig.getKafkaProps());
```
srdo: Why do we need a separate consumer for the metric? Also if we really need two consumers, you should create this via the factory. Otherwise you prevent us from stubbing/spying on this consumer.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -568,6 +584,8 @@ private void subscribeKafkaConsumer() {
     public void deactivate() {
         try {
             shutdown();
+            metricsConsumer.close();
```
srdo: I think this belongs in shutdown.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutConfig.java
```diff
@@ -667,6 +672,16 @@ private void setNonNullSerDeKafkaProp(SerializableDeserializer<K> keyDes, Class<
             return this;
         }
 
+        /**
+         * The time period that metrics data in bucketed into.
+         * @param metricsTimeBucketSizeInSecs
+         * @return
```
srdo: Nit: Don't add empty `@return`, just leave it off.

pom.xml
```diff
@@ -278,7 +278,7 @@
         <storm.kafka.artifact.id>kafka_2.10</storm.kafka.artifact.id>
 
         <!-- kafka version used by new storm-kafka-client spout code -->
-        <storm.kafka.client.version>0.10.0.0</storm.kafka.client.version>
+        <storm.kafka.client.version>0.11.0.2</storm.kafka.client.version>
```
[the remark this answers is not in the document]
srdo: We need to upgrade to at least 0.10.1.0, since that's where the `beginningOffsets` and `endOffsets` methods were added to the KafkaConsumer. I think upgrading the consumer and adding this patch would probably be a breaking change, because the spout won't work with 0.10.0.x client libraries/brokers anymore.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/KafkaUtils.java
omkreddy: Removed KafkaUtils wrapper. moved KafkaOffsetMetric to org.apache.storm.kafka.spout.metrics package

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/KafkaUtils.java
omkreddy: Updated the code

pom.xml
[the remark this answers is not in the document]
omkreddy: yes, we need at-least 0.10.1.0 kafka clients and its a breaking change. We check for any other options

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/KafkaUtils.java
```diff
@@ -0,0 +1,114 @@
...
+                        long spoutLag = latestTimeOffset - latestCompletedOffset;
+
+                        String metricPath = topicPartition.topic()  + "/partition_" + topicPartition.partition();
+                        ret.put(metricPath + "/" + "spoutLag", spoutLag);
+                        ret.put(metricPath + "/" + "earliestTimeOffset", earliestTimeOffset);
+                        ret.put(metricPath + "/" + "latestTimeOffset", latestTimeOffset);
+                        ret.put(metricPath + "/" + "latestEmittedOffset", latestEmittedOffset);
+                        ret.put(metricPath + "/" + "latestCompletedOffset", latestCompletedOffset);
+
+                        if (!topicMetricsMap.containsKey(topicPartition.topic())) {
```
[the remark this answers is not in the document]
omkreddy: Done

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/KafkaSpoutConfigTest.java
omkreddy: added a small testcase

examples/storm-kafka-client-examples/pom.xml
omkreddy: I think we don't need kafka server jar to run  examples. Ran the tests without this dependency. But i reverted the change, since it is not related to this PR.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/KafkaUtils.java
omkreddy: Removed Throwable from catch clause.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
omkreddy: Removed the wrapper metric.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/KafkaUtils.java
omkreddy: Thats good point. Will check  if it is called from same thread or different thread, based on that I will remove ConcurrentHashMap usage. We also don't need separate consumer for the metric, it it is called from the same thread. Will check and update the PR.

arunmahadevan: @omkreddy , It seems both are invoked from the spout executor thread. The getValueAndReset is invoked from the tuple-action-fn which is invoked from [here](https://github.com/apache/storm/blob/1.x-branch/storm-core/src/clj/org/apache/storm/daemon/executor.clj#L627) and the nextTuple [here](https://github.com/apache/storm/blob/1.x-branch/storm-core/src/clj/org/apache/storm/daemon/executor.clj#L647)

pom.xml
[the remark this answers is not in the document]
HeartSaVioR: Actually we are "effectively" dropping support of Kafka 0.9.x while moving from Storm 1.0.x to 1.1.0, because critical bugs in storm-kafka-client are not ported back to 1.0.x version line. I think we should try to avoid doing it again, but not sure for the cases like this. (bump the version for utilizing new feature)

Btw, if Kafka doesn't always guarantee forward compatibility, we may need to define support versions of Kafka (and others, if we could maintain) and maintain that.

srdo:
Regarding the Kafka upgrade in this:

I think we can keep support for 0.10.0.x by adding a check to KafkaSpout.open that checks if the two new consumer methods are present via reflection, and disables the new metric if they aren't. Ideally we'd be able to use this type of approach any time we add a feature that uses a new method on the KafkaConsumer.

It would be fine by me if we only do this in 1.x to retain backward compatibility, and just leave out the check in master. Maybe we add a jira issue to 2.0.0 that says we're requiring at least 0.10.1.0, so it's visible to users upgrading to 2.0.0. 

If we really need to support old Kafka versions indefinitely, we can hopefully just keep the reflection checks, but I'd rather use Storm major versions as an opportunity to clean up the code and require newer Kafka versions.

I believe Kafka guarantees interface stability for KafkaConsumer as of 0.10.0.0. Since users can specify which kafka-clients version they want when they define their topologies, we should not need to maintain different versions of storm-kafka-client for different Kafka versions.

HeartSaVioR:
+1 on @srdo 's approach. Btw, Flink has Kafka connector modules for each major version, so if necessary, we could try out similar approach. If we want to separate 0.10.x and 0.11.x+, it might be leveraging support of exactly-once semantics in Kafka 0.11.0+.

srdo:
Sure, sounds good to me. If we can get away with a little bit of reflection I'd rather try that first, but putting out a storm-kafka-client-0.10.1, storm-kafka-client-1.0.0 etc. would be good too, as long as we have a storm-kafka-client-base project for most of the code. I just don't want to get into a situation where we have to maintain 2-3 Storm branch versions * 2-3 storm-kafka-client variants per branch.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/metrics/KafkaOffsetMetric.java
```diff
@@ -0,0 +1,114 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+public class KafkaOffsetMetric implements IMetric {
+
+    private static final Logger LOG = LoggerFactory.getLogger(KafkaOffsetMetric.class);
+    private Map<TopicPartition, OffsetManager> offsetManagers;
```
srdo: nit: It might be better to use a getter or Supplier to access the field in KafkaSpout, rather than copying the Map reference in here. If someone does e.g. `offsetManagers = new HashMap<>()` in the spout, this reference ends up pointing to the wrong thing.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/metrics/KafkaOffsetMetric.java
```diff
@@ -0,0 +1,114 @@
...
+                    TopicMetrics topicMetrics = e.getValue();
+                    ret.put(topic + "/" + "totalSpoutLag", topicMetrics.totalSpoutLag);
+                    ret.put(topic + "/" + "totalEarliestTimeOffset", topicMetrics.totalEarliestTimeOffset);
+                    ret.put(topic + "/" + "totalLatestTimeOffset", topicMetrics.totalLatestTimeOffset);
+                    ret.put(topic + "/" + "totalLatestEmittedOffset", topicMetrics.totalLatestEmittedOffset);
+                    ret.put(topic + "/" + "totalLatestCompletedOffset", topicMetrics.totalLatestCompletedOffset);
+                }
+                LOG.debug("Metrics Tick: value : {}", ret);
+                return ret;
+            } else {
```
srdo: Nit: Invert the if-else here to reduce indentation. It also allows you to avoid the "catch-all" null return in the bottom.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/metrics/KafkaOffsetMetric.java
```diff
@@ -0,0 +1,114 @@
...
+                LOG.info("Metrics Tick: Not enough data to calculate spout lag.");
+            }
+        } catch (Exception t) {
```
srdo: Sorry, I wasn't being clear with my earlier comment: Why are we catching exceptions here at all?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/metrics/KafkaOffsetMetric.java
```diff
@@ -0,0 +1,114 @@
...
+import org.apache.storm.kafka.spout.internal.OffsetManager;
+import org.apache.storm.metric.api.IMetric;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+public class KafkaOffsetMetric implements IMetric {
```
srdo: Could we add a test or two for this class? Maybe a test in KafkaSpoutSingleTopicTest that starts up a spout and checks that it can get metrics for the subscribed partition?

omkreddy: Added a simple testcase

omkreddy:
@srdo Thanks for the detailed review and suggestions. Updated the PR with review comments. Used refections approach to enable or disable metrics.  Pls take a look.

cc @HeartSaVioR  @satishd

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/metrics/KafkaOffsetMetric.java
```diff
@@ -31,77 +32,75 @@
 public class KafkaOffsetMetric implements IMetric {
 
     private static final Logger LOG = LoggerFactory.getLogger(KafkaOffsetMetric.class);
-    private Map<TopicPartition, OffsetManager> offsetManagers;
-    private KafkaConsumer kafkaConsumer;
+    private Supplier<Map<TopicPartition, OffsetManager>> offsetManagerSupplier;
```
srdo: Nit: These can be final now

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/metrics/KafkaOffsetMetric.java
```diff
@@ -31,77 +32,75 @@
...
+            topicMetrics.totalLatestTimeOffset += latestTimeOffset;
+            topicMetrics.totalLatestEmittedOffset += latestEmittedOffset;
+            topicMetrics.totalLatestCompletedOffset += latestCompletedOffset;
+        }
+
+        for(Map.Entry<String, TopicMetrics> e : topicMetricsMap.entrySet()) {
+            String topic = e.getKey();
+            TopicMetrics topicMetrics = e.getValue();
+            result.put(topic + "/" + "totalSpoutLag", topicMetrics.totalSpoutLag);
+            result.put(topic + "/" + "totalEarliestTimeOffset", topicMetrics.totalEarliestTimeOffset);
```
srdo: I'm not sure I understand a few of these metrics. What does e.g. totalEarliestTimeOffset tell the user? To me it looks like totalSpoutLag makes a lot of sense to have, but the other metric totals aren't very meaningful.

Just to give an example, if I get out a value of 10 for totalLatestTimeOffset, I don't understand what use that could have, because it's just telling me that if I add together the log end offsets of my assigned partitions, I get 10. It could be used to calculate how many messages are in all the partitions, but in that case it makes more sense to return e.g. `totalRecordsInAssignedPartitions = totalLatestTimeOffset - totalEarliestTimeOffset - 1` (disregarding log compaction).

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/KafkaSpoutSingleTopicTest.java
```diff
@@ -343,4 +344,48 @@ public void testSpoutMustRefreshPartitionsEvenIfNotPolling() throws Exception {
...
+        assertEquals(0, offsetMetric.get(SingleTopicKafkaSpoutConfiguration.TOPIC+"/totalEarliestTimeOffset").longValue());
+        // the offset of the last available message + 1.
+        assertEquals(10, offsetMetric.get(SingleTopicKafkaSpoutConfiguration.TOPIC+"/totalLatestTimeOffset").longValue());
+        assertEquals(0, offsetMetric.get(SingleTopicKafkaSpoutConfiguration.TOPIC+"/totalLatestEmittedOffset").longValue());
+        assertEquals(0, offsetMetric.get(SingleTopicKafkaSpoutConfiguration.TOPIC+"/totalLatestCompletedOffset").longValue());
+        //totalSpoutLag = totalLatestTimeOffset-totalLatestCompletedOffset
+        assertEquals(10, offsetMetric.get(SingleTopicKafkaSpoutConfiguration.TOPIC+"/totalSpoutLag").longValue());
+
+        //Emit and Ack all messages
+        for(int i = 0; i < messageCount; i++) {
```
srdo: I think you can use nextTuple_verifyEmitted_ack_resetCollector(int offset) from the superclass to do everything in this block.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -739,4 +764,9 @@ public boolean shouldPoll() {
             return !this.pollablePartitions.isEmpty();
         }
     }
+
+    @VisibleForTesting
+    KafkaOffsetMetric getKafkaOffsetMetric() {
```
srdo: Optional: We could avoid this getter by using the same trick with a factory as we use for injecting the KafkaConsumer. If we add a KafkaOffsetMetricFactory to the spout constructor, you can create the metric in the test and pass it into the spout instead of letting the spout construct the metric.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/metrics/KafkaOffsetMetric.java
omkreddy: @srdo  I agree. totalEarliestTimeOffset alone is not much useful. I just added it to make the metrics compatible with old spout (storm-kafka) metrics. 

adding totalRecordsInAssignedPartitions is good idea. I have included totalRecordsInAssignedPartitions to metrics. Can you explain why we need to decrease the count by 1?  I am not understood what do you mean by disregarding log compaction?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
omkreddy: Agreed,  but I feel current approach is simpler. maybe we can evolve when we add more metrics.

srdo: Yes, makes sense. Thanks.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/metrics/KafkaOffsetMetric.java
srdo: The reason I'm asking is because I'd rather only copy over the useful metrics from storm-kafka. It's easier to not add metrics we don't think are useful, than to remove them later.

Sorry, I made a mistake. We don't need to subtract 1. My bad.

Regarding why we disregard log compaction, if there's a partition with a log end offset (latestTimeOffset) of 10, and a beginningOffset of 0, we'd expect there to be 10 messages. When log compaction is enabled, some of the messages between 0 and 10 may have been replaced by tombstones, so they're not there anymore. The metric will still show 10 messages present, because we have no way to know if some of the messages between the beginningOffset and log end offset have been compacted away.

Great, thanks for taking the time to verify that the reflection code works.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/metrics/KafkaOffsetMetric.java
```diff
@@ -0,0 +1,118 @@
...
+            long spoutLag = latestTimeOffset - latestCompletedOffset;
+            long recordsInAssignedPartitions =  latestTimeOffset - earliestTimeOffset;
+
+            String metricPath = topicPartition.topic()  + "/partition_" + topicPartition.partition();
+            result.put(metricPath + "/" + "spoutLag", spoutLag);
+            result.put(metricPath + "/" + "earliestTimeOffset", earliestTimeOffset);
+            result.put(metricPath + "/" + "latestTimeOffset", latestTimeOffset);
+            result.put(metricPath + "/" + "latestEmittedOffset", latestEmittedOffset);
+            result.put(metricPath + "/" + "latestCompletedOffset", latestCompletedOffset);
+            result.put(metricPath + "/" + "recordsInAssignedPartitions", recordsInAssignedPartitions);
```
srdo: Nit: Sorry, now that I look at it the name here is a little weird. Maybe `recordsInPartition`/`totalRecordsInPartitions`?

omkreddy: Updated the PR.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
hmcl: If we start adding a lot of these test methods we would be better off but creating a class in the tests packaged called KafkaSpoutTest that extends KafkaSpout and use that one in the tests. All of these methods should go in this class. WE don't want this class to be very bloated.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/metrics/KafkaOffsetMetric.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    @Override
+    public Object getValueAndReset() {
+
+        Map<TopicPartition, OffsetManager> offsetManagers = offsetManagerSupplier.get();
+        KafkaConsumer kafkaConsumer = consumerSupplier.get();
+
+        if (offsetManagers == null || offsetManagers.isEmpty() || kafkaConsumer == null) {
+            LOG.info("Metrics Tick: offsetManagers or kafkaConsumer is null.");
```
hmcl: Should this be INFO level? Is this going to print this message periodically?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/metrics/KafkaOffsetMetric.java
```diff
@@ -0,0 +1,118 @@
...
+            return null;
+        }
+
+        Map<String,TopicMetrics> topicMetricsMap = new HashMap<>();
+        Set<TopicPartition> topicPartitions = offsetManagers.keySet();
+
+        Map<TopicPartition, Long> beginningOffsets= kafkaConsumer.beginningOffsets(topicPartitions);
+        Map<TopicPartition, Long> endOffsets= kafkaConsumer.endOffsets(topicPartitions);
+        Map<String, Long> result = new HashMap<>();
```
hmcl: it would be useful to have a comment saying what is in this result map

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/metrics/KafkaOffsetMetric.java
```diff
@@ -0,0 +1,118 @@
...
+import java.util.Map;
+import java.util.Set;
+
+public class KafkaOffsetMetric implements IMetric {
+
+    private static final Logger LOG = LoggerFactory.getLogger(KafkaOffsetMetric.class);
+    private final Supplier<Map<TopicPartition, OffsetManager>> offsetManagerSupplier;
+    private final Supplier<KafkaConsumer> consumerSupplier;
+
+    public KafkaOffsetMetric(Supplier offsetManagerSupplier, Supplier consumerSupplier) {
```
hmcl: What's the reasoning behind passing Supplier rather than the actual object?

srdo: The consumer is replaced when activating/deactivating the spout, so we can't just pass the object in here. The alternative is to have a setConsumer method on the metric that the spout then has to call. For the offset manager map, see https://github.com/apache/storm/pull/2480#discussion_r158958224.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/metrics/KafkaOffsetMetric.java
omkreddy: This can be at info level. updated the PR.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/metrics/KafkaOffsetMetric.java
omkreddy: added a comment and added java docs to the classs

srdo:
@omkreddy This LGTM. Could you open a PR against master as well, we usually put features on that branch before merging to the other versions?

omkreddy:
@srdo  Raised PR for master : https://github.com/apache/storm/pull/2495

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -144,10 +147,37 @@ public void open(Map conf, TopologyContext context, SpoutOutputCollector collect
         setCommitMetadata(context);
 
         tupleListener.open(conf, context);
+        if (canRegisterMetrics()) registerMetric();
```
srdo: Nit: Use braces.

omkreddy: Updated the PR.

arunmahadevan:
+1

PR #2466, opened 2017-12-18
STORM-2844: KafkaSpout Throws IllegalStateException After Committing …

hmcl:
…to Kafka When First Poll Strategy Set to EARLIEST

     - Check if commits to Kafka were committed by this topology to correctly enforce FirstPollOffsetStrategy
     - CommitMetadata includes topology id, and task id and threadName

PR #2481, opened 2017-12-26
STORM-2869: Only discard outdated records when adjusting KafkaConsume…

srdo:
…r position during commit

See https://issues.apache.org/jira/browse/STORM-2869

I also deleted SingleTopicKafkaSpoutTest, because it is a duplicate of KafkaSpoutSingleTopicTest.

Handling waitingToEmit like this should also fix some edge cases around partition reassignment while there are pending tuples for revoked partitions in waitingToEmit, but I doubt anyone would actually have been affected by that. We never revoke partitions with the built in subscription classes.


external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/SingleTopicKafkaSpoutTest.java
```diff
@@ -1,379 +0,0 @@
...
-
-            //Nothing is assigned yet, should emit nothing
-            spout.nextTuple();
-            verify(collector, never()).emit(anyString(), anyList(), any(KafkaSpoutMessageId.class));
-
-            SingleTopicKafkaUnitSetupHelper.populateTopicData(kafkaUnitRule.getKafkaUnit(), SingleTopicKafkaSpoutConfiguration.TOPIC, 1);
-            Time.advanceTime(KafkaSpoutConfig.DEFAULT_PARTITION_REFRESH_PERIOD_MS + KafkaSpout.TIMER_DELAY_MS);
-
-            //The new partition should be discovered and the message should be emitted
-            spout.nextTuple();
```
danny0405: Why delete the UT for single topic test?

srdo: It was renamed here, but the old file wasn't deleted https://github.com/apache/storm/pull/2466/files#diff-b827d459a46f189dced7382d2d551439

srdo:
Thanks for the reviews. 1.x version here https://github.com/apache/storm/pull/2484

PR #2482, opened 2017-12-27
STORM-2870 Properly shutdown ExecutorService in FileBasedEventLogger

HeartSaVioR:
* extract local variable 'scheduler' to one of fields
* gracefully shutdown the scheduler

I followed the guide of Javadoc for implementing `closeFlushScheduler()` properly.
https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ExecutorService.html

### DIFF
--- a/storm-client/src/jvm/org/apache/storm/metric/FileBasedEventLogger.java
+++ b/storm-client/src/jvm/org/apache/storm/metric/FileBasedEventLogger.java
@@ -29,8 +29,10 @@ import java.nio.file.StandardOpenOption;
 import java.util.Map;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
 
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import org.apache.storm.task.TopologyContext;
 import org.apache.storm.utils.ConfigUtils;
 import org.slf4j.Logger;
@@ -43,6 +45,7 @@ public class FileBasedEventLogger implements IEventLogger {
 
     private Path eventLogPath;
     private BufferedWriter eventLogWriter;
+    private ScheduledExecutorService flushScheduler;
     private volatile boolean dirty = false;
 
     private void initLogWriter(Path logFilePath) {
@@ -59,8 +62,13 @@ public class FileBasedEventLogger implements IEventLogger {
 
 
     private void setUpFlushTask() {
-        ScheduledExecutorService scheduler = Executors.newSingleThreadScheduledExecutor();
-        Runnable task = new Runnable() {
+        ThreadFactory threadFactory = new ThreadFactoryBuilder()
+                .setNameFormat("event-logger-flush-%d")
+                .setDaemon(true)
+                .build();
+
+        flushScheduler = Executors.newSingleThreadScheduledExecutor(threadFactory);
+        Runnable runnable = new Runnable() {
             @Override
             public void run() {
                 try {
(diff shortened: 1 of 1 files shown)


storm-client/src/jvm/org/apache/storm/metric/FileBasedEventLogger.java
```diff
@@ -59,8 +62,12 @@ private void initLogWriter(Path logFilePath) {
 
 
     private void setUpFlushTask() {
-        ScheduledExecutorService scheduler = Executors.newSingleThreadScheduledExecutor();
-        Runnable task = new Runnable() {
+        ThreadFactory threadFactory = new ThreadFactoryBuilder()
+                .setNameFormat("event-logger-flush-%d")
```
arunmahadevan: should we add `setDaemon(true)` ?

arunmahadevan:
+1 LGTM, left one comment.

storm-client/src/jvm/org/apache/storm/metric/FileBasedEventLogger.java
satishd: By default daemon is set as false, it should be set as true.

storm-client/src/jvm/org/apache/storm/metric/FileBasedEventLogger.java
```diff
@@ -120,8 +127,30 @@ protected String buildLogMessage(EventInfo event) {
...
         }
+
+        closeFlushScheduler();
+    }
+
+    private void closeFlushScheduler() {
+        if (flushScheduler != null) {
+            flushScheduler.shutdown();
+            try {
+                if (!flushScheduler.awaitTermination(1, TimeUnit.SECONDS)) {
```
satishd: nit: You may want to have multiple iterations(which can be 2 or more) with while loop instead of nested if conditionals.

storm-client/src/jvm/org/apache/storm/metric/FileBasedEventLogger.java
HeartSaVioR: @arunmahadevan @satishd 

I intended to set daemon as false, since I would like to provide a chance to shutdown gracefully, via to wait for current running task to be finished instead of killing immediately.
Typically `shutdownNow()` will trigger interrupt and run() method looks like be able to handle interrupt (though it will throw RuntimeException).

If we are in favor of guaranteeing JVM shutdown more than graceful shutdown, I'm happy to  change it.

storm-client/src/jvm/org/apache/storm/metric/FileBasedEventLogger.java
HeartSaVioR: Most of cases in codebase we call shutdownNow() and doesn't wait more. I'll follow the case.

storm-client/src/jvm/org/apache/storm/metric/FileBasedEventLogger.java
satishd: I'm in favor of guaranteeing JVM shutdown in this case as I have seen dangling worker scenarios are hard for users to debug and maintain.

HeartSaVioR: OK, makes sense. Will add `setDaemon(true)`.

HeartSaVioR: Addressed.

HeartSaVioR:
@arunmahadevan @satishd 
Thanks for reviewing. I addressed review comments. Please take a look again. I'll squash commits into one once I got +1. Thanks!

HeartSaVioR:
Thanks. Squashed.

PR #2485, opened 2017-12-28
STORM-2859: Fix a number of issues with NormalizedResources when reso…

srdo:
…urce totals are zero

Issues fixed:
* calculateMinPercentageUsedBy threw division by 0 error if total cpu or memory was 0.
* remove was quietly skipping any resources that was present in only one resource list. I think we should make sure to cover all resources in the input list, so add/remove behave like normal vector addition. 
* Added check that all resources in a rack/node are also present in the total when doing calculateAvg and calculateMin.
* calculateAvg was being inconsistent about the divisor in the average between cpu/memory and other resources. Fixed so resources that are not accounted for in the total are also not counted in the divisor. 
* calculateMin didn't handle resource totals being 0 very well. The method defined `a/0 = 0`, so if any resource in the total had a 0 value, the method would return 0 and all racks/nodes would end up being prioritized equally. Changed calculateMin and calculateAvg to skip any resources where the total of that resource is 0. I think it makes sense to disregard such resources and fall back on prioritizing by any remaining non-0 resources. calculateAvg actually already did this for cpu and memory. In case there are no non-0 resources in the total, we can just return 100% for both average and minimum to disable prioritization, since any rack/node is as good as any other.
* Fixed up NormalizedResources constructor so it doesn't call an overridable method.


storm-server/src/main/java/org/apache/storm/scheduler/resource/NormalizedResources.java
```diff
@@ -184,30 +191,25 @@ public void remove(NormalizedResources other) {
         int otherLength = other.otherResources.length;
         int length = otherResources.length;
         if (otherLength > length) {
-            double [] newResources = new double[otherLength];
+            double[] newResources = new double[otherLength];
             System.arraycopy(newResources, 0, otherResources, 0, length);
             otherResources = newResources;
         }
-        for (int i = 0; i < Math.min(length, otherLength); i++) {
+        for (int i = 0; i < otherLength; i++) {
```
revans2: I think that other may have a longer array, and if that happens then we will get an ArrayIndexOutOfBoundsException.

```
NormalizedResources A = new NormalizedResources({"cpu": 1});
NormalizedResources B = new NormalizedResources({"cpu":1, "foo": 100});

A.remove(B);
```

storm-server/src/main/java/org/apache/storm/scheduler/resource/NormalizedResources.java
```diff
@@ -246,59 +259,133 @@ public boolean couldHoldIgnoringSharedMemory(NormalizedResources other) {
...
-            if (otherResources[i] != 0.0) {
-                total += used.otherResources[i] / otherResources[i];
+        if (LOG.isTraceEnabled()) {
+            LOG.trace("Calculating avg percentage used by. Used CPU: {} Total CPU: {} Used Mem: {} Total Mem: {}"
+                + " Other Used: {} Other Total: {}", totalCpu, used.getTotalCpu(), totalMemory, used.getTotalMemoryMb(),
+                this.toNormalizedOtherResources(), used.toNormalizedOtherResources());
+        }
+
+        if (used.otherResources.length > otherResources.length) {
+            throwBecauseResourceIsMissingFromTotal(used.otherResources.length);
```
revans2: My assumption was that if a resource is not around then it's value is 0.  It looks like in this case, that if a resources is missing an exception will be thrown.  I am not sure that this will not happen in normal operation, or if a user submits a topology with a resource that is not currently supported.

storm-server/src/main/java/org/apache/storm/scheduler/resource/NormalizedResources.java
```diff
@@ -246,59 +259,133 @@ public boolean couldHoldIgnoringSharedMemory(NormalizedResources other) {
...
+        }
+
+        for (int i = 0; i < otherResources.length; i++) {
+            double totalValue = otherResources[i];
+            if (totalValue == 0.0) {
+                //Skip any resources where the total is 0, the percent used for this resource isn't meaningful.
+                //We fall back to prioritizing by cpu, memory and any other resources by ignoring this value
+                skippedResourceTypes++;
```
revans2: This looks fine to me, but could we explain in the javadoc for this method that 0 values are handled differently then might be expected?

storm-server/src/main/java/org/apache/storm/scheduler/resource/NormalizedResources.java
```diff
@@ -246,59 +259,133 @@ public boolean couldHoldIgnoringSharedMemory(NormalizedResources other) {
...
             return 0.0;
         }
-        double min = used.getTotalMemoryMb() / totalMemory;
-        min = Math.min(min, used.getTotalCpu() / getTotalCpu());
+
+        double min = 100.0;
+        if (totalMemory != 0.0) {
+            min = Math.min(min, used.getTotalMemoryMb() / totalMemory);
+        }
+        if (totalCpu != 0.0) {
```
revans2: Again the math looks OK, but could you explain that 0 % used is ignored in the javadocs, because it is not obvious to me that it what will happen.

storm-server/src/main/java/org/apache/storm/scheduler/resource/NormalizedResources.java
srdo: Yes. I'm not sure how we should handle removes if `other` contains values not in `this`. The function was previously just ignoring the "foo" value in B, which I don't think makes that much sense. 

It makes add/remove act unlike regular set addition/subtraction, `A.remove(B).add(B) != A` for example. Behavior is also different between having 0 of a resource, and not having the resource, e.g. if `A = {"cpu": 1, "foo": 0}`, we would instead get an assertion error when removing B.

If we want to actually subtract "foo" from A, we need to allow negative values in the result. Maybe we could delay the no-negative-values check to some other point, so we just disallow negative values in a final result but not in intermediate calculations?

What do you think we should do?

storm-server/src/main/java/org/apache/storm/scheduler/resource/NormalizedResources.java
srdo: I could only find references to this method and calculateMin in the two resource aware strategies, where they seem to be used to estimate how large a percentage of total cluster resources a rack or node represents. That's why I assumed that hitting this case would always be a bug, because the total cluster resources should contain all the resources of the racks.

I'm happy to instead pretend that otherResources has been padded with zeroes, and skip the zero values like we do a few lines down.

storm-server/src/main/java/org/apache/storm/scheduler/resource/NormalizedResources.java
srdo: Yes, I'll add a note to both methods.

storm-server/src/main/java/org/apache/storm/scheduler/resource/NormalizedResources.java
revans2: With where remove is used it should never go negative, so I would be fine with blowing up for all negative values so long as we call it out in the javadocs, and treat something that is not there as a 0.  I agree there is a lot about the code that is not ideal.

storm-server/src/main/java/org/apache/storm/scheduler/resource/NormalizedResources.java
revans2: This is fine, just make sure the java docs explain how this should be used and when it will throw an exception.

srdo:
@revans2 I took another look after some of your last comments, and found some more bugs. I spent some time splitting up NormalizedResources so it was easier to test. Your comments should also have been addressed. 

Most of the classes related to NormalizedResources have been moved to a new package. The move is in the last commit, so it's probably easier to exclude that one from the diff.

If you have an idea for a better place to put the static fields in NormalizedResources, I'd be happy to move them. They are referred to from a lot of different classes, and it would be better if we could create the classes somewhere during startup and inject them into the classes that need them instead of referring to these static fields.

PR #2487, opened 2017-12-29
[STORM-2873] Do not delete backpressure ephemeral node frequently

kishorvpatil:
If ephemeral znode is created once, then we can leave it as is - as other workers would look at timestamp to ensure it is not stale. This avoid deletion/creation of same ephemeral znode path at very high frequency.

### DIFF
--- a/storm-client/src/jvm/org/apache/storm/cluster/StormClusterStateImpl.java
+++ b/storm-client/src/jvm/org/apache/storm/cluster/StormClusterStateImpl.java
@@ -434,11 +434,11 @@ public class StormClusterStateImpl implements IStormClusterState {
      */
     @Override
     public void workerBackpressure(String stormId, String node, Long port, long timestamp) {
-        String path = ClusterUtils.backpressurePath(stormId, node, port);
-        boolean existed = stateStorage.node_exists(path, false);
         if (timestamp <= 0) {
             return;
         }
+        String path = ClusterUtils.backpressurePath(stormId, node, port);
+        boolean existed = stateStorage.node_exists(path, false);
         byte[] data = ByteBuffer.allocate(Long.BYTES).putLong(timestamp).array();
         if (existed) {
             stateStorage.set_data(path, data, acls);
--- a/storm-client/src/jvm/org/apache/storm/cluster/StormClusterStateImpl.java
+++ b/storm-client/src/jvm/org/apache/storm/cluster/StormClusterStateImpl.java
@@ -438,9 +438,7 @@ public class StormClusterStateImpl implements IStormClusterState {
         String path = ClusterUtils.backpressurePath(stormId, node, port);
         boolean existed = stateStorage.node_exists(path, false);
         if (existed) {
-            if (timestamp <= 0) {
-                stateStorage.delete_node(path);
-            } else {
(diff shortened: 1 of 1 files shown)


storm-client/src/jvm/org/apache/storm/cluster/StormClusterStateImpl.java
```diff
@@ -438,9 +438,7 @@ public void workerBackpressure(String stormId, String node, Long port, long time
         String path = ClusterUtils.backpressurePath(stormId, node, port);
         boolean existed = stateStorage.node_exists(path, false);
         if (existed) {
-            if (timestamp <= 0) {
-                stateStorage.delete_node(path);
-            } else {
+            if (timestamp > 0) {
```
srdo: The change seems reasonable to me. The javadoc for this method is slightly out of date now, consider updating the first line of it.

The two outermost branches here do mostly the same thing. Consider rewriting to something like 
```
if (timestamp <= 0) {
  return
}
String path = ...
boolean existed = ...
byte[] data = ...
if (existed) {
  set_data
} else {
  set_ephemeral_node
}
```

kishorvpatil:
@srdo , Please review changes addressing the comments.

storm-client/src/jvm/org/apache/storm/cluster/StormClusterStateImpl.java
```diff
@@ -437,18 +436,14 @@ public void supervisorHeartbeat(String supervisorId, SupervisorInfo info) {
     public void workerBackpressure(String stormId, String node, Long port, long timestamp) {
         String path = ClusterUtils.backpressurePath(stormId, node, port);
         boolean existed = stateStorage.node_exists(path, false);
+        if (timestamp <= 0) {
```
srdo: This check could be moved to the top of the function, if the timestamp is negative there's no reason to check if the znode exists.

srdo:
+1 LGTM. I'm a little confused by when the removed case would be hit though. As far as I can tell it's only called from daemon/worker/Worker.java, and it doesn't look to me like the timestamp parameter can be negative there?

srdo:
Nevermind, timestamp is 0 when the backpressure flag is off. I wasn't paying enough attention :).

HeartSaVioR:
+1 for the change, but only applying to 1.x version line. (-1 for merging in master branch)

We're changing backpressure model in [STORM-2306](https://issues.apache.org/jira/browse/STORM-2306) which got remove of usage of ZK for backpressure, and I think the PR is getting closer to the finish line (target to be included to Storm 2.0.0), hence I don't want to make things conflict.

revans2:
@kishorvpatil could you put up a pull request to 1.x for this?

kishorvpatil:
Ok. I will create the PR for 1.x-branch.

HeartSaVioR:
@kishorvpatil Any updates?

kishorvpatil:
@HeartSaVioR, Please find patch for 1.x-branch - https://github.com/apache/storm/pull/2529

revans2:
@kishorvpatil If we are not going to merge this into master could you close it?

kishorvpatil:
closing the PR as no longer needed for master branch.

PR #2492, opened 2018-01-03
[STORM-2883] Fix storm-kafka trident API bug

dujiashu:
fix fetchmessages offsetoutofrange bug.
https://issues.apache.org/jira/browse/STORM-2883

### DIFF
--- a/external/storm-kafka/pom.xml
+++ b/external/storm-kafka/pom.xml
@@ -57,7 +57,7 @@
                 <artifactId>maven-checkstyle-plugin</artifactId>
                 <!--Note - the version would be inherited-->
                 <configuration>
-                    <maxAllowedViolations>557</maxAllowedViolations>
+                    <maxAllowedViolations>560</maxAllowedViolations>
                 </configuration>
             </plugin>
         </plugins>
--- a/external/storm-kafka/src/jvm/org/apache/storm/kafka/trident/TridentKafkaEmitter.java
+++ b/external/storm-kafka/src/jvm/org/apache/storm/kafka/trident/TridentKafkaEmitter.java
@@ -158,13 +158,8 @@ public class TridentKafkaEmitter {
         return msgs;
     }
 
-    /**re-emit the batch described by the meta data provided.
-     *
-     *
-     * @param attempt
-     * @param collector
-     * @param partition
-     * @param meta
+    /**
+     * re-emit the batch described by the meta data provided
      */
     private void reEmitPartitionBatch(TransactionAttempt attempt, TridentCollector collector, Partition partition, Map meta) {
         LOG.info("re-emitting batch, attempt " + attempt);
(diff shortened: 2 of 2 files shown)


dujiashu:
how can I pass the checkstyle? Thanks for help.

srdo:
@dujiashu Take a look in storm-kafka/target/checkstyle-violations.xml. 

Please also raise an issue for this at https://issues.apache.org/jira, then rename this PR to contain the issue number (see the other active PRs for example). It helps us track bugfixes and lets us generate release notes. Thanks.

dujiashu:
@srdo Thanks for you help.
https://issues.apache.org/jira/browse/STORM-2883

dujiashu:
When tridentAPI, although set config.usestarttimeifoffsetoutofrange true, it would throw offsetoutofrange exception yet.
This issue fix this bug.

srdo:
This breaks the semantics of transactional spouts. Please see http://storm.apache.org/releases/1.1.1/Trident-spouts.html and the sections about transactional and opaque spouts here http://storm.apache.org/releases/1.1.1/Trident-state.html. The latter link describes exactly you case in the example for why transactional spouts should not always be used.

> You might be wondering – why wouldn't you just always use a transactional spout? They're simple and easy to understand. One reason you might not use one is because they're not necessarily very fault-tolerant. For example, the way TransactionalTridentKafkaSpout works is the batch for a txid will contain tuples from all the Kafka partitions for a topic. Once a batch has been emitted, any time that batch is re-emitted in the future the exact same set of tuples must be emitted to meet the semantics of transactional spouts. Now suppose a batch is emitted from TransactionalTridentKafkaSpout, the batch fails to process, and at the same time one of the Kafka nodes goes down. You're now incapable of replaying the same batch as you did before (since the node is down and some partitions for the topic are not unavailable), and processing will halt. 

Basically transactional non-opaque spouts guarantee that a given transaction id always corresponds to the same set of tuples. So for example if txid = 1 contains tuples [A, B, C], then if txid = 1 is reemitted, it has to still contain exactly [A, B, C]. This change will break that guarantee, because if e.g. A, B have been deleted from Kafka, we'll hit the new code and txid = 1 might now contain [C, E, F] instead. 

If you need to be able to handle offsets being deleted from Kafka (which is what you're trying to do with config.usestarttimeifoffsetoutofrange), you need to instead use an Opaque spout (https://github.com/apache/storm/blob/master/external/storm-kafka/src/jvm/org/apache/storm/kafka/trident/OpaqueTridentKafkaSpout.java).

dujiashu:
@srdo Thanks so much. I always think opaquespout is only used to resolve kafka disk broken when not set replication....

srdo:
@dujiashu Happy to help.

If you agree that we shouldn't make this change, would you mind closing this PR? Thanks.

dujiashu:
@srdo Sure, this PR should be closed.

srdo:
@dujiashu I meant could you close it on your end? I think only the author can close PRs at this time.

dujiashu:
@srdo OK. This my first use the github with make PR. I understatnd it now.

PR #2493, opened 2018-01-03
[STORM-2879] Supervisor collapse continuously when there is a expired assignment for overdue storm

danny0405:
We do not make a transaction when supervisor clean up local files for a overdue storm, if an exception occurred during deleting storm-code/ser/jar, an overdue local assignment will be left on disk.

Then when supervisor restart from the exception above, the slots will be initial and container will be recovered from LocalAssignments, the blob store will fetch the files from Nimbus/Master, but will get a KeyNotFoundException, and supervisor collapses again.

So lets make the current assignment null when we first initial a Slot and recover container from local assignment but there are no local needed files, just let the supervisor clean up the local assignment and Nimbus/master will reassign it finally.

This is the JIRA issue [STORM-2879](https://issues.apache.org/jira/browse/STORM-2879).

### DIFF
--- a/storm-server/src/main/java/org/apache/storm/daemon/supervisor/Slot.java
+++ b/storm-server/src/main/java/org/apache/storm/daemon/supervisor/Slot.java
@@ -987,7 +987,18 @@ public class Slot extends Thread implements AutoCloseable, BlobChangingCallback 
         Container container = null;
         if (currentAssignment != null) { 
             try {
-                container = containerLauncher.recoverContainer(port, currentAssignment, localState);
+                // For now we do not make a transaction when removing a topology assignment from local, an overdue
+                // assignment may be left on local disk.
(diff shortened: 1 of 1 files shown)


revans2:
@danny0405 I merged this into master, but the patch does not apply cleanly to 1.x (Slot.java moved locations)

Could you please provide a 1.x patch as well?  Also do you have a JIRA ID that I can assign the JIRA to?  Just want to make sure that you get credit for the fix.

danny0405:
@revans2 This is the JIRA issue: https://issues.apache.org/jira/browse/STORM-2879

HeartSaVioR:
@revans2 I just added @danny0405 to contributor list and assigned to the JIRA issue.
@danny0405 Could you please provide a 1.x patch as well? Thanks in advance!

PR #2494, opened 2018-01-03
STORM-2880 Minor optimisation about kafka spout.

OuYangLiang:
Based on the **single responsibility principle**, method `isAtLeastOnceProcessing()` should reside in `KafkaSpoutConfig` rather than `KafkaSpout`. This patch removes the dependency of `KafkaSpoutConfig.ProcessingGuarantee` from `KafkaSpout`.

### DIFF
--- a/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
+++ b/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
@@ -130,7 +130,7 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
 
         tupleListener = kafkaSpoutConfig.getTupleListener();
 
-        if (isAtLeastOnceProcessing()) {
+        if (kafkaSpoutConfig.isAtLeastOnceProcessing()) {
             // Only used if the spout should commit an offset to Kafka only after the corresponding tuple has been acked.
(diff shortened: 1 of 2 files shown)


srdo:
Bikeshedding incoming:

I'm not sure this is better. The two methods in ProcessingGuarantee are just named variants of equals. I think we might as well just replace the checks in KafkaSpout with == checks. I don't think e.g. `kafkaSpoutConfig.getProcessingGuarantee() == KafkaSpoutConfig.ProcessingGuarantee.AT_MOST_ONCE` is following the SRP any less than `kafkaSpoutConfig.getProcessingGuarantee().isAtMostOnceProcessing()` (the second one also violates the law of Demeter).

OuYangLiang:
@srdo I still think it makes senses, and made a few changes, please review it again, thanks.

OuYangLiang:
@HeartSaVioR thanks, you can't see previous patch because I've squashed to one commit.

srdo:
> e.g. If we rename the enum property AT_LEAST_ONCE to something else, only one file be affected.

I don't think that's a concern, because we're very unlikely to do it. The enum values are part of the public API, so changing their names would be a breaking change.

I don't think this change really makes the code better or more readable, just different. If we wanted to get rid of the utility isAtLeastOnceProcessing method in KafkaSpout, I'd rather we just replace the checks with `kafkaSpoutConfig.getProcessingGuarantee() == KafkaSpoutConfig.ProcessingGuarantee.AT_MOST_ONCE`.

OuYangLiang:
@srdo thanks for the explanation, close this PR.

PR #2500, opened 2018-01-03
STORM-2876: Work around memory leak, and try to speed up tests

revans2:
The actual fix is just the changes in storm-hdfs/pom.xml  But this added 1 min to the build time on a decently powerful box.  So to combat that I made as many of the test run in parallel.  Some of the tests didn't work well in this mode, so I disabled the parallelism for storm-hive, storm-kafka-client, and integration-test.  Storm-hdfs and storm-hive have some issues with where data is stored, and we could probably make it work with the right settings.  storm-kafka-client mostly works, but occasionally I would see a test hang and I didn't have time to debug it.

### DIFF
--- a/external/storm-hdfs/pom.xml
+++ b/external/storm-hdfs/pom.xml
@@ -223,6 +223,14 @@
         <plugins>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+		<configuration>
+                    <reuseForks>false</reuseForks>
+                    <forkCount>1</forkCount>
+                </configuration>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-jar-plugin</artifactId>
                 <version>2.2</version>
                 <executions>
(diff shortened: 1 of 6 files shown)


srdo:
+1

For storm-kafka-client I think the hangs are likely in the tests that run a local Kafka instance. I've had those tests hang if I have a Kafka instance running locally and I forget to close it before running the tests.

The build time for this seems to be about the same as the build time without this change.

HeartSaVioR:
@revans2 Is this an issue only for 2.0.0? If not could you please also provide a patch for 1.x branch? Thanks in advance!

revans2:
@HeartSaVioR I have not seen the issue on 1.x, but it is also an issue there.  It is a problem with using the mini-dfs cluster, and we use it for tests in 1.x too.  The patch applies mostly cleanly to 1.x too.  The only difference is that `<trimStackTrace>false</trimStackTrace>` is not in pom.xml so git didn't know if it should include it or not.  If you are fine with me making that change manually as I merge it in I will not bother with a 1.x pull request.

HeartSaVioR:
@revans2 Sure. I guess small change while porting back has been always acceptable.

revans2:
@HeartSaVioR thanks...

PR #2203, opened 2017-07-11
STORM-2153: New Metrics Reporting API

ptgoetz:
See JIRA for more information:

https://issues.apache.org/jira/browse/STORM-2153

PR #2504, opened 2018-01-05
STORM-2887: store metrics into RocksDB

agresch:
This PR adds a RocksDB key/value database to Nimbus for storing metrics from the Supervisors.  The implementation is replaceable by implementing the MetricStore interface and updating the storm.metricstore.class configuration option.

As metrics are stored in RocksDB, their string values (for topology id and executor Id, etc.) are converted to unique integer Ids, and these strings are also stored to the database as metadata indexed by the integer Id.  When a metric is stored, it is also aggregated with any existing metric within the same 1, 10, and 60 minute timeframe.

Metric.java contains all the relevant metric properties stored in the database.

FilterOptions.java provides all the options available to pass to scan the database for various metrics.

The Supervisor sends metrics (currently only for memory usage) to Nimbus by thrift in processMetrics() in Container.java

Nimbus receives the Supervisor metrics and inserts them in the database in processWorkerMetrics().

RocksDbStore is the RocksDB metric store implementation.

RocksDbKey provides documentation on how metric data and metadata strings are stored as a key in RocksDB.  RocksDbValue provides similar documentation for the value format.

StringMetadataCache provides an in-memory cache of the most recently used string metadata for fast lookup.  The cache size is configurable.

RocksDbMetricsWriter performs the metric writes to RocksDB, taking metrics off the queue and aggregating with existing data, and storing metadata as required.  There's a single thread inserting to prevent threading issues.  Metric database inserts in my tests typically take 80 to 150 microseconds (including the aggregation) for an individual metric.

MetricsCleaner exists to periodically delete old metrics and remove unused metadata strings.  Metric retention time is configurable.

RocksDbStoreTest provide sample code for inserting and searching for metrics.


HeartSaVioR:
I think this patch (or following up patch, at least) should address workers' metrics as well. Is it just waiting for Metrics V2? (#2203) Or to avoid backward incompatibility between old workers?

agresch:
Further metrics can be added to the database as desired.  The intent of this patch was to get the database functional.  Given the patch size and ongoing metrics work, I would suggest adding more metrics as a follow on JIRA.

storm-server/src/main/java/org/apache/storm/DaemonConfig.java
```diff
@@ -1023,6 +1023,43 @@
...
+     * RocksDB setting for length of metric retention.
+     */
+    @isInteger
+    public static final String STORM_ROCKSDB_METRIC_RETENTION_HOURS = "storm.metricstore.rocksdb.retention_hours";
+
+    /**
+     * RocksDB setting for period of metric deletion thread.
+     */
+    @isInteger
+    public static final String STORM_ROCKSDB_METRIC_RETENTION_PERIOD_HOURS = "storm.metricstore.rocksdb.retention_period_hours";
```
revans2: Could we add in deletion to the name of the config somewhere?  I am not sure that it is that clear what the config does from just the name of it.

storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
```diff
@@ -1081,6 +1088,14 @@ public Nimbus(Map<String, Object> conf, INimbus inimbus, IStormClusterState stor
             BlobStore blobStore, TopoCache topoCache, ILeaderElector leaderElector, IGroupMappingServiceProvider groupMapper)
         throws Exception {
         this.conf = conf;
+
+        this.metricsStore = null;
+        try {
+            this.metricsStore = MetricStoreConfig.configure(conf);
+        } catch (Exception e) {
+            LOG.error("Failed to initialize metric store", e);
```
revans2: Could we add in a comment that the metrics store is not critical to the operation of the cluster so if it does not come up we will not stop the cluster from coming up.

storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
```diff
@@ -1081,6 +1088,14 @@ public Nimbus(Map<String, Object> conf, INimbus inimbus, IStormClusterState stor
             BlobStore blobStore, TopoCache topoCache, ILeaderElector leaderElector, IGroupMappingServiceProvider groupMapper)
         throws Exception {
         this.conf = conf;
+
+        this.metricsStore = null;
```
revans2: Can we verify that if the metricsStore is null that we don't get any NullPointerExceptions in really bad places.

storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
```diff
@@ -4166,4 +4184,23 @@ public void shutdown() {
     public boolean isWaiting() {
         return timer.isTimerWaiting();
     }
+
+    @Override
+    public void processWorkerMetrics(WorkerMetrics metrics) throws org.apache.thrift.TException {
+        if (this.metricsStore == null) {
```
revans2: Can we add in some authorization calls before we do anything with executing the command?

```
checkAuthorization(null, null, "processWorkerMetrics");
```

Then we would need to update 

https://github.com/apache/storm/blob/7ecb3d73e8e909c01d39e03a7a7ed45a2fb81859/storm-client/src/jvm/org/apache/storm/security/auth/authorizer/SimpleACLAuthorizer.java#L52

to have processWorkerMetrics in the list.

revans2: It would also be nice to have a metric for the number of times that this method was called, like with the other RPC metrics we have.

storm-server/src/main/java/org/apache/storm/metricstore/MetricStore.java
```diff
@@ -0,0 +1,76 @@
...
+ * under the License.
+ */
+
+package org.apache.storm.metricstore;
+
+import com.codahale.metrics.Meter;
+
+import java.util.Map;
+
+public interface MetricStore {
```
revans2: Can we make this extend AutoCloseable?

storm-server/src/main/java/org/apache/storm/metricstore/rocksdb/RocksDbMetricsWriter.java
```diff
@@ -0,0 +1,306 @@
...
+ * entries are evicted from the cache and need to be written to the database.  This happens as the handleEvictedMetadata()
+ * method callback.
+ */
+public class RocksDbMetricsWriter implements Runnable, AutoCloseable {
+    private static final Logger LOG = LoggerFactory.getLogger(RocksDbMetricsWriter.class);
+    private RocksDbStore store;
+    private BlockingQueue queue;
+    private WritableStringMetadataCache stringMetadataCache;
+    private Set<Integer> unusedIds = new HashSet<>();
+    private TreeMap<RocksDbKey, RocksDbValue> insertBatch = new TreeMap<>(); // RocksDB should insert in sorted key order
```
revans2: I don't think this is being used in a thread safe way.

storm-server/src/main/java/org/apache/storm/metricstore/rocksdb/RocksDbMetricsWriter.java
```diff
@@ -0,0 +1,306 @@
...
+            store.scanRange(firstPrefix, lastPrefix, (key, value) -> {
+                unusedIds.remove(key.getMetadataStringId());
+                return true; // process all metadata
+            });
+        }
+    }
+
+    // writes multiple metric values into the database as a batch operation
+    private void processBatchInsert(Map<RocksDbKey, RocksDbValue> batchMap) throws MetricException {
+        try (WriteBatch writeBatch = new WriteBatch()) {
```
revans2: Because of the thread safety problems with the map, it might be simpler to use the WriteBatch object directly instead of trying to use the map first.

storm-server/src/main/java/org/apache/storm/metricstore/rocksdb/RocksDbStore.java
```diff
@@ -0,0 +1,636 @@
...
+        int cacheCapacity = Integer.parseInt(config.get(DaemonConfig.STORM_ROCKSDB_METADATA_STRING_CACHE_CAPACITY).toString());
+        StringMetadataCache.init(metricsWriter, cacheCapacity);
+        readOnlyStringMetadataCache = StringMetadataCache.getReadOnlyStringMetadataCache();
+        metricsWriter.init(); // init the writer once the cache is setup
+
+        // start threads after metadata cache created
+        Thread thread = new Thread(metricsCleaner, "RocksDbMetricsCleaner");
+        thread.start();
+        thread = new Thread(metricsWriter, "RocksDbMetricsWriter");
+        thread.start();
```
revans2: Should we mark these threads as daemons?  If not we should make sure that they shut down when we close the Store.

storm-server/src/main/java/org/apache/storm/metricstore/rocksdb/RocksDbStore.java
```diff
@@ -0,0 +1,636 @@
...
+
+        if (!config.containsKey(DaemonConfig.STORM_ROCKSDB_METRIC_RETENTION_HOURS)) {
+            throw new MetricException("Not a valid RocksDB configuration - Missing metric retention "
+                    + DaemonConfig.STORM_ROCKSDB_METRIC_RETENTION_HOURS);
+        }
+    }
+
+    private String getRocksDbAbsoluteDir(Map conf) throws MetricException {
+        String storePath = conf.get(DaemonConfig.STORM_ROCKSDB_LOCATION).toString();
+        if (storePath == null) {
```
revans2: storePath should never be null because we called toString on the original object just one line above.  We probably want to cast it to a String instead of calling toString on it.

storm-server/src/main/java/org/apache/storm/metricstore/rocksdb/RocksDbMetricsWriter.java
agresch: The MetricsWriter is a single thread implementation.  RocksDB documentation recommended a single writer for speed, and this simplified the caching of metadata.  If we need to switch to multiple threads, I completely agree the design is not thread safe.

storm-server/src/main/java/org/apache/storm/metricstore/rocksdb/RocksDbMetricsWriter.java
agresch: The reason the insertBatch map is being used to is force sort the keys.  RocksDB indicated sorting would be faster for doing an insert.  The comment on line 54 was trying to imply this information.  Let me know if I can be clearer.

agresch:
Will look into these issues.

HeartSaVioR:
Before looking into the patch in detail, I'd like to see which functionalities the patch would want to bring.

1. As far as I read from MetricStore, it looks like providing an aggregated point from time-range query. Do I understand correctly? Because what I've expected is like a time-series one, and it would replace metrics consumers and eventually provide time-series representation. I expect it can still replace metrics consumers for the store side btw.

2. I guess I know the answer (at least partially) and eventual following-up patch would be storing metrics into external storage (like HBase), but just to double check: how it will behave when leader Nimbus goes down and one of standby Nimbus promotes to leader? Will metrics stored previously be unavailable? If Nimbus gets leadership again, how Nimbus shows the gap while it didn't receive the metrics? (especially aggregated values) Do we want to apply interpolation, or just treat it as no metrics (hence 0 for sum and None for avg)?

HeartSaVioR:
I left the questions because this PR looks like addressing only part of STORM-2156.

The summary of issue STORM-2156 is:

```
Add RocksDB instance in Nimbus and write heartbeat based metrics to it
```

and description is:

```
There should be a RocksDB instance in Nimbus where we write metrics from the heartbeats. This should allow us to replace storage for the statistics we see in the UI and expand the abilities of UIs to allow for time series charting.
Eventually this data will likely come via thrift to Nimbus as the overall metric system is overhauled.
```

(Even STORM-2156 should have follow up issue for representing RocksDB metrics to UI.)

In order to replace the metrics data in ZK heartbeat, it should be mandatory to address worker metrics (to supervisor) to nimbus. I don't mind it would be based on STORM-2693 which transfers worker metrics into Nimbus (addressing Metrics V1), or it would be following up patch for STORM-2153 (Metrics V2) being implementation of metrics collector which communicates to Nimbus. (For latter we should migrate most of built-in metrics to Metrics V2 so that it can be available.)

Actually it only addresses metrics transfer from supervisor to nimbus which doesn't bring origin intention and benefits of issue. I guess you have follow up issues or even patches but they're opaque for me now so I don't see much benefit of this. Please also file follow up issues and group issues together (design doc explaining overall plan would be great) so that we can imagine what will be changed and how the changes improve Storm.

revans2:
@HeartSaVioR sorry about that I think we probably were not clear enough on this, what is covered and what is not currently covered by this pull request.

Reading through the description of STORM-2156 you are 100% right that this does not cover everything there. I will file a new subtask for STORM-2156 and we will update this pull request to be under it.  I am sorry about the confusion.  Thanks for calling us out on this.

For those who care here is some more history about this patch:

Originally this work was based off of the metrics v2 patch and was done by @abellina and @lavindev as an intern here.  But it ended up being a very large patch and it looked like it would take a very long time to go in.  We would have to wait for metrics v2 to go in, then get this reviewed and in along with the glue code that is not here, and then have both of them ported to master.

To try and speed it up I asked @agresch to take out just the rocksdb metrics storage piece and a small number of metrics that don't require the v2 patch, put it on master, make sure it is solid and see if we can get that in by itself.  I thought this would be great because it would at least provide the minimum metrics needed to start looking at elasticity.

The prototype that @lavindev and @abellina includes all of the integration with metrics v2, the UI updated to show both sets of metrics side by side so we could judge how close they were to each other, a real time metrics graphing proof-of-concept on the UI, and a plugin to store the metrics in HBase instead of rocksdb.  So we have code to cover all of STORM-2156 and more.  We are just trying to optimize how quickly we can get it in so we can hopefully do a 2.x release sometime this quarter.

HeartSaVioR:
@revans2 Thanks a lot for sharing current status. Actually I have been aware of the patch going through @abellina forked repo, so I expect HBase plugin and UI patch would be adopted internally, and I also expect most of metrics would be integrated with new system. I thought only metrics V2 is a blocker for your team to bring such features to Apache side, and that’s why I really would want to integrate metrics V2 sooner. I was not aware that they were prototypes instead of production ready, hence also thought it can be included to Storm 2.x sooner (2.0.0 ideally).
If this patch itself is for baseline of other features like elasticity (its plan is also not transparent and I am not aware of that) I think the patch is welcome to get reviewed independently and merged in. I wasn’t just not clear what benefits the patch will provide.

revans2:
@HeartSaVioR not a problem.  I often forget that others are not able to read my thoughts and I sometimes forget to mention the context.

The code was not really a prototype.  We intend it to be production ready, but before pushing anything large like this in we want to harden it with some stress tests.  That was something that we had not done yet.

HeartSaVioR:
@revans2 Great news! Looking forward to new news about that.

agresch:
Updated the PR JIRA to reflect the discussion and new sub-task JIRA.

agresch:
This JIRA currently only addresses metrics on the Nimbus leader.  If Nimbus goes does down and switches to the standby, the options would be to sync the RocksDB database from one server to another, send metrics from the supervisors to both Nimbi (is that the correct pluralism?), or allow metrics to be dropped.  

We plan further work to HBase as an option as well, which would limit the metric loss here.  

I'd recommend dealing with Nimbus standby options in a future JIRA, but welcome discussion.

storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
agresch: Re-verified that a null MetricStore should be safe.

agresch:
Addressed all the code review comments from @revans2 other than the inline comments.

storm-server/src/main/java/org/apache/storm/daemon/supervisor/Slot.java
```diff
@@ -936,6 +936,9 @@ static DynamicState handleRunning(DynamicState dynamicState, StaticState staticS
             }
             dynamicState = dynamicState.withProfileActions(mod, modPending);
         }
+
+        dynamicState.container.processMetrics();
```
revans2: This is something I just noticed.  `handleRunning` should be called about once a second.  I don't think we want to send metrics that frequently.  Most of the metrics are on a 30 second to 1 min frequency.  I think standardizing on 1 min for us is going to make storing and querying them a lot simpler.  Could we have a small amount of state that we can use to throttle this to once a min?

agresch: Yes, will fix.

storm-server/src/main/java/org/apache/storm/metricstore/rocksdb/RocksDbMetricsWriter.java
HeartSaVioR: Maybe better to describe the intention, especially regarding thread-safety.

agresch: I tried to do this on line 43.  I will expand that description and can add thoughts on future multithreading design.

storm-server/src/main/java/org/apache/storm/metricstore/rocksdb/MetricsCleaner.java
```diff
@@ -0,0 +1,101 @@
...
+
+    @Override
+    public void close() {
+        shutdown = true;
+    }
+
+    @Override
+    public void run() {
+        while (true) {
+            if (shutdown) {
```
HeartSaVioR: Is this a reason to have separate if statement instead of `while (!shutdown)`?

storm-server/src/main/java/org/apache/storm/metricstore/rocksdb/RocksDbMetricsWriter.java
```diff
@@ -0,0 +1,323 @@
...
+        this.stringMetadataCache = StringMetadataCache.getWritableStringMetadataCache();
+    }
+
+    /**
+     * Run routine to wait for metrics on a queue and insert into RocksDB.
+     */
+    @Override
+    public void run() {
+        while (true) {
+            if (shutdown) {
```
HeartSaVioR: Same here.

storm-server/src/main/java/org/apache/storm/metricstore/rocksdb/RocksDbStore.java
```diff
@@ -0,0 +1,639 @@
...
+
+    private String getRocksDbAbsoluteDir(Map conf) throws MetricException {
+        String storePath = (String)conf.get(DaemonConfig.STORM_ROCKSDB_LOCATION);
+        if (storePath == null) {
+            throw new MetricException("Not a vaild RocksDB configuration - Missing store location " + DaemonConfig.STORM_ROCKSDB_LOCATION);
+        } else {
+            if (new File(storePath).isAbsolute()) {
+                return storePath;
+            } else {
+                String stormHome = System.getProperty("storm.home");
```
HeartSaVioR: I guess we already have constant for "storm.home".

pom.xml
```diff
@@ -324,6 +324,7 @@
         <jool.version>0.9.12</jool.version>
         <caffeine.version>2.3.5</caffeine.version>
         <jaxb-version>2.3.0</jaxb-version>
+        <rocksdb-version>5.8.6</rocksdb-version>
```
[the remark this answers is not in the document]
agresch: No.  It was the latest version at the time.  Open to advice/suggestions.  5.8.7 and 5.9.2 now also exist.

storm-server/pom.xml
```diff
@@ -64,6 +64,10 @@
             <artifactId>auto-service</artifactId>
             <optional>true</optional>
         </dependency>
+        <dependency>
+            <groupId>org.rocksdb</groupId>
+            <artifactId>rocksdbjni</artifactId>
```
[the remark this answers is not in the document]
agresch: I tested on a mac and a RHEL vm.  The rocksDB jar also contains a win64 jni dll.  If an error is thrown creating the metrics store, everything should be treated as a noop.

agresch:
I'll put up a commit for the remaining issues (or comment if I have further questions).

storm-server/src/main/java/org/apache/storm/metricstore/rocksdb/RocksDbStore.java
agresch: I did not find one.  Can you point out which constant I should use (or where to create one)?

agresch:
Other than creating a storm.home constant, I added the changes you requested.  

I am having trouble getting the html for the new  md file generated to validate the formatting.  I ran "jekyll serve -w" and was able to see an updated link for my page in index.md on the html side, but it did not generate any html for storm-metricstore.md.  I could use some advice on how this is supposed to work.  Thanks.

storm-server/src/main/java/org/apache/storm/metricstore/rocksdb/RocksDbStore.java
HeartSaVioR: Yeah my bad. We didn't have one. Let's leave it as it is since it's a nit.

storm-server/pom.xml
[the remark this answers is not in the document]
HeartSaVioR: Good. We are good to go then.

pom.xml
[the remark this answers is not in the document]
HeartSaVioR: Actually I don't have experience with RocksDB. If you see RocksDB 5.8.6 is running properly, let's just use the version.

HeartSaVioR:
@agresch 
Could you push your documentation to this branch? I could see and try out from my side.

HeartSaVioR:
@agresch 
Never mind. I can see document file. For me, removing first blank line works. Could you try out and see it works, and if it works, could you squash commits into one and let me know?

agresch:
yes, removing the line of space fixed the issue for me.  Thanks.

HeartSaVioR:
@agresch Thanks for handling the issue. Merging.

PR #2507, opened 2018-01-08
STORM-2885: Avoid conflicts with nimbusDaemon LocalCluster Tests

revans2:


storm-server/src/test/java/org/apache/storm/nimbus/LocalNimbusTest.java
```diff
@@ -32,42 +37,35 @@
...
-        topoConf.put("storm.cluster.mode", "local"); // default is aways "distributed" but here local cluster is being used.
-        topoConf.put(Config.STORM_TOPOLOGY_SUBMISSION_NOTIFIER_PLUGIN, InmemoryTopologySubmitterHook.class.getName());
-
-        List<TopologyDetails> topologyNames =new ArrayList<>();
-        for (int i=0; i<4; i++) {
-            final String topologyName = "word-count-"+ UUID.randomUUID().toString();
-            final StormTopology stormTopology = createTestTopology();
-            topologyNames.add(new TopologyDetails(topologyName, stormTopology));
-            localCluster.submitTopology(topologyName, topoConf, stormTopology);
+        int port = Utils.getAvailablePort();
```
srdo: Since this is the only test that uses a Nimbus daemon it's probably not going to cause issues yet, but this way of getting a port is racy. Is there a way to instead get the Nimbus port from the running LocalCluster, and configure the Nimbus daemon to start on port 0?

revans2: Yes it is racy.

We have done it this way for other ports without too many issues.  The problem is that it would require a feedback path for the port from Nimbus, and possibly a few updates in other places that try to read the port out of the config.  If you really want to see this changed I can file a follow on JIRA and look into it, as it would not likely be that much work.

revans2: Oh we also explicitly check that the port is not 0 to avoid having users shoot themselves in the foot and bring up a nimbus that no one can talk to.

srdo: I'd like us to at least file a JIRA to look into changing it, even if we don't do it now. Most of the other uses of Utils.getAvailablePort should have been removed in earlier PRs (e.g. https://github.com/apache/storm/pull/2140). I want to say assigning ports this way was causing test failures on Travis, but it's been a while so I may be misremembering.

revans2: I filed STORM-2886 to address the issue.

srdo: Thanks

PR #2508, opened 2018-01-09
Updated how to get the storm-starter code

aqeelvn:
A typical use case for somebody trying this is, first setting up a working storm installation and then running the example. If the storm-starter code is cloned from github directly, it may not match the version of storm they just installed, and giving rise to an error like `Error: Could not find or load main class org.apache.storm.starter.ExclamationTopology` when the topology is submitted.

The README doesn't make it clear that the storm-starter code is also available inside the version they had downloaded from http://storm.apache.org/downloads.html

Hopefully this edit will bring more clarity.


srdo:
+1

srdo:
Thanks for contributing @aqeelvn. Keep up the good work :)

PR #2509, opened 2018-01-09
Add link to JIRA search showing open newbie labelled issues to contri…

srdo:
…buting page

Seems like a good way to make finding issues a little more convenient for new people.


revans2:
Thanks @srdo I merged this into all the 1.x lines and master.

PR #2510, opened 2018-01-10
[STORM-2892] fix PATH substitution in flux tests

erikdw:
The tests fail when the PATH environment variable has a trailing colon, despite that being a valid PATH.
This happens because it is substituted directly into the resultant YAML file, which results in invalid
YAML (since you cannot end a map's value with ":" in the raw text).

So the solution is to wrap the map value with double-quotes.

Also fix 2 typos in comments in this file.

### DIFF
--- a/flux/flux-core/src/test/resources/configs/substitution-test.yaml
+++ b/flux/flux-core/src/test/resources/configs/substitution-test.yaml
@@ -22,7 +22,7 @@
 name: "${topology.name}"
 
 # Components
-# Components are analagous to Spring beans. They are meant to be used as constructor,
+# Components are analogous to Spring beans. They are meant to be used as constructor,
 # property(setter), and builder arguments.
 #components:
 #  - id: "myComponent"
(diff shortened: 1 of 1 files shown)


erikdw:
@srdo & @revans2 : thx for the quick review-and-merge here!  Follow-up question: since this has been present since the Flux stuff was introduced, it impacts all active branches.  Should I backport and send PRs for each active branch?  I presume the cherry-pick will be clean, so maybe you can just do cherry-picks instead?  Or we just ignore it (I can just skip tests when I do builds, or patch my PATH to avoid the problem).

srdo:
It was my impression that 1.0.x was the oldest active branch. Is this not the case?

erikdw:
@srdo : that is my understanding as well.  Sorry, I didn't realize you backported this to all of the other active branches already, is that done by default for trivial changes like this?

srdo:
@erikdw I don't think there's really a convention. I went by the versions marked affected in the issue. Since it's a bugfix and cherry-picked (mostly) cleanly, it seemed to make sense to put it in all the 1.x.y branches.

If it hard required non-trivial merge work, I wouldn't have pulled it back without a new PR.

erikdw:
@srdo : makes sense, thanks for the info

PR #2514, opened 2018-01-12
STORM-2899: Remove README contributors list

srdo:
Please see https://issues.apache.org/jira/browse/STORM-2899 and the linked mailing list thread.

I'm opening this in the hope that if anyone feels strongly that we should not remove the contributors list, or that we should generate it with Spark's solution, they'll chime in on this. The mailing list thread seemed to fizzle out.


HeartSaVioR:
Merged it without 24hr since it is a change regarding documentation.

PR #2515, opened 2018-01-18
STORM-2900 Always return non null collection for config keys in AbstractHadoopNimbusPluginAutoCreds.

satishd:
STORM-2900 Always return non null collection for config keys in `AbstractHadoopNimbusPluginAutoCreds`.

Closes STORM-2900


danny0405:
+1

arunmahadevan:
+1

PR #2516, opened 2018-01-18
STORM-2900 Always return non null collection for config keys in AbstractAutoCreds.

satishd:
STORM-2900 Always return non null collection for config keys in `AbstractAutoCreds`.
@HeartSaVioR pinging for review.

Respective PR for master: #2515


arunmahadevan:
+1

PR #2519, opened 2018-01-19
STORM-2903: Fix possible NullPointerException in AbstractAutoCreds and doc cleanups

omkreddy:


omkreddy:
@arunmahadevan @HeartSaVioR  
Observed below exception while testing Hive token mechanism. 

```
        at org.apache.storm.common.AbstractAutoCreds.addTokensToUGI(AbstractAutoCreds.java:219) ~[storm-autocreds-1.2.0.3.1.0.0-526.jar:1.2.0.3.1.0.0-526]
        at org.apache.storm.common.AbstractAutoCreds.populateSubject(AbstractAutoCreds.java:118) ~[storm-autocreds-1.2.0.3.1.0.0-526.jar:1.2.0.3.1.0.0-526]
        at org.apache.storm.security.auth.AuthUtils.populateSubject(AuthUtils.java:228) ~[storm-core-1.2.0.3.1.0.0-526.jar:1.2.0.3.1.0.0-526]
        ... 10 more
2018-01-19 16:23:26.157 o.a.s.util main [ERROR] Halting process: ("Error on initialization")
```

external/storm-autocreds/src/main/java/org/apache/storm/common/AbstractAutoCreds.java
```diff
@@ -216,8 +216,7 @@ private void addTokensToUGI(Subject subject) {
                         for (Token<? extends TokenIdentifier> token : allTokens) {
                             try {
                                 LOG.debug("Current user: {}", UserGroupInformation.getCurrentUser());
-                                LOG.debug("Token from credential: {} / {}", token.toString(),
-                                        token.decodeIdentifier().getUser());
+                                LOG.debug("Extracted Token from Credentials : {}", token);
```
satishd: token.toString() will not print token.decodeIdentifier().getUser()) which may be useful in debugging. `UserGroupInformation.getCurrentUser().addToken(token)` ignores null tokens.
Better to log and skip when `token` is null like below instead of changing the current log statements. 
```
    for (Token<? extends TokenIdentifier> token : allTokens) {
        try {
            if(token == null) {
                LOG.debug("Ignoring null token");
                continue;
            }
            LOG.debug("Current user: {}", UserGroupInformation.getCurrentUser());
            LOG.debug("Token from credential: {} / {}", token.toString(),
                    token.decodeIdentifier().getUser());

            UserGroupInformation.getCurrentUser().addToken(token);
            LOG.info("Added delegation tokens to UGI.");
        } catch (IOException e) {
            LOG.error("Exception while trying to add tokens to ugi", e);
        }
    }

```

arunmahadevan:
+1, Thanks for the patch. Can you also associate it with a JIRA and raise a patch for master branch as well?

external/storm-autocreds/src/main/java/org/apache/storm/common/AbstractAutoCreds.java
arunmahadevan: It appears that the token.toString() does some decoding so might print the user info.

omkreddy:
@satishd  @arunmahadevan Actually exception is from token.decodeIdentifier(). decodeIdentifier can return null. Updated the PR with review comments.

omkreddy:
PR for master: https://github.com/apache/storm/pull/2520

external/storm-autocreds/src/main/java/org/apache/storm/common/AbstractAutoCreds.java
```diff
@@ -215,9 +215,17 @@ private void addTokensToUGI(Subject subject) {
...
+                                    LOG.debug("Ignoring null token");
+                                    continue;
+                                }
+
                                 LOG.debug("Current user: {}", UserGroupInformation.getCurrentUser());
-                                LOG.debug("Token from credential: {} / {}", token.toString(),
-                                        token.decodeIdentifier().getUser());
+                                LOG.debug("Token from Credentials : {}", token);
+
+                                if (token.decodeIdentifier() != null)
```
arunmahadevan: @omkreddy , toString already handles it and just printing the token would take care of decoding and printing the user information.
https://github.com/apache/hadoop/blob/trunk/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/security/token/Token.java#L429

omkreddy:
@arunmahadevan thanks for the review. yes, printing token should be sufficient.  reverting PR to first version

arunmahadevan:
+1

satishd:
@arunmahadevan @omkreddy I have looked into that code before putting my initial comment. Not all `TokenIdentifier` implementations have implemented `toString()` with user information. `org.apache.hadoop.hbase.security.token.AuthenticationTokenIdentifier` does not implement toString() and the logging statement was added while fixing STORM-2555 with the change [here](https://github.com/apache/storm/commit/2544b6d99f163cd2552485a629651f8e97fff8ee#diff-a81e93604bd2f50eae45a7a3bee59d88) which is related to HBase delegation tokens.

HeartSaVioR:
@omkreddy @arunmahadevan @satishd 

My bad I forgot what I was doing. Yes that was why I got user information from `token.decodeIdentified().getUser()`.

Do we want to revert the commit, or craft a follow-up patch?

omkreddy:
@satishd  @HeartSaVioR  Unfortunately token.decodeIdentified() return null value in Hive related Tokens.   So if we want this log,  we need to add null check before printing token.decodeIdentified()..getUser()

satishd:
>@satishd @HeartSaVioR Unfortunately token.decodeIdentified() returns null value in Hive related Tokens. So previous code is broken for Hive. If we want this log, we need to add null check before printing token.decodeIdentified()..getUser()

@omkreddy No harm in having null check and have good debug information in the logs. 

>Do we want to revert the commit, or craft a follow-up patch?

@HeartSaVioR Better to have a followup patch.

PR #2520, opened 2018-01-19
STORM-2903: Fix possible NullPointerException in AbstractHadoopAutoCreds and doc cleanups

omkreddy:


external/storm-autocreds/src/main/java/org/apache/storm/common/AbstractHadoopAutoCreds.java
```diff
@@ -113,10 +113,14 @@ private void addTokensToUGI(Subject subject) {
                     if (allTokens != null) {
                         for (Token<? extends TokenIdentifier> token : allTokens) {
                             try {
+                                if (token == null) {
+                                    LOG.debug("Ignoring null token");
+                                    continue;
+                                }
```
danny0405: So in what scenario the token can be null ?

PR #2521, opened 2018-01-22

hmcc:
Connected to https://issues.apache.org/jira/browse/STORM-2862; 1.1.x version of apache/storm#2475.


HeartSaVioR:
+1 prior to merge. (It's kind of porting back.)

PR #2524, opened 2018-01-22
STORM-2903: (follow up) Log token identifier and username associated with token identifier

omkreddy:


omkreddy:
@HeartSaVioR @satishd  follow-up PR for STORM-2903

PR #2526, opened 2018-01-22
STORM-2904: Document Metrics V2

ptgoetz:
Documentation only. No code changes.


docs/index.md
```diff
@@ -80,7 +80,7 @@ But small change will not affect the user experience. We will notify the user wh
 * [Distributed RPC](Distributed-RPC.html)
 * [Transactional topologies](Transactional-topologies.html)
 * [Hooks](Hooks.html)
-* [Metrics](Metrics.html)
+* [Metrics](metrics_v2.html)
```
HeartSaVioR: Maybe keeping deprecated metrics doc would help, though we would want to be clear that they're deprecated. 
There's explanation regarding built-in metrics in previous metrics doc, but it may look odd to get built-in metrics into metrics V2 doc, because only small set of metrics are available in metrics V2 while we don't migrate them. Users still need metrics V1 for them.

ptgoetz:
@HeartSaVioR Fixed.

PR #2527, opened 2018-01-23
STORM-2906 Pick HBase delegation token properly while handling HBase …

HeartSaVioR:
…auth via delegation token

Please refer [STORM-2906](https://issues.apache.org/jira/browse/STORM-2906) to see more details.

This can be clean cherry-pick against 1.x version lines.

### DIFF
--- a/external/storm-hbase/src/main/java/org/apache/storm/hbase/common/Utils.java
+++ b/external/storm-hbase/src/main/java/org/apache/storm/hbase/common/Utils.java
@@ -57,12 +57,17 @@ public class Utils {
                 // which ensures toString() implementation
                 if (token.getKind().toString().equals(TOKEN_KIND_HBASE_AUTH_TOKEN)) {
                     // use UGI from token
-                    LOG.debug("Found HBASE_AUTH_TOKEN - using the token to replace current user.");
+                    if (!foundHBaseAuthToken) {
+                        LOG.debug("Found HBASE_AUTH_TOKEN - using the token to replace current user.");
 
-                    ugi = token.decodeIdentifier().getUser();
-                    ugi.addToken(token);
+                        ugi = token.decodeIdentifier().getUser();
+                        ugi.addToken(token);
 
-                    foundHBaseAuthToken = true;
+                        foundHBaseAuthToken = true;
+                    } else {
+                        LOG.warn("Found multiple HBASE_AUTH_TOKEN - will use already found token. " +
+                                "Please enable DEBUG log level to track delegation tokens.");
+                    }
                 }
             }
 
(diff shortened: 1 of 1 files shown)


external/storm-hbase/src/main/java/org/apache/storm/hbase/common/Utils.java
```diff
@@ -47,14 +48,28 @@ public static HTable getTable(UserProvider provider, Configuration config, Strin
...
+                // token.getKind() = Text, Text is annotated by @Stringable
+                // which ensures toString() implementation
+                if (token.getKind().toString().equals(TOKEN_KIND_HBASE_AUTH_TOKEN)) {
+                    // use UGI from token
+                    LOG.debug("Found HBASE_AUTH_TOKEN - using the token to replace current user.");
+
+                    ugi = token.decodeIdentifier().getUser();
+                    ugi.addToken(token);
+
+                    foundHBaseAuthToken = true;
```
satishd: @HeartSaVioR Can there be more than one HBASE_AUTH_TOKEN token for the current UGI?

HeartSaVioR: I don't think if there's one HBase cluster needed to be connected in worker, but not sure if there're more than one HBase clusters. If we need to consider and handle multiple tokens, matching token would be completely different scenario.

satishd: I am trying to understand the current limitations. Can a worker support two or more HBase bolts talk to respective HBase clusters using delegation tokens?

HeartSaVioR: I didn't test it actually, but looks like we don't support it, since we are only using **one** of tokens. Not sure HDFS/Hive bolt supports it, too.

satishd: @HeartSaVioR There should not be multiple tokens with HBASE_AUTH_TOKEN key with the current implementation. It is better to log a message(atleast with info level) if we find multiple tokens so that it is easy to debug such scenarios.

HeartSaVioR: Makes sense. I'll update the patch.

HeartSaVioR: @satishd Addressed.

satishd:
+1 
Thanks @HeartSaVioR for addressing review comments.

PR #2530, opened 2018-01-23
STORM-2907: Exclude curator dependencies from storm-core in storm-autocreds pom

arunmahadevan:
storm-autocreds brings in the curator 4.0 jars via transitive dependency of storm-core. Even though storm-core is listed as provided scope, the app assembler plugin puts the dependency (curator 4.0) into external/storm-autocreds directory. This conflicts with the storm-druid tranquility library which depends on curator 2.6.0. 

There could be other conflicts if storm-autocreds is included in class path. 

Excluding curator dependency from storm-core resolves this.

### DIFF
--- a/external/storm-autocreds/pom.xml
+++ b/external/storm-autocreds/pom.xml
@@ -40,6 +40,18 @@
                     <groupId>org.slf4j</groupId>
                     <artifactId>log4j-over-slf4j</artifactId>
                 </exclusion>
+                <exclusion>
+                    <groupId>org.apache.curator</groupId>
+                    <artifactId>curator-client</artifactId>
+                </exclusion>
+                <exclusion>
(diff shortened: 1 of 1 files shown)


hmcl:
+1, however I don't understand why this module is using the maven app assembler plugin to copy all the jars under target/app-assembler/repo nor why all of these jars are posteriorly being copied by binary.xml onto external/storm-autocreds.

It seems to me that the root cause to this problem is not the transitive dependency, but rather the way these scripts are copying jars around.

arunmahadevan:
This is so that users could include the "external/storm-autocreds" directory into their class path and have the delegation token mechanism for the storm hdfs, hbase and hive connectors work out of the box.

hmcl:
OK, thanks.
+1

PR #2433, opened 2017-11-22
[STORM-2693] Heartbeats and assignments promotion for storm2.0

danny0405:
For large cluster support, good performance for scheduling and assignments distribution.

**For nimbus heartbeats pressure test**
Heartbeats response time:
e4a1e216-0a08-4e08-90c0-334e00db962c
The horizontal ordinate is number of nodes/supervisors which report heartbeats every 1 second, for our production cluster, heartbeat reporting interval is 5 seconds and time out 30 seconds, so this means new heartbeats mode supports at least 5 * 2000 nodes.

**For topology submission time cost**
Topology submission scheduling time:

cluster used slots | newly submit workers | after submission used slots | time cost milliseconds
-- | -- | -- | --
3700 | 100 | 3800 | 1181ms
3600 | 100 | 3700 | 886ms
3500 | 100 | 3600 | 925ms
3400 | 100 | 3500 | 1520ms
3300 | 100 | 3400 | 930ms
2000 | 100 | 2100 |  722ms
1500 | 100 | 1600 |  747ms
1000 | 100 | 1100 |  754ms
500 | 100 | 600 |  695ms
100 | 100 | 200 |  749ms

This is PR for 1.1.x-branch of heartbeats and assignments promotion
#2389

This is PR for 1.1.x-branch of assignments promotion [closed now]
#2319

PR #2531, opened 2018-01-24
STORM-2898: Support for WorkerToken authentication

revans2:
This adds in support for workers to be able to authenticate with either nimbus or drpc invocation servers using WorkerTokens.

A WorkerToken is a lot like a delegation token in Hadoop.  See the jira for the design of this.

Some of the things I did in this beyond just worker tokens.

1. I ported auth-test to java (STORM-1301).  Because I wanted to reuse the code there to verify that this feature was working properly, and I didn't want to leave it in clojure.
2. I refactored how the zookeeper ACLs are passed into the StormClusterStateImpl.  They now are done from methods in DaemonType because I though it made the code cleaner, and provided a good place to add in the new ACLs needed to secure WorkerToken private keys that need to be shared in one case among all nimbus instances and in another case with the DRPC servers.
3. I refactored the digest authentication sasl code.  I did this to make it simpler to add in support for tokens that also use the same method and get a more code reuse.
4. I deprecated the sasl plain auth transport plugin.  It was not being used by any tests.  It is totally insecure in so many ways, and I didn't want anyone to mistakenly think it was OK to use it.  If others are using it for some of their testing internally I can remove the deprecation.
5. I cleaned up some check style violations for code that I was touching.

I know this is a lot of code, so thanks to anyone who can help review it.


storm-client/src/jvm/org/apache/storm/cluster/StormClusterStateImpl.java
```diff
@@ -744,19 +767,111 @@ public Credentials credentials(String stormId, Runnable callback) {
...
+                    } catch (RuntimeException e) {
+                        //This should never happen because only the primary nimbus is active, but just in case
+                        // declare the race safe, even if we lose it.
+                        if (!Utils.exceptionCauseIsInstanceOf(KeeperException.NoNodeException.class, e)) {
+                            throw e;
+                        }
+                    }
+                }
+            } catch (RuntimeException e) {
+                //No node for basePath is OK, noting to remove
```
revans2: Spelling.  s/noting/nothing/

HeartSaVioR:
@revans2 Thanks a lot for putting great efforts to craft a patch so quickly! I'll take a look after verification of release candidates on 1.x version lines in progress.

Btw, in #2433 we would also want to let worker communicate to Supervisor so that the path of passing worker heartbeats is worker -> supervisor -> nimbus for normal case and worker -> nimbus for supervisor down. From quick look there's no supervisor token.

Do we want to remove supervisor in path and let worker always communicate nimbus? Or we keep leveraging local storage for worker -> supervisor? Or nimbus token will also allow accessing to supervisor?

HeartSaVioR:
I just forgot you don't think worker -> supervisor heartbeat passing via RPC is mandatory, hence there's no need to have worker -> supervisor token. Please let me know if I'm missing here.

storm-client/src/jvm/org/apache/storm/security/auth/kerberos/KerberosSaslTransportPlugin.java
```diff
@@ -114,22 +120,26 @@ public TTransportFactory getServerTransportFactory() throws IOException {
         //check the credential of our principal
         if (subject.getPrivateCredentials(KerberosTicket.class).isEmpty()) { 
             throw new RuntimeException("Fail to verify user principal with section \""
-                    +AuthUtils.LOGIN_CONTEXT_SERVER+"\" in login configuration file "+ login_conf);
+                    +AuthUtils.LOGIN_CONTEXT_SERVER+"\" in login configuration file "+ loginConf);
```
kishorvpatil: spacing..

storm-client/src/jvm/org/apache/storm/security/auth/kerberos/ServerCallbackHandler.java
```diff
@@ -18,79 +18,86 @@
...
-    public ServerCallbackHandler(Configuration configuration, Map<String, Object> topoConf) throws IOException {
-        if (configuration==null) return;
+    public ServerCallbackHandler(Configuration configuration) throws IOException {
+        if (configuration == null) {
+            return;
+        }
 
         AppConfigurationEntry configurationEntries[] = configuration.getAppConfigurationEntry(AuthUtils.LOGIN_CONTEXT_SERVER);
         if (configurationEntries == null) {
             String errorMessage = "Could not find a '"+AuthUtils.LOGIN_CONTEXT_SERVER+"' entry in this configuration: Server cannot start.";
```
kishorvpatil: spacing.

danny0405:
@HeartSaVioR 
BTW, worker needs to fetch assignment as a timer task to sync connections.

revans2:
@HeartSaVioR I didn't implement worker->supervisor tokens here, because for this patch there is no supervisor server yet.  #2433 adds in the supervisor server, and as such would add in the changes needed to make it work with tokens too.

I just wrote up a design proposal for storm on X at [STORM-1740](https://issues.apache.org/jira/browse/STORM-1740) that would also have some impact on it.

As such once this goes in I would be happy to make a pull request to #2433 to add in support for the supervisor tokens too.

storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
```diff
@@ -2638,6 +2647,34 @@ private int getNumOfAckerExecs(Map<String, Object> totalConf, StormTopology topo
         }
     }
 
+    private void upsertWorkerTokensInCreds(Map<String, String> creds, String user, String topologyId) {
+        if (workerTokenManager != null) {
```
danny0405: updateWorkerTokensInCreds ?

storm-server/src/test/java/org/apache/storm/security/auth/AuthTest.java
```diff
@@ -0,0 +1,637 @@
...
+    @Test(expected = RuntimeException.class)
+    public void getTransportPluginThrowsRunimeTest() {
+        Map<String, Object> conf = ConfigUtils.readStormConfig();
+        conf.put(Config.STORM_THRIFT_TRANSPORT_PLUGIN, "null.invalid");
+        AuthUtils.getTransportPlugin(ThriftConnectionType.NIMBUS, conf, null);
+    }
+
+    public static ReqContext mkImpersonatingReqContext(String impersonatingUser, String userBeingIUmpersonated, InetAddress remoteAddress) {
+        ReqContext ret = new ReqContext(mkSubject(userBeingIUmpersonated));
+        ret.setRemoteAddress(remoteAddress);
```
danny0405: userBeingImpersonated ?

storm-core/test/clj/org/apache/storm/nimbus_test.clj
```diff
@@ -46,7 +46,7 @@
   (:import [org.apache.commons.io FileUtils])
   (:import [org.json.simple JSONValue])
   (:import [org.apache.storm.daemon StormCommon])
-  (:import [org.apache.storm.cluster IStormClusterState StormClusterStateImpl ClusterStateContext ClusterUtils])
+  (:import [org.apache.storm.cluster DaemonType IStormClusterState StormClusterStateImpl ClusterStateContext ClusterUtils])
```
kishorvpatil: Harmless, but don't see tests using `DaemonType` in this file.

storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
```diff
@@ -2638,6 +2647,34 @@ private int getNumOfAckerExecs(Map<String, Object> totalConf, StormTopology topo
...
+                            // token.
+                            shouldAdd = false;
+                        }
+                    } catch (Exception e) {
+                        //The old token could not be deserialized.  This is bad, but we are going to replace it anyways so just keep going.
+                        LOG.error("Could not deserialize token info", e);
+                    }
+                }
+                if (shouldAdd) {
+                    AuthUtils.setWorkerToken(creds, workerTokenManager.createOrUpdateTokenFor(type, user, topologyId));
```
kishorvpatil: Not necessary, but some info level logs mentioning update to the worker tokens might help.

storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
revans2: from https://en.wikipedia.org/wiki/Merge_(SQL)
"A relational database management system uses SQL MERGE (also called upsert) statements to INSERT new records or UPDATE existing records depending on whether condition matches."

storm-server/src/test/java/org/apache/storm/security/auth/AuthTest.java
revans2: In SASL there is the concept of a user acting on behalf of another user, or impersonating them.

We use this with the UI.

The UI user will make a connection to nimbus and authenticate as itself, but as part of the request say that it is acting on behalf of the actual user that is looking at the page.  That way nimbus can verify not only that the UI user is allowed to impersonate other users, but also that the user it is impersonating has the right permissions to do what the UI is trying to do.

storm-core/test/clj/org/apache/storm/nimbus_test.clj
revans2: Will fix it.

storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
revans2: I think the WorkerTokenManager will log it when the token is generated.

storm-client/src/jvm/org/apache/storm/security/auth/workertoken/WorkerTokenAuthorizer.java
```diff
@@ -0,0 +1,139 @@
...
+ */
+public class WorkerTokenAuthorizer implements PasswordProvider {
+    private static final Logger LOG = LoggerFactory.getLogger(WorkerTokenAuthorizer.class);
+
+    private static IStormClusterState buildStateIfNeeded(Map<String, Object> conf, ThriftConnectionType connectionType) {
+        IStormClusterState state = null;
+
+        if (AuthUtils.areWorkerTokensEnabledServer(connectionType, conf)) {
+            try {
+                state = ClusterUtils.mkStormClusterState(conf, new ClusterStateContext(DaemonType.UNKNOWN, conf));
```
agresch: not familiar with these API.  Why are we using UNKNOWN?

storm-client/src/jvm/org/apache/storm/security/auth/workertoken/WorkerTokenAuthorizer.java
```diff
@@ -0,0 +1,139 @@
...
+        if (keyCache == null) {
+            return Optional.empty();
+        }
+        try {
+            byte[] user = Base64.getDecoder().decode(userName);
+            WorkerTokenInfo deser = Utils.deserialize(user, WorkerTokenInfo.class);
+            byte[] password = getSignedPasswordFor(user, deser);
+            return Optional.of(Base64.getEncoder().encodeToString(password).toCharArray());
+        } catch (Exception e) {
+            LOG.debug("Could not decode {}, might just be a plain digest request...", userName, e);
```
agresch: The token exceptions above aren't worth an info line if they occur?

storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
```diff
@@ -2254,14 +2262,13 @@ private void renewCredentials() throws Exception {
...
         Collection<ICredentialsRenewer> renewers = credRenewers;
-        Object lock = credUpdateLock;
         Map<String, StormBase> assignedBases = state.topologyBases();
         if (assignedBases != null) {
             for (Entry<String, StormBase> entry: assignedBases.entrySet()) {
                 String id = entry.getKey();
                 String ownerPrincipal = entry.getValue().get_principal();
                 Map<String, Object> topoConf = Collections.unmodifiableMap(Utils.merge(conf, tryReadTopoConf(id, topoCache)));
-                synchronized(lock) {
+                synchronized(credUpdateLock) {
```
agresch: was this just wrong?  Does it need update in other releases?

storm-client/src/jvm/org/apache/storm/cluster/IStormClusterState.java
```diff
@@ -140,9 +143,65 @@
...
+     * @param topologyId the id of the topology to scan.
+     */
+    void removeExpiredWorkerKeys(String topologyId);
+
+    /**
+     * Remove all of the worker keys for a given topology.  Used to clean up after a topology finishes.
+     * This is expected to only ever be called from nimbus and ideally should only ever work from nimbus.
+     * @param topologyId the topology to clean up after.
+     */
+    void removeWorkerAllKeys(String topologyId);
```
Ethanlm: Would `removeAllWorkerKeys` be a little better?

storm-client/src/jvm/org/apache/storm/cluster/IStormClusterState.java
```diff
@@ -140,9 +143,65 @@
...
+
+    /**
+     * Get a list of all topologyIds that currently have private worker keys stored, of any kind.
+     * This is expected to only ever be called from nimbus.
+     * @return the list of topology ids with any kind of private worker key stored.
+     */
+    Set<String> workerTokenTopologyKeys();
```
Ethanlm: This is not a big deal but I am a little confused by this function name. Something like `idsOfTopologyWithPrivateWorkerKeys` ?

storm-client/src/jvm/org/apache/storm/security/auth/AuthUtils.java
```diff
@@ -264,6 +269,128 @@ public static IGroupMappingServiceProvider GetGroupMappingServiceProviderPlugin(
...
+     * @return true if we can enable them, else false.
+     */
+    public static boolean areWorkerTokensEnabledServer(ThriftServer server, Map<String, Object> conf) {
+        return server.supportsWorkerTokens() && willWorkerTokensBeStoredSecurely(conf);
+    }
+
+    /**
+     * Check if worker tokens should be enabled on the server side or not (for a given server).
+     * @param connectionType the type of server this is for.
+     * @param conf the daemon configuration top be sure the tokens are secure.
```
Ethanlm: should `top` --> `to` ?

storm-client/src/jvm/org/apache/storm/security/auth/kerberos/KerberosSaslTransportPlugin.java
```diff
@@ -194,15 +223,15 @@ public TTransport connect(TTransport transport, String serverHost, String asUser
         final Subject subject = login.getSubject();
         if (subject.getPrivateCredentials(KerberosTicket.class).isEmpty()) { //error
             throw new RuntimeException("Fail to verify user principal with section \""
-                        +AuthUtils.LOGIN_CONTEXT_CLIENT+"\" in login configuration file "+ login_conf);
+                        +AuthUtils.LOGIN_CONTEXT_CLIENT+"\" in login configuration file "+ loginConf);
```
Ethanlm: spacing

storm-client/src/jvm/org/apache/storm/security/auth/workertoken/WorkerTokenSigner.java
```diff
@@ -0,0 +1,64 @@
...
+
+    /**
+     * Compute HMAC of the identifier using the secret key and return the
+     * output as password.
+     * @param identifier the bytes of the identifier
+     * @param key the secret key
+     * @return the bytes of the generated password
+     */
+    static byte[] createPassword(byte[] identifier,
+                                           SecretKey key) {
```
Ethanlm: Is the function signature in one line better?

storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
danny0405: Okey，got your idea

storm-client/src/jvm/org/apache/storm/security/auth/workertoken/WorkerTokenAuthorizer.java
revans2: We are using UNKNOWN because the DaemonType is only used for setting up ACLs when writes happen, and we are not writing.  It would be cleaner to have a ReadOnly ClusterState API, and if you want me to do that I am happy to.

storm-client/src/jvm/org/apache/storm/security/auth/workertoken/WorkerTokenAuthorizer.java
revans2: The issue is because in some cases we have to support both a regular DIGEST-MD5 and a token DIGEST-MD5.  In the regular case the user name and password have no special translation.  They are just read from a file.  In the token case they are translated.

The way we distinguish between the two is to try and deserialize it here, and if it works then it is a token, else assume it is not a token.

storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
revans2: It wasn't wrong, we just were copying the lock into a local variable to use it once.  That felt needless, and made it more difficult to find where credUpdateLock was actually used.

revans2:
I think I have addressed the review comments.  Please take a look again.  If things look good I will rebase and squash before merging.

storm-client/src/jvm/org/apache/storm/cluster/IStormClusterState.java
```diff
@@ -184,21 +184,21 @@
...
-    void removeWorkerAllKeys(String topologyId);
+    void removeAllPrivateWorkerKeys(String topologyId);
 
     /**
      * Get a list of all topologyIds that currently have private worker keys stored, of any kind.
      * This is expected to only ever be called from nimbus.
      * @return the list of topology ids with any kind of private worker key stored.
      */
-    Set<String> workerTokenTopologyKeys();
+    Set<String> idsOfTopologysWithPrivateWorkerKeys();
```
HeartSaVioR: nit: Topologies

revans2:
I will fix the nit, but I was also talking to some security people and I am going to switch the signature algorithm from "HmacSHA1" to "HmacSHA256".  Not because there are any issues with SHA1, this will just make it a bit more future proof.  Also to take full advantage of a 256bit hash I am going to up the key size to 256 bits as well.

I'll make the changes squash it and put it up again, just to be sure everyone is OK with the final code.

revans2:
The changes are done.  The build passes when I disable check style, and check style passes for the sub-projects that I modified.

If I can get a final +1 on renaming the one Method that @HeartSaVioR found and also for moving to SHA256 that would be great and I will merge it in.

PR #2532, opened 2018-01-25
STORM-2912 Revert optimization of sharing tick tuple (1.x)

HeartSaVioR:
* since it incurs side effect and messes metrics

More detail: https://issues.apache.org/jira/browse/STORM-2912

We must cherry-pick this into 1.1.x-branch and 1.0.x-branch as well, and I guess it will be no conflict.

Will also craft a patch for master branch shortly.

### DIFF
--- a/storm-core/src/clj/org/apache/storm/daemon/executor.clj
+++ b/storm-core/src/clj/org/apache/storm/daemon/executor.clj
@@ -361,13 +361,15 @@
               (and (= false (storm-conf TOPOLOGY-ENABLE-MESSAGE-TIMEOUTS))
                    (= :spout (:type executor-data))))
         (log-message "Timeouts disabled for executor " comp-id ":" (:executor-id executor-data))
-        (let [val [(AddressedTuple. AddressedTuple/BROADCAST_DEST (TupleImpl. context [tick-time-secs] Constants/SYSTEM_TASK_ID Constants/SYSTEM_TICK_STREAM_ID))]]
-          (schedule-recurring
-            (:user-timer worker)
-            tick-time-secs
-            tick-time-secs
-            (fn []
-                (disruptor/publish receive-queue val))))))))
+        (schedule-recurring
+          (:user-timer worker)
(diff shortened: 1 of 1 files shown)


HeartSaVioR:
It is really hard to think such side effect while applying optimization like this. It is fairly easy to think that tuple is immutable, and it is true for interface `Tuple`, but no longer true for `TupleImpl`. Ideally `TupleImpl` should be also immutable. 

Due to how acker works, we allow updating XOR value to `TupleImpl`, and due to how we measure metrics, we allow setting timestamp to `TupleImpl`. 

It would be better if we have some way to address without incurring performance hit.

satishd:
@HeartSaVioR Was there a reasonable perf hit because of creating tick tuple instance whenever it is executed in timer?

HeartSaVioR:
@satishd I don't know about the perf hit, and I don't think it doesn't bring actual perf hit, but anyone would think it is just weird to create same tuple every time, so easy to break it again.

satishd:
@HeartSaVioR It may be helpful to have a comment on why it was changed with JIRA number so that same change is avoided later.

danny0405:
@HeartSaVioR 
Create an object is very lightweight, is should not have any performance hit.

HeartSaVioR:
@satishd @danny0405
Sorry guys. I missed while commenting, corrected now. **I don't think it brings any perf hit.** I wrote it opposite way. ;)

HeartSaVioR:
@danny0405
If it's in critical path, it could bring perf. hit, not only from creating object but also from GC. We are targeting to emit millions of tuples in a second. 
In this case, given that time unit for recurring is second, I'm sure that we don't have any perf. hit.

HeartSaVioR:
@satishd added comments. Commit amended because it is tiny change.

ptgoetz:
+1 Nice catch. Agree on performance -- tick tuples are sent infrequently enough that the optimization isn't necessary.

arunmahadevan:
+1

PR #2534, opened 2018-01-25
STORM-2910: Fix metrics reporting from supervisor

revans2:


revans2:
Either one of the 2 commits fixes the problem with the tests taking too long.  This first one makes it so metrics are actually reported back in local mode, which will help with tests in the future.

The second commit decouples the metrics reporting from the slot state machine path, so it will not block when trying to communicate with nimbus, for example if nimbus is down.

storm-server/src/main/java/org/apache/storm/daemon/supervisor/OnlyLatestExecutor.java
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    /**
+     * Run something in the future, but replace it with the latest if it is taking too long
+     * @param key what to use to dedupe things.
+     * @param r what you want to run.
+     */
+    public void execute(final K key, Runnable r) {
+        Runnable old = latest.put(key, r);
+        if (old == null) {
```
agresch: do we care about knowing that something is dropped? Seems like we would be ignorant of any issue that requires tuning.

HeartSaVioR: Yes if we are concerned about that we may want to have WARN log, and even if not, DEBUG log might be better.

storm-server/src/main/java/org/apache/storm/daemon/supervisor/OnlyLatestExecutor.java
```diff
@@ -0,0 +1,55 @@
...
+            //It was not there before so we need to run it.
+            exec.execute(() -> {
+                Runnable run = latest.remove(key);
+                if (run != null) {
+                    run.run();;
```
HeartSaVioR: nit: double ;

revans2:
@kishorvpatil @HeartSaVioR @agresch 

I have addressed all of your review comments.  Please take another look, and if it is good I'll squash the commits and merge it in.

PR #2538, opened 2018-01-27
 STORM-2913: Add metadata to at-most-once and at-least-once commits 

srdo:
https://issues.apache.org/jira/browse/STORM-2913

This builds on STORM-2914.

I believe we can resolve STORM-2913 by committing metadata in all processing guarantee modes, rather than just AT_LEAST_ONCE. This change simply adds metadata to the AT_MOST_ONCE and NONE commit statements.

### DIFF
--- a/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
+++ b/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
@@ -23,8 +23,6 @@ import static org.apache.storm.kafka.spout.KafkaSpoutConfig.FirstPollOffsetStrat
 import static org.apache.storm.kafka.spout.KafkaSpoutConfig.FirstPollOffsetStrategy.UNCOMMITTED_EARLIEST;
 import static org.apache.storm.kafka.spout.KafkaSpoutConfig.FirstPollOffsetStrategy.UNCOMMITTED_LATEST;
 
-import com.fasterxml.jackson.core.JsonProcessingException;
-import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.annotations.VisibleForTesting;
 
 import java.io.IOException;
@@ -53,6 +51,7 @@ import org.apache.kafka.common.errors.RetriableException;
 import org.apache.storm.kafka.spout.KafkaSpoutConfig.FirstPollOffsetStrategy;
 import org.apache.storm.kafka.spout.KafkaSpoutConfig.ProcessingGuarantee;
 import org.apache.storm.kafka.spout.internal.CommitMetadata;
+import org.apache.storm.kafka.spout.internal.CommitMetadataManager;
 import org.apache.storm.kafka.spout.internal.KafkaConsumerFactory;
 import org.apache.storm.kafka.spout.internal.KafkaConsumerFactoryDefault;
 import org.apache.storm.kafka.spout.internal.OffsetManager;
@@ -72,7 +71,6 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
     //Initial delay for the commit and subscription refresh timers
     public static final long TIMER_DELAY_MS = 500;
     private static final Logger LOG = LoggerFactory.getLogger(KafkaSpout.class);
-    private static final ObjectMapper JSON_MAPPER = new ObjectMapper();
 
     // Storm
     protected SpoutOutputCollector collector;
@@ -104,8 +102,7 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
     // Triggers when a subscription should be refreshed
     private transient Timer refreshSubscriptionTimer;
     private transient TopologyContext context;
-    // Metadata information to commit to Kafka. It is unique per spout per topology.
-    private transient String commitMetadata;
+    private transient CommitMetadataManager commitMetadataManager;
     private transient KafkaOffsetMetric kafkaOffsetMetric;
 
     public KafkaSpout(KafkaSpoutConfig<K, V> kafkaSpoutConfig) {
@@ -142,7 +139,7 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
         offsetManagers = new HashMap<>();
         emitted = new HashSet<>();
         waitingToEmit = new HashMap<>();
-        setCommitMetadata(context);
+        commitMetadataManager = new CommitMetadataManager(context, kafkaSpoutConfig.getProcessingGuarantee());
 
         tupleListener.open(conf, context);
         if (canRegisterMetrics()) {
@@ -154,7 +151,7 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
 
     private void registerMetric() {
         LOG.info("Registering Spout Metrics");
-        kafkaOffsetMetric = new KafkaOffsetMetric(() -> offsetManagers, () -> kafkaConsumer);
+        kafkaOffsetMetric = new KafkaOffsetMetric(() -> Collections.unmodifiableMap(offsetManagers), () -> kafkaConsumer);
         context.registerMetric("kafkaOffset", kafkaOffsetMetric, kafkaSpoutConfig.getMetricsTimeBucketSizeInSecs());
     }
 
@@ -166,16 +163,6 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
             return false;
         }
         return true;
-    }
-
-    private void setCommitMetadata(TopologyContext context) {
-        try {
-            commitMetadata = JSON_MAPPER.writeValueAsString(new CommitMetadata(
-                context.getStormId(), context.getThisTaskId(), Thread.currentThread().getName()));
-        } catch (JsonProcessingException e) {
-            LOG.error("Failed to create Kafka commit metadata due to JSON serialization error", e);
-            throw new RuntimeException(e);
-        }
     }
 
     private boolean isAtLeastOnceProcessing() {
@@ -215,8 +202,8 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
                 retryService.retainAll(partitions);
 
                 /*
-                 * Emitted messages for partitions that are no longer assigned to this spout can't
-                 * be acked and should not be retried, hence remove them from emitted collection.
+                 * Emitted messages for partitions that are no longer assigned to this spout can't be acked and should not be retried, hence
+                 * remove them from emitted collection.
                  */
                 emitted.removeIf(msgId -> !partitions.contains(msgId.getTopicPartition()));
             }
@@ -246,7 +233,7 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
 
             if (committedOffset != null) {
                 // offset was previously committed for this consumer group and topic-partition, either by this or another topology.
-                if (isOffsetCommittedByThisTopology(newTp, committedOffset)) {
+                if (commitMetadataManager.isOffsetCommittedByThisTopology(newTp, committedOffset, Collections.unmodifiableMap(offsetManagers))) {
                     // Another KafkaSpout instance (of this topology) already committed, therefore FirstPollOffsetStrategy does not apply.
                     kafkaConsumer.seek(newTp, committedOffset.offset());
                 } else {
@@ -274,31 +261,6 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
         }
     }
 
-    /**
-     * Checks if {@link OffsetAndMetadata} was committed by a {@link KafkaSpout} instance in this topology. This info is used to decide if
-     * {@link FirstPollOffsetStrategy} should be applied
-     *
-     * @param tp topic-partition
-     * @param committedOffset {@link OffsetAndMetadata} info committed to Kafka
-     * @return true if this topology committed this {@link OffsetAndMetadata}, false otherwise
-     */
-    private boolean isOffsetCommittedByThisTopology(TopicPartition tp, OffsetAndMetadata committedOffset) {
-        try {
-            if (offsetManagers.containsKey(tp) && offsetManagers.get(tp).hasCommitted()) {
-                return true;
-            }
-
-            final CommitMetadata committedMetadata = JSON_MAPPER.readValue(committedOffset.metadata(), CommitMetadata.class);
-            return committedMetadata.getTopologyId().equals(context.getStormId());
-        } catch (IOException e) {
-            LOG.warn("Failed to deserialize [{}]. Error likely occurred because the last commit "
-                + "for this topic-partition was done using an earlier version of Storm. "
-                + "Defaulting to behavior compatible with earlier version", committedOffset);
-            LOG.trace("", e);
-            return false;
-        }
-    }
-
     // ======== Next Tuple =======
     @Override
     public void nextTuple() {
@@ -311,7 +273,10 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
                 if (isAtLeastOnceProcessing()) {
                     commitOffsetsForAckedTuples(kafkaConsumer.assignment());
                 } else if (kafkaSpoutConfig.getProcessingGuarantee() == ProcessingGuarantee.NO_GUARANTEE) {
-                    commitFetchedOffsetsAsync(kafkaConsumer.assignment());
+                    Map<TopicPartition, OffsetAndMetadata> offsetsToCommit = 
+                        createFetchedOffsetsMetadata(kafkaConsumer.assignment());
+                    kafkaConsumer.commitAsync(offsetsToCommit, null);
+                    LOG.debug("Committed offsets {} to Kafka", offsetsToCommit);
                 }
             }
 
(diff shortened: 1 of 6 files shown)


hmcl:
@srdo there is a discrepancy between the title of this pull request and the title of the associated JIRA. What problem are you trying to solve in this patch? Add meatada, or remove warnings?

If the goal is to solve the recurring WARN messages that get printed, in my opinion the obvious thing to do is to simply not log the message unless the processing guarantee is AT_LEAST_ONCE. This would be a one line change

Can you please also clarify what is the need to add metadata when running the spout in AT-MOST-ONCE mode ? Also, I think it is quite dangerous to try to mimic the behavior Kafka does by simply calling commitAsync(...) in auto.commit mode. Besides the WARN messages, is there any other problem associated with letting Kafka handle everything in auto.commit mode?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -396,7 +361,10 @@ private void setWaitingToEmit(ConsumerRecords<K, V> consumerRecords) {
                 numPolledRecords);
             if (kafkaSpoutConfig.getProcessingGuarantee() == KafkaSpoutConfig.ProcessingGuarantee.AT_MOST_ONCE) {
                 //Commit polled records immediately to ensure delivery is at-most-once.
-                kafkaConsumer.commitSync();
+                Map<TopicPartition, OffsetAndMetadata> offsetsToCommit = 
+                    createOffsetsToCommitForConsumedOffsets(kafkaConsumer.assignment());
+                kafkaConsumer.commitSync(offsetsToCommit);
```
hmcl: createFetchedOffsetsMetadata

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -396,7 +361,10 @@ private void setWaitingToEmit(ConsumerRecords<K, V> consumerRecords) {
...
+                LOG.debug("Committed offsets {} to Kafka", offsetsToCommit);
```
hmcl: Committed offsets {} synchronously to Kafka

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -469,11 +437,14 @@ private boolean emitOrRetryTuple(ConsumerRecord<K, V> record) {
...
-            if (committedOffset != null && isOffsetCommittedByThisTopology(tp, committedOffset)
+            if (isAtLeastOnceProcessing()
+                && committedOffset != null 
+                && commitMetadataManager.isOffsetCommittedByThisTopology(tp, committedOffset, offsetManagers)
                 && committedOffset.offset() > record.offset()) {
                 // Ensures that after a topology with this id is started, the consumer fetch
                 // position never falls behind the committed offset (STORM-2844)
-                throw new IllegalStateException("Attempting to emit a message that has already been committed.");
+                throw new IllegalStateException("Attempting to emit a message that has already been committed."
+                    + " This should never occur in at-least-once mode.");
```
hmcl: for at-least-once semantics.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/CommitMetadataManager.java
```diff
@@ -0,0 +1,90 @@
...
+import org.apache.storm.task.TopologyContext;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Generates and reads commit metadata.
+ */
+public class CommitMetadataManager {
+
+    private static final ObjectMapper JSON_MAPPER = new ObjectMapper();
```
hmcl: Do we want one ObjectMapper for all the KafkaSpout instances (executors), or one per executor? This will share it across all the instances. Perhaps we should have one per instance.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/CommitMetadataManager.java
```diff
@@ -0,0 +1,90 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger(CommitMetadata.class);
+    // Metadata information to commit to Kafka. It is unique per spout instance.
```
hmcl: Do you mean CommitMetadataManager.class ?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/CommitMetadataManager.java
```diff
@@ -0,0 +1,90 @@
...
+    private final ProcessingGuarantee processingGuarantee;
+    private final TopologyContext context;
+
+    /**
+     * Create a manager with the given context.
+     */
+    public CommitMetadataManager(TopologyContext context, ProcessingGuarantee processingGuarantee) {
+        this.context = context;
+        try {
+            commitMetadata = JSON_MAPPER.writeValueAsString(new CommitMetadata(
```
hmcl: Ideally commitMetadata would be passed in the constructor to facilitate unit testing. We could have a factory method in this class itself with this code

```java
public CommitMetadataManager(TopologyContext context, ProcessingGuarantee processingGuarantee, String commitMetadata)
```

```java
public static CommitMetadataManager newInstance(TopologyContext context, ProcessingGuarantee processingGuarantee) {
    return new CommitMetadataManager(context, processingGuarantee, JSON_MAPPER.writeValueAsString(new CommitMetadata(context.getStormId(), context.getThisTaskId(), Thread.currentThread().getName())));
}
```

handling the JsonProcessingException in the factory method

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -469,11 +437,14 @@ private boolean emitOrRetryTuple(ConsumerRecord<K, V> record) {
             LOG.trace("Tuple for record [{}] has already been emitted. Skipping", record);
         } else {
             final OffsetAndMetadata committedOffset = kafkaConsumer.committed(tp);
-            if (committedOffset != null && isOffsetCommittedByThisTopology(tp, committedOffset)
+            if (isAtLeastOnceProcessing()
+                && committedOffset != null 
+                && commitMetadataManager.isOffsetCommittedByThisTopology(tp, committedOffset, offsetManagers)
```
hmcl: Collections.unmodifiableMap(offsetManagers)

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/CommitMetadataManager.java
```diff
@@ -0,0 +1,90 @@
...
+            if (processingGuarantee == ProcessingGuarantee.AT_LEAST_ONCE
+                && offsetManagers.containsKey(tp)
+                && offsetManagers.get(tp).hasCommitted()) {
+                return true;
+            }
+
+            final CommitMetadata committedMetadata = JSON_MAPPER.readValue(committedOffset.metadata(), CommitMetadata.class);
+            return committedMetadata.getTopologyId().equals(context.getStormId());
+        } catch (IOException e) {
+            LOG.warn("Failed to deserialize [{}]. Error likely occurred because the last commit "
```
hmcl: We should either write in the README or as part of this message that this WARN is expected the first time a user starts this or an earlier version of the spout with commits to Kafka done by an older version of the spout.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -311,7 +273,10 @@ public void nextTuple() {
                 if (isAtLeastOnceProcessing()) {
                     commitOffsetsForAckedTuples(kafkaConsumer.assignment());
                 } else if (kafkaSpoutConfig.getProcessingGuarantee() == ProcessingGuarantee.NONE) {
-                    commitConsumedOffsets(kafkaConsumer.assignment());
+                    Map<TopicPartition, OffsetAndMetadata> offsetsToCommit = 
+                        createOffsetsToCommitForConsumedOffsets(kafkaConsumer.assignment());
+                    kafkaConsumer.commitAsync(offsetsToCommit, null);
```
hmcl: createFetchedOffsetsMetadata

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/CommitMetadataManager.java
```diff
@@ -0,0 +1,90 @@
...
+import org.apache.storm.kafka.spout.KafkaSpout;
+import org.apache.storm.kafka.spout.KafkaSpoutConfig.ProcessingGuarantee;
+import org.apache.storm.task.TopologyContext;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Generates and reads commit metadata.
+ */
+public class CommitMetadataManager {
```
hmcl: All the methods and constructors in this class should be package protected

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -142,7 +139,7 @@ public void open(Map<String, Object> conf, TopologyContext context, SpoutOutputC
         offsetManagers = new HashMap<>();
         emitted = new HashSet<>();
         waitingToEmit = new HashMap<>();
-        setCommitMetadata(context);
+        commitMetadataManager = new CommitMetadataManager(context, kafkaSpoutConfig.getProcessingGuarantee());
```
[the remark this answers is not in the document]
srdo: I think we should wait. Let's do it if someone needs it, but introducing more extension points than we need is likely to cause us more headaches down the road, because once the extension point is public it's harder for us to change if we need to because we have to consider that other people may be implementing the interface.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
srdo: Will rename

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/CommitMetadataManager.java
srdo: It's in the internal package, so if we make it package private the spout can't see it. I think it's up to users to figure out that if the class is in a package called "internal", they probably shouldn't use it. Once we start looking at Java 9 modularization we can make sure not to export this.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/CommitMetadataManager.java
srdo: Objectmappers are thread safe. I don't see anywhere in the documentation that mentions whether there are internal locks being used in the ObjectMapper, but this post suggests there aren't https://stackoverflow.com/questions/18611565/how-do-i-correctly-reuse-jackson-objectmapper#comment27462917_18618918. A quick google also suggests that it should be fine to use ObjectMapper as a singleton.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/CommitMetadataManager.java
srdo: No, the metadata is the same if you create two CommitMetadataManagers from the same spout instance.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/CommitMetadataManager.java
srdo: Yes, I'll update this message

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/CommitMetadataManager.java
srdo: I would agree if CommitMetadata weren't a POJO. It doesn't have any behavior, why do we need to stub it?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
srdo: I don't mind adding it, but is this information useful to the user?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
srdo: Good point

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/CommitMetadataManager.java
srdo: Tried to update this so it's a little more clear about when it will be printed

srdo:
@hmcl Rebased.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/CommitMetadataManager.java
hmcl: ok

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutConfig.java
```diff
@@ -453,37 +460,37 @@ public Builder(String bootstrapServers, Subscription subscription) {
...
+        String autoOffsetResetPolicy = (String) builder.kafkaProps.get(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG);
+        if (builder.processingGuarantee == ProcessingGuarantee.AT_LEAST_ONCE) {
+            if (autoOffsetResetPolicy == null) {
+                /*
+                 * If the user wants to explicitly set an auto offset reset policy, we should respect it, but when the spout is configured
+                 * for at-least-once processing we should default to seeking to the earliest offset in case there's an offset out of range
+                 * error, rather than seeking to the latest (Kafka's default). This type of error will typically happen when the consumer
+                 * requests an offset that was deleted.
+                 */
+                LOG.info("Setting consumer property '{}' to 'earliest' to ensure at-least-once processing",
```
hmcl: NIT: I would write Kafka consumer, but not a deal breaker.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpoutConfig.java
```diff
@@ -453,37 +460,37 @@ public Builder(String bootstrapServers, Subscription subscription) {
...
+            }
+        } else if (builder.processingGuarantee == ProcessingGuarantee.AT_MOST_ONCE) {
+            if (autoOffsetResetPolicy != null
+                && (!autoOffsetResetPolicy.equals("latest") && !autoOffsetResetPolicy.equals("none"))) {
+                LOG.warn("Cannot guarantee at-most-once processing with auto.offset.reset.policy other than 'latest' or 'none'."
+                    + " Some messages may be processed more than once.");
             }
-            builder.kafkaProps.put(ConsumerConfig.ENABLE_AUTO_COMMIT_CONFIG, false);
         }
+        LOG.info("Setting consumer property '{}' to 'false', because the spout does not support auto-commit",
```
hmcl: NIT: I would write Kafka consumer, but not a deal breaker.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/CommitMetadataManager.java
hmcl: I disagree. The name of the logger instance should be the class that is logging the message or one of its super classes. There is no class called CommitMetadata in the codebase, so why should we have a logger called CommitMetadata?

srdo: Right, I misunderstood. I thought you were talking about the comment in L38 (i.e. you wanted it changed to "... It is unique per CommitMetadataManager"). Will fix

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/CommitMetadataManager.java
srdo: It's my impression that doing something like that (declaring one global ObjectMapper) would be fine, except if we needed different configurations from the ObjectMappers. https://stackoverflow.com/a/3909846/8845188 (The responder is one of the Jackson devs).

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/CommitMetadataManager.java
srdo: I'm happy to declare the class final, hopefully that'll discourage people from using the class directly if the internal package hint doesn't do the trick.

Regarding creating objects in the constructor, I agree with you in the general case (DI is great), but for pure data objects like CommitMetadata I'm not sure I understand the harm?

hmcl:
+1. 

Once squashed is good to merge as far as I am concerned. Thanks a lot @srdo.

HeartSaVioR:
@srdo 
Thanks for the great work. I'd rather not review the two PRs since @hmcl already gave a nice review. If you would like to merge twos by yourself, please replace storm-kafka-client 1.1.x/1.0.x with 1.x afterwards. If you would rather let me do it please let me know. Thanks in advance!

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/CommitMetadataManager.java
hswangxiao: When I get this exception, what should I do

srdo: @hswangxiao Are you on the latest version?

PR #2539, opened 2018-01-30
STORM-2917: Check the derecated config nimbus.host

MichealShin:
There is a situation: the deployer wants to use the new nimbus config(nimbus.seeds), but still leave the blank deprecated config(nimbus.host) in storm.yaml. It will not work.
  To avoid this, the program should at least check whether the deprecated config is blank before using it to connect.


HeartSaVioR:
@MichealShin 
Could you craft the patch against 1.x branch as well? Thanks in advance!

HeartSaVioR:
Merged via cherry-picking first commit which contains all of change.

@MichealShin Could you please close this? Thanks in advance!

PR #2542, opened 2018-02-01
STORM-2853 Initialize tick tuple after initializing spouts/bolts

HeartSaVioR:
* this prevents newly-initializing executor in deactivated topology to show high CPU usage

Please refer https://issues.apache.org/jira/browse/STORM-2853 for more details.

Please also note that we assume all the tasks in executor are for same component.


revans2:
Test failures look unrelated

PR #2543, opened 2018-02-01
STORM-2853 Initialize tick tuple after initializing spouts/bolts (1.x version lines)

HeartSaVioR:
* this prevents newly-initializing executor in deactivated topology to show high CPU usage
* this is based on the fact that all the tasks in executor are for same component

I guess it would be clean cherry-pick among branches regarding 1.x version lines. (1.1.x/1.0.x)

### DIFF
--- a/storm-core/src/clj/org/apache/storm/daemon/executor.clj
+++ b/storm-core/src/clj/org/apache/storm/daemon/executor.clj
@@ -350,7 +350,7 @@
      (when (seq data-points)
        (task/send-unanchored task-data Constants/METRICS_STREAM_ID [task-info data-points]))))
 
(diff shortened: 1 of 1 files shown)


danny0405:
Seems start tick thread when we active spouts/bolts instead of when making executor-data is more reasonable，So after this patch, all the tasks in one executor will share one tick thread now?

HeartSaVioR:
@danny0405 
> So after this patch, all the tasks in one executor will share one tick thread now?

Yes, but it is actually unchanged. I'm just ensuring that tick tuples are not generated when executor is not ready to process the tuples.

revans2:
test failures look unrelated

PR #2544, opened 2018-02-02
[STORM-2932] the naming of topology localityaware configs are confusing

Ethanlm:
https://issues.apache.org/jira/browse/STORM-2932

The names are confusing. I think we should delete "percent"

### DIFF
--- a/conf/defaults.yaml
+++ b/conf/defaults.yaml
@@ -258,8 +258,8 @@ topology.disruptor.batch.size: 100
 topology.disruptor.batch.timeout.millis: 1
 topology.disable.loadaware.messaging: false
 topology.state.checkpoint.interval.ms: 1000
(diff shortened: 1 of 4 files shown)


revans2:
The changes look okay to me.  I just want to be sure that you are aware that this is a breaking change and the consequences of that.  Because this config went into 2.x and has not been released it is probably not a big deal, but if someone has started to use it they would need to change their settings.

Ethanlm:
@revans2 Yes you are right. I assumed that this can be changed since 2.x is not released.

HeartSaVioR:
@revans2 @Ethanlm 
I don't think this is a deal. We didn't release Storm 2.0.0 officially, and also that's added just 4 months ago hence even early-adopter may exist, the possibility for end-users to utilize these parameters is really low.

PR #2546, opened 2018-02-05
STORM-2934 - fix startup ClassNotFoundException when missing RocksDB jar

agresch:
Currently, only Nimbus uses STORM_METRIC_STORE_CLASS.  

If the RocksDB jar is missing on Daemons other than Nimbus, they should be able to startup successfully.  

If the jar is missing on Nimbus, the metricStore will fail to initialize, and an ERROR will be logged. Nimbus will still be able to run; it just ignores using the metricStore.

### DIFF
--- a/storm-server/src/main/java/org/apache/storm/DaemonConfig.java
+++ b/storm-server/src/main/java/org/apache/storm/DaemonConfig.java
@@ -1035,7 +1035,9 @@ public class DaemonConfig implements Validated {
      * Class implementing MetricStore.
      */
     @NotNull
-    @isImplementationOfClass(implementsClass = MetricStore.class)
+    @isString
+    // Validating class implementation could fail on non-Nimbus Daemons.  Nimbus will catch the class not found on startup
+    // and log an error message, so just validating this as a String for now.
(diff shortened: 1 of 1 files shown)


Ethanlm:
The dependency rocksdbjni.jar is present in lib/ directory after packaging so there shouldn't be exceptions during nimbus daemon startup. But changing from `isImplementationOfClass` to `isString ` makes sense to me. It's just like what we do for `SUPERVISOR_BLOBSTORE` and `NIMBUS_BLOBSTORE`

storm-server/src/main/java/org/apache/storm/DaemonConfig.java
```diff
@@ -1035,7 +1035,10 @@
      * Class implementing MetricStore.
      */
     @NotNull
-    @isImplementationOfClass(implementsClass = MetricStore.class)
+    @isString
+    //@isImplementationOfClass(implementsClass = MetricStore.class)
```
HeartSaVioR: minor: Let's remove this line then. I think below explanation is enough.

agresch:
removed comment

agresch:
squashed

PR #2548, opened 2018-02-05
STORM-2935 Add metric to track when TGT expires

agresch:


storm-client/src/jvm/org/apache/storm/daemon/worker/WorkerState.java
```diff
@@ -20,6 +20,17 @@
 
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Sets;
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.util.*;
```
Ethanlm: I believe storm community says better to not use *. 

Btw, in Intellij, you can navigate Preference--> Editor --> code style --> Java and set `class count to use import with *` to a high number to avoid auto collapse

agresch: That was just a re-order.

HeartSaVioR: Might be better to fix once we found it, but this is a kind of nit.

HeartSaVioR:
Looks like this patch proposes making a change to Metrics V1 interface which is deprecated. I'd like to see us having a plan to migrate Metrics V1 into Metrics V2 instead of keep using it.

HeartSaVioR:
I just noticed that Metrics V2 is not port-forwarded to master branch. I'd suggest waiting for Metrics V2 integration against master, and apply it unless it is urgent.

agresch:
We're trying to bring this change back to older releases internally.  Since v2 is not available and we want to backport this, could we get this in and address changing to v2 in the near future?

storm-client/src/jvm/org/apache/storm/executor/bolt/BoltExecutor.java
```diff
@@ -77,6 +83,17 @@ public void init(Map<Integer, Task> idToTask) {
...
                 BuiltinMetricsUtil.registerIconnectionClientMetrics(cachedNodePortToSocket, topoConf, userContext);
                 BuiltinMetricsUtil.registerIconnectionServerMetric(workerData.getReceiver(), topoConf, userContext);
+
+                // add any autocredential expiry metrics from the worker
+                if (workerData.getAutoCredentials() != null) {
+                    int bucketSize = ((Number) topoConf.get(Config.TOPOLOGY_BUILTIN_METRICS_BUCKET_SIZE_SECS)).intValue();
+                    for (IAutoCredentials autoCredential : workerData.getAutoCredentials()) {
+                        if (autoCredential instanceof IMetric) {
+                            IMetric metric = (IMetric)autoCredential;
+                            userContext.registerMetric(metric.getMetricName(), metric, bucketSize);
```
HeartSaVioR: `metric.getMetricName()` makes sense for only this case. I agree that's easiest way to address, but how about having new interface like `RegisteringMetrics` or another better name which has `registerMetrics(TopologyContext topoContext, Map<String, Object> topoConf)` registering metrics by itself?

This will require only change of AutoTGT when migrating Metrics V1 into Metrics V2, and also enable registering multiple metrics if needed. (You can't register multiple metrics in this way in Metrics V2, whereas there's workaround to provide Map from `getValueAndReset()` like what other built-in metrics are doing in Metrics V1.)

agresch:
@HeartSaVioR  Let me know if this looks better.

storm-client/src/jvm/org/apache/storm/security/auth/kerberos/AutoTGT.java
```diff
@@ -33,18 +33,21 @@
...
 
+import org.apache.storm.task.TopologyContext;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
  * Automatically take a user's TGT, and push it, and renew it in Nimbus.
  */
-public class AutoTGT implements IAutoCredentials, ICredentialsRenewer, IMetric {
+public class AutoTGT implements IAutoCredentials, ICredentialsRenewer, IMetric, IMetricsRegistrant {
```
HeartSaVioR: We can have individual metric instance and get rid of needs for implementing `IMetric`. It should be removed in Metrics V2 anyway.

agresch:
@HeartSaVioR - is this what you were requesting?

HeartSaVioR:
@agresch 
I meant having metric field in AutoTGT so that it can be changed to Gauge in Codahale metric for Metrics V2, but it's OK if lambda works in Metrics V1.

PR #2551, opened 2018-02-07
[STORM-2940] Dynamically set the CAPACITY value of LoadAwareShuffleGrouping

Ethanlm:
Explained in https://issues.apache.org/jira/browse/STORM-2940.

This is to fix the problem thatin some extreme cases where random.nextInt will receive currentIdx=0 and throws `java.lang.IllegalArgumentException: bound must be positive` exception

### DIFF
--- a/storm-client/src/jvm/org/apache/storm/grouping/LoadAwareShuffleGrouping.java
+++ b/storm-client/src/jvm/org/apache/storm/grouping/LoadAwareShuffleGrouping.java
@@ -85,7 +85,7 @@ public class LoadAwareShuffleGrouping implements LoadAwareCustomStreamGrouping, 
         sourceNodeInfo = new NodeInfo(context.getThisWorkerHost(), Sets.newHashSet((long) context.getThisWorkerPort()));
         taskToNodePort = context.getTaskToNodePort();
         this.targetTasks = targetTasks;
-        capacity = targetTasks.size() == 1 ? 1 : targetTasks.size() * MAX_WEIGHT;
+        capacity = targetTasks.size() == 1 ? 1 : Math.max(1000, targetTasks.size() * 5);
         conf = context.getConf();
         dnsToSwitchMapping = ReflectionUtils.newInstance((String) conf.get(Config.STORM_NETWORK_TOPOGRAPHY_PLUGIN));
         localityGroup = new HashMap<>();
--- a/storm-client/src/jvm/org/apache/storm/grouping/LoadAwareShuffleGrouping.java
+++ b/storm-client/src/jvm/org/apache/storm/grouping/LoadAwareShuffleGrouping.java
@@ -44,7 +44,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class LoadAwareShuffleGrouping implements LoadAwareCustomStreamGrouping, Serializable {
-    static final int CAPACITY = 1000;
+    private int CAPACITY;
     private static final int MAX_WEIGHT = 100;
     private static class IndexAndWeights {
         final int index;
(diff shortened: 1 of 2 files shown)


storm-client/src/jvm/org/apache/storm/grouping/LoadAwareShuffleGrouping.java
```diff
@@ -44,7 +44,7 @@
 import org.slf4j.LoggerFactory;
 
 public class LoadAwareShuffleGrouping implements LoadAwareCustomStreamGrouping, Serializable {
-    static final int CAPACITY = 1000;
+    private int CAPACITY;
```
HeartSaVioR: If we no longer use it as constant, the name should be changed to lower case.

storm-client/src/jvm/org/apache/storm/grouping/LoadAwareShuffleGrouping.java
```diff
@@ -85,6 +85,7 @@ public void prepare(WorkerTopologyContext context, GlobalStreamId stream, List<I
         sourceNodeInfo = new NodeInfo(context.getThisWorkerHost(), Sets.newHashSet((long) context.getThisWorkerPort()));
         taskToNodePort = context.getTaskToNodePort();
         this.targetTasks = targetTasks;
+        CAPACITY = targetTasks.size() == 1 ? 1 : targetTasks.size() * MAX_WEIGHT;
```
HeartSaVioR: Given that other logic still leverage max weight, will this work for targetTasks.size() == 1? If it works, It may be ideal to have test on this case to clear out wondering.

Ethanlm: `targetTasks.size() == 1`  is just for performance enhancement. Since there is only one target task, there is no need to compose a choice array and choose the target from the elements of the choice array; Set CAPACITY to 1 can avoid this. 
For the particular equation,
`int count = (int) ((indexAndWeights.weight / (double) weightSum) * CAPACITY);`
 because there is only one target, so `indexAndWeights.weight / (double) weightSum` must be 1 and count must be 1 if there is any load on that target task.

HeartSaVioR: OK we are ensuring it already.

Ethanlm:
@HeartSaVioR @revans2 Could you please take a look again? If it feels good, I will squash the commits. Thanks

storm-client/src/jvm/org/apache/storm/grouping/LoadAwareShuffleGrouping.java
```diff
@@ -85,6 +85,7 @@ public void prepare(WorkerTopologyContext context, GlobalStreamId stream, List<I
         sourceNodeInfo = new NodeInfo(context.getThisWorkerHost(), Sets.newHashSet((long) context.getThisWorkerPort()));
         taskToNodePort = context.getTaskToNodePort();
         this.targetTasks = targetTasks;
+        capacity = targetTasks.size() == 1 ? 1 : Math.max(1000, targetTasks.size() * 5);
```
HeartSaVioR: Please add tests which cover the case that task count is beyond 1000 to see if it still guarantees correct distribution with acceptable margin.

Ethanlm: I just added the tests in the last commit. Is that what you were saying? Sorry if I misunderstood it

HeartSaVioR: @Ethanlm Yes that's what I meant. Thanks for addressing.

HeartSaVioR:
+1 again.

PR #2554, opened 2018-02-08
STORM-2939 add WorkerMetricsProcessor interface

agresch:
This adds the WorkerMetricsProcessor interface.  The default implementation executes the current code that sends metric data to Nimbus for insertion, now done by NimbusMetricProcessor.

The intent is that we can replace this class for other metricstore implementations (HBase is what I am currently working on) that can instead do insertion directly.

### DIFF
--- a/storm-server/src/main/java/org/apache/storm/DaemonConfig.java
+++ b/storm-server/src/main/java/org/apache/storm/DaemonConfig.java
@@ -1031,7 +1031,7 @@ public class DaemonConfig implements Validated {
         "storm.supervisor.medium.memory.grace.period.ms";
 
     /**
-     * Class implementing MetricStore.
+     * Class implementing MetricStore.  Runs on Nimbus.
      */
     @NotNull
     @isString
@@ -1040,7 +1040,7 @@ public class DaemonConfig implements Validated {
     public static final String STORM_METRIC_STORE_CLASS = "storm.metricstore.class";
 
     /**
-     * Class implementing WorkerMetricsProcessor.
+     * Class implementing WorkerMetricsProcessor.  Runs on Supervisors.
      */
     @NotNull
     @isString
--- a/storm-server/src/main/java/org/apache/storm/DaemonConfig.java
+++ b/storm-server/src/main/java/org/apache/storm/DaemonConfig.java
@@ -33,7 +33,6 @@ import static org.apache.storm.validation.ConfigValidationAnnotations.isNoDuplic
 import static org.apache.storm.validation.ConfigValidationAnnotations.isMapEntryCustom;
 
 import org.apache.storm.container.ResourceIsolationInterface;
-import org.apache.storm.metricstore.MetricStore;
 import org.apache.storm.nimbus.ITopologyActionNotifierPlugin;
 import org.apache.storm.scheduler.blacklist.reporters.IReporter;
 import org.apache.storm.scheduler.blacklist.strategies.IBlacklistStrategy;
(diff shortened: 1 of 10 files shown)


HeartSaVioR:
Some questions:

1. Will we support combination of metricstore and metricprocessor? For example, NimbusMetricProcessor as processor but using HBaseStore as store. If metric store and metric processor should be selected as a pair to make it work, we may want to have less redundant way to configure.

2. I think we will eventually migrate worker built-in metrics. Do you have a plan or (sketched) idea of that? Because if we insert metrics to HBase store directly, it will require HBase dependencies in classpath. While it might be fine for Supervisor, sometimes it might not fine for user topologies. In this case, requiring workers to push their metrics into HBase directly would make some kinds of pains, and they would want to push metrics from user topology to daemons (either Nimbus or Supervisor) to let them handle instead.

agresch:
1) My intent was that the metricstore would be for insertion, and the metricprocessor for either forwarding the metrics to the metricstore or just extending a metricstore and doing the metric insertion in place.  The HBase metricprocessor implementation I'm working on does the latter, inserting directly from the Supervisor.

2) I do not have a plan for this yet, but we plan on tackling this.  We're planning on getting the metrics forwarded from the workers to the Supervisor and inserting from HBase there (or forwarding to Nimbus for RocksDB).  In the non-HBase implementation, we need to be able to have a path from the workers to Nimbus.

storm-server/src/main/java/org/apache/storm/DaemonConfig.java
```diff
@@ -1040,6 +1039,13 @@
     // and log an error message, so just validating this as a String for now.
     public static final String STORM_METRIC_STORE_CLASS = "storm.metricstore.class";
 
+    /**
+     * Class implementing WorkerMetricsProcessor.
```
revans2: It might be good to mention here in the javadocs that this is intended to be run on the supervisor.

storm-server/src/main/java/org/apache/storm/metricstore/MetricStoreConfig.java
```diff
@@ -41,5 +41,23 @@ public static MetricStore configure(Map conf) throws MetricException {
...
+     * @throws MetricException  on misconfiguration
+     */
+    public static WorkerMetricsProcessor configureMetricProcessor(Map conf) throws MetricException {
+
+        try {
+            String processorClass = (String)conf.get(DaemonConfig.STORM_METRIC_PROCESSOR_CLASS);
+            WorkerMetricsProcessor processor = (WorkerMetricsProcessor) (Class.forName(processorClass)).newInstance();
+            processor.prepare(conf);
+            return processor;
+        } catch (Throwable t) {
```
revans2: Can we just make this an Exception instead of a Throwable?  Most Errors are not things that we want to try and recover from.

storm-server/src/main/java/org/apache/storm/metricstore/WorkerMetricsProcessor.java
```diff
@@ -0,0 +1,40 @@
...
+
+package org.apache.storm.metricstore;
+
+import java.util.Map;
+import org.apache.storm.generated.WorkerMetrics;
+
+public interface WorkerMetricsProcessor {
+
+    /**
+     * Process insertion of worker metrics.
```
revans2: Can we mention something about thread safety in here?

storm-server/src/main/java/org/apache/storm/metricstore/WorkerMetricsProcessor.java
```diff
@@ -0,0 +1,40 @@
...
+     * @param conf Storm config map
```
revans2: Can we make sure that we distinguish between the topology config here, and the daemon config that is passed into prepare?

HeartSaVioR:
@agresch 

> We're planning on getting the metrics forwarded from the workers to the Supervisor and inserting from HBase there (or forwarding to Nimbus for RocksDB). In the non-HBase implementation, we need to be able to have a path from the workers to Nimbus.

Once you intend WorkerMetricsProcessor to be run in Supervisor, non-HBase implementation should also forward metrics to Supervisor to delegate forwarding, hence the path would be worker -> Supervisor -> Nimbus, right? Just be clear about this.

agresch:
That is correct.  With the current default RocksDB implementation, we would have workers forward to the Supervisor, and the Supervisor metric processor would be NimbusMetricProcessor - sending them to Nimbus for insertion into RocksDB.

PR #2555, opened 2018-02-12
STORM-2841 Use extended class instead of partial mock for TestHiveBolt

HeartSaVioR:
* while mocking partially, oddly it calls real method intermittently which breaks test

Please refer https://issues.apache.org/jira/browse/STORM-2841 to see the detail of failure.

I guess this is clean cherry-picking for 1.x version lines. I'll submit another one if not.


arunmahadevan:
+1

PR #2556, opened 2018-02-13
STORM-2946: Upgrade to HBase 2.0

ptgoetz:
https://issues.apache.org/jira/browse/STORM-2946

We may want to discuss changes like this in the email thread about independently versioned "external" components.


HeartSaVioR:
I'm expecting that the patch is just for the update of API change, hence questions look more important than the code change.

Several questions: 
1. Did you test the patch manually with HBase 2.0 beta 1?
2. I guess there's little chance for HBase community to introduce API change after beta, but the possibility is still open. Moreover that's not ideal to depend on beta version. Do you intend to open this for the official release of HBase 2.0 and update the version/API change once it happens?
3. Does HBase 2 client be compatible with HBase 1.x? At a glance of HBase 2 project doc (https://docs.google.com/document/d/1WCsVlnHjJeKUcl7wHwqb4z9iEu_ktczrlKHK8N4SZzs/edit#) they claim admin interface is incompatible and HBase 1 client cannot administrate HBase 2. I'm wondering about opposite case (HBase 2 client to administrate HBase 1 server), and same case for other things. In short, I'd like to determine needs to have separate hbase module for HBase 1.x and HBase 2.x.

agresch:
What is the expectation for Storm users using different versions of HBase?  

Should we have some shim layer to allow using any version?

arunmahadevan:
@ptgoetz the changes looks good.

1. Can you share the results of testing these changes with HBase 2.0
2. Are these changes compatible with HBase 1.x ? Can we try running the patched bolt with HBase 1.x?
3. Based on the result of 2, we may need to maintain separate storm-hbase modules. Are there any other alternatives ?

ptgoetz:
@arunmahadevan Yes, I will share the results of my testing and will test for backward compatibility.

I don't have high hopes for backward compatibility based on https://hbase.apache.org/book.html#hbase.versioning

If it isn't compatible, we could either maintain separate storm-hbase modules, or start versioning/releasing storm-hbase independently of storm as we've discussed doing with storm-kafka-client. I'd lean toward the latter.

ptgoetz:
@HeartSaVioR @arunmahadevan 

I tested this manually against HBase 2.0.0 beta-1 and HBase 1.1.2 using the examples as well as some custom code covering trident and core storm read/write. All tests passed. So it appears to backward compatible.

srdo:

PR #2558, opened 2018-02-14
[STORM-2950] Making FieldsGrouper.chooseTasks be consistent with 1.x version line

Ethanlm:
https://issues.apache.org/jira/browse/STORM-2950

listHashCode(keys) might return Integer.MIN_VALUE (-2147483648). Math.abs(-2147483648) equals to -2147483648.

And as @HeartSaVioR  pointed out,  to be consistent with 1.x version:
```
 (mod listHashCode(keys) numTasks)
```
We should use
```
Math.floorMod(listHashCode(keys), numTasks) 
```

### DIFF
--- a/storm-client/src/jvm/org/apache/storm/utils/TupleUtils.java
+++ b/storm-client/src/jvm/org/apache/storm/utils/TupleUtils.java
@@ -41,7 +41,7 @@ public final class TupleUtils {
     }
 
     public static <T> int chooseTaskIndex(List<T> keys, int numTasks) {
-        return Math.abs(listHashCode(keys)) % numTasks;
+        return Math.floorMod(listHashCode(keys), numTasks);
     }
 
     private static <T> int listHashCode(List<T> alist) {


storm-client/src/jvm/org/apache/storm/utils/TupleUtils.java
```diff
@@ -41,7 +41,7 @@ public static boolean isTick(Tuple tuple) {
     }
 
     public static <T> int chooseTaskIndex(List<T> keys, int numTasks) {
-        return Math.abs(listHashCode(keys)) % numTasks;
+        return Math.abs(listHashCode(keys) % numTasks);
```
HeartSaVioR: Since we are now on JDK 1.8, Math.floorMod(listHashCode(keys), numTasks) would be right implementation to make distribution be consistent with 1.x version line.

https://github.com/apache/storm/blob/a689d78f81e8783b442fc96ef2cbc7295c7a931d/storm-core/src/clj/org/apache/storm/daemon/executor.clj#L48-L56

https://docs.oracle.com/javase/8/docs/api/java/lang/Math.html#floorMod-int-int-
https://clojuredocs.org/clojure.core/mod

`Math.abs(listHashCode(keys) % numTasks)` != `Math.floorMod(listHashCode(keys), numTasks)`, but `Math.floorMod(listHashCode(keys), numTasks)` == `(mod listHashCode(keys) numTasks)`.

Yes your implementation would be right about guaranteeing positive index and might be faster, but namespace in states (being used for stateful bolts) has `task id` hence changing hash function requires state resharding. Maybe worth to see how much the difference is? We should provide state resharding anyway to support the change of task parallelism. (though changing hash function requires all of end-users to run it once.)

HeartSaVioR: Worth noting that `Math.abs(listHashCode(keys)) % numTasks` != `(mod listHashCode(keys) numTasks)` so we did it wrong, and fortunately found it.

Ethanlm: Thanks. Never thought about this. Makes sense to me.

HeartSaVioR:
@Ethanlm Thanks for applying review comment so quickly!

Now the meaning of patch would be very different from issue summary & description (as well as PR's title and commit message) which we need to fix them too. Maybe something along with `Making FieldsGrouper.chooseTasks be consistent with 1.x version line` or better explanation?

Ethanlm:
@HeartSaVioR  Will this be to long: 

[STORM-2950] fix ArrayIndexOutOfBoundsException in FieldsGrouper.chooseTasks function and make it consistent with 1.x version

HeartSaVioR:
@Ethanlm `Fixing ArrayIndexOutOfBoundsException` is the side effect of making the behavior back to 1.x version. We already know 1.x version doesn't have such issue, so it can be placed to side note (commit body message) and removed from title.

PR #2559, opened 2018-02-16
STORM-2952: Deprecate storm-kafka

srdo:
https://issues.apache.org/jira/browse/STORM-2952

Please let this sit for a few days before merging, to give people time to object. We had a discussion thread about doing this, but it's good not to sneak a change like this by quietly.

### DIFF
--- a/external/storm-kafka/README.md
+++ b/external/storm-kafka/README.md
@@ -1,6 +1,9 @@
 Storm Kafka
 ====================
 
+## Deprecated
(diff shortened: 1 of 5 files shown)


HeartSaVioR:
I think waiting more than 7 days would be enough. Will merge shortly.

srdo:
@HeartSaVioR Thanks, 7 days seems reasonable enough. I'll raise a PR later to remove storm-kafka from master.

PR #2560, opened 2018-02-17
STORM-2896: Add tool to help users migrate offsets from storm-kafka t…

srdo:
…o storm-kafka-client

I'm not sure if we want this in master as well? 

I'd like to make a similar tool for the Trident spouts, but those aren't covered in this PR. I tested this by running the storm-perf Kafka null bolt topology (with and without wildcard topics) and migrating the resulting offsets into Kafka.

### DIFF
--- a/docs/storm-kafka.md
+++ b/docs/storm-kafka.md
@@ -4,6 +4,10 @@ layout: documentation
 documentation: true
 ---
 
+## Deprecated
+Storm-kafka has been deprecated and will be removed in a future Storm release. Please upgrade to storm-kafka-client.
+If you need to migrate the committed offsets to the new spout, consider using the storm-kafka-migration tool.
+
 Provides core Storm and Trident spout implementations for consuming data from Apache Kafka 0.8.x.
 
 ##Spouts
--- a/external/storm-kafka-migration/src/main/java/org/apache/storm/kafka/migration/KafkaSpoutMigration.java
+++ b/external/storm-kafka-migration/src/main/java/org/apache/storm/kafka/migration/KafkaSpoutMigration.java
@@ -1,0 +1,163 @@
+/*
+ * Copyright 2018 The Apache Software Foundation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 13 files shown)


srdo:
Added tool to migrate Trident spouts.

HeartSaVioR:
I think it would be last call to migrate if this is shipped to master branch. So I think it would be good to have in 2.0.0. We may not want to have it in 3.0.0 version line.

srdo:
Thanks, will make a master branch version.

external/storm-kafka-migration/src/main/java/org/apache/storm/kafka/migration/KafkaTridentSpoutMigration.java
```diff
@@ -69,24 +70,40 @@ public String toString() {
         }
     }
 
+    /**
+     * Get value for key. Error if value is null or not the expected type.
+     */
+    public static <T> T getOrError(Map<String, Object> conf, String key) {
```
HeartSaVioR: Not a big deal (because the storm-kafka-migration module is just a utility), but getOrError() looks like better to place to separate class to decouple two independent classes.  Just a 2 cents and I'm OK even we keep it as it is.

external/storm-kafka-migration/src/main/java/org/apache/storm/kafka/migration/KafkaSpoutMigration.java
```diff
@@ -54,25 +55,30 @@
...
     }
 
+    /**
+     * Migrates offsets from the Zookeeper store used by the storm-kafka non-Trident spouts, to Kafka's offset store used by the
+     * storm-kafka-client non-Trident spout.
+     */
     public static void main(String[] args) throws Exception {
         if (args.length != 1) {
             System.err.println("Args: confFile");
+            return;
```
HeartSaVioR: You may want to exit with explicit return code: it will help scripts which leverages this module. Same for trident migration.

srdo:
@HeartSaVioR Thanks for reviewing. Addressed the comments. Yes, I migrated offsets from both a regular spout and a trident spout, both for a regular topic and a wildcard subscription. For the regular spout, I think it's good enough to show that the offsets a migrated to the specified consumer group, so I didn't bother setting up a storm-kafka-client spout. For Trident I set up a storm-kafka-client Trident topology and checked that the new topology picked up where the storm-kafka topology left off.

docs/storm-kafka.md
```diff
@@ -4,6 +4,10 @@ layout: documentation
 documentation: true
 ---
 
+## Deprecated
```
HeartSaVioR: +1 on adding explanation of deprecation.

PR #2561, opened 2018-02-18
STORM-2958 - Use new wait strategy model for Spout as well

roshannaik:


HeartSaVioR:
@roshannaik 
You may want to update `Performance.md` as well in this patch, since it explains old spout wait strategy.

PR #2562, opened 2018-02-19
STORM-2960 Stress importance of setting up proper OS account for Storm processes

HeartSaVioR:
While the explanation of last statement implicitly says that setting up proper OS account is needed, but explicitly mentioning it should help end-users to understand the importance while setting up OS account for Storm cluster.

This change should be clean cherry-pick for 1.x version lines.

### DIFF
--- a/docs/SECURITY.md
+++ b/docs/SECURITY.md
@@ -17,6 +17,9 @@ Authentication and Authorization. But to do so usually requires
 configuring your Operating System to restrict the operations that can be done.
 This is generally a good idea even if you plan on running your cluster with Auth.
 
+Storm's OS level security relies on running Storm processes using OS accounts that have only the permissions they need. 
+Note that workers run under the same OS account as the Supervisor daemon by default.
+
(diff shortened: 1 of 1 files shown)


docs/SECURITY.md
```diff
@@ -17,6 +17,9 @@ Authentication and Authorization. But to do so usually requires
 configuring your Operating System to restrict the operations that can be done.
 This is generally a good idea even if you plan on running your cluster with Auth.
 
+Meaning to say, Storm's OS level security is based on running Storm processes with proper OS account,
```
srdo: I think this is a little hard to understand. What do you think about "Storm's OS level security relies on running Storm processes using OS accounts that have only the permissions they need. Note that workers run under the same OS account as the Supervisor daemon by default"?

HeartSaVioR: I think that's clearer. Thanks for suggestion.

HeartSaVioR: Updated.

srdo:
+1

PR #2566, opened 2018-02-20
STORM-2965 Interpret wildcard in classpath correctly when reading con…

HeartSaVioR:
…fig from classpath

* also fix version pattern in SimpleVersion which parses version like 0.10.3 correctly

Manually tested with running Storm 1.2.1-rc1 topology.

@revans2 Please take a look. Thanks in advance.

### DIFF
--- a/storm-client/src/jvm/org/apache/storm/utils/Utils.java
+++ b/storm-client/src/jvm/org/apache/storm/utils/Utils.java
@@ -63,6 +63,8 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.zip.GZIPInputStream;
 import java.util.zip.GZIPOutputStream;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipFile;
 
 import com.google.common.collect.Lists;
 import org.apache.commons.io.FileUtils;
@@ -82,7 +84,6 @@ import org.apache.storm.generated.Nimbus;
 import org.apache.storm.generated.StormTopology;
 import org.apache.storm.generated.TopologyInfo;
 import org.apache.storm.generated.TopologySummary;
-import org.apache.storm.generated.WorkerToken;
 import org.apache.storm.security.auth.ReqContext;
 import org.apache.storm.serialization.DefaultSerializationDelegate;
 import org.apache.storm.serialization.SerializationDelegate;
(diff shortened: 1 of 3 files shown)


storm-client/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1505,35 +1508,36 @@ public static StormTopology addVersions(StormTopology topology) {
...
         for (String part: cp) {
-            File f = new File(part);
-            if (f.isDirectory()) {
-                if (defaultsConf == null) {
-                    defaultsConf = readConfIgnoreNotFound(yaml, new File(f, "defaults.yaml"));
-                }
-                
-                if (stormConf == null) {
-                    stormConf = readConfIgnoreNotFound(yaml, new File(f, "storm.yaml"));
+            String wildcardSuffix = File.separator + "*";
```
srdo: Wouldn't it also work to just check if the part ends with "*"?

storm-client/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1505,35 +1508,36 @@ public static StormTopology addVersions(StormTopology topology) {
...
+            if (part.endsWith(wildcardSuffix)) {
+                // wildcard is given
+                // in java classpath, '*' is translated to '*.jar'
+                File dir = new File(part.substring(0, part.length() - wildcardSuffix.length()));
```
srdo: Could we use Path for creating paths instead of manipulating strings? (asking because it's nice if the code is obviously portable)

storm-client/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1505,35 +1508,36 @@ public static StormTopology addVersions(StormTopology topology) {
...
+                File[] jarFiles = dir.listFiles((dir1, name) -> name.endsWith(".jar"));
+
+                if (jarFiles != null) {
```
srdo: Nit: Isn't there something really weird going on if this is null? The javadoc only mentions that this can be null if there's an IO error or the dir isn't a directory.

storm-client/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1570,4 +1574,49 @@ public static boolean isLocalhostAddress(String address) {
...
+
+        public Map<String, Object> getDefaultsConf() {
+            return defaultsConf;
+        }
+
+        public Map<String, Object> getStormConf() {
+            return stormConf;
+        }
+
+        public JarConfigReader invoke() throws IOException {
```
srdo: Nit: This method seems to be part of initializing this class. Consider making it private and calling it from the constructor.

storm-client/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1570,4 +1574,49 @@ public static boolean isLocalhostAddress(String address) {
...
+                    if (!entry.isDirectory()) {
+                        if (defaultsConf == null && entry.getName().equals("defaults.yaml")) {
+                            try (InputStream in = jarFile.getInputStream(entry)) {
+                                defaultsConf = (Map<String, Object>) yaml.load(new InputStreamReader(in));
+                            }
+                        }
+
+                        if (stormConf == null && entry.getName().equals("storm.yaml")) {
+                            try (InputStream in = jarFile.getInputStream(entry)) {
+                                stormConf = (Map<String, Object>) yaml.load(new InputStreamReader(in));
```
srdo: Nit: Consider moving the InputStreamReader into the try-with-resources, so it also gets closed.

storm-client/src/jvm/org/apache/storm/utils/Utils.java
HeartSaVioR: According to my understanding of Oracle Java classpath doc, it should only expand when file name is '*' rather than like 'abc*'. If 'abc*' is given in file name, it should not expand to abcde.jar or so.

https://docs.oracle.com/javase/8/docs/technotes/tools/unix/classpath.html

storm-client/src/jvm/org/apache/storm/utils/Utils.java
HeartSaVioR: I just get rid of manipulating string via File's method. Please take a look at new commit.

storm-client/src/jvm/org/apache/storm/utils/Utils.java
HeartSaVioR: The class is made from IntelliJ's extract feature. We are now calling readZip/readJar separately so hopefully that is OK to keep it public and be called from outside.

HeartSaVioR:
@srdo 
Addressed review comment. Please take a look again.

storm-client/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1559,8 +1577,8 @@ public static boolean isLocalhostAddress(String address) {
         Set<Integer> ids = srcMap.keySet();
         Integer largestId = ids.stream().max(Integer::compareTo).get();
         int end = largestId - start;
-        ArrayList<V> result = new ArrayList<>(Collections.nCopies(end+1 , null)); // creates array[largestId+1] filled with nulls
-        for( Map.Entry<Integer, V> entry : srcMap.entrySet() ) {
+        ArrayList<V> result = new ArrayList<>(Collections.nCopies(end + 1, null)); // creates array[largestId+1] filled with nulls
```
HeartSaVioR: Auto-corrected via IntelliJ. Let me revert if we don't want to fix unrelated thing.

storm-client/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1570,4 +1588,60 @@ public static boolean isLocalhostAddress(String address) {
...
+            return this;
+        }
+
+        private void readArchive(ZipFile zipFile) throws IOException {
+            Enumeration<? extends ZipEntry> zipEnums = zipFile.entries();
+            while (zipEnums.hasMoreElements()) {
+                ZipEntry entry = zipEnums.nextElement();
+                if (!entry.isDirectory()) {
+                    if (defaultsConf == null && entry.getName().equals("defaults.yaml")) {
+                        try (InputStream in = zipFile.getInputStream(entry); InputStreamReader isr = new InputStreamReader(in)) {
```
srdo: Nit: Since you're not using the in variable, you could builder the isr in one step

storm-client/src/jvm/org/apache/storm/utils/Utils.java
srdo: Thanks, hadn't thought of that. I was thinking of the case where the full path is "*", which I think you've fixed now.

storm-client/src/jvm/org/apache/storm/utils/Utils.java
srdo: Thanks, looks good

storm-client/src/jvm/org/apache/storm/utils/Utils.java
srdo: I'm not sure. If jarFiles is null, it looks like we'll end up returning an empty config, so I think downstream code will probably break. If that's the case, I'd prefer throwing an exception here so it's obvious what went wrong. Whether we want to throw a custom exception or just remove the null check and get an NPE in the following line makes no difference to me.

storm-client/src/jvm/org/apache/storm/utils/Utils.java
srdo: Looks good, thanks. Not sure if it makes a difference to open the jarfile with JarFile rather than ZipFile, but it can't hurt.

storm-client/src/jvm/org/apache/storm/utils/Utils.java
HeartSaVioR: Addressed.

HeartSaVioR:
@srdo Addressed another review comment.

storm-client/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1504,36 +1505,53 @@ public static StormTopology addVersions(StormTopology topology) {
...
+                File[] jarFiles = dir.listFiles((dir1, name) -> name.endsWith(".jar") || name.endsWith(".JAR"));
+
+                if (jarFiles != null) {
+                    for (File jarFile : jarFiles) {
+                        JarConfigReader jarConfigReader = new JarConfigReader(yaml, defaultsConf, stormConf, jarFile).readJar();
+                        defaultsConf = jarConfigReader.getDefaultsConf();
+                        stormConf = jarConfigReader.getStormConf();
+                    }
                 }
             } else {
```
srdo: Nit: Looks like this block doesn't need to be at another level, I think the method could be written as if ... else if ... else if ...

storm-client/src/jvm/org/apache/storm/utils/Utils.java
HeartSaVioR: Yes I also feel better to throw an exception. Throwing exception will not break Nimbus instance too.

HeartSaVioR:
@srdo 
Last commit addresses throwing Exception when dir.listFiles() returns null, and reduce indentation level.

srdo:
+1 again, thanks

PR #2569, opened 2018-02-20
STORM-2968: Exclude avro and commons-beanutils dependency from storm-autocreds

arunmahadevan:


HeartSaVioR:
Only integration-test failed.

PR #2571, opened 2018-02-21
[STORM-2969] remove @isDerivedFrom validation for topology.spout/bolt/backpressure.wait.strategy

Ethanlm:
https://issues.apache.org/jira/browse/STORM-2969

get rid of @isDerivedFrom validation for wait strategies so that users can supply their own strategies.


roshannaik:
Not clear why  `@isDerivedFrom`   wont be able to help validate inheritance of user defined classes from IWaitStrategy interface. Can you elaborate ?

Ethanlm:
@roshannaik  Sorry I am not clear. The topology submission will fail on validation because the user defined class is not present in nimbus's classpath. And we don't want to allow them to put their jars to nimbus' classpath for security reason.

roshannaik:
+1   
 Thank you.

PR #2577, opened 2018-02-26
[STORM-2977] BlowfishTupleSerializer tries to use zookeeper secret if blowfish.key is not present

Ethanlm:
https://issues.apache.org/jira/browse/STORM-2977

Tries to use Config.STORM_ZOOKEEPER_TOPOLOGY_AUTH_PAYLOAD before  throwing exceptions. Making it simpler to use


Ethanlm:
@HeartSaVioR @kishorvpatil  Could you please review it when you get a chance? Thanks very much!

PR #2580, opened 2018-02-27
STORM-2980: Update storm-starter docs concerning local mode

srdo:
https://issues.apache.org/jira/browse/STORM-2980

https://issues.apache.org/jira/browse/STORM-2447 made some changes to local mode, we should reflect these in the storm-starter documentation.


HeartSaVioR:
Only integration-test failed.

PR #2582, opened 2018-03-01
STORM-2984: Add jitter in executor heartbeat

zodvik:
The executorHeartbeatTimer is scheduled without jitter. In cases where state storage is Zookeeper and not pacemaker, introducing jitter will help spread out the I/O over time on Zookeeper. This will be helpful for large clusters.

The jitter time is `worker.heartbeat.frequency.secs`

### DIFF
--- a/storm-client/src/jvm/org/apache/storm/daemon/worker/Worker.java
+++ b/storm-client/src/jvm/org/apache/storm/daemon/worker/Worker.java
@@ -178,7 +178,8 @@ public class Worker implements Shutdownable, DaemonCommon {
             });
 
         workerState.executorHeartbeatTimer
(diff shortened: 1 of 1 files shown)


HeartSaVioR:
@zodvik 
Thanks for the contribution! Btw, we're planning to change the heartbeat mechanism to RPC which should reduce overall ZK load, so I'd rather rely on new mechanism. Please take a look at #2433.

I think the patch is still effective on 1.x version line. Could you raise a pull request on 1.x-branch?

zodvik:
@HeartSaVioR 
Makes sense. I'll close this PR.

HeartSaVioR:
@zodvik We may still want to apply this to 1.x version line. So if you struggled and the patch helped your case, please submit the patch against 1.x-branch.

PR #2583, opened 2018-03-02
STORM-2649 More detailed check of config serialization

ghajos:


storm-client/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1030,8 +1034,34 @@ private static Object normalizeConfValue(Object obj) {
...
+    @SuppressWarnings("unchecked")
+    public static boolean isValidConf(Map<String, Object> topoConfIn) {
+	Map<String, Object> origTopoConf = normalizeConf(topoConfIn);
+	Map<String, Object> deserTopoConf = normalizeConf(
+		(Map<String, Object>) JSONValue.parse(JSONValue.toJSONString(topoConfIn)));
+	return confMapDiff(origTopoConf, deserTopoConf);
+    }
+
+    @VisibleForTesting
+    static boolean confMapDiff(Map<String, Object> orig, Map<String, Object> deser) {
```
srdo: Nitpick: Consider renaming to something like "checkMapEquality", since this method checks map equality and returns a boolean indicating equality.

storm-client/test/jvm/org/apache/storm/utils/UtilsTest.java
```diff
@@ -173,4 +176,23 @@ public void isZkAuthenticationConfiguredStormServerWithPropertyTest() {
             }
         }
     }
+
+    @Test
+    public void testMapDiff() {
```
srdo: Nitpick: Consider splitting this into multiple small tests each checking one thing

storm-client/test/jvm/org/apache/storm/utils/UtilsTest.java
```diff
@@ -173,4 +176,23 @@ public void isZkAuthenticationConfiguredStormServerWithPropertyTest() {
...
+	Map<String, Object> map0 = ImmutableMap.of();
+	Assert.assertTrue("case0", Utils.confMapDiff(map0, map0));
+	Map<String, Object> map1 = ImmutableMap.of("k0", ImmutableList.of(1L, 2L), "k1", ImmutableSet.of('s', 'f'),
+		"k2", "as");
+	Assert.assertTrue("case1", Utils.confMapDiff(map1, map1));
+	Map<String, Object> map2 = ImmutableMap.of("k0", ImmutableList.of(1L, 2L), "k1", ImmutableSet.of('s', 'f'),
+		"k2", "as");
+	Assert.assertTrue("case2", Utils.confMapDiff(map2, map2));
```
srdo: Nit: This test is covered by line 186

srdo:
+1 again, looks great

storm-client/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1030,8 +1034,35 @@ private static Object normalizeConfValue(Object obj) {
...
+    @SuppressWarnings("unchecked")
+    public static boolean isValidConf(Map<String, Object> topoConfIn) {
+	Map<String, Object> origTopoConf = normalizeConf(topoConfIn);
+	Map<String, Object> deserTopoConf = normalizeConf(
+		(Map<String, Object>) JSONValue.parse(JSONValue.toJSONString(topoConfIn)));
+	return checkMapEquality(origTopoConf, deserTopoConf);
+    }
+
+    @VisibleForTesting
+    static boolean checkMapEquality(Map<String, Object> orig, Map<String, Object> deser) {
```
revans2: Please rename this as it is not generic Map Equality.  It is very specific to `isValidConf` and it is confusing to have a generic name for it.

storm-client/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1030,8 +1034,35 @@ private static Object normalizeConfValue(Object obj) {
...
+	MapDifference<String, Object> diff = Maps.difference(orig, deser);
```
revans2: We need to guard against deser being null.

We are calling `JSONValue.parse` which returns null if it cannot parse an object.  This here will end up throwing an NPE if deser is null.  It would probably be better to switch to `JSONValue.parseWithException` and catch the exception and handle it instead.

storm-client/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -1030,8 +1034,35 @@ private static Object normalizeConfValue(Object obj) {
...
+	if (diff.areEqual()) {
+	    return true;
+	}
+	for (Map.Entry<String, Object> entryOnLeft : diff.entriesOnlyOnLeft().entrySet()) {
+	    LOG.warn("Config property not serializable. Name: {} - Value: {}", entryOnLeft.getKey(),
+	            entryOnLeft.getValue());
+	}
+	for (Map.Entry<String, Object> entryOnRight : diff.entriesOnlyOnRight().entrySet()) {
```
revans2: I don't know of any situation where a key would be in one map but not the other map with the current JSON serialization/parsing. I am fine with leaving these checks in for completeness, but it might be good to change the warning messages to explain to the end user that something really odd happened with key "X", but because this should never happen in real life you might be on your own with debugging it.

storm-client/test/jvm/org/apache/storm/utils/UtilsTest.java
```diff
@@ -173,4 +176,49 @@ public void isZkAuthenticationConfiguredStormServerWithPropertyTest() {
             }
         }
     }
+
+    @Test
```
revans2: Can we have some tests for isValidConf in addition to checkMapEquality?

HeartSaVioR:
@ghajos kindly reminder.

ghajos:
Sorry for the late reply!

storm-client/test/jvm/org/apache/storm/utils/UtilsTest.java
ghajos: Found a positive test case in TestConfigValidate and added negative test cases. Should I collect some more test cases?

ghajos:
@srdo @HeartSaVioR @revans2 Thank you for the review and the patience!

PR #2585, opened 2018-03-04
STORM-2762 use guava collection where ever it is possible

ghajos:
Searched the code with git grep to find unnecessarily defined collection functions. I could only find these two.


storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
```diff
@@ -469,7 +470,7 @@ public IScheduler getForcedScheduler() {
             });
     }
 
-    private static <K, V> Map<K, V> mapDiff(Map<? extends K, ? extends V> first, Map<? extends K, ? extends V> second) {
+    private static <K, V> Map<K, V> mapValuesOnlyOnRight(Map<? extends K, ? extends V> first, Map<? extends K, ? extends V> second) {
```
HeartSaVioR: The method is checking the case if value for the key in second is different than the value for the same key in first (including the value for the same key in first doesn't exist). So while the `mapDiff` is really too ambiguous, `mapValuesOnlyOnRight` is not correct since it explains only a case for multiple cases.

If we don't find better name, fine to keep it as it is. Better name is always  welcomed!

ghajos:
@HeartSaVioR  @revans2  Thank you for the review!

PR #2590, opened 2018-03-10
STORM-2974: Add transactional non-opaque spout to storm-kafka-client

srdo:
https://issues.apache.org/jira/browse/STORM-2974

The changes outside the Trident package are because I'd like us to start using https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/consumer/MockConsumer.java for testing instead of our own mockito mocks. In order to do that we have to refer to the Consumer interface instead of KafkaConsumer. The replacement is a breaking change in a few places. I'm not expecting this to go on 1.x, and backward compatibility was already broken in https://github.com/apache/storm/pull/2300, so I haven't done any work to provide a smooth transition.

I've tried to follow the transactional spout in storm-kafka to implement this as much as possible.

There are a couple of followup tasks for this that I'll try to get done later, namely https://issues.apache.org/jira/browse/STORM-2990 and https://issues.apache.org/jira/browse/STORM-2991.


srdo:
I have tested the non-opaque spout by running the storm-kafka-client-examples topology, so testing hasn't been very extensive.

HeartSaVioR:
@srdo 
I'm sorry but could you please rebase this? Thanks in advance!

srdo:
@HeartSaVioR Rebased.

PR #2591, opened 2018-03-12
STORM-2979: WorkerHooks EOFException during run_worker_shutdown_hooks

hummelm:
[https://issues.apache.org/jira/browse/STORM-2979](https://issues.apache.org/jira/browse/STORM-2979)
when using a WorkerHook, shutdown of worker throws
java.lang.RuntimeException: java.io.EOFException 

This fix is against branch 1.0.x.

I don't think that my fix is breaking the U.T like reported by continuous-integration.
I will be happy to help integrating this patch against other branches if needed.

### DIFF
--- a/storm-core/src/clj/org/apache/storm/daemon/worker.clj
+++ b/storm-core/src/clj/org/apache/storm/daemon/worker.clj
@@ -566,9 +566,12 @@
         worker-topology-context (worker-context worker)
         hooks (.get_worker_hooks topology)]
     (dofor [hook hooks]
-      (let [hook-bytes (Utils/toByteArray hook)
-            deser-hook (Utils/javaDeserialize hook-bytes BaseWorkerHook)]
-        (.start deser-hook topo-conf worker-topology-context)))))
+      (let [savedposition (.position hook)
+            hook-bytes (Utils/toByteArray hook)
+            deser-hook (Utils/javaDeserialize hook-bytes BaseWorkerHook)
+            nop (.position hook savedposition)]
+        (.start deser-hook topo-conf worker-topology-context))
+      )))
 
 (defn run-worker-shutdown-hooks [worker]
   (let [topology (:topology worker)
--- a/storm-core/src/clj/org/apache/storm/daemon/worker.clj
+++ b/storm-core/src/clj/org/apache/storm/daemon/worker.clj
@@ -566,12 +566,10 @@
         worker-topology-context (worker-context worker)
         hooks (.get_worker_hooks topology)]
     (dofor [hook hooks]
-      (let [savedposition (.position hook)
-            hook-bytes (Utils/toByteArray hook)
-            deser-hook (Utils/javaDeserialize hook-bytes BaseWorkerHook)
-            nop (.position hook savedposition)]
-        (.start deser-hook topo-conf worker-topology-context))
-      )))
+      (let [duplicated-hook (.duplicate hook)
+            hook-bytes (Utils/toByteArray duplicated-hook)
+            deser-hook (Utils/javaDeserialize hook-bytes BaseWorkerHook)]
+        (.start deser-hook topo-conf worker-topology-context)))))
 
 (defn run-worker-shutdown-hooks [worker]
   (let [topology (:topology worker)
--- a/storm-core/src/clj/org/apache/storm/daemon/worker.clj
+++ b/storm-core/src/clj/org/apache/storm/daemon/worker.clj
@@ -285,6 +285,7 @@
       :executors executors
       :task-ids (->> receive-queue-map keys (map int) sort)
       :storm-conf storm-conf
+      :deserialized-worker-hooks (java.util.ArrayList.)
       :topology topology
       :system-topology (system-topology! storm-conf topology)
       :heartbeat-timer (mk-halting-timer "heartbeat-timer")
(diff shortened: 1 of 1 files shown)


hummelm:
After some tests I can say that the issue also affects:
1.1.X
1.2.X

I'm pretty sure that master branch (2.X)  is also affected but the worker code have been rewritten in java.
I can easily create a patch for this branch if needed.

hummelm:
After some digging I finally found a workaround.
Using a simple trick in the start method of the worker-hook (rewind of
the byteBuffer to allow the deserialization on stop)it can work as
expected, here is a simple example if someone encounter the same issue
:

package myTest.stormWorkerHook;

import java.nio.ByteBuffer;
import java.util.Map;

import org.apache.storm.Config;
import org.apache.storm.LocalCluster;
import org.apache.storm.generated.StormTopology;
import org.apache.storm.hooks.BaseWorkerHook;
import org.apache.storm.task.WorkerTopologyContext;
import org.apache.storm.trident.TridentState;
import org.apache.storm.trident.TridentTopology;
import org.apache.storm.trident.operation.BaseFunction;
import org.apache.storm.trident.operation.TridentCollector;
import org.apache.storm.trident.operation.builtin.Count;
import org.apache.storm.trident.testing.FixedBatchSpout;
import org.apache.storm.trident.testing.MemoryMapState;
import org.apache.storm.trident.tuple.TridentTuple;
import org.apache.storm.tuple.Fields;
import org.apache.storm.tuple.Values;
import org.apache.storm.utils.Utils;


public class App {

    static class myHook extends BaseWorkerHook {
        // store the size of the serialization of this class
        int mySize = 0;

        public int getMySize() {
            return mySize;
        }

        public void setMySize(int mySize) {
            this.mySize = mySize;
        }

        /**
         *
         */
        private static final long serialVersionUID = 1L;

         public void start(Map stormConf, WorkerTopologyContext context) {
               System.out.println("### START HOOK");
               ByteBuffer bf =
context.getRawTopology().get_worker_hooks().get(0);
               // getting the byte data for deserialization has move
the position
               // to the end of the buffer
               int pos = bf.position();
               // rewind the position of the buffer to allow
               // the deserialization which will occur on STOp
               bf.position(pos-mySize);

            }

            /**
             * This method is called right before a worker shuts down
             */
            @Override
            public void shutdown() {
                   System.out.println("### STOP HOOK");
            }

    }



    /*
     * Function takes the "sentence" field and emits a tuple for each word.
     */
    public static class Split extends BaseFunction {
        @Override
        public void execute(TridentTuple tuple, TridentCollector collector) {
            String sentence = tuple.getString(0);
            for(String word: sentence.split(" ")) {
                collector.emit(new Values(word));
            }
        }
    }

    // simple topo
    public static StormTopology buildTridentTopology() {
        /*
         * spout reads an infinite stream of sentences from the following source
         */
        @SuppressWarnings("unchecked")
        FixedBatchSpout spout = new FixedBatchSpout(new Fields("sentence"), 3,
                new Values("the cow jumped over the moon"),
                new Values("the man went to the store and bought some candy"),
                new Values("four score and seven years ago"),
                new Values("how many apples can you eat"),
                new Values("to be or not to be the person"));
        spout.setCycle(true);

        /*
         * TridentTopology object exposes the interface for
constructing Trident computations
         */
        TridentTopology topology = new TridentTopology();

        TridentState wordCounts =
                topology.newStream("spout1", spout)
                        .parallelismHint(16)
                        .each(new Fields("sentence"), new Split(), new
Fields("word"))
                        .groupBy(new Fields("word"))
                        .persistentAggregate(new
MemoryMapState.Factory(), new Count(), new Fields("count"))
                        .parallelismHint(16)


        ;
        StormTopology topo = topology.build();
        myHook mh =new myHook();
        // serialize one time to get the size of the byte[] result
        byte[] rawdata = Utils.javaSerialize(mh);
        // store the size inside the hook
        // it will be use to fix issue regarding byteBuffer
      mh.setMySize(rawdata.length );
      // serialize with the the size as attibute
      rawdata = Utils.javaSerialize(mh);

        ByteBuffer a = ByteBuffer.wrap(rawdata);
        topo.add_to_worker_hooks(a);

        return topo;

    }




    public static void main(String[] args) throws Exception {

        Config conf = new Config();
        conf.setMaxSpoutPending(20);

            LocalCluster cluster = new LocalCluster();
            cluster.submitTopology("wordCounter", conf, buildTridentTopology());
                Thread.sleep(10000);
                cluster.killTopology("wordCounter");
                Thread.sleep(10000);
                cluster.shutdown();

storm-core/src/clj/org/apache/storm/daemon/worker.clj
```diff
@@ -566,9 +566,12 @@
         worker-topology-context (worker-context worker)
         hooks (.get_worker_hooks topology)]
     (dofor [hook hooks]
-      (let [hook-bytes (Utils/toByteArray hook)
-            deser-hook (Utils/javaDeserialize hook-bytes BaseWorkerHook)]
-        (.start deser-hook topo-conf worker-topology-context)))))
+      (let [savedposition (.position hook)
```
HeartSaVioR: I guess we could do simpler: `(.duplicate hook)` to get duplicated ByteBuffer object, and read from that instance. It will not increase position from `hook` object.

hummelm: For sure this is a better solution.
Do I push a new commit implementing this or is there a better way to take into account your suggestion ?

Ethanlm: You can just push a new commit and squash the commits later

HeartSaVioR: I'm OK if you just make change to new squashed commit, since the diff will be one-liner or couple of lines. Sure I'm also OK to see new commit and squash later.

HeartSaVioR:
Yeah the valid patch should deserialize them only once while initializing worker state, and use them from both places (start and shutdown). Current implementation will not work properly if the implementation of worker hook is something stateful (since we deserialize and initialize two separate objects).

@hummelm 
Could you address this issue? You would want to craft a patch against master branch and 1.x-branch.

hummelm:
Yes i Will add a list of deserialized hook into the worker and use it to start.stop.

hummelm:
Here is a new implementation which is compatible with stateful workerhook.
I think this implementation is better isn't it ?

HeartSaVioR:
@hummelm 
Btw, we can't merge in this patch without PR for master branch, since master branch is the place which developments take place. Please let me know when you mind taking a look at master branch. I'll submit a patch then.

hummelm:
for master branch do I have to create a new pull request ?

HeartSaVioR:
@hummelm 
Yes since two branches are diverged so I can't port forward/back.

HeartSaVioR:
@hummelm 
Could you reflect new review comments? And please let me know your Apache JIRA ID so that I can assign the issue to you.

hummelm:
@HeartSaVioR 
Sorry what do you mean by reflect new review comments, is my last commit [https://github.com/apache/storm/pull/2591/commits/229ccaf8ea4d73b7efe0e2d7d6d6d239af97950f](https://github.com/apache/storm/pull/2591/commits/229ccaf8ea4d73b7efe0e2d7d6d6d239af97950f) not enough ? 
About my JIRA ID I don't know exactly what it is but may be "michel hummel" ?

HeartSaVioR:
@hummelm My bad. Missed 229ccaf. Please provide exact Apache JIRA ID to assign issue properly. You can log in to find exact ID. This is needed in receiving contribution step to assign you as contributor as well as leaving issue resolver.

HeartSaVioR:
Applied via c1a1c10dd7fa445148d543dbe410ec84e7670e82 and ported forward to other branches.
@hummelm Could you close this issue? It didn't close automatically since I squashed commits before pushing.

PR #2592, opened 2018-03-12
STORM-2993: Storm HDFS bolt throws ClosedChannelException when Timed rotation policy is used

arunmahadevan:
The TimedRotation should synchronize so that the bolt does not attempt to write to a stale writer.


HeartSaVioR:
@arunmahadevan Could we have a patch for master branch as well?

arunmahadevan:
@HeartSaVioR , thanks for reviewing. Raised https://github.com/apache/storm/pull/2610 for master.

PR #2593, opened 2018-03-17
STORM-2994 KafkaSpout doesn't  commit null tuples offsets

reiabreu:
Hello,

Let's kick off this pull request.
Unit tests for null tuples were missing. I'm in the process of adding them.
I'll update this PR asap. Meanwhile, these seem to be the necessary changes

### DIFF
--- a/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
+++ b/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
@@ -24,7 +24,7 @@ import static org.apache.storm.kafka.spout.KafkaSpoutConfig.FirstPollOffsetStrat
 import static org.apache.storm.kafka.spout.KafkaSpoutConfig.FirstPollOffsetStrategy.UNCOMMITTED_LATEST;
 
 import com.google.common.annotations.VisibleForTesting;
-import java.io.IOException;
+
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -49,7 +49,6 @@ import org.apache.kafka.common.errors.InterruptException;
 import org.apache.kafka.common.errors.RetriableException;
 import org.apache.storm.kafka.spout.KafkaSpoutConfig.FirstPollOffsetStrategy;
 import org.apache.storm.kafka.spout.KafkaSpoutConfig.ProcessingGuarantee;
-import org.apache.storm.kafka.spout.internal.CommitMetadata;
 import org.apache.storm.kafka.spout.internal.CommitMetadataManager;
 import org.apache.storm.kafka.spout.internal.KafkaConsumerFactory;
 import org.apache.storm.kafka.spout.internal.KafkaConsumerFactoryDefault;
@@ -280,7 +279,7 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
                 if (isAtLeastOnceProcessing()) {
                     commitOffsetsForAckedTuples(kafkaConsumer.assignment());
                 } else if (kafkaSpoutConfig.getProcessingGuarantee() == ProcessingGuarantee.NO_GUARANTEE) {
-                    Map<TopicPartition, OffsetAndMetadata> offsetsToCommit = 
+                    Map<TopicPartition, OffsetAndMetadata> offsetsToCommit =
                         createFetchedOffsetsMetadata(kafkaConsumer.assignment());
                     kafkaConsumer.commitAsync(offsetsToCommit, null);
                     LOG.debug("Committed offsets {} to Kafka", offsetsToCommit);
@@ -368,7 +367,7 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
                 numPolledRecords);
             if (kafkaSpoutConfig.getProcessingGuarantee() == KafkaSpoutConfig.ProcessingGuarantee.AT_MOST_ONCE) {
                 //Commit polled records immediately to ensure delivery is at-most-once.
-                Map<TopicPartition, OffsetAndMetadata> offsetsToCommit = 
+                Map<TopicPartition, OffsetAndMetadata> offsetsToCommit =
                     createFetchedOffsetsMetadata(kafkaConsumer.assignment());
                 kafkaConsumer.commitSync(offsetsToCommit);
                 LOG.debug("Committed offsets {} to Kafka", offsetsToCommit);
@@ -484,8 +483,11 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
                     return true;
                 }
             } else {
+                /*if a null tuple is not configured to be emitted, it should be marked as emitted and acked immediately
+                * to allow its offset to be commited to Kafka*/
                 LOG.debug("Not emitting null tuple for record [{}] as defined in configuration.", record);
-                msgId.setEmitted(false);
+                msgId.setNullTuple(true);
+                offsetManagers.get(tp).addToEmitMsgs(msgId.offset());
                 ack(msgId);
             }
         }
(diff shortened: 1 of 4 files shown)


external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -484,8 +484,11 @@ private boolean emitOrRetryTuple(ConsumerRecord<K, V> record) {
                     return true;
                 }
             } else {
+                /*if a null tuple is not configured to be emitted, it should be marked as emitted and acked immediately
+                * to allow its offset to be commited to Kafka*/
                 LOG.debug("Not emitting null tuple for record [{}] as defined in configuration.", record);
                 msgId.setEmitted(false);
```
srdo: I'd like to see this property renamed to "isFiltered" or "isNullTuple" something similar. Right now we have both a list of emitted tuples, and also the isEmitted property, and they don't always match. It seems unnecessarily confusing.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -576,6 +579,8 @@ public void ack(Object messageId) {
                     + "came from a topic-partition that this consumer group instance is no longer tracking "
                     + "due to rebalance/partition reassignment. No action taken.", msgId);
             } else {
+                //a null tuple should be added to the ack list since by definition is a direct ack
```
srdo: We should probably switch the order of the emitted contains and msgId isEmitted checks. It is not possible that isEmitted is false while the tuple is in the emitted list.

HeartSaVioR:
@reiabreu kindly reminder.

reiabreu:
Hi folks! I won't have access to a computer until next weekend. I'll pick
it up then. Thank you for understanding

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
reiabreu: I concur. I'll refactor it.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -570,20 +572,25 @@ public void ack(Object messageId) {
         // Only need to keep track of acked tuples if commits to Kafka are controlled by
         // tuple acks, which happens only for at-least-once processing semantics
         final KafkaSpoutMessageId msgId = (KafkaSpoutMessageId) messageId;
-        if (!emitted.contains(msgId)) {
-            if (msgId.isEmitted()) {
+        if (!msgId.isNullTuple()) {
```
srdo: Nit: You can reduce the nesting by a bit here by switching this to a guard clause, e.g.

```
if (is null tuple) {
  ...
  return
}
if (emitted contains) {
  ...
} else {
  ...
}
```

Up to you whether you want to change it

reiabreu: Sounds good, will update it.
Regarding tests, we need to add at least one: 

Test all messages are commited for all null tuples when Spout is not set to emit null tuples

```
@Test
    public void testShouldCommitAllMessagesIfNotSetToEmitNullTuples() throws Exception {
        final int messageCount = 10;
        prepareSpout(messageCount);

        //All null tuples should be commited, meaning they were considered by to be emitted and acked
        for(int i = 0; i < messageCount; i++) {
            spout.nextTuple();
        }

        verify(collectorMock,never()).emit(
                anyString(),
                anyList(),
                any());

        Time.advanceTime(commitOffsetPeriodMs + KafkaSpout.TIMER_DELAY_MS);
        //Commit offsets
        spout.nextTuple();

        verifyAllMessagesCommitted(messageCount);
    }
```
We would need a  new SingleTopicKafkaSpoutConfiguration with something like:

```
 private static class NullRecordExtractor implements RecordTranslator{
        
        @Override
        public List<Object> apply(ConsumerRecord record) {
            return null;

        }

        @Override
        public Fields getFieldsFor(String stream) {
            return new Fields("topic", "key", "value");
        }

        @Override
        public Object apply(Object record) {
            return null;
        }
    }
```

I was planning to extend KafkaSpoutAbstractTest on something similar to KafkaSpoutSingleTopicTest.

srdo: Great, that sounds good.

reiabreu:
KafkaSpoutAbstractTest is tightly coupled to  KafkaSpoutConfig through createSpoutConfig, meaning that to test a single configuration change, we need to create a new test class. This is something that we can redesign in the future.

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/KafkaSpoutNullTupleTest.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+
+    @Override
+    KafkaSpoutConfig<String, String> createSpoutConfig() {
+        return SingleTopicKafkaSpoutConfiguration.setCommonSpoutConfigNullTuples(
+            KafkaSpoutConfig.builder("127.0.0.1:" + kafkaUnitRule.getKafkaUnit().getKafkaPort(),
+                Pattern.compile(SingleTopicKafkaSpoutConfiguration.TOPIC)))
+            .setOffsetCommitPeriodMs(commitOffsetPeriodMs)
+            .setRetry(new KafkaSpoutRetryExponentialBackoff(KafkaSpoutRetryExponentialBackoff.TimeInterval.seconds(0), KafkaSpoutRetryExponentialBackoff.TimeInterval.seconds(0),
```
srdo: Nit: Unless it's important to the test, I'd leave this config out or use the SingleTopicKafkaSpoutConfiguration utility to set common configuration. Same for maxPollRecords.

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/KafkaSpoutNullTupleTest.java
```diff
@@ -0,0 +1,76 @@
...
+    private final int maxRetries = 3;
+
+    public KafkaSpoutNullTupleTest() {
+        super(2_000);
+    }
+
+
+    @Override
+    KafkaSpoutConfig<String, String> createSpoutConfig() {
+        return SingleTopicKafkaSpoutConfiguration.setCommonSpoutConfigNullTuples(
```
srdo: I think you can simplify this by using SingleTopicKafkaSpoutConfiguration.createKafkaSpoutConfigBuilder, and simply setting the record translator. I don't think there's a reason to put a whole new configuration in SingleTopicKafkaSpoutConfiguration. I'd also rather keep the record translator local to this class, since we're not using it in other tests.

reiabreu: Absolutely, will simplify it

srdo:
+1, thanks for your patience. Please squash to one commit, and we can merge.

reiabreu:
Thank you for guiding me through the changes. I've squashed all the commits and pushed a fresh one

srdo:
@reiabreu Merged to master. Looks like there are some small modifications to do to get it on to 1.x or earlier branches. Could you put up a PR against 1.x as well (should be pretty easy with `git cherry-pick`)?

reiabreu:
@srdo Sure, I'll have a look

PR #2594, opened 2018-03-18
storm-kafka: encapsulated Broker.host and port

krichter722:
This satisfies the checkstyle check.


srdo:
The changes look good, but I'm wondering whether it makes sense to make this kind of improvement. We're planning to remove storm-kafka from master soon, we're just waiting on a couple of other issues to resolve first https://issues.apache.org/jira/browse/STORM-2953. Unless storm-kafka ends up not getting removed, this change might end up not getting released.

HeartSaVioR:
I second @srdo, and we'd better to handle removal of storm-kafka sooner to avoid such case.

PR #2598, opened 2018-03-19
storm-kafka: added ExponentialBackoffMsgRetryManagerSpoutConfig

krichter722:
Only `ExponentialBackoffMsgRetryManager` uses `SpoutConfig.retryInitialDelayMs`, `retryDelayMultiplier`, `retryDelayMaxMs` and `retryLimit` which means that those properties should be moved to a subclass `ExponentialBackoffMsgRetryManagerSpoutConfig` and the type be enforced through class parameters. Since `ExponentialBackoffMsgRetryManager` is the only `FailedMsgRetryManager` is the only implementation there's no need to make excessive use of class parameter since that flexibility is YAGNI.

Furthermore the properties are declared redundantly in `ExponentialBackoffMsgRetryManager` and `SpoutConfig`.

I don't seem to figure out how to find which new checkstyle violations the PR introduces, so I had to guess which changes caused the temporary increase in violations.

The larger portion of commented out code is left there since checkstyle value for NumberParameter of 7 seems overkill and at least constructors should be allowed to be much longer, we should talk about that on the mailing list.

The initialization of `FailedMsgRetryManager` through reflection in undocumented and seem unnecessary.

I got the note that apache-kafka might be removed from the project. I'm working with this code and want to share my work. I'd be happy if you take the time for review.

### DIFF
--- a/external/storm-kafka/src/jvm/org/apache/storm/kafka/ExponentialBackoffMsgRetryManager.java
+++ b/external/storm-kafka/src/jvm/org/apache/storm/kafka/ExponentialBackoffMsgRetryManager.java
@@ -26,26 +26,35 @@ import java.util.Queue;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 
-public class ExponentialBackoffMsgRetryManager implements FailedMsgRetryManager {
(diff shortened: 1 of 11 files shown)


srdo:

PR #2599, opened 2018-03-19
[STORM-2996] Inaccurate description about offset and kafkaspout will make users confused

liu-zhaokun:
[https://issues.apache.org/jira/browse/STORM-2996](https://issues.apache.org/jira/browse/STORM-2996)
As you can see in the picture,we don't know which partition the offset belongs in first line,so I think we can log the partition information.And consumer.host is the same as host in Partition in second line,so I think we can remove it.


liu-zhaokun:
@HeartSaVioR 
Could you help me review this PR?

srdo:

PR #2601, opened 2018-03-19
[STORM-2998]Wrong className in LoggerFactory.getLogger method

liu-zhaokun:
[https://issues.apache.org/jira/browse/STORM-2998](https://issues.apache.org/jira/browse/STORM-2998)
If we use LoggerFactory.getLogger method in a class,the most accurate approach is use this class as a parameter.


srdo:
+1

srdo:
Thanks @liu-zhaokun, merged to master

PR #2604, opened 2018-03-24
STORM-3006-updated-DRPC-docs

rauluka:
Extended DRPCClient initialization with DRPCClient config (which is now mandatory parameter) and added storm.thrift.transport: "org.apache.storm.security.auth.plain.PlainSaslTransportPlugin" in storm.yaml example so it matches DRPCClient config.


revans2:
The docs are not identical for master, so I will manually make some updates to master, which is allowed without code review.

PR #2607, opened 2018-03-27
STORM-3011 Use default bin path in flight.bash if $JAVA_HOME  is undefined

jnioche:


jnioche:
thanks, see #2609

jnioche:
thanks @revans2

PR #2608, opened 2018-03-27
[STORM-3012] Fix the issue that nimbus will crash if pacemaker is restarted

Ethanlm:
by returning results from resending HBMessage

Detailed explanation at  https://issues.apache.org/jira/browse/STORM-3012


agresch:
Seems ok to me.  I might also add handling null to PaceMakerStateStorage.java#L195.

Ethanlm:
@agresch  This change makes sure that the responses won't contain any `null` HBMessage. So I think this should be fine.

Ethanlm:
@revans2  Filed a jira https://issues.apache.org/jira/browse/STORM-3017

PR #2611, opened 2018-03-29
STORM-2924 Fix intermittent test failure in org.apache.storm.TickTupl…

HeartSaVioR:
…eTest

* leverages STORM_TEST_TIMEOUT_MS to make timeout being flexible

This change effectively increase test timeout to 150_000 ms described in `travis-script.sh`.


HeartSaVioR:
@srdo Yes it succeeds 3 times in a row, and I can run more if we would like to see more.

PR #2619, opened 2018-04-01
[STORM-3018] Fix integration test DemoTest#testExclamationTopology fa…

danny0405:
For [STORM-2693](https://github.com/apache/storm/pull/2433) i changed the task metrics reporting interval default to 60 seconds, which is same with the DemoTest#testExclamationTopology check time.

So fix it here.

See [STORM-3018](https://issues.apache.org/jira/projects/STORM/issues/STORM-3018?filter=allopenissues)

### DIFF
--- a/conf/defaults.yaml
+++ b/conf/defaults.yaml
@@ -196,6 +196,9 @@ worker.log.level.reset.poll.secs: 30
 topology.worker.receiver.thread.count: 1
 
 # Executor metrics reporting interval.
+# Cause the ui only show built in metrics, we should keep sync with the built in metrics interval,
+# also the metrics consumer's collecting interval.
+# See topology.builtin.metrics.bucket.size.secs and storm.cluster.metrics.consumer.publish.interval.secs.
 executor.metrics.frequency.secs: 60
 
 task.heartbeat.frequency.secs: 3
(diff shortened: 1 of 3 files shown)


conf/defaults.yaml
```diff
@@ -195,7 +195,7 @@ worker.log.level.reset.poll.secs: 30
 # control how many worker receiver threads we need per worker
 topology.worker.receiver.thread.count: 1
 
-# Executor metrics reporting interval.
+# Executor metrics reporting interval, keep default val same with topology.builtin.metrics.bucket.size.secs
```
srdo: Nit: Could you add to the comment the reason why these must be kept in sync?

danny0405: Because:
1. the internal metrics interval is set up by conf: topology.builtin.metrics.bucket.size.secs, we should keep sync with it
2. Now use ZK to store our metrics and the ui only show built in metrics, the metrics consumer's collecting interval are also default to 60 seconds

srdo: Thanks, but I meant add a note to the comment. I can't tell from the current comment why the two properties need to be in sync, so I'd have to look up this PR to see why.

danny0405: Ok, got your idea.

PR #2622, opened 2018-04-04
STORM-3020: fix possible race condition in AsyncLocalizer

revans2:
There were a number of places in AsyncLocalizer that were using synchronized to try and protect some maps.  When we added in support for restarting a worker if specific blobs change one of the places that was protected before moved to a background thread and lost that protection.

Now that we are java 8 we can use lambdas to provide the same protection without the need for the locks so I removed all of the course grained locking in AsyncLocalizer.  Now the only locking is on a per-blob basis.

I tested this manually by launching topologies with lots of different types of blobs, modified the blobs and verified that everything worked correctly.  I shot things, including the supervisor, various times and verified that it was able to recover in each case.

### DIFF
--- a/storm-server/src/main/java/org/apache/storm/localizer/AsyncLocalizer.java
+++ b/storm-server/src/main/java/org/apache/storm/localizer/AsyncLocalizer.java
@@ -75,10 +75,13 @@ public class AsyncLocalizer implements AutoCloseable {
 
     private final boolean isLocalMode;
     // track resources - user to resourceSet
-    protected final ConcurrentMap<String, ConcurrentMap<String, LocalizedResource>> userFiles = new ConcurrentHashMap<>();
-    protected final ConcurrentMap<String, ConcurrentMap<String, LocalizedResource>> userArchives = new ConcurrentHashMap<>();
+    //ConcurrentHashMap is explicitly used everywhere in this class because it uses locks to guarantee atomicity for compute and
+    // computeIfAbsent where as ConcurrentMap allows for a retry of the function passed in, and would require the function to have
+    // no side effects.
+    protected final ConcurrentHashMap<String, ConcurrentHashMap<String, LocalizedResource>> userFiles = new ConcurrentHashMap<>();
+    protected final ConcurrentHashMap<String, ConcurrentHashMap<String, LocalizedResource>> userArchives = new ConcurrentHashMap<>();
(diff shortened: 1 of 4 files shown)


danny0405:
I test it, and it worked ok as before, but a KeyNotFoundException still got of master when finally cleaned the base blobs.

revans2:
Thanks for the review @danny0405 

This was not trying to fix STORM-2905.  This was a separate race condition I found when reviewing your pull request for STORM-2905.

HeartSaVioR:
@revans2 
Could you elaborate how protection works?

At the first time I thought it leverages atomicity of compute*, but reading description of compute* in Map, looks like guaranteeing atomicity is not forced to implementations.

https://docs.oracle.com/javase/8/docs/api/java/util/Map.html#compute-K-java.util.function.BiFunction-

and HashMap doesn't mention about atomicity (`topologyBlobs` is a HashMap):

https://docs.oracle.com/javase/8/docs/api/java/util/HashMap.html#compute-K-java.util.function.BiFunction-

whereas ConcurrentHashMap mentions about atomicity so at least this guarantees atomicity:

https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html#compute-K-java.util.function.BiFunction-

It may worth noting that implementation requirement in ConcurrentMap looks like a bit different from what ConcurrentHashMap provides. `The default implementation may retry these steps when multiple threads attempt updates including potentially calling the remapping function multiple times.` It doesn't mention locking but retrying.

https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentMap.html#compute-K-java.util.function.BiFunction-

revans2:
@HeartSaVioR 

Great catch I forgot to update the normal has Map/HashMap to a ConcurrentHashMap.  Yes the guarantees of ConcurrentMap allow for retry and we do have side effects in some of the computes.

I will update the comments as well to say what requirements we have for the type.  The computeIfAbsent methods have no side effects, so we don't need to worry about them as much.  I'll see if I can come up with a way to make it so we don't need as strong of a guarantee for `compute` so perhaps we could use other implementations.

revans2:
@HeartSaVioR thanks for the review.  I fixed your concerns by making them all ConcurrentHashMaps and adding a note about why they need to be that.  I could not find a good way to remove the sideeffects.

PR #2631, opened 2018-04-11
[STORM-3025] Optimize Cluster methods with Caching to avoid loopy loops 

kishorvpatil:
Optimizing the `Cluster` methods that are called from with high frequency calls to speed-up scheduling time on new topologies on large clusters.

### DIFF
--- a/storm-server/src/main/java/org/apache/storm/scheduler/resource/normalization/NormalizedResourceRequest.java
+++ b/storm-server/src/main/java/org/apache/storm/scheduler/resource/normalization/NormalizedResourceRequest.java
@@ -187,29 +187,4 @@ public class NormalizedResourceRequest implements NormalizedResourcesWithMemory 
     public NormalizedResources getNormalizedResources() {
         return this.normalizedResources;
     }
-
-    public void removeOffHeap(final double offHeap) {
-        this.offHeap += offHeap;
-    }
-
-    public void remove(WorkerResources value) {
-        this.normalizedResources.remove(value);
-        //The resources are already normalized
-        Map<String, Double> resources = value.get_resources();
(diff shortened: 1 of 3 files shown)


storm-server/src/main/java/org/apache/storm/scheduler/resource/normalization/NormalizedResourceRequest.java
```diff
@@ -187,4 +187,29 @@ public double getTotalCpu() {
     public NormalizedResources getNormalizedResources() {
         return this.normalizedResources;
     }
+
+    public void removeOffHeap(final double offHeap) {
```
agresch: where is this used?  Why increment?

storm-server/src/main/java/org/apache/storm/scheduler/resource/normalization/NormalizedResourceRequest.java
```diff
@@ -187,4 +187,29 @@ public double getTotalCpu() {
...
+        this.offHeap += offHeap;
+    }
+
+    public void remove(WorkerResources value) {
```
agresch: were these used?

storm-server/src/main/java/org/apache/storm/scheduler/resource/normalization/NormalizedResourceRequest.java
kishorvpatil: This is unused method. I was trying some thing else. Let me delete this method.

storm-server/src/main/java/org/apache/storm/scheduler/resource/normalization/NormalizedResourceRequest.java
kishorvpatil: Not used

kishorvpatil:
@agresch I have addressed the code review comments.

PR #2633, opened 2018-04-12
STORM-3028 HdfsSpout: handle empty file in case of ackers

ghajos:
There is a problem with all the storm-hdfs junit tests. The test results may not be valid.

### DIFF
--- a/external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/HdfsSpout.java
+++ b/external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/HdfsSpout.java
@@ -101,7 +101,6 @@ public class HdfsSpout extends BaseRichSpout {
     private final AtomicBoolean commitTimeElapsed = new AtomicBoolean(false);
     private Timer commitTimer;
     private boolean fileReadCompletely = true;
-    private boolean newReader = false;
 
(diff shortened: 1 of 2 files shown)


ghajos:
@srdo Can you please take a look at?

revans2:
@ghajos any plans to address the question from @srdo

ghajos:
@revans2 @srdo I'm sorry! There is no reason to set both commit frequency count and commit frequency sec. I hope that this is already addressed in the last commit.

external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/HdfsSpout.java
```diff
@@ -221,14 +221,18 @@ public void nextTuple() {
...
                     reader = pickNextFile();
                     if (reader == null) {
                         LOG.debug("Currently no new files to process under : " + sourceDirPath);
                         return;
                     } else {
                         fileReadCompletely = false;
+                        newReader = true;
                     }
+                } else {
+                    newReader = false;
```
srdo: Nit: This branch is a no op, newReader is already false

ghajos: Thank you for pointing this out!

PR #2634, opened 2018-04-13
[STORM-3021] Fix wrong usages of Config.TOPOLOGY_WORKERS on RAS cluster

Ethanlm:
https://issues.apache.org/jira/browse/STORM-3021

The Resource Aware Scheduler doesn't honor `Config.TOPOLOGY_WORKERS`. So we need to fix the code where it uses `Config.TOPOLOGY_WORKERS` incorrectly.  Also fix docs/javadocs to avoid confusion.

### DIFF
--- a/docs/Resource_Aware_Scheduler_overview.md
+++ b/docs/Resource_Aware_Scheduler_overview.md
@@ -184,6 +184,10 @@ The user can set some default configurations for the Resource Aware Scheduler in
     topology.worker.max.heap.size.mb: 768.0
 ```
 
+### Warning
+
+If Resource Aware Scheduling is enabled, it will dynamically calculate the number of workers and the `topology.workers` setting is ignored.
+
 <div id='Topology-Priorities-and-Per-User-Resource'/>
 
 ## Topology Priorities and Per User Resource 
--- a/storm-client/src/jvm/org/apache/storm/Config.java
+++ b/storm-client/src/jvm/org/apache/storm/Config.java
@@ -180,7 +180,8 @@ public class Config extends HashMap<String, Object> {
     /**
      * How many processes should be spawned around the cluster to execute this topology. Each process will execute some number of tasks as
      * threads within them. This parameter should be used in conjunction with the parallelism hints on each component in the topology to
-     * tune the performance of a topology.
+     * tune the performance of a topology. The number of workers will be dynamically calculated when the Resource Aware scheduler is used,
+     * in which case this parameter will not be honored.
      */
     @isInteger
     @isPositiveNumber
@@ -331,9 +332,10 @@ public class Config extends HashMap<String, Object> {
     /**
      * How many executors to spawn for ackers.
      *
-     * <p>By not setting this variable or setting it as null, Storm will set the number of acker executors
-     * to be equal to the number of workers configured for this topology. If this variable is set to 0, then Storm will immediately ack
-     * tuples as soon as they come off the spout, effectively disabling reliability.</p>
+     * <p>By not setting this variable or setting it as null, Storm will set the number of acker executors to be equal to
+     * the number of workers configured for this topology (or the estimated number of workers if the Resource Aware Scheduler is used).
+     * If this variable is set to 0, then Storm will immediately ack tuples as soon as they come off the spout,
+     * effectively disabling reliability.</p>
      */
     @isInteger
     @isPositiveNumber(includeZero = true)
@@ -351,8 +353,9 @@ public class Config extends HashMap<String, Object> {
     /**
      * How many executors to spawn for event logger.
      *
-     * <p>By not setting this variable or setting it as null, Storm will set the number of eventlogger executors
-     * to be equal to the number of workers configured for this topology. If this variable is set to 0, event logging will be disabled.</p>
+     * <p>By setting it as null, Storm will set the number of eventlogger executors to be equal to the number of workers
+     * configured for this topology (or the estimated number of workers if the Resource Aware Scheduler is used).
+     * If this variable is set to 0, event logging will be disabled.</p>
      */
     @isInteger
     @isPositiveNumber(includeZero = true)
--- a/storm-client/src/jvm/org/apache/storm/daemon/worker/Worker.java
+++ b/storm-client/src/jvm/org/apache/storm/daemon/worker/Worker.java
@@ -231,10 +231,8 @@ public class Worker implements Shutdownable, DaemonCommon {
         executorsAtom.set(newExecutors);
 
         // This thread will send out messages destined for remote tasks (on other workers)
-        // If there are no remote tasks, don't start the thread.
-        Set<Integer> remoteTasks = Sets.difference(new HashSet<>(workerState.getOutboundTasks()),
-                                                   new HashSet<>(workerState.getLocalTaskIds()));
-        if (!remoteTasks.isEmpty()) {
+        // If there are no remote outbound tasks, don't start the thread.
+        if (workerState.hasRemoteOutboundTasks()) {
             transferThread = workerState.makeTransferThread();
             transferThread.setName("Worker-Transfer");
         }
@@ -313,8 +311,7 @@ public class Worker implements Shutdownable, DaemonCommon {
     }
 
     private void setupBackPressureCheckTimer(final Map<String, Object> topologyConf) {
-        final Integer workerCount = ObjectReader.getInt(topologyConf.get(Config.TOPOLOGY_WORKERS));
-        if (workerCount <= 1) {
+        if (workerState.isSingleWorker()) {
             LOG.info("BackPressure change checking is disabled as there is only one worker");
             return;
         }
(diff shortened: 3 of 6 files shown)


storm-client/src/jvm/org/apache/storm/daemon/worker/Worker.java
```diff
@@ -305,8 +305,9 @@ private void setupFlushTupleTimer(final Map<String, Object> topologyConf, final
     }
 
     private void setupBackPressureCheckTimer(final Map<String, Object> topologyConf) {
```
agresch: remove unused topologyConf

Ethanlm: `topologyConf`  used in this function

agresch: My bad.

storm-client/src/jvm/org/apache/storm/Config.java
```diff
@@ -372,9 +374,10 @@
     /**
      * How many executors to spawn for ackers.
      *
-     * <p>By not setting this variable or setting it as null, Storm will set the number of acker executors
-     * to be equal to the number of workers configured for this topology. If this variable is set to 0,
-     * then Storm will immediately ack tuples as soon as they come off the spout, effectively disabling reliability.</p>
+     * <p>By not setting this variable or setting it as null, Storm will set the number of acker executor to be equal to
```
erikdw: nit: please restore pluralization of `executor`.  i.e., `number of acker executors`.

Ethanlm:
@revans2  @roshannaik  Could you please take a look if you get a chance as we discussed about this in https://issues.apache.org/jira/browse/STORM-3021

docs/Resource_Aware_Scheduler_overview.md
```diff
@@ -184,6 +184,10 @@ The user can set some default configurations for the Resource Aware Scheduler in
     topology.worker.max.heap.size.mb: 768.0
 ```
 
+### Warning
+
+The number of workers will be dynamically calculated by the Resource Aware Scheduler. The `Config.TOPOLOGY_WORKERS` will not be honored.
```
roshannaik: May I suggest a minor rewording:  "If Resource Aware Scheduling is enabled, it will dynamically calculate the number of workers and the `topology.workers` setting is ignored."

storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
```diff
@@ -3032,9 +3019,19 @@ public void submitTopologyWithOpts(String topoName, String uploadedJarLocation,
...
-            totalConfToSave.put(Config.TOPOLOGY_ACKER_EXECUTORS, getNumOfAckerExecs(totalConf, topology));
-            LOG.debug("Config.TOPOLOGY_ACKER_EXECUTORS set to: {}", totalConfToSave.get(Config.TOPOLOGY_ACKER_EXECUTORS));
+            // if the Resource Aware Scheduler is used,
+            // we might need to set the number of acker executors and eventlogger executors to be the estimated number of workers.
+            if (ServerUtils.isRAS(conf)) {
+                int estimatedNumWorker = ServerUtils.getEstimatedWorkerCountForRASTopo(totalConf, topology);
+                int numAckerExecs = ObjectReader.getInt(totalConf.get(Config.TOPOLOGY_ACKER_EXECUTORS), estimatedNumWorker);
+                int numEventLoggerExecs = ObjectReader.getInt(totalConf.get(Config.TOPOLOGY_EVENTLOGGER_EXECUTORS), estimatedNumWorker);
+
+                totalConfToSave.put(Config.TOPOLOGY_ACKER_EXECUTORS, numAckerExecs);
```
roshannaik: Is this a case where we are overriding the acker & eventLogger count settings that are set by user ?

storm-client/src/jvm/org/apache/storm/daemon/worker/Worker.java
```diff
@@ -305,8 +305,9 @@ private void setupFlushTupleTimer(final Map<String, Object> topologyConf, final
...
-        final Integer workerCount = ObjectReader.getInt(topologyConf.get(Config.TOPOLOGY_WORKERS));
-        if (workerCount <= 1) {
+        Set<Integer> nonLocalTasks = Sets.difference(workerState.getTaskToComponent().keySet(),
```
roshannaik: Perhaps we can consolidate these set difference calls to count single/multiworker mode into a single function ? 
- https://github.com/apache/storm/blob/master/storm-client/src/jvm/org/apache/storm/daemon/worker/Worker.java#L230

storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
roshannaik: Some other places that may need similar fixes:

- https://github.com/apache/storm/blob/master/examples/storm-loadgen/src/main/java/org/apache/storm/loadgen/CaptureLoad.java#L109

- https://github.com/apache/storm/blob/master/storm-client/src/jvm/org/apache/storm/daemon/StormCommon.java#L363-L364  

- https://github.com/apache/storm/blob/master/storm-client/src/jvm/org/apache/storm/daemon/StormCommon.java#L276 


may want to check for all usages of that setting in code base.

storm-client/src/jvm/org/apache/storm/daemon/worker/Worker.java
roshannaik: Some other places that may need similar fixes:

- https://github.com/apache/storm/blob/master/examples/storm-loadgen/src/main/java/org/apache/storm/loadgen/CaptureLoad.java#L109

- https://github.com/apache/storm/blob/master/storm-client/src/jvm/org/apache/storm/daemon/StormCommon.java#L363-L364  

- https://github.com/apache/storm/blob/master/storm-client/src/jvm/org/apache/storm/daemon/StormCommon.java#L276 


may want to check for all usages of that setting in code base.

docs/Resource_Aware_Scheduler_overview.md
revans2: +1 for the comment from @roshannaik

storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
revans2: No we are not overriding the user setting.  We are overriding the default value.

Ethanlm: +1

storm-client/src/jvm/org/apache/storm/daemon/worker/Worker.java
Ethanlm: About the `set difference calls`, I thought these two are different. The first one https://github.com/apache/storm/blob/master/storm-client/src/jvm/org/apache/storm/daemon/worker/Worker.java#L230 is to check if there is any outbound task not in local worker. This doesn't necessarily mean the topology has only one worker. 
The second (in setupBackPressureCheckTimer) is to check if all the tasks are in local which means there is only one worker.

Ethanlm: https://github.com/apache/storm/blob/master/examples/storm-loadgen/src/main/java/org/apache/storm/loadgen/CaptureLoad.java#L109 This is fine because the code also gets the num of workers from topologyInfo.

https://github.com/apache/storm/blob/master/storm-client/src/jvm/org/apache/storm/daemon/StormCommon.java#L363-L364 this is also fine because we override the default value of `TOPOLOGY_EVENTLOGGER_EXECUTORS` when we submit the topology if it's on RAS.
Similar to https://github.com/apache/storm/blob/master/storm-client/src/jvm/org/apache/storm/daemon/StormCommon.java#L276

roshannaik: Thanks for clarifications. IMO, thats a nice (eventhough small) improvement to disable interworker when there are no outbound tasks... and not just for 1worker mode. 

Minor request: would be nice to have named functions for those two set difference calls. I suspect they will start seeing more usage soon.

looks good thanks for taking care of this.

Ethanlm: `Sets.difference()` is already a named function and it's almost one line of code. I think it's OK to just use it in this way?

Ethanlm:
The failed test should be unrelated

storm-client/src/jvm/org/apache/storm/daemon/worker/Worker.java
roshannaik: I meant ... would be nice to have functions like  isSingleWorkerMode() and hasOutboundTasks() that hide the detail of whatever mechanics are being employed to figure it out. I suspect they these functions will see more usage.

Ethanlm:
@roshannaik I addressed your comments. Thanks for review. If it looks good, I will squash the commits and wait for merge

Ethanlm:
rebased.

Ethanlm:
Not sure why the build failed. The build succeeded on my repo: https://travis-ci.org/Ethanlm/storm/builds/372073451

PR #2638, opened 2018-04-19
[STORM-3034] Adding exception stacktrace for executor failures in worker

kishorvpatil:


storm-client/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -109,6 +109,8 @@
 
 import javax.security.auth.Subject;
 
+import static org.apache.commons.lang.exception.ExceptionUtils.*;
```
Ethanlm: nit: don't use *

srdo: +1. A little odd that checkstyle didn't catch this, since it's in the rules https://github.com/apache/storm/blob/master/storm-checkstyle/src/main/resources/storm/storm_checkstyle.xml#L82. @kishorvpatil could you check if maybe the number of allowed style violations for storm-client can be reduced?

storm-client/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -361,6 +363,7 @@ public void run() {
                             Time.sleep(s);
                     }
                 } catch (Throwable t) {
+                    LOG.info("Async loop Exception Stacktrace is: {} ", getStackTrace(t));
```
srdo: Why isn't the stack trace available from the log in line 372?

kishorvpatil: The method can return at line 370 without reaching 372. We had instances where users could not understand root cause of the interrupt exception.

srdo: Thanks. Wouldn't it make sense to add the throwable to the log statement there then, so both L370 and 372 print the exception? (I'm asking why we want a separate log for this stack trace)

storm-client/src/jvm/org/apache/storm/utils/Utils.java
kishorvpatil: Updated.

kishorvpatil:
@srdo  Moved the `LOG.error` before changing exception Cause.

srdo:
Sorry if I'm nitpicking you to death, but I don't think moving the log line is better. Now everyone gets an error level log, even if the exception is due to an interrupt. I'd prefer if we just include the exception in both this log statement and the one in the interrupt case.

srdo:
+1, though I'm not sure whether it is necessary to change the log level to error.

storm-client/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -363,7 +358,7 @@ public void run() {
                 } catch (Throwable t) {
                     if (Utils.exceptionCauseIsInstanceOf(
                             InterruptedException.class, t)) {
-                        LOG.info("Async loop interrupted!");
+                        LOG.error("Async loop interrupted!", t);
```
HeartSaVioR: Here we assume that it is not an error, so if my understanding is right, `info` level would make more sense. (I'd second @srdo.)

kishorvpatil:
@HeartSaVioR @srdo ., The issue in assuming that `InterruptedException` is only coming from external interruptions. Below is the actual example we noticed when Kafka Spout`TimeoutException` was wrapped in `InterruptedException`. Users could not identify where the exception was raised.

```
KafkaSpoutI_[39 39] [WARN] Expecting exception of class: class java.lang.InterruptedException, but exception chain only contains: (#<TimeoutException org.apache.kafka.common.errors.TimeoutException: Failed to get offsets by times in 40000 ms>)```

srdo:
@kishorvpatil Can you elaborate a bit? I don't understand why `TimeoutException` would be wrapped in an `InterruptedException`? Where is the log line you posted coming from, a grep of Storm didn't turn anything up?

srdo:
@kishorvpatil Sorry to keep harping on this, but I still don't really understand how we're getting `InterruptedExceptions` that need to be logged at error level.

Could you give an example of a place in the Kafka code where an `InterruptedException` wraps another exception? I did a search of the Kafka repository for places where they throw `InterruptException`, and it looks like it's always just wrapping an `InterruptedException`? The `InterruptedException` constructor doesn't appear to allow you to wrap other exceptions either (i.e it has no `new InterruptedException(cause)` constructor)?

Do you know where the log line you posted is coming from? I initially thought it looked like a JUnit expected exception, but googling "but exception chain only contains" only pops up this PR and a log posted on this issue https://issues.apache.org/jira/browse/STORM-2873.

kishorvpatil:
@srdo, You are right. This was exception I noticed on old internal version of the code package. It is clearly not necessary in 2.x latest code. Let me close this patch.

PR #2639, opened 2018-04-19
STORM-3035: fix the issue in JmsSpout.ack when toCommit is empty

arunmahadevan:


external/storm-jms/src/main/java/org/apache/storm/jms/spout/JmsSpout.java
```diff
@@ -339,26 +339,26 @@ public void nextTuple() {
...
-        if (msgId.equals(oldest)) {
-            if (msg != null) {
-                try {
-                    LOG.debug("Committing...");
-                    msg.acknowledge();
-                    LOG.debug("JMS Message acked: " + msgId);
-                    this.toCommit.remove(msgId);
-                } catch (JMSException e) {
-                    LOG.warn("Error acknowldging JMS message: " + msgId, e);
+        if (!toCommit.isEmpty()) {
```
HeartSaVioR: Could we leave the log message (at least DEBUG) so that we can see which messages are ignored while acking due to previous failure?

HeartSaVioR: Might be better to leave the log message which messages are dropped in `fail()` too.

external/storm-jms/src/main/java/org/apache/storm/jms/spout/JmsSpout.java
```diff
@@ -339,26 +339,26 @@ public void nextTuple() {
...
+            JmsMessageID oldest = this.toCommit.first();
+            if (msgId.equals(oldest)) {
+                if (msg != null) {
+                    try {
+                        LOG.debug("Committing...");
+                        msg.acknowledge();
```
HeartSaVioR: I'm sorry I'm not familiar with JMS, but could you explain how this approach guarantee ack are done for all messages? Looks like it just removes the message from `toCommit` if late ack messages come earlier.

arunmahadevan: This piece of code was already there. I am guessing its based on the JMS acknowledgement mode.

See - https://docs.oracle.com/cd/E19798-01/821-1841/bncfw/index.html

In `Session.CLIENT_ACKNOWLEDGE` - Acknowledging a consumed message automatically acknowledges the receipt of all messages that have been consumed by its session, so this logic seems fine. 

The spout is ignoring Auto acknowledgement mode, but I am not sure about the other modes like `DUPS_OK_ACKNOWLEDGE` or `SESSION_TRANSACTED` work. cc @ptgoetz who might have more context around this.

arunmahadevan: I am not sure acking the oldest message in JMS is correct even for `CLIENT_ACKNOWLEDGE`. This would ack the new messages that have been consumed in the session (and possibly emitted) even before the spout received the ACK for the message. I guess we should keep removing the message from `toCommit` and invoke the JMS ack when its the last message in `toCommit`. (assuming we dont consume any other message in the meanwhile).

arunmahadevan:
This started out as a fix to handle the exceptions in "ack" when toCommit was empty. However during the review process and testing, figured out many more issues with the current approach. Also some JMS providers like Tibco supports ACK ing individual messages, which could not be handled with the existing code. The async mode of consuming the messages was also problematic to ensure at-least once delivery even with locks/synchronization since ack-ing an individual JMS message in CLIENT_ACK mode was going to ack the messages received in the listener (even if the listener did not return).

To handle all the issues, I have refactored quite and bit and changed the approach of consuming the messages from async (listener based) to sync (receive) and introduced MessageHandlers to handle the emit/ack/fail in different ways based on the mode.

@HeartSaVioR , can you review it again and let me know what you think?

external/storm-jms/pom.xml
```diff
@@ -94,7 +94,7 @@
                 <artifactId>maven-checkstyle-plugin</artifactId>
                 <!--Note - the version would be inherited-->
                 <configuration>
-                    <maxAllowedViolations>63</maxAllowedViolations>
+                    <maxAllowedViolations>73</maxAllowedViolations>
```
HeartSaVioR: We should not increase the number, decrease or no change is only allowed. Please fix checkstyle violations based on checkstyle report and keep the number.

external/storm-jms/src/main/java/org/apache/storm/jms/spout/JmsSpout.java
```diff
@@ -18,164 +18,124 @@
...
 import org.apache.storm.spout.SpoutOutputCollector;
 import org.apache.storm.task.TopologyContext;
 import org.apache.storm.topology.OutputFieldsDeclarer;
 import org.apache.storm.topology.base.BaseRichSpout;
 import org.apache.storm.tuple.Values;
-import org.apache.storm.utils.Utils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import javax.jms.Connection;
```
HeartSaVioR: Reorganizing imports happens now, and it looks like one of checkstyle violation.

external/storm-jms/src/main/java/org/apache/storm/jms/spout/JmsSpout.java
```diff
@@ -18,164 +18,124 @@
...
      *
      * @see Session
      */
     private int jmsAcknowledgeMode = Session.AUTO_ACKNOWLEDGE;
-    /**
-     * Indicates whether or not this spout should run as a singleton.
-     */
+
+    /** Sets up the way we want to handle the emit, ack and fails. */
+    private transient MessageHandler messageHandler = new MessageHandler();
```
HeartSaVioR: This must not be `transient`, because assigning `messageHandler` happens before serialization and Spout will lose the assigned value when deserialization happens. Are any of implementations of MessageHandler non-serializable?

external/storm-jms/src/main/java/org/apache/storm/jms/spout/JmsSpout.java
```diff
@@ -18,164 +18,124 @@
...
+            case Session.CLIENT_ACKNOWLEDGE:
+                messageHandler = new ClientAckHandler();
+                break;
+            case Session.SESSION_TRANSACTED:
+                messageHandler = new TransactedSessionMessageHandler();
+                break;
             default:
-                return "UNKNOWN";
-
+                LOG.warn("Unsupported Acknowledge mode: "
```
HeartSaVioR: So we are allowing non-JMS standard mode to parameter, right? I agree this is needed to support specific JMS implementation which **extends** session acknowledge mode.

external/storm-jms/src/main/java/org/apache/storm/jms/spout/JmsSpout.java
```diff
@@ -262,42 +189,26 @@ public void onMessage(Message msg) {
...
-            this.connection.start();
-            if (this.isDurableSubscription() && this.recoveryPeriodMs > 0) {
-                this.recoveryTimer = new Timer();
-                this.recoveryTimer.scheduleAtFixedRate(new RecoveryTask(), RECOVERY_DELAY_MS, this.recoveryPeriodMs);
-            }
-
+            ConnectionFactory cf = jmsProvider.connectionFactory();
+            Destination dest = jmsProvider.destination();
+            connection = cf.createConnection();
+            session = messageHandler.createSession(connection);
```
HeartSaVioR: We may want to consider the case: users provide mode which is not in JMS standard, and also setIndividualAck() is not called.

Now the case is handled as same as AUTO_ACKNOWLEDGE because of providing default value of messageHandler. Is it intended? We prevented the case with IllegalArgumentException.

external/storm-jms/src/main/java/org/apache/storm/jms/spout/JmsSpout.java
```diff
@@ -403,50 +274,31 @@ public void ack(Object msgId) {
...
-    /**
-     * Sets the periodicity of the timer task that checks for failures and recovers the JMS session.
+     * Sets the periodicity of the timer task that
+     * checks for failures and recovers the JMS session.
      *
      * @param period desired wait period
      */
-    public void setRecoveryPeriodMs(long period) {
-        this.recoveryPeriodMs = period;
+    public void setRecoveryPeriodMs(final long period) {
```
HeartSaVioR: This method can be removed.

external/storm-jms/pom.xml
arunmahadevan: Actually I can see more violations in JMSSpout.java in master branch (43) vs the patch (2). I am not sure why. Will fix it anyways.

external/storm-jms/src/main/java/org/apache/storm/jms/spout/JmsSpout.java
arunmahadevan: Good catch. Yes I should make the MessageHandler serializable.

external/storm-jms/src/main/java/org/apache/storm/jms/spout/JmsSpout.java
arunmahadevan: Yes. Its the only way to support provider extensions.

external/storm-jms/src/main/java/org/apache/storm/jms/spout/JmsSpout.java
arunmahadevan: We can throw an exception if `setIndividualAck` is invoked and the ACK mode is still the standard ones.

external/storm-jms/src/main/java/org/apache/storm/jms/spout/JmsSpout.java
arunmahadevan: I just left it since its breaking the public API and if someone is using this in their code.

HeartSaVioR: Could we have patch for 1.x-branch with deprecating the method? I'm OK to remove the method in 2.0.0, but you're right for 1.x.

external/storm-jms/src/main/java/org/apache/storm/jms/spout/JmsSpout.java
```diff
@@ -127,8 +130,14 @@ public void setJmsAcknowledgeMode(final int mode) {
                 messageHandler = new TransactedSessionMessageHandler();
                 break;
             default:
-                LOG.warn("Unsupported Acknowledge mode: "
-                    + mode + " (See javax.jms.Session for valid values)");
+                // individual message ack-ing needs vendor specific mode
+                if (individualAcks) {
```
HeartSaVioR: This change requires end users to call individualAcks earlier than this method. Could we guide this in javadoc or in exception message? Or can we fix it? (Maybe it defers verification step in open() so violating fail-fast...)

ptgoetz:
Looks good to me. +1

I was wondering why you removed the `distributed` flag, and realized it was never properly implemented! For it to work, we would need the following method override:

```java
    @Override
    public Map<String, Object> getComponentConfiguration() {
        if(!_isDistributed) {
            Map<String, Object> ret = new HashMap<String, Object>();
            ret.put(Config.TOPOLOGY_MAX_TASK_PARALLELISM, 1);
            return ret;
        } else {
            return null;
        }
    }    
```

I can see cases where that flag would be useful, for example when connecting to a topic vs. a queue. We may want to leave the flag there and fix the override.

arunmahadevan:
@ptgoetz , thanks for the suggestion.

Yes, it was clear why the distributed flag was needed when it was not used anywhere else in the code. I was hoping that the user will set the right parallelism while consuming from topics. I can add it back.

arunmahadevan:
@ptgoetz , can you take a look again?

ptgoetz:
Still +1. Thanks @arunmahadevan

PR #2640, opened 2018-04-20
[STORM-3036] Add isRemoteBlobExists RPC interface for deciding if rem…

danny0405:
As https://github.com/apache/storm/pull/2618 has described, now we try catch a KeyNotFoundException to decide remote blob does not exist.

It is confusing because user had removed it/them on their own initiative but we still got a KeyNotFoundException on both server and client side.

So i add a new api for this case.

This is the JIRA: [STORM-3036](https://issues.apache.org/jira/browse/STORM-3036)


danny0405:
@revans2 
Can you help me to reivew this patch?

danny0405:
@revans2 
I have fixed the conflicts and rebase, thx for your review work.

danny0405:
@revans2 
Hi, Bobby, kindly reminder.

PR #2642, opened 2018-04-23
[STORM-3043] Fix NullPointerException when apply() returns null

cleroux:
A null List<Object> is a valid return value from RecordTranslator.apply() if the ConsumerRecord is invalid.


srdo:
@cleroux Thanks for the fix. The change looks good to me. Please raise an issue at https://issues.apache.org/jira/secure/Dashboard.jspa, and include the issue number in the PR title and the commit message. It helps us track which branches the fix has been applied to.

srdo:
+1, thanks @cleroux, merged to master. Please open another PR against 1.x-branch, so this fix can make it into 1.x releases too :)

PR #2644, opened 2018-04-24
STORM-3041 improve worker-launcher messages

agresch:


storm-core/src/native/worker-launcher/impl/worker-launcher.c
```diff
@@ -309,7 +309,7 @@ int set_user(const char *user) {
 }
 
 /**
- * Open a file as the node manager and return a file descriptor for it.
+ * Open a file as the worker-launcher user and return a file descriptor for it.
```
srdo: Is this referring to "node manager" because it was copy pasted, or what is node manager referring to?

Maybe we should rename the functions that refer to "nm" as well?

agresch: Possibly this code was copied from Hadoop.

agresch:
@srdo - renamed nm-related methods and variables.

srdo:
@agresch Thanks. +1. Please squash to one commit, then we can merge.

If you want this on 1.x branches as well, I think we need a separate PR for that branch (the files have moved).

PR #2647, opened 2018-04-26
STORM-3040: Improve scheduler performance

revans2:
There are a lot of different scheduler improvements.  Mostly these are either caching, storing data in multiple ways so we can look it up quickly, and finally lazily sorting nodes in a rack only when it is needed, instead of all ahead of time.

I also added in performance tests.  They currently pass on travis, but I would like to hear from others on if this solution looks good or if there is a better way for us to do performance testing.

### DIFF
--- a/storm-server/src/main/java/org/apache/storm/scheduler/Cluster.java
+++ b/storm-server/src/main/java/org/apache/storm/scheduler/Cluster.java
@@ -47,6 +47,9 @@ import org.apache.storm.utils.Utils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+/**
+ * The current state of the storm cluster.  Cluster is not currently thread safe.
+ */
 public class Cluster implements ISchedulingState {
     private static final Logger LOG = LoggerFactory.getLogger(Cluster.class);
     private static final Function<String, Set<WorkerSlot>> MAKE_SET = (x) -> new HashSet<>();
--- a/storm-server/src/main/java/org/apache/storm/scheduler/Cluster.java
+++ b/storm-server/src/main/java/org/apache/storm/scheduler/Cluster.java
@@ -28,6 +28,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 import org.apache.storm.Config;
 import org.apache.storm.Constants;
@@ -48,6 +49,9 @@ import org.slf4j.LoggerFactory;
 
 public class Cluster implements ISchedulingState {
     private static final Logger LOG = LoggerFactory.getLogger(Cluster.class);
+    private static final Function<String, Set<WorkerSlot>> MAKE_SET = (x) -> new HashSet<>();
+    private static final Function<String, Map<WorkerSlot, NormalizedResourceRequest>> MAKE_MAP = (x) -> new HashMap<>();
+
     /**
      * key: supervisor id, value: supervisor details.
      */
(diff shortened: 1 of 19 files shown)


danny0405:
@revans2 
In total, this is a good promotion direction.

So this promotion mainly focus on repetitive computation and some eagerly computed data structure, right?

What is the average promotion percentage when applied this patch?

revans2:
@danny0405 In my tests `TestResourceAwareScheduler.testLargeTopologiesCommon` went from about 7 mins to about 7 seconds. For `TestResourceAwareScheduler.testLargeTopologiesOnLargeClustersGras` I don't have a before value because I killed it after an hour.  The after is about 7 seconds per topology, or about a min and a half.

revans2:
@danny0405 @kishorvpatil With some recent changes to master my patch started to fail with some checkstyle issues.  I have rebased and fixed all of the issues.  Please take a look again, specifically the second commit and let me know.

revans2:
Oh I forgot I also added back in something I messed up before and added back in anti-affinity to GRAS.

danny0405:
@revans2 
I approve with you promotion totally.

The only concern are all kinds of cache we use here, now storm has many caches not just for scheduling. I just think if we can make a disk-storage-backend for all of such caches master needs. Disk cache has better fault-tolerance and is much cheeper than memory, but this is another promotion direction, and has nothing to do with this patch.

BTW: thx for your nice work.

storm-server/src/main/java/org/apache/storm/scheduler/Cluster.java
```diff
@@ -48,6 +49,9 @@
 
 public class Cluster implements ISchedulingState {
     private static final Logger LOG = LoggerFactory.getLogger(Cluster.class);
+    private static final Function<String, Set<WorkerSlot>> MAKE_SET = (x) -> new HashSet<>();
+    private static final Function<String, Map<WorkerSlot, NormalizedResourceRequest>> MAKE_MAP = (x) -> new HashMap<>();
```
danny0405: HashSet is ok for now single daemon scheduling, we may make it thread safe when we want to support parallel scheduling, so can we add a comment about thread safe here?

revans2: I am happy to add in a comment to Cluster itself about it, as none of Cluster is currently thread safe.

As for parallel scheduling the plan that we had been thinking about was more around scheduling multiple topologies in parallel, rather then trying to make a single scheduler strategy multi-threaded, but both have advantages and disadvantages.

storm-server/src/main/java/org/apache/storm/scheduler/Cluster.java
```diff
@@ -763,6 +773,7 @@ public void setAssignments(
             assertValidTopologyForModification(assignment.getTopologyId());
         }
         assignments.clear();
+        totalResourcesPerNodeCache.clear();
```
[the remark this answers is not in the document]
revans2: I tried that, but it didn't have the performance boost I was hoping for.  The vast majority of the performance problem came from recomputing the value each time we wanted to sort, with for GRAS is once per executor.  So without the cache for a large topology we were recomputing things hundreds of thousands of times.  With the cache it is only how many nodes are in the cluster, which ends up being relatively small.  In reality the noise between runs drowned out any improvement, so I opted to not do the change.

revans2:
@danny0405 I added in the comments about thread safety like you suggested.

danny0405:
@revans2
Thx for your work, the storm-core travis check still fails, we should fix that.

revans2:
@danny0405 the failure is a known race condition around netty and is not related to this change.

PR #2648, opened 2018-04-27
STORM-3013: Keep KafkaConsumer open when storm-kafka-client spout is …

srdo:
…deactivated, in order to keep metrics working

See https://issues.apache.org/jira/browse/STORM-3013

The changes here are to fix a crash that occur if the spout is deactivated and Storm asks for metrics. The KafkaConsumer is closed when the spout deactivates, and the metrics provider tries to use the consumer to generate a few metrics.

The suggested fix is to not close the KafkaConsumer when deactivating the spout. There's no real reason to do it as far as I can tell, and having to replace the consumer makes the spout a bit more complex than it would otherwise be.


HeartSaVioR:
@srdo 
The patch wasn't applied to 1.x-branch cleanly. Could you please raise a PR for 1.x-branch as well? Thanks in advance!

PR #2652, opened 2018-05-01
STORM-3046: Ensure KafkaTridentSpoutEmitter handles empty batches cor…

srdo:
…rectly when they occur at the beginning of the stream

See https://issues.apache.org/jira/browse/STORM-3046


HeartSaVioR:
@srdo 
The patch wasn't applied to 1.x-branch cleanly. Could you please raise a PR for 1.x-branch as well? Thanks in advance!

PR #2653, opened 2018-05-01
[WIP] Detach storm-kafka-client release from Storm proper

srdo:
Opening this to get feedback on detaching the storm-kafka-client release cycle from Storm, as discussed on the mailing list. I'm putting this up so we can get an idea what we need to do to detach storm-kafka-client from Storm's release cycle, before we decide whether we want to try it.

I've made the following changes:
* Detached storm-kafka-client from Storm. It still uses the Storm parent POM, but is otherwise an independent project
* Copied the release_notes script so we can generate release notes for storm-kafka-client. Exclude storm-kafka-client issues from Storm release notes
* Addded a set of profiles to the storm-kafka-client POM so we can build against both Storm 1.x and 2.x. We can build against 1.x by doing e.g. `mvn clean install -Dstorm.release.line=1.x -Dstorm.version=1.2.2-SNAPSHOT`
* Reverted to raw maps in e.g. the spout open method. Having the generic types is nice, but prevents the spout from running on 1.x.

I ran into two API issues against Storm 1.x when trying to get this version running on both 1.x and 2.x. One is the map generics mentioned above, the other is a change I'd like to backport to 1.x here https://github.com/apache/storm/pull/2650. The issue is the type of the third parameter to IOpaquePartitionedTridentSpout.Emitter.getPartitionsForTask, which was changed in master.

Edit: Just to make clear, this version of the spout won't run on 1.x unless STORM-3047 goes in.

### DIFF
--- a/dev-tools/storm-kafka-client/release_notes.py
+++ b/dev-tools/storm-kafka-client/release_notes.py
@@ -19,7 +19,7 @@
 
 Depends on https://pypi.python.org/pypi/jira/, please use pip to install this module.
 
-Generates release notes for a Storm release by generating an HTML doc containing some introductory information about the
+Generates release notes for a Storm-kafka-client release by generating an HTML doc containing some introductory information about the
  release with links to the Storm docs followed by a list of issues resolved in the release. The script will fail if it finds
  any unresolved issues still marked with the target release. You should run this script after either resolving all issues or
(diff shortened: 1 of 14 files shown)


srdo:
Closing due to lack of interest.

PR #2655, opened 2018-05-02
fix STORM-3051

lujiefsi:

### Bug:

1.  callee CgroupCenter#getSubSystems return null when meet exception:
<pre>
} catch (Exception e) {
LOG.error("Get subSystems error {}", e);
}
return null;
</pre>
but its caller use it without check:
<pre>
public boolean isSubSystemEnabled(SubSystemType subSystemType) {
  Set<SubSystem> subSystems = this.getSubSystems();
  for (SubSystem subSystem : subSystems) {
     if (subSystem.getType() == subSystemType) {
     return true;
  }
  }
  return false;
}
</pre>
other callee and caller pair that have same problem.

2. callee RAS_Node#getUsedSlots and caller RAS_Node#totalSlotsUsed
3. CgroupCenter#getHierarchies and caller CgroupCenter#isMounted,

### DIFF
--- a/storm-client/src/jvm/org/apache/storm/container/cgroup/CgroupCenter.java
+++ b/storm-client/src/jvm/org/apache/storm/container/cgroup/CgroupCenter.java
@@ -97,9 +97,11 @@ public class CgroupCenter implements CgroupOperation {
     @Override
     public boolean isSubSystemEnabled(SubSystemType subSystemType) {
         Set<SubSystem> subSystems = this.getSubSystems();
-        for (SubSystem subSystem : subSystems) {
-            if (subSystem.getType() == subSystemType) {
-                return true;
+        if (subSystems != null){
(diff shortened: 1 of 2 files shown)


storm-client/src/jvm/org/apache/storm/container/cgroup/CgroupCenter.java
```diff
@@ -97,9 +97,11 @@ public synchronized static CgroupCenter getInstance() {
     @Override
     public boolean isSubSystemEnabled(SubSystemType subSystemType) {
         Set<SubSystem> subSystems = this.getSubSystems();
-        for (SubSystem subSystem : subSystems) {
-            if (subSystem.getType() == subSystemType) {
-                return true;
+        if (subSystems != null){
```
HeartSaVioR: nit: space between `)` and `{`

PR #2656, opened 2018-05-02
STORM-3049:a potential NPE in SupervisorSimpleACLAuthorizer#permit SimpleACLAuthorizer#permit

lujiefsi:

Bug:

callee ReqContext#principal have 12 callers, 10 of them have null checker like:
<pre>
public boolean permit(ReqContext context, String operation, Map<String, Object> topoConf) {
    return context.principal() != null ? users.contains(context.principal().getName()) : false;
}
</pre>
but SupervisorSimpleACLAuthorizer#permit  and SimpleACLAuthorizer#permit have no, just like:
<pre>
   //SupervisorSimpleACLAuthorizer#permit 
   String principal = context.principal().getName();
</pre>

### DIFF
--- a/storm-client/src/jvm/org/apache/storm/security/auth/authorizer/SimpleACLAuthorizer.java
+++ b/storm-client/src/jvm/org/apache/storm/security/auth/authorizer/SimpleACLAuthorizer.java
@@ -18,7 +18,6 @@ import java.util.Collection;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
-
 import org.apache.storm.Config;
 import org.apache.storm.security.auth.AuthUtils;
 import org.apache.storm.security.auth.IAuthorizer;


storm-client/src/jvm/org/apache/storm/security/auth/authorizer/SimpleACLAuthorizer.java
```diff
@@ -129,7 +129,7 @@ public void prepare(Map<String, Object> conf) {
      */
     @Override
     public boolean permit(ReqContext context, String operation, Map<String, Object> topoConf) {
-        String principal = context.principal().getName();
+    	String principal = context.principal() != null ? context.principal().getName() : null;
```
HeartSaVioR: I'm not sure when `context.principal()` is being null, but leaving as null could possibly incur another NPE. If it is a really bad case, better not to  permit with (maybe) log message.

@revans2 What do you think?

PR #2657, opened 2018-05-02
STORM-3048 : A Potential NPE

lujiefsi:

Bug:

Cluster#getAssignmentById has 20 callers, 18 callers have null checker like this:
<pre>
SchedulerAssignment assignment = cluster.getAssignmentById(td.getId());
if (assignment != null) {
    cpuNeeded -= getCpuUsed(assignment);
     memoryNeeded -= getMemoryUsed(assignment);
}
</pre>
the caller have no null checker :ConstraintSolverStrategy#checkSpreadSchedulingValid ConstraintSolverStrategy#checkConstraintsSatisfied.

I was feel uncertain about whether add null checker here, so i only add assert. If this bug are false positive,  please tell me.

### DIFF
--- a/storm-server/src/main/java/org/apache/storm/scheduler/resource/strategies/scheduling/ConstraintSolverStrategy.java
+++ b/storm-server/src/main/java/org/apache/storm/scheduler/resource/strategies/scheduling/ConstraintSolverStrategy.java
@@ -98,7 +98,7 @@ public class ConstraintSolverStrategy extends BaseResourceAwareStrategy {
      */
     private static boolean checkConstraintsSatisfied(Cluster cluster, TopologyDetails topo) {
         LOG.info("Checking constraints...");
-        assert(cluster.getAssignmentById(topo.getId())!=null);
+        assert (cluster.getAssignmentById(topo.getId()) != null);
         Map<ExecutorDetails, WorkerSlot> result = cluster.getAssignmentById(topo.getId()).getExecutorToSlot();
         Map<ExecutorDetails, String> execToComp = topo.getExecutorToComponent();
         //get topology constraints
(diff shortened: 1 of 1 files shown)


storm-server/src/main/java/org/apache/storm/scheduler/resource/strategies/scheduling/ConstraintSolverStrategy.java
```diff
@@ -98,6 +98,7 @@ public static boolean validateSolution(Cluster cluster, TopologyDetails td) {
      */
     private static boolean checkConstraintsSatisfied(Cluster cluster, TopologyDetails topo) {
         LOG.info("Checking constraints...");
+        assert(cluster.getAssignmentById(topo.getId())!=null);
```
Ethanlm: nit: space before and after `!=`.  same for the below code changes.

revans2:
I am getting a failure now of

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:2.17:check (validate) on project storm-server: You have 792 Checkstyle violations. The maximum number of allowed violations is 783. -> [Help 1]
```

Specifically

```
/Users/evans/src/apache-commit/storm-server/src/main/java/org/apache/storm/scheduler/resource/strategies/scheduling/ConstraintSolverStrategy.java
	101: WhitespaceAround: 'assert' is not followed by whitespace. Empty blocks may only be represented as {} when not part of a multi-block statement (4.1.3)
	101: WhitespaceAround: '!=' is not preceded with whitespace.
	101: WhitespaceAround: '!=' is not followed by whitespace. Empty blocks may only be represented as {} when not part of a multi-block statement (4.1.3)
	139: WhitespaceAround: 'assert' is not followed by whitespace. Empty blocks may only be represented as {} when not part of a multi-block statement (4.1.3)
	139: WhitespaceAround: '!=' is not preceded with whitespace.
	139: WhitespaceAround: '!=' is not followed by whitespace. Empty blocks may only be represented as {} when not part of a multi-block statement (4.1.3)
	178: WhitespaceAround: 'assert' is not followed by whitespace. Empty blocks may only be represented as {} when not part of a multi-block statement (4.1.3)
	178: WhitespaceAround: '!=' is not preceded with whitespace.
	178: WhitespaceAround: '!=' is not followed by whitespace. Empty blocks may only be represented as {} when not part of a multi-block statement (4.1.3)
```

PR #2659, opened 2018-05-03
STORM-3042 restore topology.acker.cpu.pcore.percent

agresch:
TOPOLOGY_ACKER_CPU_PCORE_PERCENT and other similar settings in adjustResourcesForExec() were orphaned accidentally by a previous pull request.  This adds the functionality back.

When we deployed the change to clusters that were near full with some topologies that had this setting, because we were not previously honoring it, the nodes now looked over 100% full and were throwing exceptions during scheduling.  

The fix for this was to instead zero out the resources and track that the issue is occurring.  We had also logged that the error was occurring, and found that the logging became excessive, which also slowed scheduling.  A further change was added by @revans2 to limit the logging.

### DIFF
--- a/storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
+++ b/storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
@@ -531,13 +531,13 @@ public class Nimbus implements Iface, Shutdownable, DaemonCommon {
     @SuppressWarnings("deprecation")
     private static <T extends AutoCloseable> TimeCacheMap<String, T> fileCacheMap(Map<String, Object> conf) {
         return new TimeCacheMap<>(ObjectReader.getInt(conf.get(DaemonConfig.NIMBUS_FILE_COPY_EXPIRATION_SECS), 600),
-                                  (id, stream) -> {
-                                      try {
-                                          stream.close();
-                                      } catch (Exception e) {
-                                          throw new RuntimeException(e);
-                                      }
-                                  });
+            (id, stream) -> {
+                try {
+                    stream.close();
+                } catch (Exception e) {
+                    throw new RuntimeException(e);
+                }
+            });
     }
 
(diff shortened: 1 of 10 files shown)


storm-server/src/main/java/org/apache/storm/scheduler/TopologyDetails.java
```diff
@@ -293,7 +295,7 @@ public Double getTotalMemReqTask(ExecutorDetails exec) {
      * Gets the total memory resource list for a set of tasks that is part of a topology.
      *
      * @param executors all executors for a topology
-     * @return Map<ExecutorDetails   ,       Double> , a map of the total memory requirement for all tasks in topology topoId.
+     * @return Map<ExecutorDetails>  a map of the total memory requirement for all tasks in topology topoId.
```
Ethanlm: Looks not right to me. `@return Set<SharedMemory>` ?

storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
```diff
@@ -533,11 +533,11 @@ private static StormBase make(TopologyStatus status) {
     private static <T extends AutoCloseable> TimeCacheMap<String, T> fileCacheMap(Map<String, Object> conf) {
         return new TimeCacheMap<>(ObjectReader.getInt(conf.get(DaemonConfig.NIMBUS_FILE_COPY_EXPIRATION_SECS), 600),
                                   (id, stream) -> {
-                                      try {
-                                          stream.close();
-                                      } catch (Exception e) {
-                                          throw new RuntimeException(e);
-                                      }
+                try {
```
Ethanlm: The format seems off

agresch:
@Ethanlm - updated the issues you mentioned.

HeartSaVioR:
@agresch Could you please rebase? After rebase we can review quickly and merge this in.

agresch:
@HeartSaVioR - rebased.

PR #2643, opened 2018-04-23
STORM-3039 handle slot ports in TIME_WAIT state

ghajos:
When worker is killed slot port remains in TIME_WAIT state. Since worker process is killed it is secure to start a new worker process on the same port.

PR #2660, opened 2018-05-03
STORM-3056: Add a test for quickly rebinding to a port

raghavgautam:
Adding test as per discussion on https://github.com/apache/storm/pull/2643#issuecomment-385003933

### DIFF
--- a/storm-core/test/jvm/org/apache/storm/messaging/netty/NettyTest.java
+++ b/storm-core/test/jvm/org/apache/storm/messaging/netty/NettyTest.java
@@ -17,6 +17,7 @@
 package org.apache.storm.messaging.netty;
 
 import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.Matchers.lessThan;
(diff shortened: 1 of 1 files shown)


raghavgautam:
@HeartSaVioR @ghajos Please review.

ghajos:
@raghavgautam Thanks for the quick reaction!

Unfortunately the newly introduced test passes without #2643, so it does not test the original issue. Since  in this modification both the server and the client is closed gracefully the port won't stuck in TIME_WAIT state, no need to use SO_REUSEADDR to restart new instances. To test #2643 the server port should stuck in TIME_WAIT state and then a new server instance should be started successfully.

I still think that it is out of scope to test an operating system level TCP setting in a Storm test.

cc: @HeartSaVioR

raghavgautam:
@ghajos Can you tell me how you tested your patch ?

raghavgautam:
@ghajos Ping.

kishorvpatil:
@raghavgautam Can you please address the merge conflicts?

PR #2661, opened 2018-05-05
[STORM-3055] 1.1.x remove conext connection cache

pczb:
worker has already cache connection use supervisor_id + port
and if context and worker both cache connection, there can be some problem describe in jira


revans2:
@pczb I have the same comments that I put on the master branch version of the pull request.

pczb:
port corresponding change #2669 to 1.1.x

PR #2662, opened 2018-05-06
STORM-3059: Fix NPE when the processing guarantee is not AT_LEAST_ONC…

srdo:
…E and the spout filters out a null tuple

https://issues.apache.org/jira/browse/STORM-3059


srdo:
@HeartSaVioR Whoops, added missing file.

PR #2664, opened 2018-05-07
STORM-2884: Remove storm-druid

revans2:
STORM-2884 is an issue with differences in the version of dependencies in druid tranquility conflicting with versions in storm. I personally think the right way to fix this issue is to drop tranquility.  The code has not seen a release since Jun 29, 2016.

https://github.com/druid-io/tranquility/releases

Honestly the project looks like it is close to dead.  Couple that with the fact that druid supports ingesting results from kafka.

http://druid.io/blog/2013/08/30/loading-data.html

and to me there really is not much reason to try and maintain this plugin.


arunmahadevan:
@revans2 , I don't think we should completely drop the plugin since there are users using it. There were a few options proposed https://github.com/apache/storm/pull/2498#issuecomment-355423380 and I guess we could pursue one (like shading curator within storm-druid) based on consensus. 

The version conflict between storm's and the connectors dependencies is going to be there in future as well so I guess this needs a more reasonable solution than dropping the connector. 

The druid page seem to recommend using the tranquility library due to limitations with realtime ingestion - http://druid.io/docs/latest/ingestion/stream-ingestion.html

srdo:
I think shading Curator in storm-druid seems like the best option, and maybe poking a bit at the Druid guys to get a release of Tranquility out. If it turns out that they are dropping Tranquility entirely, I agree that it makes sense not to support this connector anymore, but their documentation is still pointing to Tranquility as the library streaming producers should use (e.g. http://druid.io/docs/latest/tutorials/quickstart.html)

revans2:
@arunmahadevan and @srdo 

https://issues.apache.org/jira/browse/STORM-2882 is for adding back in shading to the storm client.  I am happy to take that up next.  My real concern is the long term support for tranquility.  It looks like there is consensus that dropping the druid bolt/state is not what is wanted, so I will close this pull request.

arunmahadevan:
@revans2 thanks for understanding the concerns and agree that we should do something about the long term support for tranquility. We can probably poke them like @srdo suggested to get out a new release with updated dependencies and see if we can make progress.

Irrespective of this, it makes sense to shade common libraries (like curator) into storm client to avoid conflicts with user topology code and thanks @revans2 for offering to take this up.

srdo:
I sent them an email here https://lists.apache.org/thread.html/4114eb25ff1e570927aa283c4127f8c83fb3bbee631ddebeee729f4c@%3Cdev.druid.apache.org%3E.

I think if it turns out that Tranquility is dead, we should consider removing storm-druid unless someone is willing to try to rewrite without Tranquility. It doesn't make sense to me to maintain a connector where the underlying client isn't being updated.

arunmahadevan:
Lets wait for a response. I think we can add some notes saying that storm-druid is deprecated and shade its dependencies than completely removing it since there are some users using it. This can be addressed as a part of STORM-2884.

srdo:
Got a response https://lists.apache.org/thread.html/22fa072edfdcb5a2fdb06e8c99daf8b2635da78e87c28e72b332ac0e@%3Cdev.druid.apache.org%3E. Sounds like Kafka is the recommended way to ingest into Druid, and there's not much development going into Tranquility, so I think deprecating storm-druid would be reasonable.

How about deprecating on 1.x and removing from 2.x (or just raise an issue to remove it from 3.x if we want to keep it around for a while longer)?

arunmahadevan:
I think if we can shade all the storm-druid dependencies we could probably leave it there.
At-least a few of the other storm connectors have not been updated in a while but we still keep it. Similarly we can just mark storm-druid deprecated (and recommend kafka based ingestion) in the README.md or somewhere. By 3.0 timeline, if there are no users we can even consider dropping it.

PR #2665, opened 2018-05-07
STORM-2988 Error on initialization of server mk-worker

dbist:
ran tests locally on storm-core and everything is clean.

### DIFF
--- a/storm-core/src/jvm/org/apache/storm/metrics2/reporters/JmxStormReporter.java
+++ b/storm-core/src/jvm/org/apache/storm/metrics2/reporters/JmxStormReporter.java
@@ -19,6 +19,7 @@ package org.apache.storm.metrics2.reporters;
 
 import com.codahale.metrics.JmxReporter;
 import com.codahale.metrics.MetricRegistry;
+import org.apache.storm.Config;
 import org.apache.storm.daemon.metrics.MetricsUtils;
 import org.apache.storm.metrics2.filters.StormMetricsFilter;
 import org.apache.storm.utils.Utils;
@@ -67,7 +68,7 @@ public class JmxStormReporter implements StormReporter {
     }
 
     public static String getMetricsJMXDomain(Map reporterConf) {
-        return Utils.getString(reporterConf, JMX_DOMAIN);
+        return Utils.getString(reporterConf.get(Config.STORM_DAEMON_METRICS_REPORTER_PLUGIN_DOMAIN), null);
     }
 
     @Override
--- a/storm-core/src/jvm/org/apache/storm/metrics2/reporters/JmxStormReporter.java
+++ b/storm-core/src/jvm/org/apache/storm/metrics2/reporters/JmxStormReporter.java
@@ -19,7 +19,6 @@ package org.apache.storm.metrics2.reporters;
 
 import com.codahale.metrics.JmxReporter;
 import com.codahale.metrics.MetricRegistry;
-import org.apache.storm.Config;
 import org.apache.storm.daemon.metrics.MetricsUtils;
 import org.apache.storm.metrics2.filters.StormMetricsFilter;
 import org.apache.storm.utils.Utils;
@@ -68,7 +67,7 @@ public class JmxStormReporter implements StormReporter {
     }
 
     public static String getMetricsJMXDomain(Map reporterConf) {
-        return Utils.getString(reporterConf.get(Config.STORM_DAEMON_METRICS_REPORTER_PLUGIN_DOMAIN), null);
+        return Utils.getString(reporterConf.get(JMX_DOMAIN), null);
     }
 
     @Override


storm-core/src/jvm/org/apache/storm/metrics2/reporters/JmxStormReporter.java
```diff
@@ -67,7 +68,7 @@ public void prepare(MetricRegistry metricsRegistry, Map<String, Object> stormCon
     }
 
     public static String getMetricsJMXDomain(Map reporterConf) {
-        return Utils.getString(reporterConf, JMX_DOMAIN);
+        return Utils.getString(reporterConf.get(Config.STORM_DAEMON_METRICS_REPORTER_PLUGIN_DOMAIN), null);
```
priyank5485: JMX_DOMAIN used initially is different than Config.STORM_DAEMON_METRICS_REPORTER_PLUGIN_DOMAIN

I think we should stick to JMX_DOMAIN config key unless @ptgoetz  thinks this is good.

arunmahadevan: I think user should be able to specify a domain so it cannot be a fixed value like JMX_DOMAIN. So the proposed changes looks fine.

arunmahadevan: I think JMX_DOMAIN was intended to be the default value of the domain.

So may be `Utils.getString(reporterConf.get(Config.STORM_DAEMON_METRICS_REPORTER_PLUGIN_DOMAIN), JMX_DOMAIN)`

cc @ptgoetz

priyank5485: @arunmahadevan I am not saying we should return that as default value. I am suggesting to use that as the key to look up the conf map.

dbist: happy to oblige, would like to draw a consensus as in branch 2.x I can't just import Config class, be it due to moving of the class from store-core to storm-client or something else? It's explicitly looking for the key in the Config class?

ptgoetz: @priyank5485 is correct. JMX_DOMAIN is the key for the lookup from the r Porter config.

ptgoetz: ‘return Utils.getString(reporterConf, JMX_DOMAIN);’

Should be something like 

‘return Utils.getString(reporterConf.get(JMX_DOMAIN), “localhost”);’

(Apologies, not on a computer atm)

HeartSaVioR: `Utils.getString(reporterConf.get(JMX_DOMAIN), null)` would keep the intention and avoid NPE.

dbist:
thank you @priyank5485 @arunmahadevan @ptgoetz and @HeartSaVioR for the lively review, integrated changes, please review

arunmahadevan:
@ptgoetz , any reason to use a different constant while this [config](https://github.com/apache/storm/blob/1.x-branch/storm-core/src/jvm/org/apache/storm/Config.java#L179) exists? And JmxStormReporter seem to use the constants from Config [here](https://github.com/apache/storm/blob/1.x-branch/storm-core/src/jvm/org/apache/storm/metrics2/reporters/JmxStormReporter.java#L41) and [here](https://github.com/apache/storm/blob/1.x-branch/storm-core/src/jvm/org/apache/storm/metrics2/reporters/JmxStormReporter.java#L46) ? 

Is there a documentation to let the users know the right key to use?

ptgoetz:
> @ptgoetz , any reason to use a different constant while this config exists? And JmxStormReporter seem to use the constants from Config here and here ?

@arunmahadevan, to keep them logically separated so either could be potentially deprecated down the road.

HeartSaVioR:
@arunmahadevan @ptgoetz 
Let's "reuse all" or "duplicate all", not reuse some of them partially. And let's leave explanation on JmxStormReporter, since there's no explanation on JmxStormReporter from metrics v2 doc as well as storm.yaml.example.

https://github.com/apache/storm/blob/master/docs/metrics_v2.md
https://github.com/apache/storm/blob/master/conf/storm.yaml.example

arunmahadevan:
We should add docs and an example usage in the storm.yaml.example (which has graphite and console). @dbist will be great if you address this for JMXStormReporter as a part of this PR.

ptgoetz:
@HeartSaVioR Agreed, but those issues can be addressed outside the scope of this bug (i.e. separate JIRA). For now we should focus on getting this fix in since it is currently holding up the release.

dbist:
@ptgoetz you've read my mind, this fix is holding up a release and I can work on the separate jira to address the doc, not a problem. We can target that for next release.

dbist:
doc update jira is https://issues.apache.org/jira/browse/STORM-3062
please review this PR and 2666 for branch 2 fix.

ptgoetz:
+1

HeartSaVioR:
Go ahead. The issue is effectively blocker so I don't want to block this for non-critical thing. Moreover the configurations are already exposed (though it is not documented yet) so someone may have concerned with changing them. 

Ideally we may get a chance to fix it for 2.0, but don't have strong voice to do it.

HeartSaVioR:
The build failure is unrelated to this patch. Build from local clearly succeeds.

HeartSaVioR:
Thanks @dbist I squashed and merged the patch into 1.x-branch via c26e4ad.

Unfortunately, github auto close only works with master branch. Could you please close this issue? Thanks in advance!

PR #2667, opened 2018-05-08
STORM-3063: Fix minor pom issues

srdo:
https://issues.apache.org/jira/browse/STORM-3063

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -231,11 +231,6 @@
 
     </developers>
 
-
-    <prerequisites>
-        <maven>3.0.0</maven>
-    </prerequisites>
-
     <scm>
         <connection>scm:git:https://git-wip-us.apache.org/repos/asf/storm.git</connection>
         <developerConnection>scm:git:https://git-wip-us.apache.org/repos/asf/storm.git</developerConnection>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -1275,6 +1270,25 @@
                     <useStandardDocletOptions>true</useStandardDocletOptions>
                 </configuration>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-enforcer-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>enforce-maven</id>
```
hmcl: since there is only one execution, is 'id' really necessary ?

hmcl:
+1

pom.xml
srdo: No, it could be left out. However I think since this plugin is so generic, it's helpful to have a descriptive name for the execution.

hmcl: Agree. Would 'enforce-maven-version' be more descriptive?

srdo: Yes, that's better. Will change.

PR #2436, opened 2017-11-27
STORM-2833: Use the original host name when closing (2.x)

revans2:

PR #2669, opened 2018-05-10
[STORM-3055] remove conext connection cache

pczb:
workerState has already cache connection use supervisor_id + port
and if context and worker both cache connection, there can be some problem describe in jira
jira link: https://issues.apache.org/jira/browse/STORM-3055

### DIFF
--- a/storm-client/src/jvm/org/apache/storm/messaging/netty/Client.java
+++ b/storm-client/src/jvm/org/apache/storm/messaging/netty/Client.java
@@ -78,10 +78,7 @@ public class Client extends ConnectionWithStatus implements IStatefulObject, ISa
     private final StormBoundedExponentialBackoffRetry retryPolicy;
     private final ClientBootstrap bootstrap;
     private final InetSocketAddress dstAddress;
-    //The actual name of the host we are trying to connect to so that
-    // when we remove ourselves from the connection cache there is no concern that
-    // the resolved host name is different.
-    private final String dstHost;
+
     /**
      * The channel used for all write operations from this client to the remote destination.
      */
@@ -114,7 +111,6 @@ public class Client extends ConnectionWithStatus implements IStatefulObject, ISa
      * Whether the SASL channel is ready.
      */
     private final AtomicBoolean saslChannelReady = new AtomicBoolean(false);
-    private final Context context;
     private final HashedWheelTimer scheduler;
     private final MessageBuffer batcher;
     // wait strategy when the netty channel is not writable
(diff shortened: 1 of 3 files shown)


pczb:
@revans2 could you help review this, i see your pull request #2436 is related to connection cache . and the travis fail is not related to this pr.

revans2:
The original code added this so when shutting down the context we could be sure that all ongoing connections were also terminated as cleanly as possible.  After this change that is only true for the server side and not the client side.

What are the ramifications of that?  I think it was just defensive programming so it is probably fine to make this change from that perspective, but I would to make sure that my understanding of the problem matches yours.

From the comments in STORM-3055 the error appears to be triggered when a supervisor is shut down, wiped clean (so it gets a new supervisor id), and then brought back up.  At that point nimbus schedules the worker on the same node/slot as before, but with a new supervisor ID.  This confuses the connection caching because when updating the connections it gets a list of connections to shut down and a separate list of connections to create.  The new connections are created, but in this case a new one is not created because we already have it open.  Then the old unneeded connections are torn down, but in this case the connection is needed.

Looking at the javadocs for IContext. It looks like the caching really does violate the spec, but it is a bit of a gray area.

https://github.com/apache/storm/blob/53f38bc31f2fd315a520ba6b86c0a60be08381cc/storm-client/src/jvm/org/apache/storm/messaging/IContext.java#L48-L57

I am fine with removing the caching like this JIRA does, but I do want to see the code cleaned up, because without the caching there is a lot of extra code that can be removed.

storm-client/src/jvm/org/apache/storm/messaging/netty/Client.java
```diff
@@ -451,7 +451,6 @@ public int getPort() {
     public void close() {
         if (!closing) {
             LOG.info("closing Netty Client {}", dstAddressPrefixedName);
-            context.removeClient(dstHost, dstAddress.getPort());
```
revans2: This line is the only reason to pass context into the Client.  Please remove all references to context if we are going to remove this.

storm-client/src/jvm/org/apache/storm/messaging/netty/Context.java
```diff
@@ -65,20 +65,8 @@ public synchronized IConnection bind(String storm_id, int port) {
      * establish a connection to a remote server
      */
     public synchronized IConnection connect(String storm_id, String host, int port, AtomicBoolean[] remoteBpStatus) {
```
revans2: This no longer needs to be synchronized.  That was used to protect connections, and is no longer needed if all we do is create a new Client.

storm-client/src/jvm/org/apache/storm/messaging/netty/Context.java
```diff
@@ -65,20 +65,8 @@ public synchronized IConnection bind(String storm_id, int port) {
...
-        IConnection connection = connections.get(key(host, port));
```
revans2: Please rename connections to `serverConnections` as it is only for Servers now.  Also it would be nice to change it back to a list or a set instead of a map, because we will no longer be using it as a cache, so we don't need to pull out a specific value any longer.  We just want it so we can close them on termination.

srdo:
As far as I can tell all connections are still being shut down with this change. The connections cached by WorkerState are closed during worker shutdown in https://github.com/apache/storm/blob/14b0b4fc5e0945456769fd58a3595188e3dea234/storm-client/src/jvm/org/apache/storm/daemon/worker/Worker.java#L449, and the context is shut down a few lines further down. 

The WorkerState.refreshConnections method also makes sure to never create a connection for a NodeInfo that is already present, so I don't think we're leaking connections there that would need to be picked up by context.term. 

Would like to see the cleanup @revans2 mentioned, as well as a change in the IContext docs so it's specified that the created connections are new.

storm-client/src/jvm/org/apache/storm/messaging/netty/Client.java
pczb: done

pczb:
@srdo add doc specification, i was wonder if we need remove synchronized when bind port
@revans2  just as srdo says the connection will be closed by the worker before worker shutdown.
i will squash all commit after review

srdo:
LGTM, thanks for addressing comments

revans2:
Once you make the corresponding changes on #2661 I would be happy to merge them both in.

revans2:
Actually we just hit this in production so thanks again @pczb for finding and fixing this.

PR #2670, opened 2018-05-10
STORM-3068: STORM_JAR_JVM_OPTS are not passed to storm-kafka-monitor …

rajkrrsingh:
added STORM_JAR_JVM_OPTS to java command in storm-kafka-monitor file to pass java configuration

### DIFF
--- a/bin/storm-kafka-monitor
+++ b/bin/storm-kafka-monitor
@@ -49,4 +49,4 @@ if [ -z "$JAVA_HOME" ]; then
 else
   JAVA="$JAVA_HOME/bin/java"
 fi
-exec $JAVA $STORM_JAAS_CONF_PARAM -cp $STORM_BASE_DIR/lib-tools/storm-kafka-monitor/storm-kafka-monitor-*.jar org.apache.storm.kafka.monitor.KafkaOffsetLagUtil "$@"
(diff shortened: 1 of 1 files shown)


priyank5485:
@rajkrrsingh Is the presumption that STORM_JAR_JVM_OPTS will be set? Have you checked if its okay when the variable is not set? Can you add it to README as well regarding this change?

rajkrrsingh:
@priyank5485 I actually verified that STORM_JAR_JVM_OPTS is not propagating to the java args, the test I ran to confirm is 

without the change java command look like
```
+ exec /usr/jdk64/jdk1.8.0_112/bin/java -cp /usr/hdp/2.6.2.0-205/storm/toollib/storm-kafka-monitor-1.1.0.2.6.2.0-205.jar org.apache.storm.kafka.monitor.KafkaOffsetLagUtil
```

while with the change and setting STORM_JAR_JVM_OPTS=-Xmx1024m the command look like
```
+ exec /usr/jdk64/jdk1.8.0_112/bin/java -Xmx1024m -cp /usr/hdp/2.6.2.0-205/storm/toollib/storm-kafka-monitor-1.1.0.2.6.2.0-205.jar org.apache.storm.kafka.monitor.KafkaOffsetLagUtil
```

priyank5485:
@rajkrrsingh When you say STORM_JAR_JVM_OPTS is not propagating to the java args, where is it not propagating from? I don't see it being set anywhere in this script. And i see that its sourcing the storm-env.sh script. So i was just suggesting just adding a check like we have for JAVA_HOME to handle case where its not defined.

priyank5485:
+1(NB)

PR #2675, opened 2018-05-14
STORM-3061: Upgrade lots of dependencies

revans2:
This upgrades lots of dependencies, including thrift.  The thrift changes are based off of an earlier patch done by @arunmahadevan 

Please skip over the generated files and look mostly at the other files.

I have only tested that the unit tests pass and that a single node cluster comes up and runs, with similar performance.

I did make a few minor changes in some places to be able to remove dependencies that are not used/needed any more (hbase-server from storm-hbase just to get some string functions).


revans2:
The travis failure is an interesting one, sql-core is failing, but none of the tests are failing, the junit JVM exist badly as a part of shutdown.  I have not been able to reproduce it locally, but I'll keep trying.

revans2:
The failure looks like it might be related to #2674

HeartSaVioR:
Could we break this PR down to non-external and externals (per module)? This is too hard to follow and do manual tests. Passing UTs doesn't represent for some external modules to guarantee working correctly.

revans2:
I'll try to break it down to more manageable chunks that are split up by component as much as possible.  There are a lot of the dependencies that cross boundaries though I will call those out in each individual pull request.

@srdo I am going to close this pull request so I can break it up.  I will look at/fix what you have called out.

PR #2676, opened 2018-05-15
STORM-3073: Uncap pendingEmits for bolt executors, and prevent LoadSpout from overflowing pendingEmits in spout executors

srdo:
https://issues.apache.org/jira/browse/STORM-3073

The first commit contains the changes I made to ExclamationTopology to provoke the error. I'll remove it after review, it is just included so it's easier to understand the error.

There are two changes in this PR. The first is to uncap the pendingEmits queue for bolt executors. It's currently capped at 1024 elements, which makes it dangerous for bolts to emit more than 1024 tuples in an execute invocation. If the bolt executor is experiencing backpressure and tries to add the tuples to pendingEmits, the queue size will be exceeded and the worker will crash.

The second change is to make LoadSpout emit failed tuples from nextTuple instead of doing it from fail. Since the spout executor is also limited to 1024 tuples in the pending queue, it is likely that the spout executor will exceed the queue limit and crash if a bunch of tuples fail at the same time (e.g. due to timeout) while the spout is adding tuples to pendingEmits. Since the spout won't call nextTuple if there are tuples in pendingEmits, we can just move the retries to that method to prevent the spout from exceeding the pendingEmits limit.

### DIFF
--- a/examples/storm-loadgen/src/main/java/org/apache/storm/loadgen/LoadSpout.java
+++ b/examples/storm-loadgen/src/main/java/org/apache/storm/loadgen/LoadSpout.java
@@ -18,11 +18,13 @@
 
 package org.apache.storm.loadgen;
 
+import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Queue;
 import java.util.stream.Collectors;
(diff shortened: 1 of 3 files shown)


srdo:
@HeartSaVioR You're right, uncapped the queue in both types of executor, and removed the test code.

srdo:
Thanks for the quick review

PR #2677, opened 2018-05-15
STORM-3075 fix NPE

agresch:


Ethanlm:
It looks reasonable. But would like to understand more about it. Could you elaborate on in which case tc will be null?

agresch:
@Ethanlm - I see tc being null whenever I start nimbus (at least with no topologies).

agresch:
@Ethanlm - nice catch.  Will fix the source of the problem instead....

agresch:
@Ethanlm - please take a look.  Thanks.

agresch:
@Ethanlm - done.  Added a spelling fix.

PR #2678, opened 2018-05-16
STORM-3076:Fix the bug of examples in external can not be compiled

fuhuaiyong:
In the  external package,the modules include storm-hbase,storm-hdfs and storm-kafka , can not be compiled successfully, due to the lack of specific implements of ILocalCluster in the storm-core-0.9.5


fuhuaiyong:
what is wrong

PR #2682, opened 2018-05-16
STORM-3076:Fix the bug which results in the failure of compiling

fuhuaiyong:
In the  external package,the modules include storm-hbase,storm-hdfs and storm-kafka , can not be compiled successfully, due to the lack of specific implements of ILocalCluster in the storm-core-0.9.5

### DIFF
--- a/external/storm-hdfs/pom.xml
+++ b/external/storm-hdfs/pom.xml
@@ -21,7 +21,7 @@
     <parent>
         <artifactId>storm</artifactId>
         <groupId>org.apache.storm</groupId>
-        <version>0.9.5-SNAPSHOT</version>
+        <version>0.9.3</version>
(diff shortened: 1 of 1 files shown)


fuhuaiyong:
it is a simple modification. why checks failed ?

srdo:
The travis test isn't stable outside the current master, I wouldn't worry about it on this branch.

The branch you're targeting isn't the current 0.9.x branch, the current branch is https://github.com/apache/storm/blob/0.9.x-branch

LocalCluster seems to exist in 0.9.x https://github.com/apache/storm/blob/0.9.x-branch/storm-core/src/clj/backtype/storm/LocalCluster.clj

I think it's unlikely there'll be any more 0.9.x releases, consider upgrading to a more recent release (ideally 1.2.x)

PR #2684, opened 2018-05-17
STORM-3079 add getMessage() support for thrift exceptions

agresch:


storm-client/src/jvm/org/apache/storm/utils/WrappedAuthorizationException.java
```diff
@@ -0,0 +1,17 @@
+package org.apache.storm.utils;
```
revans2: Missing the license header.

agresch: Fixed

PR #847, opened 2015-11-03
STORM-162: Load Aware Shuffle Grouping

revans2:
I am recreating this pull request. Originally there was discussion about using the velocity of a bolt in addition to the queue size in determining the routing, but I didn't have time to implement that, and we have been running with this code in production for quite a while, so I decided to just put it up, as it has a really positive impact.  This is most pronounced on heterogeneous clusters where not all nodes have the same CPU.  But also this improves the performance of the groupings implementations significantly dropping the 99th and 99.9th %-lie latencies significantly for the Latency vs Throughput test case.

I would still like to explore using velocity and network distance in the routing calculations, but in a follow on JIRA.

PR #2685, opened 2018-05-20
STORM-1294: Port netty-unit-test to Java

srdo:
https://issues.apache.org/jira/browse/STORM-1294

The batch-test was not being run, and wouldn't pass with the given configuration. It was disabled in this PR https://github.com/apache/storm/pull/847/files#diff-d183c4c873554507040b705bf17624b8. Since no one mentioned it, it looks like it might just have been a mistake. Updating the configuration to match the one used in the other tests makes the test pass, so I think we might as well include it.


storm-core/test/jvm/org/apache/storm/messaging/netty/NettyTest.java
```diff
@@ -0,0 +1,382 @@
...
+                assertThat(responseMessage.message(), is(messageBytes));
+            }
+        } finally {
+            context.term();
+        }
+    }
+
+    private Map<String, Object> largeMessageConf() {
+        Map<String, Object> conf = basicConf();
+        conf.put(Config.STORM_MESSAGING_NETTY_BUFFER_SIZE, 1024000);
```
HeartSaVioR: Looks like this was 102400, as `test-large-msg` defines.

storm-core/test/jvm/org/apache/storm/messaging/netty/NettyTest.java
```diff
@@ -0,0 +1,382 @@
...
+                    });
+            }
+        } finally {
+            context.term();
+        }
+    }
+
+    @Test
+    public void testBatch() throws Exception {
+        doTestBatch(basicConf());
```
HeartSaVioR: This conf should have `STORM-MESSAGING-NETTY-BUFFER-SIZE 1024000` as `test-batch` defines, so you may want to define another method for new configuration.

storm-core/test/jvm/org/apache/storm/messaging/netty/NettyTest.java
srdo: Nice catch

PR #2687, opened 2018-05-22
STORM-3061: thrift 0.11

revans2:
This moves to thrift 0.11

The only files changed are pom.xml and genthrift.sh.  The other files are all generated.


storm-client/src/genthrift.sh
```diff
@@ -17,7 +17,7 @@
 
 rm -rf gen-javabean gen-py py
 rm -rf jvm/org/apache/storm/generated
-thrift --gen java:beans,hashcode,nocamel,generated_annotations=undated --gen py:utf8strings storm.thrift
```
Ethanlm: Could you explain the reason to remove `hashcode` and  the impact of it? I am not familiar with it. Thanks

revans2: Not sure why we had hashcode to begin with. From the 0.9.3 compiler --help it shows
```
  csharp (C#):
...
    hashcode:        Generate a hashcode and equals implementation for classes.
```

So it is generating hash codes for the c# language.  I believe that it was also being used for python, but that is not documented in thrift.  In 0.11.0 it is still there in the help command for c#, but now if I run with out the change I get the error.

```
[FAILURE:generation:1] Error: unknown option java:hashcode
```

srdo: There used to be a hashcode option for Java, but it was removed https://github.com/apache/thrift/pull/65. As of Thrift 0.10, it now complains if you try to use thrift with invalid arguments https://issues.apache.org/jira/browse/THRIFT-3605

srdo: I think the utf8strings option may be unnecessary https://issues.apache.org/jira/browse/THRIFT-3503

revans2: @srdo do you want me to remove the utf8strings option?

srdo: No, it's fine

srdo:
+1

arunmahadevan:
+1 LGTM.

revans2:
The integration test failures look unrelated to this change.

revans2:
Yes I reran the integration tests manually and they all passed.

PR #2688, opened 2018-05-23
STORM-3061: Remove unused core dependencies

revans2:
This is for disruptor and java.jmx

### DIFF
--- a/bin/storm.py
+++ b/bin/storm.py
@@ -705,7 +705,6 @@ def nimbus(klass="org.apache.storm.daemon.nimbus.Nimbus"):
     cppaths = [CLUSTER_CONF_DIR]
     jvmopts = parse_args(confvalue("nimbus.childopts", cppaths)) + [
         "-Dlogfile.name=nimbus.log",
-        "-DLog4jContextSelector=org.apache.logging.log4j.core.async.AsyncLoggerContextSelector",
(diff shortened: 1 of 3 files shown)


bin/storm.py
```diff
@@ -705,7 +705,6 @@ def nimbus(klass="org.apache.storm.daemon.nimbus.Nimbus"):
     cppaths = [CLUSTER_CONF_DIR]
     jvmopts = parse_args(confvalue("nimbus.childopts", cppaths)) + [
         "-Dlogfile.name=nimbus.log",
-        "-DLog4jContextSelector=org.apache.logging.log4j.core.async.AsyncLoggerContextSelector",
```
HeartSaVioR: Just curious, we don't touch log4j2 dependency, so can I treat it as removing unnecessary option?

revans2: So a long time ago we tried out async logging for log4j.  It kind of helped, but also caused some issues.  It puts the log messages into a disruptor queue and has a background thread handles writing them out.  In my big patch after upgrading the version of log4j and removing disruptor from the classpath async logging started to get errors about disruptor not being on the classpath.  I don't know which caused it, or if it was a combination of things, but I thought if I was removing disruptor from the classpath this would be the right place to remove the async logging too.

HeartSaVioR: Thanks for the kind explanation. Makes totally sense.

PR #2690, opened 2018-05-23
STORM-3061: Clean up some storm-druid dependencies

revans2:
`mvn dependency:tree` showed that the only real dependency changes were for scala and jackson that went from 2.4.6 to 2.9.4.

I am not totally sure how to test this currently as there is an open issue right now that it does not work. 

If others want to try this feel welcome to, if not we can try to fix any remaining issues as a part of https://issues.apache.org/jira/browse/STORM-2884


revans2:
Because of STORM-3097 #2707 I am closing this pull request.  I might reopen it if STORM-3097 goes a different direction.

PR #2693, opened 2018-05-24
STORM-3084 fix Nimbus NPE

agresch:


agresch:
Can we get this merged?  Thanks.

PR #2696, opened 2018-05-28
STORM-3086: Add documentation for invoking static factory methods fro…

srdo:
…m Flux

https://issues.apache.org/jira/browse/STORM-3086


srdo:
The examples are slightly adjusted versions of some code that appears in the flux-core tests https://github.com/apache/storm/blob/master/flux/flux-core/src/test/resources/configs/config-methods-test.yaml#L40

PR #2699, opened 2018-06-01
STORM-3087: Make FluxBuilder.canInvokeWithArgs check whether the actu…

srdo:
…al argument type is assignable to Number before deciding that a method with a primitive parameter can be invoked

1.x version of https://github.com/apache/storm/pull/2695


danny0405:
Looks good, thx for your contribution.

PR #2700, opened 2018-06-02
[STORM-3093] Cache the storm id to executors mapping on master to avo…

danny0405:
# What this patch for
Now nimbus will collect all the topologies's conf/topology-ser/storm-base to compute in a scheduling round, which is a very heavy work. The scheduling will still take to minutes even we now change to RPC heartbeats and assignment distribution.

So i decide to redesign the scheduler, so we can only schedule the topologies that need to: that have dead workers or not enough number workers.

Here i checkout out the code and found that the id->executors mapping is computed every time for every topology, which is really a heavy computation and totally not that necessary, because this mapping is fixed invariable for a topology unless we rebalance or kill it.

So i refactor the code a little here, and this is more powerful after the scheduler is resigned for delta-scheduling[ which is very lightweight even there are thousands of topologies on one cluster.]

For now this is enough for us.

This is the JIRA: [STORM-3093](https://issues.apache.org/jira/browse/STORM-3093).

### DIFF
--- a/storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
+++ b/storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
@@ -284,8 +284,8 @@ public class Nimbus implements Iface, Shutdownable, DaemonCommon {
         Assignment oldAssignment = state.assignmentInfo(topoId, null);
         state.removeStorm(topoId);
         notifySupervisorsAsKilled(state, oldAssignment, nimbus.getAssignmentsDistributer());
-        BlobStore store = nimbus.getBlobStore();
         nimbus.getHeartbeatsCache().getAndUpdate(new Dissoc<>(topoId));
+        nimbus.getIdToExecutors().getAndUpdate(new Dissoc<>(topoId));
         return null;
     };
     private static final TopologyStateTransition DO_REBALANCE_TRANSITION = (args, nimbus, topoId, base) -> {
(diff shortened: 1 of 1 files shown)


Ethanlm:
Looks good. Do you have any performance test results to share so that we can have an idea how much performance gain this patch gets? It's fine if you don't have it

danny0405:
@Ethanlm 
Sorry i have not did the benchmark test, but we can see the promotion through the storm-core module building time, it's about 10 seconds faster on my laptop.

I can add in the test cases if if it needed.

revans2:
The travis failure looks unrelated.

storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
```diff
@@ -1333,6 +1336,21 @@ private AssignmentDistributionService getAssignmentsDistributer() {
...
+        return idToExecutors;
+    }
+
+    private Set<List<Integer>> getOrUpdateExecutors(String topoId, StormBase base, Map<String, Object> topoConf,
+                                                    StormTopology topology)
+        throws IOException, AuthorizationException, InvalidTopologyException, KeyNotFoundException {
+        Set<List<Integer>> executors = idToExecutors.get().get(topoId);
+        if (null == executors) {
+            executors = new HashSet<>(computeExecutors(topoId, base, topoConf, topology));
+            idToExecutors.getAndUpdate(new Assoc<>(topoId, executors));
```
kishorvpatil: Do we intend to create new HashMap every time? If not, would `idToExecutors.get().put(topoId, executors);` should suffice?

PR #2703, opened 2018-06-05
[STORM-3094] : Added topology name validation at client side

ManoharVanam:
Current Behavior : Execute topology with invalid topology name is throwing exception after uploading the jar.

Improvement : Validating topology name at client side before uploading the jar.

### DIFF
--- a/storm-client/src/jvm/org/apache/storm/StormSubmitter.java
+++ b/storm-client/src/jvm/org/apache/storm/StormSubmitter.java
@@ -231,6 +231,9 @@ public class StormSubmitter {
                     if (topologyNameExists(name, client)) {
                         throw new RuntimeException("Topology with name `" + name + "` already exists on cluster");
                     }
+                    if (!Utils.isValidKey(name)) {
+                        throw new IllegalArgumentException(name + " does not appear to be a valid topology name.");
+                    }
 
                     // Dependency uploading only makes sense for distributed mode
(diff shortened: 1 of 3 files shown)


danny0405:
Sorry i don't think put the validation at Client side is a good idea, cause many kind of clients may interact with Master through thrift RPC, we must keep all the clients sync on this validation rule if we moved it to client side.

Now the client kind i know:

- java script 
- java
- python

Even if we all modified and sync the rule between all the client, we still can not prevent some private modification to break it.

revans2:
@danny0405 This is just a performance/usability optimization.  The checks are still happening on the server side, so the other clients will still get an exception when they try to upload a jar, just like they did before, but the java client will not even try to upload the jar.  It will fail faster.

danny0405:
@revans2 
You are right, i just thought if we can keep all these clients sync in rule, but fail fast just for this is ok.

revans2:
I agree it would be great to keep all of the clients in sync, but the java is the only high level client that we support.  The only other client we generate the code for is a python client, but it is just the generated thrift code, not a high level API like for storm submitter.

ManoharVanam:
Yes , here we are handling two things,
1. Fail fast without uploading the jar
2. Giving clear message to the user at the client side.

ManoharVanam:
@HeartSaVioR done.

PR #2708, opened 2018-06-07
STORM-3098: Fixed method filterChangingBlobsFor()

zd-project:


zd-project:
I extended the test to cover the testing of filterChangingBlob now.

PR #2710, opened 2018-06-08
STORM-3099: Extend metrics on supervisor, workers and DRPC

zd-project:
This PR depends on #2743 and #2771.

New supervisor level metrics: 

- Worker Kill/Restart Statistics
	- [x] Kill Count by Category - assignment change/HB too old/Heap space/blob change?
	- [ ] Worker Suicide Cnt - category:  internal error or Assignment Change
		- [x] - Implementation reports all process exit cases instead. Does not actually differentiate suicide from normal exit.
	- [x] Worker idle period
		- The metrics records the duration machine spent in each state (in histogram) and how many times it transitions into/out from a certain state.
	- [x] Time to Actually Kill worker (from identifying need by supervisor and actual change in the state of the worker) - (This is only an estimation, accuracy affected by SleepTime)
		- This does not tracks `storm kill_workers` as it's a client-side command and does not invoke StormMetricsRegistry.
	- [x] Time to start worker for topology from reading assignment for the first time.
	- [x] Worker cleanup time
		- This does not tracks duration for `cleanupForRelaunch`.
- [x] Supervisor Level Metrics:
	- [x] Supervisor restart Count
		- Simply report every time it restarts.
	- [x] Blobstore (Request to download time)
		- [x] Download time individual blob (inside localizer) localizer gettting requst to actually download hdfs request to finish
			- I assume this to be [the complete process] from initiating download to commit download to local blob cache and inform relative workers
		- [x] Download rate individual blob (inside localizer)
			- This is tracks the actual download rate of a blob retrieval, in MB/s
		- [x] Supervisor localizer thread blob download - how long (outside localizer)
			- I put this inside async localizer as it turns out to be better suited for purpose. This tracks the time for a topology blob download request to be completely processed.
		- [x] Blob update is also considered.
	- [x] Blobstore Update due to Version change Cnts
    - [x] Worker exceptions count for clean-up, kill and force-kill.
    - [x] Internal Error/Exception Cnt. on running external process/linux commands

DRPC metrics:

- [x] Avg/Max Time to respond to Http Request

### DIFF
--- a/storm-client/src/jvm/org/apache/storm/daemon/supervisor/ClientSupervisorUtils.java
+++ b/storm-client/src/jvm/org/apache/storm/daemon/supervisor/ClientSupervisorUtils.java
@@ -18,6 +18,7 @@
 
 package org.apache.storm.daemon.supervisor;
 
+import com.codahale.metrics.Meter;
 import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
@@ -28,11 +29,15 @@ import org.apache.storm.Config;
 import org.apache.storm.shade.org.apache.commons.lang.StringUtils;
 import org.apache.storm.utils.ConfigUtils;
 import org.apache.storm.utils.ObjectReader;
+import org.apache.storm.utils.ShellUtils;
 import org.apache.storm.utils.Utils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class ClientSupervisorUtils {
+    //Worker launched through external commands, hence we count their exceptions toward shell exceptions
+    public static final Meter numWorkerLaunchExceptions = ShellUtils.numShellExceptions;
+
     private static final Logger LOG = LoggerFactory.getLogger(ClientSupervisorUtils.class);
 
     static boolean doRequiredTopoFilesExist(Map<String, Object> conf, String stormId) throws IOException {
@@ -125,7 +130,13 @@ public class ClientSupervisorUtils {
         if (environment != null) {
             procEnv.putAll(environment);
         }
-        final Process process = builder.start();
+        final Process process;
+        try {
+            process = builder.start();
+        } catch (IOException e) {
+            numWorkerLaunchExceptions.mark();
+            throw e;
+        }
         if (logPrefix != null || exitCodeCallback != null) {
             Utils.asyncLoop(new Callable<Long>() {
                 public Long call() {
--- a/storm-server/src/main/java/org/apache/storm/daemon/supervisor/Container.java
+++ b/storm-server/src/main/java/org/apache/storm/daemon/supervisor/Container.java
@@ -18,6 +18,8 @@
 
 package org.apache.storm.daemon.supervisor;
 
+import com.codahale.metrics.Meter;
+import com.codahale.metrics.Timer;
 import java.io.BufferedReader;
 import java.io.File;
 import java.io.IOException;
@@ -68,6 +70,13 @@ public abstract class Container implements Killable {
     private static final ConcurrentHashMap<Integer, TopoAndMemory> _reservedMemory =
         new ConcurrentHashMap<>();
 
+    private static final Meter numCleanupExceptions = StormMetricsRegistry.registerMeter("supervisor:num-cleanup-exceptions");
+    private static final Meter numKillExceptions = StormMetricsRegistry.registerMeter("supervisor:num-kill-exceptions");
+    private static final Meter numForceKillExceptions = StormMetricsRegistry.registerMeter("supervisor:num-force-kill-exceptions");
+    private static final Meter numForceKill = StormMetricsRegistry.registerMeter("supervisor:num-workers-force-kill");
+    private static final Timer shutdownDuration = StormMetricsRegistry.registerTimer("supervisor:worker-shutdown-duration-ns");
+    private static final Timer cleanupDuration = StormMetricsRegistry.registerTimer("supervisor:worker-per-call-clean-up-duration-ns");
+
     static {
         StormMetricsRegistry.registerGauge(
             "supervisor:current-used-memory-mb",
@@ -106,6 +115,8 @@ public abstract class Container implements Killable {
     protected String _workerId;
     protected ContainerType _type;
     private long lastMetricProcessTime = 0L;
+    private Timer.Context shutdownTimer = null;
+
     /**
      * Create a new Container.
      *
@@ -204,20 +215,34 @@ public abstract class Container implements Killable {
     @Override
     public void kill() throws IOException {
         LOG.info("Killing {}:{}", _supervisorId, _workerId);
-        Set<Long> pids = getAllPids();
+        if (shutdownTimer == null) {
+            shutdownTimer = shutdownDuration.time();
+        }
+        try {
+            Set<Long> pids = getAllPids();
 
-        for (Long pid : pids) {
-            kill(pid);
+            for (Long pid : pids) {
+                kill(pid);
+            }
+        } catch (IOException e) {
+            numKillExceptions.mark();
+            throw e;
         }
     }
 
     @Override
     public void forceKill() throws IOException {
         LOG.info("Force Killing {}:{}", _supervisorId, _workerId);
-        Set<Long> pids = getAllPids();
+        numForceKill.mark();
+        try {
+            Set<Long> pids = getAllPids();
 
-        for (Long pid : pids) {
-            forceKill(pid);
+            for (Long pid : pids) {
+                forceKill(pid);
+            }
+        } catch (IOException e) {
+            numForceKillExceptions.mark();
+            throw e;
         }
     }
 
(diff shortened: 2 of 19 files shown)


storm-server/src/main/java/org/apache/storm/daemon/supervisor/Slot.java
```diff
@@ -410,9 +420,11 @@ static DynamicState handleWaitingForBlobLocalization(DynamicState dynamicState,
         assert (dynamicState.pendingDownload != null);
         assert (dynamicState.container == null);
 
-        //Ignore changes to scheduling while downloading the topology blobs
+        // Ignore changes to scheduling while downloading the topology blobs
         // We don't support canceling the download through the future yet,
```
danny0405: nit: whitespace

danny0405:
Sorry, i didn't see any practical changes for this patch but only some comment java doc. Can you describe it?

zd-project:
I’ll make incremental commits to this pull request as it involves multiple changes, as described in the original apache issue page. Thanks.

HeartSaVioR:
@zd-project Please append "[WIP]" in the title and remove when you finished.

zd-project:
New supervisor level metrics: 

- [ ] Worker Kill/Restart Statistics
	- [x] Kill Count by Category - assignment change/HB too old/Heap space (memory limit?)
		- [x] blob change?
	- [ ] Worker Suicide Cnt - category:  internal error or Assignment Change
		- [x] - Implemented based on running status the container's main process. Does not actually reflect suicide count because it counts the normal exit as well.
	- [x] Worker idle period
		- The metrics records the duration machines spent in each state (in histogram) and how many times it transition into/out to a certain state.
	- [x] Time to Actually Kill worker (from identifying need by supervisor and actual change in the state of the worker) - (This is only an estimation, accuracy affected by SleepTime)
	- [x] Time to start worker for topology from reading assignment for the first time.
	- [x] Worker cleanup time
- [x] Supervisor Level Metrics:
	- [x] Supervisor restart Count
		- simply report everytime it restarts.
	- [x] Blobstore (Request to download time)
		- [x] download time individual blob (inside localizer) localizer gettting requst to actually download hdfs request to finish
			- I assume this to be [the complete process] from initiating download to commit download to local blob cache and inform relative workers
		- [x] download rate individual blob (inside localizer)
			- This is tracks the actual download rate of a blob retrieval, in MB/s
		- [x] supervisor localizer thread blob download - how long (outside localizer)
			- I put this inside async localizer as it turns out to be better suited for purpose. This tracks the time for a topology blob download request to be completely processed.
		- [x] Blob update is also considered.
	- [x] Blobstore Update due to Version change Cnts

zd-project:
DRPC metrics:

- [x] Avg/Max Time to respond to Http Request

HeartSaVioR:
@zd-project 
Just 2 cents: I can see couple of other changes than the issue itself: please try to avoid side-effect on the patch. Let's file issues and raise pull requests if they're real issues. It's OK to just raise a pull request if you think the patch is minor.

storm-server/src/main/java/org/apache/storm/daemon/drpc/DRPC.java
```diff
@@ -208,6 +208,7 @@ public void failRequest(String id, DRPCExecutionException e) throws Authorizatio
         }
     }
 
+    //Should this be changed to private?
```
HeartSaVioR: Leave this to the comment on PR or file an issue. TODO-like comment is tend to not be addressed afterwards.

HeartSaVioR:
Could you please rebase? I'm expecting the code diff would be reduced after that.

zd-project:
Rebase completed with one new commit added. The new TimerDecoratedAssignment is incorporated.

zd-project:
I'm kind of wondering whether decorator pattern is better than subclassing or not, for TimedPortAndAssignment, because itself is already a wrapper.

zd-project:
Two more metrics for Supervisor

- [x] Worker exceptions count for clean-up, kill and force-kill.
- [x] Internal Error/Exception Cnt. on running external process/linux commands

storm-client/src/jvm/org/apache/storm/daemon/supervisor/ClientSupervisorUtils.java
```diff
@@ -28,11 +29,14 @@
 import org.apache.storm.shade.org.apache.commons.lang.StringUtils;
 import org.apache.storm.utils.ConfigUtils;
 import org.apache.storm.utils.ObjectReader;
+import org.apache.storm.utils.ShellUtils;
 import org.apache.storm.utils.Utils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class ClientSupervisorUtils {
+    public static final Meter numWorkerLaunchExceptions = ShellUtils.numShellExceptions;
```
revans2: This all feels a bit too confusing to me.

`ShellUtils.numShellExceptions` is static and pulled in from multiple different locations, but only registered once in the supervisor. I personally would rather see it where `ClientSupervisorUtils` registers the Meter, as it is tied to the supervisor having it do it here is fine.

For ShellUtils, I would like to see it not have a static meter, but instead optionally pass a meter in when calling run, or perhaps optionally include it in the constructor.

storm-server/src/main/java/org/apache/storm/daemon/supervisor/Slot.java
```diff
@@ -1147,8 +1162,19 @@ public DynamicState withPendingLocalization(Future<Void> pendingDownload) {
          */
         public DynamicState withState(final MachineState state) {
             long newStartTime = Time.currentTimeMillis();
+            //TODO: potential lost metrics due to timing accuracy (Timer only tracks one call per millisecond)
```
revans2: Please don't leave TODOs int he code.  Either fix it, file a follow on JIRA to fix it, or accept it and just have it be a comment and not a TODO.

revans2:
Also could we squash the commits some?

storm-client/src/jvm/org/apache/storm/daemon/supervisor/ClientSupervisorUtils.java
zd-project: Logged in Jira

zd-project:
Should I squash to one per daemon?

storm-client/src/jvm/org/apache/storm/daemon/supervisor/ClientSupervisorUtils.java
Ethanlm: I would like to see some comments so people can understand why `ClientSupervisorUtils` is using the same meter with `ShellUtils`

storm-server/src/main/java/org/apache/storm/localizer/AsyncLocalizer.java
```diff
@@ -251,11 +261,18 @@ private LocalizedResource getUserFile(String user, String key) {
                                 long localVersion = blob.getLocalVersion();
                                 long remoteVersion = blob.getRemoteVersion(blobStore);
                                 if (localVersion != remoteVersion || !blob.isFullyDownloaded()) {
+                                    if (!blob.isFullyDownloaded()) {
```
Ethanlm: Should be `if (!blob.isFullyDownloaded()) {`

storm-server/src/main/java/org/apache/storm/daemon/supervisor/Container.java
```diff
@@ -346,6 +358,10 @@ public void cleanUp() throws IOException {
             _usedMemory.remove(_port);
             _reservedMemory.remove(_port);
             cleanUpForRestart();
+        } catch (IOException e) {
+            //This may or may not be reported depending on when process exits
```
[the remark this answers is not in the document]
zd-project: My bad. I thought Container runs on it's own process when I started working on this. This should not be a problem if it's just a daemon thread.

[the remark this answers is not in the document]
zd-project: I'll remove this after review is completed

storm-server/src/main/java/org/apache/storm/localizer/AsyncLocalizer.java
```diff
@@ -365,6 +384,7 @@ void addReferencesToBlobs(PortAndAssignment pna, BlobChangingCallback cb)
         if (!localResourceList.isEmpty()) {
             getBlobs(localResourceList, pna, cb);
         }
+        pna.complete();
```
Ethanlm: just one more nit: put this outside of this method since the method `addReferencesToBlobs` doesn't indicate `complete`

Ethanlm:
@HeartSaVioR @danny0405 Do you want to take a look again? Otherwise I think it's good to merge it.

Ethanlm:
@zd-project  Could you squash all the commits? Will merge this in.

PR #2711, opened 2018-06-10
STORM-3100 : Introducing CustomIndexArray to speedup HashMap lookups

roshannaik:
- Added Unit Tests
- Ran below perf numbers on my mid 2015 macbook pro for single worker.

Ran these perf topos to check perf impact:

### 1) ConstSpoutIdBoltNullBoltTopo With 1 Acker
**Before:**  Throughput= 1,067,128 /sec .  Latency = 1.038 ms
**After:**  Throughput= 1,294,511 /sec .  Latency = 1.037 ms
- Throughput improved ~21%.  No signifiant change in latency.

### 2) ConstSpoutIdBoltNullBoltTopo With 0 Acker
**Before:**  Throughput= 4,389,133
**After:**  Throughput= 4,748,744 /sec

- Throughput improved ~8%.

### 3) TVL : --rate 350000 --spouts 1 --splitters 2 --counters 2 -c topology.max.spout.pending=500 -c topology.acker.executors=2  -c topology.disable.loadaware.messaging=true

No significant change observed in Throughput or Latency.

### DIFF
--- a/examples/storm-elasticsearch-examples/src/main/java/org/apache/storm/elasticsearch/common/EsTestUtil.java
+++ b/examples/storm-elasticsearch-examples/src/main/java/org/apache/storm/elasticsearch/common/EsTestUtil.java
@@ -28,6 +28,7 @@ import org.apache.storm.tuple.Fields;
 import org.apache.storm.tuple.Tuple;
 import org.apache.storm.tuple.TupleImpl;
 import org.apache.storm.tuple.Values;
+import org.apache.storm.utils.CustomIndexArray;
 import org.elasticsearch.cluster.ClusterName;
 import org.elasticsearch.cluster.metadata.IndexMetaData;
 import org.elasticsearch.common.settings.Settings;
@@ -56,7 +57,7 @@ public final class EsTestUtil {
         GeneralTopologyContext topologyContext = new GeneralTopologyContext(
                 builder.createTopology(),
                 new Config(),
-                new HashMap<>(),
+                new CustomIndexArray<String>(0,0),
                 new HashMap<>(),
                 new HashMap<>(),
                 "") {
--- a/external/storm-elasticsearch/src/test/java/org/apache/storm/elasticsearch/common/EsTestUtil.java
+++ b/external/storm-elasticsearch/src/test/java/org/apache/storm/elasticsearch/common/EsTestUtil.java
@@ -39,6 +39,7 @@ import org.apache.storm.tuple.Fields;
 import org.apache.storm.tuple.Tuple;
 import org.apache.storm.tuple.TupleImpl;
 import org.apache.storm.tuple.Values;
+import org.apache.storm.utils.CustomIndexArray;
 import org.elasticsearch.action.admin.cluster.health.ClusterHealthResponse;
 import org.elasticsearch.action.admin.indices.delete.DeleteIndexRequest;
 import org.elasticsearch.client.Requests;
@@ -63,7 +64,7 @@ public class EsTestUtil {
     public static Tuple generateTestTuple(String source, String index, String type, String id) {
         TopologyBuilder builder = new TopologyBuilder();
         GeneralTopologyContext topologyContext = new GeneralTopologyContext(builder.createTopology(),
-                new Config(), new HashMap<>(), new HashMap<>(), new HashMap<>(), "") {
+                new Config(), new CustomIndexArray<String>(0,0), new HashMap<>(), new HashMap<>(), "") {
             @Override
             public Fields getComponentOutputFields(String componentId, String streamId) {
                 return new Fields("source", "index", "type", "id");
--- a/external/storm-hdfs/src/test/java/org/apache/storm/hdfs/bolt/format/TestSimpleFileNameFormat.java
+++ b/external/storm-hdfs/src/test/java/org/apache/storm/hdfs/bolt/format/TestSimpleFileNameFormat.java
@@ -17,6 +17,7 @@ import java.text.SimpleDateFormat;
 import java.util.HashMap;
 import java.util.Map;
 import org.apache.storm.task.TopologyContext;
+import org.apache.storm.utils.CustomIndexArray;
 import org.apache.storm.utils.Utils;
 import org.junit.Assert;
 import org.junit.Test;
@@ -69,8 +70,8 @@ public class TestSimpleFileNameFormat {
     }
 
     private TopologyContext createTopologyContext(Map<String, Object> topoConf) {
-        Map<Integer, String> taskToComponent = new HashMap<Integer, String>();
-        taskToComponent.put(7, "Xcom");
+        CustomIndexArray<String> taskToComponent = new CustomIndexArray<>(7,7);
+        taskToComponent.set(7, "Xcom");
         return new TopologyContext(null, topoConf, taskToComponent, null, null, null, null, null, null, 7, 6703, null, null, null, null,
                                    null, null);
     }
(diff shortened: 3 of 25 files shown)


roshannaik:
Looks like tester_bolt_metrics.py under storm-core is failing. I see that it can be run via mvn clojure:test under storm-core. But unable to figure out what really going wrong. Any help would be appreciated. thanks.

HeartSaVioR:
It looks like Zk issue but not 100% sure. If it is happening intermittently we might be fine. Could you setup Travis to your fork and run build couple of times?

storm-client/src/jvm/org/apache/storm/utils/CustomIndexArray.java
```diff
@@ -0,0 +1,174 @@
...
+     * Returns the number of elements (including null elements).
+     */
+    public int size() {
+        return elemCount;
+    }
+
+    /**
+     * Always returns true as this cannot be empty.
+     */
+    public boolean isEmpty() {
```
revans2: Why have this at all.  It does not implement collection so we don't need it.

storm-client/src/jvm/org/apache/storm/utils/CustomIndexArray.java
```diff
@@ -0,0 +1,174 @@
...
+        if (actualIndex < 0  ||  actualIndex >= elemCount) {
+            return false;
+        }
+        return true;
+    }
+
+    /**
+     * Returns the index range as a Set.
+     */
+    public Set<Integer> indices() {
```
revans2: Nit: we are keeping the order of the indices here, but it is not documented in the javadocs, and it is not needed, as the only place this is used is for doing a set difference.

It would also be nice to indicate that this is not a cheap call so people don't abuse it calling it a lot.

external/storm-elasticsearch/src/test/java/org/apache/storm/elasticsearch/common/EsTestUtil.java
```diff
@@ -63,7 +64,7 @@
     public static Tuple generateTestTuple(String source, String index, String type, String id) {
         TopologyBuilder builder = new TopologyBuilder();
         GeneralTopologyContext topologyContext = new GeneralTopologyContext(builder.createTopology(),
-                new Config(), new HashMap<>(), new HashMap<>(), new HashMap<>(), "") {
+                new Config(),  new CustomIndexArray<String>(0,1), new HashMap<>(), new HashMap<>(), "") {
```
revans2: nit: extra space.

examples/storm-elasticsearch-examples/src/main/java/org/apache/storm/elasticsearch/common/EsTestUtil.java
```diff
@@ -56,7 +57,7 @@ public static Tuple generateTestTuple(final String source,
         GeneralTopologyContext topologyContext = new GeneralTopologyContext(
                 builder.createTopology(),
                 new Config(),
-                new HashMap<>(),
+                new CustomIndexArray<String>(0,1),
```
revans2: Nit: java should be able to figure out that it is a `String` from the required type for GeneralTopologyContext

revans2: Also why does this need to be 0,1?

external/storm-hdfs/src/test/java/org/apache/storm/hdfs/bolt/format/TestSimpleFileNameFormat.java
```diff
@@ -69,8 +70,8 @@ public void testTimeFormat() {
     }
 
     private TopologyContext createTopologyContext(Map<String, Object> topoConf) {
-        Map<Integer, String> taskToComponent = new HashMap<Integer, String>();
-        taskToComponent.put(7, "Xcom");
+        CustomIndexArray<String> taskToComponent = new CustomIndexArray<>(0,8);
```
revans2: Why 0,8 if all we are setting is 7?

storm-client/src/jvm/org/apache/storm/daemon/Task.java
```diff
@@ -112,15 +112,15 @@ public Task(Executor executor, Integer taskId) throws IOException {
     public List<Integer> getOutgoingTasks(Integer outTaskId, String stream, List<Object> values) {
         if (debug) {
             LOG.info("Emitting direct: {}; {} {} {} ", outTaskId, componentId, stream, values);
-        }
```
revans2: I think grouping check should happen even if debug is not enabled.

storm-server/src/main/java/org/apache/storm/Testing.java
```diff
@@ -660,8 +661,8 @@ public static Tuple testTuple(List<Object> values, MkTupleParam param) {
             }
         }
 
-        Map<Integer, String> taskToComp = new HashMap<>();
-        taskToComp.put(task, component);
+        CustomIndexArray<String> taskToComp = new CustomIndexArray<>(task,task+1);
```
revans2: If lower and upper are inclusive why do we need task, task+1?

storm-client/src/jvm/org/apache/storm/utils/CustomIndexArray.java
roshannaik: I felt its intuitive and nice to maintain the integers in order even if not absolutely necessary.

storm-client/src/jvm/org/apache/storm/utils/CustomIndexArray.java
```diff
@@ -0,0 +1,174 @@
...
+            if (key > upperIndex) {
+                upperIndex = key;
+            }
+        }
+        this.baseIndex = lowerIndex;
+        this.elemCount = upperIndex - lowerIndex + 1;
+
+        this.elements = makeNullInitializedArray(elemCount);
+
+        for (Map.Entry<Integer, T> entry : src.entrySet()) {
```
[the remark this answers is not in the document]
roshannaik: Yes, calling set() directly instead of  elements.set()  would have been nice..But normally I dont call instance methods from constructors since the object is not considered as fully initialized until the constructor has finished.

examples/storm-elasticsearch-examples/src/main/java/org/apache/storm/elasticsearch/common/EsTestUtil.java
roshannaik: It is mostly for readability .. when reading this code its unclear what the deduced type will be without further checking the parameter types for the constructor. Happy to change if you prefer the other way.

roshannaik: Just any non empty range.

storm-client/src/jvm/org/apache/storm/daemon/Task.java
roshannaik: That was part of an older PR. From what I recall ... on profiling I had noticed that the grouping check was expensive in the critical path due to the fact that it needed lookups in three (now down to 2) hashmaps :   streamComponentToGrouper  &  componentGrouping. Since neither were keyed on Integer, the CustomIndexArray style optimization was not possible.

roshannaik:
Sorry for not being able to address the comments sooner. Have been traveling and should be able to get back to this soon after jul 12 th.

HeartSaVioR:
@roshannaik 

https://travis-ci.org/apache/storm/jobs/403972342
https://travis-ci.org/apache/storm/jobs/403972347
https://travis-ci.org/apache/storm/jobs/403972348
https://travis-ci.org/apache/storm/jobs/403972349

Travis CI build is failing and failures look like related to the changeset. Could you check above build result and track?

HeartSaVioR:
Looks like the build still fails and relevant to this patch.

roshannaik:
Spent sometime trying to triage this failure but unable to figure out the cause. i know the issue has to do with this patch as it passes w/o this patch.  I think it will take me some more time to get to the bottom of this.  I did notice another (kind of significant) performance issue is simple to address. Given that this is a relatively minor optimization, I am planning to provide a fix for that one first, and then revert back to this.

PR #2712, opened 2018-06-12
Update Multilang-protocol.md

MaurGi:
Adding list of supported languages and a link to the c# adapter we have published recently -

### DIFF
--- a/docs/Multilang-protocol.md
+++ b/docs/Multilang-protocol.md
@@ -7,6 +7,18 @@ This page explains the multilang protocol as of Storm 0.7.1. Versions prior to 0
 
 # Storm Multi-Language Protocol
 
+## Supported Lanugages
+
+Storm Multi-Language has implementation in the following languages:
+
+- [JavaScript](../storm-multilang/javascript)
+- [Python](../storm-multilang/python)
+- [Ruby](../storm-multilang/ruby)
+
+Third party libraries are available for the following languages:
+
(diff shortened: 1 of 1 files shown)


arunmahadevan:
Looks good.

docs/Multilang-protocol.md
```diff
@@ -7,6 +7,18 @@ This page explains the multilang protocol as of Storm 0.7.1. Versions prior to 0
 
 # Storm Multi-Language Protocol
 
+## Supported Lanugages
+
+Storm Multi-Language has implementation in the following languages:
+
+- [JavaScript](../storm-multilang/javascript)
```
HeartSaVioR: The docs will be compiled as html and placed to website, so docs should not refer outside of docs directory. You can describe maven artifacts instead, like
* Javascript: org.apache.storm / multilang-javascript
* Python: org.apache.storm / multilang-python
* Ruby: org.apache.storm / multilang-ruby

MaurGi:
thanks for the feedback, modified and resubmitted -

docs/Multilang-protocol.md
MaurGi: got it - fixed, thanks

MaurGi:
@HeartSaVioR how do I squash an existing PR in GitHub? I don't see any option on the UI, do I make a new branch? thanks -

HeartSaVioR:
@MaurGi OK never mind. I'll handle it during merging. Thanks for the contribution!

PR #2714, opened 2018-06-13
STORM-3101: Fix unexpected metrics registration in StormMetricsRegistry.

zd-project:
The current implementation is based on the idea to set the source to the running daemon. All the metrics not belonging to the source (e.g., supervisor) will be removed or rejected. This implementation also adds in the utility method for naming a metric.

Please refer to the apache issue page for the purpose of this improvement. This is actually a band-aid fix. I'm wondering if there's better approach.

### DIFF
--- a/storm-server/src/main/java/org/apache/storm/metric/StormMetricsRegistry.java
+++ b/storm-server/src/main/java/org/apache/storm/metric/StormMetricsRegistry.java
@@ -18,8 +18,13 @@ import com.codahale.metrics.Meter;
 import com.codahale.metrics.Metric;
 import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.Reservoir;
+
 import java.util.Map;
+import java.util.SortedSet;
 import java.util.concurrent.Callable;
+
+import org.apache.commons.lang.NullArgumentException;
+import org.apache.storm.cluster.DaemonType;
 import org.apache.storm.daemon.metrics.MetricsUtils;
 import org.apache.storm.daemon.metrics.reporters.PreparableReporter;
 import org.slf4j.Logger;
@@ -27,8 +32,60 @@ import org.slf4j.LoggerFactory;
 
 @SuppressWarnings("unchecked")
 public class StormMetricsRegistry {
+    // TODO: Adding a secondary reporting_registry for filtering?
     public static final MetricRegistry DEFAULT_REGISTRY = new MetricRegistry();
     private static final Logger LOG = LoggerFactory.getLogger(StormMetricsRegistry.class);
+    private static String source = null;
+
+    public static boolean setSource(DaemonType source) {
+        return setSource(source.toString());
+    }
+
+    /**
+     * Specify source of metrics to filter out accidental registry.
+     * This should be called as early as possible
+     * @param source Components all the metrics should belong to
+     * @return whether the source has been set successfully
+     */
+    public static boolean setSource(String source) throws NullArgumentException {
+        if (StormMetricsRegistry.source == null) {
+            if (source != null) {
+                StormMetricsRegistry.source = source;
+                //#getNames returns a snapshot of the registry,
+                // hence safe to perform remove
+                for (String name : DEFAULT_REGISTRY.getNames()) {
+                    if (!name.startsWith(source)) {
+                        DEFAULT_REGISTRY.remove(name);
+                    }
+                }
+                return true;
+            } else {
+                throw new NullArgumentException("Registry source");
+            }
+        } else {
+            //Ignore all subsequent changes if the source has been initialized
+            return false;
+        }
+    }
+
+    public static String name(DaemonType source, String metrics) {
+        return name(source.toString(), metrics);
+    }
+
+    /**
+     * Helper function for naming metrics.
+     * @param source Daemon/component the metrics belong to
+     * @param metrics name of the metrics
+     * @return the name of the metrics following its parenting component
+     * @throws NullArgumentException source shouldn't be null
+     */
+    public static String name(String source, String metrics) throws NullArgumentException {
+        if (source != null) {
+            return source + ":" + metrics;
+        } else {
+            throw new NullArgumentException("Registry source");
+        }
+    }
 
     public static Meter registerMeter(String name) {
         Meter meter = new Meter();
@@ -73,6 +130,12 @@ public class StormMetricsRegistry {
     }
 
     private static <T extends Metric> T register(String name, T metric) {
+        if (source != null && !name.startsWith(source)) {
+            LOG.warn("Metric {} accidentally registered and filtered", name);
+            //Mute metrics by not registering it. This is a band-aid fix
+            // TODO: Find a better solution
+            return metric;
+        }
         T ret;
         try {
             ret = DEFAULT_REGISTRY.register(name, metric);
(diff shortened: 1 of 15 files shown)


zd-project:
Some questions:
1. Currently, by rejecting a metric, StormMetricsRegistry simply discard it and log the warning without throwing any errors to the calling method/classes. Is this an appropriate practice?
2. Should we instead use a secondary registry for filtering and reporting only, apart from the primary which does include all registrations?
3. Not sure how to handle RocksDB metrics as it isn't technically a daemon.

storm-server/src/main/java/org/apache/storm/metric/StormMetricsRegistry.java
```diff
@@ -73,6 +130,12 @@ private static void startMetricsReporter(PreparableReporter reporter, Map<String
     }
 
     private static <T extends Metric> T register(String name, T metric) {
+        if (source != null && !name.startsWith(source)) {
```
agresch: I think it would be better to register with a Daemon type enum and metric name.  This would allows metrics that are daemon specific and general to work for a given daemon.

zd-project: I was originally thinking so but then I found out there are some other components not specified as 'daemon' in DaemonType but also using metrics, such as logviewer and RocksDB. I wasn't sure how to handle the case so internally I convert them all to String. It's possible to register both with a Daemon and with a name in String.

zd-project: Thanks for the clarification. Anyone else also has suggestions in the implementation?

Ethanlm:
Thanks for the contribution. It's good to exploit more on metrics functionality. 

I don't see the real need for adding filters since it's all storm internal code. But if you think it's better to have,  I would suggest to use a filter layer to make this more flexible. 

For example, 

Create a filter interface with a filter function. e.g.
```
public interface IStormMetricsRegistryFilter {
   public default boolean filter(String metricName) {
        return false;
    }
}
```
then you can have a function (e.g. addFilter) in StormMetricsRegistryFilter to add real implementation of the filter interface before StormMetricsRegistry starts to registerMeters().

The above is a very simple interface and might not be able to do much except filtering based on the `String` parameter. You can think about more on this.

I think the current implementation in this PR won't work because you are calling `setsource()` to late

zd-project:
Another implementation would be to require all metrics to be registered when at least one instance of a class has been instantiated. This will disqualify the final property of all metrics variables but we can encapsulate all the assignment to a static method and invoked at appropriate time.

For example,

```
class Scratch {
    private static Integer x = null;

    public Scratch() {
        if (Scratch.x == null) {
            x = 7;
        }
    }
}
```

srdo:
It might be good to consider whether we could move away from `static` for metrics, and consider if we could do dependency injection instead? The cause of the issue you mention on JIRA (accidentally loading metrics) seems to me to be because the metrics registration is `static`.

We're already kind of using StormMetricsRegistry/StormMetricRegistry as if they belong to the daemon classes, because the daemons are responsible for calling `start`/`stop` on them.

I think for many of the daemons (e.g. Nimbus, Supervisor, DRPCServer), we could probably instantiate a StormMetricsRegistry in the main method or main class constructor, and pass that down to any other classes that need it. That way we could move metrics registration to be non-static, so we don't run into this kind of problem. 

What do you think @zd-project?

danny0405:
@srdo 
I kind of saw the Apache Flink metrics system, is has a internal metrics layer on top of the Yammer Metrics, every Role in Flink has a MetricsGroup, one MetricsGroup has many Metrics registered in. This metrics group can be passed around along with the Role, and user can chose to register their metrics in very easily.

srdo:
@danny0405 A quick search for Role or MetricsGroup in the flink repo didn't turn up anything. Could you elaborate on what you mean, and why/how we could use Flink's mechanism here?

srdo:
@danny0405 Okay, thanks. I think the MetricRegistryImpl class looks nice https://github.com/apache/flink/blob/16ec3d7ea12c520c5c86f0721553355cc938c2ae/flink-runtime/src/main/java/org/apache/flink/runtime/metrics/MetricRegistryImpl.java. I'd think we could do something similar with StormMetricsRegistry, where we create one as part of e.g. starting Nimbus, and then that instance is passed around via injection (e.g. as in https://github.com/apache/flink/blob/4de72bbee189ab357e4d9e6fea33e27ff1ab233f/flink-runtime/src/main/java/org/apache/flink/runtime/minicluster/MiniCluster.java#L239). I assume this is what you mean?

danny0405:
@srdo Yeah, this is a valuable promotion for storm metrics and the implementation will be much cleaner.

revans2:
We have 4 separate metrics APIs right now

2 for daemons
https://github.com/apache/storm/blob/master/storm-server/src/main/java/org/apache/storm/metric/api/IClusterMetricsConsumer.java
https://github.com/apache/storm/blob/master/storm-server/src/main/java/org/apache/storm/metric/StormMetricsRegistry.java

and similarly 2 for the workers.
https://github.com/apache/storm/blob/master/storm-client/src/jvm/org/apache/storm/metric/api/IMetricsConsumer.java
https://github.com/apache/storm/blob/master/storm-client/src/jvm/org/apache/storm/metrics2/StormMetricRegistry.java

IClusterMetricsConsumer and  IMetricsConsumer are way too open so we deprecated them, or should do so.  They allowed you to send anything (literately an Object) but they had a lot of context about the metrics, a.k.a. dimensions, which process it came from, what component, etc.  The new APIs fix the Object issue, because they are based off of the dropwizard/yammer/codahale API, but it does not support dimensions. 

The Flink APIs appears to more or less fix the dimensions issue, but it is a fork of codahale/yammer.

https://github.com/apache/flink/blob/16ec3d7ea12c520c5c86f0721553355cc938c2ae/flink-metrics/flink-metrics-core/pom.xml

They have no external dependencies. 

So are you proposing that we fork just like flink does?  do you want us to use the flink-metrics-core instead as a backend?  Either of these choices will require that we have at a minimum a new API for reporting metrics to other systems, and possibly a 3rd API for registering metrics.  This is not a small amount of work, and it is possibly going to be painful to our users, who may have spent some time trying to move to the new APIs.  I am not opposed to it, as it will solve some problems and ugliness that we currently have, I just want to be sure that we understand the ramifications of a move like this.

On the plus side it would make it so we could have a single API for all metrics.

zd-project:
Looks like StormMetricRegistry has already had filtering by daemon type feature that we need. It also supports report at different intervals and such. Apart from the injection suggestion, do you think it's better to consider moving towards this registry in the future, instead of StormMetricsRegistry (Mind the 's' in the class name lol).

zd-project:
Another thought on improvement of StormMetricsRegistry in general. We should have something that works better with dependency as well. In the latest commit to #2710 that addresses metric of exceptions from external shell command, I have to register metrics from storm-client component to StormMetricsRegistry, which is in storm-server. To bypass the dependency I ended up declaring the meter in ShellUtil class but register it when launching the supervisor. But this does prompt me to think about decoupling between declaration and registration. Maybe we can have an Enum class that declares all metrics (since all of them are singleton anyway); components can make reference to them, and daemons can register/unregister them on demand when launching. It also makes maintaining existing metrics a bit easier as now we have a place to declare and organize all metrics. 
@srdo @danny0405 @revans2

zd-project:
This should be closed after #2805 is resolved.

srdo:

PR #2716, opened 2018-06-13
STORM-3102 Remove check for partition offset before every emit.

acseidel:
In kafka >0.10.2 this check became expensive, causing large performance
decreases.


srdo:
+1

PR #2718, opened 2018-06-13
STORM-3103 allow nimbus to shutdown properly

agresch:
Nimbus registers exitProcess() for the StormTimer.onKill routine.  This does not return once called.  This basically locks up the timer with active being set.

Then when Nimbus shuts down, it calls timer.close().  Since the timer is stuck in the onKill routine forever with active set on, timer close is deadlocked waiting for the join() to occur here:

https://github.com/apache/storm/blob/master/storm-client/src/jvm/org/apache/storm/StormTimer.java#L173

The proposed fix is to just set active to false before calling the onKill() for the timer.

### DIFF
--- a/storm-client/src/jvm/org/apache/storm/StormTimer.java
+++ b/storm-client/src/jvm/org/apache/storm/StormTimer.java
@@ -35,7 +35,6 @@ public class StormTimer implements AutoCloseable {
      *
      * @param name   name of the timer
      * @param onKill function to call when timer is killed unexpectedly
-     * @return StormTimerTask object that was initialized
      */
     public StormTimer(String name, Thread.UncaughtExceptionHandler onKill) {
         if (onKill == null) {
@@ -67,7 +66,7 @@ public class StormTimer implements AutoCloseable {
     }
 
     /**
-     * Same as schedule with millisecond resolution
+     * Same as schedule with millisecond resolution.
      *
      * @param delayMs     the number of milliseconds to delay before running the function
      * @param func        the function to run
@@ -249,8 +248,9 @@ public class StormTimer implements AutoCloseable {
                 } catch (Throwable e) {
                     if (!(Utils.exceptionCauseIsInstanceOf(InterruptedException.class, e))
                         && !(Utils.exceptionCauseIsInstanceOf(ClosedByInterruptException.class, e))) {
-                        this.onKill.uncaughtException(this, e);
+                        // need to set active false before calling onKill() - current implementation does not return.
                         this.setActive(false);
+                        this.onKill.uncaughtException(this, e);
                     }
                 }
             }


Ethanlm:
The code change makes sense to me. Just wonder if the whole JVM is terminated immediately after exitProcess() is being called.

agresch:
It seems not to, as Nimbus was able to continue the shutdown process cleanly.

kishorvpatil:
Good catch @agresch

danny0405:
@agresch 
Yes the 
[https://github.com/apache/storm/blob/master/storm-client/src/jvm/org/apache/storm/StormTimer.java#L173](https://github.com/apache/storm/blob/master/storm-client/src/jvm/org/apache/storm/StormTimer.java#L173)
may throw an Interrupted Exception,
but 
`} catch (Throwable e) {
                    if (!(Utils.exceptionCauseIsInstanceOf(InterruptedException.class, e))
                        && !(Utils.exceptionCauseIsInstanceOf(ClosedByInterruptException.class, e))) {
                        this.onKill.uncaughtException(this, e);
                        this.setActive(false);
                    }
                }`
here it already make a decision that if the exception is not InterruptedException.class we shall register the uncaughtException, so in what situation you mean the deadlock will occur?

agresch:
@danny0405 - The case I saw was the Timer throwing a null pointer exception from mkAssignments() - fixed in https://github.com/apache/storm/pull/2693.  This causes the onKill callback to be called and locks up the timer thread.

agresch:
If it helps, here is the callstack from our original bug report:

2018-05-24 09:27:05.636 o.a.s.d.n.Nimbus main [INFO] Starting nimbus server for storm version '2.0.0.y'
2018-05-24 09:27:06.012 o.a.s.d.n.Nimbus timer [ERROR] Error while processing event
java.lang.RuntimeException: java.lang.NullPointerException
        at org.apache.storm.daemon.nimbus.Nimbus.lambda$launchServer$37(Nimbus.java:2685) ~[storm-server-2.0.0.y.jar:2.0.0.y]
        at org.apache.storm.StormTimer$1.run(StormTimer.java:111) ~[storm-client-2.0.0.y.jar:2.0.0.y]
        at org.apache.storm.StormTimer$StormTimerTask.run(StormTimer.java:227) ~[storm-client-2.0.0.y.jar:2.0.0.y]
        at org.apache.storm.daemon.nimbus.Nimbus.readAllSupervisorDetails(Nimbus.java:1814) ~[storm-server-2.0.0.y.jar:2.0.0.y]
        at org.apache.storm.daemon.nimbus.Nimbus.computeNewSchedulerAssignments(Nimbus.java:1906) ~[storm-server-2.0.0.y.jar:2.0.0.y]
        at org.apache.storm.daemon.nimbus.Nimbus.mkAssignments(Nimbus.java:2057) ~[storm-server-2.0.0.y.jar:2.0.0.y]
        at org.apache.storm.daemon.nimbus.Nimbus.mkAssignments(Nimbus.java:2003) ~[storm-server-2.0.0.y.jar:2.0.0.y]
        at org.apache.storm.daemon.nimbus.Nimbus.lambda$launchServer$37(Nimbus.java:2681) ~[storm-server-2.0.0.y.jar:2.0.0.y]
        ... 2 more
2018-05-24 09:27:06.023 o.a.s.u.Utils timer [ERROR] Halting process: Error while processing event
java.lang.RuntimeException: Halting process: Error while processing event
        at org.apache.storm.utils.Utils.exitProcess(Utils.java:469) ~[storm-client-2.0.0.y.jar:2.0.0.y]
        at org.apache.storm.daemon.nimbus.Nimbus.lambda$new$17(Nimbus.java:484) ~[storm-server-2.0.0.y.jar:2.0.0.y]
        at org.apache.storm.StormTimer$StormTimerTask.run(StormTimer.java:252) ~[storm-client-2.0.0.y.jar:2.0.0.y]
2018-05-24 09:27:06.032 o.a.s.d.n.Nimbus Thread-12 [INFO] Shutting down master
2018-05-24 09:27:06.032 o.a.s.u.Utils Thread-13 [INFO] Halting after 10 seconds

No further shutdown was processed.  It's entirely reproducible by forcing an NPE from the timer.

srdo:
+1

PR #2719, opened 2018-06-14
STORM-3105 Update hive.version to stable hive release.

aandis:
[STORM-3105](https://issues.apache.org/jira/browse/STORM-3105)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -289,7 +289,7 @@
         <jline.version>0.9.94</jline.version>
         <clojure-data-codec.version>0.1.0</clojure-data-codec.version>
         <clojure-contrib.version>1.2.0</clojure-contrib.version>
-        <hive.version>0.14.0</hive.version>
+        <hive.version>2.3.3</hive.version>
         <hadoop.version>2.6.1</hadoop.version>
         <hdfs.version>${hadoop.version}</hdfs.version>
         <hbase.version>1.1.12</hbase.version>


aandis:
cc: @roshannaik

HeartSaVioR:
This actually require full tests on storm-hive as well as storm-autocreds (delegation token). Could you share how do you test the patch manually?

aandis:
In our case, we have copied the hive [dependencies](https://github.com/apache/storm/blob/4137328b75c06771f84414c3c2113e2d1c757c08/external/storm-hive/pom.xml#L54) into our pom with the right version and can confirm that it works. I see that [storm-hive](https://github.com/apache/storm/tree/master/external/storm-hive/src/test/java/org/apache/storm/hive) already has some tests. 

The point here is that `hive.version` should match the hive version being used by the user. Setting it here is just a default and the default may not work if there's a version mismatch with the actual hive version. To address that I have opened [STORM-3106](https://issues.apache.org/jira/browse/STORM-3106)

revans2:
The current build is failing because of an Enforcer rule.

```
...
[INFO] --- maven-enforcer-plugin:1.4.1:enforce (enforce-maven-version) @ storm-hive ---
[WARNING] Rule 1: org.apache.maven.plugins.enforcer.BannedDependencies failed with message:
Found Banned Dependency: jdk.tools:jdk.tools:jar:1.8
Use 'mvn dependency:tree' to locate the source of the banned dependencies.
...
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-enforcer-plugin:1.4.1:enforce (enforce-maven-version) on project storm-hive: Some Enforcer rules have failed. Look above for specific messages explaining why the rule failed. -> [Help 1]
...
```

srdo:

PR #2720, opened 2018-06-15
STORM-3109: Fixed incorrect conversion from relative path to absolute path for STORM_LOCAL_DIR

zd-project:
See issue: https://issues.apache.org/jira/browse/STORM-3109


srdo:
Could you explain a bit more why the conversion to absolute path is a problem? I'm having a hard time understanding why kill_workers would break if the STORM_LOCAL_DIR is given as an absolute path.

zd-project:
It would break if it’s given as a relative path. File.getCanonicalPath converts it to cwd/STORM_LOCAL_DIR

srdo:
Ah right. Thanks, that makes sense. +1

PR #2721, opened 2018-06-15
STORM-3110: Skip the user while checking isProcessAlive

arunmahadevan:
While running in secure mode, supervisor sets the worker user (in workers local state) as the user that launched the topology.

However the worker OS process does not actually run as the user (instead runs as storm user) unless `supervisor.run.worker.as.user` is also set.

If the supervisor's assignment changes, the supervisor in some cases checks if all processes are dead by matching the "pid+user". Here if the worker is running as a different user (say storm) the supervisor wrongly assumes that the worker process is dead.

Later when supervisor tries to launch a worker at that same port, it throws a bind exception

We can skip the user check while determining if the worker processes are alive.

### DIFF
--- a/storm-core/src/jvm/org/apache/storm/daemon/supervisor/Container.java
+++ b/storm-core/src/jvm/org/apache/storm/daemon/supervisor/Container.java
@@ -20,7 +20,6 @@ package org.apache.storm.daemon.supervisor;
 import java.io.BufferedReader;
 import java.io.File;
 import java.io.IOException;
-import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.Writer;
 import java.lang.ProcessBuilder.Redirect;
@@ -33,7 +32,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
-import com.google.common.annotations.VisibleForTesting;
 import org.apache.storm.Config;
 import org.apache.storm.generated.LSWorkerHeartbeat;
 import org.apache.storm.generated.LocalAssignment;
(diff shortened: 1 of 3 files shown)


arunmahadevan:
@revans2 , @HeartSaVioR, can you take a look?

danny0405:
For security, I think we need to keep the if user check, we should fix it but not through deleting it.

arunmahadevan:
Unless `supervisor.run.worker.as.user` is set, the worker process runs as "storm" user. I guess the supervisor should always check "if all processes are dead" by just looking if the worker pids are alive than doing user comparison, since there is no mapping between the user that launched the topology (e.g kerberos user) and the actual OS user that worker is running as (this is always storm).. In the "run as user" container the "kill" command is launched by switching to the OS user that worker is actually running as here - https://github.com/apache/storm/blob/1.x-branch/storm-core/src/jvm/org/apache/storm/daemon/supervisor/RunAsUserContainer.java#L55 and that should take care of the security.

arunmahadevan:
Also note that supervisor might be killing the workers in response to an assignment change triggered by nimbus (may be due to user triggered kill or rebalance) and the supervisor has no idea of which end user/action might have triggered this. The authorization checks to see if the user has permissions to kill/rebalance the topology is already happening at nimbus so I think we are good.

HeartSaVioR:
I'm not an expert of Supervisor V2 but just based on my understanding of this issue, I guess we may want to address this to two different situations:

1. `supervisor.run.worker.as.user` is set

Supervisor should check the process with `worker` account (can be read via worker state).

2. `supervisor.run.worker.as.user` is unset

Supervisor should check the process with `supervisor` account, as workers will be always launched with `supervisor` account.

In any cases, if the worker is running with different account, it might be considered as wrong process or out of sync. Ideally I think it should be corrected (kill the process) but it might open security issue (account A editing pid file for worker A1 and let Supervisor kill arbitrary process running with other account) so error or critical log message might be a best bet.

I'm not sure how many cases we can encounter above case, but unless the hole is huge, I guess checking with above logic would work.

@arunmahadevan @revans2 Please correct me if I'm missing here and/or there's better way to handle this.

HeartSaVioR:
My ideal thought depends on the assumption that we could get OS account for current JVM process in safe way. If there's no way to do it, I agree this PR might be a best bet.

revans2:
So you are running with authentication, but run as user is disabled, because you are running on windows. That is a use case that I missed, sorry about that.  

This is not a security issue as we kill the process as the user that owns the container.  When run as user is enabled we do it as the user that launched the process, when it is disabled we do it as the supervisor user.  This means the OS will prevent us from killing a process we shouldn't, even if a worker self reports a process that it should not.

The entire reason I put the user check in was to avoid a problem with PID reuse.  OSes typically rotate through PIDs in a way to avoid situations where a process dies and its PID is quickly handed to another process, but with how storm self reports PIDs through the file system... I just wanted to be cautious.

Removing the user check entirely should be fine, the issue you might run into is that if a PID is reused too quickly by a different user you might be stuck trying to kill is process that you are not allowed to and end up stuck waiting for it to die forever.  The simple fix for me would be to have the Container pass in  `System.getProperty("user.name")` for the user name, with a comment about why, and have RunAsUserContainer pass the real user in.

arunmahadevan:
@revans2 , i have refactored the code so that we check for "user.name" in the normal container and the result of `getWorkerUser` in the RunAsUserContainer. 

The `getWorkerUser` is used in other places to set the user in the workers local state and apparently used to do some authorization checks in the log viewer. I did not want to affect that behavior so I introduced a new function to return the OS user that the worker is expected to run as that invokes the `getWorkerUser` in the run as container.

This issue happens in posix environment as well and most of the users dont set the `supervisor.run.worker.as.user` config but run storm in secure mode.

PR #2726, opened 2018-06-19
STORM-3090 - Fix bug when different topics use the same offset for a partition

choojoyq:
In the current implementation of `ZkCoordinator` deleted partition managers are used as state holders for newly created partition managers. Such behaviour was introduced in the scope of [this](https://issues.apache.org/jira/browse/STORM-2296) ticket. However existing lookup is based only on partition number.

```
Map<Integer, PartitionManager> deletedManagers = new HashMap<>();
for (Partition id : deletedPartitions) {
    deletedManagers.put(id.partition, _managers.remove(id));
}
for (PartitionManager manager : deletedManagers.values()) {
    if (manager != null) manager.close();
}
LOG.info(taskPrefix(_taskIndex, _totalTasks, _taskId) + " New partition managers: " + newPartitions.toString());

for (Partition id : newPartitions) {
    PartitionManager man = new PartitionManager(
        _connections,
        _topologyInstanceId,
        _state,
        _topoConf,
        _spoutConfig,
        id,
        deletedManagers.get(id.partition));
    _managers.put(id, man);
```
Which is definitely incorrect as the same task is able to manage multiple partitions with the same number but for different topics. In this case all new partition managers obtain the same offset value from a random deleted partition manager (as `HashMap` is used). And all fetch requests for the new partition managers fail with `TopicOffsetOutOfRangeException`. Some of them are recovered via this logic if assigned offset is smaller than the real one, but other continue to repetitively fail with offset out of range exception preventing fetching messages from Kafka.
```
if (offset > _emittedToOffset) {
    _lostMessageCount.incrBy(offset - _emittedToOffset);
    _emittedToOffset = offset;
    LOG.warn("{} Using new offset: {}", _partition, _emittedToOffset);
}
```
I assume that state holder lookup should be based both on topic and partition number.

### DIFF
--- a/external/storm-kafka/src/jvm/org/apache/storm/kafka/ZkCoordinator.java
+++ b/external/storm-kafka/src/jvm/org/apache/storm/kafka/ZkCoordinator.java
@@ -17,6 +17,7 @@
  */
 package org.apache.storm.kafka;
 
+import kafka.common.TopicAndPartition;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.apache.storm.kafka.trident.GlobalPartitionInformation;
@@ -33,8 +34,8 @@ public class ZkCoordinator implements PartitionCoordinator {
     int _totalTasks;
     int _taskId;
     String _topologyInstanceId;
-    Map<Partition, PartitionManager> _managers = new HashMap();
-    List<PartitionManager> _cachedList = new ArrayList<PartitionManager>();
+    Map<Partition, PartitionManager> _managers = new HashMap<>();
+    List<PartitionManager> _cachedList = new ArrayList<>();
     Long _lastRefreshTime = null;
     int _refreshFreqMs;
     DynamicPartitionConnections _connections;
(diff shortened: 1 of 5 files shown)


HeartSaVioR:
@choojoyq
First of all, thanks for your contribution. We are deprecating storm-kafka from 1.x version and plan to remove this entirely in 2.0 (there're couple of pull requests to be reviewed before that, though). 

So you may need to submit this PR against 1.x-branch.

choojoyq:
@HeartSaVioR 
Thanks for the fast reply. I pointed this pull request to 1.x-branch. 
Does it make sense to open one more pull request for 1.0.x-branch as well?

HeartSaVioR:
@choojoyq Let us port back first if it is OK to be merged.

choojoyq:
@HeartSaVioR 
Hi, do you probably know release date of the next storm version where this pull request potentially will be merged? We are facing this issue pretty often in our cluster which requires restarting of storm topology with ``ignoreZkOffsets`` flag, so we consider temporary solution with releasing of our own version of ``storm-kafka`` module if the next official version will not be released soon. Thanks.

external/storm-kafka/src/test/org/apache/storm/kafka/ZkCoordinatorTest.java
```diff
@@ -140,6 +140,42 @@ public void testPartitionManagerRecreate() throws Exception {
...
+        for (PartitionManager partitionManager : partitionManagersBeforeRefresh) {
+            partitionManager._emittedToOffset = 100L;
+            partitionManager._committedTo = 100L;
+        }
+
+        waitForRefresh();
+
+        when(reader.getBrokerInfo()).thenReturn(TestUtils.buildPartitionInfoList(
+            TestUtils.buildPartitionInfo("topic1", 1, 9093),
+            TestUtils.buildPartitionInfo("topic3", 1, 9093)));
```
srdo: Nit: Seems weird to exclude topic 2, that can't happen in a real setup. Shouldn't you be checking that topic1 and topic2 can't overwrite each others' offsets instead (e.g by making them different a few lines up)?

external/storm-kafka/src/jvm/org/apache/storm/kafka/ZkCoordinator.java
```diff
@@ -107,14 +110,14 @@ public void refresh() {
                         _stormConf,
                         _spoutConfig,
                         id,
-                        deletedManagers.get(id.partition));
+                        deletedManagers.get(new TopicAndPartition(id.topic, id.partition)));
```
srdo: Are you sure this lookup makes sense? The deletedManagers only contains partitions that this task is no longer responsible for (curr - mine). That partition shouldn't be able to also be in newPartitions (mine - curr). Won't this always return null (unless you're hitting the bug that this PR is trying to fix, namely that offsets are being copied from the PartitionManager for e.g. topic1 to the PartitionManager for topic2)?

srdo:
The changes look good to me, +1. Left some comments, but I don't think it makes sense to put too much effort into making the tests more realistic, or refactoring, given that storm-kafka is being removed.

external/storm-kafka/src/jvm/org/apache/storm/kafka/ZkCoordinator.java
choojoyq: I believe that this lookup exists for preserving information about partitions from other brokers in case of rebalancing. If for example one of brokers fail and its partitions are reassigned to other brokers this lookup should find information about this partition and reuse it during recreation of the same partition but on other host. ``equals()`` and ``hashCode()`` in Partition are based on topic, partition and host while this lookup in based only on topic and partition.

srdo: Good point, never mind then.

external/storm-kafka/src/test/org/apache/storm/kafka/ZkCoordinatorTest.java
choojoyq: The idea of the test that before rebalancing some ``Task`` was responsible for 2 partitions from topics 1 and 2 from broker 9092, but after rebalancing of Kafka brokers it became responsible for the same partition for topic 1 and another partition from topic 3 which are hosted now on broker 9093. So ``Task`` should preserve partition information about topic1 as it was managed by this task before and it knows how many messages was already emitted, committed, etc. While partition from topic3 is a brand new and so it should be created from scratch without reusing of any information. Before the fix information for topic3 would be reused from topic1 as lookup of previously managed partitions was performed only on partition number ignoring topic.

srdo: Thanks, it makes sense. Since there's only one task, topic 2 can't be removed though. It's not really important, the test makes sense regardless.

srdo:
Is this ready to merge @choojoyq?

choojoyq:
@srdo i think so, thanks for the review

PR #2728, opened 2018-06-19
STORM-3116: Set storm.home for clojure tests and during LocalCluster …

srdo:
…runs

https://issues.apache.org/jira/browse/STORM-3116

### DIFF
--- a/storm-server/src/main/java/org/apache/storm/LocalCluster.java
+++ b/storm-server/src/main/java/org/apache/storm/LocalCluster.java
@@ -135,6 +135,7 @@ public class LocalCluster implements ILocalClusterTrackedTopologyAware, Iface {
     private final List<Supervisor> supervisors;
     private final IStateStorage state;
     private final IStormClusterState clusterState;
+    private final String stormHomeBackup;
     private final List<TmpPath> tmpDirs;
     private final InProcessZookeeper zookeeper;
     private final IContext sharedContext;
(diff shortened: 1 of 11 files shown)


storm-server/src/main/java/org/apache/storm/LocalCluster.java
```diff
@@ -193,6 +194,10 @@ private LocalCluster(Builder builder) throws Exception {
             this.supervisors = new ArrayList<>();
             TmpPath nimbusTmp = new TmpPath();
             this.tmpDirs.add(nimbusTmp);
+            stormHomeBackup = System.getProperty(ConfigUtils.STORM_HOME);
```
HeartSaVioR: I'm not 100% sure that we need to use temporary directory for storm home while launching local cluster. Did you find any issue regarding this?

srdo: The local cluster starts up Nimbus, which tries to start RockDbStore, which reads storm.home. So far it's been erroring out, but because Nimbus swallows the error it hasn't been an issue, apart from the meaningless error logs. I figure we might as well set storm.home when using a local cluster, since it's always set in a real daemon, and that's the environment we're trying to simulate.

HeartSaVioR: Thanks for the details. Now I can understand the issue clearly.

PR #2732, opened 2018-06-21
STORM-3117 prevent deleting blobs while topologies still active

agresch:
Prevented crashing nimbus if blob is deleted during a getClusterInfo call.  However, once nimbus is restarted, if there was an active topology with a missing blob, nimbus will be stuck validatingAcls and not be able to schedule.

To get around this, I prevented deleting blobs unless a topology is not active.  As a bonus, this should also fix race conditions we have seen between the scheduling loop and a kill topology occurring with the cleanup thread deleting the blobs.


storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
```diff
@@ -3557,6 +3570,14 @@ public ByteBuffer downloadBlobChunk(String session) throws AuthorizationExceptio
     @Override
     public void deleteBlob(String key) throws AuthorizationException, KeyNotFoundException, TException {
         try {
+            String topoName = ConfigUtils.getIdFromBlobKey(key);
+            if (topoName != null) {
+                if (isTopologyActiveOrActivating(stormClusterState, topoName)) {
+                    String message = "Attempting to delete blob " + key + " from under active topology " + topoName;
+                    LOG.warn(message);
+                    throw new TException(message);
```
revans2: Could we make this an AuthorizationException?

agresch: I thought of that and wondered if that would be confusing to the users since it is a temporary exception.  Willing to change it though.

HeartSaVioR: Yeah may be better to have a specific exception to clearly state the situation.

storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
```diff
@@ -2672,7 +2680,12 @@ private ClusterSummary getClusterInfoImpl() throws Exception {
                 summary.set_assigned_memoffheap(resources.getAssignedMemOffHeap());
                 summary.set_assigned_cpu(resources.getAssignedCpu());
             }
-            summary.set_replication_count(getBlobReplicationCount(ConfigUtils.masterStormCodeKey(topoId)));
+            try {
+                summary.set_replication_count(getBlobReplicationCount(ConfigUtils.masterStormCodeKey(topoId)));
+            } catch (KeyNotFoundException e) {
+                // This could fail if a blob gets deleted by mistake.  Don't crash nimbus.
+                LOG.error("Unable to find blob entry", e);
```
HeartSaVioR: Does it represent one of blobs for a topology is broken? If then we may need to have specific error message, not general one like this. And if it can't be restored automatically, we may want to guide the way how to fix this situation.

agresch: @HeartSaVioR  - This change I made first in isolation (when still allowing the blob to be deleted).  This prevents nimbus from crashing, but a nimbus restart will still have issues.  With the other change, this line may no longer be necessary, but I would rather be defensive.  It should not hit unless we have further race conditions the other code check is missing.

storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
revans2: Thinking about it I think adding in a new exception would be the best way to go.  All of the APIs throw a TException so adding in a new more specific TException would not break any clients any worse than throwing a raw TException would.

agresch: Makes sense to me.  I prefer this to an AuthorizationException.

agresch:
@HeartSaVioR @revans2 - switched thrift exception to new IllegalStateException.

revans2:
@agresch looks like you missed adding in the new file

agresch:
@revans2 - just added it

PR #2733, opened 2018-06-23
STORM-3119: Build Storm with Java 10

srdo:
https://issues.apache.org/jira/browse/STORM-3119

There are some changes in this PR that are only tangentially related to building with JDK10, namely fixes to some flaky tests that would have made it a pain to add extra test runs to every PR.

Changes:
* Add Java 10 to the Travis build matrix. Storm-cassandra and storm-hive are excluded, because Cassandra and Hive aren't Java 9+ compatible yet.
* Add relevant Java EE APIs to components that use Jersey. The Java EE APIs aren't available without add-modules in Java 9+, so we should include them explicitly as dependencies. This is already done in most of the components, but missed a few spots.
* The integration test Vagrant box has been updated so the user's name is vagrant rather than ubuntu, fix this in the scripts
* Added environment variable to allow people to use Java 10 in integration tests run via Vagrant
* Fixed the integration test so it runs on Java 10. The default worker.childopts include removed JVM parameters related to GC logging. There doesn't appear to be parameters that work on both Java 8 and 10, so we'll need to add to the documentation how people should override the worker.childopts, if they want to run on Java 10.
* Some of the Time code didn't make sense, e.g. checking for null on the final THREAD_SLEEP_TIMES_NANOS map. It was left over from an earlier refactor of Time. Cleaned it up, and deleted the deprecated Time methods. Also made sure that when simulated time is advanced, idle threads are removed from the THREAD_SLEEP_TIMES_NANOS immediately. When the LocalCluster waits for the topology to be idle, it checks whether all timer threads are in the THREAD_SLEEP_TIMES_NANOS map. It looks to me like there was no guarantee that sleeping threads actually got a chance to run when time was advanced, because they remained in the THREAD_SLEEP_TIMES_NANOS map until they happened to be scheduled. With bad luck, a thread might end up being counted as idle when it just hadn't exited the Time.sleepUntilNanos loop yet.
* The metrics-tests were doing assertions in a racy way. We can't guarantee that metrics fall in exact buckets, because the metrics are updated from the executor async loops, which aren't included in the threads covered by LocalCluster.waitForIdle. I've replaced the checks with validation that there are a certain number of metrics buckets, and that the metrics counters sum to the expected value.

Example of a failing metrics test:
```
classname: org.apache.storm.metrics-test / testname: test-custom-metric-with-multi-tasks
expected: (clojure.core/= [1 0 0 0 0 0 2] (clojure.core/subvec (org.apache.storm.metrics-test/lookup-bucket-by-comp-id-&-metric-name! "2" "my-custom-metric") 0 N__3207__auto__))
  actual: (not (clojure.core/= [1 0 0 0 0 0 2] [1 0 0 0 0 0 0]))
      at: test_runner.clj:105
```
In this case the missing 2 counts happen in a later bucket than the test is checking for, because the executor async loop happened to not have run when the metrics system added the last bucket.

Let me know if I need to split this PR up, or if a change is unclear. I ran the tests a few times on Travis, and they appear stable.

### DIFF
--- a/dev-tools/travis/travis-script.sh
+++ b/dev-tools/travis/travis-script.sh
@@ -31,7 +31,14 @@ else
   # Travis only has 3GB of memory, lets use 1GB for build, and 1.5GB for forked JVMs
   export MAVEN_OPTS="-Xmx1024m"
   
-  mvn --batch-mode test -fae -Pnative,all-tests '-P!include-shaded-deps' -Prat -pl "$2"
+  TEST_MODULES=$2
+  
+  if [ "$TRAVIS_JDK_VERSION" == "oraclejdk10" ] && [ "$TEST_MODULES" == "'!storm-client,!storm-server,!storm-core'" ]
+  then 
+    TEST_MODULES = '!storm-client,!storm-server,!storm-core,!external/storm-cassandra,!external/storm-hive'
+  fi
+  
+  mvn --batch-mode test -fae -Pnative,all-tests '-P!include-shaded-deps' -Prat -pl "$TEST_MODULES"
   BUILD_RET_VAL=$?
   
   for dir in `find . -type d -and -wholename \*/target/\*-reports`;
--- a/dev-tools/travis/travis-script.sh
+++ b/dev-tools/travis/travis-script.sh
@@ -16,37 +16,49 @@ echo "Ruby version   :  " `ruby -v`
 echo "NodeJs version :  " `node -v`
 echo "Maven version  :  " `mvn -v`
 
+set -x
+
 STORM_SRC_ROOT_DIR=$1
 
 TRAVIS_SCRIPT_DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
 
 cd ${STORM_SRC_ROOT_DIR}
-  
-if [ "$2" == "INTEGRATION-TEST" ]
+
+if [ "$2" == "Integration-Test" ]
+  then
+  exec ./integration-test/run-it.sh
+elif [ "$2" == "Client" ]
 then
-  /bin/bash ./integration-test/run-it.sh
-else
-  # We should be concerned that Travis CI could be very slow because it uses VM
-  export STORM_TEST_TIMEOUT_MS=150000
-  # Travis only has 3GB of memory, lets use 1GB for build, and 1.5GB for forked JVMs
-  export MAVEN_OPTS="-Xmx1024m"
-  
-  TEST_MODULES=$2
-  
-  if [ "$TRAVIS_JDK_VERSION" == "oraclejdk10" ] && [ "$TEST_MODULES" == "'!storm-client,!storm-server,!storm-core'" ]
+  TEST_MODULES=storm-client
+elif [ "$2" == "Server" ]
+then
+  TEST_MODULES=storm-server
+elif [ "$2" == "Core" ]
+then
+  TEST_MODULES=storm-core
+elif [ "$2" == "External" ]
+then
+  if [ "$TRAVIS_JDK_VERSION" == "oraclejdk10" ]
   then 
-    TEST_MODULES = '!storm-client,!storm-server,!storm-core,!external/storm-cassandra,!external/storm-hive'
+    TEST_MODULES='!storm-client,!storm-server,!storm-core,!external/storm-cassandra,!external/storm-hive'
+  else
+    TEST_MODULES='!storm-client,!storm-server,!storm-core'
   fi
-  
-  mvn --batch-mode test -fae -Pnative,all-tests '-P!include-shaded-deps' -Prat -pl "$TEST_MODULES"
-  BUILD_RET_VAL=$?
-  
-  for dir in `find . -type d -and -wholename \*/target/\*-reports`;
-  do
-    echo "Looking for errors in ${dir}"
-    python ${TRAVIS_SCRIPT_DIR}/print-errors-from-test-reports.py "${dir}"
-  done
-  
-  exit ${BUILD_RET_VAL}
+fi
+# We should be concerned that Travis CI could be very slow because it uses VM
+export STORM_TEST_TIMEOUT_MS=150000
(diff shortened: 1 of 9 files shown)


HeartSaVioR:
> The metrics-tests were doing assertions in a racy way. We can't guarantee that metrics fall in exact buckets, because the metrics are updated from the executor async loops, which aren't included in the threads covered by LocalCluster.waitForIdle. I've replaced the checks with validation that there are a certain number of metrics buckets, and that the metrics counters sum to the expected value.

If it is not relevant to JDK 10, it would be better to split this PR up to have individual fix for flaky test. While someone wonders why we consider about Java 10 at all, the fix for flaky test is straightforward to be reviewed and merged sooner.

srdo:
Fair enough, will split this up.

srdo:
Pulled metrics test changes and Time changes into https://github.com/apache/storm/pull/2735 and https://github.com/apache/storm/pull/2734. Will rebase this once those get merged.

dev-tools/travis/travis-script.sh
```diff
@@ -31,7 +31,14 @@ else
   # Travis only has 3GB of memory, lets use 1GB for build, and 1.5GB for forked JVMs
   export MAVEN_OPTS="-Xmx1024m"
   
-  mvn --batch-mode test -fae -Pnative,all-tests '-P!include-shaded-deps' -Prat -pl "$2"
+  TEST_MODULES=$2
+  
+  if [ "$TRAVIS_JDK_VERSION" == "oraclejdk10" ] && [ "$TEST_MODULES" == "'!storm-client,!storm-server,!storm-core'" ]
+  then 
+    TEST_MODULES = '!storm-client,!storm-server,!storm-core,!external/storm-cassandra,!external/storm-hive'
```
revans2: We have overloaded TEST_MODULES from being a pass-through to maven already for integration tests, lets just bite the bullet and turn it into an enum instead, so it is not as unexpected.

storm-client/src/jvm/org/apache/storm/utils/Time.java
```diff
@@ -185,8 +173,17 @@ public static void advanceTimeNanos(long nanos) {
...
+            long newTime = SIMULATED_CURR_TIME_NANOS.addAndGet(nanos);
+            Iterator<AtomicLong> sleepTimesIter = THREAD_SLEEP_TIMES_NANOS.values().iterator();
+            while (sleepTimesIter.hasNext()) {
+                AtomicLong curr = sleepTimesIter.next();
+                if (SIMULATED_CURR_TIME_NANOS.get() >= curr.get()) {
+                    sleepTimesIter.remove();
+                }
+            }
+            LOG.debug("Advanced simulated time to {}", newTime);
+        }
```
revans2: nit: I don't think this change is needed.  every time we add in a thread to the THREAD_SLEEP_TIMES_NANOS, there is a finally to remove it at the end.

storm-core/pom.xml
```diff
@@ -184,6 +184,10 @@
             <artifactId>mockito-core</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.awaitility</groupId>
+            <artifactId>awaitility</artifactId>
```
revans2: nit: can we make this test here too, just so it is clear that it is a test only API?

storm-core/test/resources/log4j2-test.xml
```diff
@@ -23,7 +23,8 @@
         </Console>
     </Appenders>
     <Loggers>
-        <Logger name="org.apache.zookeeper" level="WARN"/>
+        <Logger name="org.apache.storm.shade.org.apache.zookeeper" level="WARN"/>
+        <Logger name="org.apache.storm.shade.org.apache.curator" level="WARN"/>
```
revans2: Thanks for finding/fixing this.

dev-tools/travis/travis-script.sh
srdo: Could you give an example what this would look like? Googling "travis enum" didn't pop up anything.

storm-client/src/jvm/org/apache/storm/utils/Time.java
srdo: Responded in the other PR

storm-core/pom.xml
srdo: Responded in the other PR

storm-core/test/resources/log4j2-test.xml
srdo: It got pretty unreadable :) I'll take a look whether we need to do this in other modules as well.

revans2:
Just merged in your other PR, you might want to rebase to resolve the conflict.

srdo:
@revans2 Rebased

dev-tools/travis/travis-script.sh
revans2: Sorry I wasn't clear enough.  In `.travis.yaml` we have a set of MODULES that we build in parallel.  Currently these are passed unchanged to this script.  For most of them they are maven modules that just get passed directly to mvn, but in the case of INTEGRATION-TEST it has nothing to do with the maven modules and now with this change we are actually modifying the modules.  Because of this I would prefer to see us put in a layer of abstraction so `.travis.yaml` looks more like

```
env:
  - MODULES=CLIENT
  - MODULES=SERVER
  - MODULES=CORE
  - MODULES=EXTERNAL
  - MODULES=INTEGRATION-TEST
```

where the mapping is totally in travis-script.sh

revans2:
I assume that you got the integration tests to pass in travis under JDK10?  I ran them manually and got a failure too.

srdo:
@revans2 Yes, they run for me locally and on https://travis-ci.org/srdo/storm/builds/397471375. I'll try to figure out why the test is flaky, but it might take a while.

revans2:
I would almost rather disable the integration test for java 10 until we can make it work cleanly.

srdo:
I'll try running it a bunch of times, see if I can figure out why it's failing.

PR #2734, opened 2018-06-24
STORM-3120: Clean up leftover null checks in Time, ensure idle thread…

srdo:
…s get to run when cluster time is advanced

https://issues.apache.org/jira/browse/STORM-3120

Some of the Time code didn't make sense, e.g. checking for null on the final THREAD_SLEEP_TIMES_NANOS map. It was left over from an earlier refactor of Time. Cleaned it up, and deleted the deprecated Time methods. Also made sure that when simulated time is advanced, idle threads are removed from the THREAD_SLEEP_TIMES_NANOS immediately. When the LocalCluster waits for the topology to be idle, it checks whether all timer threads are in the THREAD_SLEEP_TIMES_NANOS map. It looks to me like there was no guarantee that sleeping threads actually got a chance to run when time was advanced, because they remained in the THREAD_SLEEP_TIMES_NANOS map until they happened to be scheduled. With bad luck, a thread might end up being counted as idle when it just hadn't exited the Time.sleepUntilNanos loop yet.

### DIFF
--- a/storm-client/src/jvm/org/apache/storm/utils/Time.java
+++ b/storm-client/src/jvm/org/apache/storm/utils/Time.java
@@ -12,10 +12,13 @@
 
 package org.apache.storm.utils;
 
+import java.util.Iterator;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.locks.LockSupport;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -30,26 +33,7 @@ public class Time {
     private static final AtomicLong AUTO_ADVANCE_NANOS_ON_SLEEP = new AtomicLong(0);
     private static final Map<Thread, AtomicLong> THREAD_SLEEP_TIMES_NANOS = new ConcurrentHashMap<>();
     private static final Object SLEEP_TIMES_LOCK = new Object();
-    private static final AtomicLong SIMULATED_CURR_TIME_NANOS = new AtomicLong(0);
-
-    @Deprecated
-    public static void startSimulating() {
-        synchronized (Time.SLEEP_TIMES_LOCK) {
-            Time.SIMULATING.set(true);
-            Time.SIMULATED_CURR_TIME_NANOS.set(0);
-            Time.THREAD_SLEEP_TIMES_NANOS.clear();
-            Time.AUTO_ADVANCE_NANOS_ON_SLEEP.set(0);
-            LOG.warn("Simulated Time Starting...");
-        }
-    }
-
-    @Deprecated
-    public static void stopSimulating() {
-        synchronized (Time.SLEEP_TIMES_LOCK) {
-            Time.SIMULATING.set(false);
-            LOG.warn("Simulated Time Ending...");
-        }
-    }
+    private static final AtomicLong SIMULATED_CURR_TIME_NANOS = new AtomicLong(0); 
 
     public static boolean isSimulating() {
         return SIMULATING.get();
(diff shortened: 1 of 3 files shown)


storm-server/src/main/java/org/apache/storm/LocalCluster.java
```diff
@@ -196,6 +196,9 @@ private LocalCluster(Builder builder) throws Exception {
             this.tmpDirs.add(nimbusTmp);
             stormHomeBackup = System.getProperty(ConfigUtils.STORM_HOME);
             TmpPath stormHome = new TmpPath();
+            if (!stormHome.getFile().mkdirs()) {
```
HeartSaVioR: I might be going to too strict, but this looks like follow-up of STORM-3116 (not relevant to Time) which would be better to have individual PR. The change is minor so no need to file a new issue. Just would like to avoid things getting mixed up in a commit.

Will continue reviewing other change in PR.

srdo: Yes, it is. Do you want me to put this into a STORM-3116 PR, or what do you mean?

HeartSaVioR: I hope to see this being separated PR, tagging as STORM-3116 or just tagging as MINOR. Either is fine.

srdo: Moved to https://github.com/apache/storm/pull/2736

storm-client/src/jvm/org/apache/storm/utils/Time.java
```diff
@@ -82,45 +66,49 @@ public static void sleepUntilNanos(long targetTimeNanos) throws InterruptedExcep
...
-            }
+            THREAD_SLEEP_TIMES_NANOS.remove(Thread.currentThread());
         }
     }
-
+    
     public static void sleep(long ms) throws InterruptedException {
         if (ms > 0) {
-            sleepUntil(currentTimeMillis() + ms);
+            if (SIMULATING.get()) {
```
HeartSaVioR: Is there any reason to not use `sleepUntil`? If you think `sleepUntil` is broken, the method must be fixed since the method is exposed to `public`. If not, why not using `sleepUntil`?

storm-client/src/jvm/org/apache/storm/utils/Time.java
```diff
@@ -82,45 +66,49 @@ public static void sleepUntilNanos(long targetTimeNanos) throws InterruptedExcep
...
+                Thread.sleep(ms);
+            }
         }
     }
 
-    public static void sleepNanos(long nanos) throws InterruptedException {
+    public static void parkNanos(long nanos) throws InterruptedException {
         if (nanos > 0) {
-            sleepUntilNanos(nanoTime() + nanos);
+            if (SIMULATING.get()) {
```
HeartSaVioR: Same here, and we may note that we're renaming public method, though Time class is more likely internal class though.

HeartSaVioR:
Looks good to me overall except two comments above. (I forgot to comment...)

storm-client/src/jvm/org/apache/storm/utils/Time.java
srdo: `sleepUntil` works fine. It's just a little silly in the case where we aren't simulating, because this method would calculate `currentTimeMillis() + ms`, and `sleepUntil` would then do the reverse to calculate how many ms to pass to `Thread.sleep`.

storm-client/src/jvm/org/apache/storm/utils/Time.java
srdo: In this case, I'd like to use `LockSupport.parkNanos` if we're not simulating, rather than sleeping for the closest number of milliseconds as the method did before.

srdo: Whoops, forgot to respond to the renaming bit: Yes, the method is renamed. If we want to port this to 1.x, we can easily deprecate sleepNanos and provide both methods.

storm-client/src/jvm/org/apache/storm/utils/Time.java
```diff
@@ -185,8 +173,17 @@ public static void advanceTimeNanos(long nanos) {
...
+            long newTime = SIMULATED_CURR_TIME_NANOS.addAndGet(nanos);
+            Iterator<AtomicLong> sleepTimesIter = THREAD_SLEEP_TIMES_NANOS.values().iterator();
+            while (sleepTimesIter.hasNext()) {
+                AtomicLong curr = sleepTimesIter.next();
+                if (SIMULATED_CURR_TIME_NANOS.get() >= curr.get()) {
+                    sleepTimesIter.remove();
+                }
+            }
+            LOG.debug("Advanced simulated time to {}", newTime);
+        }
```
revans2: nit: I don't think this code change is needed, as you fixed the bug with checking for THREAD_SLEEP_TIMES_NANOS == null, so the finally should work.  If it does not work, then java itself has some bugs in it.

srdo: That's not why this change is added. I want to make sure that when a test does `LocalCluster.advanceClusterTime`, all sleeping threads that are supposed to wake up actually get a chance to run before `waitForIdle` is checked again. The issue is that if we don't remove the thread from the map here, there isn't (as far as I can tell) a guarantee that the sleeping threads actually woke up.

For example, if a test has a loop
```
while (some condition)
  some code 
  LocalCluster.advanceClusterTime
```
the first call to advanceClusterTime will wait until all threads are sleeping. If thread 1 is told to sleep for 1ms, it will enter the THREAD_SLEEP_TIMES_NANOS map. On the next iteration, `LocalCluster.advanceClusterTime` will increment the time, which should cause thread 1 to wake up. If the testing thread is "too fast", it might recheck `waitForIdle` and catch thread 1 before it manages to wake up and remove itself from THREAD_SLEEP_TIMES_NANOS. This causes LocalCluster.advanceClusterTime to return before the cluster is actually idle, because thread 1 will wake up and start running again after advanceClusterTime returns.

The effect is you can't be sure that all threads are actually idle once LocalCluster.advanceClusterTime returns.

revans2: OK +1

PR #2736, opened 2018-06-24
MINOR: Mkdir for storm.home when running LocalCluster

srdo:
Turns out Rocksdb doesn't handle creating parent directories, so https://github.com/apache/storm/pull/2728 doesn't prevent it from throwing an exception when Nimbus is started.

That said, we're still getting an exception when Nimbus is started from https://github.com/apache/storm/blob/4137328b75c06771f84414c3c2113e2d1c757c08/storm-server/src/main/java/org/apache/storm/metricstore/rocksdb/StringMetadataCache.java#L64, so this doesn't by itself prevent the metrics store from failing to initialize in tests.


HeartSaVioR:
@srdo Thanks for the patch. Merged into master.

PR #2737, opened 2018-06-24
(1.x) STORM-3122 Avoid supervisor being crashed due to race condition between "async localizer" and "update blob" timer thread

HeartSaVioR:
There's race condition between "async localizer" and "update blob" timer thread.

When worker is shutting down, reference count for blob will be 0 and supervisor will remove actual blob file. There's also "update blob" timer thread which tries to keep blobs updated for downloaded topologies. While updating topology it should read some of blob files already downloaded assuming these files should be downloaded before, and the assumption is broken because of async localizer.

@arunmahadevan suggested an approach to fix this: "updateBlobsForTopology" can just catch the FIleNotFoundException and skip updating the blobs in case it can't find the stormconf, and the approach looks simplest fix so I'll provide a patch based on suggestion.

Btw, it doesn't apply to master branch, since in master branch all blobs are synced up separately (no need to read stormconf to enumerate topology related blobs), and update logic is already fault-tolerance (skip to next sync when it can't pull the blob).

### DIFF
--- a/storm-core/src/jvm/org/apache/storm/daemon/supervisor/timer/UpdateBlobs.java
+++ b/storm-core/src/jvm/org/apache/storm/daemon/supervisor/timer/UpdateBlobs.java
@@ -17,6 +17,7 @@
  */
 package org.apache.storm.daemon.supervisor.timer;
 
+import java.io.FileNotFoundException;
 import java.util.HashMap;
 import org.apache.storm.Config;
 import org.apache.storm.daemon.supervisor.Supervisor;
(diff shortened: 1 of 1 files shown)


HeartSaVioR:
Travis build is failing but I know the Travis build for 1.x version line has been unstable for a long time.
Will track the build with my fork: https://travis-ci.org/HeartSaVioR/storm/builds/396178467

HeartSaVioR:
`storm-core` tests are passing in above link. `!storm-core` test for JDK 7 is failing due to known issue (HDFS).

danny0405:
@HeartSaVioR 
Seems to related to this PR https://github.com/apache/storm/pull/2618.

HeartSaVioR:
@danny0405 
Yeah I haven't had time to go through #2618 since the code change is quite huge and async localizer is complicated. While this patch can look like an workaround to others, this patch is quite simple to avoid the issue, since we expect such race condition will not happen in the next update.

arunmahadevan:
+1, LGTM

PR #2738, opened 2018-06-27
 Storm 1311 Migration of UI from Clojure to Jersey/Grizzly I

govind-menon:
This PR contains the basic frameworks and the GET requests. PUT's and storm.py changes to follow


revans2:
Can I suggest perhaps a different path forward?  instead of creating the jersey/jackson server part first, instead start migrating the ui implementation to java first while writing some unit tests for them, and then put in the final part that replaces the routing.

i.e.

replace `cluster-summary` with `UIHelpers.getClusterInfo`, write unit tests for `UIHelpers.getClusterInfo` and then verify that the UI works as expected.

Then try replacing `json-resonse` everywhere in `core.clj` with something new in `UIHelpers`...

govind-menon:
PR is missing a lot of changes - will put up again in some time.

PR #2739, opened 2018-06-27
[STORM-3125/3126/3127]: Refactoring components for Supervisor and DRPC

zd-project:
This is a supplement issue page to STORM-3099 (PR #2710), separating out refactoring work from newly implemented metrics.

A few misc bugs discovered during refactoring have been incorporated in this PR as well, including STORM-3126/3127.

### DIFF
--- a/storm-server/src/main/java/org/apache/storm/daemon/supervisor/Slot.java
+++ b/storm-server/src/main/java/org/apache/storm/daemon/supervisor/Slot.java
@@ -45,6 +45,7 @@ import org.apache.storm.metric.StormMetricsRegistry;
 import org.apache.storm.metricstore.WorkerMetricsProcessor;
 import org.apache.storm.scheduler.ISupervisor;
 import org.apache.storm.shade.com.google.common.annotations.VisibleForTesting;
+import org.apache.storm.utils.EnumUtil;
 import org.apache.storm.utils.LocalState;
 import org.apache.storm.utils.ObjectReader;
 import org.apache.storm.utils.Time;
@@ -56,21 +57,25 @@ public class Slot extends Thread implements AutoCloseable, BlobChangingCallback 
     private static final Logger LOG = LoggerFactory.getLogger(Slot.class);
     private static final Meter numWorkersLaunched =
         StormMetricsRegistry.registerMeter("supervisor:num-workers-launched");
-    private static final Meter numWorkersKilledProcessExit =
-        StormMetricsRegistry.registerMeter("supervisor:num-workers-killed-process-exit");
-    private static final Meter numWorkersKilledMemoryViolation =
-        StormMetricsRegistry.registerMeter("supervisor:num-workers-killed-memory-violation");
-    private static final Meter numWorkersKilledHBTimeout =
-        StormMetricsRegistry.registerMeter("supervisor:num-workers-killed-hb-timeout");
-    private static final Meter numWorkersKilledHBNull =
-        StormMetricsRegistry.registerMeter("supervisor:num-workers-killed-hb-null");
+
+    private enum KillReason {
+        ASSIGNMENT_CHANGED, BLOB_CHANGED, PROCESS_EXIT, MEMORY_VIOLATION, HB_TIMEOUT, HB_NULL;
+
+        @Override
+        public String toString() {
+            return EnumUtil.toMetricName(this);
+        }
+
+    }
+
+    private static final Map<KillReason, Meter> numWorkersKilledFor = EnumUtil.toEnumMap(KillReason.class,
+        killReason -> StormMetricsRegistry.registerMeter("supervisor:num-workers-killed-" + killReason.toString()));
     private static final Meter numForceKill =
         StormMetricsRegistry.registerMeter("supervisor:num-workers-force-kill");
     private static final long ONE_SEC_IN_NANO = TimeUnit.NANOSECONDS.convert(1, TimeUnit.SECONDS);
     private final AtomicReference<LocalAssignment> newAssignment = new AtomicReference<>();
     private final AtomicReference<Set<TopoProfileAction>> profiling = new AtomicReference<>(new HashSet<>());
 
-    ;
     private final BlockingQueue<BlobChanging> changingBlobs = new LinkedBlockingQueue<>();
     private final StaticState staticState;
     private final IStormClusterState clusterState;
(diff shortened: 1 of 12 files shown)


storm-server/src/main/java/org/apache/storm/daemon/drpc/DRPC.java
```diff
@@ -208,6 +208,7 @@ public void failRequest(String id, DRPCExecutionException e) throws Authorizatio
         }
     }
 
+    //Should this be changed to private?
```
revans2: If no one is calling it from outside go ahead.

storm-server/src/main/java/org/apache/storm/daemon/supervisor/Slot.java
```diff
@@ -89,54 +94,56 @@ public Slot(AsyncLocalizer localizer, Map<String, Object> conf,
...
+                    } else {
+                        // Make the assignment null to let slot clean up the disk assignment.
+                        currentAssignment = null;
+                    }
+                } catch (ContainerRecoveryException e) {
+                    //We could not recover container will be null.
+                }
+
+                newAssignment = currentAssignment;
+                if (/* currentAssignment != null && */container == null) {
```
revans2: nit: can we remove the comment from in here.  I think it detracts from the readability.

storm-server/src/main/java/org/apache/storm/daemon/supervisor/Slot.java
```diff
@@ -241,64 +255,54 @@ static DynamicState prepareForNewAssignmentNoWorkersRunning(DynamicState dynamic
...
-        staticState.iSupervisor.killedWorker(staticState.port);
-        dynamicState.container.kill();
-        Future<Void> pendingDownload = null;
-        if (dynamicState.newAssignment != null) {
-            pendingDownload = staticState.localizer.requestDownloadTopologyBlobs(dynamicState.newAssignment, staticState.port,
-                                                                                 staticState.changingCallback);
+        //Skip special case if `storm kill_workers` is already invoked
+        Boolean isDead = dynamicState.container.areAllProcessesDead();
+        if (!isDead) {
+            //TODO: Is this necessary?
```
revans2: We probably want to inform iSupervisor of the kill each time.

storm-server/src/main/java/org/apache/storm/daemon/supervisor/Slot.java
```diff
@@ -241,64 +255,54 @@ static DynamicState prepareForNewAssignmentNoWorkersRunning(DynamicState dynamic
...
+                next = dynamicState.withState(MachineState.KILL)
+                        .withPendingLocalization(dynamicState.newAssignment, pendingDownload);
+                break;
+
+            case BLOB_CHANGED:
+                next = dynamicState.withState(MachineState.KILL_BLOB_UPDATE);
+                break;
+
+            case PROCESS_EXIT: case MEMORY_VIOLATION:
+            case HB_TIMEOUT: case HB_NULL:
```
revans2: nit: can we put each of these on their own line.  It just feels cleaner to me that way.

storm-server/src/main/java/org/apache/storm/daemon/supervisor/Slot.java
```diff
@@ -397,22 +403,26 @@ private static DynamicState filterChangingBlobsFor(DynamicState dynamicState, fi
     }
 
     /**
-     * State Transitions for WAITING_FOR_BLOB_LOCALIZATION state.
+     * State Transitions for WAITING_FOR_BLOB_LOCALIZATION state. This is when slot is waiting for
+     * the blob to download before the container is launched/relaunched for the first time for an assignment.
+     * for the first time and
```
revans2: nit: this small fragment after the end of the sentence does not totally make since.

storm-server/src/main/java/org/apache/storm/daemon/supervisor/Slot.java
```diff
@@ -826,6 +835,10 @@ public void addProfilerActions(Set<TopoProfileAction> actions) {
         }
     }
 
+    /**
+     * get the workerID (nullable) from CURRENT container, if existed, or return null
```
revans2: Nit: need a '.' at the end of the end.

storm-server/src/main/java/org/apache/storm/localizer/LocalizedResource.java
```diff
@@ -85,26 +83,26 @@
     private final Path baseDir;
     private final Path versionFilePath;
     private final Path symlinkPath;
-    private final boolean uncompressed;
+    private final boolean shouldUnzip;
```
revans2: nit can we call this `shouldUncompress` or `shouldExpandArchive`?  We support more than just zip.  And this goes everywhere that we have Unzip.

storm-server/src/main/java/org/apache/storm/localizer/LocallyCachedBlob.java
```diff
@@ -59,57 +59,47 @@ protected LocallyCachedBlob(String blobDescription, String blobKey) {
...
+    protected DownloadMeta fetch(ClientBlobStore store, String key,
+                                 IOFunction<Long, Path> pathSupplier,
+                                 IOFunction<File, OutputStream> outStreamSupplier)
+            throws KeyNotFoundException, AuthorizationException, IOException {
+
         try (InputStreamWithMeta in = store.getBlob(key)) {
             long newVersion = in.getVersion();
-            if (newVersion == currentVersion) {
-                LOG.warn("The version did not change, but going to download again {} {}", currentVersion, key);
+            if (newVersion == getLocalVersion()) {
```
revans2: nit: getLocalVersion may not always be cheap.  It would be good to cache the result instead of calling it twice.

storm-server/src/main/java/org/apache/storm/localizer/LocallyCachedBlob.java
```diff
@@ -162,6 +152,31 @@ protected static long getSizeOnDisk(Path p) throws IOException {
...
+            return Files.size(p);
+        } else {
+            //We will not follow sym links
+            return Files.walk(p)
+                    .filter((subp) -> Files.isRegularFile(subp, LinkOption.NOFOLLOW_LINKS))
+                    .mapToLong((subp) -> {
+                        try {
+                            return Files.size(subp);
+                        } catch (IOException e) {
+                            LOG.warn("Could not get the size of ");
```
revans2: This does not look like it is complete.  Could not get the size of what?

storm-server/src/main/java/org/apache/storm/utils/EnumUtil.java
```diff
@@ -0,0 +1,40 @@
...
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+package org.apache.storm.utils;
+
+import java.util.EnumMap;
+import java.util.function.Function;
+
+public class EnumUtil {
+    public static String toString(Enum type) {
```
revans2: Could you put some documentation on this.  It was not totally clear why this was being called, but now it looks like it is for the metrics name

storm-server/src/main/java/org/apache/storm/localizer/LocalizedResource.java
zd-project: If no one disagrees. I was not specific about compression method when I thought about renaming but I see why it might be misleading.

storm-server/src/main/java/org/apache/storm/localizer/LocallyCachedBlob.java
zd-project: I merely moved this code up according to the checkstyle. I didn't make any change to the code itself.

storm-server/src/main/java/org/apache/storm/utils/EnumUtil.java
zd-project: Yes it's a helper for grouped metrics for now. Would you suggest a better naming?

storm-server/src/main/java/org/apache/storm/daemon/supervisor/Slot.java
zd-project: I noticed that comments in the iSupervisor said the purpose of `killedWorker()` is actually to send a "task finished" update. If this is the case then we shouldn't inform iSupervisor when task halts abnormally.

storm-server/src/main/java/org/apache/storm/daemon/drpc/DRPC.java
zd-project: It is not called publicly, but this looks like an asynchronous method which may be used later.

storm-server/src/main/java/org/apache/storm/localizer/LocallyCachedBlob.java
```diff
@@ -59,57 +59,47 @@ protected LocallyCachedBlob(String blobDescription, String blobKey) {
...
-                while ((read = in.read(buffer)) > 0) {
-                    outStream.write(buffer, 0, read);
+                int read;
+                while ((read = in.read(buffer)) >= 0) {
+                    out.write(buffer, 0, read);
                     totalRead += read;
                 }
             }
             long expectedSize = in.getFileLength();
             if (totalRead != expectedSize) {
```
zd-project: This was used only by locallyCachedTopologyBlob previously. I was wondering why localizedResources doesn't care about checking downloaded file size.

storm-server/src/main/java/org/apache/storm/daemon/drpc/DRPC.java
revans2: If we need it later we can make it public then

storm-server/src/main/java/org/apache/storm/daemon/supervisor/Slot.java
revans2: OK

storm-server/src/main/java/org/apache/storm/localizer/LocallyCachedBlob.java
revans2: The code change is fine, I just don't want to call the API twice if we don't have to.

storm-server/src/main/java/org/apache/storm/localizer/LocallyCachedBlob.java
revans2: That may be, but if you could fix it that would be great.

storm-server/src/main/java/org/apache/storm/utils/EnumUtil.java
revans2: You might call it toMetricName instead.

revans2:
All my manual tests passed looks good +1.  Would like to see the nits addressed, but they are very minor.

storm-server/src/main/java/org/apache/storm/daemon/supervisor/Slot.java
zd-project: Okay. I just want to make sure that this is indeed the desired behavior.

storm-server/src/main/java/org/apache/storm/localizer/LocallyCachedBlob.java
zd-project: I agree with the caching, but could this be a premature optimization?

storm-server/src/main/java/org/apache/storm/localizer/LocallyCachedBlob.java
revans2: It was mostly about where we were seeing errors at the time.

zd-project:
Should I also rebase it onto current master branch?

HeartSaVioR:
@zd-project No it is good to merge. Just merged.

PR #2742, opened 2018-06-28
STORM-3061: Update commons libraries

revans2:


srdo:
+1 (relying on the tests here)

PR #2743, opened 2018-06-28
[STORM-3130]: Add Wrappers for Timer registration and timed object

zd-project:
The idea was originally conceived because a few Storm server metrics were requested on timing procedure's efficiency (e.g., how long does it take for a file/jar to be uploaded, how long does it takes to actually launch an assignment in the worker). Since many operations are processed asynchronously, there's no easy way to track the timing. However I noticed that in many cases, even the operation itself is convoluted, the object on which the operation applied is likely to be passed along all method invocation, which gives us a way to keep track the time of a procedure by observing the lifespan of the targeted object.

I've tried to attach the timer to existing object through decorator or through inheritance, and they seem both to be working in certain situations. But to me it's still not ideal to have different patterns that implements the same functionality. I'd like to come up with a generic approach this kind of problem so it'd be easier to track time this way. I'd appreciate if you can have any suggestions on how I should tackle this problem.

### DIFF
--- a/storm-server/src/main/java/org/apache/storm/daemon/nimbus/TimedWritableByteChannel.java
+++ b/storm-server/src/main/java/org/apache/storm/daemon/nimbus/TimedWritableByteChannel.java
@@ -1,0 +1,47 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license agreements.
+ * See the NOTICE file distributed with this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+package org.apache.storm.daemon.nimbus;
+
+import com.codahale.metrics.Timer;
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.nio.channels.WritableByteChannel;
+
+import org.apache.storm.metric.timed.TimedResource;
+
+public class TimedWritableByteChannel extends TimedResource<WritableByteChannel> implements WritableByteChannel {
+    public TimedWritableByteChannel(WritableByteChannel measured, Timer timer) {
+        super(measured, timer);
+    }
+
+    @Override
+    public int write(ByteBuffer src) throws IOException {
+        return getMeasured().write(src);
+    }
+
+    @Override
+    public boolean isOpen() {
+        return getMeasured().isOpen();
+    }
+
+    @Override
+    public void close() throws IOException {
+        try {
+            super.close();
+        } catch (Exception e) {
+            throw (IOException) e;
+        }
+    }
+}
--- a/storm-server/src/main/java/org/apache/storm/daemon/nimbus/TimedWritableByteChannel.java
+++ b/storm-server/src/main/java/org/apache/storm/daemon/nimbus/TimedWritableByteChannel.java
@@ -22,6 +22,7 @@ import java.nio.channels.WritableByteChannel;
 import org.apache.storm.metric.timed.TimedResource;
 
 public class TimedWritableByteChannel extends TimedResource<WritableByteChannel> implements WritableByteChannel {
+
     public TimedWritableByteChannel(WritableByteChannel measured, Timer timer) {
         super(measured, timer);
     }
--- a/storm-server/src/main/java/org/apache/storm/daemon/nimbus/TimedWritableByteChannel.java
+++ b/storm-server/src/main/java/org/apache/storm/daemon/nimbus/TimedWritableByteChannel.java
@@ -42,6 +42,9 @@ public class TimedWritableByteChannel extends TimedResource<WritableByteChannel>
         try {
             super.close();
         } catch (Exception e) {
+            //WritableByteChannel is a Channel which implements Closeable.
+            // Hence although declared AutoCloseable super#close here should only throws IOException
+            //We rethrow to conform the signature
             throw (IOException) e;
         }
     }
--- a/storm-server/src/main/java/org/apache/storm/daemon/supervisor/TimerDecoratedAssignment.java
+++ b/storm-server/src/main/java/org/apache/storm/daemon/supervisor/TimerDecoratedAssignment.java
@@ -15,29 +15,20 @@
 package org.apache.storm.daemon.supervisor;
 
 import com.codahale.metrics.Timer;
-import java.util.List;
-import java.util.concurrent.atomic.AtomicReference;
 
-import org.apache.storm.generated.ExecutorInfo;
 import org.apache.storm.generated.LocalAssignment;
 import org.apache.storm.metric.timed.TimerDecorated;
 
 public class TimerDecoratedAssignment extends LocalAssignment implements TimerDecorated {
-    //TODO: Does this have to volatile?
-    private final AtomicReference<Timer.Context> timingRef;
+    private final Timer.Context timing;
 
     public TimerDecoratedAssignment(LocalAssignment other, Timer timer) {
         super(other);
-        timingRef = new AtomicReference<>(timer.time());
-    }
-
-    @Override
-    public boolean hasStopped() {
-        return hasStopped(timingRef);
+        timing = timer.time();
     }
 
     @Override
     public long stopTiming() {
-        return stopTiming(timingRef);
+        return stopTiming(timing);
     }
 }
--- a/storm-server/src/main/java/org/apache/storm/daemon/supervisor/TimerDecoratedAssignment.java
+++ b/storm-server/src/main/java/org/apache/storm/daemon/supervisor/TimerDecoratedAssignment.java
@@ -1,0 +1,52 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license agreements.
+ * See the NOTICE file distributed with this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+package org.apache.storm.daemon.supervisor;
+
+import com.codahale.metrics.Timer;
+import java.util.List;
+import java.util.concurrent.atomic.AtomicReference;
+
+import org.apache.storm.generated.ExecutorInfo;
+import org.apache.storm.generated.LocalAssignment;
+import org.apache.storm.metric.timed.TimerDecorated;
+
+public class TimerDecoratedAssignment extends LocalAssignment implements TimerDecorated {
+    //TODO: Does this have to volatile?
+    private final AtomicReference<Timer.Context> timingRef = new AtomicReference<>();
+
+    public TimerDecoratedAssignment(Timer.Context timing) {
+        this.timingRef.set(timing);
+    }
+
+    public TimerDecoratedAssignment(String topology_id, List<ExecutorInfo> executors, Timer.Context timing) {
+        super(topology_id, executors);
+        this.timingRef.set(timing);
+    }
+
+    public TimerDecoratedAssignment(LocalAssignment other, Timer.Context timing) {
+        super(other);
+        this.timingRef.set(timing);
+    }
+
+    @Override
+    public boolean hasStopped() {
+        return hasStopped(timingRef);
+    }
+
+    @Override
+    public long stopTiming() {
+        return stopTiming(timingRef);
+    }
+}
--- a/storm-server/src/main/java/org/apache/storm/daemon/supervisor/TimerDecoratedAssignment.java
+++ b/storm-server/src/main/java/org/apache/storm/daemon/supervisor/TimerDecoratedAssignment.java
@@ -24,20 +24,11 @@ import org.apache.storm.metric.timed.TimerDecorated;
 
 public class TimerDecoratedAssignment extends LocalAssignment implements TimerDecorated {
     //TODO: Does this have to volatile?
-    private final AtomicReference<Timer.Context> timingRef = new AtomicReference<>();
+    private final AtomicReference<Timer.Context> timingRef;
 
-    public TimerDecoratedAssignment(Timer.Context timing) {
-        this.timingRef.set(timing);
-    }
-
-    public TimerDecoratedAssignment(String topology_id, List<ExecutorInfo> executors, Timer.Context timing) {
-        super(topology_id, executors);
-        this.timingRef.set(timing);
-    }
-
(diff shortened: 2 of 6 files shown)


zd-project:
Explanation on abstraction and usage.

```java
//Interface that provides #stopTiming and #hasStopped helper method
public interface TimerDecorated extends AutoCloseable {}

//Subclass of LocalAssignment
public class TimerDecoratedAssignment extends LocalAssignment implements TimerDecorated {}

//Base implementation of Timed decorators
public class Timed<T> implements TimerDecorated {}
public class TimedResource<T extends AutoCloseable> extends Timed<T> {}
public class TimedWritableByteChannel extends TimedResource<WritableByteChannel> implements WritableByteChannel {}
```

- Tracking time on the topology jar upload to Nimbus
	- Originally, Nimbus gets notified of the upload through callback `beginFileUpload()`, `uploadChunk()`, and `finishFileUpload()`, all of which depends on a map from `Path` to `WritableByteChannel`. `WritableByteChannel` is the actual object that handles data I/O, so by implementing `WritableByteChannel` with a decorator `TimedWritableByteChannel` , we're able to track the timing with only one line change in `Nimbus.java` 


```java
public String beginFileUpload() throws AuthorizationException, TException {
        try {
	        ...
            String fileloc = ...;
-	        uploaders.put(fileloc, Channels.newChannel(new FileOutputStream(fileloc)));            
+           uploaders.put(fileloc, new TimedWritableByteChannel(Channels.newChannel(new FileOutputStream(fileloc)), fileUploadDuration));
            return fileloc;
        } catch (Exception e) {
			...;
        }
    }
```

- Tracking time for a worker to launch an assignment
	- A worker slot receives assignment asynchronously and execute based on a State Machine model. We can attach the timer to the assignment itself to minimize the code refactoring. However, because `LocalAssignment` itself is generated from Thrift protocol and does not have a complete interface, it's not ideal to use a decorator. Hence I choose to inherit from `LocalAssignment` instead and declare `TimerDecoratedAssignment` class that implements `TimerDecorated` directly. This way it's less obtrusive to the existing code.

storm-server/src/main/java/org/apache/storm/metric/timed/TimerDecorated.java
```diff
@@ -0,0 +1,50 @@
...
+
+    public long stopTiming();
+
+    /**
+     * Stop the timer for measured object. Should be executed only once.
+     *
+     * @return Time a object is in use, or under measurement, in nanoseconds.
+     * @throws NullPointerException this method is called more than once.
+     */
+    default long stopTiming(final AtomicReference<Timer.Context> timingRef) throws NullPointerException {
```
revans2: You don't have to declare that NullPointerException is thrown by this code.

storm-server/src/main/java/org/apache/storm/metric/timed/TimerDecorated.java
```diff
@@ -0,0 +1,50 @@
...
+        Timer.Context timing = timingRef.get();
+        if (timing != null) {
+            long lapses = timing.stop();
+            timingRef.set(null);
```
revans2: This is potentially a problem, because AtomicReference implies that it is being used from multiple threads.  If that is true, then there is a race here, where you might have multiple threads actually call stop on it at the same time, because the get does not atomically set it to null.

storm-server/src/main/java/org/apache/storm/metric/timed/TimerDecorated.java
```diff
@@ -0,0 +1,50 @@
...
+            return lapses;
+        }
+        throw new NullPointerException("Timer for this object has stopped and cleared");
```
revans2: So the return value, `lapses`, is never used anywhere, and I don't see a lot of value in throwing an NPE for doing a double close of a resource.  I would prefer to have it have an assertion so that when the code is under test we validate that we are not doing a double free, but not really worry about it in production, because it has already been closed.

storm-server/src/main/java/org/apache/storm/metric/timed/TimedResource.java
```diff
@@ -0,0 +1,29 @@
...
+import com.codahale.metrics.Timer;
+
+public class TimedResource<T extends AutoCloseable> extends Timed<T> {
+    public TimedResource(T measured, Timer timer) {
+        super(measured, timer);
+    }
+
+    @Override
+    public void close() throws Exception {
+        super.close();
```
revans2: Could we please change this a bit?  Stopping the timer is not nearly as important as actually closing the resource, and if stopping the timer throws any kind of exception it will not stop the resource.  Could we do something more like.

```
try {
    super.close();
} finally {
    getMeasured().close();
}
```

This should guarantee that we close the resource always, and if an exception is thrown in both cases the resource exception takes precedent and in newer versions of java the other exception will show up as a Suppressed exception.

storm-server/src/main/java/org/apache/storm/metric/timed/Timed.java
```diff
@@ -0,0 +1,44 @@
...
+
+package org.apache.storm.metric.timed;
+
+import com.codahale.metrics.Timer;
+
+import java.util.concurrent.atomic.AtomicReference;
+
+public class Timed<T> implements TimerDecorated {
+    private final T measured;
+    //TODO: Does this have to volatile?
```
revans2: No AtomicReference handles making all operations thread safe.

storm-server/src/main/java/org/apache/storm/daemon/supervisor/TimerDecoratedAssignment.java
```diff
@@ -0,0 +1,52 @@
...
+import com.codahale.metrics.Timer;
+import java.util.List;
+import java.util.concurrent.atomic.AtomicReference;
+
+import org.apache.storm.generated.ExecutorInfo;
+import org.apache.storm.generated.LocalAssignment;
+import org.apache.storm.metric.timed.TimerDecorated;
+
+public class TimerDecoratedAssignment extends LocalAssignment implements TimerDecorated {
+    //TODO: Does this have to volatile?
```
revans2: no

storm-server/src/main/java/org/apache/storm/daemon/supervisor/TimerDecoratedAssignment.java
```diff
@@ -0,0 +1,52 @@
...
+    private final AtomicReference<Timer.Context> timingRef = new AtomicReference<>();
+
+    public TimerDecoratedAssignment(Timer.Context timing) {
+        this.timingRef.set(timing);
+    }
+
+    public TimerDecoratedAssignment(String topology_id, List<ExecutorInfo> executors, Timer.Context timing) {
```
revans2: Variable Naming is CamelCase in java. `topologyId`

storm-server/src/main/java/org/apache/storm/metric/timed/TimerDecorated.java
zd-project: I've just changed so to getAndSet(). In actual case there might not be multiple threads calling it. I still need to investigate a bit.

storm-server/src/main/java/org/apache/storm/metric/timed/TimerDecorated.java
zd-project: It can be used if we need the time somewhere else, say when calculating download rate. But the reason for not closing timer multiple time is that Timer.stop under the hood is not idempotent. So calling it multiple time can actually mess up the metrics.

storm-server/src/main/java/org/apache/storm/metric/timed/Timed.java
zd-project: According to Stack Overflow, AtomicReference behaves almost the same as volatile https://stackoverflow.com/questions/281132/java-volatile-reference-vs-atomicreference
But the reason I'd like to use AtomicReference is to mimic pass-by-reference behavior in C++ so I can use the default method in interface to avoid code duplication.

storm-server/src/main/java/org/apache/storm/metric/timed/TimerDecorated.java
zd-project: Can you explain a bit on when I should declare throwing exceptions?

revans2: The high level type hierarchy for Throwable for this purpose looks like

```
Error extends Throwable
Exception extends Throwable
RuntimeException extends Exception
```

Errors come from the runtime and don't have to be declared.  They typically mean something is wrong with the JVM.

Exceptions are things that you expect to see happen when the code is running and most of these have to be declared when thrown except for RuntimeException and its subclasses.

RuntimeException is a select group of exceptions where someone decided that they happen enough that it would be a real pain to force everyone to declare them all the time.  Null Pointer Dereference and Divide by 0 are examples of these.  For good or bad they don't have to be declared and by convention they don't get declared.

zd-project:
This PR will affect the implementation of #2710

storm-server/src/main/java/org/apache/storm/daemon/nimbus/TimedWritableByteChannel.java
```diff
@@ -0,0 +1,48 @@
...
+    public boolean isOpen() {
+        return getMeasured().isOpen();
+    }
+
+    @Override
+    public void close() throws IOException {
+        try {
+            super.close();
+        } catch (Exception e) {
+            throw (IOException) e;
```
srdo: If you know this is an IOException, you should change the caught type instead.

storm-server/src/main/java/org/apache/storm/metric/timed/TimedResource.java
```diff
@@ -0,0 +1,33 @@
...
+    public TimedResource(T measured, Timer timer) {
+        super(measured, timer);
+    }
+
+    @Override
+    public void close() throws Exception {
+        try {
+            super.close();
+        } finally {
+            getMeasured().close();
```
srdo: Why do we want this in a finally block?

revans2: This was something I suggested.  I was concerned that if `super.close()` could throw an exception then `getMeasured.close()` which is what we are wrapping would never be closed.  This lets us know it will be called no matter what, and if both of them throw an exception (who knows why) than the exception from `getMeasured().close()` will mask the `super.close()` exception, but because of java8 it will store it as a masked exception.

storm-server/src/main/java/org/apache/storm/daemon/supervisor/TimerDecoratedAssignment.java
revans2: Please remove the TODO comment.  All TODOs in the code should either be fixed right now, or have a JIRA filed for them.

storm-server/src/main/java/org/apache/storm/metric/timed/TimerDecorated.java
```diff
@@ -0,0 +1,47 @@
...
+
+public interface TimerDecorated extends AutoCloseable {
+
+    boolean hasStopped();
+
+    default boolean hasStopped(final AtomicReference<Timer.Context> timing) {
+        return timing.get() == null;
+    }
+
+    long stopTiming();
```
revans2: I am a little concerned that we have both close and stopTiming, but if anyone calls stopTiming and then close the assertion will be thrown.  It is like you have a method that you want everyone to implement, but never use.

Could you add some javadocs to this method in particular, but also to the others as well so that developers understand how to use these methods, as it is not the most obvious.  It would also be nice if the javadocs could point to one of the implementations as an example of how to extend this as a cross cutting concern.

zd-project: I'm actually modeling so after the implementation of dropwizard's Timer class. Basically the object can behave like an Autocloseable and put in try-with-resource block or specialized collections (such as TimeCacheMap). In a way, close method is not for users to call and user should either invoke stop timing or rely on `Autocloseable`'s behavior. But yeah now I think for robustness maybe `stopTiming` should be idempotent instead of only a one time call.

zd-project: Also I'm not sure if the current design is best for extensibility. I can think of a scenario when you want to have different timer or track time in different phase (say you want to measure not only how long it takes to launch an assignment but how long it runs on a worker). Should I make it more flexible so it's easier to maintain in the future?

storm-server/src/main/java/org/apache/storm/metric/timed/TimedResource.java
srdo: Makes sense

storm-server/src/main/java/org/apache/storm/daemon/nimbus/TimedWritableByteChannel.java
srdo: The superclass `close` can throw any Exception (e.g. from `getMeasured().close()`), so I don't think you can do this cast safely. You probably need to wrap any non-IOException and rethrow as unchecked.

If you want to avoid this, you would probably need to ensure `Timed<T extends Closeable>` instead of `Autocloseable`, not sure how many classes only implement the latter.

zd-project: getMeasured() is guaranteed the return type of `WritableByteChannel` here I believe, because this class extends `TimedResource<WritableByteChannel>`.

zd-project: getMeasured().close() should be instantiated with `Channel#close()` here I believe, because the class extends `TimedResource<WritableByteChannel>`. `getMeasured` should always return type `WritableByteChannel`

srdo: You're right. Nevermind then. I can't think of a better structure either.

zd-project: I'll comment this to improve it's readability

HeartSaVioR:
@zd-project 
You can raise a pull request on the branch which is on top of this PR's branch (so that new PR will contain all commits from this PR and newer). We can review and eventually this one first, and you can rebase new PR to latest master after merging.

zd-project:
Can anyone, if available, help continue the review process of this PR? I have more than 3 PR depending on this and counting. This makes rebasing a bit hard for me. @HeartSaVioR @revans2 @srdo 

Thanks in advance!

srdo:
As I see it there are still some issues that were commented on but are not resolved.

* The TODOs should be removed/fixed
* Comment here about adding a comment https://github.com/apache/storm/pull/2743/files#r199600941
* Comment here about making stopTiming idempotent https://github.com/apache/storm/pull/2743/files#r199531108

Other than that, this looks pretty good to me.

zd-project:
I didn't make Timed object enforced the idempotent effect of `stopTiming` for now because it's kind of overkill for now. For current implementation the timed object should be discarded soon as timing finished.

srdo:
@zd-project In your last commit, you removed the `AtomicReference` wrapper for the `Timer.Context`. Is it because you expect the `Timer.Context` to only be used by one thread?

zd-project:
stopTiming is only be called once (manually or via `#close()`) so far. I think for extensibility we could have another implementation, say `ConcurrentTimed<T> implements TimerDecorated`. I just don't see the need right now.

zd-project:
You should be able to see the usage in [here](https://github.com/apache/storm/pull/2754/files#diff-9c4945e8e924565ca1f071435f4711e9R3711), [here](https://github.com/apache/storm/pull/2710/commits/73b934420c2fd7ad441d3c0467376871d915b230) and [here](https://github.com/apache/storm/pull/2710/commits/778f30711af35732a5a13e1317dda6df905133c4)

srdo:
I think the changes here are fine, but I'd like to not merge it here, since the added code is all dead code in this PR. It would be better to review and merge one of the PRs that actually use this functionality.

Ethanlm:
merged in https://github.com/apache/storm/pull/2710

zd-project:
Merged in #2710

PR #2744, opened 2018-06-29
[STORM-3132] Avoid NPE in the Values Constructor

kishorvpatil:
`Values` construction could end up throwing NPE.


revans2:
So this changes a `Values(null)` into a `Values()` That is not really what I would expect.  I would prefer to have `Values(null)` insert a `null` into the list. `Values(null, "A")` does not turn into `Values("A")`.

HeartSaVioR:
+1 to @revans2.

kishorvpatil:
@revans2 @HeartSaVioR  Now  constructor allows `null` values.

storm-client/src/jvm/org/apache/storm/tuple/Values.java
```diff
@@ -23,9 +23,13 @@ public Values() {
     }
 
     public Values(Object... vals) {
-        super(vals.length);
-        for (Object o : vals) {
-            add(o);
+        super(vals != null ? vals.length : 0);
+        if (vals != null && vals.length != 0 ) {
```
HeartSaVioR: `vals.length != 0` should be removed in this condition, cause condition for else statement would be `vals == null || vals.length == 0`, which `Values(List())` goes into else statement and comes out of `Values(null)` instead of no element.

kishorvpatil:
@HeartSaVioR Removed unwanted condition.

storm-client/test/jvm/org/apache/storm/tuple/ValuesTest.java
```diff
@@ -0,0 +1,49 @@
...
+    @Test
+    public void testNoArgsToValues() {
+        Values vals = new Values();
+        Assert.assertTrue("Failed to add null to Values", vals.size() == 0);
+    }
+
+    @Test
+    public void testNullArgsToValues() {
+        Values vals = new Values(null);
+        Assert.assertTrue("Failed to add null to Values", vals.size() == 1);
```
HeartSaVioR: I'm sorry to comment this too late, but let's make sure checking elements in values. Same applies to other tests.

storm-client/src/jvm/org/apache/storm/tuple/Values.java
```diff
@@ -23,9 +23,13 @@ public Values() {
     }
 
     public Values(Object... vals) {
-        super(vals.length);
-        for (Object o : vals) {
-            add(o);
+        super(vals != null ? vals.length : 0);
```
Ethanlm: I think this can be `super(vals != null ? vals.length : 1);` since `if (vals==null)` we are going to add `null` to it. The length will be 1.

kishorvpatil: Making change

storm-client/test/jvm/org/apache/storm/tuple/ValuesTest.java
kishorvpatil: @HeartSaVioR, Make changes to unit tests to explicitly check elements in values.

rzo1 (approved):

PR #2745, opened 2018-06-29
STORM-3135: Allow JCQueueTest to retry interrupting the consumer if t…

srdo:
…he queue happens to be full

https://issues.apache.org/jira/browse/STORM-3135

### DIFF
--- a/storm-client/src/jvm/org/apache/storm/daemon/worker/WorkerTransfer.java
+++ b/storm-client/src/jvm/org/apache/storm/daemon/worker/WorkerTransfer.java
@@ -88,9 +88,6 @@ class WorkerTransfer implements JCQueue.Consumer {
 
     @Override
     public void accept(Object tuple) {
-        if (tuple == JCQueue.INTERRUPT) {
-            throw new RuntimeException(new InterruptedException("Worker Transfer Thread interrupted"));
-        }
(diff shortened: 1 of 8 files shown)


srdo:
The failure is unrelated, raised an issue at https://issues.apache.org/jira/browse/STORM-3137 to track.

srdo:
@revans2 Good point. It probably doesn't hurt much in the real code, because the queue is only shut down when the worker or executor is shut down. It could be an issue for tests with LocalCluster though.

Took a different approach. As far as I can tell, there isn't a great reason that we need to interrupt the async loops through the JCQueue. We can just interrupt the threads directly, which is already happening during worker/executor shutdown.

I think the reason to interrupt via the queue is because the wait strategies now include use of LockSupport.parkNanos, so there's no guarantee that the executors will hit a Thread.sleep and throw an InterruptedException. We can fix this by checking manually for interrupts, either in the async loop, or in the wait strategies where we're calling parkNanos.

revans2:
Sorry I don't think I submitted my nit.  We are calling `isInterrupted` in asyncLoop.  Is this going to leave the state of the thread as interrupted after we throw the exception.

srdo:
Yes, the thread is still interrupted after the exception is thrown. I'll change it to use `interrupted` instead.

srdo:
Fixed.

PR #2746, opened 2018-06-29
STORM-3136: Fix flaky integration test, clean up that code to be more…

srdo:
… readable

https://issues.apache.org/jira/browse/STORM-3136

* Deduplicated the test spouts and bolts a bit. 
* Made sure all test spouts are waiting a reasonable amount of time between emits. If the log rolls during the tests, the tests will fail. I think this is the reason for the flakiness. It was fixed for some tests in an earlier PR, but I missed that the spout code had so many duplicates. 
* Added some documentation
* Replaced a bunch of Guava code with streams. Deleted unused code. Simplified some classes, e.g. TimeDataWindow

### DIFF
--- a/integration-test/src/main/java/org/apache/storm/st/topology/window/TimeDataIncrementingSpout.java
+++ b/integration-test/src/main/java/org/apache/storm/st/topology/window/TimeDataIncrementingSpout.java
@@ -1,0 +1,63 @@
+/*
+ * Copyright 2018 The Apache Software Foundation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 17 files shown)


srdo:
Ran this 4 times on top of the changes from https://github.com/apache/storm/pull/2733 on Travis, runs of the integration test passed.

integration-test/src/main/java/org/apache/storm/st/topology/window/TimeDataIncrementingSpout.java
```diff
@@ -0,0 +1,64 @@
...
+import org.apache.storm.task.TopologyContext;
+import org.apache.storm.topology.OutputFieldsDeclarer;
+import org.apache.storm.topology.base.BaseRichSpout;
+import org.apache.storm.tuple.Values;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class TimeDataIncrementingSpout extends BaseRichSpout {
+        private static final Logger LOG = LoggerFactory.getLogger(TimeDataIncrementingSpout.class);
+        private final Random rng = new Random();
```
HeartSaVioR: unused?

srdo: Yes, nice catch

PR #2747, opened 2018-06-29
STORM-3134: Improve upload-creds user experience

revans2:
This also fixes a bug in upload-credentials when it was translated to java by adding in optionalArg to CLI

### DIFF
--- a/storm-core/src/jvm/org/apache/storm/command/AdminCommands.java
+++ b/storm-core/src/jvm/org/apache/storm/command/AdminCommands.java
@@ -30,6 +30,7 @@ import org.apache.storm.cluster.ClusterStateContext;
 import org.apache.storm.cluster.ClusterUtils;
 import org.apache.storm.cluster.DaemonType;
 import org.apache.storm.cluster.IStormClusterState;
+import org.apache.storm.generated.Credentials;
 import org.apache.storm.nimbus.NimbusInfo;
 import org.apache.storm.shade.org.apache.zookeeper.ZkCli;
 import org.apache.storm.utils.ConfigUtils;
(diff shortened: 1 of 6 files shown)


storm-core/src/jvm/org/apache/storm/command/UploadCredentials.java
```diff
@@ -52,7 +61,37 @@ public static void main(String[] args) throws Exception {
...
+                    Map<String, Object> commandLine = Utils.readCommandLineOpts();
+                    List<String> clCreds = (List<String>)commandLine.get(Config.TOPOLOGY_AUTO_CREDENTIALS);
+                    List<String> topoCreds = (List<String>)topologyConf.get(Config.TOPOLOGY_AUTO_CREDENTIALS);
+                    
+                    if (clCreds != null) {
+                        Set<String> extra = new HashSet<>(clCreds);
+                        if (topoCreds != null) {
+                            extra.removeAll(topoCreds);
+                        }
+                        if (!extra.isEmpty()) {
```
agresch: This is great.  But shouldn't we also check for missing credentials?

revans2: In the case where we are missing credentials the old ones will be preserved, but I think you are right I'll add it in.

revans2:
@agresch I added in the check that you wanted.

storm-core/src/jvm/org/apache/storm/command/AdminCommands.java
```diff
@@ -109,6 +133,7 @@ public void printCliHelp(String command, PrintStream out) {
     static {
         COMMANDS.put("remove_corrupt_topologies", new RemoveCorruptTopologies());
         COMMANDS.put("zk_cli", new ZkCli());
+        COMMANDS.put("creds", new CredentialsDebug());
```
agresch: is there documentation that should be updated for these commands?

HeartSaVioR: +1 If we don't have any documentation for admin commands, it should be useful to have one, otherwise end users wouldn't even know its existence.

revans2:
@agresch @HeartSaVioR I added in the docs, and also updated the docs for a few other missing commands.

revans2:
I got a 404 response from a REST api for the supervisor page `/api/v1/supervisor`.

PR #2750, opened 2018-07-02
STORM-3141: Fix NPE in WorkerState.transferLocalBatch, and refactor B…

srdo:
…ackpressureTracker to get rid of placeholder JCQueue

https://issues.apache.org/jira/browse/STORM-3141

### DIFF
--- a/storm-client/src/jvm/org/apache/storm/daemon/worker/BackPressureTracker.java
+++ b/storm-client/src/jvm/org/apache/storm/daemon/worker/BackPressureTracker.java
@@ -59,7 +59,7 @@ public class BackPressureTracker {
      * @return true if an update was recorded, false if taskId is already under BP
      */
     public boolean recordBackPressure(Integer taskId) {
-        return tasks.get(taskId).backpressure.getAndSet(true);
+        return tasks.get(taskId).backpressure.getAndSet(true) == false;
     }
 
     // returns true if there was a change in the BP situation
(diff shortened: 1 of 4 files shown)


storm-client/src/jvm/org/apache/storm/daemon/worker/BackPressureTracker.java
```diff
@@ -22,56 +22,53 @@
...
 
     /***
      * Record BP for a task.
      * This is called by transferLocalBatch() on NettyWorker thread
      * @return true if an update was recorded, false if taskId is already under BP
      */
-    public boolean recordBackPressure(Integer taskId, JCQueue recvQ) {
-        return tasks.put(taskId, recvQ) == NONE;
+    public boolean recordBackPressure(Integer taskId) {
+        return tasks.get(taskId).backpressure.getAndSet(true);
```
revans2: The return value of this method has switched.  getAndSet is going to return the previous value, which is true if it already was under backpressure and false if it was not.  This is the opposite of

```
* @return true if an update was recorded, false if taskId is already under BP
```

This is used to control when the server sends messages to the client about backpressure, and could make it happen not frequently enough and the beginning and then way too frequently after that.

srdo: Nice catch. Fixed and added some tests.

PR #2753, opened 2018-07-05
STORM-3143: Fixed bug of unnecessary inclusion of empty search result

zd-project:
`FindNMatches()` didn't correctly filter out empty match result in `substringSearch()` and hence send back an empty map to user. I don't know if this the desired behavior but a fix to current behavior will make metrics for logviewer easier to implement.

An example of current behavior:

```Json
{
    "fileOffset": 1,
    "searchString": "sdf",
    "matches": [
        {
            "searchString": "sdf",
            "fileName": "word-count-1-1530815972/6701/worker.log",
            "matches": [],
            "port": "6701",
            "isDaemon": "no",
            "startByteOffset": 0
        }
    ]
}
```
Desired behavior:

```Json
{
    "fileOffset": 1,
    "searchString": "sdf",
    "matches": []
}
```

### DIFF
--- a/storm-webapp/src/main/java/org/apache/storm/daemon/logviewer/handler/LogviewerLogSearchHandler.java
+++ b/storm-webapp/src/main/java/org/apache/storm/daemon/logviewer/handler/LogviewerLogSearchHandler.java
@@ -426,7 +426,8 @@ public class LogviewerLogSearchHandler {
 
             int newCount = matchCount + ((List<?>)theseMatches.get("matches")).size();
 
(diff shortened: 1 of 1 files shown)


HeartSaVioR:
The code looks good to me. Is this manually tested with Logviewer UI? I'd just want to check this to skip checking manually as I'm reviewing bunch of PRs at once.

zd-project:
I have manually test against a few cases and they seems to be fine. I don't know how much more rigorous I should test on but judging from the context the change is pretty much self explanatory. Should I remove the comments after the review?

PR #2761, opened 2018-07-12
STORM-2947: Remove deprecated field and method from Thrift

srdo:
Part of https://issues.apache.org/jira/browse/STORM-2947


storm-client/src/storm.thrift
```diff
@@ -206,10 +206,8 @@ struct NimbusSummary {
 
 struct ClusterSummary {
   1: required list<SupervisorSummary> supervisors;
-  //@deprecated, please use nimbuses.uptime_secs instead.
-  2: optional i32 nimbus_uptime_secs = 0;
-  3: required list<TopologySummary> topologies;
-  4: required list<NimbusSummary> nimbuses;
+  2: required list<TopologySummary> topologies;
+  3: required list<NimbusSummary> nimbuses;
```
revans2: Can we please follow the guidelines for maintaining wire compatibility?  this way old clients still have a chance of working with the newer servers.

```
struct ClusterSummary {
   1: required list<SupervisorSummary> supervisors;
   // removed (DO NOT REUSE) 2:
   3: required list<TopologySummary> topologies;
   4: required list<NimbusSummary> nimbuses;
}
```

srdo: Sure. Do you have a link to the guidelines, I didn't see anything in DEVELOPER.md?

srdo:
Should I just comment out beginFileDownload in the thrift file, or does it need to stay in Nimbus and LocalCluster commented out as well?

revans2:
@srdo Only comment it out in the thrift code.  Delete the java code.  The only reason to comment it out instead of deleting it is so there is a visual clue to anyone editing the file in the future that this API used to be here, and we should avoid reusing it.  Just on the off chance that we recreate the same API but the semantics are incompatible with each other and we get really odd errors.

storm-client/src/storm.thrift
srdo: Thanks. Is it worth it to link to in DEVELOPER,md, or somewhere else do you think?

storm-client/src/jvm/org/apache/storm/generated/ClusterSummary.java
```diff
@@ -63,8 +60,6 @@ public static _Fields findByThriftId(int fieldId) {
       switch(fieldId) {
         case 1: // SUPERVISORS
           return SUPERVISORS;
-        case 2: // NIMBUS_UPTIME_SECS
-          return NIMBUS_UPTIME_SECS;
         case 3: // TOPOLOGIES
           return TOPOLOGIES;
```
danny0405: Should we keep the sequence number as it is now ?

srdo: It's my impression that if we don't keep the field ids as they are now, it won't be possible for people to do a rolling upgrade. E.g. if we changed TOPOLOGIES to have id 2, then someone doing a rolling upgrade might have a machine send ClusterSummary with TOPOLOGIES id 3 to an upgraded machine where TOPOLOGIES has id 2, which breaks deserialization.

HeartSaVioR:
@revans2 Could you revisit this PR?

storm-client/src/jvm/org/apache/storm/generated/ClusterSummary.java
srdo: This is generated code. The comment about this is here https://github.com/apache/storm/pull/2761/files#diff-61bbff6c872f2afa10a3b695c1cfe4a3R209

PR #2762, opened 2018-07-12
STORM-3148: Avoid threading issues with kryo

revans2:
I am going to run some perf tests just to be sure, but the only real impact should be slightly more memory usage, and no more than we have to 1.x topologies.


revans2:
I am seeing some issues with this patch under heavy load let me debug it before merging anything.

revans2:
The error appears to be related to a separate JIRA around backpressure status, but I want to be sure.

revans2:
I fixed the issues with backpressure.  It turns out that the system bolt was reporting backpressure too.  So I updated the code to filter it out on the sending side, but to deal with the error appropriately on the receiving side.

revans2:
Oh and the perf tests look good now.  I was able to do 100,000 sentences per second on my MBP on ThroughputVsLatency.  I couldn't really compare the memory usage as even without my change it would not run because of the backpressure issue.

storm-client/src/jvm/org/apache/storm/messaging/netty/StormClientHandler.java
```diff
@@ -47,12 +47,20 @@ public void channelRead(ChannelHandlerContext ctx, Object message) throws Except
             BackPressureStatus status = (BackPressureStatus) message;
             if (status.bpTasks != null) {
                 for (Integer bpTask : status.bpTasks) {
-                    remoteBpStatus[bpTask].set(true);
+                    try {
+                        remoteBpStatus[bpTask].set(true);
+                    } catch (ArrayIndexOutOfBoundsException e) {
```
HeartSaVioR: Could this happen in normal situation? I guess we don't expect it. I don't think catching exception would hurt, but just to be sure about this.

storm-client/test/jvm/org/apache/storm/messaging/netty/BackPressureStatusTest.java
```diff
@@ -0,0 +1,35 @@
...
+import org.apache.storm.shade.io.netty.buffer.UnpooledByteBufAllocator;
+import org.apache.storm.utils.Utils;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+class BackPressureStatusTest {
+
+    @Test
+    void bufferTest() throws IOException {
```
srdo: What is this test demonstrating?

storm-client/src/jvm/org/apache/storm/messaging/netty/StormClientHandler.java
srdo: I'm also not sure why this can happen?

revans2: So there are two issues happening in the jira.  Previously when under heavy load a backpressure messsage would be sent.  Under the old code this might contain a -1 which would cause an ArrayIndexOutOFBoundsException to happen.  When it did some backpressure status updates in the message might be lost (specifically the ones that turn off backpressure) and the topology would stop processing.  I fixed the cause of the -1 indexes, but I thought it would be best to also guard against future problems and if there is one we can process the rest of the status updates.

storm-client/test/jvm/org/apache/storm/messaging/netty/BackPressureStatusTest.java
revans2: The original problem was because kryo was throwing an exception trying to serialize a BackPressureStatus message.  I wrote this test initially to try and debug what was happening, and I ended up serializing millions of BackPressureStatus in loops to try and verify that it was working properly.  When I failed I remembered that kryo was not thread safe and started looking around there.  If you want me to delete it I can.

revans2: Actually I'll just delete it because there is no reason to keep it, especially because we don't verify the result in any way.

storm-client/src/jvm/org/apache/storm/messaging/netty/StormClientHandler.java
HeartSaVioR: OK. Makes sense. I know message of exception will contain the index which will be included to log message, but explicitly write in log message would help users to find the index while grepping since it can be found within same line.

srdo: Makes sense, thanks.

storm-client/src/jvm/org/apache/storm/messaging/netty/StormClientHandler.java
```diff
@@ -47,12 +47,20 @@ public void channelRead(ChannelHandlerContext ctx, Object message) throws Except
...
+                        LOG.error("BP index out of bounds {}", e);
```
srdo: Nit: Is it intentional that the stack trace is stripped here?

storm-client/src/jvm/org/apache/storm/messaging/netty/StormClientHandler.java
```diff
@@ -47,12 +47,24 @@ public void channelRead(ChannelHandlerContext ctx, Object message) throws Except
             BackPressureStatus status = (BackPressureStatus) message;
             if (status.bpTasks != null) {
                 for (Integer bpTask : status.bpTasks) {
-                    remoteBpStatus[bpTask].set(true);
+                    try {
+                        remoteBpStatus[bpTask].set(true);
+                    } catch (ArrayIndexOutOfBoundsException e) {
+                        //Just in case we get something we are confused about
+                        // we can continue processing the rest of the tasks
+                        LOG.error("BP index out of bounds {}", e);
```
srdo: Is it intentional that the stack trace is stripped here?

srdo:
+1

storm-client/src/jvm/org/apache/storm/messaging/netty/StormClientHandler.java
HeartSaVioR: Ah! I realized `e` is bound to {} so my suggestion is already applied (show index within same line), but no stack trace.

storm-client/src/jvm/org/apache/storm/messaging/netty/StormClientHandler.java
srdo: I'm a little surprised that an uncaught exception here doesn't crash the worker though? I'd expect the worker to crash, and for the worker to come back up with backpressure disabled?

storm-client/src/jvm/org/apache/storm/messaging/netty/StormClientHandler.java
revans2: Yes, because when this was happening it was happening everywhere, and there is only one line under the try catch block so there isn't a lot of confusion about what happened.

storm-client/src/jvm/org/apache/storm/messaging/netty/StormClientHandler.java
revans2: There is no way to disable backpressure right now.  It didn't crash because this is running under netty.   Netty will catch the exception and pass it off to an exception handler/the future.  I didn't trace down the exact path and why we were not handling this exception to cause it to come down, but I will file a follow on JIRA to examine it.

PR #2763, opened 2018-07-13
STORM-3150: Improve Gauge registration methods and refactored code

zd-project:
STORM-3150

### DIFF
--- a/storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
+++ b/storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
@@ -1936,13 +1936,13 @@ public class Nimbus implements Iface, Shutdownable, DaemonCommon {
     }
 
     private double fragmentedMemory() {
-        Double res = nodeIdToResources.get().values().parallelStream().filter(x -> isFragmented(x) == true)
+        Double res = nodeIdToResources.get().values().parallelStream().filter(this::isFragmented)
                                       .mapToDouble(SupervisorResources::getAvailableMem).filter(x -> x > 0).sum();
         return res.intValue();
     }
 
     private int fragmentedCpu() {
-        Double res = nodeIdToResources.get().values().parallelStream().filter(x -> isFragmented(x) == true)
+        Double res = nodeIdToResources.get().values().parallelStream().filter(this::isFragmented)
                                       .mapToDouble(SupervisorResources::getAvailableCpu).filter(x -> x > 0).sum();
         return res.intValue();
     }
@@ -2808,20 +2808,25 @@ public class Nimbus implements Iface, Shutdownable, DaemonCommon {
                                     });
 
             StormMetricsRegistry.registerGauge("nimbus:num-supervisors", () -> state.supervisors(null).size());
-            StormMetricsRegistry.registerGauge("nimbus:fragmented-memory", () -> fragmentedMemory());
-            StormMetricsRegistry.registerGauge("nimbus:fragmented-cpu", () -> fragmentedCpu());
-            StormMetricsRegistry.registerGauge("nimbus:available-memory", () -> nodeIdToResources.get().values().parallelStream()
-                                                                                                 .mapToDouble(
-                                                                                                     SupervisorResources::getAvailableMem)
-                                                                                                 .sum());
-            StormMetricsRegistry.registerGauge("nimbus:available-cpu", () -> nodeIdToResources.get().values().parallelStream().mapToDouble(
-                SupervisorResources::getAvailableCpu).sum());
-            StormMetricsRegistry.registerGauge("nimbus:total-memory", () -> nodeIdToResources.get().values().parallelStream()
-                                                                                             .mapToDouble(SupervisorResources::getTotalMem)
-                                                                                             .sum());
-            StormMetricsRegistry.registerGauge("nimbus:total-cpu", () -> nodeIdToResources.get().values().parallelStream()
-                                                                                          .mapToDouble(SupervisorResources::getTotalCpu)
-                                                                                          .sum());
+            StormMetricsRegistry.registerGauge("nimbus:fragmented-memory", this::fragmentedMemory);
+            StormMetricsRegistry.registerGauge("nimbus:fragmented-cpu", this::fragmentedCpu);
+            StormMetricsRegistry.registerGauge("nimbus:available-memory", () -> nodeIdToResources.get().values()
+                .parallelStream()
+                .mapToDouble(SupervisorResources::getAvailableMem)
+                .sum());
+            StormMetricsRegistry.registerGauge("nimbus:available-cpu", () -> nodeIdToResources.get().values()
+                .parallelStream()
+                .mapToDouble(SupervisorResources::getAvailableCpu)
+                .sum());
+            StormMetricsRegistry.registerGauge("nimbus:total-memory", () -> nodeIdToResources.get().values()
+                .parallelStream()
+                .mapToDouble(SupervisorResources::getTotalMem)
+                .sum());
+            StormMetricsRegistry.registerGauge("nimbus:total-cpu", () -> nodeIdToResources.get().values()
+                .parallelStream()
+                .mapToDouble(SupervisorResources::getTotalCpu)
+                .sum());
+
             StormMetricsRegistry.startMetricsReporters(conf);
 
             if (clusterConsumerExceutors != null) {
--- a/storm-server/src/main/java/org/apache/storm/metric/StormMetricsRegistry.java
+++ b/storm-server/src/main/java/org/apache/storm/metric/StormMetricsRegistry.java
@@ -19,7 +19,7 @@ import com.codahale.metrics.Metric;
 import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.Reservoir;
 import java.util.Map;
-import java.util.concurrent.Callable;
+
 import org.apache.storm.daemon.metrics.MetricsUtils;
 import org.apache.storm.daemon.metrics.reporters.PreparableReporter;
 import org.slf4j.Logger;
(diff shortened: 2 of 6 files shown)


storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
```diff
@@ -2809,21 +2809,19 @@ public void launchServer() throws Exception {
                                         }
                                     });
 
+            //Be cautious using method reference instead of lambda. subexpression preceding :: will be evaluated only upon evaluation
```
srdo: Nit: Not sure about this comment, it's not really specific to this code, but applies to any use of method references anywhere.

storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
```diff
@@ -2809,21 +2809,19 @@ public void launchServer() throws Exception {
...
-            StormMetricsRegistry.registerGauge("nimbus:fragmented-cpu", () -> fragmentedCpu());
+            StormMetricsRegistry.registerGauge("nimbus:fragmented-memory", this::fragmentedMemory);
+            StormMetricsRegistry.registerGauge("nimbus:fragmented-cpu", this::fragmentedCpu);
             StormMetricsRegistry.registerGauge("nimbus:available-memory", () -> nodeIdToResources.get().values().parallelStream()
-                                                                                                 .mapToDouble(
-                                                                                                     SupervisorResources::getAvailableMem)
-                                                                                                 .sum());
-            StormMetricsRegistry.registerGauge("nimbus:available-cpu", () -> nodeIdToResources.get().values().parallelStream().mapToDouble(
-                SupervisorResources::getAvailableCpu).sum());
+                .mapToDouble(SupervisorResources::getAvailableMem).sum());
```
srdo: Nit: I think the readability of method chaining is improved by having one call per line, e.g. 

```
x.stream
  .map
  .collect
```
rather than
```
x.stream
.map.collect
```
It's a super minor complaint though.

storm-server/src/main/java/org/apache/storm/metric/StormMetricsRegistry.java
```diff
@@ -27,52 +28,63 @@
...
+    public static <V> Gauge<V> registerGauge(final String name, final Callable<V> fn) {
+        return register(name, () -> {
+            try {
+                return fn.call();
+            } catch (Exception e) {
+                LOG.error("Error getting gauge value for {}", name, e);
             }
-        };
-        return register(name, gauge);
+            return null;
```
srdo: Why does it make sense to return null rather than 0?

storm-server/src/main/java/org/apache/storm/metric/StormMetricsRegistry.java
```diff
@@ -27,52 +28,63 @@
...
-            @Override
-            public Integer getValue() {
-                try {
-                    return (Integer) fn.call();
-                } catch (Exception e) {
-                    LOG.error("Error getting gauge value for {}", name, e);
-                }
-                return 0;
+    /**
+     * Register a gauge with provided callback.
```
srdo: Nit: Consider adding a note that this handles exceptions thrown from the callback.

storm-server/src/main/java/org/apache/storm/metric/StormMetricsRegistry.java
```diff
@@ -27,52 +28,63 @@
...
-        };
-        return register(name, gauge);
+            return null;
+        });
     }
 
-    public static void registerProvidedGauge(final String name, Gauge gauge) {
+    /**
+     * Register a provided gauge. Use this method if custom gauges is needed or
+     * no checked exceptions should be handled.
```
srdo: Nit: Consider rephrasing as "or exceptions should not be handled"

storm-server/src/main/java/org/apache/storm/metric/StormMetricsRegistry.java
```diff
@@ -27,52 +28,63 @@
...
         register(name, gauge);
     }
 
     public static Histogram registerHistogram(String name, Reservoir reservoir) {
-        Histogram histogram = new Histogram(reservoir);
-        return register(name, histogram);
+        return register(name, new Histogram(reservoir));
+    }
+
+    public static void registerAll(final String prefix, MetricSet metrics) {
```
srdo: Why is this method necessary?

storm-server/src/main/java/org/apache/storm/metric/StormMetricsRegistry.java
zd-project: Should I note so in the JavaDoc then?

storm-server/src/main/java/org/apache/storm/metric/StormMetricsRegistry.java
zd-project: This is used in #2764. Should move the change there?

storm-server/src/main/java/org/apache/storm/metric/StormMetricsRegistry.java
srdo: Yes please :)

storm-server/src/main/java/org/apache/storm/metric/StormMetricsRegistry.java
srdo: Having it here is fine, but I'm not sure I understand what it does differently from just calling the register method directly? Is it just for clarity?

storm-server/src/main/java/org/apache/storm/metric/StormMetricsRegistry.java
zd-project: Because we are using generic type now. In addition reporters from metrics package seem to be handling null case gracefully so I think it's safe to use. That being said I should probably add this in the doc or notify people in some way if they want to inherit custom reporter.

storm-server/src/main/java/org/apache/storm/metric/StormMetricsRegistry.java
zd-project: For wrapper to the private `register()` method. I don't see any problem if we should make `register` public though.

storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
zd-project: Sorry I was writing this to myself because I was not super familiar with Java lambda. Should probably delete it as it's not relevant to code.

storm-server/src/main/java/org/apache/storm/metric/StormMetricsRegistry.java
srdo: I think it's a little weird to have this and `unregister` be public, but not the generic `register` method.

srdo: I don't have a strong opinion on this, but making the `register` method public might make for a nicer API, especially since `unregister` can unregister any kind of metric.

zd-project: I'll have to think about this a bit more, since I will have a few more PR depending on this. The way `MetricRegistry` handles `MetricSet` under the hood is a bit wacky, and they haven't supported `removeAll(MetricSet set)` method in their release yet. Do you have any advice of improving `StormMetricsRegistry` in general?

zd-project: Improved in #2771

srdo:
+1, thanks for your patience. Please squash to one commit, and I'll merge.

zd-project:
Okay thanks. Can you also help reviewing #2771? If you think merging two PR is better.

srdo:
Yes, I'll take a look.

storm-server/src/main/java/org/apache/storm/metric/StormMetricsRegistry.java
```diff
@@ -27,52 +28,67 @@
...
+     * Register a gauge with provided callback. This swallows all exceptions
+     * thrown from the callback, consider using {@link #registerProvidedGauge(String, Gauge)}
+     * if no exceptions will be thrown by the callable.
+     *
+     * @param name name of the gauge
+     * @param fn callback that measures
+     * @param <V> type of measurement the callback returns, also the type of gauge
+     * @return registered gauge
+     */
+    public static <V> Gauge<V> registerGauge(final String name, final Callable<V> fn) {
```
revans2: Could we just change this to take a guage instead?

```
public static <V> Gauge<V> registerGauge(final String name, final Gauge<V> gauge) {
    return register(name, gauge);
}
```
We are not calling this from clojure anymore so we didn't need the Callable any more, and for the most part there are no code changes to make this happen.

storm-server/src/main/java/org/apache/storm/scheduler/blacklist/BlacklistScheduler.java
```diff
@@ -89,13 +89,8 @@ public void prepare(Map<String, Object> conf) {
...
         blacklistHost = new HashSet<>();
 
-        StormMetricsRegistry.registerGauge("nimbus:num-blacklisted-supervisor", new Callable<Integer>() {
-            @Override
-            public Integer call() throws Exception {
-                //nimbus:num-blacklisted-supervisor + none blacklisted supervisor = nimbus:num-supervisors
-                return blacklistHost.size();
-            }
-        });
+        //nimbus:num-blacklisted-supervisor + none blacklisted supervisor = nimbus:num-supervisors
```
revans2: nit: I have no idea what this comment means, so could we remove it?  Unless you understand it, then can we make it clearer?

storm-server/src/main/java/org/apache/storm/metric/StormMetricsRegistry.java
zd-project: Okay, should I first merge change in this PR into #2771 and then port them to where they are actually invoked?

storm-server/src/main/java/org/apache/storm/metric/StormMetricsRegistry.java
zd-project: Callable seems exists only for swallowing exceptions now. I don't know if this is part of the original design but I don't see a problem eliminating so.

storm-server/src/main/java/org/apache/storm/scheduler/blacklist/BlacklistScheduler.java
zd-project: It looks to me that there's a typo: nimbus:num-blacklisted-supervisor + non-blacklisted supervisor = nimbus:num-supervisors

storm-server/src/main/java/org/apache/storm/metric/StormMetricsRegistry.java
revans2: Please do I don't see a reason to keep it here

zd-project:
All nits are addressed and PR is rebased on current master branch. Do you want to merge this PR right now or merge this together with #2771?

srdo:
@zd-project I think we can merge this immediately. Could you please squash back to one commit?

srdo:
@zd-project Merged, thanks for your patience. JIRA is down right now, will update the ticket when it comes back up.

PR #2765, opened 2018-07-14
[STORM-3152]Storm has supported ipv6 but Troubleshooting.md didn't update

liu-zhaokun:
[https://issues.apache.org/jira/browse/STORM-3152](https://issues.apache.org/jira/browse/STORM-3152)
As storm's socket get inetaddress by java's InetAddress class,I think there is no difficulty to run Topology on ipv6.And I have test it on ipv6,all socket can be established on ipv6's ip.

### DIFF
--- a/docs/Troubleshooting.md
+++ b/docs/Troubleshooting.md
@@ -27,7 +27,6 @@ Possible symptoms:
 
 Solutions:
 
- * Storm doesn't work with ipv6. You can force ipv4 by adding `-Djava.net.preferIPv4Stack=true` to the supervisor child options and restarting the supervisor. 
  * You may have a misconfigured subnet. See the solutions for `Worker processes are crashing on startup with no stack trace`
 
 ### Topology stops processing tuples after awhile


revans2:
There are still a few places in the UI and in some test code that look like we are splitting on ":" to separate out a host from a port.  This is not going to work on raw IPv6 addresses.  I would prefer it if we updated the comment to say that storm supports IPv6, but there may still be some issues with raw IPv6 Addresses.

liu-zhaokun:
@revans2
Thanks for your reply.I will try to modify the way to separate out a host from a port.

liu-zhaokun:
@revans2
Thanks for your reply.I will try to modify the way to separate out a host from a port.

rzo1:
We are currently cleaning up old issues with stale discussions. The last comment here was made 5 years ago and Storm as evolved. If this is still an issue or relevant for your work, feel free to re-open and to rebase / fix the related PR.

PR #2766, opened 2018-07-17
STORM-2972: Replace storm-kafka with storm-kafka-client in storm-sql-…

srdo:
…kafka

https://issues.apache.org/jira/browse/STORM-2972

The new kafka scheme syntax is stolen from [Apache Camel](https://github.com/apache/camel/blob/master/components/camel-kafka/src/main/docs/kafka-component.adoc), since I figure we are doing something very similar to them.

The primary limitation here compared to the storm-kafka version is that you can't include metadata (e.g. partition) in the emitted tuple, because I had to adapt the Scheme API to the RecordTranslator API used in storm-kafka-client. Emitting metadata with a Scheme in the storm-kafka spout is handled by [special case instanceof'ing the scheme](https://github.com/apache/storm/blob/af42f434f4a4c3d9087c6058b359033736d3b5e8/external/storm-kafka/src/jvm/org/apache/storm/kafka/KafkaUtils.java#L134) to see if it fits one that takes metadata, and I'd really prefer not to do that.

If we need to be able to emit metadata as well, we can just allow people to override the RecordTranslatorSchemeAdapter. I'll leave that as a task for later, unless someone objects.

### DIFF
--- a/docs/storm-sql-example.md
+++ b/docs/storm-sql-example.md
@@ -9,7 +9,7 @@ This page is written by "how-to" style so you can follow the step and learn how 
 
 ## Preparation
 
-This page assumes that Apache Zookeeper, Apache Storm and Apache Kafka is installed locally and running with properly configured.
+This page assumes that Apache Zookeeper, Apache Storm and Apache Kafka are installed locally and running properly configured.
 For convenience, this page assumes that Apache Kafka 0.10.0 is installed via `brew`.
 
 We'll use below tools to prepare the JSON data which will be fed to the input data source. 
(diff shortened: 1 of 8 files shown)


srdo:
I tried out storm-sql-kafka with a small example similar to the orders example at https://storm.apache.org/releases/2.0.0-SNAPSHOT/storm-sql.html. I adjusted the docs for the storm-sql-example page, but did not go through and run it.

docs/storm-sql-example.md
```diff
@@ -291,9 +291,4 @@ That's it! Supposing we have UDF which queries geo location via remote ip, we ca
 
 ## Summary
 
-We looked through several simple use cases for Storm SQL to learn Storm SQL features. If you haven't looked at [Storm SQL integration](storm-sql.html) and [Storm SQL language](storm-sql-reference.html), you need to read it to see full supported features. 
-
-Note that Storm SQL is running on Trident, which is micro-batch, and also no strong typed. Sink doesn't actually check the type.
```
HeartSaVioR: Nice finding! Thanks for handling this.

docs/storm-sql-internal.md
```diff
@@ -18,18 +18,16 @@ Figure 1 describes the workflow of executing a SQL query in StormSQL. First, use
 <p>Figure 1: Workflow of StormSQL.</p>
 </div>
 
-Note: Trident Topology is now replaced with normal Storm topology leveraging Streams API.
```
HeartSaVioR: Yeah I forgot to remove the line after renewing diagrams. Nice finding.

PR #2768, opened 2018-07-18
STORM-3156: Remove the transactional topology API

srdo:
https://issues.apache.org/jira/browse/STORM-3156

I'm not very familiar with the transactional topology API, so please let me know if I missed any classes.

I left the transactional topology documentation in place as a reference in the Trident documentation, because I think the documentation is well written and provides a good explanation of the problems solved by Trident, and how Storm arrived at the Trident API.

Also I noted that the tutorial page wasn't linked from the index, so I added a link.


HeartSaVioR:
I don't see any use case on using transactional API (even when I start contribution on Apache Storm) so completely OK to remove, but feel safer to ask @revans2 to verify Bobby hasn't have use case on transactional API once more...

HeartSaVioR:
@revans2 Could you review this? I'm softly +1 but assuming there's no usage on transactional API.

danny0405:
@HeartSaVioR 
I don't see any usage on transactional API either, but i agree to wait for @revans2 's reply.

revans2:
Sorry about my slow reply.  I have been heads down on some other issues lately.  I do know of some people who have used the transaction API, but Trident is the replacement for it.  It uses the same concepts behind the scenes, but with a higher level API.

PR #2769, opened 2018-07-19
STORM-3159: Fixed a potential file resource leak.

zd-project:
The file is not properly put in the try-with-resource/try-finally block.

### DIFF
--- a/storm-server/src/main/java/org/apache/storm/utils/ServerUtils.java
+++ b/storm-server/src/main/java/org/apache/storm/utils/ServerUtils.java
@@ -636,15 +636,14 @@ public class ServerUtils {
      * @throws IOException
      */
     public static long zipFileSize(File myFile) throws IOException {
-        RandomAccessFile raf = new RandomAccessFile(myFile, "r");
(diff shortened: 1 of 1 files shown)


zd-project:
This will affect the implementation of #2754

danny0405:
So, in what case the file resource leak will leak ?

zd-project:
The file resource was not wrapped in try-finally or try-with-resource block. In case of IOException, resource may not be closed properly

storm-server/src/main/java/org/apache/storm/utils/ServerUtils.java
```diff
@@ -635,16 +635,16 @@ public static void unZip(File inFile, File toDir) throws IOException {
      *
      * @throws IOException
      */
+    //Used only by logviewer, considering moving to web-app?
```
agresch: even if used only in one place, if it is a general purpose function, I think it makes sense to have in a global place.  I'd remove this comment.

srdo:
+1. Please raise an issue for this, and squash to one commit that contains the issue number.

PR #2773, opened 2018-07-24
Blobstore sync bug fix

jiangzhileaf:
1.Bug fix for blob sync frequency with time unit error.
2.Bug fix for blob sync delete file, add catch NoSuchFileException.
3.Bug fix for blob sync update blob flie, add catch FileNotFoundException

### DIFF
--- a/storm-server/src/main/java/org/apache/storm/blobstore/BlobStoreUtils.java
+++ b/storm-server/src/main/java/org/apache/storm/blobstore/BlobStoreUtils.java
@@ -12,6 +12,7 @@
 
 package org.apache.storm.blobstore;
 
+import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.HashSet;
@@ -191,6 +192,8 @@ public class BlobStoreUtils {
                     out.close();
                 }
                 isSuccess = true;
+            } catch(FileNotFoundException fnf) {
+                LOG.warn("FileNotFoundException", fnf);
             } catch (IOException | AuthorizationException exception) {
                 throw new RuntimeException(exception);
             } catch (KeyNotFoundException knf) {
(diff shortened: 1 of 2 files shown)


storm-server/src/main/java/org/apache/storm/blobstore/LocalFsBlobStore.java
```diff
@@ -197,7 +198,7 @@ public void run() {
                     throw new RuntimeException(e);
                 }
             }
-        }, 0, ObjectReader.getInt(conf.get(DaemonConfig.NIMBUS_CODE_SYNC_FREQ_SECS)));
+        }, 0, ObjectReader.getInt(conf.get(DaemonConfig.NIMBUS_CODE_SYNC_FREQ_SECS))*1000);
```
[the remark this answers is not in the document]
HeartSaVioR: Nice catch!

storm-server/src/main/java/org/apache/storm/blobstore/BlobStoreUtils.java
```diff
@@ -191,6 +192,8 @@ public static boolean downloadUpdatedBlob(Map<String, Object> conf, BlobStore bl
                     out.close();
                 }
                 isSuccess = true;
+            } catch(FileNotFoundException fnf) {
```
HeartSaVioR: I'd rather leave it as it is, unless we feel confident that it is OK to swallow here.

revans2: I agree with @HeartSaVioR here.

HeartSaVioR:
@jiangzhileaf Any updates?

storm-server/src/main/java/org/apache/storm/blobstore/BlobStoreUtils.java
arunmahadevan: Btw, theres a race condition where the topology gets killed while the blob download is going on and apparently it throws a FNF exception here when invoking out.close().

arunmahadevan:
Seems to fix a few race conditions, may be we can take this forward.

HeartSaVioR:
@arunmahadevan I'd be really happy if you can take forward with this. Looks like we lost @jiangzhileaf.

arunmahadevan:
@HeartSaVioR created https://github.com/apache/storm/pull/2871 and linked to a JIRA. 

Maybe good to get this in 2.0 release.

srdo:

PR #2774, opened 2018-07-24
Bug fix for BaseResourceAwareStrategy to avoid NullPointerException

jiangzhileaf:
Bug fix for BaseResourceAwareStrategy to avoid NullPointerException, when a supervisor is in bad list but alive.

### DIFF
--- a/storm-server/src/main/java/org/apache/storm/scheduler/resource/strategies/scheduling/BaseResourceAwareStrategy.java
+++ b/storm-server/src/main/java/org/apache/storm/scheduler/resource/strategies/scheduling/BaseResourceAwareStrategy.java
@@ -357,9 +357,12 @@ public abstract class BaseResourceAwareStrategy implements IStrategy {
             ObjectResources rack = new ObjectResources(rackId);
             racks.add(rack);
             for (String nodeHost : nodeHosts) {
-                for (RAS_Node node : hostnameToNodes(nodeHost)) {
-                    rack.availableResources.add(node.getTotalAvailableResources());
-                    rack.totalResources.add(node.getTotalAvailableResources());
+                List<RAS_Node> nodes = hostnameToNodes(nodeHost);
+                if(nodes != null) {
+                    for (RAS_Node node : nodes) {
+                        rack.availableResources.add(node.getTotalAvailableResources());
+                        rack.totalResources.add(node.getTotalAvailableResources());
+                    }
                 }
             }
 


storm-server/src/main/java/org/apache/storm/scheduler/resource/strategies/scheduling/BaseResourceAwareStrategy.java
```diff
@@ -357,9 +357,12 @@ private AllResources createClusterAllResources() {
             ObjectResources rack = new ObjectResources(rackId);
             racks.add(rack);
             for (String nodeHost : nodeHosts) {
-                for (RAS_Node node : hostnameToNodes(nodeHost)) {
-                    rack.availableResources.add(node.getTotalAvailableResources());
-                    rack.totalResources.add(node.getTotalAvailableResources());
+                List<RAS_Node> nodes = hostnameToNodes(nodeHost);
+                if(nodes != null) {
+                    for (RAS_Node node : nodes) {
```
danny0405: we initialize the nodes with new HashMap<>(), so it will never be a null, and this fix is not necessary

danny0405:
Sorry, we initialize the nodes with new HashMap<>(), so it will never be a null, and this fix is not necessary

storm-server/src/main/java/org/apache/storm/scheduler/resource/strategies/scheduling/BaseResourceAwareStrategy.java
danny0405: Oh, my fault, the nodes may be null. +1 BTW.

storm-server/src/main/java/org/apache/storm/scheduler/resource/strategies/scheduling/BaseResourceAwareStrategy.java
```diff
@@ -357,9 +357,12 @@ private AllResources createClusterAllResources() {
             ObjectResources rack = new ObjectResources(rackId);
             racks.add(rack);
             for (String nodeHost : nodeHosts) {
-                for (RAS_Node node : hostnameToNodes(nodeHost)) {
-                    rack.availableResources.add(node.getTotalAvailableResources());
-                    rack.totalResources.add(node.getTotalAvailableResources());
+                List<RAS_Node> nodes = hostnameToNodes(nodeHost);
+                if(nodes != null) {
```
revans2: nit: need a space after the `if`

revans2:
@jiangzhileaf I am trying to wrap my head around which situations would cause this to happen.  It appears that in some cases you can get a host name from the network topography that has no corresponding RAS_Node for it.  But the network topography is created from the list of supervisors.

https://github.com/apache/storm/blob/9cc5b72c1fae78c60061b9f3c3f3d41277ca93cf/storm-server/src/main/java/org/apache/storm/scheduler/Cluster.java#L172

and are the RAS_Nodes.

https://github.com/apache/storm/blob/9cc5b72c1fae78c60061b9f3c3f3d41277ca93cf/storm-server/src/main/java/org/apache/storm/scheduler/resource/RAS_Nodes.java#L80

So the only thing I can think of that would cause this to fail is that somehow the the implementation of DNSToSwitchMapping being used has violated the contract of the API and returned more than it should.

https://github.com/apache/storm/blob/9cc5b72c1fae78c60061b9f3c3f3d41277ca93cf/storm-client/src/jvm/org/apache/storm/networktopography/DNSToSwitchMapping.java#L24-L35

If we want to support that use case, which I think is fine to make it more generic, I would rather have us fix-up the results returned by the network mapping.  Also we would need to update the javadocs to explain what the new contract is.

rzo1:
We are currently cleaning up old issues with stale discussions. The last comment here was made 5 years ago and Storm as evolved. If this is still an issue or relevant for your work, feel free to re-open and to rebase / fix the related PR.

PR #2776, opened 2018-07-27
STORM-3161 Local mode should force setting min replication count to 1

HeartSaVioR:
When topology.min.replication.count is set to more than 1, nimbus in local mode
never achieve condition for replication, hence stuck on handling blobs.

We should force set it to 1 in local mode to avoid this situation.

Will also submit patch for 1.x-branch.


danny0405:
@HeartSaVioR 
What is the wrong stake trace, to be stuck on handling blobs, you mean topologies will not get any jar/conf/ser ?

HeartSaVioR:
@danny0405 It is not an error but stuck with no update. 

https://github.com/apache/storm/blob/3b5f9e7c75b378881060f454d39e6bf653e0bcb3/storm-core/src/clj/org/apache/storm/daemon/nimbus.clj#L502

In local mode we always launch only one nimbus, hence never achieve replication if min replication count is set to higher than 1.

arunmahadevan:
+1

PR #2777, opened 2018-07-27
(1.x) STORM-3161 Local mode should force setting min replication count to 1

HeartSaVioR:
When topology.min.replication.count is set to more than 1, nimbus in local mode
never achieve condition for replication, hence stuck on handling blobs.

We should force set it to 1 in local mode to avoid this situation.

Also set storm.home to new temporary directory to avoid using actual log directory.

Patch for master branch: #2776


arunmahadevan:
+1

PR #2779, opened 2018-07-29
STORM-3163: Make ShellBolt logger setup calls occur in-thread to support MDC

mal:
Prior to this the logger set up would occur in a separate thread (the
parent), to the one doing the logging. This meant that MDC values, which
are thread local, could not persist from the set up phase. By doing set
up in the same thread, MDC values are now reliably available during log
calls.

https://issues.apache.org/jira/browse/STORM-3163


mal:
Solved by leveraging `log4j2`'s `isThreadContextMapInheritable` property.

PR #2786, opened 2018-08-01
STORM-3168 prevent AsyncLocalizer cleanup from crashing

agresch:
Just catching and logging any error that may be occurring.


srdo:
Seems reasonable to me, but do we really want to catch the Error subtypes as well?

agresch:
If an error occurred, what should be the expected behavior?  Shutdown?

srdo:
Yes, we probably don't want to keep running if we hit an Error.

agresch:
@srdo - updated to shutdown on errors.  Filed https://issues.apache.org/jira/browse/STORM-3174 to standardize exit codes.  I'm not clear on why we're selecting the values we are.

srdo:
Thanks. +1

PR #2787, opened 2018-08-01
STORM-3169: Correctly convert configured minutes to milliseconds

zd-project:
See https://issues.apache.org/jira/browse/STORM-3169 for details

### DIFF
--- a/storm-webapp/src/main/java/org/apache/storm/daemon/logviewer/utils/LogCleaner.java
+++ b/storm-webapp/src/main/java/org/apache/storm/daemon/logviewer/utils/LogCleaner.java
@@ -44,6 +44,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.SortedSet;
 import java.util.TreeSet;
+import java.util.concurrent.TimeUnit;
 import java.util.function.BinaryOperator;
 import java.util.stream.StreamSupport;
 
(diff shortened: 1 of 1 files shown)


zd-project:
Seems this PR is passing on my repo but failed here. There may be some unrelated bugs out there.

storm-webapp/src/test/java/org/apache/storm/daemon/logviewer/utils/LogCleanerTest.java
```diff
@@ -87,8 +88,9 @@ public void testMkFileFilterForLogCleanup() throws IOException {
 
         final long nowMillis = Time.currentTimeMillis();
         final long cutoffMillis = logCleaner.cleanupCutoffAgeMillis(nowMillis);
-        final long oldMtimeMillis = cutoffMillis - 500;
-        final long newMtimeMillis = cutoffMillis + 500;
+        final long interval = TimeUnit.MINUTES.toMillis(500);
```
srdo: I don't think this change is necessary. The test is mocking some files for right before and right after the cutoff. Why do we need to make it 500 minutes before/after the cutoff instead of 500 milliseconds?

srdo: Not sure I follow. The cutoffMillis is calculated based on the 60 minutes, so would be `now - 60 minutes`. It should be the same cutoffMillis used by the log cleaner. 500 milliseconds before the cutoff is before the cutoff, right?

srdo: Just to try to be a little more clear, the test needs a timestamp before the cutoff, and a timestamp after the cutoff. The cutoff is `now - 60 minutes`. `now - 60 minutes - 500 ms` is before the cutoff. I don't understand why it instead needs to be `now - 60 minutes - 500 minutes`?

zd-project: Ah I missed that. My bad. You're right.

srdo:
+1, thanks for the quick fix. Please squash and we can merge.

zd-project:
Squashed.

PR #2788, opened 2018-08-01
STORM-3170: Fixed bug to eliminate invalid file deletion

zd-project:
https://issues.apache.org/jira/browse/STORM-3170

### DIFF
--- a/storm-webapp/src/main/java/org/apache/storm/daemon/logviewer/utils/DirectoryCleaner.java
+++ b/storm-webapp/src/main/java/org/apache/storm/daemon/logviewer/utils/DirectoryCleaner.java
@@ -24,13 +24,14 @@ import java.nio.file.DirectoryStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.ArrayList;
-import java.util.Comparator;
+import java.util.HashSet;
 import java.util.List;
 import java.util.PriorityQueue;
 import java.util.Set;
 import java.util.Stack;
 import java.util.regex.Pattern;
 
+import org.apache.storm.utils.Utils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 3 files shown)


zd-project:
This will affect #2754

storm-webapp/src/test/java/org/apache/storm/daemon/logviewer/testsupport/MockRemovableFileBuilder.java
```diff
@@ -0,0 +1,29 @@
...
+
+import org.mockito.Mockito;
+
+import java.io.File;
+
+public class MockRemovableFileBuilder extends MockFileBuilder {
+    @Override
+    public File build() {
+        File mockFile = super.build();
+        Mockito.when(mockFile.delete()).thenReturn(true);
```
srdo: This stubbing is a little weird. It'll make the files pretend to be deleteable, but once you delete the file it'll still say it exists, and calling delete again will still return true.

Is the reason we're mocking File rather than creating temporary files for performance reasons?

zd-project: I'm simply leveraging what's already out there. Maybe it's easier to control the behavior of a file in this way? I'll look into the implementation to see if I can alter behavior after calling `#delete`

storm-webapp/src/main/java/org/apache/storm/daemon/logviewer/utils/DirectoryCleaner.java
```diff
@@ -133,10 +123,15 @@ public int compare(File f1, File f2) {
...
                 File file = stack.pop();
-                toDeleteSize -= file.length();
-                LOG.info("Delete file: {}, size: {}, lastModified: {}", file.getCanonicalPath(), file.length(), file.lastModified());
-                file.delete();
-                deletedFiles++;
+                final String canonicalPath = file.getCanonicalPath();
+                final long fileSize = file.length();
+                final long lastModified = file.lastModified();
+                //Original implementation doesn't actually check if delete succeeded or not.
+                if (file.delete()) {
```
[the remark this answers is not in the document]
zd-project: I'm wondering if we should use forcedelete here directly?

storm-webapp/src/test/java/org/apache/storm/daemon/logviewer/testsupport/MockRemovableFileBuilder.java
revans2: @zd-project you can control it more fully, but it gets a little more complicated.  Not a big deal but if it is causing problems it can be done.

srdo:
+1, thanks for fixing the issue with nondeleteable files. The solution looks good. Please squash and we can merge.

storm-webapp/src/test/java/org/apache/storm/daemon/logviewer/testsupport/MockRemovableFileBuilder.java
zd-project: Should I still concern about this issue then? @revans2 @srdo

srdo: No, it's just me nitpicking. We can always replace it if it becomes an issue to use in tests.

zd-project: Okay I'll just file a minor jira then

revans2:
Still +1

PR #2790, opened 2018-08-03
STORM-3175 - Allow usage of custom Callback.

dfdemar:
[STORM-3175](https://issues.apache.org/jira/browse/STORM-3175)

### DIFF
--- a/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/bolt/KafkaBolt.java
+++ b/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/bolt/KafkaBolt.java
@@ -24,6 +24,7 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
 import org.apache.kafka.clients.producer.Callback;
 import org.apache.kafka.clients.producer.KafkaProducer;
+import org.apache.kafka.clients.producer.Producer;
 import org.apache.kafka.clients.producer.ProducerRecord;
 import org.apache.kafka.clients.producer.RecordMetadata;
 import org.apache.storm.kafka.bolt.mapper.FieldNameBasedTupleToKafkaMapper;
@@ -58,10 +59,11 @@ public class KafkaBolt<K, V> extends BaseTickTupleAwareRichBolt {
 
     public static final String TOPIC = "topic";
 
-    private KafkaProducer<K, V> producer;
+    private Producer<K, V> producer;
     private OutputCollector collector;
     private TupleToKafkaMapper<K,V> mapper;
     private KafkaTopicSelector topicSelector;
+    private PreparableCallback providedCallback;
     private Properties boltSpecifiedProperties = new Properties();
     /**
      * {@see KafkaBolt#setFireAndForget(boolean)} for more details on this. 
@@ -98,6 +100,16 @@ public class KafkaBolt<K, V> extends BaseTickTupleAwareRichBolt {
         return this;
     }
 
+    /**
+     * Sets a user defined callback for use with the KafkaProducer.
+     * @param producerCallback user defined callback
+     * @return this
+     */
+    public KafkaBolt<K, V> withProducerCallback(PreparableCallback producerCallback) {
+        this.providedCallback = producerCallback;
+        return this;
+    }
+
     @Override
     public void prepare(Map<String, Object> topoConf, TopologyContext context, OutputCollector collector) {
         LOG.info("Preparing bolt with configuration {}", this);
@@ -118,6 +130,10 @@ public class KafkaBolt<K, V> extends BaseTickTupleAwareRichBolt {
             }
         }
 
+        if (providedCallback != null) {
+            providedCallback.prepare(topoConf, context);
+        }
+
         producer = mkProducer(boltSpecifiedProperties);
         this.collector = collector;
     }
@@ -125,8 +141,30 @@ public class KafkaBolt<K, V> extends BaseTickTupleAwareRichBolt {
     /**
      * Intended to be overridden for tests.  Make the producer with the given props
      */
-    protected KafkaProducer<K, V> mkProducer(Properties props) {
+    protected Producer<K, V> mkProducer(Properties props) {
         return new KafkaProducer<>(props);
+    }
+
+    /**
+     * Creates the Callback to send to the Producer. Using this Callback will also execute
+     * the user defined Callback, if provided.
+     */
+    private Callback createProducerCallback(final Tuple input) {
+        return (ignored, e) -> {
+            synchronized (collector) {
+                if (e != null) {
+                    collector.reportError(e);
+                    collector.fail(input);
+                } else {
+                    collector.ack(input);
+                }
+
+                // User defined Callback
+                if (providedCallback != null) {
+                    providedCallback.onCompletion(ignored, e);
+                }
+            }
+        };
     }
 
     @Override
@@ -142,21 +180,11 @@ public class KafkaBolt<K, V> extends BaseTickTupleAwareRichBolt {
                 Callback callback = null;
 
                 if (!fireAndForget && async) {
-                    callback = new Callback() {
-                        @Override
-                        public void onCompletion(RecordMetadata ignored, Exception e) {
-                            synchronized (collector) {
-                                if (e != null) {
-                                    collector.reportError(e);
-                                    collector.fail(input);
-                                } else {
-                                    collector.ack(input);
-                                }
-                            }
-                        }
-                    };
+                    callback = createProducerCallback(input);
+                } else if (providedCallback != null) {
+                    callback = providedCallback;
                 }
-                Future<RecordMetadata> result = producer.send(new ProducerRecord<K, V>(topic, key, message), callback);
+                Future<RecordMetadata> result = producer.send(new ProducerRecord<>(topic, key, message), callback);
                 if (!async) {
                     try {
                         result.get();
@@ -207,7 +235,7 @@ public class KafkaBolt<K, V> extends BaseTickTupleAwareRichBolt {
     public void setAsync(boolean async) {
         this.async = async;
     }
-    
+
     @Override
     public String toString() {
         return "KafkaBolt: {mapper: " + mapper 
--- a/external/storm-kafka-client/src/test/java/org/apache/storm/kafka/bolt/KafkaBoltTest.java
+++ b/external/storm-kafka-client/src/test/java/org/apache/storm/kafka/bolt/KafkaBoltTest.java
@@ -17,73 +17,147 @@
  */
 package org.apache.storm.kafka.bolt;
 
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
 import static org.mockito.ArgumentMatchers.any;
-import static org.mockito.ArgumentMatchers.argThat;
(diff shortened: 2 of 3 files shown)


external/storm-kafka-client/src/main/java/org/apache/storm/kafka/bolt/KafkaBolt.java
```diff
@@ -129,6 +140,27 @@ public void prepare(Map<String, Object> topoConf, TopologyContext context, Outpu
...
+            synchronized (collector) {
+                if (e != null) {
+                    collector.reportError(e);
+                    collector.fail(input);
+                } else {
+                    collector.ack(input);
+                }
+
+                // User defined Callback
+                if (providedCallback != null) {
```
dfdemar: The original behavior of the Callback is pretty important so we probably don't want a user-defined Callback to completely erase the original behavior. If the intent _**is**_ to totally redefine the Callback then one can simply subclass `KafkaBolt` and explicitly override `mkProducerCallback()`.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/bolt/KafkaBolt.java
```diff
@@ -62,6 +62,7 @@
     private OutputCollector collector;
     private TupleToKafkaMapper<K,V> mapper;
     private KafkaTopicSelector topicSelector;
+    private Callback providedCallback;
```
srdo: I think we need to define our own interface for this. Callback isn't serializable, so Storm won't be able to transfer it from the Nimbus submitter to the worker JVM.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/bolt/KafkaBolt.java
```diff
@@ -129,6 +140,27 @@ public void prepare(Map<String, Object> topoConf, TopologyContext context, Outpu
         return new KafkaProducer<>(props);
     }
 
+    /**
+     * Make the producer Callback. Using this Callback will also execute the user defined Callback, if provided.
+     */
+    protected Callback mkProducerCallback(final Tuple input) {
```
srdo: I'm not sure we want to make this protected. The default makes use of some private fields (e.g. collector), so subclasses won't be able to implement this properly, and I'm also not really understanding why someone would need to override this, since the code that is already here is pretty important to the bolt working correctly.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/bolt/KafkaBolt.java
srdo: Also I'm not sure exactly what someone implementing this interface would need, but maybe we should add a prepare method to the interface as well, so people who need some configuration or the topology context can get access? What do you think?

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/bolt/KafkaBoltTest.java
```diff
@@ -37,38 +39,48 @@
 import org.apache.storm.tuple.Tuple;
 import org.junit.Test;
 import org.mockito.ArgumentMatcher;
+import org.mockito.internal.util.MockUtil;
```
srdo: I would prefer not to use internal classes. I think https://static.javadoc.io/org.mockito/mockito-core/2.20.0/org/mockito/Mockito.html#mocking_details can do the same thing.

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/bolt/KafkaBoltTest.java
```diff
@@ -37,38 +39,48 @@
...
         bolt.withTopicSelector("MY_TOPIC");
-        
+
+        return bolt;
+    }
+
+    @SuppressWarnings({ "unchecked", "serial" })
+    @Test
+    public void testSimple() {
+        final KafkaProducer<String, String> producer = mock(KafkaProducer.class);
```
srdo: I'm wondering if we would be better off using https://kafka.apache.org/10/javadoc/org/apache/kafka/clients/producer/MockProducer.html? Our own stubbing ends up doing some weird things, e.g. returning null from `send`.

dfdemar: That's a possibility, though `KafkaBolt` has `KafkaProducer` as a private field so `MockProducer` can't be used. That could be fixed by changing that field to a `Producer` interface instead.

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/bolt/KafkaBoltTest.java
dfdemar: Nice, didn't know about that one.

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/bolt/KafkaBoltTest.java
srdo: That would be fine.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/bolt/KafkaBolt.java
dfdemar: I can change to `private`.

dfdemar: Fair point. I'll change it to `private`.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/bolt/KafkaBolt.java
dfdemar: What about something like this? Not sure if `OutputCollector` is necessary or not.

```java
public interface PreparedCallback extends Callback, Serializable {
    void prepare(Map<String, Object> topoConf, TopologyContext context, OutputCollector outputCollector);
}
```

So if the `PreparedCallback` is not null, then KafkaBolt would call `.prepare()` during `KafkaBolt.prepare()`.

The one concern I have is that it might be open for abuse by tempting people to use it to inject logic into the bolt preparation that's unrelated to Kafka publishing.

srdo: That looks good to me. I'm also not sure about the collector, but I'd maybe lean toward not including it. The callback won't have access to incoming tuples, so the only use would be if someone wanted the callback to emit a new unanchored tuple.

I wouldn't worry about people misusing the prepare method. They can just as easily override the bolt's prepare method.

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/bolt/KafkaBoltTest.java
```diff
@@ -17,73 +17,140 @@
...
+        LOG.info("GOT {} ->", arg);
+        LOG.info("{}, {}, {}", arg.topic(), arg.key(), arg.value());
+        assertThat(arg.topic(), is("MY_TOPIC"));
+        assertThat(arg.key(), is(key));
+        assertThat(arg.value(), is(value));
+
+        // Complete the send
+        producer.completeNext();
+        verify(collector).ack(testTuple);
+        verify(customCallback).onCompletion(any(), any());
```
srdo: Nit: Can we be more specific about the parameters here, or would that be hard to do?

dfdemar: Do you mean in regards to the `any()`? I'll see if I can make it a little better.

srdo: Yes. It's nitpicking though, so if it's not super easy don't worry about it. Either way, please squash to one commit when you want this merged.

dfdemar: No problem.

dfdemar: I was able to make it a little better. I went ahead and squashed the commits and it should be okay to merge now.

dfdemar:
@srdo Ready to merge!

uddhavarote:
Hey, thanks for this update. Without this update, one would end up (re)writing a bolt like KafkaBolt with similar functionality but different  `OnCompletion` function, like I am doing currently for extracting the `timestamp` (message's ingest time) from `RecordMetadata` and passing it to the next bolt with current `Tuple` fields. 

However, even with this update, I am afraid if I will be able to do the similar `OnCompletion` thing I am currently doing by rewriting the bolt.
The implemented `OnCompletion`  function looks  like ↓
```
@Override
public void onCompletion(RecordMetadata metadata, Exception e) {
  boolean hasNextBolt = appConf.getBooleanOrDefault(, true);
  synchronized (collector) {
    if (e != null) {
      collector.reportError(e);
    } else {
        if (hasNextBolt) {
          Long field1 = metadata.timestamp(); 
          Long field2 = input.getLongByField(...);
          String field3 = input.getStringByField(...);
          collector.emit( input, new Values(field1, field2, field3));
       }
        collector.ack(input);
    }
  }
}
```
Since, there is no reference to the`Tuple` object in the  `PreparableCallback` interface, how can something like above be achieved? 
If not achievable, is it a good idea to give reference to `Tuple` in the said interface allowing further new emits (in `OnCompletion()`) of tuples composed of new fields `field1`and fields from current tuple `field2, field3`?

srdo:
@uddhavarote I wasn't sure if anyone needed the ability to emit new tuples after writing to Kafka, because you can still do further processing by adding another topology (or a new spout to your current topology) that reads from your target topic. There are some minor drawbacks to doing more processing after writing to Kafka the way your snippet does it. Either you emit the new tuple unanchored and you won't get at-least-once processing for the part of the topology after the KafkaBolt, or you anchor it to the input tuple like you do in your snippet, and risk adding the same tuples to Kafka multiple times. If the bolt after the KafkaBolt fails the tuple, the entire topology is repeated, including the KafkaBolt.

I'm not opposed to adding the collector to `prepare` as well as giving you access to the input tuple in the callback, but I think you should be aware of the drawbacks of laying out your topology like this.


Just for illustration, what I'm proposing is this 
```
Data Source -> Your topology here -> KafkaBolt -> Topic A -> KafkaSpout -> Bolt handling metadata timestamp
```

rather than 
```
Data Source -> Your topology here -> KafkaBolt -> Topic A
                                               -> Bolt handling metadata timestamp
```

uddhavarote:
@srdo Thanks for the details. Yeah, I am aware of the drawbacks. But I think emitting this next tuple into a different stream, not `default` stream,  should not replay the tuples in case of failure. That way, we create disjoint streams after the Kafka bolt.

The proposed topology looks good, however, say access to `Topic A` is not available due to the authorization or only `RecordMetadata` is required, there is no point in reading the whole topic.  I think one would be better off emitting the next tuple in the same topology rather than writing a new topology to read the data from `RecordMetadata`.

I would like to suggest that the access to the `Tuple` be provided. along with a note about the consequences in such a case.

PR #2791, opened 2018-08-03
STORM-3176: KafkaSpout commit offset occurs CommitFailedException which leads to worker dead

wangzzu:
KafkaSpout use the commitAsync api of Consumer, if the interval time between the calling of the  consumer.poll() more than `max.poll.interval.ms` or the heartbeat of consumer timeout, that will occur CommitFailedException,  and then the worker will die, the log like this:

```
2018-07-31 19:19:03.341 o.a.s.util [ERROR] Async loop died!
org.apache.mtkafka.clients.consumer.CommitFailedException: Commit cannot be completed since the group has already rebalanced and assigned the partitions to another member. This means that the time between subsequent calls to poll() was longer th
an the configured max.poll.interval.ms, which typically implies that the poll loop is spending too much time message processing. You can address this either by increasing the session timeout or by reducing the maximum size of batches returned in
poll() with max.poll.records.
at org.apache.kafka.clients.consumer.internals.ConsumerCoordinator.sendOffsetCommitRequest(ConsumerCoordinator.java:698) ~[stormjar.jar:?]
at org.apache.kafka.clients.consumer.internals.ConsumerCoordinator.commitOffsetsSync(ConsumerCoordinator.java:577) ~[stormjar.jar:?]
at org.apache.kafka.clients.consumer.KafkaConsumer.commitSync(KafkaConsumer.java:1126) ~[stormjar.jar:?]
at org.apache.kafka.clients.consumer.KafkaConsumer.commitSync(KafkaConsumer.java:XXX) ~[stormjar.jar:?]
at org.apache.storm.kafka.spout.KafkaSpout.commitOffsetsForAckedTuples(KafkaSpout.java:430) ~[stormjar.jar:?]
at org.apache.storm.kafka.spout.KafkaSpout.nextTuple(KafkaSpout.java:264) ~[stormjar.jar:?]
at org.apache.storm.daemon.executor$fn__10936$fn__10951$fn__10982.invoke(executor.clj:647) ~[storm-core-1.1.2-mt001.jar:?]
at org.apache.storm.util$async_loop$fn__553.invoke(util.clj:484) [storm-core-1.1.2-mt001.jar:?]
at java.lang.Thread.run(Thread.java:745) [?:1.8.0_112]
2018-07-31 19:19:03.342 o.a.s.d.executor [ERROR]
```

I find it will catch the Exception in the auto-commit mode of consumer, the source code is:

```java
private void maybeAutoCommitOffsetsSync(long timeoutMs) {
    if (autoCommitEnabled) {
        Map<TopicPartition, OffsetAndMetadata> allConsumedOffsets = subscriptions.allConsumed();
        try {
            log.debug("Sending synchronous auto-commit of offsets {} for group {}", allConsumedOffsets, groupId);
            if (!commitOffsetsSync(allConsumedOffsets, timeoutMs))
                log.debug("Auto-commit of offsets {} for group {} timed out before completion",
                        allConsumedOffsets, groupId);
        } catch (WakeupException | InterruptException e) {
            log.debug("Auto-commit of offsets {} for group {} was interrupted before completion",
                    allConsumedOffsets, groupId);
            // rethrow wakeups since they are triggered by the user
            throw e;
        } catch (Exception e) {
            // consistent with async auto-commit failures, we do not propagate the exception
            log.warn("Auto-commit of offsets {} failed for group {}: {}", allConsumedOffsets, groupId,
                    e.getMessage());
        }
    }
}
```

I think KafkaSpout should do like this, catch the Exception avoid to worker die. And when the msg ack failed, Spout should judge the offset of the msgID is larger than the last commit offset(Spout can guarantee that these msgs which offset less than the last commit offset are all ack), if not, the msg should not retry.

### DIFF
--- a/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
+++ b/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
@@ -37,6 +37,7 @@ import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 import org.apache.commons.lang.Validate;
+import org.apache.kafka.clients.consumer.CommitFailedException;
 import org.apache.kafka.clients.consumer.Consumer;
 import org.apache.kafka.clients.consumer.ConsumerRebalanceListener;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
@@ -282,7 +283,12 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
                 } else if (kafkaSpoutConfig.getProcessingGuarantee() == ProcessingGuarantee.NO_GUARANTEE) {
                     Map<TopicPartition, OffsetAndMetadata> offsetsToCommit =
                         createFetchedOffsetsMetadata(consumer.assignment());
-                    consumer.commitAsync(offsetsToCommit, null);
+                    try {
+                        consumer.commitAsync(offsetsToCommit, null);
+                    }catch (CommitFailedException e) {
+                        // catch the CommitFailedException
+                        LOG.warn("Commit offsets {} failed for group {}: {}", offsetsToCommit.toString());
+                    }
                     LOG.debug("Committed offsets {} to Kafka", offsetsToCommit);
                 }
             }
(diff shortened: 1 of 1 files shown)


srdo:
@wangzzu Hi, I'm not sure this should be happening. We're not using the `KafkaConsumer.subscribe` API anymore, so Kafka shouldn't be managing the partition assignment.

Can you confirm that you're using storm-kafka-client 1.2.0? Could you please post your KafkaSpoutConfig?

wangzzu:
@srdo sorry, the version I used is 1.1.2. This problem has already fixed in 1.2.0.

srdo:
@wangzzu Yes, but it should also have been fixed in 1.1.2. Please make sure you're not using one of the deprecated Subscription subtypes, e.g. https://github.com/apache/storm/blob/v1.1.2/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/NamedSubscription.java

wangzzu:
@srdo I found that we are using the deprecated Subscription subtypes actually, thx.

PR #2792, opened 2018-08-04
Add the getName() method in order to obtain the applied label to the Trident stream.

alex-87:
Being able to get the applied label of a Trident stream could be useful in case of complex Trident topology design. The getter returns the `String` corresponding to the `_name` attribute of the `Stream` object.


srdo:
The changes look good to me. Please raise an issue at https://issues.apache.org/jira and rename this PR and the commit message to contain the issue number.

HeartSaVioR:
@alex-87 Sorry I'm late on party. Could you follow up @srdo guide?

PR #2794, opened 2018-08-06
STORM-3180 Total executors in Cluster Summary in main UI page is not …

HeartSaVioR:
…exposed even a topology is running

* rename the field of /cluster/summary output: `totalExecutors` to `executorsTotal`

Please compare with doc regarding the output format of API:
https://github.com/apache/storm/blob/master/docs/STORM-UI-REST-API.md#apiv1clustersummary-get

### DIFF
--- a/storm-webapp/src/main/java/org/apache/storm/daemon/ui/UIHelpers.java
+++ b/storm-webapp/src/main/java/org/apache/storm/daemon/ui/UIHelpers.java
@@ -571,7 +571,7 @@ public class UIHelpers {
         result.put("slotsTotal", totalSlots);
         result.put("slotsFree", totalSlots - usedSlots);
         result.put("tasksTotal", totalTasks);
-        result.put("totalExecutors", totalExecutors);
+        result.put("executorsTotal", totalExecutors);
 
         result.put("totalMem", supervisorTotalMemory);
         result.put("totalCpu", supervisorTotalCpu);


srdo:
Grepping for "totalExecutors" also pops up a couple of other locations in storm-webapp.

```
./storm-webapp/src/main/java/org/apache/storm/daemon/ui/UIHelpers.java:536:        int totalExecutors =
./storm-webapp/src/main/java/org/apache/storm/daemon/ui/UIHelpers.java:574:        result.put("totalExecutors", totalExecutors);
./storm-webapp/src/main/java/org/apache/storm/daemon/ui/UIHelpers.java:631:        result.put("totalExecutors", ownerResourceSummary.get_total_executors());
./storm-webapp/src/main/java/org/apache/storm/daemon/ui/WEB-INF/js/script.js:541:        data: 'totalExecutors',
./storm-webapp/src/main/java/org/apache/storm/daemon/ui/WEB-INF/owner.html:148:                    //owner,totalTopologies,totalTasks,totalExecutors,totalWorkers
./storm-webapp/src/main/java/org/apache/storm/daemon/ui/WEB-INF/templates/owner-page-template.html:51:            <td>{{totalExecutors}}</td>
```

Do any of the others also need to be changed?

HeartSaVioR:
@srdo No, unfortunately we don't apply consistent naming so `totalExecutors` is right for others. Renaming them would break clients which rely on the REST API, so I'm not sure we would be better to fix them though I think it is ideal to fix.

srdo:
Okay, just wanted to be sure it wasn't an oversight. +1.

PR #2795, opened 2018-08-06
STORM-3182: Removing superfluous topo id check from owner resource AP…

govind-menon:
…I request


govind-menon:
@HeartSaVioR Fix for 3182. Working on the others.

PR #2799, opened 2018-08-09
STORM-3183: Fix for visualization on Storm API

govind-menon:


HeartSaVioR:
screen shot 2018-08-10 at 8 13 44 pm

Looks like there is still some missed spot: could you check with this?

govind-menon:
@HeartSaVioR Fixed the component page.

PR #2798, opened 2018-08-08
STORM-3184: Mask the plaintext passwords from the logs

arunmahadevan:
Introduce a `Password` config annotation and use it to mark configs that are
sensitive and mask the values while logging.

PR #2801, opened 2018-08-11
STORM-3184: Mask the plaintext passwords from the logs

arunmahadevan:
Introduce a `Password` config annotation and use it to mark configs that are
sensitive and mask the values while logging.

Master port of https://github.com/apache/storm/pull/2798

### DIFF
--- a/storm-client/src/jvm/org/apache/storm/utils/ConfigUtils.java
+++ b/storm-client/src/jvm/org/apache/storm/utils/ConfigUtils.java
@@ -14,6 +14,8 @@ package org.apache.storm.utils;
 
 import java.io.File;
 import java.io.IOException;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Field;
 import java.net.URLEncoder;
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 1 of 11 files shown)


storm-client/src/jvm/org/apache/storm/utils/ConfigUtils.java
```diff
@@ -52,6 +79,16 @@ public static ConfigUtils setInstance(ConfigUtils u) {
         return oldInstance;
     }
 
+    public static Map<String, Object> maskPasswords(final Map<String, Object> conf) {
+        Maps.EntryTransformer<String, Object, Object> maskPasswords =
```
HeartSaVioR: Could we replace this with Java streams API and remove Guava dependency?

HeartSaVioR:
@arunmahadevan Looks like we don't need to introduce Guava for master branch. I understand this is annoying to apply Java 7 / Java 8 for each branch, but let's keep this approach until we ship Storm 2.0.0.

storm-client/src/jvm/org/apache/storm/utils/ConfigUtils.java
arunmahadevan: The problem is that the map is traversed and copied even if we don't intend to log the conf and I was trying to avoid it using the Guava transformer. It may not be a major issue now since the conf is logged only once at the beginning, but wanted to avoid any potential future misuse. 

I could not figure out a way to do the same with java 8 Streams API. I have refactored it to return a wrapped object which will be evaluated only when its required to log. Please take a look and let me know if it makes sense or if theres a better way.

HeartSaVioR: @arunmahadevan 
I also think this approach can replace `Utils.radactValue` via adding `@password` to STORM_ZOOKEEPER_TOPOLOGY_AUTH_PAYLOAD.
What do you think? Looks like we don't need to have both approach.

PR #2806, opened 2018-08-16
[STORM-3198] Topology submitters should be able to supply log4j2 conf

jacobtolar:
This adds a new config setting, `topology.logging.config`, that allows a topology submitter to specify an additional log4j2 config to be used via composite configuration.

When the worker starts the `-Dlog4j.configurationFile` option is modified to include the additional file if requested.

To test, I submitted a topology (FastWordCountTopology):

```
bin/storm jar -c topology.logging.config=classpath:resources/topology_log4j2.xml ../../../../../examples/storm-starter/target/storm-starter-*.jar org.apache.storm.starter.FastWordCountTopology fwc
```

topology_log4j2.xml:
```
<configuration monitorInterval="60" shutdownHook="disable">
<loggers>
    <root>
       <RegexFilter regex=".*Preparing.*" onMatch="DENY" onMismatch="ACCEPT" />
    </root>
    <Logger name="org.apache.storm.metric.LoggingMetricsConsumer">
       <RegexFilter regex=".*Preparing.*" onMatch="DENY" onMismatch="ACCEPT" />
    </Logger>
    <Logger name="STDERR">
       <RegexFilter regex=".*Preparing.*" onMatch="DENY" onMismatch="ACCEPT" />
    </Logger>
    <Logger name="STDOUT">
       <RegexFilter regex=".*Preparing.*" onMatch="DENY" onMismatch="ACCEPT" />
    </Logger>
</loggers>
</configuration>
```

With this child configuration added, the messages about [preparing bolts](https://github.com/apache/storm/blob/26d2f955251c0fa56520f6fe08cb35ef5171f321/storm-client/src/jvm/org/apache/storm/executor/bolt/BoltExecutor.java#L109) no longer appear in the worker log.

I also tested changing simply changing the log level in the child log4j2 config and that also works as expected.


pom.xml
```diff
@@ -278,7 +278,7 @@
         <netty.version>4.1.25.Final</netty.version>
         <sysout-over-slf4j.version>1.0.2</sysout-over-slf4j.version>
         <log4j-over-slf4j.version>1.6.6</log4j-over-slf4j.version>
-        <log4j.version>2.8.2</log4j.version>
+        <log4j.version>2.11.0</log4j.version>
```
jacobtolar: Note: The update to log4j2-2.11.0 is not strictly necessary. 2.8.2 supports composite configuration. However, there was [a bug](https://issues.apache.org/jira/browse/LOG4J2-2123) in merging filters from child configurations that I got fixed in 2.11.0. If a child configuration added filters and there were none in the parent, the child filter would be ignored.

I would like to add some filters in my topology, so, bumping the version to the one with the fix.

storm-server/src/main/java/org/apache/storm/daemon/supervisor/BasicContainer.java
```diff
@@ -622,6 +620,12 @@ protected String javaCmd(String cmd) {
         List<String> classPathParams = getClassPathParams(stormRoot, topoVersion);
         List<String> commonParams = getCommonParams();
 
+        String log4jConfigurationFile = getWorkerLoggingConfigFile();
```
srdo: Nit: Can this go in `getCommonParams`?

storm-client/src/jvm/org/apache/storm/Config.java
```diff
@@ -705,6 +705,13 @@
      */
     @isString(acceptedValues = { "S0", "S1", "S2", "S3" })
     public static final String TOPOLOGY_LOGGING_SENSITIVITY = "topology.logging.sensitivity";
+    /**
+     * Log file the user can use to configure Log4j2.
```
srdo: Nit: Consider mentioning that the configuration file is applied in addition to the regular config file, using these rules https://logging.apache.org/log4j/2.x/manual/configuration.html#CompositeConfiguration.

pom.xml
srdo: Unless we know of a reason not to, we could just bump to the latest version IMO.

storm-server/src/main/java/org/apache/storm/daemon/supervisor/BasicContainer.java
jacobtolar: No, it can't -- The xml file might be in the topology jar uploaded to the nimbus. That file is not on the classpath if the LogWriter feature is used. 

So I guess it could, but getCommonParams would have to take a parameter and get invoked twice. Right now it's just invoked once and the output is used for both LogWriter and Worker jvm startup scripts.

srdo: Thanks for explaining.

jacobtolar:
Yeah, sorry it took me a while to get to fixing it. Hopefully that does it.

srdo:
+1. Please squash to one commit @jacobtolar.

jacobtolar:
rebased to latest master and squashed into one commit

PR #2807, opened 2018-08-16
STORM-3199: Remove metrics-ganglia due to LGPL dependency

srdo:
https://issues.apache.org/jira/browse/STORM-3199


srdo:
Rerunning test, it had some trouble pulling dependencies from Maven

srdo:
Thanks for reviewing @HeartSaVioR. It can't, I'll open another PR against 1.x shortly

srdo:
Raised https://github.com/apache/storm/pull/2809.

On 1.x the LGPL dependency is excluded so people can install it manually. If we want to keep metrics-ganglia, we could probably make that change instead of removing the dependency entirely.

I'm not sure it makes sense for us to keep metrics-ganglia though, since the Metrics library maintainers have decided to drop it. We'll be unable to upgrade past Metrics 3.2.x if we decide to keep the dependency.

What do you think @HeartSaVioR?

HeartSaVioR:
@srdo 
I guess it worths to discuss in user/dev mailing list, cause I don't know who is using Ganglia Reporter. Actually I feel Metrics V2 may not be used actively, so IMHO the risk is small, but we may want to do some kinds of poll before removing anyway. If someone is using the Ganglia Reporter for 1.x, we can deprecate it for 1.x and remove in 2.0.

srdo:
Posted a discuss thread on the mailing list.

PR #2808, opened 2018-08-17
[STORM-3131] Support hostname-substitution for blobstore.hdfs.principal

Ethanlm:
https://issues.apache.org/jira/browse/STORM-3131

 With this feature, we can set 
```
blobstore.hdfs.principal: <headless-user>/HOSTNAME@domain
```
on nodes(nimbi, supervisors) the `HOSTNAME` will be replaced with the actual hostname on that host.

### DIFF
--- a/storm-client/src/jvm/org/apache/storm/Config.java
+++ b/storm-client/src/jvm/org/apache/storm/Config.java
@@ -19,6 +19,7 @@
 package org.apache.storm;
 
 import com.esotericsoftware.kryo.Serializer;
+import java.net.UnknownHostException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
@@ -28,6 +29,7 @@ import org.apache.storm.metric.IEventLogger;
 import org.apache.storm.policy.IWaitStrategy;
 import org.apache.storm.serialization.IKryoDecorator;
 import org.apache.storm.serialization.IKryoFactory;
+import org.apache.storm.utils.Utils;
 import org.apache.storm.validation.ConfigValidation;
 import org.apache.storm.validation.ConfigValidation.EventLoggerRegistryValidator;
 import org.apache.storm.validation.ConfigValidation.ListOfListOfStringValidator;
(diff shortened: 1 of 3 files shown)


storm-client/src/jvm/org/apache/storm/Config.java
```diff
@@ -1884,4 +1888,11 @@ public void setTopologyStrategy(String strategy) {
         this.put(Config.TOPOLOGY_SCHEDULER_STRATEGY, strategy);
     }
 
+    public static String getBlobstoreHDFSPrincipal(Map conf) throws UnknownHostException {
+        String principal = (String)conf.get(Config.BLOBSTORE_HDFS_PRINCIPAL);
+        if (principal != null) {
+            principal = principal.replace("HOSTNAME", Utils.localHostname());
```
HeartSaVioR: Hadoop picks `_HOST` as placeholder and it would have less change to be conflicted. So if we have no better reason to use alternative, using `_HOST` would be better to me.

Also I would feel more safer if we could try to parse principal and only replace placeholder when host part in principal is just a placeholder.  So `_HOST` in `yarn/_HOST@EXAMPLE.COM` can be substituted, but both `_HOST` in `yarn/_HOST2@EXAMPLE.COM` and `_HOST` in `_HOST/host1@EXAMPLE.COM` should not be substituted.

@revans2 Could you provide some suggestions?

revans2: I agree _HOST is probably preferable, as it is less likely to result in a collision by accident, and it follows with what Hadoop is doing.

revans2:
@Ethanlm I know you have a lot on your plate right now, if you want me to make the change from HOSTNAME to _HOST I can.

Ethanlm:
Sorry forgot this PR. I will do it soon. Thanks @revans2

Ethanlm:
The failed tests should be unrelated.

Ethanlm:
Could you please re-check? Thanks! @HeartSaVioR @revans2

PR #2811, opened 2018-08-21
STORM-3184: Replace the usage of redact-value with ConfigUtils.maskPasswords

arunmahadevan:
The topology submission can fail since redact-value expects a clojure map. We don't need redact-value, it can be replaced with ConfigUtils.maskPasswords. This is already done in master.


arunmahadevan:
cc @HeartSaVioR

HeartSaVioR:
Thanks @arunmahadevan , merged follow-up patch to 1.x-branch.

PR #2814, opened 2018-08-28
[STORM-3207] Fix Sasl Plugin to use WorkerToken

kishorvpatil:
The `doAsUser` is null for DRPCClient. If WorkerToken is found, it should use it.
Also, setting on `addServerDefinition` to `localhost` literal string is incorrect on server side.


storm-client/src/jvm/org/apache/storm/security/auth/kerberos/KerberosSaslTransportPlugin.java
```diff
@@ -119,7 +119,7 @@ private Login mkLogin() throws IOException {
     @Override
     public TTransport connect(TTransport transport, String serverHost, String asUser) throws IOException, TTransportException {
         WorkerToken token = WorkerTokenClientCallbackHandler.findWorkerTokenInSubject(type);
-        if (token != null && asUser != null) {
```
revans2: The reason I had `asUser != null`, is because tokens done support user impersonation, but I got it wrong, and it should have been `asUser == null`.  But this should be fine.  If it becomes an issue we can add in support for asUser if needed

revans2:
Thanks for finding and fixing these bugs.

PR #2816, opened 2018-08-29
STORM-3208 fix worker kill NPE

agresch:


danny0405:
@agresch 
So in what case `_localState.getApprovedWorkers()` will return null.

agresch:
@danny0405 - I'm not really knowledgable about any of this particular code, but the LocalState API clearly allows returning null, rather than throwing any exception.  I'm just trying to cover the callstack I provided in the JIRA.  If we feel this shouldn't happen, maybe LocalState getApprovedWorkers() should throw an exception if there are no approved workers?

revans2:
@danny0405 @agresch this should not happen under normal conditions.  If it is happening under normal conditions we need to fix some things.  The approved workers holds the list of workers that the supervisor believes it is currently running.  The only times that it would become null would be when there is an issue with the local state, like it was wiped clean too quickly by an over aggressive script meant to fix a broken node, or by a test framework that possibly got the timing wrong.

If this is happening regularly outside of odd situations like I said, then there may be a race condition that we missed and we need to look into it more deeply.

agresch:
@revans2 - this is a normal run of our integration test pipeline.  I don't have much context, just noticed this error in the logs and filed the JIRA.

revans2:
@agresch could we at least put in some logging when it is null?

agresch:
@revans2 - added a warning.

PR #2827, opened 2018-09-12
STORM-3221: Fixing utilization in clusterSummary API call

govind-menon:


storm-webapp/src/main/java/org/apache/storm/daemon/ui/UIHelpers.java
```diff
@@ -582,11 +582,9 @@ public static Response makeStandardResponse(
         result.put("schedulerDisplayResource",
                 conf.get(DaemonConfig.SCHEDULER_DISPLAY_RESOURCE));
         result.put("memAssignedPercentUtil", supervisorTotalMemory > 0
-                ? String.valueOf((supervisorTotalMemory - supervisorUsedMemory)  * 100.0
-                / supervisorTotalMemory) : "0.0");
+                ? String.valueOf((supervisorUsedMemory  * 100.0) / supervisorTotalMemory) : "0.0");
         result.put("cpuAssignedPercentUtil", supervisorTotalCpu > 0
-                ? String.valueOf((supervisorTotalCpu - supervisorUsedCpu) * 100.0
-                / supervisorTotalCpu) : "0.0");
+                ? String.valueOf((supervisorUsedCpu * 100.0) / supervisorTotalCpu) : "0.0");
```
revans2: nit: Can we fix the rounding while we are at it?

PR #2828, opened 2018-09-12
STORM-3217: Fixing getComponentPage API call

govind-menon:


revans2:
missing metrics

This might need to be a separate JIRA, but there are a lot of missing metrics on the topology page too.

PR #2829, opened 2018-09-12
STORM-3222: Fix KafkaSpout internals to use LinkedList instead of ArrayList

arunmahadevan:
KafkaSpout internally maintains a waitingToEmit list per topic partition and keeps removing the first item to emit during each nextTuple. The implementation uses an ArrayList which results in un-necessary traversal and copy for each tuple.

### DIFF
--- a/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
+++ b/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
@@ -30,6 +30,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -350,7 +351,7 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
 
     private void setWaitingToEmit(ConsumerRecords<K, V> consumerRecords) {
         for (TopicPartition tp : consumerRecords.partitions()) {
-            waitingToEmit.put(tp, new ArrayList<>(consumerRecords.records(tp)));
+            waitingToEmit.put(tp, new LinkedList<>(consumerRecords.records(tp)));
         }
     }
 
@@ -536,7 +537,7 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
                         //Discard the pending records that are already committed
                         waitingToEmit.put(tp, waitingToEmitForTp.stream()
                             .filter(record -> record.offset() >= committedOffset)
-                            .collect(Collectors.toList()));
+                            .collect(Collectors.toCollection(LinkedList::new)));
                     }
                 }
 


arunmahadevan:
Also I am not sure why the nextTuple only emits a single tuple wheres ideally it should emit whatever it can emit in a single nextTuple call which is more efficient. However the logic appears too complicated to refactor.

@srdo @HeartSaVioR

hmcl:
@arunmahadevan what's the motivation to change this to LinkedList?

nextTuple emits only a single tuple because that's the contract of the method nextTuple, which must be honored. This was thoroughly discussed in the patch with the initial code implementation.

arunmahadevan:
>what's the motivation to change this to LinkedList?

Its mentioned in the description. Heres the relevant code for some more details - https://github.com/apache/storm/blob/master/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java#L422

>nextTuple emits only a single tuple because that's the contract of the method nextTuple, which must be honored. This was thoroughly discussed in the patch with the initial code implementation.

It can emit one or more and ideally should emit whatever it has to emit at that point.
https://github.com/apache/storm/blob/master/storm-client/src/jvm/org/apache/storm/spout/ISpout.java#L72

hmcl:
@revans2 if I am not mistaken I recall that when you reviewed the initial PR you suggested that a call to nextTuple should send only one tuple/record. Can you please refresh my mind on the reasons why? If I am confusing with a different patch, please disregard this message. Thanks.

srdo:
@arunmahadevan I think the reason for only emitting one tuple at a time was to let Storm enforce the max.spout.pending limit. Maybe it would be better to accept some overflow in order to get the spout running faster, since people can still control how large the overflow can get by setting the consumer's max.poll.records (which I don't believe existed at the time the spout was originally written). I wanted to point to the PR where this was discussed, but I can't find it.

If you want to try emitting as many tuples as possible, it should be possible by removing this break: https://github.com/apache/storm/pull/2829/files#diff-7d7cbc8f5444fa7ada7962033fc31c5eR425

+1 for the current changes.

HeartSaVioR:
@srdo is right. The reason is to enforce "max.spout.pending.limit", but the concept is brought when backpressure is not in place. IMHO, the concept is the thing we should drop whenever backpressure is stabilized.

While I'm wondering backpressure would work correctly without any critical performance hit on 1.x (we disabled by default AFAIK), STORM-2306 renewed the mechanism of backpressure which we may be OK to rely on backpressure. If my memory is right, providing "max.spout.pending" was still the way to optimize even with STORM-2306, but even without this it worked correctly with sane performance.

At least for Storm 2.0 we could try out relying only backpressure and drop the concept if we are happy with the result.

revans2:
The reason we want to send a single tuple is because of how we do flow control in the spouts.  If you want to send more it can be more efficient, but you risk going over the max.spout.pending amount that is set by the user, and in 2.x where an emit can block, you risk having the spout block as well.

arunmahadevan:
@revans2 thanks for merging. Raised https://github.com/apache/storm/pull/2835 for 1.x-branch.

>If you want to send more it can be more efficient, but you risk going over the max.spout.pending amount that is set by the user, and in 2.x where an emit can block, you risk having the spout block as well.

If the emit can block in 2.x it might be an issue to address because its not in spout contract to necessarily emit only one record. I thought it could overflow and the max.spout.pending is not that relevant with the back pressure changes.

HeartSaVioR:
If my understanding is right, there's pendingEmits (unbounded) which comes into play when it can't push tuple immediately, so emit should not block, nextTuple should not block as well. If emit can block in spout it could be pretty much a big risk (ack/fail is not handled as well and tuples will fail into timeout, and it may raise backpressure again).

@roshannaik Could you confirm?

revans2:
Sorry my bad, there is an overflow.

https://github.com/apache/storm/blob/4605ae0a34858309171e726e1924b9a37695c977/storm-client/src/jvm/org/apache/storm/executor/ExecutorTransfer.java#L94-L111

The pause happens outside of the spouts in waiting for the queue to not be full any more just like before with the spouts.

roshannaik:
@HeartSaVioR  

1) pendingEmitsQ prevents nextTuple() from blocking when downstream queue is full.   It holds the overflow emits  (one or more) that occurred within a **single** nextTuple() invocation.  If Spout executor notices this Q is not empty, it will try to process any ACKs before entering wait strategy. Purpose of this Q is to prevent deadlock under BP. 

2) **topology.max.spout.pending**  Looked like a candidate for elimination given that STORM-2306's BP. Although it is not absolutely necessary to use it in 2.0.. during benchmarking I noticed that it could sometimes have sizable impact on performance. Why ? it remains a mystery. It is now a perf related tunable (in ACK mode).. and not really a BP mechanism.

roshannaik:
@arunmahadevan ..  need to be careful when that we are not doing too many emits in a single nextTuple()... it  can cause a OOM situation by flooding the pendingQ. 

Worth checking if there is any measurable gain in combining many emits into  a single nextTuple() for KafkaSpout.

srdo:
Even if we change the spout to emit all the fetched records in one `nextTuple` call, the number and size of records returned in a fetch is limited by the `max.poll.records` KafkaConsumer setting (500 by default), as well as the `fetch.max.bytes` setting (50MB by default).

PR #2832, opened 2018-09-14
STORM-3205 Optimization in TuplImpl

roshannaik:
See [Jira](https://issues.apache.org/jira/browse/STORM-3205) description for measurements and details


roshannaik:
travis failure seems unrelated.

PR #2833, opened 2018-09-14
STORM-3225: Use MediaType for check

revans2:
Full disclosure I am really just pushing back some changes that @govind-menon did, but was missed when the majority of UI changes going to java were put in open source.


storm-webapp/src/main/java/org/apache/storm/daemon/ui/filters/AuthorizedUserFilter.java
```diff
@@ -79,7 +80,7 @@
     public static Response makeResponse(Exception ex, ContainerRequestContext request, int statusCode) {
         String callback = null;
 
-        if (request.getMediaType().toString().contains("application/json")) {
+        if (request.getMediaType() != null && request.getMediaType().equals(MediaType.APPLICATION_JSON_TYPE)) {
```
srdo: Nit: I think this can be simplified to `MediaType.APPLICATION_JSON_TYPE.equals(request.getMediaType())`

srdo:
+1

PR #2836, opened 2018-09-15
STORM-3162: Cleanup heartbeats cache and make it thread safe

revans2:
This is an alternative to #2800 

@zd-project I had such trouble really understanding what was happening with the heartbeat cache, because it was spread throughout so many files that I couldn't really be sure that #2800 would fix the issues or not.

I decided that I needed to refactor the code to really understand how it worked, and provide a clean interface to the cache, which is what I did.  Once I had that it became clear to me that #2800 would fix the immediate issue, but there were other races in the code, especially with the caches for the executors.  This is probably not a big deal but I felt it would be best to just fix it too.

### DIFF
--- a/storm-client/src/jvm/org/apache/storm/stats/ClientStatsUtil.java
+++ b/storm-client/src/jvm/org/apache/storm/stats/ClientStatsUtil.java
@@ -1,0 +1,202 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.storm.stats;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import org.apache.storm.generated.ClusterWorkerHeartbeat;
+import org.apache.storm.generated.ExecutorInfo;
+import org.apache.storm.generated.ExecutorStats;
+import org.apache.storm.generated.GlobalStreamId;
+import org.apache.storm.shade.com.google.common.collect.Lists;
+import org.apache.storm.utils.Time;
+
+/**
+ * Stats calculations needed by storm client code.
+ */
+public class ClientStatsUtil {
+    public static final String SPOUT = "spout";
+    public static final String BOLT = "bolt";
+    static final String EXECUTOR_STATS = "executor-stats";
+    static final String UPTIME = "uptime";
+    public static final String TIME_SECS = "time-secs";
+    public static final ToGlobalStreamIdTransformer TO_GSID = new ToGlobalStreamIdTransformer();
+    public static final IdentityTransformer IDENTITY = new IdentityTransformer();
+
+    /**
+     * Convert a List<Long> executor to java List<Integer>.
+     */
+    public static List<Integer> convertExecutor(List<Long> executor) {
+        return Lists.newArrayList(executor.get(0).intValue(), executor.get(1).intValue());
+    }
+
+    /**
+     * Make and map of executors to empty stats.
+     * @param executors the executors as keys of the map.
+     * @return and empty map of executors to stats.
+     */
+    public static Map<List<Integer>, ExecutorStats> mkEmptyExecutorZkHbs(Set<List<Long>> executors) {
+        Map<List<Integer>, ExecutorStats> ret = new HashMap<>();
+        for (Object executor : executors) {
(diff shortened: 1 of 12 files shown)


storm-client/src/jvm/org/apache/storm/stats/ClientStatsUtil.java
```diff
@@ -0,0 +1,202 @@
...
+    static class ToGlobalStreamIdTransformer implements KeyTransformer<GlobalStreamId> {
+        @Override
+        public GlobalStreamId transform(Object key) {
+            if (key instanceof List) {
+                List l = (List) key;
+                if (l.size() > 1) {
+                    return new GlobalStreamId((String) l.get(0), (String) l.get(1));
+                }
+            }
+            return new GlobalStreamId("", key.toString());
```
kishorvpatil: Should this be marked to _DEFAULT_STREAM_ID_ or is this expected to be ignored?

storm-server/src/main/java/org/apache/storm/daemon/nimbus/HeartbeatCache.java
```diff
@@ -0,0 +1,229 @@
...
+    /**
+     * Remove a specific topology from the cache.
+     * @param topoId the id of the topology to remove.
+     */
+    public void removeTopo(String topoId) {
+        cache.remove(topoId);
+    }
+
+    /**
+     * Update the heartbeats for a topology with no heartbeats that came in.
```
srdo: I don't understand what this is saying.

storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
```diff
@@ -4624,7 +4569,7 @@ public boolean isRemoteBlobExists(String blobKey) throws AuthorizationException,
         return true;
     }
 
-    private static final class Assoc<K, V> implements UnaryOperator<Map<K, V>> {
+    static final class Assoc<K, V> implements UnaryOperator<Map<K, V>> {
```
srdo: Assoc and Dissoc don't seem to be used in new places, so is making them package private still necessary?

storm-server/src/main/java/org/apache/storm/daemon/nimbus/HeartbeatCache.java
```diff
@@ -0,0 +1,229 @@
...
+        for (List<Integer> exec : allExecutors) {
+            List<Long> longExec = new ArrayList<>(exec.size());
+            for (Integer num : exec) {
+                longExec.add(num.longValue());
+            }
+
+            Long startTime = execToStartTimes.get(longExec);
+            ExecutorCache executorCache = topoCache.get(exec);
+            //null isTimedOut means worker never reported any heartbeat
+            Boolean isTimedOut = executorCache == null ? null : executorCache.isTimedOut();
```
srdo: Nit: Would probably be simpler to set isTimedOut to false if executorCache is null. That avoids the extra null check below.

storm-server/pom.xml
```diff
@@ -171,7 +171,7 @@
                 <artifactId>maven-checkstyle-plugin</artifactId>
                 <!--Note - the version would be inherited-->
                 <configuration>
-                    <maxAllowedViolations>780</maxAllowedViolations>
+                    <maxAllowedViolations>853</maxAllowedViolations>
```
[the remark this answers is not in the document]
revans2: Yes it is all the issues that came with StatsUtil, which is why the other one went in a good direction.  I can try and clean them up more if you want.

storm-server/src/main/java/org/apache/storm/daemon/nimbus/HeartbeatCache.java
```diff
@@ -0,0 +1,229 @@
...
+    }
+
+    /**
+     * Update the heartbeats for a topology with no heartbeats that came in.
+     * @param topoId the id of the topology to look at.
+     * @param taskTimeoutSecs the timeout to know if they are too old.
+     */
+    public void timeoutOldHeartbeats(String topoId, Integer taskTimeoutSecs) {
+        Map<List<Integer>, ExecutorCache> topoCache = cache.computeIfAbsent(topoId, MAKE_MAP);
+        //if not executor beats, refresh is-timed-out of the cache which is done by master
```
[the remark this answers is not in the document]
revans2: The comment came from the original code as I was doing refactoring.  I'll make it better.

storm-server/src/main/java/org/apache/storm/daemon/nimbus/HeartbeatCache.java
revans2: Yes it is very confusing.  I'll fix it.

storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
revans2: Yes, I'll make them private again.  This was because as I was refactoring I kept the Assoc and  Dissoc until the very end when I fixed the threading.

storm-server/src/main/java/org/apache/storm/daemon/nimbus/HeartbeatCache.java
revans2: Yes I agree.  Will do it.  Again this was code that was already there, and I just refactored it, but I am happy to fix it.

storm-server/pom.xml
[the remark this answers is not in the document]
srdo: No, it's fine.

revans2:
@srdo I think I addressed all of your review comments.

srdo:
Looks good, thanks. +1

PR #2840, opened 2018-09-17
STORM-3147: Add metrics based on ClusterSummary

srdo:
Rebase and update of https://github.com/apache/storm/pull/2764.

@zd-project Please take a look when you have a chance.

### DIFF
--- a/storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
+++ b/storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
@@ -18,8 +18,12 @@
 
 package org.apache.storm.daemon.nimbus;
 
+import com.codahale.metrics.CachedGauge;
+import com.codahale.metrics.DerivativeGauge;
 import com.codahale.metrics.Histogram;
 import com.codahale.metrics.Meter;
+import com.codahale.metrics.MetricSet;
+import com.codahale.metrics.SlidingTimeWindowReservoir;
 import com.codahale.metrics.Timer;
 import java.io.File;
 import java.io.FileInputStream;
@@ -50,6 +54,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Function;
 import java.util.function.UnaryOperator;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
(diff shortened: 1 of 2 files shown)


storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
```diff
@@ -2922,19 +2948,20 @@ public void launchServer() throws Exception {
                                         }
                                     });
 
-            metricsRegistry.registerGauge("nimbus:num-supervisors", () -> state.supervisors(null).size());
-            metricsRegistry.registerGauge("nimbus:fragmented-memory", this::fragmentedMemory);
-            metricsRegistry.registerGauge("nimbus:fragmented-cpu", this::fragmentedCpu);
-            metricsRegistry.registerGauge("nimbus:available-memory", () -> nodeIdToResources.get().values().parallelStream()
-                .mapToDouble(SupervisorResources::getAvailableMem)
+            metricsRegistry.registerGauge("nimbus:total-available-memory-non-negative", () -> nodeIdToResources.get().values()
```
revans2: There appears to be some oddness with the naming of the metrics.  It would be nice to have some kind of convention for them.

Perhaps we can have something like 

 * 'nimbus:cluster:available-memory-mb' to show that this came from nimbus, it is about the cluster and it is the available memory in MB.
 * 'nimbus:cluster:available-cpu-pct'
 * 'nimbus:cluster:total-memory-mb'
 * 'nimbus:cluster:total-cpu-pct'

etc.

storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
```diff
@@ -4767,4 +4797,205 @@ public IScheduler getForcedScheduler() {
...
+        private final Histogram topologiesNumExecutors = registerHistogram.apply("topologies:num-executors");
+        private final Histogram topologiesNumWorker = registerHistogram.apply("topologies:num-workers");
+        private final Histogram topologiesUptime = registerHistogram.apply("topologies:uptime-secs");
+        private final Histogram topologiesReplicationCount = registerHistogram.apply("topologies:replication-count");
+        private final Histogram topologiesRequestedMemOnHeap = registerHistogram.apply("topologies:requested-mem-on-heap");
+        private final Histogram topologiesRequestedMemOffHeap = registerHistogram.apply("topologies:requested-mem-off-heap");
+        private final Histogram topologiesRequestedCpu = registerHistogram.apply("topologies:requested-cpu");
+        private final Histogram topologiesAssignedMemOnHeap = registerHistogram.apply("topologies:assigned-mem-on-heap");
+        private final Histogram topologiesAssignedMemOffHeap = registerHistogram.apply("topologies:assigned-mem-off-heap");
+        private final Histogram topologiesAssignedCpu = registerHistogram.apply("topologies:assigned-cpu");
```
revans2: I am sorry, but I am really having a hard time of seeing a histogram of these being valuable, but perhaps I am wrong.  I would much rather see dimensions/tags being used, but since those are not available yet, I am okay with this.

srdo:
I'm hoping @zd-project will help review this, since this is a rebase of his PR. I don't really have the context to know whether these added metrics are useful.

revans2:
I worked with @zd-project fairly closely to come up with the goals for the metrics.  I just think we can do better on some of the metrics, but it is not something we have to do right now.  I am +1 for merging this in as is, but I think we will need some follow on JIRA to clean things up and document them better.

PR #2842, opened 2018-09-18
STORM-3229:  Add in better error reporting

revans2:


revans2:
@danny0405 I changed the log to info.  That is a great point, because the only time you would see that under normally would be if someone had configured the cluster to use digest auth, which we really only have around for testing purposes.

PR #2844, opened 2018-09-21
 STORM-3232: Display on the UI all versions of storm that are supported

revans2:
Also display the full version info from STORM-621


revans2:
Some example screen shots. On the last one I had to remove the git repo and user name because of work requirements :(

multiple

multipletoggle

basic

toggled

Ethanlm:
Why is `Topologies matching 1.0 will run under 2.0-unittests`? Should it be `run under 2.0.0-SNAPSHOT` since the nimbus version is `2.0.0-SNAPSHOT`?

revans2:
@Ethanlm I used the same config that I added as a part of the unit tests for this.  It has garbage data in it, but made it so I didn't have to fully setup multiple versions on my test cluster.

PR #2852, opened 2018-09-27
STORM-3237 track Nimbus mkAssignment failures

agresch:


storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
```diff
@@ -2850,6 +2853,7 @@ public void launchServer() throws Exception {
                                             }
                                             doCleanup();
                                         } catch (Exception e) {
+                                            this.mkAssignmentsErrors.mark();
```
kishorvpatil: should we have separate tracker for cleanup failures and `doCleanup()` ?

agresch: My thought was any exception would be worth investigating in any case.  I would hope that these issues would be rare.  Let me know if you want it more narrowly applied just to mkAssignment.

kishorvpatil: Yes, let's apply this narrowly to just `mkAssignments`

storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
```diff
@@ -2162,7 +2165,12 @@ private boolean isReadyForMKAssignments() throws Exception {
     }
 
     private void mkAssignments() throws Exception {
-        mkAssignments(null);
+        try {
+            mkAssignments(null);
```
revans2: Do we want to make the change inside `mkAssignments(String scratchTopoId)` so we can count errors during the rebalance command as well?

agresch: I can change that.

agresch:
@revans2 @kishorvpatil - made the change requested by Bobby.

PR #2853, opened 2018-09-29
STORM-2963 Adding notes on GC to Performance.md

roshannaik:


HeartSaVioR:
@roshannaik No worries. Please comment when you're done. Adding `WIP` would be also good to determine in this case.

roshannaik:
i had already updated the PR yesterday before above comment ... so its ready.

PR #2856, opened 2018-10-01
STORM-3241: Hides the event log link when no host is valid

d2r:


d2r:
Thanks for reviews @kishorvpatil, @revans2, @Ethanlm, I have merged this to master.

PR #2858, opened 2018-10-02
STORM-3242: Adds "examples" and "externals" profiles

d2r:
The existing modules have been moved into like-named Maven profiles that
are enabled by default.

They can be disabled, for example, with:
```
mvn clean install -DskipTests=true -P '!examples,!externals'
```

### DIFF
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -262,6 +262,8 @@ The following commands must be run from the top-level directory.
 
 If you wish to skip the unit tests you can do this by adding `-DskipTests` to the command line. 
 
+If you wish to skip the examples and external modules, you can do this by adding `-P '!examples,!externals'` to the command line.
+
 In case you modified `storm.thrift`, you have to regenerate thrift code as java and python code before compiling whole project.
 
 ```sh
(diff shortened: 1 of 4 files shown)


d2r:
Agree. If there are errors/failures in travis-ci, then I will take a look.

srdo:
This looks good, but I'm wondering if Flux and the sql modules should be added to externals as well? They depend on storm-kafka-client, so the build won't work with externals disabled unless you happen to already have storm-kafka-client in your local repo.

revans2:
Sorry @srdo is right we need to make sure that we have the dependencies setup properly.

d2r:
Yes, thank you @srdo for pointing this out.

Would it be all right merely to move `flux` and `sql` to the new `externals` profile, or would we need make source code/layout changes as well to keep things organized?

srdo:
I'd be fine with just moving them to the new profile. It already contains the integration test, which also isn't in externals.

d2r:
OK this is done.
Tested using:
```Bash
rm -rf ~/.m2/repository/
mvn clean install -DskipTests=true -U -P '!examples,!externals'
```

srdo:
+1, thanks for addressing my comment. Please squash to one commit containing the issue number.

d2r:
Squashed; old commit was 9aeac5e517d6f5b1a7c1aba0afc0385a99144471.

srdo:
Thanks @d2r, merged to master.

PR #2860, opened 2018-10-03
STORM-1318: Fixes & ports shell submission to Java

d2r:


revans2:
The code change looks fine, but could you file a follow on JIRA to clean up the pom.xml now that clojure is only needed for tests, or you could do it here, I am fine either way.

d2r:
> could you file a follow on JIRA to clean up the pom.xml now that clojure is only needed for tests

Created [STORM-3243](https://issues.apache.org/jira/browse/STORM-3243).

PR #2862, opened 2018-10-03
STORM-3238: Reduces scope of wrapper, corrects variable

d2r:
https://www.w3schools.com/js/js_this.asp

> In a function definition, this refers to the "owner" of the function.

When we wrapped this in an anonymous function, the data that `this` pointed to changed. So it is a simple fix to pull the variable from the outer scope.


d2r:
Thank you @kishorvpatil for the review. I have merged this to master branch.

PR #2869, opened 2018-10-11
Revert "[STORM-3233] Updated zookeeper client to version 3.4.13 which…

revans2:
… fixes various issues including ZOOKEEPER-2184 that prevents ZooKeeper Java clients working in dynamic IP (container / cloud) environment."

This reverts commit 6d77f0a977d2276daa2f985864b41abac4f59c3f.


agresch:
What was the reason for the upgrade?  Do we need a JIRA to upgrade later?  At what point?

PR #2871, opened 2018-10-11
[STORM-3252] Bug fix for blobstore sync

arunmahadevan:
1.Bug fix for blob sync frequency with time unit error.
2.Bug fix for blob sync delete file, add catch NoSuchFileException.
3.Bug fix for blob sync update blob flie, add catch FileNotFoundException

### DIFF
--- a/storm-server/src/main/java/org/apache/storm/blobstore/BlobStoreUtils.java
+++ b/storm-server/src/main/java/org/apache/storm/blobstore/BlobStoreUtils.java
@@ -12,6 +12,7 @@
 
 package org.apache.storm.blobstore;
 
+import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.HashSet;
@@ -191,6 +192,8 @@ public class BlobStoreUtils {
                     out.close();
                 }
                 isSuccess = true;
+            } catch(FileNotFoundException fnf) {
+                LOG.warn("FileNotFoundException", fnf);
             } catch (IOException | AuthorizationException exception) {
                 throw new RuntimeException(exception);
             } catch (KeyNotFoundException knf) {
--- a/storm-server/src/main/java/org/apache/storm/blobstore/BlobStoreUtils.java
+++ b/storm-server/src/main/java/org/apache/storm/blobstore/BlobStoreUtils.java
@@ -193,7 +193,7 @@ public class BlobStoreUtils {
                 }
                 isSuccess = true;
             } catch(FileNotFoundException fnf) {
-                LOG.warn("FileNotFoundException", fnf);
+                LOG.warn("Blobstore file for key '{}' does not exist or got deleted before it could be downloaded.", key, fnf);
             } catch (IOException | AuthorizationException exception) {
                 throw new RuntimeException(exception);
             } catch (KeyNotFoundException knf) {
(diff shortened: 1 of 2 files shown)


arunmahadevan:
Cherry picked commits from https://github.com/apache/storm/pull/2773 and linked with a JIRA.

storm-server/src/main/java/org/apache/storm/blobstore/BlobStoreUtils.java
```diff
@@ -191,6 +192,8 @@ public static boolean downloadUpdatedBlob(Map<String, Object> conf, BlobStore bl
                     out.close();
                 }
                 isSuccess = true;
+            } catch(FileNotFoundException fnf) {
+                LOG.warn("FileNotFoundException", fnf);
```
kishorvpatil: rethrow fnf?

arunmahadevan: It crashes the nimbus if the exception is propagated. (we have seen this happening if topology gets killed while the non-leader nimbus tries to download the blob). It may be better to swallow and let `downloadUpdatedBlob` return false here?

revans2: Line 271 of this file is already dealing with the race. so in theory eating the FileNotFound or rethrowing it as a KeyNotFound exception should be fine.  I am just trying to understand where the FileNotFound is coming from.

The remoteBlobStore should only be getting thrift exceptions.  The only place in the code I see that could throw a FileNotFoundException is when we are interacting with the local blob, but even then the blobstore code should be changing it into a KeyNotFound or some other thrift exception.

kishorvpatil: I see. but then, any other `IOException` such as failure to reach Blobstore etc could cause nimbus restart. I would treat all IOExceptions in same manner. i.e. log them and return `false`.

revans2: Okay I think I see it now.  On the local file output stream it can show up if the file/directory is deleted before we do the commit.  It would be good to document this, preferably in the LOG message, because FileNotFoundException is redundant, it is already going to log that and knowing why this is only a warning is preferable.

storm-server/src/main/java/org/apache/storm/blobstore/BlobStoreUtils.java
```diff
@@ -191,6 +192,8 @@ public static boolean downloadUpdatedBlob(Map<String, Object> conf, BlobStore bl
                     out.close();
                 }
                 isSuccess = true;
+            } catch(FileNotFoundException fnf) {
+                LOG.warn("Blobstore file for key '%s' does not exist or got deleted before it could be downloaded.", key, fnf);
```
revans2: I think you want '{}' instead of '%s'

arunmahadevan: Thanks for catching.

arunmahadevan:
@revans2 , updated the log message, please check if it makes sense. 

@kishorvpatil I am not sure if we want to swallow IOException, since it may be due to some serious problem which we want to propagate and nimbus cannot continue in that case.

kishorvpatil:
> @kishorvpatil I am not sure if we want to swallow IOException, since it may be due to some serious problem which we want to propagate and nimbus cannot continue in that case.
Agreed. We don't want any other `IOException`.

PR #2873, opened 2018-10-11
STORM-3243: Removes and cleans up Clojure dependencies

d2r:


docs/STORM-UI-REST-API.md
```diff
@@ -1572,15 +1572,6 @@ Sample Response:
...
-}
-```
-
```
d2r: Removed because it contains a reference to `conjure`, which is no longer used in the UI. I did not think it necessary to add a new example, so I just removed it.

PR #2882, opened 2018-10-17
STORM-3260: Add in support to print some state

revans2:


revans2:
@kishorvpatil 

I didn't intend the data to be true JSON.  There are some things Thrift supports that JSON does not, like non-string map keys.  Storm uses those in our thrift data structures.

The goal is to be able to get something out that a human can look at and understand.

To me `not set` is less ambiguous than `"not set"`, `""`, or `{}` because it would never parse as JSON and it is actually not set. A `null` would probably be fine seeing how setting something to null in thrift is the same as not setting it.

kishorvpatil:
@revans2 , ok, looking at output, I thought it was trying to output JSON data. Thanks for the explanation.

kishorvpatil:
Travis-ci build failures seems unrelated to the changes.

PR #2883, opened 2018-10-18
fix(JDBC): Syntax error on INSERT

Benvorth:


Benvorth:
Apache Phoenix does not provide an INSERT statement. Instead UPSERT is the only way to insert data to HBase via Phoenix.

PR #2890, opened 2018-10-21
STORM-3268: Improve integration test stability

srdo:
https://issues.apache.org/jira/browse/STORM-3268

Basically the integration test can't handle logs rolling during the test. To fix this we added a bit of time between spout emits, but I don't think it was enough. I've put a limit on how many tuples the spouts will emit during the tests, which should prevent the logs from growing too much.

Some tests use multiple bolt/spout tasks, and don't wait for all executors to be started. In some cases tests could fail because e.g. two spout instances were supposed to run, but one executor was slow to start so the log for that executor wasn't picked up when the test asked Nimbus for logs for the spout component. The tests now check that Nimbus has all expected executors listed.

The tests extract spout/bolt logs by looking for a magic string. If multiple components were running in the same worker, the log reading logic could pick up e.g. logs from the bolt while looking for logs from the spout. The magic string now includes the component id the log reader is looking for.

Some tests were slow due to building lots of useless error messages, the error messages are constructed lazily now.

The tests aren't perfectly stable, but I think this is due to an issue in Storm where tuples can be lost if the spout starts emitting before the bolt executor is ready to receive. I've seen the spout emit a bunch of tuples before the bolt worker started, and those tuples end up timing out.

### DIFF
--- a/integration-test/src/main/java/org/apache/storm/st/topology/TestableTopology.java
+++ b/integration-test/src/main/java/org/apache/storm/st/topology/TestableTopology.java
@@ -17,14 +17,18 @@
 
 package org.apache.storm.st.topology;
 
+import java.util.concurrent.TimeUnit;
 import org.apache.storm.generated.StormTopology;
 
 public interface TestableTopology {
     String DUMMY_FIELD = "dummy";
-    //Some tests rely on reading the worker log. If emits are too close together and too much is logged, the log might roll, breaking the test.
-    int MIN_SLEEP_BETWEEN_EMITS_MS = 10;
-    int MAX_SLEEP_BETWEEN_EMITS_MS = 100;
+    int TIMEDATA_SLEEP_BETWEEN_EMITS_MS = 20;
+    //Some tests rely on reading the worker log. If there are too many emits and too much is logged, the log might roll, breaking the test.
+    //Ensure the time based windowing tests can emit for 5 minutes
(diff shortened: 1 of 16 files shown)


d2r:
Depends on #2889

integration-test/src/main/java/org/apache/storm/st/topology/TestableTopology.java
```diff
@@ -17,14 +17,18 @@
...
+import java.util.concurrent.TimeUnit;
 import org.apache.storm.generated.StormTopology;
 
 public interface TestableTopology {
     String DUMMY_FIELD = "dummy";
-    //Some tests rely on reading the worker log. If emits are too close together and too much is logged, the log might roll, breaking the test.
-    int MIN_SLEEP_BETWEEN_EMITS_MS = 10;
-    int MAX_SLEEP_BETWEEN_EMITS_MS = 100;
+    int TIMEDATA_SLEEP_BETWEEN_EMITS_MS = 20;
+    //Some tests rely on reading the worker log. If there are too many emits and too much is logged, the log might roll, breaking the test.
```
d2r: I wonder if we could instead configure the logs to roll less often, since that is the real issue?

srdo: Yes, we could disable log rolling by customizing the cluster's `log4j2.xml`. I wanted to avoid doing that because it means you can't (reliably) run the integration test against your own cluster without having to edit `log4j2.xml` as well.

d2r: OK, I assumed there was a way for the test itself to manipulate this. If not, then no worries.

PR #2893, opened 2018-10-23
STORM-3272 allow worker-launcher to delete dead symlinks

agresch:
When a topology folder contains a symlink to a missing file, the worker-launcher would not remove the symlink.  This prevents force delete from deleting the folder when cleaning up a topology.  

When a cluster is short on resources, this can cause all sorts of issues on the supervisor when rescheduling the same topology when this folder still exists.

### DIFF
--- a/storm-core/src/native/worker-launcher/impl/worker-launcher.c
+++ b/storm-core/src/native/worker-launcher/impl/worker-launcher.c
@@ -593,14 +593,17 @@ int recursive_delete(const char *path, int supervisor_owns_dir) {
     return UNABLE_TO_BUILD_PATH;
   }
 
+  struct stat file_stat;
+
   if(access(path, F_OK) != 0) {
     if(errno == ENOENT) {
-      return 0;
-    }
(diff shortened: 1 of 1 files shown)


storm-core/src/native/worker-launcher/impl/worker-launcher.c
```diff
@@ -593,14 +593,17 @@ int recursive_delete(const char *path, int supervisor_owns_dir) {
...
   if(access(path, F_OK) != 0) {
     if(errno == ENOENT) {
-      return 0;
-    }
-    // Can probably return here, but we'll try to lstat anyway.
-  }
+       // we need to handle symlinks that target missing files.
+       if((lstat(path, &file_stat) != 0) || ((file_stat.st_mode & S_IFMT) != S_IFLNK)) {
+         return 0;
+       }
```
d2r: If we fail to `lstat` the path, do we want to log a message as we do below?

agresch: @d2r @revans2 - added logging

PR #2896, opened 2018-10-26
[Storm-3261] Fix broken links in documentation

contradictioned:
As I wrote in the [issue](https://issues.apache.org/jira/browse/STORM-3261) I found that some pages were not generated due to missing YAML headers. Since the files are linked from the index page, I belive they are intended to be rendered.  
I added them and by the way also fixed some broken links.


docs/storm-sql-reference.md
```diff
@@ -189,7 +189,7 @@ may refer to tables in the FROM clause of an enclosing query.
 
 *selectWithoutFrom* is equivalent to VALUES,
 but is not standard SQL and is only allowed in certain
-[conformance levels]({{ site.apiRoot }}/org/apache/calcite/sql/validate/SqlConformance.html#isFromRequired--).
+[conformance levels](https://calcite.apache.org/apidocs/org/apache/calcite/sql/validate/SqlConformance.html#isFromRequired--).
```
d2r: Can we avoid hard-coding the host?

contradictioned: In other instances, there is just a link beginning with `javadocs` leading to the local copy of generated javadoc (e.g., to http://storm.apache.org/releases/1.2.2/javadocs/... ), but `org.apache.calcite.sql` is not included there. So I don't think so.

d2r: Right, makes sense.

srdo:
@d2r You can test this by checking out https://github.com/apache/storm-site and following the guide for publishing the documentation

danny0405:
@contradictioned Great work, BTW, can we rename the title to "[Storm-3261] intent of this PR"  format ?

PR #2899, opened 2018-11-01
[STORM-3265] flight.bash fall back to use java utils directly if $JAVA_HOME not set or java not avaialble in $BINPATH

Ethanlm:
https://issues.apache.org/jira/browse/STORM-3265

If JAVA_HOME is not set, or `/usr/bin/java` not available, we should just use java directly


Ethanlm:
Last commit prepended "[STORM-3265]" to the commit message

PR #2901, opened 2018-11-06
[STORM-3271] Docker support:  launch storm workers in docker containers

Ethanlm:
Spent a lot of effort on this.

This patch let storm supervisor launches the worker in a docker container so that user code is isolated from each other and from the host machine. This is a security enhancement.

Detailed explanation of this are available at docs/Docker-support.md in this pr.

I tested it manually (only on `RHEL7`, didn't test it on other os systems) and tested with secure cluster setup. Also tested with our own integration test suites with docker support turned off and proved that this pr won't break storm.


Ethanlm:
We are testing this (with additional related commits) more thoroughly. Will update the pull request after. Additionally, we are working on using `runc` to manage containers directly to avoid many issues from docker itself.

PR #2902, opened 2018-11-06
[STORM-3282] Fix RAS worker count estimation

kishorvpatil:


kishorvpatil:
The jenkins failure seems unrelated.

storm-server/src/main/java/org/apache/storm/utils/ServerUtils.java
```diff
@@ -691,8 +691,10 @@ public static boolean isRAS(Map<String, Object> conf) {
 
     public static int getEstimatedWorkerCountForRASTopo(Map<String, Object> topoConf, StormTopology topology)
         throws InvalidTopologyException {
-        return (int) Math.ceil(getEstimatedTotalHeapMemoryRequiredByTopo(topoConf, topology) /
-                               ObjectReader.getDouble(topoConf.get(Config.WORKER_HEAP_MEMORY_MB)));
+        Double defaultWorkerMaxHeap = ObjectReader.getDouble(topoConf.get(Config.WORKER_HEAP_MEMORY_MB));
+        Double topologyWorkerMaxHeap = ObjectReader.getDouble(topoConf.get(Config.TOPOLOGY_WORKER_MAX_HEAP_SIZE_MB));
```
Ethanlm: Will `topoConf.get(Config.WORKER_HEAP_MEMORY_MB)` be null? If so, this will throw exception `IllegalArgumentException("Don't know how to convert null to double");`

kishorvpatil: There is default of this value - https://github.com/apache/storm/blob/master/conf/defaults.yaml#L332 and https://github.com/apache/storm/blob/master/conf/defaults.yaml#L332 avoids it. But we can 
So I don't anticipate that happening, but I am adding defaults.

Ethanlm: Oh. ok. If there is a default value in yaml file, then that's fine. I think we don't want to have to maintain two default values both in default.yaml and also in the code.

kishorvpatil: The default in the code  is last resort, as will be picked if we ever remove defaults from yaml

PR #2906, opened 2018-11-12
STORM-3123 - add support for Kafka security config in storm-kafka-monitor

arunmahadevan:


arunmahadevan:
Have tested the changes with Kafka broker running in 2-way SSL and unsecure modes and able to see the lags. The change on the kafka spout side is to return the KafkaConfig properties in getComponentConfiguration so that the lag util can use it.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
```diff
@@ -692,11 +692,28 @@ public String toString() {
         configuration.put(configKeyPrefix + "topics", getTopicsString());
 
         configuration.put(configKeyPrefix + "groupid", kafkaSpoutConfig.getConsumerGroupId());
-        configuration.put(configKeyPrefix + "bootstrap.servers", kafkaSpoutConfig.getKafkaProps().get("bootstrap.servers"));
-        configuration.put(configKeyPrefix + "security.protocol", kafkaSpoutConfig.getKafkaProps().get("security.protocol"));
+        for (Entry<String, Object> conf: kafkaSpoutConfig.getKafkaProps().entrySet()) {
+            if (conf.getValue() != null && isPrimitiveOrWrapper(conf.getValue().getClass())) {
```
HeartSaVioR: nit: Might be better to leave a log message for dropped configuration keys from here. Maybe DEBUG is fine since I guess they're only used for storm-kafka-monitor.

external/storm-kafka-monitor/src/main/java/org/apache/storm/kafka/monitor/NewKafkaSpoutOffsetQuery.java
```diff
@@ -27,12 +27,15 @@
     private final String consumerGroupId; // consumer group id for which the offset needs to be calculated
     private final String bootStrapBrokers; // bootstrap brokers
     private final String securityProtocol; // security protocol to connect to kafka
+    private final String consumerConfig; // security configuration file to connect to secure kafka
```
HeartSaVioR: nit: IMHO we may be able to have better name (like securityConfFilePath?) to represent what comment says. Other parameters look like self-described but I couldn't imagine new parameter points to the file by its name.

storm-core/src/jvm/org/apache/storm/utils/TopologySpoutLag.java
```diff
@@ -68,14 +77,38 @@
...
-        String securityProtocol = (String) jsonConf.get(CONFIG_KEY_PREFIX + "security.protocol");
+        String securityProtocol = (String) jsonConf.get(SECURITY_PROTOCOL_CONFIG);
         if (securityProtocol != null && !securityProtocol.isEmpty()) {
             commands.add("-s");
             commands.add(securityProtocol);
         }
         return commands;
     }
 
+    private static File getExtraPropertiesFile(Map<String, Object> jsonConf) {
```
HeartSaVioR: nit: maybe using `create` or `build` or so instead of `get` would be clear to represent that new (temporary) file is generated.

external/storm-kafka-monitor/src/main/java/org/apache/storm/kafka/monitor/NewKafkaSpoutOffsetQuery.java
arunmahadevan: it can be any properties. renamed to `consumerPropertiesFileName`.

arunmahadevan:
@HeartSaVioR thanks for reviewing. Addressed comments.

PR #2907, opened 2018-11-14
STORM-2990, STORM-3279: Fix issue where Kafka Trident spout could ign…

srdo:
…ore EARLIEST and LATEST, and make EARLIEST and LATEST only take effect on topology deploy

See https://issues.apache.org/jira/browse/STORM-2990 and https://issues.apache.org/jira/browse/STORM-3279.

3279 is fixed by always returning a metadata from the emitter, even if no tuples were emitted. If no tuples are emitted, the metadata will now indicate that `consumer.position() - 1` was emitted.

2990 is fixed by inspecting the lastBatchMeta topology id, and deciding whether to start over based on that. 

I expanded the emitter tests to cover these cases as well, and tried to make them a bit more readable by splitting them into multiple classes. The Kafka client dependency has been bumped to 0.11.0.0, since the MockConsumer wasn't behaving properly in 0.10.1.0.

The only other change is to replace the KafkaUnit internals so we no longer depend on Kafka's internal test tooling.


srdo:
Test failure looks unrelated.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/trident/KafkaTridentSpoutEmitter.java
```diff
@@ -218,26 +228,27 @@ private void emitTuple(TridentCollector collector, ConsumerRecord<K, V> record)
...
                 consumer.seekToBeginning(Collections.singleton(tp));
-            } else if (firstPollOffsetStrategy == LATEST) {
+            } else if (firstPollOffsetStrategy == LATEST && isFirstPollSinceTopologyWasDeployed) {
                 LOG.debug("First poll for topic partition [{}], seeking to partition end", tp);
                 consumer.seekToEnd(Collections.singleton(tp));
             } else if (lastBatchMeta != null) {
                 LOG.debug("First poll for topic partition [{}], using last batch metadata", tp);
                 consumer.seek(tp, lastBatchMeta.getLastOffset() + 1);  // seek next offset after last offset from previous batch
-            } else if (firstPollOffsetStrategy == UNCOMMITTED_EARLIEST) {
+            } else if (firstPollOffsetStrategy == UNCOMMITTED_EARLIEST || firstPollOffsetStrategy == EARLIEST) {
```
janithkv: Will we ever hit  firstPollOffsetStrategy == EARLIEST  case here ?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/trident/KafkaTridentSpoutEmitter.java
```diff
@@ -218,26 +228,27 @@ private void emitTuple(TridentCollector collector, ConsumerRecord<K, V> record)
...
                 LOG.debug("First poll for topic partition [{}] with no last batch metadata, seeking to partition beginning", tp);
                 consumer.seekToBeginning(Collections.singleton(tp));
-            } else if (firstPollOffsetStrategy == UNCOMMITTED_LATEST) {
+            } else if (firstPollOffsetStrategy == UNCOMMITTED_LATEST || firstPollOffsetStrategy == LATEST) {
```
janithkv: Will we ever hit firstPollOffsetStrategy == LATEST case here ?

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/trident/KafkaTridentSpoutEmitter.java
srdo: No, doesn't look like it. Good catch.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/trident/KafkaTridentSpoutEmitter.java
```diff
@@ -170,15 +170,25 @@ public void reEmitPartitionBatch(TransactionAttempt tx, TridentCollector collect
...
+                currentBatch = new KafkaTridentSpoutBatchMetadata(
+                    records.get(0).offset(),
+                    records.get(records.size() - 1).offset(),
+                    topologyContext.getStormId());
+            } else {
+                //Build new metadata based on the consumer position.
+                //We want the next emit to start at the current consumer position,
+                //so make a meta that indicates that position - 1 is the last emitted offset
+                //This helps us avoid cases like STORM-3279, and simplifies the seek logic.
+                long lastEmittedOffset = consumer.position(currBatchTp) - 1;
```
revans2: Is there ever a chance that `lastEmittedOffset` could end up being negative? I am thinking specifically if we start up a topology reading from a new/empty topic.

srdo: Yes, then the starting offset may be 0 and lastEmittedOffset could be -1.

The `seek` method used by `emitPartitionBatchNew` always adds 1 to the last batch meta offset, so we'd be seeking to 0 or larger.

`reEmitPartitionBatch` does use the last batch meta offset without adding 1, but if we hit this case there are no emits in the batch, so Trident shouldn't replay it. I'm happy to add guards for -1 to `reEmitPartitionBatch` though, since I don't feel certain that Trident could never replay an empty batch.

Any other reason you can see that we should be aware of negative offsets?

srdo:
Regarding looping offsets, I don't believe Kafka really does anything special to handle offset value overflow (quick google turned up http://mail-archives.apache.org/mod_mbox/kafka-users/201510.mbox/%3CCAE+KEuHikS71hEam=imt4K2OOirYf5hed=f3Yhu3eyM3uOwxYg@mail.gmail.com%3E and http://mail-archives.apache.org/mod_mbox/kafka-users/201111.mbox/%3cCAFbh0Q3wXeivW9G+xre5eN8-SLVprdNLWyKmd+CcW6n=A=ZSHQ@mail.gmail.com%3e).

I think the assumption is that a long is large enough that wrapping around realistically won't happen.

srdo:
Added a check to `reEmitPartitionBatch`. Let me know if it looks reasonable, and I'll squash.

srdo:
Rebased to fix conflicts and squashed to one commit.

external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/trident/KafkaTridentSpoutEmitter.java
revans2: okay sounds good.

revans2:
+1

srdo:
Thanks for the reviews, merged to master.

PR #2908, opened 2018-11-15
STORM-3276: Updated Flux to deal with storm local correctly

revans2:


storm-server/src/main/java/org/apache/storm/LocalCluster.java
```diff
@@ -1197,8 +1238,9 @@ public IBolt makeAckerBoltImpl() {
      * When running a topology locally, for tests etc.  It is helpful to be sure that the topology is dead before the test exits.  This is
      * an AutoCloseable topology that not only gives you access to the compiled StormTopology but also will kill the topology when it
      * closes.
-     *
+     * ```
```
kishorvpatil: Minor one. Not sure if ``` turns it into pre formatted code. Do we need to use `<code`> tag?

jnioche:
Tested on a StormCrawler topology, works fine in local and remote mode.
one quick question though: shouldn't --sleep be deprecated? --local-ttl is used for all the local jobs instead

flux/flux-core/src/main/java/org/apache/storm/flux/Flux.java
```diff
@@ -52,17 +52,22 @@
 public class Flux {
     private static final Logger LOG = LoggerFactory.getLogger(Flux.class);
 
+    @Deprecated
```
srdo: Given that current scripts to start a Flux topology will have to be changed to make this work, is there any value to preserving these options? If someone is using one of these, most likely their script will end up failing or doing the wrong thing.

storm-server/src/main/java/org/apache/storm/LocalCluster.java
```diff
@@ -323,6 +342,13 @@ private static boolean areAllWorkersWaiting() {
         }
     }
 
+    private static final long DEFAULT_ZK_PORT = 2181;
```
srdo: Nit: Please move this to the other constants so it's not in the middle of method declarations.

flux/flux-core/src/main/java/org/apache/storm/flux/Flux.java
jnioche: IMHO this it is acceptable for a major release.  Projects need a drastic cleanup once in a while.

srdo: Sure, what I meant was if we're going to break the API anyway I think it makes sense to remove the options entirely, rather than deprecating them.

jnioche: Got you! Your comment was pretty clear, I just need another coffee :-)

revans2:
@srdo @kishorvpatil @jnioche 

Sorry this has taken so long.  I just rebased/squashed and @srdo I addressed your nit.  Please take another look.

srdo:
+1.

I'm still wondering why we need to preserve command line options here https://github.com/apache/storm/pull/2908#discussion_r236358832 but I think it's fine either way.

PR #2915, opened 2018-11-23
[STORM-3291]Worker can't run as the user who submitted the topology

liu-zhaokun:
[https://issues.apache.org/jira/browse/STORM-3291](https://issues.apache.org/jira/browse/STORM-3291)
Without principal, worker can't be launched as the user who submitted the topology even we set "supervisor.run.worker.as.user" to true.Because the submitterUser will be overwrited by the user who launched nimbus.Even in ui,we can see the owner is the user who launch nimbus.


liu-zhaokun:
@revans2 could you help me to review this PR? thanks

liu-zhaokun:
@revans2 
Thanks for your reply.I really want to launch worker as who submit it.I think TOPOLOGY_SUBMITTER_USER is for this,but this config doesn't work.I know your concerns，but is there any other way to launch worker as users?

revans2:
@liu-zhaokun Currently the only way that the user name is transmitted from the client to the server is through authentication because authentication is secure and we didn't want to accidentally enable a false sense of security when auth is disabled.  We set the owner of the topology to the name of the user running nimbus because that is who the topology was truly running as.

If you really want to make this happen you will need to do one of two things.  Either 
1. install some type of authentication, could be digest which would not be too difficult to get working, or 
2. modify the clients to set a config with the user they want the topology to run as and then have nimbus honor it.

If you do the second option I would want a config to enable this new behavior. Something like

```java
    /**
     * If true nimbus will honor the config user.name from the topology conf, and set it as the
     * owner of the topology.  This only has an impact when authentication is disabled.  Be 
     * very careful when using this when supervisor.run.worker.as.user is enabled as any user
     *  can become nearly any user on the box.
     */
    @isBoolean
     public static final String NIMBUS_USE_UNSAFE_USER_CONFIG = "nimbus.use.unsafe.user.config";
```

Just so you know `user.name` is the name of the system property that java sets with the name of the unix user your java process is running as, but it is not safe to assume it is always correct because it can be modified/overwritten.

liu-zhaokun:
@revans2 Hi,is my revision consistent with your thoughts? And I also have a question,whether we can't use ResourceAwareScheduler if supervisor.run.worker.as.user is false?

liu-zhaokun:
Could you help me to retrigger CI?

PR #2918, opened 2018-12-04
STORM-3295 allow blacklist scheduling to function properly with multi…

agresch:
…ple supervisors on a host

If any supervisor on a host remains blacklisted, BlacklistScheduler.getBlacklistHosts() still considers the host blacklisted.  This change forces all supervisors on a supervisor to be released, which will free the host.

It may be nicer to consider blacklisting strictly based on supervisors, but that is open to discussion, and a much larger change than this.

### DIFF
--- a/storm-server/src/main/java/org/apache/storm/scheduler/blacklist/strategies/RasBlacklistStrategy.java
+++ b/storm-server/src/main/java/org/apache/storm/scheduler/blacklist/strategies/RasBlacklistStrategy.java
@@ -22,7 +22,6 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.TreeMap;
 
 import org.apache.storm.generated.InvalidTopologyException;
 import org.apache.storm.scheduler.Cluster;
@@ -106,21 +105,5 @@ public class RasBlacklistStrategy extends DefaultBlacklistStrategy {
             }
         }
         return readyToRemove;
-    }
-
-    private Map<String, Set<String>> createHostToSupervisorMap(final List<String> blacklistedNodeIds, Cluster cluster) {
-        Map<String, Set<String>> hostToSupervisorMap = new TreeMap<>();
-        for (String supervisorId : blacklistedNodeIds) {
-            String hostname = cluster.getHost(supervisorId);
-            if (hostname != null) {
-                Set<String> supervisorIds = hostToSupervisorMap.get(hostname);
-                if (supervisorIds == null) {
-                    supervisorIds = new HashSet<>();
-                    hostToSupervisorMap.put(hostname, supervisorIds);
-                }
-                supervisorIds.add(supervisorId);
-            }
-        }
-        return hostToSupervisorMap;
     }
 }
--- a/storm-server/src/main/java/org/apache/storm/scheduler/blacklist/strategies/RasBlacklistStrategy.java
+++ b/storm-server/src/main/java/org/apache/storm/scheduler/blacklist/strategies/RasBlacklistStrategy.java
@@ -22,6 +22,8 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.TreeMap;
+
 import org.apache.storm.generated.InvalidTopologyException;
 import org.apache.storm.scheduler.Cluster;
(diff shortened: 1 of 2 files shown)


storm-server/src/main/java/org/apache/storm/scheduler/blacklist/strategies/RasBlacklistStrategy.java
```diff
@@ -79,25 +81,46 @@
...
+                    for (String supervisorId : supervisorIds) {
+                        SupervisorDetails sd = availableSupervisors.get(supervisorId);
+                        if (sd != null) {
+                            NormalizedResourcesWithMemory sdAvailable = cluster.getAvailableResources(sd);
+                            int sdAvailableSlots = cluster.getAvailablePorts(sd).size();
+                            readyToRemove.add(supervisorId);
+                            shortage.remove(sdAvailable, cluster.getResourceMetrics());
+                            shortageSlots -= sdAvailableSlots;
+                            LOG.info("Releasing {} with {} and {} slots leaving {} and {} slots to go", supervisorId,
+                                    sdAvailable, sdAvailableSlots, shortage, shortageSlots);
```
danny0405: So what is the purpose we need to release supervisors blacklist grouping by nodes. What if the node is stuck or broken ? If we shuffle the release by nodes, may be the risk would be spread evenly.

agresch: If there are two supervisors on a host that are blacklisted and we release one, one supervisor will remain blacklisted.  This single blacklisted supervisor will be returned here:

https://github.com/apache/storm/blob/master/storm-server/src/main/java/org/apache/storm/scheduler/blacklist/BlacklistScheduler.java#L167

Then this code will see that the same host that both supervisors are on is blacklisted, preventing any scheduling on that node:

https://github.com/apache/storm/blob/master/storm-server/src/main/java/org/apache/storm/scheduler/blacklist/BlacklistScheduler.java#L167

I think a better overall long term solution is to blacklist supervisors instead of hosts, but that touches a lot more code.  In the short term I think this is a relatively small tradeoff to allow blacklisting to work with multiple supervisors per host.

danny0405: @agresch Got your idea.
In my personal opinion, for production, if a node has a blacklist supervisor(which means the supervisor does not send any HBs for some time), most of the cases it because the node machine itself has some problems(for now there are a few causes like: disk is full or network is in disconnection), so a safer way is we never schedule workers to the node if there are some blacklist supervisors on it.

If you want to make use of the healthy supervisor on the node(has some blacklist supervisors also), at least there is a decision to make sure the supervisor is healthy, we can do this through checking the heartbeats of it.

agresch: It sounds like you don't agree even with the existing code that releases blacklisted supervisors when needed for scheduling? 

Should I be looking at converting all the existing code to blacklist supervisors instead of hosts to get things functional when there are more than one supervisor on a host?

Thanks.

storm-server/src/main/java/org/apache/storm/scheduler/blacklist/strategies/RasBlacklistStrategy.java
```diff
@@ -79,25 +81,46 @@
...
                         }
                     }
+                    // make sure we've handled all supervisors on the host before we break
+                    if (!shortage.areAnyOverZero() && shortageSlots <= 0) {
+                        // we have enough resources now...
+                        break;
```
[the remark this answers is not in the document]
agresch: I am doing what you are indicating....  I create a list of supervisors on the host and release all the supervisors before breaking.....

agresch:
@revans2 - I will look at addressing DefaultBlacklistStrategy.

agresch:
@revans2 @danny0405 - Made changes requested by @revans2.  Let me know if we should have a separate JIRA for switching blacklisting to work strictly by supervisor instead of doing this mismatch between host and supervisors.  I think this trade off of releasing all supervisors on a host is acceptable.

danny0405:
@agresch If this strategy only happens when cluster resource is full originally, i'm ok with it. At least it's a better way.

PR #2922, opened 2018-12-08
STORM-3301: Fix case where KafkaSpout could emit tuples that were alr…

srdo:
…eady committed

https://issues.apache.org/jira/browse/STORM-3301

I made a slight change to log compaction handling, the spout will now commit an acked offset after a gap, even if it's the only one. Previously there had to be at least 2 acked offsets after the gap before commit.


srdo:
The test failure is due to broken stubbing, I'll fix it soon

PR #2923, opened 2018-12-08
STORM-3301: Fix case where KafkaSpout could emit tuples that were alr…

srdo:
…eady committed

1.x version of https://github.com/apache/storm/pull/2922


SrinivasVadla4421:
solution need

PR #2924, opened 2018-12-09
STORM-1289: Port integration-test.clj to Java

srdo:
https://issues.apache.org/jira/browse/STORM-1289

I have also ported a few tests to JUnit 5 and changed the IntegrationTest and PerformanceTest categories to be JUnit 5 Tags instead. I think it improves readability of the root pom. Let me know if these changes are too noisy here, and I'll move them to a separate PR.


srdo:
Approved as part of https://github.com/apache/storm/pull/2927

PR #2925, opened 2018-12-10
STORM-3302: Ensures we close streams to HDFS

d2r:


storm-client/src/jvm/org/apache/storm/blobstore/BlobStore.java
```diff
@@ -288,13 +288,15 @@ public void createBlob(String key, InputStream in, SettableBlobMeta meta, Subjec
             while ((len = in.read(buffer)) > 0) {
                 out.write(buffer, 0, len);
             }
-            out.close();
-        } catch (AuthorizationException | IOException | RuntimeException e) {
```
d2r: The documentation for this method indicates that we throw the above exceptions, however they were being caught instead of thrown. This change will result in the exceptions being thrown while still helping to guarantee the `out` stream is closed as the `in` stream was already being ensured to be closed.

storm-client/src/jvm/org/apache/storm/blobstore/BlobStore.java
```diff
@@ -288,13 +288,15 @@ public void createBlob(String key, InputStream in, SettableBlobMeta meta, Subjec
...
-            if (out != null) {
-                out.cancel();
-            }
         } finally {
-            in.close();
+            try {
+                if (out != null) {
```
agresch: we're always calling cancel here instead of close() for the normal case.

storm-client/src/jvm/org/apache/storm/dependency/DependencyUploader.java
```diff
@@ -154,11 +154,22 @@ private boolean uploadDependencyToBlobStore(String key, File dependency)
...
-            Files.copy(dependency.toPath(), blob);
-            blob.close();
+            AtomicOutputStream blob = null;
+            try {
+                blob = getBlobStore().createBlob(key, new SettableBlobMeta(acls));
+                Files.copy(dependency.toPath(), blob);
+                blob.close();
 
-            uploadNew = true;
+                uploadNew = true;
```
agresch: shouldn't we call blob = null here to prevent a dupe cancel after close?

storm-server/src/main/java/org/apache/storm/blobstore/BlobStoreUtils.java
```diff
@@ -204,6 +204,14 @@ public static boolean downloadUpdatedBlob(Map<String, Object> conf, BlobStore bl
             } catch (Exception exp) {
                 // Logging an exception while client is connecting
                 LOG.error("Exception", exp);
+            } finally {
+                if (null != out) {
```
agresch: same here, shouldn't we have nulled out out on close?

storm-client/src/jvm/org/apache/storm/blobstore/BlobStore.java
d2r: That is intentional. Calling `cancel` will also cancel writes associated with the stream, and will clean up any transient resources ("part" files) that may have been left around as a result of an error. It seems best to do this in the error case, and I do not think it will do any harm in other cases.

storm-client/src/jvm/org/apache/storm/dependency/DependencyUploader.java
d2r: Actually we probably want to remove the call to `close`, since it will happen in the `finally` block now.

storm-client/src/jvm/org/apache/storm/blobstore/BlobStore.java
d2r: I see what you mean. We need to preserve a distinction between cleaning up in the normal case and cleaning up in the error case.

In the normal case, we want `close`. In the error case, we want `cancel`.

storm-client/src/jvm/org/apache/storm/blobstore/BlobStore.java
Ethanlm: It looks to me that the exception is still caught instead of thrown?

d2r: Formerly they were caught. Removing these lines will result in them being thrown as both the method documentation and declarations intend.

We may not want this change however, and instead we could update the declaration and the documentation such that the behavior remains unchanged. I am willing to do either one, as this PR is focused on eliminating the possibility of leaking some sockets.

storm-client/src/jvm/org/apache/storm/dependency/DependencyUploader.java
```diff
@@ -154,11 +154,23 @@ private boolean uploadDependencyToBlobStore(String key, File dependency)
...
+            } finally {
+                try {
+                    if (blob != null) {
+                        blob.cancel();
+                    }
+                } catch (IOException throwaway) {
+                    // Ignore.
```
kishorvpatil: It would be useful to log Error/warning in case there is `IOException`.

kishorvpatil: Ignore that. This exception will appear only while attempting to cancel the blob connection. We should ignore it.

d2r:
I have squashed commits. This was a destructive to the branch's history; the old commit was 
368b4fa.

kishorvpatil:
@d2r  Thank you for the fix and squashing the commits.

PR #2928, opened 2018-12-14
STORM-3270: Build Storm with Java 11, excluding some incompatible mod…

srdo:
…ules

https://issues.apache.org/jira/browse/STORM-3270

All Hadoop-related modules are excluded from tests, since Hadoop doesn't work with Java 11 yet. Same for Cassandra.

Libraries I'm aware of that need to be updated, other than Netty which is bumped in this PR:
* Clojure to 1.10 (not released yet, currently going through RCs)
* Hadoop/Hive/HBase need to be updated once https://issues.apache.org/jira/browse/HADOOP-15338 is resolved and released.
* Cassandra needs to be updated to 4.0.0. This update should be easy, since we only depend on it in tests, the production code only relies on the Datastax driver, which should be compatible with most Cassandra versions.
* Dropwizard Metrics should probably be updated to 4.x, but we're blocked by https://issues.apache.org/jira/browse/CASSANDRA-14667. I'm not sure how critical the fixes for Metrics are.
* Kafka should probably be upgraded to 2.1.0. I don't think it is an issue for storm-kafka-client, since users can just specify another client library version, but for storm-kafka-monitor it might be necessary for us to upgrade.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,7 @@ sudo: required
 language: java
 jdk:
   - oraclejdk8
+  - openjdk11
 before_install:
   - rvm reload
   - rvm use 2.4.2 --install
--- a/dev-tools/travis/travis-script.sh
+++ b/dev-tools/travis/travis-script.sh
@@ -38,9 +38,9 @@ then
   TEST_MODULES=storm-core
 elif [ "$2" == "External" ]
 then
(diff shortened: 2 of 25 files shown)


srdo:
The build failed because of a hiccup when downloading dependencies. See https://travis-ci.org/srdo/storm/builds/468124888 instead.

PR #2930, opened 2018-12-26
STORM-3274: Migrates storm CLI to using argparse making documentation…

govind-menon:
… more accessible

### DIFF
--- a/bin/storm.py
+++ b/bin/storm.py
@@ -15,31 +15,31 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-from __future__ import print_function
 
-import os
-import random
-import re
-import shlex
-import tempfile
-import uuid
-import subprocess as sub
+
+import argparse
 import json
-
+import os
+import shlex
+import subprocess
 import sys
+from random import randint
 
-try:
-    # python 3
-    from urllib.parse import quote_plus
-except ImportError:
-    # python 2
-    from urllib import quote_plus
-try:
-    # python 3
-    import configparser
-except ImportError:
-    # python 2
+from argparse import HelpFormatter
+from operator import attrgetter
+
+class SortingHelpFormatter(HelpFormatter):
+
+    def add_arguments(self, actions):
+        actions = sorted(actions, key=attrgetter('option_strings'))
+        super(SortingHelpFormatter, self).add_arguments(actions)
+
+if sys.version_info[0] == 2:
     import ConfigParser as configparser
+    from urllib import quote_plus
+else:
+    import configparser
+    from urllib.parse import quote_plus
 
 def is_windows():
     return sys.platform.startswith('win')
@@ -49,81 +49,13 @@ def identity(x):
 
 def cygpath(x):
     command = ["cygpath", "-wp", x]
-    p = sub.Popen(command,stdout=sub.PIPE)
+    p = subprocess.Popen(command, stdout=subprocess.PIPE)
     output, errors = p.communicate()
     lines = output.split(os.linesep)
     return lines[0]
 
-def init_storm_env():
-    global CLUSTER_CONF_DIR
-    ini_file = os.path.join(CLUSTER_CONF_DIR, 'storm_env.ini')
-    if not os.path.isfile(ini_file):
-        return
-    config = configparser.ConfigParser()
-    config.optionxform = str
-    config.read(ini_file)
-    options = config.options('environment')
-    for option in options:
-        value = config.get('environment', option)
-        os.environ[option] = value
-        
-def get_java_cmd():
-    cmd = 'java' if not is_windows() else 'java.exe'
-    if JAVA_HOME:
-        cmd = os.path.join(JAVA_HOME, 'bin', cmd)
-    return cmd
-
-normclasspath = cygpath if sys.platform == 'cygwin' else identity
-STORM_DIR = os.sep.join(os.path.realpath( __file__ ).split(os.sep)[:-2])
-USER_CONF_DIR = os.path.expanduser("~" + os.sep + ".storm")
-STORM_CONF_DIR = os.getenv('STORM_CONF_DIR', None)
-
-if STORM_CONF_DIR == None:
-    CLUSTER_CONF_DIR = os.path.join(STORM_DIR, "conf")
-else:
-    CLUSTER_CONF_DIR = STORM_CONF_DIR
-
-if (not os.path.isfile(os.path.join(USER_CONF_DIR, "storm.yaml"))):
-    USER_CONF_DIR = CLUSTER_CONF_DIR
-
-STORM_WORKER_LIB_DIR = os.path.join(STORM_DIR, "lib-worker")
-STORM_LIB_DIR = os.path.join(STORM_DIR, "lib")
-STORM_TOOLS_LIB_DIR = os.path.join(STORM_DIR, "lib-tools")
-STORM_WEBAPP_LIB_DIR = os.path.join(STORM_DIR, "lib-webapp")
-STORM_BIN_DIR = os.path.join(STORM_DIR, "bin")
-STORM_LOG4J2_CONF_DIR = os.path.join(STORM_DIR, "log4j2")
-STORM_SUPERVISOR_LOG_FILE = os.getenv('STORM_SUPERVISOR_LOG_FILE', "supervisor.log")
-
-init_storm_env()
-
-CONFIG_OPTS = []
-CONFFILE = ""
-JAR_JVM_OPTS = shlex.split(os.getenv('STORM_JAR_JVM_OPTS', ''))
-JAVA_HOME = os.getenv('JAVA_HOME', None)
-JAVA_CMD = get_java_cmd(); 
-if JAVA_HOME and not os.path.exists(JAVA_CMD):
-    print("ERROR:  JAVA_HOME is invalid.  Could not find bin/java at %s." % JAVA_HOME)
-    sys.exit(1)
-STORM_EXT_CLASSPATH = os.getenv('STORM_EXT_CLASSPATH', None)
-STORM_EXT_CLASSPATH_DAEMON = os.getenv('STORM_EXT_CLASSPATH_DAEMON', None)
-DEP_JARS_OPTS = []
-DEP_ARTIFACTS_OPTS = []
-DEP_ARTIFACTS_REPOSITORIES_OPTS = []
-DEP_MAVEN_LOCAL_REPOSITORY_DIRECTORY = None
-DEP_PROXY_URL = None
-DEP_PROXY_USERNAME = None
-DEP_PROXY_PASSWORD = None
-
-def get_config_opts():
-    global CONFIG_OPTS
-    return "-Dstorm.options=" + ','.join(map(quote_plus,CONFIG_OPTS))
-
-if not os.path.exists(STORM_LIB_DIR):
-    print("******************************************")
-    print("The storm client can only be run from within a release. You appear to be trying to run the client from a checkout of Storm's source code.")
-    print("\nYou can download a Storm release at http://storm.apache.org/downloads.html")
-    print("******************************************")
-    sys.exit(1)
+def get_config_opts(storm_config_opts):
+    return "-Dstorm.options=" + ','.join([quote_plus(s) for s in storm_config_opts])
 
 def get_jars_full(adir):
     files = []
@@ -132,11 +64,7 @@ def get_jars_full(adir):
     elif os.path.exists(adir):
         files = [adir]
 
-    ret = []
-    for f in files:
-        if f.endswith(".jar"):
-            ret.append(os.path.join(adir, f))
-    return ret
+    return [os.path.join(adir, f) for f in files if f.endswith(".jar")]
 
 # If given path is a dir, make it a wildcard so the JVM will include all JARs in the directory.
 def get_wildcard_dir(path):
@@ -147,30 +75,18 @@ def get_wildcard_dir(path):
         ret = [path]
     return ret
 
-def get_classpath(extrajars, daemon=True, client=False):
-    ret = get_wildcard_dir(STORM_DIR)
-    if client:
-        ret.extend(get_wildcard_dir(STORM_WORKER_LIB_DIR))
-    else :
-        ret.extend(get_wildcard_dir(STORM_LIB_DIR))
-    ret.extend(get_wildcard_dir(os.path.join(STORM_DIR, "extlib")))
-    if daemon:
-        ret.extend(get_wildcard_dir(os.path.join(STORM_DIR, "extlib-daemon")))
-    if STORM_EXT_CLASSPATH != None:
-        ret.append(STORM_EXT_CLASSPATH)
-    if daemon and STORM_EXT_CLASSPATH_DAEMON != None:
-        ret.append(STORM_EXT_CLASSPATH_DAEMON)
-    ret.extend(extrajars)
-    return normclasspath(os.pathsep.join(ret))
+def get_java_cmd():
+    cmd = 'java' if not is_windows() else 'java.exe'
+    if JAVA_HOME:
+        cmd = os.path.join(JAVA_HOME, 'bin', cmd)
+    return cmd
 
-def confvalue(name, extrapaths, daemon=True):
-    global CONFFILE
+def confvalue(name, storm_config_opts, extrapaths, daemon=True):
     command = [
-        JAVA_CMD, "-client", get_config_opts(), "-Dstorm.conf.file=" + CONFFILE,
-        "-cp", get_classpath(extrapaths, daemon), "org.apache.storm.command.ConfigValue", name
+        JAVA_CMD, "-client", get_config_opts(storm_config_opts), "-Dstorm.conf.file=" + CONF_FILE,
+        "-cp", get_classpath(extrajars=extrapaths, daemon=daemon), "org.apache.storm.command.ConfigValue", name
     ]
-    p = sub.Popen(command, stdout=sub.PIPE)
-    output, errors = p.communicate()
+    output = subprocess.Popen(command, stdout=subprocess.PIPE).communicate()[0]
     # python 3
     if not isinstance(output, str):
         output = output.decode('utf-8')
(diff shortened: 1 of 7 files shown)


govind-menon:
Tests will be forthcoming

storm-client/pom.xml
```diff
@@ -240,6 +240,29 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
+            <groupId>org.codehaus.mojo</groupId>
+            <artifactId>exec-maven-plugin</artifactId>
+            <executions>
+                <execution>
+                    <configuration>
+                        <executable>python2.7</executable>
```
srdo: I'd like the storm-client build to still work on Windows. One option would be to only run this execution when the OS is UNIX-y, e.g.
```
<profile>
    <activation>
        <os><family>unix</family></os>
    </activation>
    <plugin-config-here/>
</profile>
```

srdo: Or even better, set the executable name to python2.7 on Linux, and python on Windows.

govind-menon: @srdo I've added a profile that would execute only on Unix - I didn't add one for Windows because that would change the file path separators that we're asserting on. I can change the test to make that configureable if you want

bin/storm.py
```diff
@@ -132,13 +56,8 @@ def get_jars_full(adir):
...
         files = [adir]
 
-    ret = []
-    for f in files:
-        if f.endswith(".jar"):
-            ret.append(os.path.join(adir, f))
-    return ret
+    return [os.path.join(adir, f) for f in files if f.endswith(".jar")]
 
-# If given path is a dir, make it a wildcard so the JVM will include all JARs in the directory.
```
srdo: Nit: Is this comment no longer relevant?

bin/storm.py
```diff
@@ -156,49 +95,92 @@ def get_classpath(extrajars, daemon=True, client=False):
...
-def confvalue(name, extrapaths, daemon=True):
-    global CONFFILE
-    command = [
-        JAVA_CMD, "-client", get_config_opts(), "-Dstorm.conf.file=" + CONFFILE,
-        "-cp", get_classpath(extrapaths, daemon), "org.apache.storm.command.ConfigValue", name
-    ]
-    p = sub.Popen(command, stdout=sub.PIPE)
-    output, errors = p.communicate()
-    # python 3
-    if not isinstance(output, str):
```
srdo: Is this check not necessary for Python 3 anymore?

bin/storm.py
```diff
@@ -296,787 +239,1044 @@ def exec_storm_class(klass, jvmtype="-server", jvmopts=[], extrajars=[], args=[]
...
+    and shows query plan instead of submitting topology.
+    """
+
+    sub_parser = subparsers.add_parser("sql", help=command_help)
+
+    add_client_jar_options(sub_parser)
+
+    sub_parser.add_argument("sql_file", metavar="sql-file")
+    sub_parser.add_argument(
+        "topology_name", metavar="topology-name", help="should be --explain to activate explain mode"
```
srdo: Nit: Can we use an argument group to do this instead of putting --explain into topology-name? https://docs.python.org/3/library/argparse.html#mutual-exclusion

storm-client/pom.xml
srdo: Thanks, it looks good. I'm wondering if we can get away with also doing `pip install mock --user` as part of the Maven run, or is that a bad idea?

govind-menon: @srdo It's a little messier to do that and better to leave it up to the environment to do it's own test setup. --user is for then there's no virtualenv (which will vary from machine to machine) - which is a good example of how the host itself should do the setup.

srdo:
This is in reply to https://github.com/apache/storm/pull/2930#discussion_r245726485, for some reason Github won't let me put another comment there.

Makes sense. I think we should at least note that `mock` is a prerequisite in https://github.com/apache/storm/blob/master/DEVELOPER.md#prerequisites. Maybe add a line that users should install `pip` and then do `pip install mock --user`.

srdo:
Would it make sense to run the tests for Python 3 as well as Python 2.7?

srdo:
When I run the build, a storm.pyc file is generated in the storm/bin directory. This file ends up in the storm-dist distribution. I'm guessing this isn't intentional?

srdo:
Can we make running `storm.py` with 0 arguments equivalent to running `storm.py -h`? Running just `storm.py` gives the following output, which isn't very nice:

```
PS E:\apache-storm-2.0.1-SNAPSHOT\bin> ./storm
usage: storm [-h] [--config CONFIG] [-c C]
             {jar,localconfvalue,remoteconfvalue,local,sql,kill,upload-credentials,blobstore,heartbeats,activate,list,deactivate,rebalance,get-errors,kill_workers,admin,shell,repl,nimbus,pacemaker,supervisor,ui,logviewer,drpc-client,drpc,dev-zookeeper,version,classpath,server_classpath,monitor}
             ...
storm: error: too few arguments
```

srdo:
Nit: The existing help message prints the commands alphabetically. We might want to do the same with the new help message. Quick google suggests it is possible https://stackoverflow.com/questions/12268602/sort-argparse-help-alphabetically.

bin/storm.py
```diff
@@ -296,787 +239,1044 @@ def exec_storm_class(klass, jvmtype="-server", jvmopts=[], extrajars=[], args=[]
...
+    print(args.conf_name + ": " + confvalue(args.conf_name, args.c, [USER_CONF_DIR]))
+
+
+def print_remoteconfvalue(args):
+    print(args.conf_name + ": " + confvalue(args.conf_name, args.c, [CLUSTER_CONF_DIR]))
+
+
+def initialize_main_command():
+    main_parser = argparse.ArgumentParser(prog="storm")
+    main_parser.add_argument("--config", default=None, help="Override default storm conf")
```
srdo: Nit: It isn't obvious from the description what the difference between `-c` and `--config` is, or how to use them. Can we update the descriptions? For `--config` something like "Override default storm conf file", and for `-c` something like "Override storm conf properties, e.g. key=val,key2=val2".

govind-menon:
@srdo 

Addressed most of your comments

- Tests are run for python2 and python 3
- Help is sorted alphabetically now
- running just storm prints the entire help
- The mutual exclusion change isn't doable if there are other positional arguments (which there are for that command)

bin/storm.py
govind-menon: There are positional arguments added on L520 that don't allow mutual exclusion

srdo:
Thanks. It looks great. A few final things:

* When running `mvn clean install -DskipTests`, the tests for the script are still run. Can we skip the executions if `skipTests` is true?
* Running the tests still generates a .pyc file in storm/bin, which ends up in the distribution. I think we should either remove it after the tests run, or explicitly exclude it from the storm-dist build.
* The help is sorted alphabetically, which is good. Can we do the same for the positional arguments (e.g. jar, logviewer, kill)?
* I would still like us to describe the new dependency on `mock` in DEVELOPER.md, as mentioned here https://github.com/apache/storm/pull/2930#issuecomment-452016830. I think it is ideal if people can blindly follow the setup guide and build Storm, having the build fail because there are extra setup steps that aren't mentioned in the guide could be demotivating to new people.

srdo:
The .pyc issue can be solved by adding `<PYTHONDONTWRITEBYTECODE>true</PYTHONDONTWRITEBYTECODE>` to the environment variables for the two `exec-maven-plugin` executions

govind-menon:
@srdo I've addressed the 4 comments you had - thank you for the review

kishorvpatil:
@govind-menon The `node-health-check` command  is missing from `storm.py` after this change.

govind-menon:
@kishorvpatil Fixed + added a test for it

kishorvpatil:
@govind-menon 
The CLI is not recognizing -c options are configuration parameters.

bin/storm.py
```diff
@@ -234,60 +228,21 @@ def resolve_dependencies(artifacts, artifact_repositories, maven_local_repos_dir
...
-        JAVA_CMD, jvmtype,
-        "-Ddaemon.name=" + daemonName,
-        get_config_opts(),
-        "-Dstorm.home=" + STORM_DIR,
-        "-Dstorm.log.dir=" + storm_log_dir,
-        "-Djava.library.path=" + confvalue("java.library.path", extrajars, daemon),
-        "-Dstorm.conf.file=" + CONFFILE,
-        "-cp", get_classpath(extrajars, daemon, client=client),
-    ] + jvmopts + [klass] + list(args)
+                   JAVA_CMD, jvmtype,
```
kishorvpatil: excess tabulation?

govind-menon: No I was formatting it so it was more readable

bin/storm.py
```diff
@@ -296,787 +251,1090 @@ def exec_storm_class(klass, jvmtype="-server", jvmopts=[], extrajars=[], args=[]
...
+
+    zk_cli_parser.add_argument(
+        "-s", "--server", default=None, help="""Set the connection string to use,
+        defaults to storm connection string"""
+    )
+
+    zk_cli_parser.add_argument(
+        "-t", "--time-out", default=None, help="""Set the timeout to use, defaults to storm
+            zookeeper timeout."""
+    )
```
d2r: * `type=check_positive`?
* Does this have to be positive? What happens on `0`?
* What are the units?

bin/storm.py
```diff
@@ -296,787 +251,1090 @@ def exec_storm_class(klass, jvmtype="-server", jvmopts=[], extrajars=[], args=[]
...
+    add_common_options(sub_parser)
+
+
+def initialize_nimbus_subcommand(subparsers):
+    command_help = """
+    Launches the nimbus daemon. This command should be run under
+    supervision with a tool like daemontools or monit.
+
+    See Setting up a Storm cluster for more information.
+    (http://storm.apache.org/documentation/Setting-up-a-Storm-cluster
```
d2r: Missing a closing `)` parenthesis

bin/storm.py
```diff
@@ -296,787 +251,1090 @@ def exec_storm_class(klass, jvmtype="-server", jvmopts=[], extrajars=[], args=[]
...
+    Provides a very simple way to send DRPC requests. The server and port are picked from the configs.
     """
-    global DEP_JARS_OPTS, DEP_ARTIFACTS_OPTS, DEP_ARTIFACTS_REPOSITORIES_OPTS, DEP_MAVEN_LOCAL_REPOSITORY_DIRECTORY, DEP_PROXY_URL, DEP_PROXY_USERNAME, DEP_PROXY_PASSWORD
 
-    local_jars = DEP_JARS_OPTS
-    artifact_to_file_jars = resolve_dependencies(DEP_ARTIFACTS_OPTS, DEP_ARTIFACTS_REPOSITORIES_OPTS, DEP_MAVEN_LOCAL_REPOSITORY_DIRECTORY, DEP_PROXY_URL, DEP_PROXY_USERNAME, DEP_PROXY_PASSWORD)
+    sub_parser = subparsers.add_parser("drpc-client", help=command_help, formatter_class=SortingHelpFormatter)
+
+    sub_parser.add_argument(
+        "-f", "--function", default=None, help="""If a -f argument is supplied to set the function name all of the arguments are treated
```
d2r: Suggest `If the -f argument`...

bin/storm.py
```diff
@@ -156,49 +107,92 @@ def get_classpath(extrajars, daemon=True, client=False):
...
+
+    if JAVA_HOME and not os.path.exists(JAVA_CMD):
+        print("ERROR:  JAVA_HOME is invalid.  Could not find bin/java at %s." % JAVA_HOME)
+        sys.exit(1)
+
+    if not os.path.exists(STORM_LIB_DIR):
+        print("*" * 20)
+        print('''The storm client can only be run from within a release. 
+You appear to be trying to run the client from a checkout of Storm's source code.
+You can download a Storm release at http://storm.apache.org/downloads.html")''')
```
d2r: All of the links can be updated to use `https`.

bin/storm.py
```diff
@@ -296,787 +251,1090 @@ def exec_storm_class(klass, jvmtype="-server", jvmopts=[], extrajars=[], args=[]
...
+    """
+    sub_parser = subparsers.add_parser("kill", help=command_help, formatter_class=SortingHelpFormatter)
+
+    sub_parser.add_argument("topology-name")
+
+    sub_parser.add_argument(
+        "-w", "--wait-time-secs",
+        help="""override the length of time Storm waits between deactivation and shutdown""",
+        default=None
+    )
```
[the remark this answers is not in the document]
d2r: Do we want a `check_nonnegative` function?

bin/storm.py
```diff
@@ -296,787 +251,1090 @@ def exec_storm_class(klass, jvmtype="-server", jvmopts=[], extrajars=[], args=[]
...
+
+def add_common_options(parser):
+    parser.add_argument("--config", default=None, help="Override default storm conf file")
+    parser.add_argument(
+        "-storm_config_opts", "-c", action="append", default=[],
+        help="Override storm conf properties , e.g. nimbus.ui.port=4443"
+    )
+
+def add_topology_jar_options(parser):
+    parser.add_argument("topology_jar_path", help="will upload the jar at topology-jar-path when the topology is submitted.")
```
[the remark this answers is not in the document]
govind-menon: I didn't understand this part - they're positional arguments so it doesn't change the invocation?

bin/storm.py
```diff
@@ -296,787 +251,1090 @@ def exec_storm_class(klass, jvmtype="-server", jvmopts=[], extrajars=[], args=[]
...
+    sub_parser.set_defaults(func=heartbeats)
+    add_common_options(sub_parser)
+
+
+def initialize_activate_subcommand(subparsers):
+    sub_parser = subparsers.add_parser(
+        "activate", help="Activates the specified topology's spouts.", formatter_class=SortingHelpFormatter
+    )
+
+    sub_parser.add_argument("topology-name")
```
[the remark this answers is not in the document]
govind-menon: The java command only takes one topology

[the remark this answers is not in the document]
d2r: Yes, this is correct. I misread the current version.

bin/storm.py
```diff
@@ -296,787 +251,1090 @@ def exec_storm_class(klass, jvmtype="-server", jvmopts=[], extrajars=[], args=[]
...
+    initialize_supervisor_subcommand(subparsers)
+    initialize_ui_subcommand(subparsers)
+    initialize_logviewer_subcommand(subparsers)
+    initialize_drpc_client_subcommand(subparsers)
+    initialize_drpc_subcommand(subparsers)
+    initialize_dev_zookeeper_subcommand(subparsers)
+    initialize_version_subcommand(subparsers)
+    initialize_classpath_subcommand(subparsers)
+    initialize_server_classpath_subcommand(subparsers)
+    initialize_monitor_subcommand(subparsers)
```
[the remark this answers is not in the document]
d2r: This has been addressed.

bin/storm.py
d2r: Has this question been addressed?

bin/storm.py
```diff
@@ -296,787 +251,1090 @@ def exec_storm_class(klass, jvmtype="-server", jvmopts=[], extrajars=[], args=[]
...
+        help="sets the number of seconds the local cluster will run for before it shuts down",
+        default=LOCAL_TTL_DEFAULT
+    )
+
+    sub_parser.add_argument(
+        "--java-debug",
+        help="lets you turn on java debugging and set the parameters passed to -agentlib:jdwp on the JDK" +
+             "e.g transport=dt_socket,address=localhost:8000 will open up a debugging server on port 8000",
+        default=None
+    )
```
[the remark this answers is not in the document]
d2r: This has been addressed.

bin/storm.py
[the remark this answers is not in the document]
d2r: This has been addressed.

bin/storm.py
```diff
@@ -296,787 +251,1090 @@ def exec_storm_class(klass, jvmtype="-server", jvmopts=[], extrajars=[], args=[]
...
+    sub_parser = subparsers.add_parser(
+        "rebalance", help=command_help, formatter_class=SortingHelpFormatter
+    )
+
+    sub_parser.add_argument("-w", "--wait", default=None)
+
+    sub_parser.add_argument(
+        "-n", "--num-workers", default=None,
+        help="change the number of requested workers"
+    )
```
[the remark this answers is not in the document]
d2r: This has been addressed.

bin/storm.py
```diff
@@ -296,787 +251,1090 @@ def exec_storm_class(klass, jvmtype="-server", jvmopts=[], extrajars=[], args=[]
...
+
+    sub_parser.add_argument(
+        "-e", "--executors", action="append", default=[],
+        help="change the number of executors for a given component"
+    )
```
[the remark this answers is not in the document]
govind-menon: I added the validation for this in the method itself. Also the default=[] is equivalent to nargs='*'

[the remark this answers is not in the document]
d2r: Ok sounds good.

bin/storm.py
d2r: This has been addressed.

bin/storm.py
d2r: This has been addressed.

bin/storm.py
```diff
@@ -296,787 +251,1090 @@ def exec_storm_class(klass, jvmtype="-server", jvmopts=[], extrajars=[], args=[]
...
+        "replication", formatter_class=SortingHelpFormatter
+    )
+    replication_parser.add_argument("KEY")
+    replication_parser.add_argument(
+        "--read", action="store_true", help="Used to read the replication factor of the blob"
+    )
+    replication_parser.add_argument(
+        "--update", action="store_true", help=" It is used to update the replication factor of a blob."
+    )
+    replication_parser.add_argument("-r", "--replication-factor", default=None)
```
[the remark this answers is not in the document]
d2r: This has been addressed.

bin/storm.py
```diff
@@ -296,787 +251,1090 @@ def exec_storm_class(klass, jvmtype="-server", jvmopts=[], extrajars=[], args=[]
...
+    and shows query plan instead of submitting topology.
+    """
+
+    sub_parser = subparsers.add_parser("sql", help=command_help, formatter_class=SortingHelpFormatter)
+
+    add_client_jar_options(sub_parser)
+
+    sub_parser.add_argument("sql_file", metavar="sql-file")
+    sub_parser.add_argument(
+        "topology_name", metavar="topology-name", help="should be --explain to activate explain mode"
```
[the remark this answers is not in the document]
d2r: This has been addressed by implementing the mutually exclusive groups.

bin/storm.py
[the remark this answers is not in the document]
d2r: Yes this was fixed.

bin/storm.py
d2r: Yes this has been resolved already by keeping the needed checks.

PR #2931, opened 2019-01-03
STORM-3307: Fixes error time on UI for time of last error on componen…

govind-menon:
…t page


kishorvpatil:
@govind-menon  Thank you for the patch.

PR #2932, opened 2019-01-03
STORM-3274: Adds mock to Travis pipeline so it can run Python tests f…

govind-menon:
…or cli migration


srdo:
I think the build is failing because requirements.txt is being caught by the rat check. 

The change itself looks fine. I'm not sure it makes sense by itself though, maybe it should just be part of https://github.com/apache/storm/pull/2930?

govind-menon:
@srdo Will do

PR #2934, opened 2019-01-07
STORM-3310: Make JCQueueTest wait for consumer to read all queued ite…

srdo:
…ms before terminating

https://issues.apache.org/jira/browse/STORM-3310

I've only seen the test fail once, but it should be pretty easy to see that the test has a race. The tests all expect some messages to go through the queue, but the `run` method doesn't make sure the producers have actually produced anything, and even if they do, the consumer may be shut down before it can read the messages the producers put in the queue.


srdo:
I'd recommend looking at the diff via https://github.com/apache/storm/pull/2934/files?w=1, since the indentation changes make it hard to tell what changed.

PR #2935, opened 2019-01-09
Doc clarification on returning `null` from RecordTranslator.apply()

esamson:
According to KafkaSpout implementation, record is discarded if
Builder.setEmitNullTuples is set to `false`, instead of `true`.


srdo:
+1

srdo:
Thanks @esamson, merged to master

PR #2936, opened 2019-01-09
STORM-3312: Upgrade Guava to latest version where possible

srdo:
https://issues.apache.org/jira/browse/STORM-3312

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -276,8 +276,8 @@
         <httpclient.version>4.5.6</httpclient.version>
         <jctools.version>2.0.1</jctools.version>
         <jgrapht.version>0.9.0</jgrapht.version>
-        <guava.version>16.0.1</guava.version>
-        <auto-service.version>1.0-rc3</auto-service.version>
+        <guava.version>27.0.1-jre</guava.version>
+        <auto-service.version>1.0-rc4</auto-service.version>
         <netty.version>4.1.30.Final</netty.version>
         <sysout-over-slf4j.version>1.0.2</sysout-over-slf4j.version>
         <log4j-over-slf4j.version>1.6.6</log4j-over-slf4j.version>
--- a/examples/storm-hdfs-examples/pom.xml
+++ b/examples/storm-hdfs-examples/pom.xml
@@ -26,6 +26,11 @@
     </parent>
 
     <artifactId>storm-hdfs-examples</artifactId>
+    
+    <properties>
+        <!-- Required downgrade by hadoop-hdfs 2.6.1 -->
+        <guava.version>16.0.1</guava.version>
+    </properties>
 
     <dependencies>
         <dependency>
@@ -38,6 +43,11 @@
             <groupId>org.apache.storm</groupId>
             <artifactId>storm-hdfs</artifactId>
             <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+            <version>${guava.version}</version>
         </dependency>
     </dependencies>
 
--- a/examples/storm-hive-examples/pom.xml
+++ b/examples/storm-hive-examples/pom.xml
@@ -26,6 +26,11 @@
     </parent>
 
     <artifactId>storm-hive-examples</artifactId>
+    
+    <properties>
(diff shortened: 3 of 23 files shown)


examples/storm-hbase-examples/pom.xml
```diff
@@ -26,6 +26,11 @@
     </parent>
 
     <artifactId>storm-hbase-examples</artifactId>
+    
+    <properties>
+        <!-- Required downgrade by hbase-client 1.4.4 -->
```
HeartSaVioR: Just curious, how did you explore whether downgrade is required?

srdo: It's been a while, so I might be misremembering, but I believe I used `mvn dependency:tree` and checked whether each module had any dependencies that used Guava less than version 21.0. According to the Guava README, that was the last version to remove a non-beta API (https://github.com/google/guava#important-warnings point 2). My understanding is that Guava versions are backwards compatible starting at 21.0.

The downgrade may not actually be required, but I don't want to do the testing required to figure that out. At some point someone will upgrade the direct dependency, and then we can remove the downgrade.

srdo:
Yes, let's wait. 2.0.0 is big enough already :)

examples/storm-hbase-examples/pom.xml
HeartSaVioR: Thanks for explaining! I would also think this would be our best try with limited resource.

HeartSaVioR:
@srdo Could you rebase? I think this can be handled along with STORM-3349 and STORM-3350, and still have a slot to include in 2.0.0.

srdo:
I'll rebase soon, but I'd like to merge the Hadoop stuff first. It's likely to affect where we need to manually downgrade Guava.

HeartSaVioR:
Yeah sure. That sounds better.

pom.xml
```diff
@@ -266,8 +266,8 @@
         <httpclient.version>4.5.6</httpclient.version>
         <jctools.version>2.0.1</jctools.version>
         <jgrapht.version>0.9.0</jgrapht.version>
-        <guava.version>16.0.1</guava.version>
-        <auto-service.version>1.0-rc3</auto-service.version>
+        <guava.version>27.0.1-jre</guava.version>
+        <auto-service.version>1.0-rc4</auto-service.version>
```
d2r: Just curious, why is the version changed for `auto-service` here? Is it related to guava?

srdo: Yes. Rc3 uses Guava 19, while 4 uses 23.5. I think as of Guava 21, there have been no breaking changes going forward.

srdo:
Test failure is caused by the issue fixed in https://github.com/apache/storm/pull/2943. That particular test has a habit of hitting the failing assertion in Slot for some reason.

pom.xml
d2r: OK makes sense.

HeartSaVioR:
Just reran the build which failed. Will +1 again and merge once the build passes.

srdo:
@HeartSaVioR Thanks for reviewing. I'll rebase this and other PRs now that the Slot fix is merged. Otherwise I suspect the test will keep failing.

HeartSaVioR:
@srdo Ah yes right. Travis CI doesn't seem to try to merge the change with latest master. Maybe better to have that but that would not be easy with free CI options. Will see whether tests on PRs are passing.

srdo:
I was wrong, it sounds like Travis does merge with master before building https://docs.travis-ci.com/user/pull-requests/#how-pull-requests-are-built. Still have to trigger the builds manually after changes on master though.

HeartSaVioR:
Good to hear if Travis CI can deal with it (TBH a bit confused here - not sure why I thought it didn't): that means we could rerun the build again just before merging to ensure it also works with latest master.

srdo:
Yes. Might be good if PRs are ready to merge for a while, but most of the time I think PRs are only made ready to merge shortly before they actually get merged, so in most cases there shouldn't be a need to do it.

HeartSaVioR:
Sure, not intended to bring more steps/process here. We just found safer way to merge. :)

srdo:
:thumbsup:

PR #2939, opened 2019-01-17
STORM-3315: Upgrade to Kryo 4

srdo:
https://issues.apache.org/jira/browse/STORM-3315

Tested compatibility with Kryo 3.0.3 serialization by running the TVL topology for a couple of minutes with two supervisors, one using the unmodified Storm code, and one running this patch.

### DIFF
--- a/external/storm-kinesis/src/main/java/org/apache/storm/kinesis/spout/KinesisConnectionInfo.java
+++ b/external/storm-kinesis/src/main/java/org/apache/storm/kinesis/spout/KinesisConnectionInfo.java
@@ -84,6 +84,7 @@ public class KinesisConnectionInfo implements Serializable {
 
     private byte[] getKryoSerializedBytes (final Object obj) {
         final Kryo kryo = new Kryo();
+        kryo.getFieldSerializerConfig().setOptimizedGenerics(true);
         final ByteArrayOutputStream os = new ByteArrayOutputStream();
         final Output output = new Output(os);
         kryo.setInstantiatorStrategy(new StdInstantiatorStrategy());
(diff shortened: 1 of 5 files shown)


srdo:
I think making an automated test for this would be a big hassle. The test would need to download Storm 1.2.2 (or some other version using Kryo 3), and set it up in a cluster with a build of the current Storm version. We'd then need to run something like our integration test, and scan the logs for errors.

Regarding how I decided which changes to make for Kryo 4, I'm just trusting the Kryo 4 release notes, which indicates that Kryo 3 and 4 are compatible if you set `setOptimizedGenerics(true)` (see https://github.com/EsotericSoftware/kryo/releases/tag/kryo-parent-4.0.0)

HeartSaVioR:
Sorry I think I wasn't clear enough. What I meant was serializing TreeMap to simulate SO issue referred in STORM-3315 instead of checking compatibility with old version of Storm (or Kyro).

srdo:
Sorry, I misunderstood.

I don't think we should add a test in Storm. It doesn't sound like the issue has anything to do with Kryo's configuration, but was simply due to a bug in Kryo that has been fixed. 

I think the issue was fixed in https://github.com/EsotericSoftware/kryo/pull/342, which includes tests.

HeartSaVioR:
OK. I don't have strong voice on adding (possibly redundant) test. Just would like to confirm upgrading Kyro version would resolve the ask on SO, because this patch started from there. Have you go through manual test?

srdo:
No. I'll try it out manually.

srdo:
I can't reproduce the issue based on the description in the Kryo issue. Storm seems to serialize the example fine even with Kryo 3.0.3. 

Maybe we can get the person on SO with the problem to try this branch. Otherwise, I'm not sure how to proceed.

srdo:
Closing, out of date. If this is interesting to anyone, feel free to use this PR for future work.

PR #2943, opened 2019-01-21
STORM-3319: Fix failing assertion in Slot

srdo:
https://issues.apache.org/jira/browse/STORM-3319

In addition to fixing the failing assertion, I've also fixed a deadlock between AsyncLocalizer and Slot.
The latch AsyncLocalizer is waiting for is counted down by Slot, but Slot may be stuck trying to enter other synchronized methods on the blob instance.


srdo:
Rebasing to master to check if this fixes test instability

srdo:
Looks like it fixes the instability.

PR #2945, opened 2019-01-21
STORM-3321: Fix race in LocalCluster regarding Nimbus leadership, red…

srdo:
…uce poll timers for Nimbus and supervisor to speed up tests and avoid timeouts

https://issues.apache.org/jira/browse/STORM-3321

I also updated an error message to print how long Slots take to shut down, previously it would print 1000ms no matter how long the shutdown took.

The timer interval reductions in LocalCluster are set to only happen if time simulation is disabled. There are some tests (e.g. nimbus_test.clj) that use time simulation and want those timers to be longer, so it seemed easier to just leave it alone for them.

### DIFF
--- a/storm-client/src/jvm/org/apache/storm/nimbus/ILeaderElector.java
+++ b/storm-client/src/jvm/org/apache/storm/nimbus/ILeaderElector.java
@@ -15,6 +15,8 @@ package org.apache.storm.nimbus;
 import java.io.Closeable;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.TimeUnit;
+import org.apache.storm.shade.com.google.common.annotations.VisibleForTesting;
 
 /**
  * The interface for leader election.
@@ -52,6 +54,15 @@ public interface ILeaderElector extends Closeable {
      * @return the current leader's address , may return null if no one has the lock.
      */
     NimbusInfo getLeader();
+    
+    /**
+     * Wait for the caller to gain leadership. This should only be used in single-Nimbus clusters, and is only useful to allow testing
+     * code to wait for a LocalCluster's Nimbus to gain leadership before trying to submit topologies.
+     *
+     * @return true is leadership was acquired, false otherwise
+     */
+    @VisibleForTesting
+    boolean awaitLeadership(long timeout, TimeUnit timeUnit) throws InterruptedException;
 
     /**
      * Get list of current nimbus addresses.
--- a/storm-client/src/jvm/org/apache/storm/testing/FixedTupleSpout.java
+++ b/storm-client/src/jvm/org/apache/storm/testing/FixedTupleSpout.java
@@ -22,7 +22,6 @@ import org.apache.storm.task.TopologyContext;
 import org.apache.storm.topology.IRichSpout;
 import org.apache.storm.topology.OutputFieldsDeclarer;
 import org.apache.storm.tuple.Fields;
-import org.apache.storm.utils.Utils;
 
 import static org.apache.storm.utils.Utils.get;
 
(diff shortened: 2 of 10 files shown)


storm-server/src/main/java/org/apache/storm/LocalCluster.java
```diff
@@ -231,6 +233,10 @@ private LocalCluster(Builder builder) throws Exception {
             } else {
                 this.clusterState = builder.clusterState;
             }
+            if (!Time.isSimulating()) {
+                //Ensure Nimbus assigns topologies as quickly as possible
+                conf.put(DaemonConfig.NIMBUS_MONITOR_FREQ_SECS, 1);
```
HeartSaVioR: Is this change also useful for other cases (non-test) as well? If we are unsure about it, could we setup global config for non time simulating test and add this to there instead?

storm-server/src/main/java/org/apache/storm/LocalCluster.java
```diff
@@ -690,6 +700,10 @@ public synchronized Supervisor addSupervisor(Number ports, Map<String, Object> c
         }
         superConf.put(Config.STORM_LOCAL_DIR, tmpDir.getPath());
         superConf.put(DaemonConfig.SUPERVISOR_SLOTS_PORTS, portNumbers);
+        if (!Time.isSimulating()) {
+            //Monitor for assignment changes as often as possible, so e.g. shutdown happens as fast as possible.
```
HeartSaVioR: same here

storm-client/src/jvm/org/apache/storm/nimbus/ILeaderElector.java
```diff
@@ -52,6 +54,15 @@
...
     NimbusInfo getLeader();
+    
+    /**
+     * Wait for the caller to gain leadership. This should only be used in single-Nimbus clusters, and is only useful to allow testing
+     * code to wait for a LocalCluster's Nimbus to gain leadership before trying to submit topologies.
+     *
+     * @return true is leadership was acquired, false otherwise
+     */
+    @VisibleForTesting
+    boolean awaitLeadership(long timeout, TimeUnit timeUnit) throws InterruptedException;
```
HeartSaVioR: I assume we don't think ILeaderElector as user facing (public) API so don't mind about backward compatibility.

srdo: No, I don't think this API is supposed to be used by anyone other than Nimbus.

storm-server/src/main/java/org/apache/storm/LocalCluster.java
srdo: I think this is still useful outside testing, since the topology will start faster, and topologies will usually be submitted right after bootup of the local Nimbus. Are we intending for people to use LocalCluster for cases other than testing though?

srdo: By "testing" I mean both automatic tests and manual tests

HeartSaVioR: OK makes sense. Doesn't seem to be a big deal as I think about this again.

storm-client/src/jvm/org/apache/storm/testing/FixedTupleSpout.java
```diff
@@ -131,8 +130,6 @@ public void nextTuple() {
             String id = UUID.randomUUID().toString();
             _pending.put(id, ft);
             _collector.emit(ft.stream, ft.values, id);
-        } else {
-            Utils.sleep(100);
```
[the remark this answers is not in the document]
srdo: Storm already sleeps between nextTuple calls if nothing is emitted based on the sleep strategy. The extra sleep here doesn't do much for us I think, and manually sleeping in nextTuple is a bad habit I'd like to discourage (it blocks other message processing for the spout, e.g. ack handling).

See https://github.com/apache/storm/blob/b07413670fa62fec077c92cb78fc711c3bda820c/storm-client/src/jvm/org/apache/storm/executor/spout/SpoutExecutor.java#L224

PR #2950, opened 2019-01-31
STORM-3327: Use UTF-8 for URL encoding in storm-webapp

srdo:
https://issues.apache.org/jira/browse/STORM-3327

### DIFF
--- a/storm-client/src/jvm/org/apache/storm/utils/Utils.java
+++ b/storm-client/src/jvm/org/apache/storm/utils/Utils.java
@@ -32,14 +32,17 @@ import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
 import java.io.OutputStreamWriter;
 import java.io.Serializable;
+import java.io.UnsupportedEncodingException;
 import java.lang.management.ManagementFactory;
 import java.lang.management.ThreadInfo;
 import java.net.InetAddress;
 import java.net.ServerSocket;
 import java.net.URL;
 import java.net.URLDecoder;
+import java.net.URLEncoder;
 import java.net.UnknownHostException;
 import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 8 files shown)


storm-client/src/jvm/org/apache/storm/utils/Utils.java
```diff
@@ -246,7 +248,11 @@ replace below code with split(",") once 'storm.cmd' is fixed to send url-encoded
               */
             String[] configs = commandOptions.split(",(?![^\\[\\]{}]*(]|}))");
             for (String config : configs) {
-                config = URLDecoder.decode(config);
+                try {
+                    config = URLDecoder.decode(config, StandardCharsets.UTF_8.name());
```
HeartSaVioR: Could we add this to util class so that we don't need to specify `StandardCharsets.UTF_8.name()` all the time? Maybe two versions throwing original UnsupportedEncodingException and RuntimeException wrapping UnsupportedEncodingException.

storm-webapp/src/main/java/org/apache/storm/daemon/ui/UIHelpers.java
```diff
@@ -1312,21 +1326,27 @@ private static Double nullToZero(Double value) {
...
+            throw Utils.wrapInRuntime(e);
+        }
         result.put("uptime", prettyUptimeSec(executorSummary.get_uptime_secs()));
         result.put("uptimeSeconds", executorSummary.get_uptime_secs());
         String host = executorSummary.get_host();
         result.put("host", host);
         int port = executorSummary.get_port();
         result.put("port", port);
+        
+        ComponentAggregateStats componentAggregateStats = executorAggregateStats.get_stats();
```
HeartSaVioR: Are below changes relevant / necessary? Because it looks like not relevant to this issue.

srdo: Checkstyle started complaining that the variable declarations were too far from first use.

storm-client/src/jvm/org/apache/storm/utils/Utils.java
srdo: Good idea. Fixed.

storm-webapp/src/main/java/org/apache/storm/daemon/ui/UIHelpers.java
HeartSaVioR: Ah yes thanks for fixing this too!

PR #2952, opened 2019-02-01
STORM-3329 move HttpForwardingMetricsConsumer to storm-client

agresch:


Ethanlm:
I want to ping @revans2  to take a look since last time he moved this class from metrics to storm-loadgen. I wonder if there is any concerns.

revans2:
My concern would be with how it is intended to be used.  By keeping it in the loadgen package it is not likely to be used in production, but if it is a part of storm-client we need to be sure that it is obvious that it is not production ready.  I am also a little nervous that it might expose some http server classes that we are shading.

PR #2957, opened 2019-02-08
STORM-3335 allow timing out when scheduling a topology

agresch:
This is based on a couple of internal PRs @govind-menon committed to our older version of storm.

### DIFF
--- a/storm-server/src/main/java/org/apache/storm/scheduler/resource/ResourceAwareScheduler.java
+++ b/storm-server/src/main/java/org/apache/storm/scheduler/resource/ResourceAwareScheduler.java
@@ -17,6 +17,11 @@ import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.stream.Collectors;
 import org.apache.storm.Config;
 import org.apache.storm.DaemonConfig;
@@ -45,6 +50,8 @@ public class ResourceAwareScheduler implements IScheduler {
     private ISchedulingPriorityStrategy schedulingPriorityStrategy;
     private IConfigLoader configLoader;
     private int maxSchedulingAttempts;
+    private int schedulingTimeoutSeconds;
+    private ExecutorService backgroundScheduling;
 
     private static void markFailedTopology(User u, Cluster c, TopologyDetails td, String message) {
         markFailedTopology(u, c, td, message, null);
@@ -78,6 +85,15 @@ public class ResourceAwareScheduler implements IScheduler {
         configLoader = ConfigLoaderFactoryService.createConfigLoader(conf);
         maxSchedulingAttempts = ObjectReader.getInt(
             conf.get(DaemonConfig.RESOURCE_AWARE_SCHEDULER_MAX_TOPOLOGY_SCHEDULING_ATTEMPTS), 5);
+        schedulingTimeoutSeconds = ObjectReader.getInt(
+                conf.get(DaemonConfig.SCHEDULING_TIMEOUT_SECONDS_PER_TOPOLOGY), 60);
+        backgroundScheduling = Executors.newFixedThreadPool(1);
+    }
+
+    @Override
+    public void cleanup() {
+        LOG.info("Cleanup ResourceAwareScheduler scheduler");
+        backgroundScheduling.shutdown();
     }
 
     @Override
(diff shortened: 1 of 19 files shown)


storm-server/src/main/java/org/apache/storm/scheduler/resource/ResourceAwareScheduler.java
```diff
@@ -78,6 +85,9 @@ public void prepare(Map<String, Object> conf) {
         configLoader = ConfigLoaderFactoryService.createConfigLoader(conf);
         maxSchedulingAttempts = ObjectReader.getInt(
             conf.get(DaemonConfig.RESOURCE_AWARE_SCHEDULER_MAX_TOPOLOGY_SCHEDULING_ATTEMPTS), 5);
+        schedulingTimeoutSeconds = ObjectReader.getInt(
+                conf.get(DaemonConfig.SCHEDULING_TIMEOUT_SECONDS_PER_TOPOLOGY), 60);
+        backgroundScheduling = Executors.newFixedThreadPool(1);
```
srdo: I think we need to add a method that allows cleanup of `IScheduler`, similar to the `cleanup` method we have on `IBolt`. This executor needs to be shut down, otherwise the thread will leak if this class gets used in tests.

storm-server/src/main/java/org/apache/storm/scheduler/resource/ResourceAwareScheduler.java
```diff
@@ -136,10 +146,24 @@ private void scheduleTopology(TopologyDetails td, Cluster cluster, final User to
...
+                    () -> finalRasStrategy.schedule(toSchedule, td)
+                );
+                try {
+                    result = schedulingFuture.get(schedulingTimeoutSeconds, TimeUnit.SECONDS);
+                } catch (TimeoutException te) {
+                    markFailedTopology(topologySubmitter, cluster, td, "Scheduling took too long for "
+                            + td.getId() + " using strategy " + rasStrategy.getClass().getName() + " timeout after "
+                            + schedulingTimeoutSeconds + " seconds.");
+                    schedulingFuture.cancel(true);
+                    rasStrategy.stop();
```
srdo: Consider dropping the `stop` method and make the scheduler check for interrupts instead. You're already sending an interrupt to it via the `cancel` call above.

storm-server/src/main/java/org/apache/storm/scheduler/resource/ResourceAwareScheduler.java
```diff
@@ -136,10 +146,24 @@ private void scheduleTopology(TopologyDetails td, Cluster cluster, final User to
...
             try {
-                SchedulingResult result = rasStrategy.schedule(toSchedule, td);
+                SchedulingResult result = null;
+                Future<SchedulingResult> schedulingFuture = backgroundScheduling.submit(
+                    () -> finalRasStrategy.schedule(toSchedule, td)
+                );
+                try {
+                    result = schedulingFuture.get(schedulingTimeoutSeconds, TimeUnit.SECONDS);
+                } catch (TimeoutException te) {
+                    markFailedTopology(topologySubmitter, cluster, td, "Scheduling took too long for "
```
srdo: Nit: Consider referencing the config parameter users should change if they want to raise the timeout, so it is obvious what users can do when they get this message.

agresch:
@srdo - I made the changes you suggested.

storm-server/src/main/java/org/apache/storm/scheduler/DefaultScheduler.java
```diff
@@ -101,6 +101,10 @@ public void prepare(Map<String, Object> conf) {
         //noop
     }
 
+    @Override
+    public void cleanup() {
+    }
```
danny0405: Most of the cleanup() method seems a no-op, can we define a default implementation of it ?

srdo:
@agresch Thanks, it looks great. I'm wondering if we can put the setup and cleanup in `@Before/@After` for many of the tests? Ideally we'd avoid leaking threads even when the tests fail, so cleanup should be either in an `@After` or in try-finally.

agresch:
somehow missed some test files, more changes to come...

storm-server/src/main/java/org/apache/storm/scheduler/DefaultScheduler.java
agresch: @danny0405 - Updated.  Please take a look.

agresch:
@srdo - please look at the latest couple commits for the test cleanup.

srdo:
+1, thanks for addressing comments.

storm-server/src/main/java/org/apache/storm/scheduler/resource/strategies/scheduling/ConstraintSolverStrategy.java
```diff
@@ -331,6 +331,10 @@ protected SolverResult backtrackSearch(SearcherState state) {
             return new SolverResult(state, false);
         }
 
+        if (Thread.currentThread().isInterrupted()) {
```
Ethanlm: Why do we need to this? (I don't know how this is used and trying to learn)

agresch: My remembrance from previous googling is that a cancel on the future causes the thread to be interrupted, but it is up to that thread to respond to the interrupt.  If the scheduling thread is stuck in a long loop, it will just keep running.

Ethanlm: Thanks!

agresch:
@Ethanlm - please check again.  Thanks.

srdo:
@agresch Thank you for your patience. Please squash to one commit, and I'll merge.

agresch:
@srdo - squashed

PR #2958, opened 2019-02-19
STORM-3341: DRPCSpout should check before connecting to DrpcServer

MichealShin:
Now DRPCSpout always try to connect all Drpc Servers in the method 'open'. If one Drpc server shut down, the topology will restart over and over again.

DRPCSpout should check before connecting to DrpcServer.


kishorvpatil:
If any one DRPC daemon out of n is down/unavailable during opening of Spout, the DRPCSpout would not serve any request in the future.

PR #2963, opened 2019-02-25
[MINOR] Update release verify script - 'sha' to 'sha512'

HeartSaVioR:
Apache Storm's release process has changed to use sha512 instead of sha, but it is not reflected to release verify script. This patch applies to the change.


srdo:
+1

PR #2970, opened 2019-03-07
STORM-3354: Fix Nimbus not reentering leadership election in some cas…

srdo:
…es, and quit leadership election cleanly when Nimbus shuts down

https://issues.apache.org/jira/browse/STORM-3354

There are two issues here. 

First that the LeaderElectorImp doesn't close the LeaderLatch when closed, which can cause a crash during Nimbus shutdown. 

Second that if LeaderListenerCallback.leaderCallback decides not to be leader due to missing blobs, I believe Nimbus will stop trying to acquire leadership entirely. I haven't seen this happen, but I can't find any code that restarts the LeaderLatch.

### DIFF
--- a/storm-server/src/main/java/org/apache/storm/nimbus/LeaderListenerCallback.java
+++ b/storm-server/src/main/java/org/apache/storm/nimbus/LeaderListenerCallback.java
@@ -23,6 +23,7 @@ import javax.security.auth.Subject;
 import com.codahale.metrics.Meter;
 import org.apache.commons.io.IOUtils;
 import org.apache.storm.Config;
+import org.apache.storm.DaemonConfig;
 import org.apache.storm.blobstore.BlobStore;
 import org.apache.storm.blobstore.InputStreamWithMeta;
 import org.apache.storm.cluster.ClusterUtils;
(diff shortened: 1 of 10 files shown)


storm-server/src/main/java/org/apache/storm/nimbus/LeaderListenerCallback.java
```diff
@@ -60,31 +62,35 @@
...
      * @param zk zookeeper CuratorFramework client
-     * @param leaderLatch LeaderLatch
+     * @param leaderElector Leader elector
      * @param blobStore BlobStore
      * @param tc TopoCache
      * @param clusterState IStormClusterState
      * @param acls zookeeper acls
      */
-    public LeaderListenerCallback(Map conf, CuratorFramework zk, LeaderLatch leaderLatch, BlobStore blobStore,
+    public LeaderListenerCallback(Map conf, CuratorFramework zk, BlobStore blobStore, ILeaderElector leaderElector,
```
HeartSaVioR: nit: param orders changed while it's not reflected to javadoc

srdo: Fixed.

PR #2974, opened 2019-03-12
STORM-3355: Use supervisor.worker.shutdown.sleep.secs to set worker s…

srdo:
…uicide delay to allow users to configure how long they're willing to wait for orderly shutdown

https://issues.apache.org/jira/browse/STORM-3355

This will not be useful on Windows, as there is no easy way that I can see for shutting down JVMs cleanly. I dug around for a while and it looks like the JVM devs also haven't found a reasonable way to do orderly shutdowns on Windows https://bugs.openjdk.java.net/browse/JDK-8056139. We can probably be excused for not doing better.

On Linux it works as expected. Tested by running ExclamationTopology with a `cleanup` that sleeps for 15 seconds, and setting the timeout parameter to 30 seconds. The worker was allowed to shut down cleanly.

### DIFF
--- a/storm-client/src/jvm/org/apache/storm/daemon/worker/Worker.java
+++ b/storm-client/src/jvm/org/apache/storm/daemon/worker/Worker.java
@@ -127,7 +127,9 @@ public class Worker implements Shutdownable, DaemonCommon {
         Worker worker = new Worker(conf, null, stormId, assignmentId, Integer.parseInt(supervisorPort),
                                    Integer.parseInt(portStr), workerId);
         worker.start();
-        Utils.addShutdownHookWithForceKillIn1Sec(worker::shutdown);
+        int workerShutdownSleepSecs = ObjectReader.getInt(conf.get(Config.SUPERVISOR_WORKER_SHUTDOWN_SLEEP_SECS));
+        LOG.info("Adding shutdown hook with kill in {} secs", workerShutdownSleepSecs);
+        Utils.addShutdownHookWithDelayedForceKill(worker::shutdown, workerShutdownSleepSecs);
     }
 
     public void start() throws Exception {
--- a/conf/defaults.yaml
+++ b/conf/defaults.yaml
@@ -161,7 +161,7 @@ supervisor.run.worker.as.user: false
 supervisor.worker.start.timeout.secs: 120
 #how long between heartbeats until supervisor considers that worker dead and tries to restart it
 supervisor.worker.timeout.secs: 30
-#how many seconds to sleep for before shutting down threads on worker
+#How many seconds to allow for graceful worker shutdown when killing workers before resorting to force kill
 supervisor.worker.shutdown.sleep.secs: 3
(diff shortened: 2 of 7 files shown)


storm-client/src/jvm/org/apache/storm/daemon/worker/Worker.java
```diff
@@ -127,7 +127,9 @@ public static void main(String[] args) throws Exception {
         Worker worker = new Worker(conf, null, stormId, assignmentId, Integer.parseInt(supervisorPort),
                                    Integer.parseInt(portStr), workerId);
         worker.start();
-        Utils.addShutdownHookWithForceKillIn1Sec(worker::shutdown);
+        LOG.info("Adding shutdown hook with kill in {} secs", ObjectReader.getInt(conf.get(Config.SUPERVISOR_WORKER_SHUTDOWN_SLEEP_SECS)));
```
HeartSaVioR: `ObjectReader.getInt(conf.get(Config.SUPERVISOR_WORKER_SHUTDOWN_SLEEP_SECS))` will be executed twice once end users configure to print INFO or lower. Would be better to have separate variable here.

srdo: Good catch, will fix

srdo:
I'm not sure it's an issue that it's non-deterministic. The force kill only happens if the worker fails to shut down within the time limit, which should hopefully be rare. If the time limit is hit, does it matter whether the supervisor `kill -9`'s the worker JVM, or whether the worker JVM suicides?

I assumed the reason we want the worker to suicide is in case the supervisor issues a soft kill, and then crashes. I'm not sure how the supervisor handles it if it is restarted and the worker is a "zombie" hanging in the middle of shutdown?

If that's not the reason, I'm not sure. Relying on the supervisor's kill might be fine?

HeartSaVioR:
https://github.com/apache/storm/commit/2c2570fcd707b71d69db81736ba703707cee3ae7

I found origin commit to introduce 1 sec suicide but Github fails to link the PR and I can't get the reason 
of doing it from only commit message. Maybe it's simply (and blindly) added for all of JVM processes? Daemon processes definitely need it.

I wasn't sure about worker, but if we concern about supervisor being crashed just after issuing soft kill, that would make sense.

Btw, I think you're right. Once it reaches force kill it wouldn't matter who triggers it.

srdo:
The reason for replacing `Runtime.halt` with `Runtime.exit` everywhere is most likely that `Runtime.exit` invokes shutdown hooks before quitting, where `Runtime.halt` just kills the JVM immediately. It looks to me like that commit is an attempt to do best-effort cleanup (e.g. running `Worker.shutdown`) on shutdown, where the previous behavior was to just suicide the JVM immediately.

Since the behavior before that commit was for the worker to just die immediately on shutdown, I think the shutdown hook was probably added in the worker because that was the same way it was done for the daemons.

I'm still not sure the shutdown hook is necessary for the worker, but I don't think it hurts either.

PR #2976, opened 2019-03-13
STORM-3357: Bump Clojure version to 1.10

srdo:
https://issues.apache.org/jira/browse/STORM-3357

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -257,7 +257,7 @@
         <license.skipAggregateAddThirdParty>true</license.skipAggregateAddThirdParty>
 
         <!-- dependency versions -->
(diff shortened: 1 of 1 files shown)


HeartSaVioR:
@kishorvpatil Could you please review this? I don't have any clojure topology for production use case, so would like to double check if we have the env.

kishorvpatil:
> @kishorvpatil Could you please review this? I don't have any clojure topology for production use case, so would like to double check if we have the env.

Let me test this and get back to you.

HeartSaVioR:
@kishorvpatil Thanks for the quick update!

PR #2978, opened 2019-03-21
STORM-3311: Use Java 7 Files API for file IO

srdo:
https://issues.apache.org/jira/browse/STORM-3311

Primary changes:
* Replace File with Path
* Try to replace String with Path in as many places as it makes sense, and get rid of as much manual path string construction as possible.
* Ensure IO streams are try-with-resource'd where applicable
* Ensure Files.list return stream is try-with-resource'd
* Use nio Files API to open files instead of java.io APIs
* Ensure we are explicit about which charset we use with IO streams/string-byte conversion.
* Atomic file move is possible on Windows under some conditions, e.g. the move cannot be across device/partition boundaries. Make AdvancedFsOps offer best-effort atomic move for Windows as well.
* Files.exists is slow on Java 8 https://bugs.openjdk.java.net/browse/JDK-8153414. Replaced all occurrences with File.exists.

In order to avoid older file APIs creeping in again I've added some checkstyle rules to try to catch them.

The changes here aren't a perfect fix for Windows compatibility, since Windows' file deletion handling even with FILE_SHARE_DELETE is kind of bad. We can delete files with open handles, but they won't disappear from the parent directory's file list until all handles are closed. This prevents deletion of the parent directory. It is possible to work around this (https://boostgsoc13.github.io/boost.afio/doc/html/afio/FAQ/deleting_open_files.html) but I've chosen not to, since I'm not comfortable about moving worker files (maybe sensitive?) into the tmp dir before deletion.

Windows can do POSIX-style deletes with the behavior we want, but it's currently not exposed in a way that isn't a huge pain to access https://wpdev.uservoice.com/forums/110705-universal-windows-platform/suggestions/37133743-add-file-disposition-posix-semantics-to-public-api. Hopefully that API will become more accessible at some point, and we can look at making use of it then.


examples/storm-hdfs-examples/src/main/java/org/apache/storm/hdfs/bolt/SequenceFileTopology.java
```diff
@@ -14,6 +14,8 @@
 
 import java.io.FileInputStream;
```
kishorvpatil: Remove unused imports?

srdo: Yes, will get this done. I'll probably use my IDE to organize the imports, so it might bloat the diff. Will put it in a new commit.

srdo:
I'm going to postpone this until 2.0.0 is released. I'd rather not pile this change on top of that release.

examples/storm-hdfs-examples/src/main/java/org/apache/storm/hdfs/bolt/SequenceFileTopology.java
srdo: Should be fixed now, fixed imports for all modules. The changes are in the second commit.

srdo:
This is ready for review, since Storm 2.0.0 has released.

srdo:
Closing, out of date

PR #2980, opened 2019-03-25
STORM-3361: Include dependency license file in distributions, clean u…

srdo:
…p license merges a bit, make storm-dist use NOTICE/LICENSE from project root

https://issues.apache.org/jira/browse/STORM-3361

The change here is so we include a DEPENDENCY-LICENSES file in our distributions and source listing our dependencies and their licenses.

Regarding the new file, it is generated by running the license plugin as described in DEVELOPER. Ideally we will regenerate it before each release. It is being committed to git so it will be present in source releases. If we can generate it instead as part of the storm-dist/source build, that would probably be better though. 

There were NOTICE/LICENSE files in storm-dist/binary that are being included in distributions. To me it looks like they are supposed to be the same files as the ones we have in the project root, but I don't see why we would want to have a copy in storm-dist instead of just reading the root files.


srdo:
I would like to eventually be able to make the license file a little nicer. We currently have a bunch of extra headers for dependencies with multiple licenses. Ideally we could shove all Apache licensed dependencies under the Apache license header, even if they are dual licensed under GPL as well. I don't think there's a way to do it with the plugin currently though.

srdo:
Here's what changed:

* Some dependencies (JUnit) were compile scope when they shouldn't be.
* Removed the NOTICE/LICENSE files from storm-dist, and replaced them with updated versions called LICENSE-binary/NOTICE-binary in the project root.
* Made some adjustments to storm-dist to make it easier to get an idea of the dependencies being included. Specifically, I've added modules to storm-dist to make dependencies of final-package visible to the license plugin. Before, we were just copying files from unrelated modules. This should allow us to use the license plugin to get a mostly complete list of dependencies for storm-dist/binary. I believe the only missing licenses are shaded-deps, which were easily added using `mvn dependency:tree`.
* NOTICE-binary is now the source NOTICE, plus appended NOTICEs from all storm-dist/binary dependencies.
* LICENSE-binary is now source LICENSE, plus slightly edited output from the license plugin for storm-dist/binary.

As far as I can tell, the only part missing is regarding copyright notices. Many licenses (e.g. MIT) start with a copyright declaration, which must be included in downstream work. I'm not sure whether we can get that automatically.

srdo:
I think the copyright notice issue is resolved. The license check plugin can download licenses. Did that, deleted the duplicates and added the ones the license plugin failed to download. The licenses are in the licenses directory, which is now included in both source and binary distributions.

srdo:
This is ready for review, will squash once review is done.

srdo:
Just to summarize what the aim of this change is (and hopefully help reviewers):

LICENSE and NOTICE contain licenses and notices for our source, following the guide at http://www.apache.org/dev/licensing-howto.html. These are the L&N included in our storm-dist/source distribution.

LICENSE-binary is LICENSE, plus a list of dependencies of the storm-dist/binary project (i.e. everything we put in the binary distribution), plus the list of shaded artifacts from shaded-deps. This follows the http://www.apache.org/dev/licensing-howto.html#binary guideline. I've put the license texts in the license directory, and mention the link in LICENSE-binary. 

The addition to LICENSE-binary is generated by following the steps in DEVELOPER.md to get the license list for storm-dist/binary, editing the output a bit by hand, and adding the shaded-deps dependencies by hand. The licenses in the `licenses` directory are downloaded by following the steps in DEVELOPER.md to download licenses, then renaming the fetched files to be a bit more readable.

NOTICE-binary is NOTICE, plus concatenated NOTICEs from storm-dist/binary dependencies. The appended text is generated using the steps in DEVELOPER.md (the `collect_license_files` script). The script is a lightly modified version of the script Flink uses https://github.com/apache/flink/blob/master/tools/releasing/collect_license_files.sh. This follows the http://www.apache.org/dev/licensing-howto.html#mod-notice guideline.

LICENSE-binary and NOTICE-binary are renamed and included as LICENSE/NOTICE in the binary distribution.

DEPENDENCY-LICENSES is included in either distribution, and simply lists all Storm's dependencies and their licenses. The licenses are included in the `licenses` directory. The file is there to cover https://apache.org/legal/resolved.html#appropriately-labelled-condition for category B dependencies. I'm not positive we need this file, but it would be to make users aware of category B dependencies we don't distribute directly, but are needed to use e.g. connector modules.

srdo:
Thanks for reviewing @kishorvpatil. Squashed and rebased. Will merge once tests pass.

PR #2981, opened 2019-03-26
Issue tracker [STORM-3362] Solved: eventHubSpout uses a blocking receiver in nextTuple()

ysyyork:


srdo:
Thanks for the contribution.

The change looks fine, but I think we should let people configure the timeout. Could we add it to https://github.com/apache/storm/blob/8d4432233c8776247da520b5935b7eebfc267ba5/external/storm-eventhubs/src/main/java/org/apache/storm/eventhubs/spout/EventHubSpoutConfig.java ?

ysyyork:
Good suggestion. How about this? @srdo

srdo:
LGTM, thanks. Please update the commit message to contain the JIRA issue (see the other PRs for example).

Will merge after the 24 hour waiting period, assuming tests pass :)

ysyyork:
Sure, done

srdo:
Great, +1 pending tests.

srdo:
Test failure is unrelated.

ysyyork:
yeah, i didn't get it either. do I need to do anything?

srdo:
Not for this PR, no.

ysyyork:
Alright thx

external/storm-eventhubs/src/main/java/org/apache/storm/eventhubs/spout/EventHubSpoutConfig.java
```diff
@@ -42,6 +42,7 @@
     // disabling filter
     private String connectionString;
     private String topologyName;
+    private int receiverTimeoutInMillis = 10; // default
```
HeartSaVioR: Is `10ms` a realistic value for timeout? Since it's the default value and end users don't touch unless they face timeout error, the value would need to cover most of cases. We know it should be short to let Spout handles ack/fail fast, but `10ms` sounds too small to reach remote-end.

ysyyork: you are right. I think this depends on the how big is your payload in event hub. eventhub has a 1 MB limit for packet size. If we considering worst case scenario, then the bandwidth requirements are 1MB/0.01sec = 100M/sec which seems a really high requirements. based on this, let's tune this to like 100mills so the bandwidth requirements become 10M/sec. do you think this is reasonable? let me know. My current application, the payload is pretty small. only like several bytes so it doesn't matter.

srdo: I'm not familiar with the EH client library, so I don't know if it fetches data in a background thread, but assuming it only fetches data when polled, latency to EH will be a concern. With a default of 100ms, you will only be able to fetch data at all if the latency is less than 100ms.

The Kafka spout uses 200ms by default, but Kafka can be run "near" the spout (in network hops terms, e.g. in the same LAN), while EH is a hosted service. It might be good to set it conservatively, e.g. to 500-1000ms, and let people fine tune it manually.

ysyyork: sounds good. let me set it to 1 sec then to be safe

PR #2982, opened 2019-03-28
STORM-3363: Migrate Aether to maven-resolver as Aether is donated to ASF

HeartSaVioR:
Please refer [STORM-3363](https://issues.apache.org/jira/browse/STORM-3363) for more details.

Other than migrating Aether to maven-resolver, this patch also contains a minor fix which storm cli incorrectly parses `artifacts` and `artifactRepositories`. Looks like STORM-3274 changed the behavior but not tested the case.
(Oddly my local dev. cannot run tests because of somewhat issue with `mock` and `python3`. I'll see whether the change breaks cli test in Travis CI build, and fix it if any.)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -320,9 +320,8 @@
         <clojure.test.set>!integration.*</clojure.test.set>
         <skipITs>true</skipITs>
 
-        <aetherVersion>1.1.0</aetherVersion>
-        <mavenVersion>3.1.0</mavenVersion>
-        <wagonVersion>1.0</wagonVersion>
+        <maven-resolver.version>1.3.3</maven-resolver.version>
+        <maven.version>3.6.0</maven.version>
         <azure-eventhubs.version>0.13.1</azure-eventhubs.version>
         <jersey.version>2.27</jersey.version>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -320,9 +320,8 @@
         <clojure.test.set>!integration.*</clojure.test.set>
         <skipITs>true</skipITs>
 
-        <aetherVersion>1.1.0</aetherVersion>
-        <mavenVersion>3.1.0</mavenVersion>
-        <wagonVersion>1.0</wagonVersion>
+        <maven-resolver.version>1.3.3</maven-resolver.version>
+        <maven.version>3.6.0</maven.version>
```
srdo: This might be a bit out of scope, but maybe we should use this variable in the enforcer plugin below, so we require at least this Maven version?

HeartSaVioR: Maybe we don't need to align maven version for build and maven version for storm-submit-tool. The propery has been only used for storm-submit-tool.

srdo: Ok. It probably doesn't matter then.

PR #2984, opened 2019-03-29
[STORM-3365] Configuration to disable Topology Lag Monitoring

kishorvpatil:
Sometimes Kafka broker is unreachable/security can cause issue..making UI hang for too long.  So adding feature to allow disabling Topology Lag monitoring.

### DIFF
--- a/conf/defaults.yaml
+++ b/conf/defaults.yaml
@@ -98,6 +98,7 @@ ui.header.buffer.bytes: 4096
 ui.http.creds.plugin: org.apache.storm.security.auth.DefaultHttpCredentialsPlugin
 ui.pagination: 20
 ui.disable.http.binding: true
+ui.disable.spout.lag.monitoring: true
 
 logviewer.port: 8000
 logviewer.childopts: "-Xmx128m"
--- a/storm-server/src/main/java/org/apache/storm/DaemonConfig.java
+++ b/storm-server/src/main/java/org/apache/storm/DaemonConfig.java
@@ -317,6 +317,12 @@ public class DaemonConfig implements Validated {
     public static final String UI_DISABLE_HTTP_BINDING = "ui.disable.http.binding";
 
     /**
+     * This controls whether Storm UI would not monitor Spout lag.
+     */
+    @isBoolean
+    public static final String UI_DISABLE_SPOUT_LAG_MONITORING = "ui.disable.spout.lag.monitoring";
+
+    /**
      * This controls wheather Storm Logviewer should bind to http port even if logviewer.port is > 0.
      */
     @isBoolean
--- a/storm-server/src/main/java/org/apache/storm/DaemonConfig.java
+++ b/storm-server/src/main/java/org/apache/storm/DaemonConfig.java
@@ -317,7 +317,7 @@ public class DaemonConfig implements Validated {
     public static final String UI_DISABLE_HTTP_BINDING = "ui.disable.http.binding";
 
     /**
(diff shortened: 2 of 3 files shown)


storm-server/src/main/java/org/apache/storm/DaemonConfig.java
```diff
@@ -316,6 +316,12 @@
     @isBoolean
     public static final String UI_DISABLE_HTTP_BINDING = "ui.disable.http.binding";
 
+    /**
+     * This controls wheather Storm UI would not monitor Spout lag.
```
agresch: spelling of whether.

kishorvpatil: Fixed.

storm-server/src/main/java/org/apache/storm/DaemonConfig.java
```diff
@@ -316,6 +316,12 @@
     @isBoolean
     public static final String UI_DISABLE_HTTP_BINDING = "ui.disable.http.binding";
 
+    /**
+     * This controls whether Storm UI would not monitor Spout lag.
```
srdo: Nit: If I didn't already know about storm-kafka-monitor, this would be unclear to me. I would also be confused why this doesn't do anything for non-Kafka spouts. How about "This controls whether Storm UI displays spout lag for the Kafka spout"

storm-webapp/src/main/java/org/apache/storm/daemon/ui/UIHelpers.java
```diff
@@ -1618,7 +1619,8 @@ private static Double nullToZero(Double value) {
      * @return getTopologyLag.
      */
     public static Map<String, Map<String, Object>> getTopologyLag(StormTopology userTopology, Map<String,Object> config) {
-        return TopologySpoutLag.lag(userTopology, config);
+        Boolean disableLagMonitoring = ObjectReader.getBoolean(config.get(DaemonConfig.UI_DISABLE_SPOUT_LAG_MONITORING), false);
```
srdo: Nit: Consider adding the default to defaults.yaml instead of here. That way it's obvious to users what the default is.

srdo:
Agree with Jungtaek, getting rid of storm-kafka-monitor would be great.

storm-server/src/main/java/org/apache/storm/DaemonConfig.java
kishorvpatil: Addressed.

kishorvpatil:
@HeartSaVioR I totally agree that we should use spout metrics published to nimbus and make it generalized UI display and REST API response. At that point we can get rid of `storm-kafka-monitor` component.

conf/defaults.yaml
```diff
@@ -98,6 +98,7 @@ ui.header.buffer.bytes: 4096
 ui.http.creds.plugin: org.apache.storm.security.auth.DefaultHttpCredentialsPlugin
 ui.pagination: 20
 ui.disable.http.binding: true
+ui.disable.spout.lag.monitoring: true
```
srdo: Shouldn't this be false?

kishorvpatil: @srdo  I would rather turn off the feature by default. I can make it `false` if you think otherwise.

srdo: No, it's fine by me. I just wanted to be sure it wasn't a mistake.

PR #2988, opened 2019-04-04
[STORM-3233] Update zookeeper version to 3.4.14, master

pacham7:


srdo:
The change looks good. Please include the issue number in the commit message, it makes it easier for us to track why a change was made.

pacham7:
Done... I think.

srdo:
@pacham7 Could you squash back to one commit?

srdo:
So there is only one commit with the message "[STORM-3233] Update zookeeper version to 3.4.14"

pacham7:
@srdo Done, thanks.

srdo:
Test failure looks like a bit of flakiness in the integration test. Might see if I can reproduce it at some point, but it is an issue on current master as well.

Thanks for the PR. +1. We can merge after the 24 hour waiting period.

PR #2993, opened 2019-04-04
STORM-3374 remove unnecessary stacktrace from log

agresch:


srdo:
Do you have some examples of the kind of exception that gets caught here? It seems weird that we only log this line at info level, since we already filter connection errors out.

agresch:
Here's an older one I kept from a ways back: https://jira.apache.org/jira/browse/STORM-3374?focusedCommentId=16810783&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-16810783

srdo:
Thanks. Maybe we could log the stack trace at debug or trace level, and keep the info level log without a stack trace? The reason I'm worried about removing it is that the exception handler catches any kind of exception coming out of the Netty channel handlers, so I don't think we can be sure the exception is innocent.

Alternatively we could log IOExceptions at info level without stack trace (with stack trace at debug/trace level maybe), and any other type of exception at error level?

agresch:
@srdo - updated as you suggested

srdo:
+1, thanks.

srdo:
Please squash and I'll merge

agresch:
@srdo - squashed

PR #2994, opened 2019-04-12
0.9.x branch

tjqiulu:
hi, developer,
I just use storm v0.9.6
here is security issue by Fortify scanning within script.js
as follow are detail:
Abstract:

The method toggleSys() in script-.js sends unvalidated data to a web browser on line 59, which can result in the browser executing malicious code.

Explanation:

Cross-site scripting (XSS) vulnerabilities occur when:

1. Data enters a web application through an untrusted source. In the case of DOM-based XSS, data is read from a URL parameter or other value within the browser and written back into the page with client-side code. In the case of Reflected XSS, the untrusted source is typically a web request, while in the case of Persisted (also known as Stored) XSS it is typically a database or other back-end data store.

2. The data is included in dynamic content that is sent to a web user without being validated. In the case of DOM Based XSS, malicious content gets executed as part of DOM (Document Object Model) creation, whenever the victim's browser parses the HTML page.

The malicious content sent to the web browser often takes the form of a segment of JavaScript, but may also include HTML, Flash or any other type of code that the browser may execute. The variety of attacks based on XSS is almost limitless, but they commonly include transmitting private data like cookies or other session information to the attacker, redirecting the victim to web content controlled by the attacker, or performing other malicious operations on the user's machine under the guise of the vulnerable site.

function toggleSys() {
    var sys = $.cookies.get('sys') || false;
    sys = !sys;

    var exDate = new Date();
    exDate.setDate(exDate.getDate() + 365);

    $.cookies.set('sys', sys, {'path':'/', 'expiresAt':exDate.toUTCString()});
    **window.location = window.location;**
}

could you give some solution?
please

br.
Qiu lu


srdo:
Hi @tjqiulu 

0.9.x is no longer maintained. Please upgrade to a current version.

The code you are pointing to still exists in master https://github.com/apache/storm/blob/1976cb332bc6e8dceaec278fb1c3927627981081/storm-webapp/src/main/java/org/apache/storm/daemon/ui/WEB-INF/js/script.js#L88. I believe it is a false positive from your scanning tool. All that line does is reload the page.

That said, I would be fine with seeing it replaced with Location.reload https://developer.mozilla.org/en-US/docs/Web/API/Location/reload. You can raise an issue at https://issues.apache.org/jira to track this change.

PR #2997, opened 2019-04-14
[STORM-3380] Use Objects.equals to avoid possible NullPointerException

bd2019us:
Hello,
I found that if the Map "data" does not have the key "TYPE", the String "compType" may cause potential risk of NullPointerException since it is immediately used after initialization and there is no null checker. One recommended API is Objects.equals(String,String) which can avoid this exception.

### DIFF
--- a/storm-server/src/main/java/org/apache/storm/stats/StatsUtil.java
+++ b/storm-server/src/main/java/org/apache/storm/stats/StatsUtil.java
@@ -2004,7 +2004,7 @@ public class StatsUtil {
         win2stats.put(FAILED, ClientStatsUtil.getMapByKey(data, WIN_TO_FAILED));
 
         String compType = (String) data.get(TYPE);
-        if (compType.equals(ClientStatsUtil.SPOUT)) {
+        if (Objects.equals(compType,ClientStatsUtil.SPOUT)) {
             ret.set_component_type(ComponentType.SPOUT);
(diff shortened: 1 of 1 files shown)


srdo:
@bd2019us Thanks, looks great. Could you squash to one commit, and we can merge after the 24-hour waiting period?

storm-server/src/main/java/org/apache/storm/stats/StatsUtil.java
```diff
@@ -2004,7 +2005,7 @@ private static ComponentPageInfo thriftifyCompPageData(
         win2stats.put(FAILED, ClientStatsUtil.getMapByKey(data, WIN_TO_FAILED));
 
         String compType = (String) data.get(TYPE);
-        if (compType.equals(ClientStatsUtil.SPOUT)) {
+        if (Objects.equals(compType,ClientStatsUtil.SPOUT)) {
```
HeartSaVioR: It looks like only avoiding NPE, which also makes misleading as component being interpreted as bolt. We should see what are the actual cases of `data.get(TYPE)` being null, and decide how to fix based on the case.

Ethanlm: I agree. Maybe also consider throwing exceptions if it will ever be null. Need to understand it more

PR #3000, opened 2019-04-17
STORM-3381: Exclude spotbugs-annotations from Zookeeper, regen DEPEND…

srdo:
…ENCY-LICENSES

https://issues.apache.org/jira/browse/STORM-3381


hmcl:
+1

DEPENDENCY-LICENSES
```diff
@@ -584,6 +582,7 @@ List of third-party dependencies grouped by their license type.
         * Woodstox (org.codehaus.woodstox:woodstox-core-asl:4.4.1 - http://woodstox.codehaus.org)
         * Xerces2 Java Parser (xerces:xercesImpl:2.9.1 - http://xerces.apache.org/xerces2-j)
         * XML Commons External Components XML APIs (xml-apis:xml-apis:1.3.04 - http://xml.apache.org/commons/components/external/)
+        * zookeeper (org.apache.zookeeper:zookeeper:3.4.14 - no url defined)
```
hmcl: NIT:  missing the URL

srdo: The file is autogenerated. It will just get overwritten. Best fix is to get Zookeeper to add an URL to their pom. As far as I can tell they're in the middle of a migration to Maven, and the new pom has an URL in the license section https://github.com/apache/zookeeper/blob/ff47fc3dfda3487ca3a5aca2ea16831b2aebc7e2/pom.xml#L54

srdo: I think we can override urls for the dependencies that are missing them. I'll check. If that works I'll put it up in a later PR.

srdo: I checked and it doesn't look like there's a way to override the urls currently. The missing file format (the stuff in THIRD-PARTY.properties) only allows you to set a license name, not a dependency URL.

The element I pointed to before was wrong, it's actually this one that needs to be set https://github.com/apache/zookeeper/blob/ff47fc3dfda3487ca3a5aca2ea16831b2aebc7e2/pom.xml#L46

Making these overridable is going to need a fix in the license plugin, I'll raise an issue, but I can't see how it can be fixed without replacing the missing file format.

PR #3001, opened 2019-04-25
[STORM-3384] set-log-level command throws wrong exception if the topology is not running

Ethanlm:
https://issues.apache.org/jira/browse/STORM-3384

Fix
```
Exception in thread "main" java.lang.IllegalArgumentException: No matching field found: IllegalArgumentException for class java.lang.String
```


srdo:
+1

Ethanlm:
The failure seems to be irrelevant .

PR #3003, opened 2019-04-25
[STORM-3384] set-log-level command throws wrong exception if the topology is not running

Ethanlm:
cherry-pick from https://github.com/apache/storm/pull/3001


srdo:
+1

Ethanlm:
The failure seems to be irrelevant .

PR #3008, opened 2019-05-15
STORM-3392 Topology page should show components even if workers haven…

dandsager1:
…'t started

### DIFF
--- a/storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
+++ b/storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
@@ -84,6 +84,8 @@ import org.apache.storm.generated.AlreadyAliveException;
 import org.apache.storm.generated.Assignment;
 import org.apache.storm.generated.AuthorizationException;
 import org.apache.storm.generated.BeginDownloadResult;
+import org.apache.storm.generated.Bolt;
+import org.apache.storm.generated.BoltAggregateStats;
 import org.apache.storm.generated.ClusterSummary;
 import org.apache.storm.generated.CommonAggregateStats;
 import org.apache.storm.generated.ComponentAggregateStats;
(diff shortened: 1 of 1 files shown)


dandsager1:
Topology page sample
Screen Shot 2019-05-16 at 10 26 18 AM

srdo:
Looks good, but could you elaborate on why it is useful to show these placeholders?

dandsager1:
This is useful when there are no metrics, which happens when starting up a topology.

srdo:
I meant why is it useful? I get that it has an effect when the topology starts, but why is it useful to you that the placeholders are shown, rather than the UI being empty?

d2r:
@srdo I think in a multi-tenant environment, where some of us have to try and diagnose another user's topology that is not working well, it is would be nice if the UI at least displays the topology's components regardless of whether they have yet sent any metrics. At least that is a case when I would find it useful.

In fact it would be nice to enhance the Topology Visualization in the same way, but that probably is a more involved change and might best be done separately.

srdo:
@d2r Sounds good, thanks for explaining. +1.

storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
```diff
@@ -4129,6 +4136,83 @@ public TopologyPageInfo getTopologyPageInfo(String topoId, String window, boolea
...
+     * Add placeholder AggStats allowing topology page to show components before AggStats are populated.
+     *
+     * @param topoPageInfo  topology page info holding bolt AggStats
+     * @param topology      storm topology used to get bolt names
+     */
+    private void maybeAddPlaceholderBoltAggStats(TopologyPageInfo topoPageInfo, StormTopology topology) {
+        if (topoPageInfo.get_id_to_bolt_agg_stats().isEmpty()) {
+            Map<String, Bolt> bolts = topology.get_bolts();
+            for (Entry<String, Bolt> entry : bolts.entrySet()) {
+                if (Utils.isSystemId(entry.getKey())) {
```
Ethanlm: Sorry I am too late to review. But why do we want to skip system bolts?

dandsager1: The idea was to show active topology components. It is easy to filter based on the includeSys param passed into getTopologyPageInfo

PR #3009, opened 2019-05-17
Fix class name of CsvStormReporter

contradictioned:
In the documentation about metrics v2, the class name of the csv is `CsvReporter`, when it actually should be `CsvStormReporter`.

Is it enough to merge this into master, or should I also ask to merge this into a branch like `1.1.x`, such that it can be fixed on all relevant online docs?


HeartSaVioR:
> Is it enough to merge this into master, or should I also ask to merge this into a branch like 1.1.x, such that it can be fixed on all relevant online docs?

I guess we can port it back while merging the patch.

srdo:
Thanks @contradictioned. I'll merge it here and to 1.x-branch, and then fix up the 1.2.2/2.0.0 docs on the site.

PR #3010, opened 2019-05-17
STORM-3394 Fix placeholder system bolt filtering for Topology page

dandsager1:
Placeholder system components now filtered by Show/Hide System Stats. Always filters out the **__system** bolt since this is suppressed on the non-placeholder web page.

### DIFF
--- a/storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
+++ b/storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
@@ -4081,7 +4081,7 @@ public class Nimbus implements Iface, Shutdownable, DaemonCommon {
                 setResourcesDefaultIfNotSet(boltResources, entry.getKey(), topoConf);
                 commonStats.set_resources_map(boltResources.get(entry.getKey()).toNormalizedMap());
             }
-            maybeAddPlaceholderBoltAggStats(topoPageInfo, topology);
+            maybeAddPlaceholderBoltAggStats(topoPageInfo, topology, includeSys);
 
             if (workerSummaries != null) {
                 topoPageInfo.set_workers(workerSummaries);
@@ -4145,7 +4145,7 @@ public class Nimbus implements Iface, Shutdownable, DaemonCommon {
     private void maybeAddPlaceholderSpoutAggStats(TopologyPageInfo topoPageInfo, StormTopology topology) {
         if (topoPageInfo.get_id_to_spout_agg_stats().isEmpty()) {
             Map<String, SpoutSpec> spouts = topology.get_spouts();
-            for (Entry<String, SpoutSpec> entry : spouts.entrySet()) {
+            for (String spoutName : spouts.keySet()) {
                 // component
                 ComponentAggregateStats placeholderComponentStats = new ComponentAggregateStats();
                 placeholderComponentStats.set_type(ComponentType.SPOUT);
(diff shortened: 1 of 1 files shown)


storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
```diff
@@ -4176,12 +4176,13 @@ private void maybeAddPlaceholderSpoutAggStats(TopologyPageInfo topoPageInfo, Sto
...
+     * @param includeSys    whether to show system bolts
      */
-    private void maybeAddPlaceholderBoltAggStats(TopologyPageInfo topoPageInfo, StormTopology topology) {
+    private void maybeAddPlaceholderBoltAggStats(TopologyPageInfo topoPageInfo, StormTopology topology, boolean includeSys) {
         if (topoPageInfo.get_id_to_bolt_agg_stats().isEmpty()) {
             Map<String, Bolt> bolts = topology.get_bolts();
-            for (Entry<String, Bolt> entry : bolts.entrySet()) {
-                if (Utils.isSystemId(entry.getKey())) {
+            for (String boltName : bolts.keySet()) {
+                if (!includeSys && Utils.isSystemId(boltName) || boltName.equals("__system")) {
```
HeartSaVioR: `Utils.isSystemId` already covers `boltName.equals("__system")`, so you just need to call `Utils.isSystemId`.

dandsager1: When not using placeholder components, there's some filtering on the **__system** bolt that suppresses it's displaying even when the "Show System Stats" option is selected.

The check for __system is trying to keep the placeholder and non-placeholder output in sync.

I can remove the check if it's decided that's a cleaner way to go.

HeartSaVioR: Is your intention `(!includeSys && Utils.isSystemId(boltName)) || boltName.equals("__system"))`? Then let's add bracket there to clarify the intention.

I'm not sure where you pointed out as `some filtering`, but now I think it may be natural to filter out system bolts, as it runs per executor and the main purpose for now is shipping metrics V1 which tracking these bolts may not give meaningful value.

dandsager1: Added parentheses for clarity

storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
```diff
@@ -4176,12 +4176,13 @@ private void maybeAddPlaceholderSpoutAggStats(TopologyPageInfo topoPageInfo, Sto
...
+     * @param includeSys    whether to show system bolts
      */
-    private void maybeAddPlaceholderBoltAggStats(TopologyPageInfo topoPageInfo, StormTopology topology) {
+    private void maybeAddPlaceholderBoltAggStats(TopologyPageInfo topoPageInfo, StormTopology topology, boolean includeSys) {
         if (topoPageInfo.get_id_to_bolt_agg_stats().isEmpty()) {
             Map<String, Bolt> bolts = topology.get_bolts();
-            for (Entry<String, Bolt> entry : bolts.entrySet()) {
-                if (Utils.isSystemId(entry.getKey())) {
+            for (String boltName : bolts.keySet()) {
+                if ((!includeSys && Utils.isSystemId(boltName)) || boltName.equals("__system")) {
```
Ethanlm: `Constants.SYSTEM_COMPONENT_ID` Is a better instead of `__system`

dandsager1: Changed to use Constants.SYSTEM_COMPONENT_ID

Ethanlm:
We should squash the commits before merge.

PR #3011, opened 2019-05-22
STORM-3395 Fix Placeholder Topology page components show task counts

dandsager1:
Placeholder components num_tasks and num_executors now show expected value

### DIFF
--- a/storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
+++ b/storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
@@ -4144,19 +4144,13 @@ public class Nimbus implements Iface, Shutdownable, DaemonCommon {
      */
     private void maybeAddPlaceholderSpoutAggStats(TopologyPageInfo topoPageInfo, StormTopology topology) {
         if (topoPageInfo.get_id_to_spout_agg_stats().isEmpty()) {
-            Map<String, SpoutSpec> spouts = topology.get_spouts();
-            for (String spoutName : spouts.keySet()) {
+            for (Entry<String, SpoutSpec> entry : topology.get_spouts().entrySet()) {
                 // component
                 ComponentAggregateStats placeholderComponentStats = new ComponentAggregateStats();
                 placeholderComponentStats.set_type(ComponentType.SPOUT);
 
                 // common aggregate
-                CommonAggregateStats commonStats = new CommonAggregateStats();
-                commonStats.set_num_executors(0);
-                commonStats.set_num_tasks(0);
-                commonStats.set_emitted(0);
-                commonStats.set_transferred(0);
-                commonStats.set_acked(0);
+                CommonAggregateStats commonStats = getPlaceholderCommonAggregateStats(entry.getValue());
                 placeholderComponentStats.set_common_stats(commonStats);
 
                 // spout aggregate
@@ -4166,7 +4160,7 @@ public class Nimbus implements Iface, Shutdownable, DaemonCommon {
                 specificStats.set_spout(spoutAggStats);
                 placeholderComponentStats.set_specific_stats(specificStats);
 
-                topoPageInfo.get_id_to_spout_agg_stats().put(spoutName, placeholderComponentStats);
+                topoPageInfo.get_id_to_spout_agg_stats().put(entry.getKey(), placeholderComponentStats);
             }
         }
     }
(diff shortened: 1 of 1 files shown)


storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
```diff
@@ -4214,6 +4203,22 @@ private void maybeAddPlaceholderBoltAggStats(TopologyPageInfo topoPageInfo, Stor
...
     }
 
+    private CommonAggregateStats getPlaceholderCommonAggregateStats(Object component) {
+        // common aggregate
+        CommonAggregateStats commonStats = new CommonAggregateStats();
+
+        // set num_tasks and expected num_executors
+        Map<String, Object> jsonMap = StormCommon.componentConf(component);
+        int numTasks = (int) (long) jsonMap.getOrDefault(Config.TOPOLOGY_TASKS, 0);
+        commonStats.set_num_executors(numTasks);
```
HeartSaVioR: Just to double check, is it safe to assume number of executors is same as number of tasks? If there's safe way to get the value it would be better to apply that, otherwise 0 (even -1) might make less confusion.

dandsager1: Num_executors does not have to be same as num_tasks. Changing num_executors to 0

Ethanlm: You can use `StormCommon.numStartExecutors` to get num_executors.

https://github.com/apache/storm/blob/master/storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java#L2392-L2405

dandsager1: Thanks @Ethanlm. I'll pull in numExecutors from your suggestion.

storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
```diff
@@ -4214,6 +4203,31 @@ private void maybeAddPlaceholderBoltAggStats(TopologyPageInfo topoPageInfo, Stor
...
     }
 
+    private CommonAggregateStats getPlaceholderCommonAggregateStats(Object component) {
+        // common aggregate
+        CommonAggregateStats commonStats = new CommonAggregateStats();
+
+        // get num_executors
+        int numExecutors = 0;
+        try {
+            numExecutors = "(component);
```
HeartSaVioR: This code becomes odd. Please make sure you've run tests locally before push.

dandsager1: Sorry, not sure how this happened. I ran tests locally before committing and then must have fat fingered the file before committing.

dandsager1: Fix has been committed.

storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
```diff
@@ -4214,6 +4203,31 @@ private void maybeAddPlaceholderBoltAggStats(TopologyPageInfo topoPageInfo, Stor
...
 
+    private CommonAggregateStats getPlaceholderCommonAggregateStats(Object component) {
+        // common aggregate
+        CommonAggregateStats commonStats = new CommonAggregateStats();
+
+        // get num_executors
+        int numExecutors = 0;
+        try {
+            numExecutors = StormCommon.numStartExecutors(component);
+        } catch (InvalidTopologyException e) {
```
Ethanlm: It looks like we can remove `throws InvalidTopologyException` from `numStartExecutors` and then we don't need to catch the `InvalidTopologyException`

storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
```diff
@@ -4214,6 +4203,31 @@ private void maybeAddPlaceholderBoltAggStats(TopologyPageInfo topoPageInfo, Stor
...
+            // ignore
+        }
+
+        // get num_tasks
+        Map<String, Object> jsonMap = StormCommon.componentConf(component);
+        int numTasks = (int) (long) jsonMap.getOrDefault(Config.TOPOLOGY_TASKS, 0);
```
Ethanlm: Why do we need to cast twice: `(int) (long) `? Also it's probably better to have 

`jsonMap.getOrDefault(Config.TOPOLOGY_TASKS, numExecutors);`. But in theory, `TOPOLOGY_TASKS` will be set during submission if it's not specified by the topology. So it should be fine.

storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
dandsager1: Leaving as is

storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
dandsager1: Removed the ugly cast and now using numExecutors as default value

Ethanlm:
Could you please squash the commits? I can then merge it

Ethanlm:
All the build passed except
``` 
9968.10
JDK: openjdk11 Java
MODULES=Integration-Test
 32 min 1 sec
```
Should be unrelated. Merging it

PR #3013, opened 2019-05-25
STORM-3396:fix uploading dependency jars too slow when storm client and server located in different IDC

StaticMian:
when storm client and server is locating in different IDC(one is in Beijing, while another in Shanghai), uploading dependency jars may take a very long long time(in my case, 31minutes!)...

    when I digged into this, I found that in DependencyUploader,  method "uploadDependencyToBlobStore" using JDK NIO's Files.copy to upload local jars to remote Blob server. In Files.copy(Path, OutputStream), the buffer size is 8k by default, given that latency between Beijing and Shanghai is about 20ms, a dependency fat jar of 360M finally cost me 'a lunch time' to finish uploading!!!

### DIFF
--- a/storm-client/src/jvm/org/apache/storm/dependency/DependencyUploader.java
+++ b/storm-client/src/jvm/org/apache/storm/dependency/DependencyUploader.java
@@ -20,11 +20,13 @@ package org.apache.storm.dependency;
 
 import java.io.File;
 import java.io.IOException;
+import java.io.InputStream;
 import java.nio.file.Files;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
+import org.apache.storm.Config;
 import org.apache.storm.blobstore.AtomicOutputStream;
 import org.apache.storm.blobstore.BlobStoreAclHandler;
 import org.apache.storm.blobstore.ClientBlobStore;
@@ -35,18 +37,22 @@ import org.apache.storm.generated.KeyAlreadyExistsException;
 import org.apache.storm.generated.KeyNotFoundException;
 import org.apache.storm.generated.SettableBlobMeta;
 import org.apache.storm.shade.com.google.common.annotations.VisibleForTesting;
+import org.apache.storm.utils.ObjectReader;
 import org.apache.storm.utils.Utils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.apache.storm.shade.org.apache.commons.io.IOUtils;
 
 public class DependencyUploader {
     public static final Logger LOG = LoggerFactory.getLogger(DependencyUploader.class);
 
     private final Map<String, Object> conf;
     private ClientBlobStore blobStore;
+    private int uploadChuckSize;
 
     public DependencyUploader() {
         conf = Utils.readStormConfig();
+        this.uploadChuckSize = ObjectReader.getInt(conf.get(Config.STORM_BLOBSTORE_DEPENDENCY_JAR_UPLOAD_CHUCK_SIZE_BYTES), 1024 * 1024);
     }
 
     public void init() {
@@ -157,7 +163,9 @@ public class DependencyUploader {
             AtomicOutputStream blob = null;
             try {
                 blob = getBlobStore().createBlob(key, new SettableBlobMeta(acls));
-                Files.copy(dependency.toPath(), blob);
+                try(InputStream in = Files.newInputStream(dependency.toPath())) {
+                    IOUtils.copy(in, blob, this.uploadChuckSize);
+                }
                 blob.close();
                 blob = null;
 
--- a/conf/defaults.yaml
+++ b/conf/defaults.yaml
@@ -144,6 +144,7 @@ nimbus.blobstore.class: "org.apache.storm.blobstore.LocalFsBlobStore"
 nimbus.blobstore.expiration.secs: 600
 
 storm.blobstore.inputstream.buffer.size.bytes: 65536
+storm.blobstore.dependency.jar.upload.chuck.size.bytes: 1048576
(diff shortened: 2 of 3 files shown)


storm-client/src/jvm/org/apache/storm/dependency/DependencyUploader.java
```diff
@@ -44,9 +51,12 @@
 
     private final Map<String, Object> conf;
     private ClientBlobStore blobStore;
+    private static final int DEFAULT_CHUCK_SIZE = 1024 * 1024;
```
srdo: I think this should go in https://github.com/apache/storm/blob/master/conf/defaults.yaml so it's obvious to users what the default value is.

storm-client/src/jvm/org/apache/storm/dependency/DependencyUploader.java
```diff
@@ -157,7 +167,14 @@ private boolean uploadDependencyToBlobStore(String key, File dependency)
             AtomicOutputStream blob = null;
             try {
                 blob = getBlobStore().createBlob(key, new SettableBlobMeta(acls));
-                Files.copy(dependency.toPath(), blob);
+//              Files.copy(dependency.toPath(), blob);
```
srdo: Please delete this.

storm-client/src/jvm/org/apache/storm/dependency/DependencyUploader.java
```diff
@@ -176,6 +193,22 @@ private boolean uploadDependencyToBlobStore(String key, File dependency)
         return uploadNew;
     }
 
+    private long doUploadDependencyJar(Path source, OutputStream destination) throws IOException{
```
srdo: I think the idea here is great, but we should try to use existing tools instead. How about https://commons.apache.org/proper/commons-io/javadocs/api-2.5/org/apache/commons/io/IOUtils.html#copyLarge(java.io.InputStream,%20java.io.OutputStream,%20byte[])?

storm-client/src/jvm/org/apache/storm/dependency/DependencyUploader.java
```diff
@@ -176,6 +193,22 @@ private boolean uploadDependencyToBlobStore(String key, File dependency)
...
+        Objects.requireNonNull(destination);
+        try (InputStream in = source.getFileSystem().provider().newInputStream(source)) {
```
srdo: How about https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#newInputStream-java.nio.file.Path-java.nio.file.OpenOption...- instead?

storm-client/src/jvm/org/apache/storm/dependency/DependencyUploader.java
```diff
@@ -157,7 +167,14 @@ private boolean uploadDependencyToBlobStore(String key, File dependency)
             AtomicOutputStream blob = null;
             try {
                 blob = getBlobStore().createBlob(key, new SettableBlobMeta(acls));
-                Files.copy(dependency.toPath(), blob);
+//              Files.copy(dependency.toPath(), blob);
+
+                /***
+                 * By default, {@link java.nio.file.Files#copy(Path, OutputStream)} copy bytes only 8k each time,
```
srdo: I think this comment should be deleted before we merge, as it only makes sense when looking at the current diff.

storm-client/src/jvm/org/apache/storm/dependency/DependencyUploader.java
StaticMian: thanks for the heads up !

storm-client/src/jvm/org/apache/storm/dependency/DependencyUploader.java
StaticMian: ok...

storm-client/src/jvm/org/apache/storm/dependency/DependencyUploader.java
StaticMian: clever! thanks for the advice...

storm-client/pom.xml
```diff
@@ -83,6 +83,11 @@
             <artifactId>kryo</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>commons-io</groupId>
```
srdo: This is not necessary. IOUtils is already available somewhere under the `org.apache.storm.shade` namespace. Try removing this again, and letting your IDE add the import in DependencyUploader

StaticMian: Thank you for your patience！

storm-client/src/jvm/org/apache/storm/dependency/DependencyUploader.java
```diff
@@ -154,10 +160,15 @@ private boolean uploadDependencyToBlobStore(String key, File dependency)
             acls.add(new AccessControl(AccessControlType.OTHER,
                                        BlobStoreAclHandler.READ));
 
+            InputStream in = null;
             AtomicOutputStream blob = null;
             try {
+                in = Files.newInputStream(dependency.toPath());
```
srdo: Please use try-with-resources as you did before, it is better and safer.

StaticMian: Good idea! But, does two 'try's look pretty? also I want to keep the code style same with blob...
ok, I'll try the try-with-resources

StaticMian: Thanks again for your advice. This is my first time to commit a PR,  I am learning how to do it formally. To save your time, when that being done, I'll notice you.

srdo:
+1

PR #3020, opened 2019-06-01
[STORM-3000] Add missing @Override annotations

krichter722:
Adding missing `@Override` annotations helps detecting inheritance issues when generics are used more intensively.

In order to avoid missing `@Override` annotations in the future PMD is configured and used to verify the code state.

Is based on changes proposed in #3019.

### DIFF
--- a/examples/storm-jms-examples/src/main/java/org/apache/storm/jms/example/SpringJmsProvider.java
+++ b/examples/storm-jms-examples/src/main/java/org/apache/storm/jms/example/SpringJmsProvider.java
@@ -63,10 +63,12 @@ public class SpringJmsProvider implements JmsProvider {
 		this.destination = (Destination)context.getBean(destinationBean);
 	}
 
+	@Override
 	public ConnectionFactory connectionFactory() throws Exception {
 		return this.connectionFactory;
 	}
 
+	@Override
 	public Destination destination() throws Exception {
 		return this.destination;
 	}
--- a/examples/storm-hbase-examples/src/main/java/org/apache/storm/hbase/topology/TotalWordCounter.java
+++ b/examples/storm-hbase-examples/src/main/java/org/apache/storm/hbase/topology/TotalWordCounter.java
@@ -32,6 +32,7 @@ public class TotalWordCounter implements IBasicBolt {
     private static final Random RANDOM = new Random();
     private BigInteger total = BigInteger.ZERO;
 
+    @Override
     public void prepare(Map<String, Object> topoConf, TopologyContext context) {
     }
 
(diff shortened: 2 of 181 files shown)


examples/storm-jms-examples/src/main/java/org/apache/storm/jms/example/SpringJmsProvider.java
```diff
@@ -63,10 +63,12 @@ public SpringJmsProvider(String appContextClasspathResource, String connectionFa
 		this.destination = (Destination)context.getBean(destinationBean);
 	}
 
+        @Override
```
d2r: Minor: indentation this file

krichter722:
@d2r @srdo Thanks for you review, I included your requests.

srdo:
+1

krichter722:
> For this PR it is minor, and fixing the other white-space is out of scope anyway. I'll follow up on this white-space separately.

I already did that for some projects. I'll open them soon.

d2r:
> I already did that for some projects. I'll open them soon.

That would be great!
I wanted to do three things:
* Normalize white-space on master branch
* Figure out why `mvn checkstyle:checkstyle` does not work, because it [should have caught this](https://github.com/apache/storm/blob/b9ab5e56a9ad1776022ed6d8520c0268a666fea0/storm-checkstyle/src/main/resources/storm/storm_checkstyle.xml#L56-L61)
* Perhaps add an [.editorconfig](https://editorconfig.org/) to our repo, to help make it easier in avoiding these types of errors.

srdo:
@d2r The module has a number of allowed violations, it may be that the violations weren't caught because it didn't increase the number of violations, or maybe the allowed count is too high https://github.com/apache/storm/blob/master/examples/storm-jms-examples/pom.xml#L108

d2r:
> it may be that the violations weren't caught because it didn't increase the number of violations, or maybe the allowed count is too high

I see you have put some thought into this already: This could be addressed by your jira issue [STORM-3071](https://issues.apache.org/jira/browse/STORM-3071).

krichter722:
#3025 covers some of the issues. Since fixing one issue for all project requires the readjustment of maxAllowedViolations which is painful. I suggest pushing getting all projects to have maxAllowedViolations == 0. It's the least troublesome approach in my opinion.

This can be merged, right?

srdo:
@krichter722 Yes, thanks for your patience.

Regarding adjusting maxAllowedViolations, I don't see why it is painful?

krichter722:
> Regarding adjusting maxAllowedViolations, I don't see why it is painful?

You need to count how many violations you fixed and change the number and do that for every project, otherwise new violations can be committed. I think this requires a lot of effort on top of fixing all violations which needs to be done anyway.

srdo:
Ah right, sorry I misunderstood. You are right, that is painful.

PR #3021, opened 2019-06-06
[STORM-3406] Allow customizing storm-hdfs FileReader

kishorvpatil:


srdo:
It looks like we're constructing instances of FileReader by reflection. We're calling two different constructors:

https://github.com/apache/storm/blob/22440747aa0f2ed3f3d26baec9140f79bb82f644/external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/HdfsSpout.java#L699

https://github.com/apache/storm/blob/22440747aa0f2ed3f3d26baec9140f79bb82f644/external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/HdfsSpout.java#L699

I think if we're going to make the interface public, we should also have a factory interface that requires factories to have those two methods. We can then replace the code I linked with calls to the factory methods. That way users won't have to read the code to know that they need to implement constructors that match the calls linked above.

PR #3023, opened 2019-06-10
[STORM-3401] Add OpenLab CI configuration for ARM64 build

Yikun:
This patch adds the CI configuration to enable the support for arm
build in OpenLab.

After this, each pull request in theopenlab/storm will trigger the
storm-arm64-build job which verified the arm build on ARM cluster.


rzo1:
@Yikun Is this still a thing?

PR #3024, opened 2019-06-13
STORM-3408 - update rocks version

thecoop:
https://issues.apache.org/jira/browse/STORM-3408

The current version of rocks doesn't work on some installs of windows 10. Update to the most recent 5.x version


srdo:
The changes look good, and going by https://github.com/facebook/rocksdb/wiki/RocksDB-Compatibility-Between-Different-Releases we should be fine to upgrade to new versions without requiring users to do any manual migration steps.

+1

PR #3025, opened 2019-06-15
[STORM-3415] Checkstyle jms examples

krichter722:


examples/storm-jms-examples/src/main/java/org/apache/storm/jms/example/ExampleJmsTopology.java
```diff
@@ -41,7 +45,12 @@
     public static final String JMS_TOPIC_SPOUT = "JMS_TOPIC_SPOUT";
     public static final String ANOTHER_BOLT = "ANOTHER_BOLT";
 
-    @SuppressWarnings("serial")
+    /**
+     * The main method.
+     * @param args takes the topology name as first argument
+     * @throws Exception any expection occuring durch cluster setup or operation
+     */
+    @SuppressWarnings({"serial", "checkstyle:VariableDeclarationUsageDistance"})
```
srdo: This seems like a bad way to get violation count to 0. We're just hiding the violation now.

examples/storm-jms-examples/src/main/java/org/apache/storm/jms/example/GenericBolt.java
```diff
@@ -15,27 +15,28 @@
...
 /**
  * A generic <code>org.apache.storm.topology.IRichBolt</code> implementation
  * for testing/debugging the Storm JMS Spout and example topologies.
- * <p>
- * For debugging purposes, set the log level of the
+ *
+ * <p>For debugging purposes, set the log level of the
  * <code>org.apache.storm.contrib.jms</code> package to DEBUG for debugging
- * output.
+ * output.</p>
```
srdo: nit: I don't think p tags need to be closed.

krichter722: They're not necessary in HTML, but possible. I don't like the idea and it's been changed in XHTML. Do you insist?

examples/storm-jms-examples/src/main/java/org/apache/storm/jms/example/ExampleJmsTopology.java
krichter722: In my opinion it's valid to make exceptions if they're thought through. Especially is the exception is associated with concrete line of code rather than just represented by a count.

Breaking up this method would reduce it's readability in my opinion. After all, I don't care, but I'm starting the discussion in order to figure out whether you're strictly against it (I might do similar work) or whether you accept it if you agree.

examples/storm-jms-examples/src/main/java/org/apache/storm/jms/example/GenericBolt.java
srdo: No, it's fine.

krichter722:
Can anyone tell me, why master builds and this branch doesn't because of missing files in `storm-checkstyle` if nothing has been changed in the only commit of this PR?

examples/storm-jms-examples/src/main/java/org/apache/storm/jms/example/ExampleJmsTopology.java
srdo: I agree in principle, but in this case the violation is just saying to move the `jmsTopicProvider` declaration closer to where it is first used. I don't think you need to break up the method.

srdo:
I think it's an issue inherited from https://github.com/apache/storm/pull/3020/files, looks like it can't find the storm/pmd-ruleset file added to config here https://github.com/apache/storm/pull/3020/files#diff-600376dffeb79835ede4a0b285078036R1257

Master may have passed because the Maven repo was cached, and contained storm-checkstyle from an earlier build.

srdo:
Okay, figured out the issue. The travis script builds storm-shaded-deps first, separately from the rest of the build. This means storm-checkstyle isn't build first, which means it's using whatever version is cached.

I'll fix it.

srdo:
@krichter722 The build should be back to green.

krichter722:
@srdo Thanks for the pointer. This can be merged, right?

srdo:
@krichter722 Thanks for your patience. I'd still like to move `jmsTopicProvider` down so we don't have to ignore the error in https://github.com/apache/storm/pull/3025#discussion_r294055492

srdo:
+1, thanks @krichter722

srdo:
Oh whoops, wait we need a JIRA issue for this. Could you create one, and rename the PR and commit message to contain the issue number?

krichter722:
@srdo Sure, done. I didn't want to sneak in the warnings suppression :) (and I don't think that you think that I did)

PR #3029, opened 2019-06-18
copyright year set to 2019

jnioche:
Have I just won the prize for the most trivial contribution ever?


srdo:
Thanks for noticing and fixing.

PR #3031, opened 2019-06-19
STORM-3409: Build shaded deps with dependencies, instead of as a loos…

srdo:
…e module, never upload Storm artifacts to Travis build cache to ensure a clean slate for each build. Remove license plugin configuration until we can upgrade to 1.21, as it breaks clean builds.

https://issues.apache.org/jira/browse/STORM-3409

This issue seems to be the cause of the current Travis failures.

I'll add the license plugin back as soon as 1.21 releases. The current version is breaking the build, as it causes Maven to try to resolve e.g. storm-client before it is built.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,10 @@ env:
 dist: trusty
 sudo: required
 
+before_cache:
+# Never upload Storm artifacts to the build cache, ensuring a clean slate every time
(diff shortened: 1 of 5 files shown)


srdo:
@Ethanlm Thanks for reviewing

PR #3034, opened 2019-06-25
STORM-3422: TupleCaptureBolt is not thread-safe

JanecekPetr:
Fixes [STORM-3422](https://issues.apache.org/jira/browse/STORM-3422).

Note, **this is a preview PR meant as a base for more discussion**.

There are two commits, two different approaches to the same problem, please take a look. The latter, I believe, is simpler and easier to understand.

The synchronization is expected to be completely uncontended, we simply need a way to safely publish the changes made by one thread or the other. Either way, the usage of the class feels dirty, but there isn't much we can do now, is there?

### DIFF
--- a/storm-client/src/jvm/org/apache/storm/testing/TupleCaptureBolt.java
+++ b/storm-client/src/jvm/org/apache/storm/testing/TupleCaptureBolt.java
@@ -17,22 +17,29 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
 import org.apache.storm.task.OutputCollector;
 import org.apache.storm.task.TopologyContext;
 import org.apache.storm.topology.IRichBolt;
 import org.apache.storm.topology.OutputFieldsDeclarer;
 import org.apache.storm.tuple.Tuple;
 
-
 public class TupleCaptureBolt implements IRichBolt {
-    public static final transient Map<String, Map<String, List<FixedTuple>>> emitted_tuples = new HashMap<>();
 
-    private String name;
+    /*
+     * Even though normally bolts do not need to care about thread safety, this particular bolt is different.
+     * It maintains a static field that is prepopulated before the topology starts, is written into by the topology,
+     * and is then read from after the topology is completed - all of this by potentially different threads.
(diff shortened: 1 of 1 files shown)


srdo:
I think I like the ListMultiMap approach a little better, since it's not as easy to forget to synchronize a reference to the map. I would also expect there to be some contention on the map field, as it will be updated by every tuple in the local topology.

You're right that using a static field for this is a bit hacky. Best alternative I can think of would be creating a field in LocalCluster, and figuring out a way to share it with these bolts. It would probably be more complicated though. Either way, I don't think this is too bad, as the TupleCaptureBolt isn't supposed to be used outside a LocalCluster, and we can make sure to e.g. clear the field between topology submissions.

If you can think of a good alternative, I don't think anyone would miss this bolt :)

JanecekPetr:
> I would also expect there to be some contention on the map field, as it will be updated by every tuple in the local topology.

D'oh! Of course. That was bad.

I updated the PR with another take - this time taking advantage of the *happens-before* guarantees provided by [`ConcurrentHashMap#compute()`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html#compute-K-java.util.function.BiFunction-). This now allows parallelism among different bolts, and will correctly work even if two instances of the same bolt are executed at the same time. The change is simpler, and there's no need to synchronize on the list anymore. The cost is that the inner lambda is a capturing one and has therefore be instantiated with every iteration...

Do you guy have any benchmarks impacted by this? I could not find one. Or do we not really care in tests?

(and I rebased the branch)

srdo:
Looks great, +1 assuming tests pass. I don't think performance in tests is terribly important.

JanecekPetr:
Thanks! How do I manage backporting? Do you guys take care of those things, or should I cherrypick manually and open another PR once this one gets merged?

srdo:
Normally we could cherrypick it for you, but this uses Java 8 features and 1.x targets Java 7. If you still want this backported, we need a new PR.

JanecekPetr:
> 1.x targets Java 7

Aha. Nevermind then, it's a possible, but annoying fix and apparently not many people care as they'd discover this too.

Thanks for your help and rapid responses!

PR #3035, opened 2019-06-25
STORM-3417: Fix checkstyle violations in sql-core

krichter722:


sql/storm-sql-core/src/jvm/org/apache/storm/sql/javac/CompilingClassLoader.java
```diff
@@ -125,7 +125,8 @@ public CompilingClassLoader(
     }
 
     /**
-     * @return Whether compilation was successful.
+     * Compiles source code to byte code.
```
srdo: Wasn't this check disabled?

krichter722: We removed the rule to add missing Javadoc on public methods and classes. The rules to validate already present Javadoc were present and remained untouched. I think this makes sense because if Javadoc is present it should be valid. It could also be removed but deciding that slows down the PR workflow in my opinion.

srdo: Makes sense, thanks.

krichter722:
@srdo This can be merged, right?

srdo:
@krichter722 Yes, soon. We generally wait at least 24 hours before merging, so others get a chance to review.

PR #3037, opened 2019-06-25
STORM-3420: Fix checkstyle violations in storm-mqtt

krichter722:


external/storm-mqtt/src/main/java/org/apache/storm/mqtt/common/MqttOptions.java
```diff
@@ -88,7 +87,7 @@ public String getWillTopic() {
      * If set the server will publish the client's Will message to the specified topics if the client has an unexpected
      * disconnection.
      *
-     * @param willTopic
+     * @param willTopic the will topic
```
srdo: This is minor, but does removing the `@param` cause the checkstyle check to fail? If not, I'd rather get rid of these until someone has something useful to write here. "the will topic" is just noise.

krichter722: Good catch. I though that parameters would be enforced as soon as Javadoc is present. I agree 110% that those are pointless.

srdo:
+1 once test passes

PR #3044, opened 2019-06-27
STORM-3426: Remove AbbreviationAsWordInName from checkstyle setup

krichter722:
This rule requires a lot of unconstructive renamings and doesn't have a huge impact on code quality.


srdo:
Is it a large effort to suppress the current violations instead?

krichter722:
No, I'd say a few dozens.

srdo:
Ok, thanks. Then I think it would be better to keep the rule, and suppress current violations. What do you think?

PR #3052, opened 2019-06-29
STORM-3436: Set TupleInfo root id, so we can log it when topology deb…

srdo:
…ug logging is enabled

### DIFF
--- a/storm-client/src/jvm/org/apache/storm/executor/spout/SpoutOutputCollectorImpl.java
+++ b/storm-client/src/jvm/org/apache/storm/executor/spout/SpoutOutputCollectorImpl.java
@@ -149,6 +149,7 @@ public class SpoutOutputCollectorImpl implements ISpoutOutputCollector {
             info.setTaskId(this.taskId);
             info.setStream(stream);
             info.setMessageId(messageId);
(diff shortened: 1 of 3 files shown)


srdo:
This is a regression since 1.x, as that version correctly logs the root id.

storm-client/src/jvm/org/apache/storm/executor/spout/SpoutOutputCollectorImpl.java
```diff
@@ -172,7 +173,7 @@ public void reportError(Throwable error) {
             globalTupleInfo.setValues(values);
             globalTupleInfo.setMessageId(messageId);
             globalTupleInfo.setTimestamp(0);
-            globalTupleInfo.setId("0:");
+            globalTupleInfo.setRootId(0);
```
Ethanlm: This could be `globalTupleInfo.setRootId(rootId);` since rootId is 0 in this case. But I am fine with it

srdo: Thanks, fixed.

PR #3061, opened 2019-07-01
STORM-3446: storm-maven-plugins: fix all checkstyle warnings

krichter722:


storm-buildtools/storm-maven-plugins/src/main/java/org/apache/storm/maven/plugin/versioninfo/VersionInfoMojo.java
```diff
@@ -113,6 +113,7 @@ private String getBuildTime() {
      * @return SCM in use for this build
      * @throws Exception if any error occurs attempting to determine SCM
      */
+    @SuppressWarnings("checkstyle:AbbreviationAsWordInName")
```
srdo: I think for stuff in this module, feel free to just change the names. This isn't part of a user-facing API.

krichter722: I'm starting to not like this rule :)

srdo: It's fine for new code I think. I agree that it's annoying for our existing code.

PR #3062, opened 2019-07-01
STORM-3447: jms: fix all checkstyle warnings

krichter722:


external/storm-jms/src/main/java/org/apache/storm/jms/spout/JmsMessageID.java
```diff
@@ -14,20 +14,21 @@
...
+    public JmsMessageID(long sequence, String jmsId) {
+        this.jmsId = jmsId;
         this.sequence = sequence;
     }
 
-
-    public String getJmsID() {
-        return this.jmsID;
+    @SuppressWarnings("checkstyle:AbbreviationAsWordInName")
+    public String getJmsId() {
```
srdo: I don't think we should rename this, unless users can't directly access this class?

external/storm-jms/src/main/java/org/apache/storm/jms/trident/TridentJmsSpout.java
```diff
@@ -193,7 +185,7 @@ public Fields getOutputFields() {
         private final long rotateTimeMillis;
         private final int maxBatchSize;
         private final String name;
-        private final Logger LOG = LoggerFactory.getLogger(JmsEmitter.class);
+        private final Logger log = LoggerFactory.getLogger(JmsEmitter.class);
```
srdo: Feel free to make the field static instead.

krichter722: Then the surrounding private class JmsEmitter would need to be static as well. That'd mean touching more than just style. The private class is longer than the surrounding class though, so it could be moved up.

srdo: Why does the surrounding class need to be static too?

srdo: Okay, it doesn't compile. Makes sense. I think leaving it like this is fine.

srdo:
+1. will wait for tests

PR #3067, opened 2019-07-01
STORM-3442 Add owner to supervisor summary

dandsager1:


storm-webapp/src/main/java/org/apache/storm/daemon/ui/WEB-INF/js/script.js
```diff
@@ -332,6 +332,14 @@ var makeWorkerStatsTable = function (response, elId, parentId, type) {
                     data;
             }
         },
+        {
+            data: 'owner',
```
Ethanlm: Will this also add "owner" column to the topology summary page? If so, it will be redundant since the topology summary page already has "owner" information

storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
```diff
@@ -4051,7 +4052,9 @@ public TopologyPageInfo getTopologyPageInfo(String topoId, String window, boolea
                                                            nodeToHost,
                                                            workerToResources,
                                                            includeSys,
-                                                           true); //this is the topology page, so we know the user is authorized
+                                                           true, //this is the topology page, so we know the user is authorized
+                                                           null,
+                                                           owner);
```
Ethanlm: With this change, the other `aggWorkerStats` function is no longer used by anyone and can be removed.

storm-webapp/src/main/java/org/apache/storm/daemon/ui/WEB-INF/js/script.js
dandsager1: Correct. This also adds owner to the topology summary page. The duplicate owner can be removed at the cost of adding logic to script.js. It seems a wash to me.

storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
dandsager1: removing unused overload

storm-webapp/src/main/java/org/apache/storm/daemon/ui/WEB-INF/js/script.js
dandsager1: Removing duplicate owner is simple. Making this change and also removing "owner" tooltip from  topology-page-template

PR #3074, opened 2019-07-04
STORM-3456: cassandra: fix all checkstyle warnings

krichter722:


external/storm-cassandra/src/main/java/org/apache/storm/cassandra/trident/state/OpaqueTupleStateMapper.java
```diff
@@ -99,6 +98,7 @@ public Values toValues(OpaqueValue<ITuple> tuple) {
             prev = null;
         }
 
+        Long currTx = (Long) values.get(index++);
```
srdo: I think this introduces a bug. The index is supposed to be 0 here, but the for loops above can increment it. Maybe initialize index to 1 above, and then just do get(0) here.

external/storm-cassandra/src/main/java/org/apache/storm/cassandra/trident/state/TransactionalTupleStateMapper.java
```diff
@@ -87,6 +86,7 @@ public Values toValues(TransactionalValue<ITuple> tuple) {
             curr = null;
         }
 
+        Long txId = (Long) values.get(index++);
```
srdo: Same issue here

srdo:
+1.

Please fix comments in a new commit on future PRs, Github's force push diffs aren't very good yet (e.g https://github.com/apache/storm/compare/f4698976dc16f6c6303f8877ef82418ec42c8c57..e7f2c919a1a99621450a945e7ad473c0d8641a9a contains a bunch of changes outside this module), so it's hard to tell exactly what changed. It's faster to review only the new changes rather than looking at the full diff again. Thanks.

PR #3075, opened 2019-07-04
STORM-3457: hdfs: fix all checkstyle warnings

krichter722:


external/storm-hdfs/src/main/java/org/apache/storm/hdfs/spout/Configs.java
```diff
@@ -33,79 +34,93 @@
     public static final String TEXT = "text";
     public static final String SEQ = "seq";
     /**
+     * HDFS URI.
```
srdo: Consider moving the comments on the fields into the Javadoc. E.g. here you could put "Required - HDFS name node".

srdo:
+1

PR #3085, opened 2019-07-07
MINOR: Remove unused dev-tools/test-ns.py

krichter722:
… order to avoid unnecessary clutter in the logs


srdo:
+1 for the change, but I'm not sure what this file is for. It looks like it used to be for running tests in storm-core when it was still mainly Clojure code. I think this file is here because running tests in a specific namespace was a bit involved. Since we've switched to Java, this file seems a little redundant.

There doesn't appear to be any references to test-ns from build scripts, and I don't think that anyone is still using it(?). Maybe we should just delete it.

krichter722:
I have to admit that it's an old branch and that I just rebased it and opened the PR. I didn't check whether it was still used. I could reuse this PR to delete the file. What do you think?

srdo:
Deleting the file would be fine by me.

srdo:
+1

srdo:
Thanks @krichter722

PR #3091, opened 2019-07-21
STORM-3472: Add tests missing for STORM-3411, make the download file name generat…

srdo:
…ion code less error prone

### DIFF
--- a/storm-webapp/src/main/java/org/apache/storm/daemon/logviewer/utils/LogviewerResponseBuilder.java
+++ b/storm-webapp/src/main/java/org/apache/storm/daemon/logviewer/utils/LogviewerResponseBuilder.java
@@ -75,11 +75,10 @@ public class LogviewerResponseBuilder {
     /**
      * Build a Response object representing download a file.
      *
-     * @param host host address
      * @param contentDispositionName The name to set in the Content-Disposition header
      * @param file file to download
      */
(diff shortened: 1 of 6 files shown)


Ethanlm:
Sorry for the late reply.   I will take a look when I get a chance.

storm-webapp/src/main/java/org/apache/storm/daemon/logviewer/utils/LogviewerResponseBuilder.java
```diff
@@ -76,24 +76,18 @@ public static Response buildSuccessJsonResponse(Object entity, String callback,
...
-    public static Response buildDownloadFile(String host, File file, Meter numFileDownloadExceptions) throws IOException {
-        String fname;
-        try {
-            String topoInfo = file.getParentFile().getParentFile().getName();
-            String port = file.getParentFile().getName();
-            fname = String.format("%s-%s-%s-%s", host, port, topoInfo, file.getName());
-        } catch (NullPointerException e) {
-            fname = file.getName();
-        }
+    public static Response buildDownloadFile(String host, String contentDispositionName,
```
Ethanlm: The `String host` parameter can be removed.

srdo: Good catch, fixed

PR #3092, opened 2019-07-25
STORM-3474 Large fragmented cluster scheduling time test

dandsager1:


dandsager1:
Added reason for test in JIRA

storm-server/src/main/java/org/apache/storm/scheduler/resource/strategies/scheduling/ConstraintSolverStrategy.java
```diff
@@ -41,6 +41,9 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+// When making changes to scheduling strategy, be sure to run performance performance unit test
+// TestResourceAwareScheduler::testLargeFragmentedClusterScheduling
```
Ethanlm: I don't think this is a good practice to link to a unit test. And we don't need to worry about reminding developers to run a specific unit test because running unit tests is something that will just happen. Developers will run all the related unit tests or at least let Travis run the unit tests.

dandsager1: removed

storm-server/src/test/java/org/apache/storm/scheduler/resource/TestResourceAwareScheduler.java
```diff
@@ -1067,6 +1068,208 @@ public void minCpuWorkerSplitFails() {
...
+        }
+    }
+
+    /*
+     * Test time to schedule large cluster scheduling with fragmentation
+     */
+    private TimeBlockResult testLargeClusterSchedulingTiming(int numNodes, String Strategy) {
+        Config config = null;
+        if (Strategy.equals(DefaultResourceAwareStrategy.class.getName())) {
+            config = createClusterConfig(10, 10, 0, null);
```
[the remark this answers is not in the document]
dandsager1: I don't know how this would be structured. I can imagine an interface class and then 3 derived classes, each with its own createClusterConfig method, but it sounds like you have something different in mind?

As a recent convert from C++, I think of using function pointers, but that concept is not in Java.

storm-server/src/test/java/org/apache/storm/scheduler/resource/TestResourceAwareScheduler.java
```diff
@@ -1067,6 +1068,208 @@ public void minCpuWorkerSplitFails() {
...
+        LOG.info("TestLargeFragmentedClusterScheduling took {} ms", Time.currentTimeMillis() - startTime);
+        for (int strategyIdx = 0; strategyIdx < numStrategies; ++strategyIdx) {
+            double medianFirstBlockTime = getMedianBlockTime(testResults[strategyIdx], true);
+            double medianLastBlockTime = getMedianBlockTime(testResults[strategyIdx], false);
+            double ratio = medianLastBlockTime / medianFirstBlockTime;
+            LOG.info("{}, FirstBlock {}, LastBlock {} ratio {}", strategies[strategyIdx], medianFirstBlockTime, medianLastBlockTime, ratio);
+        }
+
+        // Check last block scheduling time does not get significantly slower
+        final double[] acceptedStrategyTimeRatios = {6.96, 7.78, 7.75};
```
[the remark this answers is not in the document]
dandsager1: Added comment explaining values were obtained empirically described in comment block above.

dandsager1:
The intent is to ensure that adverse changes to scheduling are not discovered at deploy time. It could also be used as a tool for making scheduling faster.

storm-server/src/test/java/org/apache/storm/scheduler/resource/TestResourceAwareScheduler.java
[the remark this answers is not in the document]
srdo: No, I meant make the Config a parameter of this method, so `testLargeClusterSchedulingTiming(int numNodes, Config config)`. You can then either do `Map<String, Config> strategyToConfig`, or have a static inner class containing the strategy name and Config in line https://github.com/apache/storm/pull/3092/files#diff-36af6e89e1390a52764241458869fd76R1142.

Not that I think you should use them here, but Java does have something similar to function pointers (not as laissez faire though): https://docs.oracle.com/javase/tutorial/java/javaOO/methodreferences.html

storm-server/src/test/java/org/apache/storm/scheduler/resource/TestResourceAwareScheduler.java
```diff
@@ -1067,6 +1068,215 @@ public void minCpuWorkerSplitFails() {
...
+        for (Entry<String, TimeBlockResult> strategyResult : strategyToTimeBlockResults.entrySet()) {
+            TimeBlockResult strategyTimeBlockResult = strategyResult.getValue();
+            double medianFirstBlockTime = getMedianValue(strategyTimeBlockResult.firstBlockTime);
+            double medianLastBlockTime = getMedianValue(strategyTimeBlockResult.lastBlockTime);
+            double ratio = medianLastBlockTime / medianFirstBlockTime;
+
+            double slowSchedulingThreshold = 1.5;
+            String msg = "Strategy " + strategyResult.getKey() + " scheduling is significantly slower for mostly full fragmented cluster\n";
+            msg += "Ratio was " + ratio + " Max allowed is " + (slowSchedulingThreshold * ratio);
+            assertTrue(msg, ratio < slowSchedulingThreshold * strategyToAcceptedBlockTimeRatios.get(strategyResult.getKey()));
```
srdo: This is not something you have to change, but as you mentioned you are new to Java, I thought I'd mention this as you may be unfamiliar with JUnit and Hamcrest:

There's an `assertThat` method you can use to do this kind of assertion in a way where the error message will be useful automatically. Writing this as `assertThat(ratio, lessThan(slowSchedulingThreshold * strategyToAcceptedBlockTimeRatios.get(strategyResult.getKey()))` gets you an automatic error message similar to what you've written manually here.

srdo:
+1 assuming tests pass. If the test running time becomes an issue, we can either try to make scheduling faster, or put it in a category of slow tests that aren't run by default, but do run on CI.

storm-server/src/test/java/org/apache/storm/scheduler/resource/TestResourceAwareScheduler.java
dandsager1: Thanks for the info. I'll be sure to use `assertThat` in the future.

PR #3093, opened 2019-07-25
STORM-3476 don't query remote files on cleanup if target size is acce…

agresch:
…ptable

I'm seeing supervisors querying files over and over on the namenode from this method and never deleting them, adding namenode load.  If we're below the cleanup target size, it seems we should just not bother doing these checks.


srdo:
+1, I guess there's no harm to not cleaning up unused blobs unless we're over the size limit.

PR #3094, opened 2019-07-26
STORM-3477 switch HDFS blobstore isRemoteBlobExists to use file exist…

agresch:
…s for lower overhead

### DIFF
--- a/external/storm-hdfs-blobstore/src/main/java/org/apache/storm/hdfs/blobstore/HdfsBlobStore.java
+++ b/external/storm-hdfs-blobstore/src/main/java/org/apache/storm/hdfs/blobstore/HdfsBlobStore.java
@@ -336,6 +336,27 @@ public class HdfsBlobStore extends BlobStore {
         }
     }
 
+    /**
+     * Checks if a blob exists.
+     *
+     * @param key blobstore key
+     * @param who subject
+     * @throws AuthorizationException if authorization is failed
+     * @throws KeyNotFoundException if key is not found
+     */
+    public boolean blobExists(String key, Subject who) throws AuthorizationException, KeyNotFoundException {
+        who = checkAndGetSubject(who);
+        validateKey(key);
+        SettableBlobMeta meta = getStoredBlobMeta(key);
+        aclHandler.hasPermissions(meta.get_acl(), READ, who, key);
+        try {
+            BlobStoreFile blobStoreFile = hbs.read(DATA_PREFIX + key);
+            return blobStoreFile.exists();
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
     @Override
     public Iterator<String> listKeys() {
         try {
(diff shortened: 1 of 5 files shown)


external/storm-hdfs-blobstore/src/main/java/org/apache/storm/hdfs/blobstore/HdfsBlobStore.java
```diff
@@ -336,6 +336,27 @@ public InputStreamWithMeta getBlob(String key, Subject who)
...
+     * @throws KeyNotFoundException if key is not found
+     */
+    public boolean blobExists(String key, Subject who) throws AuthorizationException, KeyNotFoundException {
+        who = checkAndGetSubject(who);
+        validateKey(key);
+        SettableBlobMeta meta = getStoredBlobMeta(key);
+        aclHandler.hasPermissions(meta.get_acl(), READ, who, key);
+        try {
+            BlobStoreFile blobStoreFile = hbs.read(DATA_PREFIX + key);
+            return blobStoreFile.exists();
```
Ethanlm: We don't need this.  
For every key,  there will be two files `meta_<key>` and `data_<key>`. Every operation to the blob will first check and get `meta_<key>`.  If the blob is deleted, it throws `KeyNotFoundException`. If it exists, we don't need to continue to check `data_<key>` because we can simply assume `data_<key>` exists. Because if `meta_<key>`exists but  `data_<key>` doesn't, there will be RuntimeException with the client tries to download or access to the blob. 

In this way, we can reduce the hdfs operation further.

agresch: So basically we're saying that a KeyNotFoundException indicates the blobDoes not exist from line 350 and any other non-AuthorizationException would return true?

Ethanlm: There will no  non-AuthorizationException (after removing L352 to L357).

Ethanlm: I think it's more clear if we consider the blob exists when `meta_<key>` exists.

agresch:
@Ethanlm - see if this matches what you expect.  Thanks.

Ethanlm:
Yeah that's what I was thinking about.  Based on this, I am thinking it's probably better to check `KeyNotFoundException` inside the `blobExists` function and if it happens, returns false.   Then `isRemoteBlobExists` can simply invoke `blobExists` function.

Do we want to add a  `blobExists` function to `LocalFsBlobStore` too? So they can have the same functionality?

agresch:
blobexists() is added strictly for isRemoteBlobExists(), which LocalFsBlobStore does not need.  I think we can forgo adding it.

I agree with your first comment about KeyNotFoundException.  Will rework it.

agresch:
@Ethanlm - please check again.

PR #3095, opened 2019-07-29
STORM-3478: Upgrade to JUnit 5.5.1 and Mockito 3.0.0

srdo:
Also bumping Mockito version because why not, it's been a while.


external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/KafkaSpoutMessagingGuaranteeTest.java
```diff
@@ -90,8 +90,6 @@ public void testAtMostOnceModeCommitsBeforeEmit() throws Exception {
 
         spout.nextTuple();
 
-        when(consumerMock.position(partition)).thenReturn(1L);
```
srdo: Just for context, these are unused stubs, which Mockito complains about.

PR #3097, opened 2019-08-05
STORM-3475 Add ConstraintSolverStrategy Unit Test

dandsager1:
exposes scheduling StackOverflowError with large number of executors/constraints

### DIFF
--- a/storm-server/src/test/java/org/apache/storm/scheduler/resource/strategies/scheduling/TestConstraintSolverStrategy.java
+++ b/storm-server/src/test/java/org/apache/storm/scheduler/resource/strategies/scheduling/TestConstraintSolverStrategy.java
@@ -37,7 +37,9 @@ import org.apache.storm.scheduler.resource.strategies.priority.DefaultScheduling
 import org.apache.storm.utils.Time;
 import org.apache.storm.utils.Utils;
 import org.junit.Assert;
-import org.junit.Test;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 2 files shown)


storm-server/src/test/java/org/apache/storm/scheduler/resource/strategies/scheduling/TestConstraintSolverStrategy.java
```diff
@@ -194,6 +196,51 @@ protected SolverResult backtrackSearch(SearcherState state) {
...
+        addContraints("bolt-1", "bolt-2", constraints);
+        addContraints("bolt-1", "bolt-0", constraints);
+        addContraints("bolt-1", "spout-0", constraints);
+
+        config.put(Config.TOPOLOGY_RAS_CONSTRAINTS, constraints);
+        TopologyDetails topo = genTopology("testTopo-" + parallelismMultiplier, config, 10, 10, 30 * parallelismMultiplier, 30 * parallelismMultiplier, 31414, 0, "user");
+        Topologies topologies = new Topologies(topo);
+
+        Map<String, SupervisorDetails> supMap = genSupervisors(30 * parallelismMultiplier, 30, 3500, 35000);
+        Cluster cluster = new Cluster(new INimbusTest(), new ResourceMetrics(new StormMetricsRegistry()), supMap, new HashMap<String, SchedulerAssignmentImpl>(), topologies, config);
```
Ethanlm: we should reuse `makeCluster` function. For example: https://github.com/apache/storm/pull/3098/files: 
Create a `public Cluster makeCluster(Topologies topologies, Map<String, SupervisorDetails> supMap)` for this .

storm-server/src/test/java/org/apache/storm/scheduler/resource/strategies/scheduling/TestConstraintSolverStrategy.java
```diff
@@ -194,6 +196,51 @@ protected SolverResult backtrackSearch(SearcherState state) {
...
+
+        ResourceAwareScheduler scheduler = new ResourceAwareScheduler();
+        scheduler.prepare(config);
+        scheduler.schedule(topologies, cluster);
+
+        boolean scheduleSuccess = isStatusSuccess(cluster.getStatus(topo.getId()));
+
+        if (parallelismMultiplier == 5) {
```
Ethanlm: I would like this to be a range. Maybe `<=5` or if you can find the largest number that pass the test. 
 So if `@ValueSource(ints = { 5, 20 })` is changed, it won't surprise people

storm-server/src/test/java/org/apache/storm/scheduler/resource/strategies/scheduling/TestConstraintSolverStrategy.java
```diff
@@ -202,6 +204,50 @@ protected SolverResult backtrackSearch(SearcherState state) {
...
+
+        ResourceAwareScheduler scheduler = new ResourceAwareScheduler();
+        scheduler.prepare(config);
+        scheduler.schedule(topologies, cluster);
+
+        boolean scheduleSuccess = isStatusSuccess(cluster.getStatus(topo.getId()));
+
+        if (parallelismMultiplier <= 5) {
+            Assert.assertTrue(scheduleSuccess);
+        } else if (parallelismMultiplier == 20) {
```
Ethanlm: Maybe I was not clear. I think we should find a number/range so that we can have
```
if (parallelismMultiplier <= xx) {
  ...
} else {
  ...
}
```
But I am fine with the current code.

PR #3104, opened 2019-08-16
[STORM-3484] Add Cluster Blacklist to UI to show what supervisors are…

RuiLi8080:
… blacklisted

### DIFF
--- a/storm-client/src/storm.thrift
+++ b/storm-client/src/storm.thrift
@@ -194,6 +194,7 @@ struct SupervisorSummary {
   9: optional double used_cpu;
   10: optional double fragmented_mem;
   11: optional double fragmented_cpu;
+  12: optional bool blacklisted;
 }
 
 struct NimbusSummary {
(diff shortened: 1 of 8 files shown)


agresch:
We still have a change to push back that allows multiple supervisor ids on a single host for use on NUMA.  While doing this, we ran into some bugs where host/node ids were not necessarily synonymous.

For clarity, could you update the naming from of methods/variables something like: getBlacklistedSupervisors() to getBlacklistedSupervisorsHosts()?  This will help prevent issues going forward.

We still have an open task to push this NUMA code back.

RuiLi8080:
The getBlacklistedSupervisors() call added here is intended to return a list of SupervisorSummaries. I notice that the SupervisorSummary objects are usually referred as supervisors in the thrift file. I guess this method naming could make more sense here?
I did some renaming in the BlacklistScheduler to differentiate Host and Id. I hope this helps to address your comments.

RuiLi8080:
Addressed comments by including blacklist information into a column of supervisor summary table.

storm-client/src/storm.thrift
```diff
@@ -759,6 +759,7 @@ service Nimbus {
   void createStateInZookeeper(1: string key); // creates state in zookeeper when blob is uploaded through command line
 
   // need to add functions for asking about status of storms, what nodes they're running on, looking at task logs
+  set<string> getBlacklistedSupervisorIds()
```
Ethanlm: We don't need to expose this function to client.  `getSupervisorSummary` and `getSupervisorPageInfo` are the two places that needs this information. They are all inside nimbus.java. We can just have a private `getBlacklistedSupervisorIds` function in `nimbus.java` .

RuiLi8080: Yes you are right. Addressed.

RuiLi8080:
Tested. Blacklist information propagates to UI correctly. Thanks to @Ethanlm for the help

PR #3106, opened 2019-08-19
[STORM-3491] BoltReaderRunnable shouldn't throw IllegalArgumentException for sync command

Ethanlm:
Refer to https://issues.apache.org/jira/browse/STORM-3491


srdo:
I'm wondering if we should revert back to the old behavior of accepting any string? Someone else may be depending on that behavior.

Ethanlm:
Makes sense. Fixing it

srdo:
+1, thanks

Ethanlm:
Thanks for the review

PR #3107, opened 2019-08-19
MINOR: remove scm tag in pom.xml

Ethanlm:
I think this is not needed. And it can be confusing when it's not updated correctly

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -242,7 +242,6 @@
     <scm>
         <connection>scm:git:https://gitbox.apache.org/repos/asf/storm.git</connection>
         <developerConnection>scm:git:https://gitbox.apache.org/repos/asf/storm.git</developerConnection>
-        <tag>v2.0.0</tag>
         <url>https://gitbox.apache.org/repos/asf/storm</url>
     </scm>
 


Ethanlm:
resolved the conflicts.

srdo:
Isn't this getting set by the release plugin? If so, it will show up again next time we do a release candidate.

When is the tag tag not being updated correctly?

srdo:
https://stackoverflow.com/questions/23718601/what-is-the-usage-of-maven-pom-xml-tag-element-inside-of-scm-when-you-are

Ethanlm:
I wanted to remove this from master because master is not a release branch. It doesn't make much sense to have `v2.0.0` in master branch especially we are going to move master to `2.2.x-SNAPSHOT`

srdo:
Yes, and that makes sense. But I think the tag will just show up again next time someone makes an RC. It only stays gone if we never do a release from master without branching, and then also never merge the release branches into master.

Ethanlm:
We shouldn't do a minor version release without branching. I don't see why we would merge a release branch to master. What we should do is cherry-pick, right?

srdo:
Yes, you are right. +1. I don't see the harm of doing this, worst case the tag just shows up again.

Ethanlm:
Thanks for the review

PR #3108, opened 2019-08-20
[STORM-3492] Add config to prevent good supervisor with bad workers f…

RuiLi8080:
…rom going to blacklist when necessary

- This is a configurable parameter, and it won't change default behaviors

### DIFF
--- a/storm-server/src/main/java/org/apache/storm/DaemonConfig.java
+++ b/storm-server/src/main/java/org/apache/storm/DaemonConfig.java
@@ -138,6 +138,17 @@ public class DaemonConfig implements Validated {
     public static final String BLACKLIST_SCHEDULER_STRATEGY = "blacklist.scheduler.strategy";
 
     /**
+     * Whether {@link org.apache.storm.scheduler.blacklist.BlacklistScheduler} will assume the supervisor is bad
+     * based on bad slots or not.
+     * A bad slot indicates the situation where the nimbus doesn't receive heartbeat from the worker in time,
+     * it's hard to differentiate if it's because of the supervisor node or the worker itself.
+     * If this is set to true, the scheduler will consider a supervisor is bad when seeing bad slots in it.
+     * Otherwise, the scheduler will assume a supervisor is bad only when it does not receive supervisor heartbeat in time.
+     */
+    @IsBoolean
+    public static final String BLACKLIST_SCHEDULER_ASSUME_SUPERVISOR_BAD_BASED_ON_BAD_SLOT = "blacklist.scheduler.assume.supervisor.bad.based.on.bad.slot";
+
+    /**
      * Whether we want to display all the resource capacity and scheduled usage on the UI page. You MUST have this variable set if you are
      * using any kind of resource-related scheduler.
      * <p/>
--- a/storm-server/src/main/java/org/apache/storm/scheduler/blacklist/BlacklistScheduler.java
+++ b/storm-server/src/main/java/org/apache/storm/scheduler/blacklist/BlacklistScheduler.java
@@ -55,6 +55,7 @@ public class BlacklistScheduler implements IScheduler {
     protected EvictingQueue<HashMap<String, Set<Integer>>> badSupervisorsToleranceSlidingWindow;
     protected int windowSize;
     protected volatile Set<String> blacklistedSupervisorIds;     // supervisor ids
+    private boolean blacklistOnBadSlots;
     private Map<String, Object> conf;
 
     public BlacklistScheduler(IScheduler underlyingScheduler, StormMetricsRegistry metricsRegistry) {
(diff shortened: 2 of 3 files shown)


storm-server/src/main/java/org/apache/storm/scheduler/blacklist/BlacklistScheduler.java
```diff
@@ -132,20 +134,20 @@ private void badSupervisors(Map<String, SupervisorDetails> supervisors) {
...
+            for (Map.Entry<String, SupervisorDetails> entry : supervisors.entrySet()) {
+                String key = entry.getKey();
+                SupervisorDetails supervisorDetails = entry.getValue();
+                if (cachedSupervisors.containsKey(key)) {
+                    Set<Integer> badSlots = badSlots(supervisorDetails, key);
+                    if (badSlots.size() > 0) { //supervisor contains bad slots
+                        badSupervisors.put(key, badSlots);
+                    }
+                } else {
+                    cachedSupervisors.put(key, supervisorDetails.getAllPorts()); //new supervisor to cache
```
Ethanlm: I think this line should execute anyway no matter if `blacklistOnBadSlots` is true or false

storm-server/src/main/java/org/apache/storm/scheduler/blacklist/BlacklistScheduler.java
```diff
@@ -89,6 +90,7 @@ public void prepare(Map<String, Object> conf) {
         badSupervisorsToleranceSlidingWindow = EvictingQueue.create(windowSize);
         cachedSupervisors = new HashMap<>();
         blacklistHost = new HashSet<>();
+        blacklistOnBadSlots = ObjectReader.getBoolean(this.conf.get(DaemonConfig.BLACKLIST_SUPERVISOR_ON_BAD_SLOT), true);
```
Ethanlm: We don't need default value here since it's already in `defaults.yaml`

RuiLi8080: Other variables using this format have default value set in default.yaml as well. And the ObjectReader.getBoolean() function require a default input. Should I still make change here?

storm-server/src/main/java/org/apache/storm/DaemonConfig.java
```diff
@@ -137,6 +137,13 @@
     @IsImplementationOfClass(implementsClass = IBlacklistStrategy.class)
     public static final String BLACKLIST_SCHEDULER_STRATEGY = "blacklist.scheduler.strategy";
 
+    /**
+     * Whether the blacklistscheduler will put a supervisor to blacklist when there is a bad slot/worker
```
Ethanlm: We could clarify this a little bit here. It's not really put into the blacklist directly because of a bad slot. 

Also in the javadoc, we can refer to blacklist scheduler using {@link org.apache.storm.scheduler.blacklist.BlacklistScheduler}.

storm-server/src/main/java/org/apache/storm/scheduler/blacklist/BlacklistScheduler.java
Ethanlm: Okay. I am fine with it for now.

storm-server/src/main/java/org/apache/storm/DaemonConfig.java
```diff
@@ -137,6 +137,14 @@
     @IsImplementationOfClass(implementsClass = IBlacklistStrategy.class)
     public static final String BLACKLIST_SCHEDULER_STRATEGY = "blacklist.scheduler.strategy";
 
+    /**
+     * Whether the {@link org.apache.storm.scheduler.blacklist.BlacklistScheduler} will put a
+     * (supervisor:slots) map to badSupervisorsToleranceSlidingWindow
```
Ethanlm: This will be too concrete about the implementations  and not easy for users to understand.  Maybe something like:

Whether {@link org.apache.storm.scheduler.blacklist.BlacklistScheduler}  will assume the supervisor is bad based on bad slots or not.  When some slot is bad, meaning the nimbus doesn't receive heartbeat from the worker in time, it's hard to determine if it's because of the supervisor node or the worker itself having issues.  If this is set to true, the scheduler will simply assume the supervisor is bad if there is any bad slot; otherwise, the scheduler will simply assume it's not a supervisor problem.

RuiLi8080: Appreciate the suggestions. All make sense to me. Please check the latest changes.

storm-server/src/main/java/org/apache/storm/DaemonConfig.java
```diff
@@ -137,6 +137,17 @@
...
+    /**
+     * Whether {@link org.apache.storm.scheduler.blacklist.BlacklistScheduler} will assume the supervisor is bad
+     * based on bad slots or not.
+     * A bad slot indicates the situation where the nimbus doesn't receive heartbeat from the worker in time,
+     * it's hard to differentiate if it's because of the supervisor node or the worker itself.
+     * If this is set to true, the scheduler will consider a supervisor is bad when seeing bad slots in it.
+     * Otherwise, the scheduler will assume a supervisor is bad only when it does not receive supervisor heartbeat in time.
+     */
+    @IsBoolean
+    public static final String BLACKLIST_SUPERVISOR_BASED_ON_BAD_SLOT = "blacklist.scheduler.assume.supervisor.bad.based.on.bad.slot;
```
Ethanlm: Let's make the variable name and the value consistent.

Ethanlm:
I merged the other pr so this pr has some conflicts. Please rebase and I will merge it if travis build passes

PR #3109, opened 2019-08-21
Set pom version to 2.2.0-SNAPSHOT

Ethanlm:
We created 2.1.x-branch for 2.1.x release. We then should move master to 2.2.0-SNAPSHOT.

This pr is the result of 

`mvn versions:set -DnewVersion=2.2.0-SNAPSHOT  -P dist,rat,externals,examples`

### DIFF
--- a/examples/storm-elasticsearch-examples/pom.xml
+++ b/examples/storm-elasticsearch-examples/pom.xml
@@ -21,7 +21,7 @@
     <parent>
         <artifactId>storm</artifactId>
         <groupId>org.apache.storm</groupId>
-        <version>2.0.1-SNAPSHOT</version>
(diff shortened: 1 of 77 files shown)


Ethanlm:
Not sure if I should just commit the change directly to master after I created 2.1.x-branch or make a pull request like this so people can review.

srdo:
I think usually it's done without a PR. I think the only thing people could give their opinion on is the version number, and hopefully we'd agree that on the mailing list before branching.

+1.

Ethanlm:
Thanks. So next time when we do branching, we can discuss in the mailing list. Then the release manager can go ahead, do branching, update master pom version, and continue. I will write this down and update the doc later so everyone will be clear about this.

PR #3111, opened 2019-08-21
 STORM-3493: Allow overriding python interpreter by environment variable

hsn10:
This is specially needed on FreeBSD because there is no "python" = there is python2.7

### DIFF
--- a/bin/storm
+++ b/bin/storm
@@ -34,7 +34,9 @@ while [ -h "${PRG}" ]; do
 done
 
 # check for version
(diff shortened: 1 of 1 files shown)


Ethanlm:
It looks good. Could you please file a jira at https://issues.apache.org/jira/projects/STORM and update your PR summary and commit message to include the jira number? 

It can be something like `STORM-XXXX: ...` or `[STORM-XXXX] ...`

Thanks.

PR #3112, opened 2019-08-21
[STORM-3494] use HdfsLoginUtil singleton to login to hdfs

Ethanlm:
refer to https://issues.apache.org/jira/browse/STORM-3494

Use HdfsLoginUtil singleton to login to hdfs exactly once for a process.


storm-client/test/jvm/org/apache/storm/utils/ConfigUtilsTest.java
```diff
@@ -92,32 +92,32 @@ public void getValueAsList_nonStringList() {
     }
 
     @Test
-    public void getBlobstoreHDFSPrincipal() throws UnknownHostException {
```
agresch: do we want to keep this test as is for validating deprecated configs?  It looked like they should still work?

dev-tools/travis/travis-script.sh
```diff
@@ -43,7 +43,7 @@ elif [ "$2" == "External" ]
 then
   if [ "$TRAVIS_JDK_VERSION" == "openjdk11" ]
   then 
-    TEST_MODULES='!storm-client,!storm-server,!storm-core,!storm-webapp,!storm-shaded-deps,!external/storm-cassandra,!external/storm-hive,!external/storm-hdfs,!external/storm-hbase,!sql/storm-sql-external/storm-sql-hdfs,!external/storm-hdfs-blobstore'
+    TEST_MODULES='!storm-client,!storm-server,!storm-core,!storm-webapp,!storm-shaded-deps,!external/storm-cassandra,!external/storm-hive,!external/storm-hdfs,!external/storm-hbase,!sql/storm-sql-external/storm-sql-hdfs,!external/storm-hdfs-blobstore,!external/storm-hdfs-login'
```
Ethanlm: Not sure why we have different `TEST_MODULES` on openjdk11 in first place. I added it here because `storm-hdfs-login` is similar to `storm-hdfs-blobstore`

storm-client/test/jvm/org/apache/storm/utils/ConfigUtilsTest.java
Ethanlm: Makes sense. Will fix it.

storm-client/src/jvm/org/apache/storm/utils/HdfsLoginUtil.java
```diff
@@ -0,0 +1,81 @@
...
+     * We use the single HdfsLoginPlugin instance because it may use some local state/cache.
+     * @param conf the storm configuration
+     * @return the logged in subject
+     */
+    public Subject logintoHdfs(Map<String, Object> conf) {
+
+        if (hdfsLoginPlugin == null) {
+            synchronized (HdfsLoginUtil.class) {
+                if (hdfsLoginPlugin == null) {
+                    String className = (String) conf.get(Config.STORM_HDFS_LOGIN_PLUGIN);
```
agresch: any reason not to default this?

Ethanlm: We will then introduce hdfs dependencies to `storm-client`. I think (but I am not sure) it's also part of the reason that we separate storm-hdfs-blobstore out.

Ethanlm:
Travis build failed due to non default config of `storm.hdfs.login.plugin` available. Will look into it.

dev-tools/travis/travis-script.sh
srdo: It's because cassandra and all of Hadoop don't run on Java 11 yet. See https://issues.apache.org/jira/browse/STORM-2798

Ethanlm: Thank you for clarification

Ethanlm:
Will redo this when ready

PR #3113, opened 2019-08-21
STORM-3482 Implement One Worker Per Component Option

dandsager1:


Ethanlm:
Could you please resolve the conflicts? This PR looks almost ready to merge

storm-client/src/jvm/org/apache/storm/Config.java
```diff
@@ -321,6 +321,12 @@
      */
     @IsBoolean
     public static final String TOPOLOGY_RAS_ONE_EXECUTOR_PER_WORKER = "topology.ras.one.executor.per.worker";
+    /**
+     * Whether to limit each worker to one component. This is useful for debugging topologies to clearly identify workers that
+     * are slow/crashing and for estimating resource requirements and capacity.
+     */
```
Ethanlm: Also I think we can add some comment here to be more clear about which one will be used if both of these are set true.

storm-client/src/jvm/org/apache/storm/Config.java
```diff
@@ -318,9 +318,19 @@
     /**
      * Whether to limit each worker to one executor. This is useful for debugging topologies to clearly identify workers that
      * are slow/crashing and for estimating resource requirements and capacity.
+     * If both TOPOLOGY_RAS_ONE_EXECUTOR_PER_WORKER and TOPOLOGY_RAS_ONE_COMPONENT_PER_WORKER are enabled,
```
Ethanlm: We could use javadoc style comments to link the constants or classes. For example, in this case,  `{@link #TOPOLOGY_RAS_ONE_EXECUTOR_PER_WORKER}` But I am fine with it as it is

storm-server/src/main/java/org/apache/storm/scheduler/resource/RasNode.java
```diff
@@ -362,15 +364,40 @@ public void assignSingleExecutor(WorkerSlot ws, ExecutorDetails exec, TopologyDe
      * @return true if it would fit else false
      */
     public boolean wouldFit(WorkerSlot ws, ExecutorDetails exec, TopologyDetails td) {
-        assert nodeId.equals(ws.getNodeId()) : "Slot " + ws + " is not a part of this node " + nodeId;
```
Ethanlm: I didn't notice this before. Why do we want to delete this line

storm-client/src/jvm/org/apache/storm/Config.java
```diff
@@ -318,9 +318,19 @@
...
      * are slow/crashing and for estimating resource requirements and capacity.
+     * If both {@link #TOPOLOGY_RAS_ONE_EXECUTOR_PER_WORKER} and {@link #TOPOLOGY_RAS_ONE_COMPONENT_PER_WORKER} are enabled,
+     * {@link #TOPOLOGY_RAS_ONE_COMPONENT_PER_WORKER} is ignored.
      */
     @IsBoolean
     public static final String TOPOLOGY_RAS_ONE_EXECUTOR_PER_WORKER = "topology.ras.one.executor.per.worker";
+    /**
+     * Whether to limit each worker to one component. This is useful for debugging topologies to clearly identify workers that
+     * are slow/crashing and for estimating resource requirements and capacity.
+     * If both TOPOLOGY_RAS_ONE_EXECUTOR_PER_WORKER and TOPOLOGY_RAS_ONE_COMPONENT_PER_WORKER are enabled,
```
Ethanlm: Ideally we want to use javadoc style here too. But it's okay

PR #3115, opened 2019-08-23
STORM-3490: Add checkstyle rule RedundantModifier

krichter722:


storm-client/src/jvm/org/apache/storm/cluster/IStateStorage.java
```diff
@@ -192,7 +192,7 @@
      *
      * @param listener A StateStorageListener to handle changing cluster state events.
      */
-    void add_listener(final ConnectionStateListener listener);
```
Ethanlm: here too

storm-client/src/jvm/org/apache/storm/topology/base/BaseTickTupleAwareRichBolt.java
```diff
@@ -56,5 +56,5 @@ protected void onTickTuple(final Tuple tuple) {
      *
      * @param tuple The input tuple to be processed.
      */
-    protected abstract void process(final Tuple tuple);
```
Ethanlm: here too

storm-server/src/main/java/org/apache/storm/scheduler/resource/strategies/scheduling/BaseResourceAwareStrategy.java
```diff
@@ -137,8 +137,8 @@ protected boolean scheduleExecutor(
     }
 
     protected abstract TreeSet<ObjectResources> sortObjectResources(
-        final AllResources allResources, ExecutorDetails exec, TopologyDetails topologyDetails,
-        final ExistingScheduleFunc existingScheduleFunc
+        AllResources allResources, ExecutorDetails exec, TopologyDetails topologyDetails,
```
Ethanlm: here too

krichter722:
@Ethanlm Thanks for your review. All `final` modifier are removed on abstract methods (in abstract classes or interfaces) where they have no purpose because they don't influence the parameter of the overridden method (`final` can be added or omitted there no matter how the superclass' method looks like).

The `public` modifiers are generally removed where they can't be used, e.g. on package-private (inner) classes. All changes made sense for me, but please review carefully again.

Ethanlm:
It makes sense to me. I will find some time and go through it again. Thanks for explaining.

storm-client/src/jvm/org/apache/storm/validation/ConfigValidationUtils.java
```diff
@@ -143,15 +143,15 @@ public void validateField(String pd, String name, Object field)
     /**
      * Declares methods for validating configuration values.
      */
-    public static interface FieldValidator {
+    interface FieldValidator {
```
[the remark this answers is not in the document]
krichter722: Correct, only `static` is redundant here, I corrected that in an added commit. Thanks for the review.

PR #3116, opened 2019-08-23
Remove caching of $HOME/.m2/repository

krichter722:
Remove caching of $HOME/.m2/repository because it causes missing build dependencies to be unspotted

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,6 +51,5 @@ script:
   - /bin/bash ./dev-tools/travis/travis-script.sh `pwd` $MODULES
 cache:
   directories:
-    - "$HOME/.m2/repository"
     - "$HOME/.rvm"
     - "$NVM_DIR"


srdo:
Could you elaborate on what type of error this caching would hide?

krichter722:
> Could you elaborate on what type of error this caching would hide?

Maven artifacts which are referenced in a POM, but are no longer available in remote repositories. The build thus fails for everyone who doesn't have the dependency cached.

srdo:
> Maven artifacts which are referenced in a POM, but are no longer available in remote repositories

Maven artifacts aren't supposed to disappear this way. As far as I know you can't remove artifacts from e.g. Maven Central. I don't think we should remove caching to account for artifacts being deleted from remote repositories, since it shouldn't be happening. 

I'd rather wait to remove caching until we see build instability due to this. Did you see this situation happen?

krichter722:
SNAPSHOT versions can disappear. It's common practice to delete all snapshot artifacts as soon as the corresponding release is made. During development SNAPSHOT versions can be used even though they'll probably/should be replaced with non-SNAPSHOT versions before a release (of Storm). That's something you want to detect in CI afaik.

srdo:
I don't think we use SNAPSHOT versions of third party dependencies. Using them would prevent us from releasing until the third party project produces a non-SNAPSHOT release, as we can't release Storm with a dependency on a SNAPSHOT component. 

For the in-development code, we still tend not to have introduce SNAPSHOT dependencies on third party dependencies during development. At least I've never seen it happen. The only SNAPSHOTs that are part of Storm during development tends to be Storm's own modules. We clear those out before every build here https://github.com/apache/storm/blob/e60443a59a7d9ad70ebb9f28ae281f836e0bc1a1/.travis.yml#L29

krichter722:
Right, then I'll follow your proposal to wait until a build fails due to the problem. I agree that it shouldn't happen, but only in theory practice is equals to theory.

PR #3117, opened 2019-08-23
STORM-3495 TestConstraintSolverStrategy is not stable on travis

dandsager1:


dandsager1:
Decrease number of executors to schedule to avoid stack overflow.

storm-server/src/test/java/org/apache/storm/scheduler/resource/strategies/scheduling/TestConstraintSolverStrategy.java
```diff
@@ -210,7 +210,7 @@ protected SolverResult backtrackSearch(SearcherState state) {
      * Cluster has sufficient resources for scheduling to succeed but can fail due to StackOverflowError.
      */
     @ParameterizedTest
-    @ValueSource(ints = {5, 20})
+    @ValueSource(ints = {1, 20})
```
Ethanlm: I would also update L241
 https://github.com/apache/storm/blob/master/storm-server/src/test/java/org/apache/storm/scheduler/resource/strategies/scheduling/TestConstraintSolverStrategy.java#L241-L248 to reflect the change. 

Maybe change to 

```
if (parallelismMultiplier <= 1) {
   xxxx;
} else {
   yyyy;
}
```

srdo:
I don't have an opinion on the actual change, but would it be possible to make whatever code is recursing do iterations instead? It seems bad that we hit a stack overflow on a pretty small number of executors.

dandsager1:
It's not a straightforward conversion to an iterative method. I'll file a jira on converting to iterative.

srdo:
Thanks

PR #3118, opened 2019-08-27
STORM-3500: Fix spelling error in property storm.blobstore.dependency…

srdo:
….jar.upload.chuck.size.bytes

This is minor, but we might as well fix it before the spelling is set in stone. This needs to also go in 2.1.x


srdo:
Applied to master and 2.1.x

PR #3119, opened 2019-08-27
Revert "STORM-3476 don't query remote files on cleanup if target size…

srdo:
… is acceptable"

I've reopened https://issues.apache.org/jira/browse/STORM-3476 as it is causing an issue when topologies are killed. I verified that this issue is causing the problem in the supervisor by running a topology with the current code and this code commented out.

This needs to be applied here and in 2.1.x.


srdo:
Me too.

srdo:
Applied to master and 2.1.x-branch

PR #3120, opened 2019-08-29
[STORM-3501] local cluster worker restarts because of missing resources folder

diogopmonteiro:
Creating resources directory if it does not exist in resources jar or in classpath.
Moved a bit of code to make mocking easier.

Happy to hear about your thoughts.


diogopmonteiro:
I mocked AdvancedFSOps because the other unit tests on AsyncLocalizerTest are also mocking it. So, in my opinion we could open an other issue to change them all to not mock the filesystem. 

Ok, it seems better with that getResources method in ServerUtils. Changed.

srdo:
Ok, raised https://issues.apache.org/jira/browse/STORM-3504.

The rest of the changes look good. +1 if tests pass.

diogopmonteiro:
Thanks @srdo for your collaboration on this issue.

PR #3125, opened 2019-09-05
STORM-3506 prevent topo conf from overriding some system properties

agresch:


Ethanlm:
Actually after reviewing this again, I think this might not work as expected.  The code below this is
`Map<String, Object> totalConfToSave = Utils.merge(otherConf, topoConf);` and the `totalConfToSave` is actually the one that got used for topology. If this `STORM_CGROUP_HIERARCHY_DIR` config exists in `otherConf` but not in `topoConf`, it will eventually stay in `totalConfToSave`

agresch:
@Ethanlm - will look into your comment when I get a chance.

agresch:
@Ethanlm - I think you are right.  This change would fix if it's set by the topology, but not if a differing setting is on the client's storm.yaml.  Thanks for pointing this out.

agresch:
@Ethanlm - could you take a look again when you get a chance?

Ethanlm:
also applied to https://github.com/apache/storm/tree/2.1.x-branch

PR #3126, opened 2019-09-09
STORM-3504: changed AsyncLocalizerTest to not mock AdvancedFSOps

diogopmonteiro:
Removed AdvancedFSOps mocks from AsyncLocalizerTest, tests now use a temp path to execute.

### DIFF
--- a/storm-server/src/test/java/org/apache/storm/localizer/AsyncLocalizerTest.java
+++ b/storm-server/src/test/java/org/apache/storm/localizer/AsyncLocalizerTest.java
@@ -18,7 +18,6 @@ import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
-import java.io.OutputStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
@@ -94,7 +93,8 @@ public class AsyncLocalizerTest {
     private ClientBlobStore mockblobstore = mock(ClientBlobStore.class);
 
     private static String getTestLocalizerRoot() {
-        File f = new File("./target/" + Thread.currentThread().getStackTrace()[2].getMethodName() + "/localizer/");
+        File f = new File("./target/" + Thread.currentThread().getStackTrace()[2].getMethodName() + "/localizer/")
+                .getAbsoluteFile();
         f.deleteOnExit();
         return f.getPath();
     }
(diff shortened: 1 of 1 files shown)


storm-server/src/test/java/org/apache/storm/localizer/AsyncLocalizerTest.java
```diff
@@ -94,7 +93,8 @@
     private ClientBlobStore mockblobstore = mock(ClientBlobStore.class);
 
     private static String getTestLocalizerRoot() {
-        File f = new File("./target/" + Thread.currentThread().getStackTrace()[2].getMethodName() + "/localizer/");
+        File f = new File("./target/" + Thread.currentThread().getStackTrace()[2].getMethodName() + "/localizer/")
+                .getAbsoluteFile();
         f.deleteOnExit();
```
srdo: I think we should use TmpPath instead of deleteOnExit to clean up test files, as they then get deleted when the test is done rather than when the JVM shuts down. Here's an example https://github.com/apache/storm/blob/23cd491da1ddc3cb684eb0a196101af5af64f6a9/storm-server/src/test/java/org/apache/storm/LocalStateTest.java#L31

diogopmonteiro: Thanks for the suggestion, it's better that way.
Will change.

diogopmonteiro: Changed all tests that used that `baseDir` to use a `TmpPath`. 

I went a bit further and cleaned the tests a bit. They should be equivalent, but I think that now they are more readable. I hope that's ok.

storm-server/src/test/java/org/apache/storm/localizer/AsyncLocalizerTest.java
```diff
@@ -86,68 +83,66 @@
...
-            Future<Void> f = bl.requestDownloadBaseTopologyBlobs(pna, null);
+        ReflectionUtils mockedReflectionUtils = mock(ReflectionUtils.class);
+        ServerUtils mockedServerUtils = mock(ServerUtils.class);
+
+        ReflectionUtils previousReflectionUtils = ReflectionUtils.setInstance(mockedReflectionUtils);
+        ServerUtils previousServerUtils = ServerUtils.setInstance(mockedServerUtils);
+
+        // cannot use automatic resource management here in this try because the AsyncLocalizer depends on a config map,
+        // which should take the storm local dir, and that storm local dir is declared in the try-with-resources.
+        AsyncLocalizer victim = null;
```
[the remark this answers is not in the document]
diogopmonteiro: I think it makes it easier to identify the object under test.

PR #3127, opened 2019-09-10
[STORM-3507] add greylist for superviosrs which are forced out of blacklist due to low resources

RuiLi8080:
* Add greylist to cluster state
* When releasing supervisors due to low resources, put these released supervisors to greylist
* When scheduling, these greylisted supervisors will be scheduled only after having gone through other supervisors

### DIFF
--- a/storm-server/src/main/java/org/apache/storm/scheduler/Cluster.java
+++ b/storm-server/src/main/java/org/apache/storm/scheduler/Cluster.java
@@ -27,7 +27,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
-import java.util.function.Function;
 import java.util.stream.Collectors;
 import org.apache.storm.Config;
 import org.apache.storm.Constants;
@@ -84,6 +83,7 @@ public class Cluster implements ISchedulingState {
     private final ResourceMetrics resourceMetrics;
     private SchedulerAssignmentImpl assignment;
     private Set<String> blackListedHosts = new HashSet<>();
+    private List<String> greyListedSupervisors = new ArrayList<>();
     private INimbus inimbus;
     private double minWorkerCpu = 0.0;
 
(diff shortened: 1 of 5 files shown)


storm-server/src/main/java/org/apache/storm/scheduler/Cluster.java
```diff
@@ -1085,4 +1093,14 @@ public void updateFrom(Cluster other) {
...
         return minWorkerCpu;
     }
+
+    public List<String> getQuarantinedSupervisors() {
+        return quarantinedSupervisors;
+    }
+
+    public void setQuarantinedSupervisors(Set<String> quarantinedSupervisors) {
+        this.quarantinedSupervisors.clear();
+        LOG.info("Supervisors: {} were forced to quarantine due to resource shortage", quarantinedSupervisors);
```
govind-menon: Nit: were forced to be quarantined

RuiLi8080: Thanks Govind. Addressed

storm-server/src/main/java/org/apache/storm/scheduler/blacklist/strategies/DefaultBlacklistStrategy.java
```diff
@@ -85,9 +85,10 @@ public void prepare(Map<String, Object> conf) {
             }
         }
         Set<String> toRelease = releaseBlacklistWhenNeeded(cluster, new ArrayList<>(blacklist.keySet()));
+        cluster.setQuarantinedSupervisors(toRelease);
```
Ethanlm: I don't like this to be a side effect of this `getBlacklist` function. But I am still looking for a better way to better resolve this.

Ethanlm: I apologize for taking so long to update on this. I have been busy on other things. 

I couldn't find a better way to solve this problem without breaking backwards compatibility of `IBlacklistStrategy` interface.  But to make the code more clear, here are my suggestions:

1. don't use `quarantined`. The meaning of `quarantined` can be confusing, especially under the context here, we are not saying these nodes can't be used. They are just lowest priority. I think `graylist` is a better word in this context. `blacklist` means don't use it, `graylist` means don't use it unless have to.
2. Add comments about this side effect (updating graylist) in this `getBlacklist` function

We should refactor the blacklistScheduler code in the future if possible.

Ethanlm:
Looks good. Please update the title and squash the commits. Thanks

RuiLi8080:
@Ethanlm Thanks for reviewing. I have updated the pull-request. Waiting for Travis build.

PR #3128, opened 2019-09-10
STORM-3509 Improved RAS scheduling

dandsager1:
Don't scheduling topology if topology resource requirements exceeds total available on cluster.

### DIFF
--- a/storm-server/src/main/java/org/apache/storm/scheduler/resource/ResourceAwareScheduler.java
+++ b/storm-server/src/main/java/org/apache/storm/scheduler/resource/ResourceAwareScheduler.java
@@ -14,6 +14,7 @@ package org.apache.storm.scheduler.resource;
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -25,6 +26,7 @@ import java.util.concurrent.TimeoutException;
 import java.util.stream.Collectors;
 import org.apache.storm.Config;
 import org.apache.storm.DaemonConfig;
+import org.apache.storm.metric.StormMetricsRegistry;
 import org.apache.storm.scheduler.Cluster;
 import org.apache.storm.scheduler.IScheduler;
 import org.apache.storm.scheduler.SchedulerAssignment;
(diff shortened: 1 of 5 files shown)


storm-server/src/main/java/org/apache/storm/scheduler/resource/ResourceAwareScheduler.java
```diff
@@ -247,13 +234,169 @@ private void scheduleTopology(TopologyDetails td, Cluster cluster, final User to
...
+     * Ideally we would simply track NormalizedResources, but shared topology memory complicates things.
+     * Topologies with shared memory may use more than the SharedMemoryLowerBound, and topologyRequiredResources
+     * ignores shared memory.
+     *
+     * Resources are tracked in two ways:
+     * 1) AvailableResources. Track cluster available resources and required topology resources.
+     * 2) RemainingRequiredResources. Start with required topology resources, and deduct for partially scheduled and evicted topologies.
+     */
+    private class TopologySchedulingResources {
+        ResourceMetrics metrics;
```
Ethanlm: It is not very clear to me why we need this

storm-server/src/main/java/org/apache/storm/scheduler/resource/ResourceAwareScheduler.java
```diff
@@ -247,13 +234,169 @@ private void scheduleTopology(TopologyDetails td, Cluster cluster, final User to
...
+        double topologyRequiredNonSharedMemory;
+        double topologySharedMemoryLowerBound;
+
+        NormalizedResourceOffer remainingRequiredTopologyResources;
+        double remainingRequiredTopologyMemory;
+        double topologyScheduledMemory;
+
+        TopologySchedulingResources(Cluster cluster, TopologyDetails td) {
+            remainingResourcesSet = false;
+            metrics = new ResourceMetrics(new StormMetricsRegistry());
```
Ethanlm: I think we shouldn't create a new registry here.

storm-server/src/main/java/org/apache/storm/scheduler/resource/ResourceAwareScheduler.java
dandsager1: `metrics` variable is used in NormalizedResourceOffer.remove(). It tracks when resources become negative. During scheduling, negative resources indicate a scheduling failure; otherwise resources should not become negative. I don't want to use Cluster metrics because (1) it causes testSchedulingNegativeResources to fail and (2) erroneous logging of (scheduling only) negative resources.

Maybe pass null metrics and change code to handle handle that?

Ethanlm: Yes I agree to change the code to handle it. 

We can have another function 
```
 public boolean remove(NormalizedResourcesWithMemory other) {
    return remove(other, null);
}
```
And then change `public boolean remove(NormalizedResourcesWithMemory other, ResourceMetrics resourceMetrics) ` to handle the case where `resourceMetrics` is null.

dandsager1: removing `metrics` and making above changes

storm-server/src/main/java/org/apache/storm/scheduler/resource/ResourceAwareScheduler.java
```diff
@@ -247,13 +234,169 @@ private void scheduleTopology(TopologyDetails td, Cluster cluster, final User to
...
+
+            // available resources
+            clusterAvailableResources = cluster.getNonBlacklistedClusterAvailableResources(Collections.emptyList());
```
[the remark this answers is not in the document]
dandsager1: returns lower bound since blacklisted supervisors do not contribute

PR #3131, opened 2019-09-19
STORM-3510: Track overflow count per taskId for resending backpressur…

cjljohnson:
…e status

This is a fix to the WorkerState.transferLocalBatch logic for keeping track of how often to resend backpressure status updates as the overflow queue grows.

Previously a single counter was shared and overwritten for all executors and not retained between mini-batches.

This fix stores the overflow count in a map keyed by taskId and stored in BackPressureTracker.java so values from different executors don't overwrite each other and a status update is only sent once per 10000 tuples rather than every mini batch once the threshold has been reached.

### DIFF
--- a/storm-client/src/jvm/org/apache/storm/daemon/worker/BackPressureTracker.java
+++ b/storm-client/src/jvm/org/apache/storm/daemon/worker/BackPressureTracker.java
@@ -19,7 +19,6 @@
 package org.apache.storm.daemon.worker;
 
 import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -40,7 +39,6 @@ import org.slf4j.LoggerFactory;
 public class BackPressureTracker {
     static final Logger LOG = LoggerFactory.getLogger(BackPressureTracker.class);
     private final Map<Integer, BackpressureState> tasks;
-    private final Map<Integer, Integer> lastOverflowCount;
     private final String workerId;
 
     public BackPressureTracker(String workerId, Map<Integer, JCQueue> localTasksToQueues) {
@@ -49,11 +47,14 @@ public class BackPressureTracker {
             .collect(Collectors.toMap(
                 entry -> entry.getKey(),
                 entry -> new BackpressureState(entry.getValue())));
-        this.lastOverflowCount = new HashMap<>();
     }
 
-    private void recordNoBackPressure(Integer taskId) {
-        tasks.get(taskId).backpressure.set(false);
+    public BackpressureState getBackpressureState(Integer taskId) {
+        return tasks.get(taskId);
+    }
+
+    private void recordNoBackPressure(BackpressureState state) {
+        state.backpressure.set(false);
     }
 
     /**
@@ -63,8 +64,8 @@ public class BackPressureTracker {
      *
      * @return true if an update was recorded, false if taskId is already under BP
      */
-    public boolean recordBackPressure(Integer taskId) {
-        return tasks.get(taskId).backpressure.getAndSet(true) == false;
+    public boolean recordBackPressure(BackpressureState state) {
+        return state.backpressure.getAndSet(true) == false;
     }
 
     // returns true if there was a change in the BP situation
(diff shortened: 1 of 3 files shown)


storm-client/src/jvm/org/apache/storm/daemon/worker/BackPressureTracker.java
```diff
@@ -39,6 +40,7 @@
 public class BackPressureTracker {
     static final Logger LOG = LoggerFactory.getLogger(BackPressureTracker.class);
     private final Map<Integer, BackpressureState> tasks;
+    private final Map<Integer, Integer> lastOverflowCount;
```
srdo: I think we could optimize a bit here by having a single map of taskId -> BackPressureState, and just add the last overflow count to the BackPressureState object. That way, transferLocalBatch can look up the BackPressureState once rather than looking up in both maps here.

storm-client/src/jvm/org/apache/storm/daemon/worker/WorkerState.java
```diff
@@ -88,6 +88,7 @@
 
     private static final Logger LOG = LoggerFactory.getLogger(WorkerState.class);
     private static final long LOAD_REFRESH_INTERVAL_MS = 5000L;
+    private static final int RESEND_BACKPRESSURE_SIZE = 10000;
```
srdo: Am wondering if it makes sense to make this configurable?

storm-client/src/jvm/org/apache/storm/daemon/worker/BackPressureTracker.java
cjljohnson: I agree that it would be cleaner to store it in the BackpressureState object, but can you clarify what you mean by transferLocalBatch will only need to look it up once?  The BackpressureState object isn't exposed beyond the BackPressureTracker class, so you are paying for two map lookups either way. Unless i'm misunderstanding?

storm-client/src/jvm/org/apache/storm/daemon/worker/WorkerState.java
cjljohnson: I can make this change if someone more authoritative than myself concurs that it is a good idea.

storm-client/src/jvm/org/apache/storm/daemon/worker/BackPressureTracker.java
srdo: Yes, you are right, but I think we could make BackPressureStatus public, and change the BackPressureTracker API so we have a `getBackPressureStatus` method for getting the status, and then `recordBackPressure` and `setLastOverflowCount` and similar methods could just take the `BackPressureStatus` instance. Alternatively we could put `recordBackPressure` and similar methods in `BackPressureStatus`.

If we do either of these things, we can get away with only looking up the BackPressureStatus once.

storm-client/src/jvm/org/apache/storm/daemon/worker/WorkerState.java
srdo: I don't have a strong opinion, just wanted to know what you thought. If you don't see a need, we can always do it later if someone feels it is necessary.

storm-client/src/jvm/org/apache/storm/daemon/worker/WorkerState.java
```diff
@@ -414,18 +408,16 @@ public void refreshConnections() {
...
             Map<NodeInfo, IConnection> next = new HashMap<>(prev);
             for (NodeInfo nodeInfo : newConnections) {
-                next.put(nodeInfo,
-                         mqContext.connect(
-                             topologyId,
-                             //nodeHost is not null here, as newConnections is only non-empty if assignment was not null above.
-                             nodeHost.get(nodeInfo.get_node()),    // Host
-                             nodeInfo.get_port().iterator().next().intValue(),       // Port
-                             workerTransfer.getRemoteBackPressureStatus()));
+                next.put(nodeInfo, mqContext.connect(topologyId,
```
srdo: Please either move mqContext.connect back on its own line, or make a new variable so we can separate the connect and put calls. It's not obvious at a glance now which call the parameters below belongs to.

srdo:
Looks great. +1 assuming tests pass.

storm-client/src/jvm/org/apache/storm/daemon/worker/WorkerState.java
cjljohnson: Looks like i accidentally hit the auto-style shortcut at some point. I'll revert all the style changes that aren't part of the changes.

storm-client/src/jvm/org/apache/storm/daemon/worker/WorkerState.java
srdo: Sounds good. Thanks for considering it.

PR #3132, opened 2019-09-26
STORM-3186: Customizable configuration for metric reporting interval

jainrish:


agresch:
This looks fine, except I'm not clear why this is only modified for ConsolePreparableReporter.

storm-server/src/main/java/org/apache/storm/DaemonConfig.java
```diff
@@ -74,6 +74,12 @@
     @IsString
     public static final String STORM_DAEMON_METRICS_REPORTER_PLUGIN_DOMAIN = "storm.daemon.metrics.reporter.plugin.domain";
 
+    /**
+     * We report the metrics with this interval period.
+     */
+    @IsString
+    public static final String STORM_DAEMON_METRICS_REPORTER_INTERVAL_PERIOD = "storm.daemon.metrics.reporter.interval.period";
```
srdo: Please rename so we include the time unit, e.g. "storm.daemon.metrics.reporter.interval.secs"

jainrish:
> This looks fine, except I'm not clear why this is only modified for ConsolePreparableReporter.

@agresch I have modified it for CsvPreparableReporter.java as well. These are the only two classes which had hardcoded reporting for 10 seconds.

storm-server/src/main/java/org/apache/storm/DaemonConfig.java
jainrish: sure, will do

agresch:
Thanks.  I also see ScheduledStormReporter using REPORT_PERIOD and REPORT_PERIOD_UNITS for setting this.  Just wondering if we should switch that as well for consistency?  I like the clarity of this change better.  @srdo have any thoughts on this?

srdo:
I think the ScheduledStormReporter classes are for the metrics reporters for the workers (the parts topologies can hook into), while this is for the daemons. I think having separate parameters makes sense, as the worker metrics config is topology level, while this new setting is at the cluster level.

Don't have an opinion on whether we should add a parameter for setting the time unit as well, do you think it would be useful?

As an aside, we really should rename StormMetricsRegistry. We have StormMetricRegistry for the workers, and StormMetricsRegistry for the daemons. It's easy to mix them up.

jainrish:
> I think the ScheduledStormReporter classes are for the metrics reporters for the workers (the parts topologies can hook into), while this is for the daemons. I think having separate parameters makes sense, as the worker metrics config is topology level, while this new setting is at the cluster level.
> 
> Don't have an opinion on whether we should add a parameter for setting the time unit as well, do you think it would be useful?
> 
> As an aside, we really should rename StormMetricsRegistry. We have StormMetricRegistry for the workers, and StormMetricsRegistry for the daemons. It's easy to mix them up.

@srdo Let me know if its required to have a configuration for time unit as well. I would like to work on it.

storm-server/src/main/java/org/apache/storm/DaemonConfig.java
```diff
@@ -74,6 +74,12 @@
     @IsString
     public static final String STORM_DAEMON_METRICS_REPORTER_PLUGIN_DOMAIN = "storm.daemon.metrics.reporter.plugin.domain";
 
+    /**
+     * We report the metrics with this interval period.
+     */
+    @IsString
+    public static final String STORM_DAEMON_METRICS_REPORTER_INTERVAL_PERIOD = "storm.daemon.metrics.reporter.interval.secs";
```
srdo: Please also rename the variable

srdo:
@jainrish If you think a time unit would be useful, feel free to add it. Up to you whether to do it in this PR or in a later issue.

storm-server/src/main/java/org/apache/storm/DaemonConfig.java
jainrish: done!!

jainrish:
> @jainrish If you think a configuration for time unit would be useful, feel free to add it. Up to you whether to do it in this PR or in a later issue.

@srdo I will prefer to do it in a separate issue. Let me know if I should raise the jira.

srdo:
@jainrish Feel free.

srdo:
Looks good, just a few minor nits left:

Please add the default setting to https://github.com/apache/storm/blob/master/conf/defaults.yaml. That way it's visible to users what the default is. Then you can remove the default value from the Java classes.

storm-server/src/main/java/org/apache/storm/daemon/metrics/reporters/ConsolePreparableReporter.java
```diff
@@ -46,13 +50,14 @@ public void prepare(MetricRegistry metricsRegistry, Map<String, Object> topoConf
             builder.convertDurationsTo(durationUnit);
         }
         reporter = builder.build();
+        reportingIntervalPeriod = ObjectReader.getInt(topoConf.get(DaemonConfig.STORM_DAEMON_METRICS_REPORTER_INTERVAL_SECS), 10);
```
srdo: I'd like to see topoConf renamed here and in the other preparable reporters. The name is misleading, it isn't the topology config, but the daemon config. Renaming to e.g. "daemonConf" would make this clearer.

Ethanlm: Agreed

storm-server/src/main/java/org/apache/storm/daemon/metrics/reporters/ConsolePreparableReporter.java
```diff
@@ -17,13 +17,17 @@
...
+import org.apache.storm.DaemonConfig;
 import org.apache.storm.daemon.metrics.ClientMetricsUtils;
+import org.apache.storm.utils.ObjectReader;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class ConsolePreparableReporter implements PreparableReporter {
     private static final Logger LOG = LoggerFactory.getLogger(ConsolePreparableReporter.class);
     ConsoleReporter reporter = null;
+    Integer reportingIntervalPeriod = null;
```
Ethanlm: It would be better to change `reportingIntervalPeriod` to `reportingIntervalSecs`

storm-server/src/main/java/org/apache/storm/daemon/metrics/reporters/CsvPreparableReporter.java
```diff
@@ -18,14 +18,18 @@
...
 import org.apache.storm.daemon.metrics.ClientMetricsUtils;
 import org.apache.storm.daemon.metrics.MetricsUtils;
+import org.apache.storm.utils.ObjectReader;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class CsvPreparableReporter implements PreparableReporter {
     private static final Logger LOG = LoggerFactory.getLogger(CsvPreparableReporter.class);
     CsvReporter reporter = null;
+    Integer reportingIntervalPeriod = null;
```
Ethanlm: here too: It would be better to change reportingIntervalPeriod to reportingIntervalSecs

PR #3138, opened 2019-10-10
STORM-3520: Fixes drpc-client function argument validation in Storm CLI

govind-menon:


Ethanlm:
We need this plus other CLI fixes in 2.1.x-branch too.

PR #3140, opened 2019-10-10
STORM-3188: Remove try-catch block from getAndResetWorkerHeartbeats

daisyxchen:


storm-server/src/main/java/org/apache/storm/daemon/supervisor/timer/ReportWorkerHeartbeats.java
```diff
@@ -50,13 +50,8 @@ public void run() {
...
         Map<String, LSWorkerHeartbeat> localHeartbeats;
-        try {
-            localHeartbeats = SupervisorUtils.readWorkerHeartbeats(this.conf);
-            return getSupervisorWorkerHeartbeatsFromLocal(localHeartbeats);
-        } catch (Exception e) {
-            LOG.error("Read local worker heartbeats error, skipping heartbeats for this round, msg:{}", e.getMessage());
-            return null;
-        }
+        localHeartbeats = SupervisorUtils.readWorkerHeartbeats(this.conf);
+        return getSupervisorWorkerHeartbeatsFromLocal(localHeartbeats);
```
kishorvpatil: The method is called in from run method https://github.com/apache/storm/blob/master/storm-server/src/main/java/org/apache/storm/daemon/supervisor/timer/ReportWorkerHeartbeats.java#L45-L49
This try catch loop ensures that thread does not die.
Not sure what this PR is trying to accomplish. Please elaborate.

daisyxchen: According to the request in ticket https://issues.apache.org/jira/browse/STORM-3188, is the try catch necessary?

srdo: I think it is. Maybe we should close the issue as won't fix.

PR #3141, opened 2019-10-11
STORM-3518: Add support to (un)favored nodes with range format

daisyxchen:
Hostnames like `host[1-2].yahoo.com`, `hostA3[1-4].yahoo.com` will be supported.

### DIFF
--- a/storm-server/src/test/java/org/apache/storm/scheduler/resource/strategies/scheduling/TestDefaultResourceAwareStrategy.java
+++ b/storm-server/src/test/java/org/apache/storm/scheduler/resource/strategies/scheduling/TestDefaultResourceAwareStrategy.java
@@ -628,7 +628,7 @@ public class TestDefaultResourceAwareStrategy {
         t1Conf.putAll(config);
         final List<String> t1FavoredHostNames = Arrays.asList("host-41", "host-42", "host-43");
         t1Conf.put(Config.TOPOLOGY_SCHEDULER_FAVORED_NODES, t1FavoredHostNames);
-        final List<String> t1UnfavoredHostIds = Arrays.asList("host-1", "host-2", "host-3");
+        final List<String> t1UnfavoredHostIds = Arrays.asList("host-1", "host-2", "host-3", "host[11-13]", "host[2-1]");
         t1Conf.put(Config.TOPOLOGY_SCHEDULER_UNFAVORED_NODES, t1UnfavoredHostIds);
         //generate topologies
         TopologyDetails topo1 = genTopology("topo-1", t1Conf, 8, 0, 2, 0, CURRENT_TIME - 2, 10, "user");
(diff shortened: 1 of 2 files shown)


storm-server/src/test/java/org/apache/storm/scheduler/resource/strategies/scheduling/TestDefaultResourceAwareStrategy.java
```diff
@@ -628,7 +628,7 @@ public void testMultipleRacksWithFavoritism() {
         t1Conf.putAll(config);
         final List<String> t1FavoredHostNames = Arrays.asList("host-41", "host-42", "host-43");
         t1Conf.put(Config.TOPOLOGY_SCHEDULER_FAVORED_NODES, t1FavoredHostNames);
-        final List<String> t1UnfavoredHostIds = Arrays.asList("host-1", "host-2", "host-3");
+        final List<String> t1UnfavoredHostIds = Arrays.asList("host-1", "host-2", "host-3", "host[11-13]", "host[2-1]");
```
govind-menon: Could we add an assert to check the correct host ids were unfavored?

daisyxchen: I ran the local test but it is weird that none of the host ids in topo1 favored list show up in the assigned cluster. 
The original test code didn't test favored or unfavored hostIds at all.
```
        for (WorkerSlot ws : assignment.getSlotToExecutors().keySet()) {
            String hostName = rs.idToNode(ws.getNodeId()).getHostname();
            String rackId = resolvedSuperVisors.get(hostName);
            Assert.assertTrue(ws + " is neither on a favored node " + t1FavoredHostNames + " nor the highest priority rack (rack-0)",
                t1FavoredHostNames.contains(hostName) || "rack-0".equals(rackId));
            Assert.assertFalse(ws + " is a part of an unfavored node " + t1UnfavoredHostIds,
                t1UnfavoredHostIds.contains(hostName));
        }
```
The assigned hostIds are always host-001, host-000, host-002, host-003. Is that expected in local unit test?

@revans2 , since you wrote the test code, do you have any insights about the issue?

Ethanlm:
Thanks for working on this. 

The idea of unfavored/favored nodes is that users can specify it in their topologies so when nimbus schedules the topology, it will try to scheduled tasks to favored nodes first, and try to avoid scheduling on unfavored nodes. 

There are multiple places where unfavored/favored nodes are being used. 

1. https://github.com/apache/storm/blob/master/storm-server/src/main/java/org/apache/storm/scheduler/resource/strategies/scheduling/GenericResourceAwareStrategy.java#L102-L103

2. https://github.com/apache/storm/blob/master/storm-server/src/main/java/org/apache/storm/scheduler/resource/strategies/scheduling/ConstraintSolverStrategy.java#L257-L258

3. https://github.com/apache/storm/blob/master/storm-server/src/main/java/org/apache/storm/scheduler/resource/strategies/scheduling/DefaultResourceAwareStrategy.java#L63-L64

We can have common methods in Config.java to get favored/unfavored nodes. And replace above three.

kishorvpatil:
@daisyxchen would you like to revisit this?

PR #3144, opened 2019-10-15
STORM-3524 create complete blobstore path in case parent directory ge…

agresch:
…ts deleted

Based on the exception I saw on a supervisor, I think a race condition existed between a cleanup of the localizer for an older topology and a new one starting where a parent directory was deleted right before a download.

Files.createDirectory() then throws an exception because the parent directory does not exist and can never succeed.

Files.createDirectories() should guarantee the directory gets created.  Also added logging for removing the user to better debug in the future.

### DIFF
--- a/storm-server/src/main/java/org/apache/storm/localizer/LocalizedResource.java
+++ b/storm-server/src/main/java/org/apache/storm/localizer/LocalizedResource.java
@@ -144,6 +144,7 @@ public class LocalizedResource extends LocallyCachedBlob {
 
     static void completelyRemoveUnusedUser(Path localBaseDir, String user) throws IOException {
         Path userFileCacheDir = getLocalUserFileCacheDir(localBaseDir, user);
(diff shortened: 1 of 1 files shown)


storm-server/src/main/java/org/apache/storm/localizer/LocalizedResource.java
```diff
@@ -144,6 +144,7 @@ private static Path constructBlobWithVersionFileName(Path baseDir, String key, l
 
     static void completelyRemoveUnusedUser(Path localBaseDir, String user) throws IOException {
         Path userFileCacheDir = getLocalUserFileCacheDir(localBaseDir, user);
+        LOG.info("completelyRemoveUnusedUser {} for directory {}", user, userFileCacheDir);
```
Ethanlm: This method is about deleting `baseDir/supervisor/usercache/user1` directory. It might be better to log that. `userFileCacheDir` is just part of it.

agresch:
@Ethanlm - made the requested changes.

PR #3146, opened 2019-10-17
STORM-3521: Adds ugly workaround to deal with topology main arguments…

govind-menon:
… with non unix compliant flags


bin/storm.py
```diff
@@ -1464,18 +1464,32 @@ def print_server_classpath(args):
...
     exec_storm_class(
         "org.apache.storm.command.Monitor", storm_config_opts=args.storm_config_opts,
-        args=remove_common_options(sys.argv[2:]),
+        main_class_args=remove_common_options(sys.argv[2:]),
         jvmtype="-client",
         extrajars=[USER_CONF_DIR, STORM_BIN_DIR])
 
+def is_not_conflicting_jar_argument(raw_arg):
+    # TODO refactor underlying Topology main classes to avoid this ugly hack
+    conflicting_flags = ["-hdfsConf"]
```
Ethanlm: Why is it specific to `-hdfsConf`?

bin/storm.py
```diff
@@ -1401,7 +1401,7 @@ def drpc_client(args):
 
     exec_storm_class(
         "org.apache.storm.command.BasicDrpcClient",
-        args=remove_common_options(sys.argv[2:]), storm_config_opts=args.storm_config_opts,
+        main_class_args=remove_common_options(sys.argv[2:]), storm_config_opts=args.storm_config_opts,
```
Ethanlm: why do we need to change this

storm-client/test/py/test_storm_cli.py
```diff
@@ -67,6 +67,25 @@ def test_jar_command(self):
...
+        ], self.mock_execvp, mock.call(
+            self.java_cmd, [
+                self.java_cmd, '-client','-Ddaemon.name=', '-Dstorm.options=',
+                '-Dstorm.home=' + self.storm_dir + '', '-Dstorm.log.dir=' + self.storm_dir + "/logs",
+                '-Djava.library.path=', '-Dstorm.conf.file=', '-cp',
+                '' + self.storm_dir + '/*:' + self.storm_dir + '/lib:' + self.storm_dir
+                + '/extlib:/path/to/jar.jar:' + self.storm_dir + '/conf:' + self.storm_dir + '/bin:',
+                '-Dstorm.jar=/path/to/jar.jar', '-Dstorm.dependency.jars=', '-Dstorm.dependency.artifacts={}',
+                'some.Topology.Class', '-name', 'run-topology', 'randomArgument', '-randomFlag', 'randomFlagValue',
+                '-rotateSize', '0.0001', 'dfs.namenode.kerberos.principal.pattern=hdfs/*.EV..COM', '-hdfsConf'
```
Ethanlm: Why is this correct?
`'dfs.namenode.kerberos.principal.pattern=hdfs/*.EV..COM', '-hdfsConf'`? Should `-hdfsConf` appear before `'dfs.namenode.kerberos.principal.pattern=hdfs/*.EV..COM'`

bin/storm.py
govind-menon: This is just to make the name more descriptive

bin/storm.py
govind-menon: That's the ugly part - the conflict is that the way the HdfsTopology specifies the argument is non-unix compliant. And there's no consistent way to identify all such possible flags so I'm only filtering this one for now and add more later

storm-client/test/py/test_storm_cli.py
govind-menon: Good catch - the filtered flags are re-added at the end. The argument is of the form -hdfsConf=someValue so the order shouldn't matter

bin/storm.py
Ethanlm: Any idea why flags starting with `-h` will fail, e.g. `-htt`?
```
storm jar: error: argument -h/--help: ignored explicit argument 'tt'
```

govind-menon: Because -htt is equivalent to -h -tt. The correct way to specify it is --htt. The HdfsTopology acceps -htt (which is non-Unix compliant it should be single dashes for single characters or double dashes for multiple characters)

Ethanlm: I think I finally find a work around: set `allow_abbrev=False`.

govind-menon: That doesn't solve the underlying issue - there are single flag arguments that we had previously in the CLI. Setting that will make it backwards incompatible (like the -c)

Ethanlm: I had an offline discussion with @govind-menon and verified this would work. 
If `allow_abbrev` is true, then `-htt` will be parsed as `-h tt`. And that's why it failed. 

The only problem is `allow_abbrev` is not available in python2.

PR #3147, opened 2019-10-18
[STORM-3526]: Always override Login Config from System Property

kishorvpatil:


agresch:
1) Pacemaker calls ConfigUtils.overrideLoginConfigWithSystemProperty(ConfigUtils.readStormConfig()), basically duplicating the code.  Can you fix this?

2) Worker calls ConfigUtils.overrideLoginConfigWithSystemProperty(ConfigUtils.readSupervisorStormConf(conf, topologyId)).  Seems like we should add this to readSupervisorStormConf() as well?

Ethanlm:
I understand the general idea of it. But curious what are we trying to fix here?  Also there is a similar function https://github.com/apache/storm/blob/master/storm-client/src/jvm/org/apache/storm/utils/Utils.java#L299 . It can be confusing if `Utils.readStormConfig` and  `ConfigUitls. readStormConfig` return different values

kishorvpatil:
@Ethanlm , The _Utils.readStormConfig_ is used by _ConfigUtils.readStormConfig_ in variably. The _Utils.readStormConfig_ is used at many other places, while _ConfigUtils.readStormConfig_  is used by daemons. Many a times, daemons should be able to read system property passed as argument to JVM  to override default config file setting for login config. The daemons need to stick to single configuration defined in the yaml file. You may have more than one daemon running using their individual jvm arguments ( system. properties)  to customize login config.
Without this change, we can not override the login config using jvm argument.

Ethanlm:
The one in Worker.java mentioned by @agresch  should be fixed too

kishorvpatil:
Closing this as #3176  is merged.

PR #3153, opened 2019-10-24
STORM-3523 revert handling of FNF.  Prevent AsyncLocalizer deadlock.

agresch:
Ended up finding that we were running the AsyncLocalizer with 3 threads, which ended up causing a deadlock.  This was the root cause of the FileNotFoundException.

### DIFF
--- a/conf/defaults.yaml
+++ b/conf/defaults.yaml
@@ -135,7 +135,7 @@ transactional.zookeeper.port: null
 
 ## blobstore configs
 supervisor.blobstore.class: "org.apache.storm.blobstore.NimbusBlobStore"
-supervisor.blobstore.download.thread.count: 5
+supervisor.blobstore.download.thread.count: 3
 supervisor.blobstore.download.max_retries: 3
 supervisor.localizer.cache.target.size.mb: 10240
 supervisor.localizer.cleanup.interval.ms: 30000
--- a/conf/defaults.yaml
+++ b/conf/defaults.yaml
@@ -135,7 +135,7 @@ transactional.zookeeper.port: null
 
 ## blobstore configs
 supervisor.blobstore.class: "org.apache.storm.blobstore.NimbusBlobStore"
-supervisor.blobstore.download.thread.count: 3
+supervisor.blobstore.download.thread.count: 5
 supervisor.blobstore.download.max_retries: 3
 supervisor.localizer.cache.target.size.mb: 10240
 supervisor.localizer.cleanup.interval.ms: 30000
--- a/storm-server/src/main/java/org/apache/storm/localizer/AsyncLocalizer.java
+++ b/storm-server/src/main/java/org/apache/storm/localizer/AsyncLocalizer.java
@@ -21,7 +21,6 @@ package org.apache.storm.localizer;
 import com.codahale.metrics.Meter;
 import com.codahale.metrics.Timer;
 import java.io.File;
-import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.nio.file.DirectoryStream;
 import java.nio.file.Files;
@@ -450,15 +449,7 @@ public class AsyncLocalizer implements AutoCloseable {
             topoConfBlob.removeReference(pna);
         }
 
-        List<LocalResource> localResources;
-        try {
-            localResources = getLocalResources(pna);
-        } catch (FileNotFoundException e) {
-            LOG.warn("Local resources for {} no longer available", pna, e);
-            return;
-        }
-
-        for (LocalResource lr : localResources) {
+        for (LocalResource lr : getLocalResources(pna)) {
             try {
                 removeBlobReference(lr.getBlobName(), pna, lr.shouldUncompress());
             } catch (Exception e) {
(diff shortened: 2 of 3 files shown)


agresch:
Added a commit that should prevent deadlocks.  The problem is that as workers launch (also the update/cleanup tasks), they use a thread and then block on a compute that waits on a download to complete.  If these tasks use all of the threads in the executor service, no threads exist to do a download, so they will deadlock. 

My solution here is to just guarantee we have at least one thread for tasks and downloads.  We should never have blockage then.

Updated the thread count to have similar performance to the previous default.

storm-server/src/main/java/org/apache/storm/localizer/AsyncLocalizer.java
```diff
@@ -213,7 +214,7 @@ private LocalizedResource getUserFile(String user, String key) {
             blobPending.compute(topologyId, (tid, old) -> {
                 CompletableFuture<Void> ret = old;
                 if (ret == null) {
-                    ret = CompletableFuture.supplyAsync(new DownloadBlobs(pna, cb), execService);
+                    ret = CompletableFuture.supplyAsync(new DownloadBlobs(pna, cb), taskExecService);
```
Ethanlm: Should we use `downloadExecService` here

conf/defaults.yaml
```diff
@@ -135,7 +135,7 @@ transactional.zookeeper.port: null
 
 ## blobstore configs
 supervisor.blobstore.class: "org.apache.storm.blobstore.NimbusBlobStore"
-supervisor.blobstore.download.thread.count: 5
+supervisor.blobstore.download.thread.count: 3
```
Ethanlm: I am not sure if we want to change default value because users might rely on it

storm-server/src/main/java/org/apache/storm/localizer/AsyncLocalizer.java
agresch: No, I don't believe so.  The "DownloadBlobs" is basically creating an object that acts as a task that then waits for a download to finish through the getBlobs() call.  getBlobs() calls and waits on downloadOrUpdate(), which runs in the download executor.

Here is a callstack for this path from one of the jstacks I took on a deadlocked supervisor:

Thread 70026: (state = BLOCKED)
 - sun.misc.Unsafe.park(boolean, long) @bci=0 (Compiled frame; information may be imprecise)
 - java.util.concurrent.locks.LockSupport.park(java.lang.Object) @bci=14, line=175 (Interpreted frame)
 - java.util.concurrent.CompletableFuture$Signaller.block() @bci=19, line=1693 (Interpreted frame)
 - java.util.concurrent.ForkJoinPool.managedBlock(java.util.concurrent.ForkJoinPool$ManagedBlocker) @bci=59, line=3313 (Interpreted frame)
 - java.util.concurrent.CompletableFuture.waitingGet(boolean) @bci=129, line=1729 (Interpreted frame)
 - java.util.concurrent.CompletableFuture.get() @bci=11, line=1895 (Interpreted frame)
 - org.apache.storm.localizer.AsyncLocalizer.getBlobs(java.util.List, org.apache.storm.localizer.PortAndAssignment, org.apache.storm.localizer.BlobChangingCallback) @bci=226, line=557 (Interpreted frame)
 - org.apache.storm.localizer.AsyncLocalizer.addReferencesToBlobs(org.apache.storm.localizer.PortAndAssignment, org.apache.storm.localizer.BlobChangingCallback) @bci=19, line=383 (Interpreted frame)
 - org.apache.storm.localizer.AsyncLocalizer.lambda$null$7(org.apache.storm.localizer.PortAndAssignment, org.apache.storm.localizer.BlobChangingCallback, java.lang.String, java.lang.String, java.util.concurrent.CompletableFuture) @bci=34, line=218 (Interpreted frame)
 - org.apache.storm.localizer.AsyncLocalizer$$Lambda$75.apply(java.lang.Object, java.lang.Object) @bci=24 (Interpreted frame)
 - java.util.concurrent.ConcurrentHashMap.compute(java.lang.Object, java.util.function.BiFunction) @bci=298, line=1877 (Interpreted frame)
 - org.apache.storm.localizer.AsyncLocalizer.lambda$requestDownloadTopologyBlobs$8(java.lang.String, org.apache.storm.localizer.PortAndAssignment, org.apache.storm.localizer.BlobChangingCallback, java.lang.Void) @bci=14, line=212 (Interpreted frame)
 - org.apache.storm.localizer.AsyncLocalizer$$Lambda$73.apply(java.lang.Object) @bci=20 (Interpreted frame)
 - java.util.concurrent.CompletableFuture.uniCompose(java.util.concurrent.CompletableFuture, java.util.function.Function, java.util.concurrent.CompletableFuture$UniCompose) @bci=84, line=952 (Interpreted frame)
 - java.util.concurrent.CompletableFuture$UniCompose.tryFire(int) @bci=29, line=926 (Interpreted frame)
 - java.util.concurrent.CompletableFuture$Completion.exec() @bci=2, line=443 (Interpreted frame)
 - java.util.concurrent.ForkJoinTask.doExec() @bci=10, line=289 (Interpreted frame)
 - java.util.concurrent.ForkJoinPool$WorkQueue.runTask(java.util.concurrent.ForkJoinTask) @bci=21, line=1056 (Interpreted frame)
 - java.util.concurrent.ForkJoinPool.runWorker(java.util.concurrent.ForkJoinPool$WorkQueue) @bci=35, line=1692 (Interpreted frame)
 - java.util.concurrent.ForkJoinWorkerThread.run() @bci=24, line=157 (Interpreted frame)

agresch:
@Ethanlm - restored default thread count.

storm-server/src/main/java/org/apache/storm/localizer/AsyncLocalizer.java
Ethanlm: Your explanation makes sense. 

The stack trace seems to irrelevant though.

PR #3154, opened 2019-10-25
MINOR: update RELEASING doc to include more details

Ethanlm:
Add some details to make things more clear

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -5,6 +5,13 @@ the Storm release process.
 
 ---
 
+# Release Policy
+
+Apache Storm follows the basic idea of [Semantic Versioning](https://semver.org/). Given a version number MAJOR.MINOR.PATCH, increment the:
+ 1. MAJOR version when you make incompatible API changes,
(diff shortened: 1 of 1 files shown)


RELEASING.md
```diff
@@ -56,3 +83,5 @@ Ensure you have a signed GPG key, and that the GPG key is listed in the Storm KE
 1. Go to http://repository.apache.org and drop the staging repository.
 
 2. Delete the staged distribution files from https://dist.apache.org/repos/dist/dev/storm/
+
+3. Delete the git tag.
```
srdo: I don't think we should delete the git tag. Keeping it around makes it easier for people to find the source in git later, without having to go download the source zip (which would then not have any history). 

I think keeping it around is/has been standard procedure on this and other Apache projects

Ethanlm: This is about deleting the git tag when release candidate fails.  If it succeeds, we should keep the git tag.

srdo: Ah, whoops.

PR #3155, opened 2019-10-27
STORM-3066: Implement support for using list elements in properties in FluxParser

efgpinto:
This MR adds support for accessing elements of list properties in configurations of FluxParser.
The reference of list items can be done as:
```yaml
  # test variable substitution for list element
  list.element.property.target: ${a.list.property[0]}
```
Additionally, refactors the code a bit more efficient and lighter on memory.

### DIFF
--- a/flux/flux-core/src/main/java/org/apache/storm/flux/parser/FluxParser.java
+++ b/flux/flux-core/src/main/java/org/apache/storm/flux/parser/FluxParser.java
@@ -18,12 +18,17 @@
 
 package org.apache.storm.flux.parser;
 
-import java.io.ByteArrayOutputStream;
+import java.io.BufferedReader;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.InputStreamReader;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Properties;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
 import org.apache.storm.flux.model.BoltDef;
 import org.apache.storm.flux.model.IncludeDef;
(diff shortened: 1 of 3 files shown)


flux/flux-core/src/main/java/org/apache/storm/flux/parser/FluxParser.java
```diff
@@ -139,37 +149,49 @@ public static Properties parseProperties(String propertiesFile, boolean resource
...
-            for (Object key : properties.keySet()) {
-                str = str.replace("${" + key + "}", properties.getProperty((String)key));
+        reader.lines().forEach(line -> {
+            Matcher m = propertyPattern.matcher(line);
+            if (m.find()) {
+                String target = "${" + m.group("var") + "}";
+                Optional<String> replacement = getPropertyReplacement(properties, m, envSubstitution);
+                replacement.ifPresent(propValue -> bos.append(line.replace(target, propValue)));
+                if (!replacement.isPresent()) {
+                    LOG.warn("Could not found replacement for property: " + m.group("var"));
```
srdo: Nit: "Could not find"

flux/flux-core/src/main/java/org/apache/storm/flux/parser/FluxParser.java
```diff
@@ -139,37 +149,49 @@ public static Properties parseProperties(String propertiesFile, boolean resource
...
+        BufferedReader reader = new BufferedReader(new InputStreamReader(in));
 
-        // TODO substitution implementation is not exactly efficient or kind to memory...
-        String str = bos.toString();
-        // properties file substitution
-        if (properties != null) {
-            LOG.info("Performing property substitution.");
-            for (Object key : properties.keySet()) {
-                str = str.replace("${" + key + "}", properties.getProperty((String)key));
+        reader.lines().forEach(line -> {
```
srdo: This is minor, but I'm wondering if this could be made slightly easier to read using `.map` and `.collect` along with the `joining` Collector https://docs.oracle.com/javase/8/docs/api/java/util/stream/Collectors.html#joining--

That way you wouldn't have to add the new line to the StringBuilder manually, and there'd be less to do in the `else`-cases where things aren't matched.

flux/flux-core/src/main/java/org/apache/storm/flux/parser/FluxParser.java
```diff
@@ -139,37 +149,49 @@ public static Properties parseProperties(String propertiesFile, boolean resource
...
-    private static TopologyDef loadYaml(Yaml yaml, InputStream in, Properties properties, boolean envSubstitution) throws IOException {
-        ByteArrayOutputStream bos = new ByteArrayOutputStream();
+    private static TopologyDef loadYaml(Yaml yaml, InputStream in, Properties properties, boolean envSubstitution) {
         LOG.info("loading YAML from input stream...");
-        int b = -1;
-        while ((b = in.read()) != -1) {
-            bos.write(b);
-        }
+        StringBuilder bos = new StringBuilder();
+        BufferedReader reader = new BufferedReader(new InputStreamReader(in));
```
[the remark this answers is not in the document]
efgpinto: Thanks, updated using try-with-resources

flux/flux-core/src/main/java/org/apache/storm/flux/parser/FluxParser.java
efgpinto: Thanks for the suggestion. Better now I think. Can you take a look? 🙇

srdo:
+1

PR #3156, opened 2019-10-27
STORM-3211: Fix NPE in WindowedBoltExecutor on getComponentConfiguration

efgpinto:
Avoid NPE in WindowedBoltExecutor when the wrapped bolt returns a null configuration.


srdo:
+1

PR #3160, opened 2019-10-30
STORM-3530 Improve scheduling failure messages

dandsager1:
When topology fails to schedule due to insufficient resources, add to topology status message:
1. Specific scheduling failure text
2. Include available resources along with required

### DIFF
--- a/storm-server/src/main/java/org/apache/storm/scheduler/resource/strategies/scheduling/BaseResourceAwareStrategy.java
+++ b/storm-server/src/main/java/org/apache/storm/scheduler/resource/strategies/scheduling/BaseResourceAwareStrategy.java
@@ -60,6 +60,15 @@ public abstract class BaseResourceAwareStrategy implements IStrategy {
     private final Map<String, List<RasNode>> hostnameToNodes = new HashMap<>();
     private final Map<String, List<RasNode>> rackIdToNodes = new HashMap<>();
     protected RasNodes nodes;
+    private String schedulingFailText;
+
+    public String getSchedulingFailText() {
+        return schedulingFailText;
+    }
+
+    public void clearSchedulingFailText() {
+        schedulingFailText = "";
+    }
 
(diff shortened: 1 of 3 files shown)


storm-server/src/main/java/org/apache/storm/scheduler/resource/strategies/scheduling/BaseResourceAwareStrategy.java
```diff
@@ -126,7 +135,9 @@ protected boolean scheduleExecutor(
         } else {
             String comp = td.getExecutorToComponent().get(exec);
             NormalizedResourceRequest requestedResources = td.getTotalResources(exec);
-            LOG.error("Not Enough Resources to schedule Task {} - {} {}", exec, comp, requestedResources);
+            schedulingFailText = "Not Enough Resources to schedule Task "
+                    + exec + " - " + comp + " " + requestedResources;
+            LOG.error(schedulingFailText);
```
agresch: Can we make this a warn message?  It's always bothered me that this reports as an ERROR.  There's no error in the code flow, the cluster just does not have enough resources.

dandsager1: Changed to warn

storm-server/src/main/java/org/apache/storm/scheduler/resource/strategies/scheduling/IStrategy.java
```diff
@@ -41,4 +41,10 @@
      *     successful.
      */
     SchedulingResult schedule(Cluster schedulingState, TopologyDetails td);
+
+    // Get topology scheduling failure text.
+    String getSchedulingFailText();
```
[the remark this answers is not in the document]
dandsager1: Backed out change to IStrategy

PR #3161, opened 2019-10-30
MINOR: create a checklist for testing a new Storm release

Ethanlm:
This creates a checklist so new devs have a better idea on reviewing a release

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -1,7 +1,6 @@
-# Committer documentation
+# Release
 
-This document summarizes information relevant to Storm committers.  It includes information about
-the Storm release process.
+This document includes information about the Storm release process.
 
 ---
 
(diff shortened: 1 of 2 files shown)


RELEASING.md
```diff
@@ -85,3 +84,24 @@ gpg --print-md SHA512 RELEASE_NOTES.html > RELEASE_NOTES.html.sha512
...
 3. Delete the git tag.
+
+# How to vote on a release candidate
+
+We encourage everyone to review and vote on a release candidate to make an Apache Storm release more reliable and trustworthy.
+
+Below is a checklist that one could do to review a release candidate. 
+Please note this list is not exhaustive and only includes some of the common steps. Feel free to add your own tests.
+
+1. Verify files such as *.asc, *.sha512;
```
srdo: I think we should link to the utility script in dev-tools/rc for this. I didn't know about it until I happened across it recently, would be good to raise visibility.

RELEASING.md
```diff
@@ -85,3 +84,24 @@ gpg --print-md SHA512 RELEASE_NOTES.html > RELEASE_NOTES.html.sha512
 2. Delete the staged distribution files from https://dist.apache.org/repos/dist/dev/storm/
 
 3. Delete the git tag.
+
+# How to vote on a release candidate
```
srdo: I'm wondering if this belongs at the top of the file? The audience for it is probably broader than the audience for how to author a release.

Ethanlm: make sense

RELEASING.md
Ethanlm: I think the original order is fine. We need to have release candidate then we can review and vote. It's more natural to me

srdo: I think your comments got switched somehow.

Ok, it's fine. I'd just like this guidance to be visible to people.

srdo:
+1

PR #3162, opened 2019-11-07
Add thrift.supervisor.port to documentation

cjljohnson:
I spent a whole lot of time trying to figure out why my workers wouldn't start, and it ended up being that i had to open port 6628 (thrift.supervisor.port) on the workers.

This caused me a lot of headache as it is omitted from the documentation.

Can someone confirm that i have the client for this port correct?  I put Nimbus but it might be Zookeeper or other workers, i'm not sure exactly how it works


srdo:
Yes, as I recall it is Nimbus that talks to the supervisor port.

+1, thanks for updating the docs. I'm a little unsure why we have the SECURITY file in two locations, but can figure that out separately.

PR #3164, opened 2019-11-10
STORM-3529: Catch and Log Kafka RetriableException

OliverMD:


external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/metrics/KafkaOffsetMetric.java
```diff
@@ -76,8 +77,17 @@ public Object getValueAndReset() {
...
-        Map<TopicPartition, Long> beginningOffsets = consumer.beginningOffsets(topicPartitions);
-        Map<TopicPartition, Long> endOffsets = consumer.endOffsets(topicPartitions);
+        Map<TopicPartition, Long> beginningOffsets;
+        Map<TopicPartition, Long> endOffsets;
+
+        try {
+            beginningOffsets = consumer.beginningOffsets(topicPartitions);
+            endOffsets = consumer.endOffsets(topicPartitions);
+        } catch (RetriableException e) {
+            LOG.error("Failed to get offsets from Kafka", e);
```
srdo: Consider lowering to WARN level, and write that we will retry on next metrics tick. I think this is only a concern if it is recurring, so error level seems a little strong.

external/storm-kafka-client/src/test/java/org/apache/storm/kafka/spout/KafkaSpoutSingleTopicTest.java
```diff
@@ -428,4 +427,33 @@ public void testOffsetMetrics() throws Exception {
...
+        assertEquals(offsetMetric.get(SingleTopicKafkaSpoutConfiguration.TOPIC+"/totalLatestCompletedOffset").longValue(), 0);
+        //totalSpoutLag = totalLatestTimeOffset-totalLatestCompletedOffset
+        assertEquals(offsetMetric.get(SingleTopicKafkaSpoutConfiguration.TOPIC+"/totalSpoutLag").longValue(), 10);
+
+        //Emit all messages and check that they are emitted. Ack the messages too
+        for (int i = 0; i < messageCount; i++) {
+            nextTuple_verifyEmitted_ack_resetCollector(i);
+        }
+
+        commitAndVerifyAllMessagesCommitted(messageCount);
```
OliverMD: @srdo Do you think it would make sense to not include this as the test above already tests this?

srdo: Is this bit necessary?

srdo: I think we can leave it out

srdo:
+1 assuming the test passes.

PR #3165, opened 2019-11-13
'kafkaSpoutConfig' can be 'protected'

jacobtolar:
Extending classes may need access to the spout configuration.

### DIFF
--- a/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
+++ b/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaSpout.java
@@ -74,7 +74,7 @@ public class KafkaSpout<K, V> extends BaseRichSpout {
     protected SpoutOutputCollector collector;
 
     // Kafka
-    private final KafkaSpoutConfig<K, V> kafkaSpoutConfig;
+    protected final KafkaSpoutConfig<K, V> kafkaSpoutConfig;
     private final ConsumerFactory<K, V> kafkaConsumerFactory;
     private final TopicAssigner topicAssigner;
     private transient Consumer<K, V> consumer;


srdo:
I'm not sure about this. The spout wasn't really written with subclassing in mind, and the majority of the implementation is private. I also don't think we want to guarantee that we won't break the internal implementation of the spout in minor/patch releases. If you have a use case, maybe we could add a hook somewhere instead?

jacobtolar:
@srdo Thanks for the thoughts. 

In our case, we just want to get at the record translator inside 'open' from the config. We actually just need its streams for our current use case. What we're doing is creating per-stream metrics in the spout. When acking is on, we produce a histogram of total time a given tuple spends in the topology (by stream) from appearing in the spout until it acks or fails.

What we have works, it always just feels silly keeping another reference to something already in the parent.

It sounds like you'd prefer something like an overridable hook called at the end of open() that passes in the config or list of streams or something?

srdo:
@jacobtolar No, I want to discourage subclassing entirely. I'd much rather encourage people to wrap the spout than to extend it, as that's less likely to break if we make changes internally in the spout.

What you're trying to do sounds very reasonable, but it sounds like you could do it by composition as well. So you'd have something like
```
class YourSpout extends BaseRichSpout {
private final KafkaSpout wrapped;
private final KafkaSpoutConfig config;

public YourSpout(KafkaSpoutConfig config) {
  wrapped = new KafkaSpout(config)
  this.config = config;
}
}
```

and then you'd provide implementations of the spout methods that call the wrapped methods, and also do your tracking. This 

If you need additional hooks for e.g. knowing when a tuple has been discarded by the spout, we could add them to https://github.com/apache/storm/blob/master/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/KafkaTupleListener.java

What do you think?

kishorvpatil:
@jacobtolar  can we close this PR?

jacobtolar:
Yes, it's fine. Our existing code has this working by extending the kafka spout but it's a little hacky. 
When we have time or a need to we'll change to what has been recommended here by @srdo . 

Thanks!

PR #3166, opened 2019-11-13
STORM-3536 Add Generic-resources.md

dandsager1:


srdo:
The doc looks good. I'm wondering if you need to add a link to https://github.com/apache/storm/blob/master/docs/index.md, or some other page? Otherwise there won't be anything linking to this document.

dandsager1:
Added link from index and fixed link from Generic-resources

docs/Generic-resources.md
```diff
@@ -0,0 +1,39 @@
...
+```
+Example of Usage:
+```
+   supervisor.resources.map: {"gpu.count" : 2.0}
+```
+
+
+### Generic Resources in UI
+
+![Storm Cluster UI](https://user-images.githubusercontent.com/50145295/68497812-06d1b280-021b-11ea-9f96-8946bf4193fc.png)
```
srdo: I think you can put this image in the /images directory instead, that way we avoid the site breaking if this image disappears.

srdo:
+1

PR #3167, opened 2019-11-15
STORM-3538 Add Meter for sendSupervisorAssignments exception

dandsager1:


storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
```diff
@@ -1654,6 +1657,10 @@ private AssignmentDistributionService getAssignmentsDistributer() {
         return assignmentsDistributer;
     }
 
+    private StormMetricsRegistry getMetricsRegistry() {
```
agresch: This seems redundant since it is private.  Any real reason for having this?

dandsager1: It's used to access class member in static methods in the same way nimbus.getAssignmentsDistributer().

I'm fine with using the member directly but then should do the same with getAssignmentsDistributer()

agresch: Thanks.  I missed that.

PR #3169, opened 2019-11-15
STORM-3540 fix Pacemaker connection issues

agresch:
"Exception occurred in Pacemaker" message was not logging cause properly until I changed the message for some reason.

Problem:

1) ThriftNettyClientCodec.initChannel() is called, which we map to PacemakerClient.channelReady().
2) We get a pacemaker message to write
3) We see the channel as ready and write/flush the message
4) The flush causes a NotYetConnected exception
5) This triggers a reconnect
6) The reconnect causes us to go to step 1) and we get stuck in a loop failing to talk to Pacemaker

The solution I think is more appropriate:

PacemakerClientHandler.channelActive() should be the one calling PacemakerClient.channelReady().

### DIFF
--- a/storm-client/src/jvm/org/apache/storm/pacemaker/codec/ThriftNettyClientCodec.java
+++ b/storm-client/src/jvm/org/apache/storm/pacemaker/codec/ThriftNettyClientCodec.java
@@ -72,7 +72,7 @@ public class ThriftNettyClientCodec extends ChannelInitializer<Channel> {
                 throw new RuntimeException(e);
             }
         } else {
-            client.channelReady(ch);
+            // no work for AuthMethod.NONE
         }
 
         pipeline.addLast("PacemakerClientHandler", new PacemakerClientHandler(client));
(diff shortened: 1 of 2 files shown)


storm-client/src/jvm/org/apache/storm/pacemaker/codec/ThriftNettyClientCodec.java
```diff
@@ -72,7 +72,7 @@ protected void initChannel(Channel ch) throws Exception {
                 throw new RuntimeException(e);
             }
         } else {
-            client.channelReady(ch);
+            // no work for AuthMethod.NONE
```
srdo: Consider removing the else block entirely

srdo:
+1, good catch

storm-client/src/jvm/org/apache/storm/pacemaker/codec/ThriftNettyClientCodec.java
agresch: My personal preference is to leave it so we know AuthMethod.NONE was not skipped by accident.

srdo: Ok, makes sense.

PR #3171, opened 2019-11-20
STORM-3543 Change task hook info objects to use basic for loops to reduce garbag…

cjljohnson:
A big drive was made in 2.0 to avoid iterators on the critical path to reduce garbage collection.

When task hooks are used, an iterator is used for each info object to pass it to all hooks.

As this is on the critical path when task hooks are used, the same reasoning should apply and the iterators be replaced with for loops.


srdo:
Good catch, +1. Please amend the commit so the issue number is included in the message.

cjljohnson:
Updated the commit message.

PR #3178, opened 2019-12-09
STORM-3511 Upgrade Thrift to 0.13.0

marjoral:


srdo:
I think you need to also run storm-client/gen_thrift.sh, using the upgraded Thrift version.

PR #3185, opened 2019-12-18
STORM-3557 allow health checks to pass on timeout and add meter to track

agresch:


storm-server/src/main/java/org/apache/storm/healthcheck/HealthChecker.java
```diff
@@ -40,7 +45,7 @@
     private static final String TIMEOUT = "timeout";
     private static final String FAILED_WITH_EXIT_CODE = "failed_with_exit_code";
 
-    public static int healthCheck(Map<String, Object> conf) {
+    public static int healthCheck(Map<String, Object> conf, Supervisor supervisor) {
```
Ethanlm: We only need `metricsRegistry` here instead of `Supervisor`.

storm-server/src/main/java/org/apache/storm/healthcheck/HealthChecker.java
```diff
@@ -66,10 +71,23 @@ public static int healthCheck(Map<String, Object> conf) {
...
             return 1;
+        } else if (results.contains(TIMEOUT)) {
+            LOG.warn("The supervisor healthchecks timedout!!!");
+            if (supervisor != null) {
+                Meter timeoutMeter = supervisor.getMetricsRegistry().getMeter(Constants.SUPERVISOR_HEALTH_CHECK_TIMEOUTS);
+                if (timeoutMeter != null) {
+                    timeoutMeter.mark();
+                }
+            }
+            Boolean failOnTimeouts = ObjectReader.getBoolean(conf.get(DaemonConfig.STORM_HEALTH_CHECK_FAIL_ON_TIMEOUTS), true);
```
Ethanlm: We can remove this since it's already in the `default.yaml` file to avoid duplicates.
 
Or remove the one in default.yaml

agresch:
@Ethanlm - made the changes you requested.

Ethanlm:
Restarted travis to see if it will pass

PR #3188, opened 2020-01-03
[STORM-3563] Fix broken link for apache-maven download and cache maven directory

Ethanlm:
1. maven-3.6.1 got removed from the mirror which breaks the travis build. Change to point to  archive.apache.org.

2. Cache apache-maven directory to avoid frequent download.


srdo:
+1

Ethanlm:
I will not wait for 24hrs since it's non-code change and it blocks every other merge.

PR #3189, opened 2020-01-08
[STORM-3494] use HdfsLoginUtil singleton to login to hdfs

Ethanlm:
Please refer to https://issues.apache.org/jira/browse/STORM-3494

Use HdfsLoginUtil singleton to login to hdfs exactly once for a process.

This will break backwards-compatibility. I didn't find a better way to resolve this. Suggestions are very welcome.

### DIFF
--- a/dev-tools/travis/travis-script.sh
+++ b/dev-tools/travis/travis-script.sh
@@ -43,7 +43,7 @@ elif [ "$2" == "External" ]
 then
   if [ "$TRAVIS_JDK_VERSION" == "openjdk11" ]
   then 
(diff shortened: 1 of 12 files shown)


agresch:
Could we assume if the deprecated blobstore configs are set, STORM_HDFS_LOGIN_PLUGIN should use this default implementation?  Not sure I like this, but would prevent breaking things?

Generally looks good to me.

Ethanlm:
We don't want to add hadoop as a dependency to storm itself. So the loginPlugin is in a separate module.  Hence to use this loginPlugin, users need to explicitly copy jar files to external-daemon otherwise there would be ClassNotFoundException.  So we can't use it as the default.

PR #3190, opened 2020-01-08
STORM-3544 Ability to find topologies with low worker uptime on UI 

agresch:
This is @dandsager1 's work to create a new page to see all topologies with the min worker uptime on one page.  I've updated the generated thrift.

We've had issues where nodes are bad or there were bugs were some topologies could not launch workers.  This page allows one place to quickly find topologies that may be having issues with nodes, rather than having to go to each topology page individually.

Screen Shot 2020-01-08 at 11 21 45 AM


Ethanlm:
Can we modify TopologySummary instead of adding a new structure?

agresch:
I think the reason was there was concern about the REST API fetching data about all the workers for all the topologies to get the min uptime.  By having a separate page, this only gets hit when specifically looking for this information.  Most of the time most users would not care.

If that is not a concern, I would agree moving it to Topology summary makes sense.

rzo1:
Feel free to re-open, if this is still a thing.

PR #3194, opened 2020-01-14
[STORM-3565] add support for adding dimensions to metric data points

RuiLi8080:
A dimensional metric collects information into multiple sub-categories (dimensions). 
Add an interface `IDimensionalMetric` extending `IMetric` in hope for supporting a better fine-grained metric reporting.
`DimensionalCountMetric` serves as an example for creating and using dimensional metric


kishorvpatil:
Having _DimensionalCountMetric_ is excluding all other IMetric implementations that would required Dimensional implementation.

storm-client/src/jvm/org/apache/storm/metric/api/IDimensionalMetric.java
```diff
@@ -0,0 +1,32 @@
...
+public interface IDimensionalMetric extends IMetric {
+
+    /**
+     * Dimensions is a collection of additional key-value metadata which is optional.
+     * It could be useful for further fine-grained measurement for a metric.
+     * It may also help better grouping in downstream metric monitoring system.
+     *
+     * @return a map of dimension name to a sub-metric
+     */
+    Map<Map<String, String>, CountMetric> getDimToMetric();   // dimensions : sub-metric
```
kishorvpatil: Using `CountMetric` makes this interface specific to `CountMetric`.

RuiLi8080: @kishorvpatil My mistake, thank you for pointing out. My intention was to set this to IMetric. When implementing, map value could be dynamically typed by implementing IMetric. Corrected.

RuiLi8080:
Re-worked on this PR. My thinking is to add a dimensions field in IMetricConsumer.DataPoint as an additional K-V metadata map. Data points are collected from metric measurement. The optional additional information could be helpful when downstream monitoring system want to organize or group those datapoints.

storm-client/src/jvm/org/apache/storm/metric/api/IMetric.java
```diff
@@ -12,16 +12,39 @@
...
      *     of calling this should be that the value is reset.
      */
     Object getValueAndReset();
+
+
+    /**
+     * Check whether this metric is carrying additional dimension map.
+     * @return a boolean value.
+     */
+    default boolean isDimensional() {
```
Ethanlm: Why do we need this function?

PR #3195, opened 2020-01-16
[STORM-3567] fix UI showing wrong resource info when topo is not scheduled

RuiLi8080:
Currently topology UI page is showing wrong resource request info if the topo is not scheduled.
This PR:
1. Fix the getTopologyPageInfo method to calculate the correct resource map even if they are not populated yet
2. Rename the component-level resource_map related keys in order to prevent name collision of mustache.render. Now even if the client failed to retrieve the data for resources, it will be left blank instead of getting same named value from topology level.

Example: this wc2 topo is not scheduled and the assigned on-heap mem is 0. Now the components are showing the right requested value.

Screen Shot 2020-01-16 at 2 32 57 PM

### DIFF
--- a/storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
+++ b/storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
@@ -4169,21 +4169,10 @@ public class Nimbus implements Iface, Shutdownable, DaemonCommon {
             }
 
             Map<String, Object> topoConf = Utils.merge(conf, common.topoConf);
-            Map<String, NormalizedResourceRequest> spoutResources = ResourceUtils.getSpoutsResources(topology, topoConf);
-            for (Entry<String, ComponentAggregateStats> entry : topoPageInfo.get_id_to_spout_agg_stats().entrySet()) {
-                CommonAggregateStats commonStats = entry.getValue().get_common_stats();
-                setResourcesDefaultIfNotSet(spoutResources, entry.getKey(), topoConf);
-                commonStats.set_resources_map(spoutResources.get(entry.getKey()).toNormalizedMap());
-            }
-            maybeAddPlaceholderSpoutAggStats(topoPageInfo, topology);
 
-            Map<String, NormalizedResourceRequest> boltResources = ResourceUtils.getBoltsResources(topology, topoConf);
-            for (Entry<String, ComponentAggregateStats> entry : topoPageInfo.get_id_to_bolt_agg_stats().entrySet()) {
-                CommonAggregateStats commonStats = entry.getValue().get_common_stats();
-                setResourcesDefaultIfNotSet(boltResources, entry.getKey(), topoConf);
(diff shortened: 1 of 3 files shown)


storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
```diff
@@ -4242,20 +4231,30 @@ public TopologyPageInfo getTopologyPageInfo(String topoId, String window, boolea
...
-     * Add placeholder AggStats allowing topology page to show components before AggStats are populated.
+     * If aggStats are not populated, compute common and component(spout) agg and create placeholder stat.
+     * This allow the topology page to show component spec even the topo is not scheduled.
+     * Otherwise, just fetch data from current topoPageInfo.
      *
      * @param topoPageInfo topology page info holding spout AggStats
      * @param topology     storm topology used to get spout names
      */
-    private void maybeAddPlaceholderSpoutAggStats(TopologyPageInfo topoPageInfo, StormTopology topology) {
+    private void addSpoutAggStats(TopologyPageInfo topoPageInfo, StormTopology topology, Map<String, Object> topoConf) {
```
Ethanlm: Nit: Java doc is missing @param topology topoConf

storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
```diff
@@ -4264,23 +4263,35 @@ private void maybeAddPlaceholderSpoutAggStats(TopologyPageInfo topoPageInfo, Sto
...
+     * If aggStats are not populated, compute common and component(bolt) agg and create placeholder stat.
+     * This allow the topology page to show component spec even the topo is not scheduled.
+     * Otherwise, just fetch data from current topoPageInfo.
      *
      * @param topoPageInfo  topology page info holding bolt AggStats
      * @param topology      storm topology used to get bolt names
      * @param includeSys    whether to show system bolts
      */
-    private void maybeAddPlaceholderBoltAggStats(TopologyPageInfo topoPageInfo, StormTopology topology, boolean includeSys) {
+    private void addBoltAggStats(TopologyPageInfo topoPageInfo, StormTopology topology,
```
Ethanlm: nit: missing javadoc @param topoConf

RuiLi8080: Corrected both and force-push.

RuiLi8080:
Talked to @Ethanlm offline that mustache name collision won't happen in this case with the getTopologyPageInfo fix. Since this bug fix will probably go to 2.1 branch as well, we want to minimize the backward compatibility impact. So I revert the html part of changes. Retested and it works the same.

Ethanlm:
There is merge conflict against 2.1.x-branch so I couldn't cherry-pick the commits. Can you please file a separate PR against 2.1.x-branch? 
Mostly the conflict comes from the "UI showing genericResource" feature introduced in master branch.

PR #3196, opened 2020-01-17
[STORM-3568] prevent user from changing log level with empty logger name from topo UI

RuiLi8080:
A minor fix to topo UI. This change alert user when they try to change log level on topology page without providing a logger name.
Currently user can do this and get a 500 Server Error at the bottom which could be confusing.

Screen Shot 2020-01-17 at 3 26 22 PM


RuiLi8080:
Thanks. That's actually @agresch's catch.

Ethanlm:
Travis failure is not related. Merging this.

PR #3197, opened 2020-01-22
MINOR: The expected timoout of 5seconds might fail in launching conta…

kishorvpatil:
…iner and heartbeats.
The increase in Heartbeat timeouts guarantees some slowness in VMs


Ethanlm:
I am not sure about this. The `firstHbTimeoutMs` is `120_000ms`, which is used when the container is launched and heartbeat for the first time.

kishorvpatil:
This patch is not fixing the issue on all cases, so closing it.

PR #3198, opened 2020-01-23
[STORM-3568] prevent user from changing log level with empty logger name from topo UI

RuiLi8080:
Cheery-pick from [#3196](https://github.com/apache/storm/pull/3195)


Ethanlm:
Travis failure is not related.

PR #3199, opened 2020-01-23
[STORM-3567] fix UI showing wrong resource info when topo is not sche…

RuiLi8080:
…duled


Ethanlm:
Note that the explanation is in https://github.com/apache/storm/pull/3195. This PR is for 2.1.x-branch

PR #3202, opened 2020-01-31
STORM-3572 prevent NPEs when doing visualization with restarting exec…

agresch:
…utors

### DIFF
--- a/storm-server/src/main/java/org/apache/storm/stats/StatsUtil.java
+++ b/storm-server/src/main/java/org/apache/storm/stats/StatsUtil.java
@@ -830,6 +830,9 @@ public class StatsUtil {
         Map<String, Map<K, Double>> ret = new HashMap<>();
 
         Map<String, Map<K, List>> expands = expandAveragesSeq(avgSeq, countSeq);
+        if (expands == null) {
+            return ret;
+        }
         for (Map.Entry<String, Map<K, List>> entry : expands.entrySet()) {
             String k = entry.getKey();
 
@@ -2304,6 +2307,9 @@ public class StatsUtil {
             } else {
                 initVal = mergeWithAddPair(initVal, expandAverages(avg, count));
             }
+        }
+        if (initVal == null) {
+            initVal = new HashMap<>();
         }
         return initVal;
     }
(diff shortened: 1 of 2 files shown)


storm-webapp/src/main/java/org/apache/storm/daemon/ui/resources/StormApiResource.java
```diff
@@ -299,6 +299,9 @@ public Response getTopology(@PathParam("id") String id,
                     ),
                     callback
             );
+        } catch (RuntimeException e) {
```
Ethanlm: why do we need this

storm-server/src/main/java/org/apache/storm/stats/StatsUtil.java
```diff
@@ -830,6 +830,9 @@ private static TopologyPageInfo postAggregateTopoStats(Map task2comp, Map exec2n
         Map<String, Map<K, Double>> ret = new HashMap<>();
 
         Map<String, Map<K, List>> expands = expandAveragesSeq(avgSeq, countSeq);
+        if (expands == null) {
```
Ethanlm: this is not needed since `expandAveragesSeq` will not return null

storm-webapp/src/main/java/org/apache/storm/daemon/ui/resources/StormApiResource.java
agresch: This provided the callstack that let me debug this.  If a different error occurs, I would like to be able to see the callstack.

storm-server/src/main/java/org/apache/storm/stats/StatsUtil.java
agresch: I can remove if desired. I would prefer to be safe.

storm-webapp/src/main/java/org/apache/storm/daemon/ui/resources/StormApiResource.java
Ethanlm: Okay. 
You might want to update the log message since this function is not only about getting visualization

storm-server/src/main/java/org/apache/storm/stats/StatsUtil.java
Ethanlm: I am fine with it

storm-webapp/src/main/java/org/apache/storm/daemon/ui/resources/StormApiResource.java
agresch: Need to fix this file.  Thanks for catching

agresch:
@Ethanlm - force push changed just StormApiResource.java to address your find.

PR #3203, opened 2020-02-06
[STORM-3574] Rebalance accepts config change affecting system component.

bipinprasad:
Example when topology.metrics.consumer.register is specified to add metrics consumer.
 Change rebalance() to start new tasks by calling computeExecutors and consequently
 StormCommon.stormTaskInfo and StormCommon.systemTopology.

### DIFF
--- a/storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
+++ b/storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
@@ -3346,6 +3346,12 @@ public class Nimbus implements Iface, Shutdownable, DaemonCommon {
             if (subject != null) {
                 options.set_principal(subject.getPrincipals().iterator().next().getName());
             }
+            String topoId = toTopoId(topoName);
+            StormTopology stormTopology = tryReadTopology(topoId, topoCache);
+            StormBase base = new StormBase();
+            base.set_name(topoName);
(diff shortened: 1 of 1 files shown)


storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
```diff
@@ -3346,6 +3346,12 @@ public void rebalance(String topoName, RebalanceOptions options)
             if (subject != null) {
                 options.set_principal(subject.getPrincipals().iterator().next().getName());
             }
+            String topoId = toTopoId(topoName);
+            StormTopology stormTopology = tryReadTopology(topoId, topoCache);
+            StormBase base = new StormBase();
+            base.set_name(topoName);
+            idToExecutors.getAndUpdate(new Assoc<>(topoId,
+                    new HashSet<>(computeExecutors(topoId, base, topoConf, stormTopology))));
```
Ethanlm: When rebalance happens, 
the code here  https://github.com/apache/storm/blob/master/storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java#L1749 will remove the entry and let it recompute. 
So I think we don't this?

bipinprasad:
Resolved by STORM-3580

Ethanlm:
Why is it related to  STORM-3580?

PR #3206, opened 2020-02-07
[STORM-3578] ClientAuthUtils.insertWorkerTokens removed tokens incorrectly

Ethanlm:
Explained in the JIRA.


Ethanlm:
updated the JIRA number

PR #3207, opened 2020-02-10
[STORM-3580] Processs -c option to storm.py for Rebalance

bipinprasad:
storm.py passes utf encoded "-c" parameters as csv jvm property "storm.options". This system property is not being used in rebalance client command when passing information to Nimbus. Change Rebalance to use this system property.

### DIFF
--- a/storm-core/src/jvm/org/apache/storm/command/Rebalance.java
+++ b/storm-core/src/jvm/org/apache/storm/command/Rebalance.java
@@ -20,6 +20,7 @@ import org.apache.storm.generated.Nimbus;
 import org.apache.storm.generated.RebalanceOptions;
 import org.apache.storm.utils.NimbusClient;
 import org.apache.storm.utils.Utils;
+import org.json.simple.JSONObject;
 import org.json.simple.JSONValue;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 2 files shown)


storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
```diff
@@ -3356,6 +3356,12 @@ public void rebalance(String topoName, RebalanceOptions options)
             if (subject != null) {
                 options.set_principal(subject.getPrincipals().iterator().next().getName());
             }
+            String topoId = toTopoId(topoName);
```
Ethanlm: I think the change here should not be part of this PR

bipinprasad: Rebalance will not work without the change below. Should the change to Nimus.java be part of a different Jira then?

bipinprasad: Reverted Nimbus.java rebalance() changes.

storm-core/src/jvm/org/apache/storm/command/Rebalance.java
```diff
@@ -57,6 +58,15 @@ public static void main(String[] args) throws Exception {
         }
 
         Map<String, Object> confOverrides = (Map<String, Object>) cl.get("t");
+        Map<String, Object> jvmOpts = Utils.readCommandLineOpts(); // values in -Dstorm.options (originally -c in storm.py)
```
Ethanlm: This is mostly a question:  Do we want to include default `storm.yaml` into this too? 

for example: 
https://github.com/apache/storm/blob/master/storm-core/src/jvm/org/apache/storm/command/ShellSubmission.java#L35 
this reads `storm.yaml` and commandLineOpts

storm-server/src/main/java/org/apache/storm/daemon/nimbus/Nimbus.java
Ethanlm: Thanks

storm-core/src/jvm/org/apache/storm/command/Rebalance.java
Ethanlm: Offline discussion with Bipin :

> I don’t think we want that - that would be a reset+. Since rebalance should an incremental change to the currently running Topo.

PR #3210, opened 2020-02-18
[STORM-3583] Supervisor Logs and ignores if blob is not found

kishorvpatil:
Avoid supervisor restart is blob is not found for some reason. if any other IOExceptions such as misconfiguration etc causing IOException the supervisor should shutdown.


Ethanlm:
Is there a JIRA for this?
Also https://github.com/apache/storm/blob/master/storm-server/src/main/java/org/apache/storm/localizer/AsyncLocalizer.java#L462-L463
this catches exception and wrap it as a IOException. 

Will `releaseSlotFor` ever throw a FileNotFoundException?

kishorvpatil:
> Is there a JIRA for this?
STORM-3583

PR #3211, opened 2020-02-19
[STORM-3584] Support getting version info from a wildcard classpath entry

Ethanlm:
https://issues.apache.org/jira/browse/STORM-3584

Current VersionInfo.getFromClasspath method only tries to get version info from the specific property file under a directory, or jar/zip files. It should support a classpath entry like `/<parent-path>/*`.

### DIFF
--- a/storm-client/src/jvm/org/apache/storm/utils/VersionInfo.java
+++ b/storm-client/src/jvm/org/apache/storm/utils/VersionInfo.java
@@ -172,7 +172,10 @@ public final class VersionInfo {
                 try {
                     Path parent = p.getParent();
                     List<String> children = new ArrayList<>();
-                    Files.list(parent).forEach(path -> children.add(path.toString()));
+                    Files.list(parent)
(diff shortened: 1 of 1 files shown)


storm-client/src/jvm/org/apache/storm/utils/VersionInfo.java
```diff
@@ -172,7 +172,10 @@ private static IVersionInfo getFromClasspath(List<String> classpath, final Strin
                 try {
                     Path parent = p.getParent();
                     List<String> children = new ArrayList<>();
-                    Files.list(parent).forEach(path -> children.add(path.toString()));
+                    Files.list(parent)
+                        //avoid infinite recursion
```
Ethanlm: infinite recursion can happen if there is a file named `*` under `/<parent-path>/` directory

kishorvpatil: Good finding.

PR #3212, opened 2020-02-21
[STORM-3587] Allow Scheduler futureTask to gracefully exit before TimeoutException.

bipinprasad:
ResourceAwareScheduler creates a FutureTask with timeout specified in DaemonConfig.

ConstraintSolverStrategy uses the the another configuration variable to determine when to terminate its effort. Limit this value so that it terminates at most slightly before TimeoutException. This graceful exit allows result (and its error) to be available in ResourceAwareScheduler.

### DIFF
--- a/storm-server/src/main/java/org/apache/storm/scheduler/resource/ResourceAwareScheduler.java
+++ b/storm-server/src/main/java/org/apache/storm/scheduler/resource/ResourceAwareScheduler.java
@@ -166,7 +166,7 @@ public class ResourceAwareScheduler implements IScheduler {
                     Future<SchedulingResult> schedulingFuture = backgroundScheduling.submit(
                         () -> finalRasStrategy.schedule(toSchedule, td));
                     try {
-                        result = schedulingFuture.get(schedulingTimeoutSeconds, TimeUnit.SECONDS);
+                        result = schedulingFuture.get(schedulingTimeoutSeconds + 1, TimeUnit.SECONDS);
                     } catch (TimeoutException te) {
                         markFailedTopology(topologySubmitter, cluster, td, "Scheduling took too long for "
                                 + td.getId() + " using strategy " + rasStrategy.getClass().getName() + " timeout after "
@@ -226,7 +226,17 @@ public class ResourceAwareScheduler implements IScheduler {
                         //Only place we fall though to do the loop over again...
                     } else { //Any other failure result
                         //The assumption is that the strategy set the status...
-                        topologySubmitter.markTopoUnsuccess(td, cluster);
+                        String msg = "";
+                        if (result.getStatus() != null) {
+                            msg += result.getStatus().name() + ":";
+                        }
+                        if (result.getErrorMessage() != null && !result.getErrorMessage().isEmpty()) {
+                            msg += result.getErrorMessage();
+                        }
+                        if (result.getMessage() != null && !result.getMessage().isEmpty()) {
+                            msg += " " + result.getMessage();
+                        }
+                        topologySubmitter.markTopoUnsuccess(td, cluster, msg);
                         return;
                     }
                 }
--- a/storm-server/src/main/java/org/apache/storm/scheduler/resource/ResourceAwareScheduler.java
+++ b/storm-server/src/main/java/org/apache/storm/scheduler/resource/ResourceAwareScheduler.java
@@ -166,7 +166,7 @@ public class ResourceAwareScheduler implements IScheduler {
                     Future<SchedulingResult> schedulingFuture = backgroundScheduling.submit(
                         () -> finalRasStrategy.schedule(toSchedule, td));
                     try {
-                        result = schedulingFuture.get(schedulingTimeoutSeconds + 1, TimeUnit.SECONDS);
+                        result = schedulingFuture.get(schedulingTimeoutSeconds, TimeUnit.SECONDS);
                     } catch (TimeoutException te) {
                         markFailedTopology(topologySubmitter, cluster, td, "Scheduling took too long for "
                                 + td.getId() + " using strategy " + rasStrategy.getClass().getName() + " timeout after "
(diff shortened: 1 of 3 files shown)


storm-server/src/main/java/org/apache/storm/scheduler/resource/ResourceAwareScheduler.java
```diff
@@ -166,7 +166,7 @@ private void scheduleTopology(TopologyDetails td, Cluster cluster, final User to
                     Future<SchedulingResult> schedulingFuture = backgroundScheduling.submit(
                         () -> finalRasStrategy.schedule(toSchedule, td));
                     try {
-                        result = schedulingFuture.get(schedulingTimeoutSeconds, TimeUnit.SECONDS);
+                        result = schedulingFuture.get(schedulingTimeoutSeconds + 1, TimeUnit.SECONDS);
```
Ethanlm: Why do we need ` + 1`

storm-server/src/main/java/org/apache/storm/scheduler/resource/User.java
```diff
@@ -79,15 +79,18 @@ public String getId() {
         return ret;
     }
 
-    public void markTopoUnsuccess(TopologyDetails topo, Cluster cluster) {
+    public void markTopoUnsuccess(TopologyDetails topo, Cluster cluster, String msg) {
         unsuccess.add(topo);
         if (cluster != null) {
-            cluster.setStatus(topo.getId(), "Scheduling Attempted but topology is invalid");
+            if (msg == null) {
+                msg = "Scheduling Attempted but topology is invalid";
```
Ethanlm: failed to schedule does not necessarily `the topology is invalid`.

storm-server/src/main/java/org/apache/storm/scheduler/resource/ResourceAwareScheduler.java
bipinprasad: The timeout is check in strategy to determine when terminate. However, if the future task is killed at or around the same time - this results in a TimeOut exception on the task and the result is not propagated back to the caller. +1 gives an additional second before the FutureTask is rudely terminated and allows the result to be returned and examined for the actual message in the result.

storm-server/src/main/java/org/apache/storm/scheduler/resource/User.java
bipinprasad: That message is a generic default - same as prior default. I believe there is one other caller to this method.

storm-server/src/main/java/org/apache/storm/scheduler/resource/ResourceAwareScheduler.java
Ethanlm: I think we should discuss about whether we want to do this or not. Essentially this applies to every timeout. If our "timeout=x second" in Storm means things will fail/time out at `x+1` seconds, then everywhere with timeout configs, we need `+1` to make the semantic consistent. I think this is not very necessary

storm-server/src/main/java/org/apache/storm/scheduler/resource/ResourceAwareScheduler.java
```diff
@@ -226,7 +226,17 @@ private void scheduleTopology(TopologyDetails td, Cluster cluster, final User to
                         //Only place we fall though to do the loop over again...
                     } else { //Any other failure result
                         //The assumption is that the strategy set the status...
-                        topologySubmitter.markTopoUnsuccess(td, cluster);
+                        String msg = "";
```
Ethanlm: This can be replaced by result.toString()

Ethanlm: This can be replaced by result.toString()

storm-server/src/main/java/org/apache/storm/scheduler/resource/ResourceAwareScheduler.java
bipinprasad: Should not apply to every timeout. If the scheduled task is a cooperating task that is also using the same timeout to determine when to stop, then we have this situation where scheduler is interrupting the FutureTask before the task is allowed to gracefully exit and return a result.

If the scheduled task is a non-cooperating task (i.e. .not using the timeout), then it is fine to use the specified number.

storm-server/src/main/java/org/apache/storm/scheduler/resource/ResourceAwareScheduler.java
bipinprasad: Changed

storm-server/src/main/java/org/apache/storm/scheduler/resource/ResourceAwareScheduler.java
Ethanlm: Probably I misunderstood. Could you point out to me where we are using `SCHEDULING_TIMEOUT_SECONDS_PER_TOPOLOGY` too?  It looks to me here is the only place the schedulingTimeoutSeconds is being used. I don't see cooperations.

bipinprasad: I see your point. 

When I think about cooperating process, is that the strategy is a time-bound task and part of the same code base running in the same JVM - so there never should be a a need to kill a FutureTask except as a precaution against bug introduced inadvertently.

Current ConstraintSolver uses a different (but redundant) config variable for time limit - which is accidentally set to the same default value.

In light of this - it may be better to explicitly pass "max" time limit to the constraint solver. And then determine how much the margin needs to be, and then add the margin to the FutureTask timeout. Note that this extra margin (and the timeout exception should only happen in exceptional  case when there is bug in ConstraintSolver. Normally it will/should exit by the timeout duration.
And the result should be available.

Ethanlm: It makes sense to me. Thanks.

Ethanlm: Can you please add some brief comments about the purpose of `+1` so future me will not be surprised when I come back to this. Thanks

bipinprasad: Pushed the change into ConstrainstSolverStrategy where there is millisecond granularity and avoid hitting the ceiling. Removed +1 from ResourceAwareStrategy.

